PR #13353, opened 2023-11-14
AsyncHtmlLoader update due to fetching errors in not available webpages

pedro-inf-custodio:
- **Description:** When running the `qa_chain` with the `WebResearchRetriever` class as a retriever, the code would throw an error even with other possible/valid sources. This would happen when there was no access/connection to an URL.
  - **Issue:** None,
  - **Dependencies:** None,

Please make sure your PR is passing linting and testing before submitting. Run `make format`, `make lint` and `make test` to check this locally.

See contribution guidelines for more information on how to write/run tests, lint, etc: 
https://github.com/langchain-ai/langchain/blob/master/.github/CONTRIBUTING.md

If you're adding a new integration, please include:
  1. a test for the integration, preferably unit tests that do not rely on network access,
  2. an example notebook showing its use. It lives in `docs/extras` directory.

If no one reviews your PR within a few days, please @-mention one of @baskaryan, @eyurtsev, @hwchase17.

PR #13401, opened 2023-11-15
WebResearchRetriever error handling in urls with connection error

pedro-inf-custodio:
- **Description:** Added a method `fetch_valid_documents` to `WebResearchRetriever` class that will test the connection for every url in `new_urls` and remove those that raise a `ConnectionError`.
  - **Issue:** [Previous PR](https://github.com/langchain-ai/langchain/pull/13353),
  - **Dependencies:** None,
  - **Tag maintainer:** @efriis 

Please make sure your PR is passing linting and testing before submitting. Run `make format`, `make lint` and `make test` to check this locally.

See contribution guidelines for more information on how to write/run tests, lint, etc: 
https://github.com/langchain-ai/langchain/blob/master/.github/CONTRIBUTING.md

If you're adding a new integration, please include:
  2. an example notebook showing its use. It lives in `docs/extras` directory.


libs/langchain/langchain/retrievers/web_research.py
```diff
@@ -154,6 +154,17 @@ def search_tool(self, query: str, num_search_results: int = 1) -> List[dict]:
         query_clean = self.clean_search_query(query)
         result = self.search.results(query_clean, num_search_results)
         return result
+    
+    def fetch_valid_documents(self, url_list: List[str]) -> List[str]:
```
efriis, line 158:
```suggestion
    def _fetch_valid_documents(self, url_list: List[str]) -> List[str]:
```
make it semiprivate

libs/langchain/langchain/retrievers/web_research.py
```diff
@@ -195,6 +206,9 @@ def _get_relevant_documents(
         # Check for any new urls that we have not processed
         new_urls = list(urls.difference(self.url_database))
 
+        # Removes urls with connection errors from list
+        new_urls = self.fetch_valid_documents(new_urls)
```
efriis, line 210: this will significantly slow down execution. 

I think the best option would be a combination approach of what you did here and in #13353, where you add a config param to `AsyncHtmlLoader` that defaults to False of `ignore_load_errors`, and you construct the loader here with that set to true.

What do you think?

libs/langchain/langchain/retrievers/web_research.py
```diff
@@ -154,6 +154,17 @@ def search_tool(self, query: str, num_search_results: int = 1) -> List[dict]:
         query_clean = self.clean_search_query(query)
         result = self.search.results(query_clean, num_search_results)
         return result
+    
+    def fetch_valid_documents(self, url_list: List[str]) -> List[str]:
+        clean_urls = []
+        for url in url_list:
+            try:
+                Requests().get(url).status_code
```
efriis, line 162: let's avoid adding extra requests to the path if we can

libs/langchain/langchain/retrievers/web_research.py
pedro-inf-custodio: It does make sense to me. I've done the `try ... except ...` in the `_scrape` method to avoid putting a lot of logic in the `load` method.

efriis (approved):
Thanks @pedro-inf-custodio

PR #13322, opened 2023-11-14
Added llama2-13b-chat-v1 on bedrock

sungeuns:
- **Description:** Added `llama2-13b-chat-v1` for Bedrock client
  - **Issue:** #13316
  - **Dependencies:** None

Please make sure your PR is passing linting and testing before submitting. Run `make format`, `make lint` and `make test` to check this locally.

See contribution guidelines for more information on how to write/run tests, lint, etc: 
https://github.com/langchain-ai/langchain/blob/master/.github/CONTRIBUTING.md

If you're adding a new integration, please include:
  1. a test for the integration, preferably unit tests that do not rely on network access,
  2. an example notebook showing its use. It lives in `docs/extras` directory.

If no one reviews your PR within a few days, please @-mention one of @baskaryan, @eyurtsev, @hwchase17.

Issue #13316, opened 2023-11-14
Add support for Bedrock Llama 2 13b model (meta.llama2-13b-chat-v1)


mirodrr:
### System Info

I get following error just by adding model parameters to existing code that works with other models 

"Malformed input request: 2 schema violations found, please reformat your input and try again."

```
model_name = "meta.llama2-13b-chat-v1"
model_kwargs = {
"max_gen_len": 512,
"temperature": 0.2,
 "top_p": 0.9
}
bedrock_boto = boto3.client("bedrock-runtime", "us-east-1")
bedrock_llm = Bedrock(model_id=model_name, client=bedrock_boto,model_kwargs=model_kwargs)
bedrock_llm("Hello!")
```

### Who can help?

@hwchase17
@agola11

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

1. Attempt to call the new llama2 bedrock model like so:

```
model_name = "meta.llama2-13b-chat-v1"
model_kwargs = {
"max_gen_len": 512,
"temperature": 0.2,
 "top_p": 0.9
}
bedrock_boto = boto3.client("bedrock-runtime", "us-east-1")
bedrock_llm = Bedrock(model_id=model_name, client=bedrock_boto,model_kwargs=model_kwargs)
bedrock_llm("Hello!")
```

### Expected behavior

The Bedrock class would work successfully as it does for other BedRock models

CarloNicolini:
Same error here with langchain 0.0.336.

The DOSU bot is confusing and wrong. Solution is simpler, there is a reason why you get the 2-schema validation errors. 
If you look at the body of the `Bedrock._prepare_input_and_invoke` the cases of 'cohere', 'anthropic', and other providers are **manually set**. 
There is the need to manually include the 'meta' provider by adding all the parts of the code containing match cases with the provider 'meta'.
In practice in the `body` one should modify the `inputText` key renaming it into `prompt`.

PR #13403, opened 2023-11-15
Add llama2-13b-chat-v1 support to `chat_models.BedrockChat`

WouterDurnez:
Hi 👋 We are working with Llama2 on Bedrock, and would like to add it to Langchain. We saw a [pull request](https://github.com/langchain-ai/langchain/pull/13322) to add it to the `llm.Bedrock` class, but since it concerns a chat model, we would like to add it to `BedrockChat` as well.

  - **Description:** Add support for Llama2 to `BedrockChat` in `chat_models`  
  - **Issue:** the issue # it fixes (if applicable) [#13316](https://github.com/langchain-ai/langchain/issues/13316)
  - **Dependencies:** any dependencies required for this change `None`
  - **Tag maintainer:** /
  - **Twitter handle:** `@SimonBockaert @WouterDurnez`


3coins (approved):
@WouterDurnez 
Thanks for making this change. LGTM!

baskaryan:
thanks @WouterDurnez!

libs/langchain/langchain/chat_models/bedrock.py
```diff
@@ -26,6 +27,8 @@ def convert_messages_to_prompt(
     ) -> str:
         if provider == "anthropic":
             prompt = convert_messages_to_prompt_anthropic(messages=messages)
```
pressoholics, line 29: @baskaryan @WouterDurnez If using a bedrock anthropic model this always resolves to NotImplementedError due to incorrect implementation of If statement. This should be elseif or immediately return the prompt value once match with anthropic. This is currently an breaking change.

WouterDurnez: 🙈 https://github.com/langchain-ai/langchain/pull/13684 --> added a fix

WouterDurnez: 🙈 Sorry, missed that. Added a fix but then saw the PR above. Thx!

PR #13404, opened 2023-11-15
add azure ai data document loader

samuel100:
This PR adds an "Azure AI data" document loader, which allows Azure AI users to load their registered data assets as a document object in langchain.


baskaryan:
updated to add lazy loading support, let me know if it looks good to you @samuel100!

hwchase17 (approved):
seems pretty reasonable!

PR #13405, opened 2023-11-15
Passthrough kwargs in runnable lambda

nfcampos:


efriis (approved):

libs/langchain/langchain/schema/runnable/base.py
```diff
@@ -2509,6 +2513,7 @@ def invoke(
                 self._invoke,
                 input,
                 self._config(config, self.func),
+                **kwargs,
```
eyurtsev, line 2516: should we update the signature of `invoke` in Runnable? it'll probably catch a few other places where kwargs are missing?

eyurtsev (approved):
looks good

PR #13408, opened 2023-11-15
RunnableLambda: create afunc instance from func when not provided

### LINKED ISSUE #13407: RunnableLambda: returned runnable called synchronously when using ainvoke
### System Info

langchain on master branch

### Who can help?

_No response_

### Information

- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Memory
- [ ] Tools / Toolkits
- [X] Chains
- [ ] Callbacks/Tracing
- [X] Async

### Reproduction

```
from langchain.schema.runnable import RunnableLambda

import asyncio


def idchain_sync(__input):
    print(f'sync chain call: {__input}')
    return __input

async def idchain_async(__input):
    print(f'async chain call: {__input}')
    return __input

idchain = RunnableLambda(func=idchain_sync,afunc=idchain_async)




def func(__input):
    return idchain


asyncio.run(RunnableLambda(func).ainvoke('toto'))
#printss 'sync chain call: toto' instead of 'async chain call: toto'                                                                                                                                                  

```

### Expected behavior

LCEL's route can cause chains to be silently run synchronously, while the user uses ainvoke...

When calling a RunnableLambda A returning a chain B with ainvoke, we would expect the new chain B to be called with ainvoke;
However, if the function provided to RunnableLambda A is not async, then the chain B will be called with invoke, silently causing all the rest of the chain to be called synchronously.

nbondoux:
Fixes #13407.

This workaround consists in letting the RunnableLambda create its self.afunc from its self.func when self.afunc is not provided; the change has no dependency.


efriis (commented):
Do you have an example of a RunnableLambda that will have different `invoke` or `ainvoke` output from this?

I believe `ainvoke` works already by running the default `asyncio` executor on the lambda's `func` attribute, so I think this change will not be needed.

efriis:
I'll close this for now, and let me know if there's some functionality the current runnable-default solution is missing!

nbondoux:
> I'll close this for now, and let me know if there's some functionality the current runnable-default solution is missing!

The example is provided in issue #13407 ; ainvoke will run invoke, and thus starts a synchronous flow on any chain returned by the RunnableLambda. It seems very problematic to me as this behaviour is silent and, I would say, unexpected.

efriis:
Ah I see the issue. I misunderstood the case this was handling. Appreciate the explanation! Just brain dumping:

If you call `ainvoke` on a RunnableLambda that generates another RunnableLambda, that lambda will end up getting synchronously invoked in the asyncio event loop logic.

This PR solves this problem by bypassing that delegation, generating an `afunc` that does the same thing, which can then be `ainvoke`'d in the proper way.

Reopening, and going to think about if there's a good way of doing this without setting `afunc` on initialization, maybe by computing afunc within `ainvoke` if the attribute isn't set.

efriis:
Sorry got busy. Great change @nbondoux ! Outlined a preferred solution here, and will come back to this tomorrow. If you have some extra cycles, here's my todo list for then :)

- get it passing lint
- write unit test(s)

Appreciate your explanation above!

nbondoux:
@efriis : I did one small fix on your code, linted, added a unit test. However, test_stream_log_retriever still has to be fixed as its outputs are now disordered (which is certainly a proof that my fix is useful).

efriis (approved):
Thanks @nbondoux ! Fixed the test

nbondoux:
@efriis : I merged the origin/master with refactor, I hope I did the right things !

nbondoux:
@efriis pinging for merge if it is fine with you ! Note that test_stream_log_retriever was also rewritten on main branch; I kept the new version from main branch on this PR, but maybe I should have preferred yours.

yours: https://github.com/langchain-ai/langchain/blob/e9fd64c9480ececd5f6672f7f7046075bfe768fd/libs/langchain/tests/unit_tests/schema/runnable/test_runnable.py#L2154
main: https://github.com/langchain-ai/langchain/blob/686162670e2fe15fb906999da84d36a273b2f25e/libs/core/tests/unit_tests/runnables/test_runnable.py#L2118

nfcampos (approved):
Thanks!

PR #13413, opened 2023-11-15
Fix Runnable Lambda Afunc Repr

hinthornw:
Otherwise, you get an error when using async functions.

h/t to Chris Ruppelt


Cruppelt:
Thanks!

hinthornw:
need to run syrupy

eyurtsev (approved):

PR #13418, opened 2023-11-15
Runnable with message history

baskaryan:
Screenshot 2023-11-15 at 1 12 20 PM

https://dev.smith.langchain.com/public/c1c06afb-2e4f-4b53-bf37-6926003e7fa9/r


libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,211 @@
...
+            elif "output" in run.outputs:
+                output_messages = load(run.outputs["output"])
+                if isinstance(output_messages, BaseMessage):
+                    output_messages = [output_messages]
+            else:
+                raise ValueError(
+                    f"Output stored in run as dict but cannot extract messages: "
+                    f"{run.outputs}"
+                )
+        elif isinstance(run.outputs, BaseMessage):
```
hinthornw, line 180: These branches would never be reached. `run.outputs` is always a dict

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+    def get_input_schema(
+        self, config: Optional[RunnableConfig] = None
+    ) -> Type[BaseModel]:
+        super_schema = super().get_input_schema(config)
+        if super_schema.__custom_root_type__ is not None:
+            from langchain.schema.messages import BaseMessage
+
+            fields: Dict = {}
+            if self.input_messages_key:
+                fields[self.input_messages_key] = (str, ...)
```
baskaryan, line 152: should this be Union[str, BaseMessage, Sequence[BaseMessage]] to be safer?

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+    RunnableWithMessageHistory must always be called with a config that contains
+        {"configurable": {"thread_id": "<THREAD_ID>"}} and can optionally include
+        {"configurable": {"thread_id": "<THREAD_ID>", "user_id": "<USER_ID>"}}.
+
+    Example:
+        .. code-block:: python
+
+            from langchain.schema.prompts import ChatPromptTemplate, MessagesPlaceholder
+            from langchain.chat_models import ChatAnthropic
+            from langchain.schema.runnables import RunnableWithMessageHistory
```
eyurtsev, line 50: code block is incomplete. Could we add some more example code?

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+
+if TYPE_CHECKING:
+    from langchain.callbacks.tracers.schemas import Run
+    from langchain.schema.messages import BaseMessage
+
+MessagesOrDictWithMessages = Union[Sequence["BaseMessage"], Dict[str, Any]]
+
+
+class RunnableWithMessageHistory(RunnableBindingBase):
+    """
```
eyurtsev, line 35: The only way I will be happy is if we don't have a new line here. I'll probably be happy with a PS5 as well though (if i'm honest).

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+            output_messages_key: Must be specified if the base runnable returns a dict
+                as output.
+            message_history_key: Must be specified if the base runnable accepts a dict
+                as input and expects a separate key for historical messages.
+            **kwargs: Arbitrary additional kwargs to pass to parent class
+                ``RunnableBindingBase`` init.
+        """
+        history_chain: Runnable = RunnableLambda(
+            self._enter_history, self._aenter_history
+        ).with_config(run_name="load_history")
```
eyurtsev, line 103: nice touch with naming the lambda

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+        messages_key = message_history_key or input_messages_key
+        if messages_key:
+            history_chain = RunnablePassthrough.assign(
+                **{messages_key: history_chain}
+            ).with_config(run_name="insert_history")
```
eyurtsev, line 108: `insert` -> `update` ?

libs/langchain/tests/unit_tests/schema/runnable/test_history.py
```diff
@@ -0,0 +1,219 @@
...
+def test_input_dict() -> None:
+    runnable = RunnableLambda(
+        lambda input: "you said: "
+        + "\n".join(
+            str(m.content) for m in input["messages"] if isinstance(m, HumanMessage)
+        )
+    )
+    factory = get_factory()
+    with_history = runnable.with_message_history(factory, input_messages_key="messages")
+    config: RunnableConfig = {"configurable": {"thread_id": datetime.datetime.now()}}
```
eyurtsev, line 44: could we avoid using datetime.datetime.now() for a thread_id? can we hard-code the thread id to reasonable ids? like 1, 2, 3, 4 etc?

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+        if "thread_id" not in config.get("configurable", {}):
+            example_input = {self.input_messages_key: "foo"}
+            example_config = {"configurable": {"thread_id": "123"}}
+            raise ValueError(
+                "thread_id is required."
+                " Pass it in as part of the config argument to .invoke() or .stream()"
+                f"\neg. chain.invoke({example_input}, {example_config})"
+            )
+        # attach message_history
+        thread_id = config["configurable"]["thread_id"]
```
eyurtsev, line 248: should we just pass a single `session_id` parameter? client can do concat logic with it if they need it

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+            **kwargs,
+        )
+
+    @property
+    def config_specs(self) -> List[ConfigurableFieldSpec]:
+        return get_unique_config_specs(
+            super().config_specs
+            + [
+                ConfigurableFieldSpec(
+                    id="user_id",
```
eyurtsev, line 127: What's the reason why we need both instead of just having a `session_id` and having the client decide how to deal with that? The reason is that we're opinionated about it's assembled (i.e., we're not passing both parameters independently)

eyurtsev (commented):
Haven't finished reviewing yet -- have to run, will try to continue in the evening

libs/langchain/langchain/schema/runnable/history.py
baskaryan: cc @nfcampos

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+        return await asyncio.get_running_loop().run_in_executor(
+            None, self._enter_history, input, config
+        )
+
+    def _exit_history(self, run: Run, config: RunnableConfig) -> None:
+        hist: BaseChatMessageHistory = config["configurable"]["message_history"]
+
+        # Get the input messages
+        inputs = load(run.inputs)
+        input_val = inputs[self.input_messages_key or "input"]
```
eyurtsev, line 226: Anyway to push the `input` literal default to initialization or does that cause an issue?

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+                "thread_id is required."
+                " Pass it in as part of the config argument to .invoke() or .stream()"
+                f"\neg. chain.invoke({example_input}, {example_config})"
+            )
+        # attach message_history
+        thread_id = config["configurable"]["thread_id"]
+        user_id = config["configurable"].get("user_id")
+        session_id = thread_id if user_id is None else f"{user_id}:{thread_id}"
+        config["configurable"]["message_history"] = self.factory(
+            session_id,
```
eyurtsev, line 252: What if instead we let the factory accept a dict with inputs so we avoid doing the string concatenation here, then users won't have to create parsing code in the factory to determine if it was only thread id or user and thread id that were provided etc

```python
    def _merge_configs(self, *configs: Optional[RunnableConfig]) -> RunnableConfig:
        config = super()._merge_configs(*configs)
        # extract thread_id
        if "thread_id" not in config.get("configurable", {}):
            example_input = {self.input_messages_key: "foo"}
            example_config = {"configurable": {"thread_id": "123"}}
            raise ValueError(
                "thread_id is required."
                " Pass it in as part of the config argument to .invoke() or .stream()"
                f"\neg. chain.invoke({example_input}, {example_config})"
            )
        # attach message_history
        thread_id = config["configurable"]["thread_id"]
        user_id = config["configurable"].get("user_id")
        config["configurable"]["message_history"] = self.factory(
            {"thread_id": thread_id, "user_id": user_id}
        )
        return config
```

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+
+                Must return as output one of:
+                - A string which can be treated as an AIMessage
+                - A BaseMessage or sequence of BaseMessages
+                - A dict with a key for a BaseMessage or sequence of BaseMessages
+
+            factory: Function that returns a new BaseChatMessageHistory given a
+                session id. Should take a single parameter `session_id` which is a
+                string.
+            input_messages_key: Must be specified if the base runnable accepts a dict
```
eyurtsev, line 92: It looks like the code is falling back on `input` being a default, so this key doesn't have to be specified

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+    output_messages_key: Optional[str] = None
+    message_history_key: Optional[str] = None
+
+    def __init__(
+        self,
+        runnable: Runnable[
+            MessagesOrDictWithMessages,
+            Union[str, BaseMessage, MessagesOrDictWithMessages],
+        ],
+        factory: Callable[[str], BaseChatMessageHistory],
```
eyurtsev, line 65: What do you think about changing the factory signature to:

```factory: Callable[[dict], BaseChatMessageHistory]```

It might be annoying client side to parse the string?

File: libs/langchain/langchain/schema/runnable/history.py [hunk 14]
```diff
@@ -0,0 +1,254 @@
...
+        history_chain: Runnable = RunnableLambda(
+            self._enter_history, self._aenter_history
+        ).with_config(run_name="load_history")
+        messages_key = message_history_key or input_messages_key
+        if messages_key:
+            history_chain = RunnablePassthrough.assign(
+                **{messages_key: history_chain}
+            ).with_config(run_name="insert_history")
+        bound = (
+            history_chain | runnable.with_listeners(on_end=self._exit_history)
```
eyurtsev, line 110: Do we need to support an async version of on_end ?

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,254 @@
...
+            return hist.messages.copy() + self._get_input_messages(input_val)
+
+    async def _aenter_history(
+        self, input: Dict[str, Any], config: RunnableConfig
+    ) -> List[BaseMessage]:
+        return await asyncio.get_running_loop().run_in_executor(
+            None, self._enter_history, input, config
+        )
+
+    def _exit_history(self, run: Run, config: RunnableConfig) -> None:
```
eyurtsev, line 221: Because this is handled by a tracer any exception in this code will be silently ignored. I don't think this is the desired behavior by default?

libs/langchain/tests/unit_tests/schema/runnable/test_history.py
```diff
@@ -0,0 +1,219 @@
...
+    output = with_history.invoke({"messages": [HumanMessage(content="hello")]}, config)
+    assert output == "you said: hello"
+    output = with_history.invoke(
+        {"messages": [HumanMessage(content="good bye")]}, config
+    )
+    assert output == "you said: hello\ngood bye"
+
```
eyurtsev, line 51: This test currently fails:

Do we want to support first argument being `str` or list of base message and necessarily a dict? (The docs in the message history runnable seem to indicate that's the case)

```python
def test_input_arg_with_history_key() -> None:
    runnable = RunnableLambda(
        lambda input: "you said: "
                      + "\n".join(
            [str(m.content) for m in input["history"] if isinstance(m, HumanMessage)]
            + [input["input"]]
        )
    )
    factory = get_factory()
    with_history = runnable.with_message_history(
        factory, message_history_key="history"
    )
    config: RunnableConfig = {"configurable": {"thread_id": datetime.datetime.now()}}
    output = with_history.invoke("hello", config)
    assert output == "you said: hello"
    output = with_history.invoke("hello", config)
    assert output == "you said: hello\ngood bye"
```

libs/langchain/langchain/schema/runnable/history.py
baskaryan: "input" only applies on exit_history when getting inputs from Run (which will wrap non-dict inputs in a dict with default key "input"), we don't want it set for enter_history (when inputs are what the user entered, not yet wrapped by default

libs/langchain/langchain/schema/runnable/base.py
```diff
@@ -820,6 +822,36 @@ def with_fallbacks(
             exceptions_to_handle=exceptions_to_handle,
         )
 
+    def with_message_history(
```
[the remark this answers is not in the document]
baskaryan: my mild preference is always for methods over classes for discoverability (if i'm already using a class it's easy to find it's methods / IDE autocomplete helps. if i need a new class i don't know where to start looking). and it doesn't seem that bad to me to have a somewhat specific method but don't feel strongly on this point

libs/langchain/langchain/schema/runnable/history.py
baskaryan: no strong opinions on my end, thoughts @nfcampos?

libs/langchain/langchain/schema/runnable/history.py
nfcampos: This is what we wanted for opengpts doesn't necessarily make sense here

libs/langchain/langchain/schema/runnable/history.py
nfcampos: I didn't do that because none of our chat history classes accept such a dict, so all would need wrappers to be used here if we go with that

libs/langchain/langchain/schema/runnable/history.py
nfcampos: Replied in other thread

libs/langchain/langchain/schema/runnable/history.py
nfcampos: With listeners runs it in thread pool. Given chat history classes don't have async methods unfortunately there's no advantage to having native async method

libs/langchain/langchain/schema/runnable/base.py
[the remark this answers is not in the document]
eyurtsev: My view is this is tackling an over-specialized method to an already large namespace -- i.e., makes it a bit easier to use memory at the expense of using using runnables in general

libs/langchain/langchain/schema/runnable/history.py
eyurtsev: With that said I'm OK with user id and thread id as being the default parametrization, but another thing we could do is to accept the parameterization through the init, and pass it along to the factory

libs/langchain/tests/unit_tests/schema/runnable/test_history.py
eyurtsev: Discussed with @baskaryan  -- it shouldn't work with str :)

libs/langchain/langchain/schema/runnable/history.py
eyurtsev: What do we think if we reduce this to `session_id` as a default to match existing implementations? 

* It's possible to parse user id and thread id from the session id, so it won't block folks from implementing that.

---

In addition, we could do the following as an enhancement if we like it in the future:

* enhance functionality to accept a factory which takes either a dict or a str as the first argument and based on the type signature invoke it appropriately
* Allows user to override default config spec by passing something during initialization (and assume this only works with a factory that takes a dict)

libs/langchain/langchain/schema/runnable/history.py
```diff
@@ -0,0 +1,288 @@
...
+    get_unique_config_specs,
+)
+
+if TYPE_CHECKING:
+    from langchain.callbacks.tracers.schemas import Run
+    from langchain.schema.messages import BaseMessage
+    from langchain.schema.runnable.config import RunnableConfig
+
+MessagesOrDictWithMessages = Union[Sequence["BaseMessage"], Dict[str, Any]]
+GetSessionHistoryCallable = Callable[..., BaseChatMessageHistory]
```
eyurtsev, line 32:
```suggestion
GetSessionHistoryCallable = Callable[[str], BaseChatMessageHistory]
```

eyurtsev (approved):
I think this looks good

libs/langchain/langchain/schema/runnable/history.py
baskaryan: might cause issues if you pass in something that takes more than str (like RedisChatMessageHistory)

PR #13434, opened 2023-11-16
updated `async-faiss` example

leo-gan:
The original notebook has the `faiss` title which is duplicated in the`faiss.jpynb`. As a result, we have two `faiss` items in the vectorstore ToC. And the first item breaks the searching order (it is placed between `A...` items).
- I updated title to `Asynchronous Faiss`.

### DIFF
--- a/docs/docs/integrations/vectorstores/async_faiss.ipynb
+++ b/docs/docs/integrations/vectorstores/async_faiss.ipynb
@@ -5,15 +5,16 @@
    "id": "683953b3",
    "metadata": {},
    "source": [
-    "# Faiss\n",
+    "# Asynchronous Faiss\n",
     "\n",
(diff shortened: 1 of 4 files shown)


efriis (commented):
@leo-gan let's actually just delete this file because it's a duplicate of faiss.ipynb with a few characters changed to make it async.

Maybe delete this and add a note to the bottom of faiss.ipynb describing how to do the guide in async way with a single example?

leo-gan:
@efriis What if I copy the asynch_fai content into the faiss with correspondent heading? IMO asynchronous code is kind of important and can be referenced when we learn asynchronous usage.

efriis:
I think it will look very redundant. If you look at the diff between the two files, they look almost identical

efriis:
If you do want to keep them separate, maybe name this one "FAISS (Async)" so they appear next to each other in the sidebar?

leo-gan:
I like idea with "Faiss (Asynch)". It is fast to switch and easy to reference.

efriis (approved):

efriis:
Thanks for fixing!

PR #13443, opened 2023-11-16
Use List instead of list

ifduyue:
Unify List usages in libs/langchain/langchain/text_splitter.py, only one place it's `list`, all other ocurrences are `List`


baskaryan:
thanks @ifduyue!

PR #13449, opened 2023-11-16
Astra DB: minor improvements to docstrings and demo notebook

hemidactylus:
This PR brings a few minor improvements to the docs, namely class/method docstrings and the demo notebook.

- A note on how to control concurrency levels to tune performance in bulk inserts, both in the class docstring and the demo notebook;
- Slightly increased concurrency defaults after careful experimentation (still on the conservative side even for clients running on less-than-typical network/hardware specs)
- renamed the DB token variable to the standardized `ASTRA_DB_APPLICATION_TOKEN` name (used elsewhere, e.g. in the Astra DB docs)
- added a note and a reference (add_text docstring, demo notebook) on allowed metadata field names.

Thank you!


baskaryan:
thank you @hemidactylus!

PR #13451, opened 2023-11-16
VertexAI LLM count_tokens method requires list of prompts

stijntratsaertit:
I encountered this during summarization with VertexAI. I was receiving an INVALID_ARGUMENT error, as it was trying to send a list of about 17000 single characters. 

The [count_tokens method](https://github.com/googleapis/python-aiplatform/blob/main/vertexai/language_models/_language_models.py#L658) made available by Google takes in a list of prompts. It does not fail for small texts, but it does for longer documents because the argument list will be exceeding Googles allowed limit. Enforcing the list type makes it work successfully.

This change will cast the input text to count to a list of that single text so that the input format is always correct.

[Twitter](https://www.x.com/stijn_tratsaert)

### DIFF
--- a/libs/langchain/langchain/llms/vertexai.py
+++ b/libs/langchain/langchain/llms/vertexai.py
@@ -288,7 +288,7 @@ class VertexAI(_VertexAICommon, BaseLLM):
             The integer number of tokens in the text.
         """
         try:
-            result = self.client.count_tokens(text)
+            result = self.client.count_tokens([text])
         except AttributeError:
             raise NotImplementedError(
                 "Your google-cloud-aiplatform version didn't implement count_tokens."


stijntratsaertit:
> Should we update an integration test too?

It would be possible to add an new test that would count the document of a large document. But as this is merely a change to adhere to the function signature, it might be a bit extreme. Let me know.
The existing tests don't change in outcome, as the previous solution worked too in most cases.

stijntratsaertit:
> Should we update an integration test too?

It would be possible to add an new test that would count the content of a large document. But as this is merely a change to adhere to the function signature, it might be a bit extreme. Let me know.
The existing tests don't change in outcome, as the previous solution worked too in most cases.

stijntratsaertit:
Can we pursue this? @lkuligin

lkuligin:
SGTM, thanks!

PR #13467, opened 2023-11-16
Bugfix: OpenAIFunctionsAgentOutputParser doesn't handle functions with no args

chrisaffirm:
**Description/Issue:** 
When OpenAI calls a function with no args, the args are `""` rather than `"{}"`. Then `json.loads("")` blows up. This PR handles it correctly.

**Dependencies:** None

### DIFF
--- a/libs/langchain/langchain/agents/output_parsers/openai_functions.py
+++ b/libs/langchain/langchain/agents/output_parsers/openai_functions.py
@@ -44,7 +44,12 @@ class OpenAIFunctionsAgentOutputParser(AgentOutputParser):
         if function_call:
             function_name = function_call["name"]
             try:
-                _tool_input = json.loads(function_call["arguments"])
+                if len(function_call["arguments"].strip()) == 0:
+                    # OpenAI returns an empty string for functions containing no args
+                    _tool_input = {}
+                else:
+                    # otherwise it returns a json object
+                    _tool_input = json.loads(function_call["arguments"])
             except JSONDecodeError:
                 raise OutputParserException(
                     f"Could not parse tool input: {function_call} because "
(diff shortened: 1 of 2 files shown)


efriis (commented):
Do you see this behavior consistently? Do you have an example call to OpenAI that would return this?

When I try to reproduce, I consistently get `'arguments': '{}'`

chrisaffirm:
>Do you see this behavior consistently? Do you have an example call to OpenAI that would return this?
>
>When I try to reproduce, I consistently get 'arguments': '{}'

The behavior is inconsistent. Here's some fairly minimal code that reproduces this for me:

```
def _build_no_arg_functions(name_desc: List[Tuple[str, str]]) -> List[Dict]:
    return [
        {
            "name": x[0],
            "description": x[1],
            "parameters": { "type": "object", "properties": {}, "required": [],},
        } for x in name_desc
    ]

off_topic_prompt = ChatPromptTemplate.from_messages(
    [
        (
            "system",
            """
I will show you a question that someone asked me about animals.

If the animal is fluffy, call the function is_fluffy. If the animal flies, call the function does_fly.
Do not answer the question. Do not say answer the question, just call one of these functions.
""",  # noqa: #501
        ),
        ("human", "{text}"),
    ]
)

off_topic_model = ChatOpenAI(model_name="gpt-3.5-turbo-0613").bind(
    functions=_build_no_arg_functions(
        [
            ("is_fluffy", "This should be called for fluffy animals"),
            ("does_fly", "This function should be called for animals that fly."),
        ]
    )
)

off_topic_chain = off_topic_prompt | off_topic_model #| AffirmOpenAIFunctionsAgentOutputParser()
```
image

This behavior is not stable w.r.t. prompt. For example suppose I change the prompt to:

```
I will show you a question that someone asked me about animals. 

If the animal is fluffy, call the function is_fluffy. If the animal flies, call the function does_fly.
If the question is about an animal that neither flies nor is fluffy, answer the question. 

Do not answer the question. Do not say answer the question, just call one of these functions.
```
Then I get valid json in the return value.

image

efriis (approved):
Worked great. Thanks @chrisaffirm ! Kudos for implementing a unit test

PR #11113, opened 2023-09-27
auto escape stop sequences in 'enforce_stop_tokens' helper function

Tomas2D:
**Description**

I've noticed that the split function interprets the inputs as regex expressions, which can sometimes lead to unexpected behaviour.

For instance: the LLM implementation can pass stop parameters to the appropriate model (or inference API) and then it will call the `enforce_stop_tokens` function to be 100% sure that there is no extra text after the stop sequence has been reached.

So where is the problem?
If you pass `stop=["."]`, the API will interpret the stop sequences as is, but the helper function interprets the `.` as regex matching any character yielding in an empty result.

Tag maintainer:
@hwchase17
@baskaryan

PR #13468, opened 2023-11-16
fixed: Missing escape in llms.utils.enforce_stop_tokens()

### LINKED ISSUE #13397: Missing escape in `llms.utils.enforce_stop_tokens()`
### System Info

Not relevant.

### Who can help?

@hwchase17
@agola11

### Information

- [ ] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Memory
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Use any LLM relying on stop words with special regex characters.

For example instanciate a `HuggingFacePipeline` LLM instance with [openchat](https://huggingface.co/openchat/openchat_3.5) model. This model uses `<|end_of_turn|>` stop words.

Since the `llms.utils.enforce_stop_tokens()` function doesn't escape the provided stop words strings the `|` chars are interpreted as part of the regex instead of the stop word. So in this case any single `<` chars in the output would trigger the
split.

### Expected behavior

Stop words should be escaped with `re.escape()` so the split only happens on the complete words.

umair313:
**Problem:**
The problem with the enforce_stop_tokens method as currently implemented is that it directly joins the stop words with the | character for the re.split function. This approach works well if the stop words don't contain any regex special characters. However, if they do, this can lead to unintended behavior because the special characters are interpreted as part of the regex pattern, not as literal characters.

**Solution**
To fix this issue, each stop word should be escaped so that any special characters it contains are treated as literals. 

fixed: #13397

### DIFF
--- a/libs/langchain/langchain/llms/utils.py
+++ b/libs/langchain/langchain/llms/utils.py
@@ -5,4 +5,8 @@ from typing import List
 
 def enforce_stop_tokens(text: str, stop: List[str]) -> str:
     """Cut off the text as soon as any stop words occur."""
-    return re.split("|".join(stop), text, maxsplit=1)[0]
+
+    escaped_stop = [re.escape(word) for word in stop]
+    pattern = "|".join(escaped_stop)
+
+    return re.split(pattern, text, maxsplit=1)[0]


efriis:
Same PR as #11113 - closing because people can escape before passing in, and anyone currently escaping would be broken by this.

umair313:
@efriis 

Thanks for your feedback. Did you guys left the issue #13397 open intentionally?

efriis:
Commented

PR #13471, opened 2023-11-16
fix for `integratons/document_loaders` sidebar

leo-gan:
The current `integrations/document_loaders/` sidebar has the `example_data` item, which is a menu with a single item: "Notebook".
It is happening because the `integrations/document_loaders/` folder has the `example_data/notebook.md` file that is used to autogenerate the above menu item.
- removed an example_data/notebook.md file. Docusaurus doesn't have simple ways to fix this problem (to exclude folders/files from an autogenerated sidebar). Removing this file didn't break any existing examples, so this fix is safe.


leo-gan:
@efriis Hi Eric, Vercel is failing for unknown reason. The [Details](https://vercel.com/langchain/langchain/ByZFMwZ2tYxCCsMhGKb38iKUzkbA) of this cannot be accessed.
Can you, please, help?

efriis (approved):
Looks like this file isn't even used in examples. Probably got caught up in some conversion because example_data/notebook.ipynb is missing for this notebook example, but people can pick a different notebook in there for now.

PR #13481, opened 2023-11-16
Resolve Buffer Clearing Issue in Summary Generation

Jyumpp:
**Description:** Despite this class intending to summarize just enough messages to fit within a set context size, the product of this class always contains *ALL* messages *PLUS* a summary system message at the front, ultimately making token count longer rather than shorter. This commit addresses the issue by fixing the single bug where the buffer is constantly reset to the internal message store even after messages are summarized and removed. 

**Issues:** Couldn't find one, misread the issue I thought was similar with my last pull request.

@efriis (no LLM this time, I may have drank too much of the kool-aid)

Reason for changes in diff order:
- Local member needed to store truncated message list
- Allow buffer to use local chat list if a moving summary exists
- Let the BaseMessage object have 'system' type specifically (required to pass type checking)
- Store truncated list when `prune` runs

### DIFF
--- a/libs/langchain/langchain/memory/summary_buffer.py
+++ b/libs/langchain/langchain/memory/summary_buffer.py
@@ -12,10 +12,15 @@ class ConversationSummaryBufferMemory(BaseChatMemory, SummarizerMixin):
     max_token_limit: int = 2000
     moving_summary_buffer: str = ""
     memory_key: str = "history"
+    chat_messages: List[BaseMessage] = []
 
     @property
     def buffer(self) -> List[BaseMessage]:
-        return self.chat_memory.messages
+        return (
+            self.chat_memory.messages
+            if self.moving_summary_buffer == ""
+            else self.chat_messages
+        )
 
     @property
     def memory_variables(self) -> List[str]:
@@ -30,7 +35,9 @@ class ConversationSummaryBufferMemory(BaseChatMemory, SummarizerMixin):
         buffer = self.buffer
         if self.moving_summary_buffer != "":
             first_messages: List[BaseMessage] = [
-                self.summary_message_cls(content=self.moving_summary_buffer)
+                self.summary_message_cls(
+                    content=self.moving_summary_buffer, type="system"
+                )
             ]
             buffer = first_messages + buffer
         if self.return_messages:
@@ -70,6 +77,7 @@ class ConversationSummaryBufferMemory(BaseChatMemory, SummarizerMixin):
             self.moving_summary_buffer = self.predict_new_summary(
                 pruned_memory, self.moving_summary_buffer
             )
+            self.chat_messages = buffer
 
     def clear(self) -> None:
         """Clear memory contents."""


efriis (commented):
Thanks for writing it up! 2 questions

1. could you post some example code or write a unit test showing the error you describe, where all messages are being included
2. is there a way to solve this without adding the `chat_messages` attribute and instead relying on the inherited `self.chat_memory.messages`?

Jyumpp:
Thanks for getting back to me! I can write up a test tomorrow to demonstrate the error, it I discovered it from ballooning context costs, it became more noticeable now that OpenAI models accept 16k context tokens. As for self.chat_memory.messages, I don't believe that's mutable, as the chat memory seems to be implemented as a BaseChatHistory, more so exposing any of the memory backends in memory/chat_message_histories rather than being able to write back to.

efriis:
Got it. I'm actually seeing the right behavior when I follow the interface described in the docs https://python.langchain.com/docs/modules/memory/types/summary_buffer without this change.

Memory is in need of some work, so appreciate your attention to it! I would recommend using the interface as described in the docs to avoid infinitely accumulating messages though.

Jyumpp:
I respectfully disagree. While it does work with the default uninitialized memory, passing a premade history object, such as with the Postgres History chat_message_history loader, messages are not mutable.

Jyumpp:
I've taken a pretty thorough reading of the memory source so far, and my team just submitted our research paper for our LLM framework for using these tools in a manufacturing setting, where actual persistent data stores are invaluable. If there are different pain points in memory you think should be addressed first, such as a possible other bug with the Postgres loader rather than this class, I'd be eager to put my time towards those instead. :)

efriis:
Got it. Do you have code showing that bug? I didn't realize you were using it with something postgres related, and I could see that being an area ripe for bugs or questionable compatibility.

The main reason this PR shouldn't be merged is because it stores duplicate information in a new `chat_messages` attribute to `self.chat_memory.messages`

Jyumpp:
My current code showing the bug is only our internal project, I can get the code cleaned up to push to our org repo to share plus a recreation test case. 

I agree the duplication is a bad idea. Instead, would it suffice to alter the ‘self.moving_summary_buffer‘ to be a tuple of the buffer string and the index of where ‘first_messages‘ ends?

Jyumpp:
I just pushed the work for our Discord bot to our org: [HERE](https://github.com/UGA-Innovation-Factory/AlfredBot/blob/main/alfred.py) 
- It requires 1 Postgres db, embeddings and chat logs stored in same db, different tables

I included 2 text files with the as_is langchain implementation, and the with_edit version that I'm using with the modifications, importing as a modified class in the project root dir.

efriis:
Very interesting. Any interest in being one of the first users of our new Runnable-compatible memory? It landed today! https://python.langchain.com/docs/expression_language/how_to/message_history#adding-message-history

I think the main issue with the current approach is that BufferSummary memory doesn't support manually passing in the `chat_history` object as this code does.

PR #13483, opened 2023-11-16
Add Batch Size kwarg to the llm start callback

hinthornw:
So you can more easily use the token counts directly from the API endpoint for batch size of 1


hinthornw:
Ah phooey gotta do against core now

PR #13492, opened 2023-11-17
docs updating `AzureML` notebooks

leo-gan:
- Added/updated descriptions and links

### DIFF
--- a/docs/docs/integrations/chat/azure_chat_openai.ipynb
+++ b/docs/docs/integrations/chat/azure_chat_openai.ipynb
@@ -7,7 +7,7 @@
    "source": [
     "# Azure OpenAI\n",
     "\n",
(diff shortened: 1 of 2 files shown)


docs/docs/integrations/chat/azure_chat_openai.ipynb
efriis, line 1: This description might be more confusing than it's worth. The old one is more concise and gets to the point?

efriis: I think you might have found a different docs page! Let's link this one instead: https://learn.microsoft.com/en-us/azure/ai-services/openai/overview

leo-gan: We have a separate page for [Azure OpenAI models](azure_chat_openai) . ^ this is about the different integration. It is based on the `AzureMLChatOnlineEndpoint` class. 
So, I think the updated version is correct.

efriis: I think you've linked the file that I commented on, which I'm pretty sure is about AzureChatOpenAI.

The other file you commented on looks fine. This file is about AzureChatOpenAI though, not the AzureML Endpoints.

leo-gan: Oops. Yes, you are right. I'm commenting about another file. Sorry. 
Yes, your links is better. Thanks!

efriis (approved):

PR #13508, opened 2023-11-17
Documentation: Reduce largest static image file size

rihardsgravis:
- **Description:** Reduce image asset file size used in documentation by running them via lossless image optimization ([tinypng](https://www.npmjs.com/package/tinypng-cli) was used in this case). Images wider than 1916px (the maximum width of an image displayed in documentation) where downsized.
  - **Issue:** No issue is created for this, but the large image file assets caused slow documentation load times
  - **Dependencies:** No dependencies affected


hwchase17 (commented):
thanks, this seems fine to me

efriis:
Would it be better to do something like this, and keep the higher resolution ones in the repo? Would also apply to new images over time instead of having to re-run this script to resize images.

https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-ideal-image

efriis:
Ah actually looks like that only applies to `<Image/>` components rather than markdown images. This lgtm too!

PR #13511, opened 2023-11-17
Fix `MarkdownHeaderTextSplitter` not recognizing tilde-fenced code blocks

unifyh:
- **Description:** Previously `MarkdownHeaderTextSplitter` did not consider tilde-fenced code blocks (https://spec.commonmark.org/0.30/#fenced-code-blocks). This PR fixes that.
   ````md
   # Bug caused by previous implementation:
   ~~~py
   foo()
   # This is a comment that would be considered header
   bar()
   ~~~
   ````

### DIFF
--- a/libs/langchain/langchain/text_splitter.py
+++ b/libs/langchain/langchain/text_splitter.py
@@ -391,16 +391,23 @@ class MarkdownHeaderTextSplitter:
         initial_metadata: Dict[str, str] = {}
 
         in_code_block = False
+        opening_fence = ""
 
         for line in lines:
             stripped_line = line.strip()
 
-            if stripped_line.startswith("```"):
-                # code block in one row
-                if stripped_line.count("```") >= 2:
+            if not in_code_block:
+                # Exclude inline code spans
+                if stripped_line.startswith("```") and stripped_line.count("```") == 1:
+                    in_code_block = True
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/text_splitter.py
```diff
@@ -391,16 +391,23 @@ def split_text(self, text: str) -> List[Document]:
...
 
-            if stripped_line.startswith("```"):
-                # code block in one row
-                if stripped_line.count("```") >= 2:
+            if not in_code_block:
+                # Exclude inline code spans
+                if stripped_line.startswith("```") and stripped_line.count("```") == 1:
+                    in_code_block = True
+                    opening_fence = "```"
+                elif stripped_line.startswith("~~~"):
```
baskaryan, line 404: do we need to do the same count check here , in case start and end of block are in a single line

unifyh: No, see https://spec.commonmark.org/0.30/#example-146
tl;dr: if 2 `~~~`'s appear in one line, the latter is considered part of info string, instead of end of block

unifyh:
Hi @baskaryan, is it good to go now?

hwchase17 (approved):

hwchase17:
thanks!

PR #13512, opened 2023-11-17
AstraDB: use includeSimilarity option instead of $similarity

nicoloboschi:
- **Description:** AstraDB is going to deprecate the `$similarity` projection property in favor of the ´includeSimilarity´ option flag. I moved all the queries to the new format. 
- **Tag maintainer:** @hemidactylus 
- **Twitter handle:** nicoloboschi

### DIFF
--- a/libs/langchain/langchain/vectorstores/astradb.py
+++ b/libs/langchain/langchain/vectorstores/astradb.py
@@ -480,12 +480,11 @@ class AstraDB(VectorStore):
             self.collection.paginated_find(
                 filter=metadata_parameter,
                 sort={"$vector": embedding},
-                options={"limit": k},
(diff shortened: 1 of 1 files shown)


efriis:
cc @hemidactylus !

efriis:
Not sure how versioning works on AstraDB, but will we need to support the old format for backwards compatibility for some people?

nicoloboschi:
> Not sure how versioning works on AstraDB, but will we need to support the old format for backwards compatibility for some people?

The `includeSImilarity` option works fine both in older and newer versions so this fix is compatible with both

hemidactylus:
Correct, as @nicoloboschi mentions the new method works with older API versions too so it is not a problem.
LGTM if that matters :)

PR #13518, opened 2023-11-17
docs: expand the desktop threshold width

dqbd:


efriis:
Does this do it? Looks like this still has the same issue

dqbd:
Closing for impracticality

PR #13524, opened 2023-11-17
updated cohere.py to support input_type for new embedding models

joeylmartin:


efriis:
Hey there! Closing because this is a breaking change without much explanation of why these changes were introduced. If you want to re-submit, a longer description would be great, and try to do it while maintaining existing behavior for people currently using the class (e.g. removing attributes in this case) :)

PR #13528, opened 2023-11-17
add llm kwargs as loading parameter

curefatih:
**Description:**
Enables runtime parameter injection while loading serialized chains.

 **Issue:**
Current approach does not supports runtime credential loading if there is multiple or conditional logic with the LLM settings while loading. 

For e.g. If we have serialized object like this one 
```python

chain_config = {
    "memory": None,
    "verbose": False,
    "prompt": {
        "input_variables": [
            "topic"
        ],
        "output_parser": None,
        "template": "Tell me a joke about {topic}:",
        "template_format": "f-string",
        "_type": "prompt"
    },
    "llm": {
        "model_name": "text-davinci-003",
        "temperature": 0.9,
        "max_tokens": 256,
        "top_p": 1,
        "frequency_penalty": 0,
        "presence_penalty": 0,
        "n": 1,
        "best_of": 1,
        "request_timeout": None,
        "logit_bias": {},
        "_type": "openai"
    },
    "output_key": "text",
    "_type": "llm_chain"
}
```
and if we try to load without environment variable it raises an error because of the api key does not exists in either env variables or inside the serialized object (as expected)

```python
from langchain.chains.loading import load_chain_from_config

chain = load_chain_from_config(chain_config)

# Did not find openai_api_key, please add an environment variable `OPENAI_API_KEY` which contains it, or pass  `openai_api_key` as a named parameter. (type=value_error)
``` 

and what I hgumbly propose in here is using the kwargs which is already defined in current version to add LLM parameters for using as settings like this one

```python
chain = load_chain_from_config(chain_config, **{
  "llm_extras": {
    "openai": {
      "openai_api_key": "<super_secret>"
    }
  }
})
```

**Dependencies:** 
No extra dependency needed.

**Tag maintainer:**
I don't know you are the right person to tag but @eyurtsev I will be very appreciated if there is any direction or review. Thanks in advance!

 **Twitter handle:**
This is not a big feature/fix bug but here it is `@cureef`

### DIFF
--- a/libs/langchain/langchain/chains/loading.py
+++ b/libs/langchain/langchain/chains/loading.py
@@ -37,9 +37,9 @@ def _load_llm_chain(config: dict, **kwargs: Any) -> LLMChain:
     """Load LLM chain from config dict."""
     if "llm" in config:
         llm_config = config.pop("llm")
-        llm = load_llm_from_config(llm_config)
+        llm = load_llm_from_config(llm_config, **(kwargs.get('llm_extras', {})))
     elif "llm_path" in config:
-        llm = load_llm(config.pop("llm_path"))
+        llm = load_llm(config.pop("llm_path"), **(kwargs.get('llm_extras', {})))
     else:
         raise ValueError("One of `llm` or `llm_path` must be present.")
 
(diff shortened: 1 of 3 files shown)


curefatih:
Hi, @efriis 👋 Is there any chance to review? Thanks in advance!

baskaryan:
thanks for pr @curefaith! we should resolve this, but also want to note that recommended way to serialize objects these days is using langchain.load module. you can see what that looks like here https://python.langchain.com/docs/modules/model_io/llms/llm_serialization. If you serialize the prompt and LLM separately using the new serialization logic then things like API keys should work out of the box. curious if that would solve your issue in this case

curefatih:
Hi @baskaryan Thank you for the review. It is totally what I was trying to do. I thought the new load was backward compatible and tried with the given example but it seems like completely changed. I have updated the serialized object now it is all good. Thank again for the review!

Issue #12240, opened 2023-10-24
Issue: Problems getting filtering to work with langchain, OpenSearchVectorSearch, similarity_search_with_score


jrabyIBM:
### Issue you'd like to raise.

Using langchain on AWS lambda function to call an AWS OpenSearch Serverless vector index.  I did not file as a bug, as I suspect that this is a usage problem, but it might be a bug.

I have built an OpenSearch vector index, using AWS OpenSearch Serverless, and are using it to perform vector queries against document chunks.  This is all working well. I would now like to add the ability to filter these searches by document (there may be hundreds or thousands of vector entries per document).  There is metadata included with each entry, which includes a DocumentId field, which is a string-ized guid.  I thought it would be simple to filter against this metadata attribute, but cannot get it working.  

The call I am making is:
```
    docs = vsearch.similarity_search_with_score(
        theQuery,
        index_name="vector-index2",
        vector_field = "vector_field",
        search_type="script_scoring",
        pre_filter=filter_dict,
        space_type="cosinesimil",
        engine="faiss",
        k = 3
    )
```

Typical unfiltered output:
{"doc": {"page_content": "seat is  installed ... facing", "metadata": {"DocumentId": "{8B5D5890-0000-C1E8-B0B6-678C8509665D}", "DocumentTitle": "2023-Toyota-RAV4-OM.pdf", "DocClassName": "Document"}}, "score": 1.8776696}]

For filter_dict, I have tried:
    filter_dict = {'DocumentId':theDocId}                                    <-- Gets an error 
    filter_dict = {'bool': {'must': {'term': {'DocumentId': theDocId} } } }  <-- doesn’t find anything
    filter_dict = {'bool': {'must': {'match': {'DocumentId': theDocId} } } } <-- doesn’t find anything
    filter_dict = {'must': {'match': {'DocumentId': theDocId} } }            <-- Gets an error
    filter_dict = {'match': {'DocumentId': theDocId} }                         <-- doesn’t find anything

theDocId is a string set to an input parameter and looks like this:  {8B5D5890-0000-C1E8-B0B6-678C8509665D}.  Results are the same if I use a string constant (i.e. '{8B5D5890-0000-C1E8-B0B6-678C8509665D}')

Same results if I use search_type=”approximate_search” in conjunction with boolean_filter=filter_dict

Any suggestions on how to get this to work?

jrabyIBM:
@dosu-bot Thanks for the quick reply!

[...]

Issue #12179, opened 2023-10-24
Issue: <Not able to filter OpenSearch vectorstore  using Filter in search_kwargs>


rchegondi:
### Issue you'd like to raise.

Hello Team,
We have a metadata field in our opensearch index which I would like use to filter the passages before sending it to LLM to generate the answer. Tried many options suggested in chat.langchain.com but none of them seems to work.

Here is an example:
I have an user with associated groups ['user_group1', 'user_group2'] and the OpenSearch index with data like 
{
'text': 'Page Content......',
'metadata': {
     'source_acl': [ {'group_name': 'user_group1',  'group_type': 'type1'}, {}, {}] 
}
}

So far I have been trying this

```
chain = ConversationalRetrievalChain.from_llm(
                llm=SagemakerEndpoint(
                    endpoint_name=constants.EMBEDDING_ENDPOINT_ARN_ALPHA,
                    region_name=constants.REGION,
                    content_handler=self.content_handler_llm,
                    client=self.sg_client
                ),
                # for experimentation, you can change number of documents to retrieve here
                retriever=opensearch_vectorstore.as_retriever(search_kwargs={'k': 3, 'filter': 'source_acl.group_name:user_group1'}),
                memory=memory,
                condense_question_prompt=custom_question_prompt,
                return_source_documents=True
            )
```

Am I missing anything here?

Thanks in Advance.

PR #13529, opened 2023-11-17
fix: filtering in opensearch was not working

RomuloPaiva01:
- changed the filtering inside the query (boolean filtering)

### DIFF
--- a/libs/langchain/langchain/vectorstores/opensearch_vector_search.py
+++ b/libs/langchain/langchain/vectorstores/opensearch_vector_search.py
@@ -208,10 +208,10 @@ def _approximate_search_query_with_boolean_filter(
         "size": k,
         "query": {
             "bool": {
-                "filter": boolean_filter,
                 subquery_clause: [
                     {"knn": {vector_field: {"vector": query_vector, "k": k}}}
                 ],
+                "filter": [boolean_filter],
             }
         },
     }


efriis:
Hey @RomuloPaiva01 ! Could you link the docs for the filter format you're matching? Seems like the recommendations in some related issues have been different

Looks like some potentially related issues:
#12240
#12179

baskaryan:
cc @naveentatikonda

nicoloboschi (approved):

RomuloPaiva01:
> Hey @RomuloPaiva01 ! Could you link the docs for the filter format you're matching? Seems like the recommendations in some related issues have been different
> 
> Looks like some potentially related issues: #12240 #12179

Hey @efriis thank you very much for the reply!

https://opensearch.org/docs/latest/search-plugins/knn/filter-search-knn/

In the documentation, boolean filtering is described as an after-filtering process. This means that the filtering step should be added after the KNN (as per my change), rather than before, as it was previously.

efriis:
Hey @RomuloPaiva01 ! I actually don't see the array format you've implemented here in the docs you linked. The closest I see is this https://opensearch.org/docs/latest/search-plugins/knn/filter-search-knn/#scoring-script-filter

which implements the existing interface, where `query.bool.filter` is a dict, not a list.

I'm going to close this for now. Could you check your opensearch version and see if maybe you're on an old version?

PR #13538, opened 2023-11-17
update OpenAI function agents' llm validation

price-deshaw:
- **Description:** This PR modifies the LLM validation in OpenAI function agents to check whether the LLM supports OpenAI functions based on a property (`supports_oia_functions`) instead of whether the LLM passed to the agent `isinstance` of `ChatOpenAI`. This allows classes that extend `BaseChatModel` to be passed to these agents as long as they've been integrated with the OpenAI APIs and have this property set, even if they don't extend `ChatOpenAI`. 
  - **Issue:** N/A
  - **Dependencies:** none


baskaryan (commented):
should we just get rid of the validation step altogether? the validation makes for a slightly nicer error but doesn't seem worth the overhead to me

price-deshaw:
> should we just get rid of the validation step altogether? the validation makes for a slightly nicer error but doesn't seem worth the overhead to me

this seems reasonable to me, especially considering the `OpenAIFunctionsAgent`s already contain `OpenAI` in their name. I updated the PR accordingly if you'd like to re-review @baskaryan

hwchase17 (approved):
thanks!

PR #13541, opened 2023-11-17
implement the summary buffer fix without insantiating new BaseMessage…

Jyumpp:
**Description:** Despite this class intending to summarize just enough messages to fit within a set context size, the product of this class always contains ALL messages PLUS a summary system message at the front, ultimately making token count longer rather than shorter. This commit addresses the issue by fixing the single bug where the buffer is constantly reset to the internal message store even after messages are summarized and removed. 

- The error does not occur with in-memory History buffers, but instead fails when trying to access and modify an immutable database (such as a Postgres DB)
- [My organization's project repo](https://github.com/UGA-Innovation-Factory/AlfredBot) contains the code that triggered this error to be discovered, as well as output Verbose=True files for the current langchain master, my initial fix, and this revised solution that doesn't replicate any extra Lists.

**Issues:** Couldn't find one, misread the issue I thought was similar with my last pull request.

@efriis 

Reason for changes in diff order:
- An index is stored with the moving summary buffer to represent up to which messages it represents
- The buffer property only returns memory outside of the summary buffer, without a duplicate List this time
- This satisfy's the need for the required member `type` for any BaseMessage constructor
- Update the prune index
- Clear index when clearing other members

### DIFF
--- a/libs/langchain/langchain/memory/summary_buffer.py
+++ b/libs/langchain/langchain/memory/summary_buffer.py
@@ -10,12 +10,12 @@ class ConversationSummaryBufferMemory(BaseChatMemory, SummarizerMixin):
     """Buffer with summarizer for storing conversation memory."""
 
     max_token_limit: int = 2000
-    moving_summary_buffer: str = ""
+    moving_summary_buffer: tuple[str, int] = "", 0
     memory_key: str = "history"
 
     @property
     def buffer(self) -> List[BaseMessage]:
-        return self.chat_memory.messages
+        return self.chat_memory.messages[self.moving_summary_buffer[1] :]
 
     @property
     def memory_variables(self) -> List[str]:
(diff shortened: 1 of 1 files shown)


efriis:
Hey @Jyumpp ! This is the third time you've opened this PR, and I understand your frustration with the class. Would you be willing to try out the more up to date method of memory that I mentioned in #13481 ? https://python.langchain.com/docs/expression_language/how_to/message_history#adding-message-history

On this older memory object (`ConversationSummaryBufferMemory`):
In your PR description, you repeatedly reference the memory not functioning like you expect, but when I use the recommended interface documented in the docs, I don't see the behavior you see. As mentioned in #13481, I would guess this has to do with your code manually attaching a postgres-backed chat message history.

This code is unfortunately not mergeable because it would be a breaking change for people using it in the documented way (public attributes change type, return types of functions are different).

I'll leave this open to hear if you're ok to use the new one, but I'll probably close this afterwards!

Jyumpp:
Thanks for your time! I'll take a look at this new method, but that guide seems to indicate that it functions *only* on histories and not memories, which I think is more likely to bury the problem than solve it. I do feel that the ability to attach a reconstructed History object is important to many scenarios should they ever extend to prod tools rather than demo toys. To overcome this, would it be better to A) include a private member as the index rather than using the tuple method B) create a new summary buffer memory type exclusively for these immutable cases or C) suffice to just use patched code in projects that happen to run into this issue.

PR #13543, opened 2023-11-17
fixed `openai_assistant` namespace

leo-gan:
BUG: langchain.agents.openai_assistant has a reference as
`from langchain_experimental.openai_assistant.base import OpenAIAssistantRunnable`
should be 
`from langchain.agents.openai_assistant.base import OpenAIAssistantRunnable`

This prevents building of the API Reference docs


efriis (approved):
Thanks! Great catch

Issue #11209, opened 2023-09-29
Return output from map chain in the final result of ConversationalRetrievalChain.


gftempleton:
### Feature request

Hi, 
I am working on Document QA, and using ConversationalRetrievalChain, with map-reduce as chain type. 
In the final result of ConversationalRetrievalChain, I want to get the individual output from each map-chain.

### Motivation

This is a separate issue, and I think there should be a parameter that we can pass, to get the output of these chains as well. 

### Your contribution

Not as of now.

Farhan-Faisal:
I will like to look into it and see whether something like this is already implemented. If not, I will try to implement it and submit a PR.

gftempleton:
Hi Farhan, 
Thank you for answering. 
I need the final result dictionary to have the intermediate results, just like we have source dictionary and other stuff. 
Thanks.

PitcairnHotdog:
Hi gftempleton,

MapReduceDocumentChain contains a flag named “return_intermediate_steps” that allows intermediate steps to be outputted along with the final result. I guess that is what you are referring to?

After looking into it, I did find out that this flag didn't work with ConversationalRetrievalChain. I got a "ValueError: "run" not supported when there is not exactly one output key." Could you please kindly confirm whether this is also the issue you encountered?

PR #13547, opened 2023-11-18
Implemented the option to include output from map chain in the final result of ConversationalRetrievalChain

paperraptor:
**Description:** 
MapReduceDocumentChain contains a flag named “return_intermediate_steps” that allows intermediate steps to be outputted along with the final result. Normally, when this flag is set to True, the final result dictionary should contain a list of intermediate results produced during the map reduce process. However, when a MapReduceDocumentChain object with “return_intermediate_steps” set to True is used as a part of a ConversationalRetrievalChain object, an error will arise, making it impossible for users to obtain intermediate results of the execution of a ConversationalRetrievalChain object with map reduce.

The main cause of the issue seems to be inside  _call() of class BaseConversationalRetrievalChain, where method run() is invoked on self.combine_docs_chain. 
image

When a MapReduceDocumentChain object with “return_intermediate_steps” set to True is evaluated, there will be two output keys, named “output_text” and “intermediate_steps”, respectively. Therefore, _ _call_ _() should be used to evaluate such an object. However, inside the method _call() of class BaseConversationalRetrievalChain, it is hardcoded that run() is to be used to evaluate the object, resulting in this error
image

This PR modifies the existing code of the ConversationalRetrievalChain to verify if the map reduce document chain passed into combine_docs_chain has the return_intermediate_result flag set to True, and add the intermediate result of map reduce to the output of ConversationalRetrievalChain accordingly.

 **Issue:** 
 #11209

**Dependencies:**
No new dependencies have been added.

I worked closely with @PitcairnHotdog on this PR


ccurme (changes requested):
Sorry for the late review. This needs:

- async implementation
- don't hard-code key names
- tests

PR #13549, opened 2023-11-18
feat: load all namespaces

andstu:
- **Description:** This change allows for the `MWDumpLoader` to load all namespaces including custom by default instead of only loading the [default namespaces](https://www.mediawiki.org/wiki/Help:Namespaces#Localisation).


baskaryan:
thanks @andstu!

PR #11240, opened 2023-09-29
improved rdf_graph schema generation

WessZumino:
Description: 

The RdfGraph loader provides RDF schema extraction to pass to the GraphSparqlQAChain for Q&A. The  RdfGraph also allows to provide short comments to each entity in the schema; I found this very useful to pass context or even small examples on how to use a particular class/attribute to the LLM that enhanced query generation in the GraphSparqlQAChain. 

In the current version however, adding comments leads to the generation of several artifacts that make the schema too long to fit the context window of the LLM. For example, adding few comments to a medium size graph led to a schema with 40k tokens from the original 200.  This was noted in issue  https://github.com/langchain-ai/langchain/issues/8907; although here it refers to OWL format, the problem persists in RDF. 

In this PR I am resolving the above issue by adding a filter to the original SPARQL query. Referring to the example above, the schema + comments now takes only 280 tokens. 

Another addition is the option to add a minimal list of prefixes to the schema. This is useful when the LLM is not able to infer the correct prefix from the URI or this is completely missing from the schema as I found in few use cases.

I am planning to discuss the use of comments/prefix as an addition to the current documentation in a follow up PR. 

Issue: https://github.com/langchain-ai/langchain/issues/8907,
Dependencies: N/A,
Tag maintainer: @jexp ,

PR #9136, opened 2023-08-11
fix sparql queries for relations in schema description

felixocker:
- **Description**: Fix for the SPARQL QA chain: fixed SPARQL queries for retrieving information about relations in the graph to create a textual description of the schema for the language model. This should resolve #8907
- **Issue**: #8907
- **Dependencies**: None
- **Tag maintainer**: @baskaryan, @hwchase17

Issue #8907, opened 2023-08-08
RdfGraph schema retrieval queries for the relation types are not linked by the correct comment variable 


LorenzBuehmann:
### System Info

langchain = 0.0.251
Python = 3.10.11

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [X] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

1. Create an OWL ontology called `dbpedia_sample.ttl` with the following:
``` turtle
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix wikidata: <http://www.wikidata.org/entity/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix prov: <http://www.w3.org/ns/prov#> .
@prefix : <http://dbpedia.org/ontology/> .



:Actor
    a owl:Class ;
    rdfs:comment "An actor or actress is a person who acts in a dramatic production and who works in film, television, theatre, or radio in that capacity."@en ;
    rdfs:label "actor"@en ;
    rdfs:subClassOf :Artist ;
    owl:equivalentClass wikidata:Q33999 ;
    prov:wasDerivedFrom <http://mappings.dbpedia.org/index.php/OntologyClass:Actor> .

:AdministrativeRegion
    a owl:Class ;
    rdfs:comment "A PopulatedPlace under the jurisdiction of an administrative body. This body may administer either a whole region or one or more adjacent Settlements (town administration)"@en ;
    rdfs:label "administrative region"@en ;
    rdfs:subClassOf :Region ;
    owl:equivalentClass <http://schema.org/AdministrativeArea>, wikidata:Q3455524 ;
    prov:wasDerivedFrom <http://mappings.dbpedia.org/index.php/OntologyClass:AdministrativeRegion> .

:birthPlace
    a rdf:Property, owl:ObjectProperty ;
    rdfs:comment "where the person was born"@en ;
[...]

PR #13555, opened 2023-11-18
improve sparql-qa-chain

felixocker:
**Description:**
* fixed the SPARQL queries for schema extraction - in some cases not all classes were loaded
* improved the prompt for creating natural language answers from the query results
* fixed the RDF example - the content was not retrieved from the graph DB due to the name not matching the one in the graph
* added an OWL example with a local file - simplified, but inspired by @LorenzBuehmann's example in #8907

**Issue:**
loosely related to issue #8907 and the discussion within PR #11240. the issues mentioned there were resolved in #9136, but made me notice another issue where classes were missing in the extracted schema

**Dependencies:**
none

**Tag maintainer:**
@baskaryan


docs/docs/use_cases/graph/graph_sparql_qa.ipynb
baskaryan, line 1: could we add to the docs an example of customizing the different prompts? imagine no prompt will be perfect for every use case so it's important to show how to customize

LorenzBuehmann:
Hi,

I can see two limitations resp. opportunities for improvement here:

1. using the local name of an IRI is not the intended way of providing a human readable name in RDF. The best practice is usually to use `rdfs:label` or similar properties (e.g. skos:prefLabel, skos:altLabel, ...). For example, in Wikidata all IRIs just have IDs like `https://www.wikidata.org/entity/Q5`, same holds for many biomedical domain ontologies.
2. we should add a config option to restrict the text to a particular language, e.g. to English, Spanish, etc. - even if the LLM can handle different languages, in my opinion it'S way cleaner to have just one language per prompt. And in RDF you can indeed have `rdfs:comment` literals in multiple languages. The current code would therefore return multiple rows for the same class or property with descriptions in different languages. This would also make the prompt larger than necessary. 

I hope my comments make sense to some extent.

docs/docs/use_cases/graph/graph_sparql_qa.ipynb
felixocker: yeah, sounds good. I added a section `customization` to the jupyter notebook in fd0ba32

felixocker:
@LorenzBuehmann thanks for your input, I fully agree.
I added support for labels in a062f4d and ad187a4 allows setting a `language` for labels and comments (default is no language)
Note that if a language is set, labels and comments should actually be available in that language for all notions, otherwise they may be ignored. But this seems like a reasonable assumption to me, at least for now.

felixocker:
@baskaryan could you have a look at this? I think it's ready to merge, but any feedback is appreciated.

PR #13561, opened 2023-11-18
Change docs: add -U flag (upgrade) to "pip install -U duckduckgo_search"

deedy5:
change all the lines like `pip install duckduckgo_search` to `pip install -U duckduckgo_search`


baskaryan:
out of curiosity, were you seeing some errors with older versions?

deedy5:
If everything works, it's okay.
I left only changes to the documentation to always install the latest version.

PR #13568, opened 2023-11-18
Add error rate

hinthornw:
To the in-memory outputs. Separate it out from the outputs so it's present in the dataframe.describe() results

### DIFF
--- a/libs/langchain/langchain/smith/evaluation/runner_utils.py
+++ b/libs/langchain/langchain/smith/evaluation/runner_utils.py
@@ -126,6 +126,14 @@ class TestResult(dict):
         return pd.DataFrame(records, index=indices)
 
 
+class EvalError(dict):
+    """Your architecture raised an error."""
+
+    def __init__(self, error: BaseException, **kwargs):
+        super().__init__(**kwargs)
+        self.error = error
+
+
 def _wrap_in_chain_factory(
     llm_or_chain_factory: MODEL_OR_CHAIN_FACTORY,
     dataset_name: str = "<my_dataset>",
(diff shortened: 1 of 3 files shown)


libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -988,21 +1008,26 @@ def _collect_test_results(
...
         feedback = all_eval_results.get(str(example.id), [])
         results[str(example.id)] = {
-            "output": output,
             "input": example.inputs,
             "feedback": feedback,
-            "execution_time": execution_time,
+            "execution_time": all_execution_time.get(str(example.id)),
         }
+        if isinstance(output, EvalError):
+            results[str(example.id)]["error"] = output.error
```
eyurtsev, line 1028: Was this `Error` previously? Could this end up being a breaking change?

libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -112,20 +105,46 @@ def to_dataframe(self) -> pd.DataFrame:
...
+                }
+            )
             records.append(r)
             indices.append(example_id)
 
         return pd.DataFrame(records, index=indices)
 
 
+class EvalError(dict):
+    """Your architecture raised an error."""
```
eyurtsev, line 136: Excuse moi

eyurtsev (approved):

libs/langchain/langchain/smith/evaluation/runner_utils.py
hinthornw: hm ya will update

PR #13571, opened 2023-11-19
MMR with Vectara

ofermend:
- **Description:** Vectara implements MMR in the backend. This allows to invoke that MMR natively.
  - **Twitter handle:** @ofermend

### DIFF
--- a/libs/langchain/langchain/vectorstores/vectara.py
+++ b/libs/langchain/langchain/vectorstores/vectara.py
@@ -281,7 +281,8 @@ class Vectara(VectorStore):
                 to minimum diversity and 1 to maximum diversity.
                 Defaults to 0.3.
                 Note: mmr_diversity_bias is equivalent 1-lambda_mult
-                where lambda_mult is the value often used in max_marginall_relevance_search()
+                where lambda_mult is the value often used in 
+                max_marginal_relevance_search()
                 We chose to use that since we believe it's more intuitive to the user.
 
         Returns:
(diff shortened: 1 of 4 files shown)


libs/langchain/langchain/vectorstores/vectara.py
```diff
@@ -251,6 +254,9 @@ def similarity_search_with_score(
         filter: Optional[str] = None,
         score_threshold: Optional[float] = None,
         n_sentence_context: int = 2,
+        mmr: bool = False,
```
baskaryan, line 257: similarity_search and mmr are meant to be two separate techniques, should we just put the mmr implementation in the mmr function? that's how all other vector stores do it

ofermend: I made it so that similarity_search_with_score is the main function that calls Vectara backend, and all others call types (e.g. similarity_search or mmr) call it with specific parameters.
I think in the path you are suggesting, perhaps move this functionality into an internal method (e.g. _vectara_search()) and all formal functions (like similarity_search, similarity_search_with_score, mmr, etc) will call that for a specific implementation.
Is that what you had in mind?

ofermend: @baskaryan any thoughts / guidance?

ofermend: Thinking about this more, I do think MMR is really an option.
I'm considering adding (in separate PR) support for two search types: default (as today) and summarization (returns both documents and a summarization so removed the need for an additional LLM call) as suggested here: https://github.com/langchain-ai/langchainjs/pull/2206

MMR can be used with either of those, and in this context I don't think it'd be good to define it as an additional searchType. 

thoughts?

ofermend: @baskaryan - any other thoughts here?

efriis:
Closing in favor of #14408

PR #13573, opened 2023-11-19
Update geodataframe.py

Hamzahmed:
here it is validating shapely.geometry.point.Point:       if not isinstance(data_frame[page_content_column].iloc[0], gpd.GeoSeries): raise ValueError(
f"Expected data_frame[{page_content_column}] to be a GeoSeries"         you need it to validate the geoSeries and not the shapely.geometry.point.Point

if not isinstance(data_frame[page_content_column], gpd.GeoSeries):
            raise ValueError(
                f"Expected data_frame[{page_content_column}] to be a GeoSeries"


baskaryan:
cc @brendancol

hwchase17 (approved):

PR #13576, opened 2023-11-19
openapi/planner.py: Deal with json in markdown output cases

arnaudgelas:
- **Description:** In openapi/planner deal with json in markdown output cases
  - **Issue:**  In some cases LLMs could return json in markdown which can't be loaded.  
  - **Dependencies:**
  - **Tag maintainer:** @eyurtsev
  - **Twitter handle:**


baskaryan:
should we use `langchain.output_parsers.json.parse_json_markdown`?

arnaudgelas:
@baskaryan Right, I was not aware of this function. Thanks for the pointer!

hwchase17 (approved):

PR #13587, opened 2023-11-20
API Reference building script update

leo-gan:
The namespaces like `langchain.agents.format_scratchpad` clogging the API Reference sidebar.
This change removes those 3-level namespaces from sidebar (this issue was discussed with @efriis )


leo-gan:
> @leo-gan could you resolve conflicts and also confirm that this will still document those classes, just within the same `agents` page? Or does this ignore classes 3+ levels deep?

@efriis Yes, exactly. All elements in those 3-level ns groups are presented in 2-level ns groups.
For example, 
`langchain.agents.format_scratchpad` group / `agents.format_scratchpad.log.format_log_to_str(...)` function
is duplicated as
`langchain.agents` group / `agents.format_scratchpad.log.format_log_to_str(...)` function

All elements in the 3-level are duplicated in the 2-level ns groups

efriis (approved):
rst diffs look good. Thanks!

PR #13399, opened 2023-11-15
Update PlanAndExecute agent to include intermediate steps

adrijanik:
- **Description:** update PlanAndExecute agent to include intermediate steps. 
  - **Issue:** [issue 4485](https://github.com/langchain-ai/langchain/issues/4485).

PR #13591, opened 2023-11-20
Adrijanik patch 1

adrijanik:


hwchase17 (commented):
can we make add a configuration option `return_intermediate_steps: bool = False` and then only do this if that is set?

baskaryan:
hey @adrijanik, were you still working on this?

baskaryan:
duplicate of #13399

PR #13594, opened 2023-11-20
Update mongodb_atlas.py to fix ImportError: cannot import name 'DriverInfo' from 'pymongo'

ruanwz:


hwchase17:
thanks, was updated in other PR! appreciate it though

PR #13618, opened 2023-11-20
office365 toolkit bug fixes

leo-gan:
Several bug fixes:
- emails: instead of `bcc` the `cc` is used.
- errors in the truncation descriptions
- no truncation of the `message_search`
Several updates:
- generalized UTC format 
- truncation limit can be changed now in _call()


hwchase17 (approved):
thanks Leo!

PR #13620, opened 2023-11-20
VoyageEmbeddings embed_general_texts

thomas0809:
- **Description:** add method  embed_general_texts in VoyageEmebddings to support input_type
  - **Issue:** 
  - **Dependencies:** 
  - **Tag maintainer:** 
  - **Twitter handle:** @Voyage_AI_


libs/langchain/langchain/embeddings/voyageai.py
```diff
@@ -167,9 +173,24 @@ def embed_query(self, text: str) -> List[float]:
         Args:
             text: The text to embed.
 
+        Returns:
+            Embedding for the text.
+        """
+        return self._get_embeddings([text], input_type="query")[0]
+
+    def embed_general_texts(
+        self, texts: List[str], input_type: Optional[str] = None
```
baskaryan, line 182: nit: could we separate args and kwargs with *
```python
self, texts: List[str], *, input_type: Optional[str]=None
```

baskaryan (approved):

PR #13623, opened 2023-11-20
Add core as dep

baskaryan:


libs/langchain/pyproject.toml
```diff
@@ -12,11 +12,16 @@ langchain-server = "langchain.server:main"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
+langchain-core = "^0.0.1"
```
efriis, line 15: Should we >= while we're in the 0.0.x?
```suggestion
langchain-core = ">=0.0.1,<0.1"
```

baskaryan: if we make breaking changes in 0.0.2 (think we will) then won't be 0.0.1 compatible

efriis (approved):

PR #13627, opened 2023-11-20
Core refac

baskaryan:
proposed changes:
- remove langchain_core/schema since no clear distinction b/n schema and non-schema modules
- make every module that doesn't end in -y plural
- where easy have 1-2 classes per file
- no more than one level of nesting


libs/core/langchain_core/output_parsers/str.py
```diff
@@ -0,0 +1,23 @@
...
+    def _type(self) -> str:
+        """Return the output parser type for serialization."""
+        return "default"
+
+    def parse(self, text: str) -> str:
+        """Returns the input text with no changes."""
+        return text
+
+
+# TODO: Deprecate
```
efriis, line 22: can we remove or leave in langchain with this?

efriis (commented):
i dig

libs/core/langchain_core/_api/__init__.py
```diff
@@ -16,9 +16,12 @@
     surface_langchain_deprecation_warnings,
     warn_deprecated,
 )
+from .path import as_import_path, get_relative_path
```
efriis, line 19: any reason for the relative instead of absolute?

libs/core/langchain_core/callbacks/manager.py
```diff
@@ -41,23 +45,16 @@
...
-from langchain_core.callbacks.tracers.stdout import ConsoleCallbackHandler
-from langchain_core.schema import (
-    AgentAction,
-    AgentFinish,
-    Document,
-    LLMResult,
-)
-from langchain_core.schema.messages import BaseMessage, get_buffer_string
-from langchain_core.schema.output import ChatGenerationChunk, GenerationChunk
+from langchain_core.tracers.langchain_v1 import LangChainTracerV1
```
efriis, line 55: leave some of the old tracer stuff in langchain?

libs/core/langchain_core/exceptions.py
efriis, line 1: Would it be valuable to have exceptions per-module instead of across all here?

libs/core/langchain_core/language_models/base.py
```diff
@@ -74,8 +74,8 @@ class BaseLanguageModel(
     @property
     def InputType(self) -> TypeAlias:
         """Get the input type for this runnable."""
-        from langchain_core.prompts.base import StringPromptValue
         from langchain_core.prompts.chat import ChatPromptValueConcrete
+        from langchain_core.prompts.str import StringPromptValue
```
efriis, line 78: any issues naming this `str` in case someone tries to import `from langchain_core.prompts import str`?

libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -1539,7 +1541,7 @@ def test_with_listeners(mocker: MockerFixture) -> None:
     )
     chat = FakeListChatModel(responses=["foo"])
 
-    chain = prompt | chat
+    chain: Runnable = prompt | chat
```
efriis, line 1544: Why are these types desirable?

libs/langchain/langchain/agents/output_parsers/openai_tools.py
efriis, line 1: core or regular?

efriis (approved):
A few questions and otherwise looks good!

libs/core/tests/unit_tests/runnables/test_runnable.py
baskaryan: making linter happy

libs/core/langchain_core/language_models/base.py
baskaryan: mm good point. someone could `import str as lc_str` but probs best to avoid that in first place, will make it string

libs/core/langchain_core/exceptions.py
baskaryan: i kinda like exceptions being in one, makes it really easy to know where to import from. but no strong feelings

libs/core/langchain_core/_api/__init__.py
baskaryan: was copying above, guessing linter complains about importing from private module?

libs/langchain/langchain/agents/output_parsers/openai_tools.py
baskaryan: what's the question?

efriis: Oh lol was this not core ever? For some reason I thought this was a change in core, but looking at the heading looks like it wasn't

efriis: I thought you'd moved `openai_tools` to core, and was asking if it should stay in regular, but looks like it was regular all along

PR #13633, opened 2023-11-21
remove openai api key from cookbook

gyliu513:


baskaryan:
thanks @gyliu513! @adolkhan maybe yours, may want to delete api key

AdkSarsen:
Thanks a lot @baskaryan!

PR #13645, opened 2023-11-21
Added partial_variables to BaseStringMessagePromptTemplate.from_template(...)

jponf:
**Description:** BaseStringMessagePromptTemplate.from_template was passing the value of partial_variables into cls(...) via **kwargs, rather than passing it to PromptTemplate.from_template. Which resulted in those *partial_variables being* lost and becoming required *input_variables*.

**twitter handle**: @josep_pon


baskaryan:
thank you @jponf

PR #13652, opened 2023-11-21
fix the problem that redis vector store will change metadata with val…

dandanwei:
- **Description:** This commit fixed the problem that Redis vector store will change the value of a metadata from 0 to empty when saving the document, which should be an un-intended behavior. 
  - **Issue:** N/A
  - **Dependencies:** N/A


baskaryan:
cc @Spartee

Spartee (approved):
Good catch. LGTM @baskaryan @tylerhutcherson

baskaryan:
thank you @dandanwei!

PR #13661, opened 2023-11-21
Add CometTracer

alexkuzmik:
Hi! I'm Alex, Python SDK Team Lead from [Comet](https://www.comet.com/site/).

This PR contains our new integration between langchain and Comet - `CometTracer` class which uses new `comet_llm` python package for submitting data to Comet.

No additional dependencies for the langchain package are required directly, but if the user wants to use `CometTracer`, `comet-llm>=2.0.0` should be installed. Otherwise an exception will be raised from `CometTracer.__init__`.

A test for the feature is included.

There is also an already existing callback (and .ipynb file with example) which ideally should be deprecated in favor of a new tracer. I wasn't sure how exactly you'd prefer to do it. For example we could open a separate PR for that.

I'm open to your ideas :)


alexkuzmik:
@baskaryan

hwchase17 (approved):

PR #13691, opened 2023-11-21
Template Metadata pt 1

efriis:


templates/basic-critique-revise/README.md
efriis, line 1: this was just a one off because it was missing

PR #13693, opened 2023-11-22
fix merge None value and non None value error

mengjincn:


mengjincn:
there is another same pr

Issue #13683, opened 2023-11-21
[Issue] Error with Bedrock when AWS region is not specified or not in environment variable


kaihe-stori:
### System Info

langchain version: 0.0.339
python: 3.9.17

### Who can help?

@hwchase17 @agola11 

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

When creating a Bedrock model without `region_name`
```py
llm_bedrock = Bedrock(model_id='anthropic.claude-instant-v1')
```
It errors out
```py
ValidationError: 1 validation error for Bedrock
__root__
  Could not load credentials to authenticate with AWS client. Please check that credentials in the specified profile name are valid. (type=value_error)
```

This works
```py
llm_bedrock = Bedrock(model_id='anthropic.claude-instant-v1', region_name='us-west-2')
```

The problem is from https://github.com/langchain-ai/langchain/blob/bfb980b96800020d90c9362aaad40d8817636711/libs/langchain/langchain/llms/bedrock.py#L203-L207
where `get_from_dict_or_env` raises an exception if region is neither specified nor in env and default is not set.

To fix it, change default from `None` to `session.region_name`.

### Expected behavior

No error is raised.

PR #13694, opened 2023-11-22
Set default region from boto3 session for Bedrock

Falydoor:
- **Description:** Set default region from boto3 session for Bedrock 
- **Issue:** #13683


jsgsamzn:
+1 while there are potential workarounds for this, this amounts to a breaking change for anyone with cross-region and/or bare-bones Bedrock clients.

hwchase17 (approved):
thanks!

PR #13714, opened 2023-11-22
Add option to prefix config keys in configurable_alts

nfcampos:

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -1205,6 +1205,7 @@ class RunnableSerializable(Serializable, Runnable[Input, Output]):
         self,
         which: ConfigurableField,
         default_key: str = "default",
+        prefix_keys: bool = False,
         **kwargs: Union[Runnable[Input, Output], Callable[[], Runnable[Input, Output]]],
     ) -> RunnableSerializable[Input, Output]:
         from langchain_core.runnables.configurable import (
@@ -1212,7 +1213,11 @@ class RunnableSerializable(Serializable, Runnable[Input, Output]):
         )
 
         return RunnableConfigurableAlternatives(
-            which=which, default=self, alternatives=kwargs, default_key=default_key
+            which=which,
+            default=self,
+            alternatives=kwargs,
+            default_key=default_key,
+            prefix_keys=prefix_keys,
         )
 
 
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -1204,6 +1204,7 @@ class RunnableSerializable(Serializable, Runnable[Input, Output]):
     def configurable_alternatives(
         self,
         which: ConfigurableField,
+        *,
         default_key: str = "default",
         prefix_keys: bool = False,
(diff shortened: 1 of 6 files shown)


libs/core/langchain_core/runnables/configurable.py
```diff
@@ -299,6 +300,8 @@ class RunnableConfigurableAlternatives(DynamicRunnable[Input, Output]):
 
     default_key: str = "default"
 
+    prefix_keys: bool
```
eyurtsev, line 303: Could you add a doc string for this key?

libs/core/langchain_core/runnables/base.py
```diff
@@ -1205,14 +1205,19 @@ def configurable_alternatives(
         self,
         which: ConfigurableField,
         default_key: str = "default",
+        prefix_keys: bool = False,
```
eyurtsev, line 1208: Could we add * to force invocation by name?

libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -1020,6 +1020,115 @@ def test_configurable_alts_factory() -> None:
...
+            name="Chat Responses",
+            options={
+                "hello": "A good morning to you!",
+                "bye": "See you later!",
+                "helpful": "How can I help you?",
+            },
+            default=["hello", "bye"],
+        ),
+        sleep=ConfigurableField(
+            id="chat_sleep",
```
eyurtsev, line 1036: Was surprising to find `sleep` in here, didn't realize it was an actual attribute on fake list chat model (it doesn't sound like it would be, but it is), maybe a comment could be helpful

libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -1020,6 +1020,115 @@ def test_configurable_alts_factory() -> None:
     assert fake_llm.with_config(configurable={"llm": "chat"}).invoke("...") == "b"
 
 
+def test_configurable_fields_prefix_keys() -> None:
```
eyurtsev, line 1023: There's a lot of stuff packed into here.. without doc-strings explaining how some of the code is supposed to work and why, i think it's hard to understand what's going on

libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -1020,6 +1020,115 @@ def test_configurable_alts_factory() -> None:
...
+            id="responses",
+            name="Chat Responses",
+            options={
+                "hello": "A good morning to you!",
+                "bye": "See you later!",
+                "helpful": "How can I help you?",
+            },
+            default=["hello", "bye"],
+        ),
+        sleep=ConfigurableField(
```
eyurtsev, line 1035: What happens if there's another configurable alternative that does not support `sleep`, but it's still makred as `is_shared`?

eyurtsev (approved):

libs/core/tests/unit_tests/runnables/test_runnable.py
nfcampos: In that case the other one doesnt use it (ie what happens already before this PR)

PR #13720, opened 2023-11-22
Update summarization.ipynb

k3ybladewielder:
After testing the example, I saw that the input_variables was docs and not doc_summaries


hwchase17:
i think this is address in separate pr

k3ybladewielder:
I think I ended up pasting the commit description with information copied from elsewhere. But, the change made was only the doc_summaries for docs. After testing the example, I saw that the input_variables was docs and not doc_summaries

k3ybladewielder:
Also i saw its alreade fixed in #13726

PR #13740, opened 2023-11-22
docs[patch]: link to LangSmith docs

leo-gan:
It happens that there is no link to the LangSmith Docs from the LangChain Docs.
Added this link


leo-gan:
@efriis FYI Eric, for some reason, Vercel didn't build docs.

PR #13741, opened 2023-11-22
docs: moved up Chains menu

leo-gan:
`Chains` now is inside `More` menu. It is the most important module of LangChain, no? The whole package is "Lang CHAIN" and the Chains are hidden somewhere, no kidding.
`Chains` menu is moved up in the top menu hierarchy.


efriis (commented):
The primary approach moving forward will be using LCEL instead of these, so let's leave it in the collapsed menu.

@baskaryan thoughts on renaming the section "Legacy Chains" to make it clearer?

leo-gan:
Moving LCEL up as the primary usage is legitimate. Still, we have `Agents` in this level but not `Chains`. I argue that `Chains` has the same, at least, importance then `Agents`. 
And LCEL is the `Language CHAIN Extression Language`, not LAEL, not LPEL. Chain is on the top of LangChain hierarchy. It is even on the logo. 🦜🔗 I'm not pushing up parrot yet ;)

hwchase17:
@leo-gan in practice, most of the chains are so application specific they are in use cases. The ones that arent are deprecated in favor of LCEL

For agents, the "Agents" sectoin can be thought of the equivalent of LCEL, but there are agent toolkits in the Use Case/Integration sections

PR #13742, opened 2023-11-22
INFRA: run LC ci after core changes

baskaryan:


efriis (approved):

eyurtsev (approved):

.github/workflows/langchain_experimental_ci.yml
```diff
@@ -13,6 +13,8 @@ on:
       - '.github/workflows/langchain_experimental_ci.yml'
       - 'libs/*'
       - 'libs/experimental/**'
+      - 'libs/langchain/**'
```
eyurtsev, line 16: I assume that we can't assume order dependency here?

baskaryan: what do you mean?

PR #13763, opened 2023-11-23
small fix matching engine AttributeError - object has no attribute

saiwas:
This PR is fixing an attributeError: object endpoint has no attribute "_public_match_client" when using gcp matching engine with private VPC network.

### DIFF
--- a/libs/langchain/langchain/vectorstores/matching_engine.py
+++ b/libs/langchain/langchain/vectorstores/matching_engine.py
@@ -232,7 +232,7 @@ class MatchingEngine(VectorStore):
         filter = filter or []
 
         # If the endpoint is public we use the find_neighbors function.
-        if self.endpoint._public_match_client:
+        if hasattr(self.endpoint, "_public_match_client"):
             response = self.endpoint.find_neighbors(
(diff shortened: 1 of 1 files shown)


saiwas:
Issue in short, we're using the latest version of langchain with GCP product [Matching Engine](https://cloud.google.com/vertex-ai/docs/vector-search/overview) in a private VPC network design. However, it always hit error **object endpoint has no attribute "_public_match_client"** because we don't deploy the matching engine index to a public endpoint.

A quick fix for it is using "hasattr" to check the "endpoint" whether has attribute "_public_match_client".

libs/langchain/langchain/vectorstores/matching_engine.py
```diff
@@ -232,7 +232,7 @@ def similarity_search_by_vector_with_score(
         filter = filter or []
 
         # If the endpoint is public we use the find_neighbors function.
-        if self.endpoint._public_match_client:
+        if hasattr(self.endpoint, "_public_match_client"):
```
hwchase17, line 235: this is checking for different things, i think we still want want to check that this is True or non Nully right?

hwchase17 (approved):

PR #13764, opened 2023-11-23
Update agents\conversational\output_parser.py

lilong669:
The regular expression  regex = r"Action: (.*?)[\n]*Action Input: (.*)", with .* being the default greedy match.This results in the action_input variable not including the part after the newline character.


hwchase17 (commented):
not sure i fully understand, we should add unit tests

hwchase17:
fixed in https://github.com/langchain-ai/langchain/pull/15082

PR #13229, opened 2023-11-10
Fix "additional_kwargs["arguments"] already exists in this message, but with a different type." for Azure openai API. 

superken01:
- **Description:** This PR addresses an issue with the OpenAI API streaming response, where initially the key (arguments) is provided but the value is None. Subsequently, it updates with {"arguments": "{\n"}, leading to a type inconsistency that causes an exception. The specific error encountered is ValueError: additional_kwargs["arguments"] already exists in this message, but with a different type. This change aims to resolve this inconsistency and ensure smooth API interactions.
  - **Issue:** None.
  - **Dependencies:** None.
  - **Tag maintainer:** @eyurtsev,

PR #13768, opened 2023-11-23
Fix Azure Openai function calling in streaming mode

hendrik-hog:
- **Description**: This PR addresses an issue with the OpenAI API streaming response, where initially the key (arguments) is provided but the value is None. Subsequently, it updates with {"arguments": "{\n"}, leading to a type inconsistency that causes an exception. The specific error encountered is ValueError: additional_kwargs["arguments"] already exists in this message, but with a different type. This change aims to resolve this inconsistency and ensure smooth API interactions.
- **Issue**: None.
- **Dependencies**: None.
- **Tag maintainer**: @eyurtsev

This is an updated version of #13229 based on the refactored code. Credit goes to @superken01.


hwchase17 (approved):

dkindlund:
Hey @hwchase17 and @hendrik-hog , was this fix also ported to https://github.com/langchain-ai/langchainjs ?

I think I'm running into this same bug within the langchainjs package, but I'm not sure if it exists... can you check around these lines:
https://github.com/langchain-ai/langchainjs/blob/main/langchain-core/src/messages/index.ts#L240

PR #13769, opened 2023-11-23
Repair Wikipedia document loader `load_max_docs` and improve test coverage.

queenvictoria:
**Description:** 

Repair Wikipedia document loader `load_max_docs` and improve test coverage.

**Issue:** 

The Wikipedia document loader was not respecting the `load_max_docs` paramater (not reported) and would always return a maximum of 10 documents. This is because the API wrapper (in `utilities/wikipedia.py`) wasn't passing `top_k_results` to the underlying [Wikipedia library](https://wikipedia.readthedocs.io/en/latest/code.html#module-wikipedia). By default this library returns 10 results.

The default number of results for the document loader has been reduced from 100 to 25. This is because loading 100 results takes a very long time and is an inconvenient default. It should possibly be 10.

In addition, the documentation for the loader reported that there was a hard limit (300) on the number of documents returned. In actuality 300 is the maximum Wikipedia query character length set by the API wrapper.

Tests have been added for the document loader (previously missing) and to test the correct numbers of documents are being returned by each class, both by default, and when overridden. Also repaired is the `assert_docs` test which has been updated to correctly test for the default metadata (which includes `source` in recent releases).

**Dependencies:** 
nil

**Tag maintainer:**
@leo-gan

**Twitter handle:**
@queenvictoria

### DIFF
--- a/libs/langchain/langchain/document_loaders/wikipedia.py
+++ b/libs/langchain/langchain/document_loaders/wikipedia.py
@@ -8,7 +8,7 @@ from langchain.utilities.wikipedia import WikipediaAPIWrapper
 class WikipediaLoader(BaseLoader):
     """Load from `Wikipedia`.
 
-    The hard limit on the number of downloaded Documents is 300 for now.
+    The hard limit on the length of the query is 300 for now.
 
     Each wiki page represents one Document.
     """
@@ -17,7 +17,7 @@ class WikipediaLoader(BaseLoader):
         self,
         query: str,
         lang: str = "en",
-        load_max_docs: Optional[int] = 100,
+        load_max_docs: Optional[int] = 25,
         load_all_available_meta: Optional[bool] = False,
         doc_content_chars_max: Optional[int] = 4000,
     ):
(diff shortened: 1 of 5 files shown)


leo-gan:
@queenvictoria Thanks for fixing this!

leo-gan (approved):
lgtm

leo-gan (approved):
lgtm

libs/langchain/langchain/document_loaders/wikipedia.py
```diff
@@ -17,7 +17,7 @@ def __init__(
         self,
         query: str,
         lang: str = "en",
-        load_max_docs: Optional[int] = 100,
+        load_max_docs: Optional[int] = 25,
```
baskaryan, line 20: why change default?

leo-gan: @baskaryan It is described in the PR description ^

leo-gan (approved):
lgtm

PR #13781, opened 2023-11-23
Allow returning tokens usage counts

Bassem-Yacoube:
Small change to allow returning tokens usage counts from OctoAI endpoints


libs/langchain/langchain/llms/octoai_endpoint.py
```diff
@@ -132,6 +132,11 @@ def _call(
                 # Send the request using the OctoAI client
                 output = octoai_client.infer(self.endpoint_url, parameter_payload)
                 text = output.get("choices")[0].get("message").get("content")
+                if (
```
hwchase17, line 135: we want to append the tokens used to the completion string? that seems weird... why?

PR #13788, opened 2023-11-23
`kaggle` document loader

leo-gan:
Added the `Kaggle` datasets as a DocumentLoader

### DIFF
--- a/docs/docs/integrations/providers/kaggle.mdx
+++ b/docs/docs/integrations/providers/kaggle.mdx
@@ -1,0 +1,17 @@
+# Kaggle
+
+>[Kaggle](https://www.kaggle.com/) is a data science competition platform and online community of data scientists and machine learning practitioners under `Google LLC`. `Kaggle` enables users to find and publish datasets, explore and build models in a web-based data science environment, work with other data scientists and machine learning engineers, and enter competitions to solve data science challenges.
+
+## Installation and Setup
+
+You need to install `kaggle` python package.
+
+```bash
+pip install kaggle
+```
+
+## Document Loader
+
+The Kaggle has lots of [datasets](https://www.kaggle.com/datasets) that can be used to train and evaluate your LLM chains.
+
+See a [usage example and detailed installation instructions](/docs/integrations/document_loaders/kaggle_dataset).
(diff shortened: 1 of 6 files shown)


libs/langchain/langchain/document_loaders/kaggle_dataset.py
```diff
@@ -0,0 +1,74 @@
...
+
+from langchain.docstore.document import Document
+from langchain.document_loaders.base import BaseLoader
+
+
+class KaggleDatasetLoader(BaseLoader):
+    """Load from `Kaggle` datasets.
+
+    Follow these steps to use this loader:
+    - Register a Kaggle account and create an API token to use this loader.
```
efriis, line 11: I haven't used the kaggle package before, and would it be possible to automate these steps in the loader instead of having it be a csv loader with some logic around the kaggle dataset format?

docs/docs/integrations/providers/kaggle.mdx
```diff
@@ -0,0 +1,17 @@
+# Kaggle
+
+>[Kaggle](https://www.kaggle.com/) is a data science competition platform and online community of data scientists and machine learning practitioners under `Google LLC`. `Kaggle` enables users to find and publish datasets, explore and build models in a web-based data science environment, work with other data scientists and machine learning engineers, and enter competitions to solve data science challenges.
+
+## Installation and Setup
+
+You need to install `kaggle` python package.
```
efriis, line 7: In the current implementation, I don't think this is true.

efriis (commented):
Little suggestion to make this even more convenient for folks using datasets from kaggle!

docs/docs/integrations/providers/kaggle.mdx
leo-gan: A good catch! Thanks! It's a reminder from experimentation.

libs/langchain/langchain/document_loaders/kaggle_dataset.py
leo-gan: It is possible. But I don't have time for it right now. I'm switching this PR to draft. Maybe I'll find time for automate it.

PR #13820, opened 2023-11-24
Adds the Yellowbrick Data Warehouse as a supported vector store

markcusack:
- **Description** An integration to allow the Yellowbrick Data Warehouse to function as a vector store


hwchase17 (approved):
this looks pretty clean and solid to me! an example notebook might be good, but not necc!

markcusack:
> this looks pretty clean and solid to me! an example notebook might be good, but not necc!

Thanks! Just raised a PR to add a notebook.

PR #13831, opened 2023-11-25
docs: `Hugging Face` platform page

leo-gan:
`Hugging Face` is definitely a platform. It includes many integrations for many modules (LLM, Embedding, DocumentLoader, Tool)
So, a doc page was added that defines Hugging Face as a platform.


efriis (commented):
Hey @leo-gan ! These two provider sections just show up differently in the sidebar, so the important distinction is a bit different than the naming of platform vs provider. Let me chat with the team about which providers we want to feature in the default-expanded section!

efriis:
I think this also has some content changes. In the meantime, would it be ok to merge this in with just the content update in the existing page?

leo-gan:
@efris I've discussed it with @hwchase17 and persuaded him.
>Harrison Chase  [2:48 PM]
>Yeah that’s fair. To be honest, we don’t see nearly as many people using them as the other providers. But I do want to push open source, so I’m down to add them

hwchase17 (approved):

PR #13834, opened 2023-11-25
Astra DB, LLM cache classes (exact-match and semantic cache)

hemidactylus:
This PR provides idiomatic implementations for the exact-match and the semantic LLM caches using Astra DB as backend through the database's HTTP JSON API. These caches require the `astrapy` library as dependency.

Comes with integration tests and example usage in the `llm_cache.ipynb` in the docs.

@baskaryan this is the Astra DB counterpart for the Cassandra classes you merged some time ago, tagging you for your familiarity with the topic. Thank you!


baskaryan:
again looks great, thanks @hemidactylus!

PR #13836, opened 2023-11-25
Astra DB, chat message history

hemidactylus:
This PR adds a chat message history component that uses Astra DB for persistence through the JSON API.
The `astrapy` package is required for this class to work.

I have added tests and a small notebook, and updated the relevant references in the other docs pages.

(@rlancemartin this is the counterpart of the Cassandra equivalent class you so helpfully reviewed back at the end of June)

Thank you!


baskaryan:
looks awesome, thanks @hemidactylus!

PR #13848, opened 2023-11-25
🎞️ add json format to OllamaCommon class

shroominic:
The Ollama Client now allows a pure [json mode](https://github.com/jmorganca/ollama/blob/main/docs/api.md#request-json-mode) which I just added to the OllamaCommon class.

This allows using it like in this example:
Screenshot 2023-11-25 at 14 11 19


baskaryan:
cc @jmorganca

shroominic:
Additionally, I added ChatLlamaCpp now. Not sure if this was the right move to add it to the same PR but its booth similar.

shroominic:
I tried to use create_chat_completion but i realized that its a bit more complex because the Dict Messages Schema changed because of the new multimodal interface and didn't know how to handle this best because LlamaCpp supports multimodal stuff but there is not a lot documentation on this...

Reverted and caused a git chaos but a squash should solve this :D

PR #13851, opened 2023-11-25
Reduce the number of tokens required to describe a Cypher/Neo4j schema

tomasonjo:
Instead of using JSON-like syntax to describe node and relationship properties we changed to a shorter and more concise schema description

Old:

```
        Node properties are the following:
        [{'properties': [{'property': 'name', 'type': 'STRING'}], 'labels': 'Movie'}, {'properties': [{'property': 'name', 'type': 'STRING'}], 'labels': 'Actor'}]
        Relationship properties are the following:
        []
        The relationships are the following:
        ['(:Actor)-[:ACTED_IN]->(:Movie)']
```

New:

```
Node properties are the following:
Movie {name: STRING},Actor {name: STRING}
Relationship properties are the following:

The relationships are the following:
(:Actor)-[:ACTED_IN]->(:Movie)
```

### DIFF
--- a/docs/docs/use_cases/graph/graph_cypher_qa.ipynb
+++ b/docs/docs/use_cases/graph/graph_cypher_qa.ipynb
@@ -48,16 +48,7 @@
    "execution_count": 2,
    "id": "0928915d",
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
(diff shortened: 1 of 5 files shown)


efriis (approved):
Thanks @tomasonjo !

PR #13853, opened 2023-11-25
RELEASE: core 0.0.6

baskaryan:


natsukium:
Thanks for maintaining these great packages.
Is it possible to tag langchain-core as `core/0.0.6` on GitHub?
I have packaged and distributed langchain with [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) and need to package this langchain-core for the update as well.
We would like to build the package from GitHub sources as much as possible, and it would be easier to package and track changes if it is tagged.

PR #13866, opened 2023-11-26
Update PGVector __query_collection with extended filters

Frank995:
- **Description:** current SQL filtering capabilities of PGVector document matching are limited to "is equal" and "is in", this extends it and generalise the code to any (potential) future other matching type 
  - **Dependencies:** none
  - **Tag maintainer:** @hwchase17 ,


Frank995:
Closed as functionalityy has been implemented recently

PR #13875, opened 2023-11-26
max length attribute for spacy splitter for large docs

kunal8164705:
For large size  documents spacy splitter doesn't work it throws an error as shown in  below screenshot.
Reason its default max_length is 1000000 and there is no option to increase it. So i added it in this PR. 

image


kunal8164705:
@baskaryan kindly merge this PR.

hwchase17 (approved):

PR #13879, opened 2023-11-26
Make contributing to docs easier

daertommy:
With this PR readers can suggest changes to the docs with an easy to use WYSIWYG markdown editor.

[This](https://holocron.so/github/pr/langchain-ai/langchain/master/editor) is how the editor looks like for this repo.

https://github.com/remorses/docusaurus-example/assets/31321188/43c8bc68-3668-4a25-bf1c-a70eceab7bcb

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -98,6 +98,9 @@ const config = {
             });
             return sidebarItems;
           },
+          editUrl: ({ docPath }) => {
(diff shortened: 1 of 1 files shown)


hwchase17:
pretty cool! we're currently thinking of revamping docs - we'll keep this option in mind when doing so. I think for something like this, we want to think it through a bit more though

daertommy:
Ok thank you

PR #13883, opened 2023-11-26
add skeleton of thought

hwchase17:


templates/skeleton-of-thought/pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[tool.poetry]
+name = "skeleton-of-thought"
+version = "0.0.1"
+description = ""
```
efriis, line 4: quick reminder to add the new fields. Will add a CI check to reject templates that are missing a description and/or the `tool.templates-hub` sections

Issue #13887, opened 2023-11-27
ObsidianLoader fails when encountering template variables in the properties frontmatter of a file


ealt:
### System Info

langchain version 0.0.340
Python version: 3.11.5

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

1. Create an Obsidian template with a [template variable](https://help.obsidian.md/Plugins/Templates#Template+variables) in the [properties](https://help.obsidian.md/Editing+and+formatting/Properties#Property+format) section of the file.
2. Attempt to load a directory containing that template file using [ObsidianLoader](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/document_loaders/obsidian.py).

```shell
$ echo -e "---\nyear: {{date:YYYY}}\n---" > vault/template.md
$ python
>>> from langchain.document_loaders.obsidian import ObsidianLoader
>>> loader = ObsidianLoader('vault')
>>> loader.load()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/.conda/envs/localai/lib/python3.11/site-packages/langchain/document_loaders/obsidian.py", line 115, in load
    front_matter = self._parse_front_matter(text)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/localai/lib/python3.11/site-packages/langchain/document_loaders/obsidian.py", line 48, in _parse_front_matter
    front_matter = yaml.safe_load(match.group(1))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/localai/lib/python3.11/site-packages/yaml/__init__.py", line 125, in safe_load
    return load(stream, SafeLoader)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/localai/lib/python3.11/site-packages/yaml/__init__.py", line 81, in load
    return loader.get_single_data()
[...]

PR #13888, opened 2023-11-27
Fixes error loading Obsidian templates

ealt:
- **Description:** Obsidian templates can include [variables](https://help.obsidian.md/Plugins/Templates#Template+variables) using double curly braces.  `ObsidianLoader` uses PyYaml to parse the frontmatter of documents. This parsing throws an error when encountering variables' curly braces. This is avoided by temporarily substituting safe strings before parsing.
  - **Issue:** #13887


libs/langchain/langchain/document_loaders/obsidian.py
```diff
@@ -44,8 +64,14 @@ def _parse_front_matter(self, content: str) -> dict:
         if not match:
             return {}
 
+        self._placeholders: Dict[str, str] = {}
```
baskaryan, line 67: does this need to be an attribute on the object? can we just pass it in to restore_template_vars and replace_template_var? prefer to avoid mutating state within the DocumentLoader object unless necessary

ealt:
I am not sure why the Vercel deployment failed, I'm getting a 404 page when I try to follow the links. Any idea if what the problem could be?

ealt:
@baskaryan just want to check, are waiting on anything else on my end?

hwchase17 (approved):

Issue #13751, opened 2023-11-22
GenericLoader.from_filesystem "exclude" not working


giancarloerra:
### System Info

Python 3.9.6, Langchain 0.0.334

### Who can help?

@eyurtsev

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

I'm experimenting with some simple code to load a local repository to test CodeLlama, but the "exclude" in GenericLoader.from_filesystem seems not working:

`from langchain.document_loaders.generic import GenericLoader
from langchain.document_loaders.parsers import LanguageParser
from langchain.text_splitter import Language

repo_path = "../../my/laravel/project/"

# Load
loader = GenericLoader.from_filesystem(
    repo_path,
    glob="**/*",
    suffixes=[".php"],
    parser=LanguageParser(
        parser_threshold=2000,
    ),
    exclude=["../../my/laravel/project/vendor/", "../../my/laravel/project/node_modules/", "../../my/laravel/project/storage/", "../../my/laravel/project/public/", "../../my/laravel/project/tests/", "../../my/laravel/project/resources/"]
)

documents = loader.load()
len(documents)
`

Am I missing something obvious? I cannot find any example...with or without the exclude, the length of docs is the same (and if I just print "documents" I see files in the folders I excluded).

### Expected behavior

I would expect that listing subpaths from the main path then these would be excluded.

giancarloerra:
I'm afraid it doesn't seem to work. I got exclusion of first level in the folders using:

"vendor/**/*"

But it doesn't seem to be possible, no matter what I do, to make it exclude ALL the content in that folder (including all its subfolders). Is the exclude working only for files or group of files?

It's quite common to have local repositories with vendor modules, that need to be excluded from any attempt at loading code for analysis. It's also quite common to have several folders to exclude, recursively.

Maybe is not yet implemented?

johnlarkin1:
[...]

PR #13894, opened 2023-11-27
Improve exclude path filesystem logic

johnlarkin1:
- **Description:** Improving the use of the exclude arguments for `FileSystemBlobLoader`
  - **Issue:** #13751
  - **Dependencies:** n/a
  - **Tag maintainer:** 
  - **Twitter handle:** this PR won't get announced, but @jlarks32

See the issue above for more detail, but I don't think that the `exclude` parameter is being handled really like what the API is indicating / what we were thinking (largely due to annoyances with pathlib's `Path.match`). Let me know if you all want me to add some more unit tests and build this out a little bit more.

### DIFF
--- a/libs/langchain/langchain/document_loaders/blob_loaders/file_system.py
+++ b/libs/langchain/langchain/document_loaders/blob_loaders/file_system.py
@@ -118,11 +118,19 @@ class FileSystemBlobLoader(BlobLoader):
 
     def _yield_paths(self) -> Iterable[Path]:
         """Yield paths that match the requested pattern."""
+
+        def match_exclude(path: Path) -> bool:
+            """Check if the path matches any of the exclude patterns."""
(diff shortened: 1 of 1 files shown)


johnlarkin1:
Hey team, I wasn't able to actually run `poetry install --with test` so if there are issues with the formatting / builds let me know how to resolve and I definitely can. 

Here was the issue i filed: https://github.com/langchain-ai/langchain/issues/13912

efriis (commented):
pathlib's match sounds like a mess! However, I'm not sure the included logic is the best way of handling it, because it's a bit too custom.

I haven't looked too deeply into this, but given the python bug is tagged only up to python3.9, does the issue resolve on newer python versions?

If possible, I'd love to resolve this without any custom path matching logic in the langchain library to overcome issues in python.

ccurme:

PR #13901, opened 2023-11-27
AnthropicFunctions  function_call compatibility

JohnnyFoulds:
- **Description:** Updates to `AnthropicFunctions` to be compatible with the OpenAI `function_call` functionality.
  - **Issue:** The functionality to indicate `auto`, `none` and a forced function_call was not completely implemented in the existing code.
  - **Dependencies:** None
  - **Tag maintainer:** @baskaryan , and any of the other maintainers if needed.
  - **Twitter handle:** None

I have specifically tested this functionality via AWS Bedrock with the Claude-2 and Claude-Instant models.


hwchase17 (approved):
thanks, i think its unclear what best to do when no tool_input is returned, but this seems reasonable i guess. may update later when real anthropic function calling comes out

PR #13914, opened 2023-11-27
docs[patch]: RAG Cookbook example fix

waseemhnyc:
### Description:
Hey 👋🏽  this is a small docs example fix. Hoping it helps future developers  who are working with Langchain.

### Problem:
Take a look at the original example code. You were not able to get the `dialogue_turn[0]` while it was a tuple.

Original code:
```python
def _format_chat_history(chat_history: List[Tuple]) -> str:
    buffer = ""
    for dialogue_turn in chat_history:
        human = "Human: " + dialogue_turn[0]
        ai = "Assistant: " + dialogue_turn[1]
        buffer += "\n" + "\n".join([human, ai])
    return buffer
```
In the original code you were getting this error:
```bash
    human = "Human: " + dialogue_turn[0].content
                        ~~~~~~~~~~~~~^^^
TypeError: 'HumanMessage' object is not subscriptable
```
### Solution:
The fix is to just for loop over the chat history and look to see if its a human or ai message and add it to the buffer.


waseemhnyc:
Thanks @baskaryan!

efriis (commented):
Hey @waseemhnyc ! Won't this break the retrieval `invoke` right below this definition? This particular cookbook example is storing chat history as a list of tuples.

hwchase17:
yeah i think this was assuming the input was `List[Tuple[str]]`

Rather than change that assumption, maybe we should just update typing to be more clear

efriis (approved):

PR #13915, opened 2023-11-27
Update openai.py: Fix the bug where openai_proxy is not effective.

leleboys:
Fix the bug where ChatOpenAI().openai_proxy = HTTP_PROXY is not effective.

Reproduce&Test:

from langchain.chat_models import ChatOpenAI
llm = ChatOpenAI()
llm.openai_proxy = "http://127.0.0.1:8001"
llm.predict("Hello")


efriis:
Closing because this won't work (httpx isn't imported where it's needed)! Also, could you try using `ChatOpenAI(openai_proxy="...")` instead, or using an environment variable like it's used in the docs?

PR #13930, opened 2023-11-27
Update prompt injection model

asofter:
- **Description:** Existing model used for Prompt Injection is quite outdated but we fine-tuned and open-source a new model based on the same model deberta-v3-base from Microsoft - [laiyer/deberta-v3-base-prompt-injection](https://huggingface.co/laiyer/deberta-v3-base-prompt-injection). It supports more up-to-date injections and less prone to false-positives.
  - **Dependencies:** No
  - **Tag maintainer:** -
  - **Twitter handle:** @alex_yaremchuk


efriis (commented):
Instead of changing the default model, can you add a parameter that allows people to configure it, leaving the default as the existing deepset one? Feel free to add a note in the docs that you can configure an alternative from laiyer!

baskaryan:
per @efriis's suggestion updated to change docs and configurability without changing default behavior, let me know if it looks ok to you @asofter!

asofter:
@efriis thank you for the suggestion!

@baskaryan thank you for the implementation. It looks great! Probably in another PR, I will add support of ONNX for faster inference but good to go for now.

libs/experimental/langchain_experimental/prompt_injection_identifier/hugging_face_identifier.py
```diff
@@ -1,36 +1,50 @@
...
+    """Tool that uses HF model to detect prompt injection attacks."""
 
     name: str = "hugging_face_injection_identifier"
     description: str = (
         "A wrapper around HuggingFace Prompt Injection security model. "
         "Useful for when you need to ensure that prompt is free of injection attacks. "
         "Input should be any message from the user."
     )
-    model: Pipeline = Field(default_factory=_model_default_factory)
+    model: Any = Field(default_factory=_model_default_factory)
```
efriis, line 35: would it be better to separate out the `model_name` and `model` params? Looks like it might be doing double duty currently.

baskaryan: it'd be cleaner code but seems like a potentially confusing UX to me

Issue #12165, opened 2023-10-23
For New Contributors: Use SecretStr for api_keys


eyurtsev:
Hello LangChain community,

We're always happy to see more folks getting involved in contributing to the LangChain codebase.

This is a good first issue if you want to learn more about how to set up
for development in the LangChain codebase.

## Goal

Your contribution will make it safer to print out a LangChain object
without having any secrets included in raw format in the string representation.

## Set up for development

Prior to making any changes in the code:

https://github.com/langchain-ai/langchain/blob/master/.github/CONTRIBUTING.md

Make sure you're able to test, format, lint from `langchain/libs/langchain`

```sh
make test
```

```sh
make format
```

```sh
make lint
```

## Shall you accept

Shall you accept this challenge, please claim one (and only one) of the modules from the list
below as one that you will be working on, and respond to this issue.

Once you've made the required code changes, open a PR and link to this issue.

## Acceptance Criteria

- invoking str or repr on the the object does not show the secret key

Integration test for the code updated to include tests that:

- confirms the object can be initialized with an API key provided via the initializer
- confirms the object can be initialized with an API key provided via an env variable

Confirm that it works:

- either re-run notebook for the given object or else add an appropriate test
that confirms that the actual secret is used appropriately (i.e.,`.get_secret_value()`)

**If your code does not use `get_secret_value()` somewhere, then it probably contains a bug!**

## Modules

- [ ] langchain/chat_models/anyscale.py @aidoskanapyanov 
- [ ] langchain/chat_models/azure_openai.py @onesolpark 
- [x] langchain/chat_models/azureml_endpoint.py @fyasla 
- [ ] langchain/chat_models/baichuan.py
- [ ] langchain/chat_models/everlyai.py @sfreisthler 
- [x] langchain/chat_models/fireworks.py @nepalprabin 
- [ ] langchain/chat_models/google_palm.py @faisalt14 
- [ ] langchain/chat_models/javelin_ai_gateway.py
- [ ] langchain/chat_models/jinachat.py
- [ ] langchain/chat_models/konko.py
- [ ] langchain/chat_models/litellm.py
- [ ] langchain/chat_models/openai.py @AnasKhan0607 
- [ ] langchain/chat_models/tongyi.py
- [ ] langchain/llms/ai21.py
- [x] langchain/llms/aleph_alpha.py @slangenbach 
- [ ] langchain/llms/anthropic.py
- [ ] langchain/llms/anyscale.py @aidoskanapyanov 
[...]

PR #13946, opened 2023-11-28
secretStr for baichuan chat model api key

nhywieza:
Description: This PR removes secrets present in raw format,
Issue: Baichuan api key was exposed when printing out the langchain object(https://github.com/langchain-ai/langchain/issues/12165)
Tag maintainer: @eyurtsev


leo-gan:
CC @eyurtsev @baskaryan BTW we have similar issue with other LLMs (secrets are not hidden by `SecretStr`)
For example,
`    openai_api_key: Optional[str] = Field(default=None, alias="api_key")` - in the `ChatOpenAI` class

eyurtsev (approved):
Looks good!

Issue #11249, opened 2023-09-30
MathpixPDFLoader never finishes


fincha:
### System Info

langchain==0.0.305
MacOS M1 Silicon

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

I have a basic methon for loading the pdf via mathpix

```
def load_pdf(path):
    loader = MathpixPDFLoader(path)
    pages = loader.load()
    return pages;

pages = load_pdf("example_data/DOF-230519AFYXBECVECGY-0028747724.pdf")

```

but if I run this, I gen only this:

Status: None, waiting for processing to complete
Status: None, waiting for processing to complete
Status: None, waiting for processing to complete
Status: None, waiting for processing to complete
Status: None, waiting for processing to complete
Status: None, waiting for processing to complete
Status: None, waiting for processing to complete
...

in backend of mathpix I see the document aldready parsed.

Any idea on this?

### Expected behavior

Expect the document to return.

fincha:
[71681972252-1.pdf](https://github.com/langchain-ai/langchain/files/12774673/71681972252-1.pdf)

I am parsing this pdf. The timeout is not an issue, the detault is 500 and the result is visible in backend of mathpix

```


[...]

PR #13949, opened 2023-11-28
Fix Mathpix PDF loader integration

chadnorvell:
- **Description:** Fixes the Mathpix PDF loader API integration. Specifically, ensures that Mathpix auth headers are provided for every request, and ensures that we recognize all errors that can occur during a request. Also, the option to provide API keys as kwargs never actually worked before, but now that's fixed too.
  - **Issue:** #11249
  - **Dependencies:** None


hwchase17 (approved):
thanks!

PR #13957, opened 2023-11-28
Use langchain core in-tree as a dev dependency

nfcampos:
Using the published version means master is broken for contributors whenever we make changes in one lib that depend on the other.

### DIFF
--- a/.github/workflows/langchain_ci.yml
+++ b/.github/workflows/langchain_ci.yml
@@ -3,19 +3,19 @@ name: libs/langchain CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
     paths:
-      - '.github/actions/poetry_setup/action.yml'
-      - '.github/tools/**'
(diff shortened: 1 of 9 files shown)


.github/workflows/langchain_ci.yml
```diff
@@ -33,77 +33,29 @@ env:
...
   pydantic-compatibility:
-    uses:
-      ./.github/workflows/_pydantic_compatibility.yml
+    uses: ./.github/workflows/_pydantic_compatibility.yml
     with:
       working-directory: libs/langchain
-      langchain-core-location: ../core
     secrets: inherit
 
-  # It's possible that langchain works fine with the latest *published* langchain-core,
```
nfcampos, line 67: Note this test was actually already duplicate, because when the main test workflow is run with `langchain-core-location` arg it always installs the in-tree version of core

efriis (approved):

baskaryan (approved):

PR #13984, opened 2023-11-28
fix test deps

baskaryan:


libs/langchain/pyproject.toml
```diff
@@ -12,7 +12,7 @@ langchain-server = "langchain.server:main"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
-langchain-core = "^0.0.6"
+langchain-core = ">=0.0.6"
```
efriis, line 15: Shouldn't we keep this as ^ because we're requiring the upgrade of core from 6 to 7?

efriis (approved):
One q

efriis (approved):

PR #13986, opened 2023-11-28
Update writer.py

Pannanja:
fixed typos


efriis:
Will leave this as-is actually! Some of these decrease performance (e.g. getting rid of "urls")

You're welcome to update these in your local installation though - that's the idea behind templates!

PR #13988, opened 2023-11-28
Audio file loader implemented with Azure speech service

kzmain:

  - **Description:** Audio file loader implemented with Azure speech service
  - **Issue:** No,
  - **Dependencies:** azure-cognitiveservices-speech,
  - **Tag maintainer:** kzmain (see below),
  - **Twitter handle:** @kzmain (Same id as the github)


libs/langchain/langchain/document_loaders/audio.py
```diff
@@ -0,0 +1,24 @@
...
+
+from pathlib import Path
+from typing import Any, Union
+
+from langchain.document_loaders.blob_loaders import FileSystemBlobLoader
+from langchain.document_loaders.generic import GenericLoader
+from langchain.document_loaders.parsers.audio import AzureSpeechServiceParser
+
+
+class AzureSpeechServiceLoader(GenericLoader):
```
eyurtsev, line 11: Made a PR here to help with the pattern for Generic Loader: https://github.com/langchain-ai/langchain/pull/14004 -- we don't want any file logic to live outside of file system blob loader.

@kzmain met know if you want to complete the refactor or if you want me to do it

libs/langchain/langchain/document_loaders/parsers/audio.py
```diff
@@ -298,3 +299,189 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
                 page_content=res.normalized_text,
                 metadata={"source": blob.source},
             )
+
+
+class AzureSpeechServiceParser(BaseBlobParser):
+    """Loads an Audio with azure.cognitiveservices.speech."""
+
+    def __init__(self, **kwargs: Any) -> None:
```
eyurtsev, line 307: 1) Could you add a link to the relevant pages on azure that explain the API?
2) Could you expand the kwargs and document the variable meanings?
3) Could we add a parameter to control the polling interval (i.e., sleep is hard-coded to 0.5 seconds right now, we should expose this to the user)

---

nit: It's common to use logger rather than print statements in production code, you could consider swapping to using logger.info / logger.error as appropriate. (not a requirement to merge since we have some other parsers that are using print right now)

kzmain: Dear eyurtsev,

I fixed all the issues in the commit [7b8372d](https://github.com/langchain-ai/langchain/pull/13988/commits/7b8372db69bfb905459400218afc71f0948a9d91):

I Added Azure Speech service official documents and code pages.
I expanded and documented the kwargs meanings.
I added a parameter to control the transcribe job's polling interval.

Furthermore, I did all the unit tests in the test_audio.py file and here's the evidence:
Screenshot 2023-12-03 at 03 26 35

 Look forward to seeing feedback from you. Thanks.

libs/langchain/langchain/document_loaders/audio.py
eyurtsev: @kzmain 

1) I updated the service loader to use generic loader (see https://github.com/langchain-ai/langchain/pull/14004)
2) I expanded the kwargs so the parameters are easy to discover and refactored some of the resolving of values

Would you be able to test that the service still works for you? I don't have credentials set up so testing help will be helpful


3) Could you  run the parser with 2 audio files at once using `from_filesystem` to confirm that transcribing more than one file at a time works?

libs/langchain/tests/integration_tests/test_audio.py
```diff
@@ -0,0 +1,49 @@
+from langchain.document_loaders.audio import AzureSpeechServiceLoader
+
+SPEECH_SERVICE_REGION = ""
+SPEECH_SERVICE_KEY = ""
+
+
+def _get_csv_file_path() -> str:
```
eyurtsev, line 7: Could you fix the testing code here?

libs/langchain/tests/integration_tests/test_audio.py
```diff
@@ -0,0 +1,49 @@
...
+    return ""
+
+
+def test_azure_speech_load_key_region_auto_detect_languages() -> None:
+    loader = AzureSpeechServiceLoader(
```
eyurtsev, line 12: `loader = AzureSpeechServiceLoader.from_filesystem(...)`

kzmain: Dear @eyurtsev ,

AzureSpeechServiceLoader doesn't inherit the `GenericLoader` but the `BaseLoader ` just like the CSVLoader and the PDFLoader. In this case, it does not use the from_filesystem() function but load the file with class initializer like: `AzureSpeechServiceLoader(file_path: str)` format.

I uploaded the unit tests success screenshot in previous conversation. I am wondering if I upload my unit test result with console result with console and the test code (without my credential), will help your code review?

kzmain: Dear @eyurtsev ,

First appreciate you soooo much for your updated my parser neat and well-structured, I found it until I got home. I updated my loader after I got home:
1. the class `AzureSpeechServiceParser` `__init__` function kwargs' type
2. changed the loader from a GenericLoader  to BaseLoader, and added the lazy_load function
3. updated the unit tests
Screenshot 2023-12-06 at 20 04 05

kzmain:
Dear @eyurtsev Can you help me to check if my commit is available to merge?

eyurtsev:
Hi @kzmain sorry for failing to follow on this PR. I'm going to close this since this is a year old and still hasn't been merged. I appreciate you contributing to the project and really apologize that it's taken so long to get a resolution.

PR #13991, opened 2023-11-28
Add MongoDB database agent

### LINKED ISSUE #3399: Connection with mongo db
Can you please help me with connecting my LangChain agent to a MongoDB database? I know that it's possible to directly connect to a SQL database using this resource [https://python.langchain.com/en/latest/modules/agents/toolkits/examples/sql_database.html](url) but I'm not sure if the same approach can be used with MongoDB. If it's not possible, could you suggest other ways to connect to MongoDB?

amiaxys:
**Description:** Created an agent designed to interact with and query a `MongoDB` database using PyMongo. The agent is very similar to [SQL Database](https://python.langchain.com/docs/integrations/toolkits/sql_database). Currently, it only supports one database per instance `MongoDatabase` class using URI, and the database name must be appended to the end of the URI. 

See `docs/docs/integrations/toolkits/mongo_database.ipynb` for an example of usage.

**Note:** To run PyMongo queries, it checks the input string using regex that it follows the `db.collection.find ...` format and runs it using `eval()`, which is bad practice and may be dangerous (if anyone finds a better solution, feel free change that). Additionally, the tests pass locally due to having MongoDB installed since the tests use the `tmp/mongodb-27017.sock` Unix socket, but do not pass Github Actions test workflows for the same reason, and since we weren't sure whether it would be good practice to edit the workflow and haven't figured out a good way to mock PyMongo's MongoClient/Database that isn't unnecessarily complex or makes the unit tests superficial, we left it as is. 

**Issue:** Closes [Connection with mongo db #3399](https://github.com/langchain-ai/langchain/issues/3399)
**Dependencies:** pymongo (dependency already existed, but changed from optional to not)


Haoming-jpg:
@eyurtsev Hi, could you please take a look at our PR? Ty!!

docs/docs/integrations/toolkits/mongo_database.ipynb
```diff
@@ -0,0 +1,202 @@
...
+    "This shows how to initialize the agent using the `OPENAI_FUNCTIONS` agent type. Note that this is an alternative to the above."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# agent_executor = create_sql_agent(\n",
```
hwchase17, line 91:
```suggestion
    "# agent_executor = create_mongo_agent(\n",
```

libs/langchain/langchain/tools/sql_database/tool.py
```diff
@@ -2,15 +2,15 @@
 """Tools for interacting with a SQL database."""
 from typing import Any, Dict, Optional
 
-from langchain_core.pydantic_v1 import BaseModel, Extra, Field, root_validator
+from langchain.pydantic_v1 import BaseModel, Extra, Field, root_validator
```
hwchase17, line 5:
```suggestion
from langchain_core.pydantic_v1 import BaseModel, Extra, Field, root_validator
```

libs/langchain/langchain/tools/sql_database/tool.py
```diff
@@ -2,15 +2,15 @@
...
 
-from langchain_core.language_models import BaseLanguageModel
+from langchain.schema.language_model import BaseLanguageModel
```
hwchase17, line 7:
```suggestion
from langchain_core.language_models import BaseLanguageModel
```

libs/langchain/langchain/tools/sql_database/tool.py
```diff
@@ -2,15 +2,15 @@
...
 from langchain.callbacks.manager import (
     AsyncCallbackManagerForToolRun,
     CallbackManagerForToolRun,
 )
 from langchain.chains.llm import LLMChain
-from langchain_core.prompts import PromptTemplate
+from langchain.prompts import PromptTemplate
```
hwchase17, line 13:
```suggestion
from langchain_core.prompts import PromptTemplate
```

docs/docs/integrations/toolkits/sql_database.ipynb
```diff
@@ -43,13 +43,73 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 9,
    "id": "65ec5bb3",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
hwchase17, line 49: should probably revert

Haoming-jpg:
Thank you sooooooo much!!! @hwchase17

PR #13996, opened 2023-11-29
Adding Initial Support for NVIDIA AI Playground LLMs

VKudlay:
- **Description:** Added NVIDIA AI Playground Initial support for a selection of models (Llama models, Mistral, etc.)
  - **Dependencies:** These models do depend on the AI Playground services in NVIDIA NGC. API keys with a significant amount of trial compute are available (10K queries as of the time of writing).

- Some additional tests can be ran in [this colab notebook](https://colab.research.google.com/drive/1I2xcp8Mq1x36BizAlzRCYDsEbCiQqRBL#revisionId=0B3KSICnsqLpGNHJVWTlVN1E4YkZyR3JLNG1yOG8xUDVkU0VvPQ)
- [A quick link to get a key](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/ai-foundation/models/codellama-13b/api) when you have an NGC account. Generate Key button at the top right of the code window.

Thanks so much for the help, and please let me know if you need anything :D


docs/docs/integrations/llms/nv_aiplay.ipynb
hinthornw, line 1: I like that this walks through some of the differences in classes. To make it easier for a reader, I think the section taken from async_llm doesn't add much: it's easiest to use invoke/ainvoke for single calls or batch/abatch for batch calls. No need to add other asyncio.gather commentary

hinthornw: For the example with the base64 encoded  : is this the recommended API for specifying multimodal content (using html tags) ?

docs/docs/integrations/chat/nv_aiplay.ipynb
hinthornw, line 1: "Interfae" is misspelled

For both of these documents, I'd probably invert the order of presentation and start with the "Recommended solution" rather than the underlying abstraction, unless you believe most people will use the NVCRModel class directly. It's doesn't implement any common APIs, so it will be harder to compose with the rest of the framework. We should lead with the abstraction that _is_ composable with the framework and then take a step back to show how you can have more configurability/control for common intentions by using this, if it is a good abstraction

cc @baskaryan

hinthornw (commented):
Thanks for the PR! My main comments/questions are around the several abstractions. I think you do a pretty good job of defining them, but if I were a new user, I'd get a bit lost on which abstraction I am supposed to use. Ideally there is 1 interface for chat models that I could use as a user, and then the ways to configure it are well documented. As it stands, it looks from the notebooks like I should use the NVCRModel class, which doesn't share a common interface with the other LLMs in langchain

VKudlay:
@hinthornw Hey William! Thanks so much for the review! I'll address the concerns in order and can make the updates! Regarding the recommended interface, NVCRModel is a backbone model, so I'm ok with de-emphasizing it (or rather, maybe moving that discussion to the end). It's probably still a good thing to note somewhere since that interface is way closer to the raw interface advertised (as the python option) in [AI Playground UI](https://catalog.ngc.nvidia.com/ai-foundation-models). 

As for the shared interfaces, are you recommending the removal of LlamaChat/LlamaLLM/MistralLLM/etc and just using NVAIPlayLLM/NVAIPlayChat? I'm mostly down for that (would probably be good to keep a divide between SteerLM-style and non-SteerLM-style defaults and the future NeVA incorporation). Alternatively, root_validator could get loaded with name-based resolutions to pick the appropriate pathways and whine when a model shouldn't be receiving something? 

Thanks again :D

libs/langchain/langchain/embeddings/nv_aiplay.py
```diff
@@ -0,0 +1,75 @@
...
+        model_name = values.get("model")
+        model_name = model_name if model_name else values["model_name"]
+        values["model_name"] = model_name
+        values["model"] = model_name
+        return values
+
+    def _embed(self, text: str, model_type: Literal["passage", "query"]) -> List[float]:
+        """Embed a single text entry to either passage or query type"""
+        if len(text) > self.max_length:
+            text = text[: self.max_length]
```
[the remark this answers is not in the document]
VKudlay: Yeah, I agree that it looks weird, but yes. From [the model documentation](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/ai-foundation/models/nvolve-29k/documentation).

image

String length is the problematic one, whereas token overflow is ok (truncated server-side). I could put in that latter check on client-side; might be a good idea...

libs/langchain/langchain/embeddings/nv_aiplay.py
```diff
@@ -0,0 +1,75 @@
...
+## NOTE: This file should not be ran in isolation as a single-file standalone.
+## Please use llms.nv_aiplay instead.
+
+
+class NVAIPlayEmbeddings(ClientModel, Embeddings):
+    """NVIDIA's AI Playground NVOLVE Question-Answer Asymmetric Model."""
+
+    client: NVCRModel = Field(NVCRModel)
+    model_name: str = Field("nvolveqa")
+    model: Optional[str] = Field(None)
```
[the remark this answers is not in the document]
VKudlay: Awesome! Yeah, I kinda hated that, but was following other implementations' examples. Will fix.

docs/docs/integrations/llms/nv_aiplay.ipynb
VKudlay: @hinthornw Yup! There is another interface I'm looking into since this API is somewhat limiting in its character count (which is why there's a minor halt on NeVA-in-LangChain integration, [with my current best-effort listed here](https://github.com/langchain-ai/langchain/commit/f3048381bb16fd2c4f20a19d50e5b6a81549abf7)). Currently, the input is re-searched server-side and the image is processed and formatted. 

image

This is actually an example of using the underlying NVCRModel client to very directly mimic the requests structure of an AI Playground model [from the current core documentation](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/ai-foundation/models/neva-22b/api).

libs/langchain/langchain/embeddings/nv_aiplay.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def aembed_batch_query(self, texts: List[str]) -> List[List[float]]:
+        """Embed query text with Asynchronous Batching."""
+        queries = [self.aembed_query(text) for text in texts]
+        return await asyncio.gather(*queries)
+
+    async def aembed_batch_documents(self, texts: List[str]) -> List[List[float]]:
+        """Embed search docs with Asynchronous Batching."""
+        passages = [self.aembed_documents([text]) for text in texts]
+        outs = await asyncio.gather(*passages)
```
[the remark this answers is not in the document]
VKudlay: Thanks for the CC. Would love some guidance on this on how to make this a proper and recommendable method! (I'm actually a content developer for DLI, so I would like to know what is recommended and where the pitfalls would come up).

libs/langchain/langchain/embeddings/nv_aiplay.py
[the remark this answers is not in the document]
hinthornw: Got it thanks for explaining

libs/langchain/langchain/embeddings/nv_aiplay.py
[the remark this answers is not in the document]
hinthornw: Thanks! That was mostly because the initial implementations of some integrations went a different way and we didn't want to make a breaking change by removing entirely

VKudlay:
Hey @hinthornw! Resolves as much of the issues I could! Wanted to hot-potato it before the weekend so I didn't have to worry about it for a few days, so no weekend rush or anything like that. 

Of note:
-  Took some creative liberties to try to compromise with the additional class options. Renamed them to much more general "GeneralChat/LLM", "CodeChat/LLM", "InstructChat/LLM", "ContextChat/LLM", "SteerChat/LLM", and "ImageChat/LLM". That way, users can safely drop these into their code and we can just swap out the defaults behind the scenes as better options come out. I have checked and there is currently not a way to derive intake schemas from our API (i.e. does this model need labels, or context-only, or a sub-512-length max_tokens argument), so I think making these intuitive grouping classes is a good idea for now.
- I've confirmed some requirements for NeVA/Fuyu, and I think it's reasonable for us to incorporate; just need to be able to pull an image from user's local, resize to 336x336 (and yes, this kinda sucks for Fuyu at the moment; they're working on it), and toss it into a tag as you noted earlier. Do you have any recommendation for packages that would cause minimal problems for LC? (PIL ok?)
- Would love suggestions on how to let aembed_batch_<queries/documents> exist purely for pedagogical reasons. What do you think I could do to make it good enough? 

Thanks again, and have an awesome weekend! :D

VKudlay:
@hinthornw @baskaryan Just to confirm, everything is good and the embedding model has been updated to add max concurrency restrictions. NeVA does not have deep incorporation yet (it's just a simple default-model class), but I've blocked off the ImageLLM/ImageChat abstraction for it and intend to update it with deeper incorporation (automatic image loading) later. If it will speed up the PR, we are happy to remove those classes for now. Regardless, things look good :D

libs/langchain/langchain/llms/nv_aiplay.py
```diff
@@ -0,0 +1,764 @@
+## NOTE: This class is intentionally implemented to subclass either ChatModel or LLM for
+##  demonstrative purposes and to make it function as a simple standalone file.
```
hinthornw, line 2: Not sure what "for demonstrative purposes" means in this context actually

libs/langchain/langchain/llms/nv_aiplay.py
```diff
@@ -0,0 +1,764 @@
...
+            "last_msg",
+            "available_functions",
+        ]
+    )
+
+    @root_validator()
+    def validate_model(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        """Validate and update model arguments, including API key and formatting"""
+        values["nvapi_key"] = get_from_dict_or_env(
+            ClientModel.desecretize(values),
```
hinthornw, line 162: Is `ClientModel.desecretize` needed?

libs/langchain/langchain/llms/nv_aiplay.py
```diff
@@ -0,0 +1,764 @@
...
+            return {"role": role, "content": content}
+        ## Support for tuple inputs
+        if type(msg) in (list, tuple):
+            return {"role": msg[0], "content": msg[1]}
+        ## Support for manually-specified default inputs to AI Playground
+        if isinstance(msg, dict) and msg.get("content"):
+            msg["role"] = msg.get("role", "user")
+            return msg
+        ## Support for LangChain Messages
+        if hasattr(msg, "content"):
```
hinthornw, line 684: Coul do an isinstance check instead

hinthornw (commented):
I think the biggest thing is it seems like it will be hard to maintain if we try to put all the parsing logic into a single NVAIBaseModel

libs/langchain/langchain/llms/nv_aiplay.py
VKudlay: Oh haha, that's actually a legacy comment. This project started out as a connector for the Deep Learning Institute to use in courses, and then the abstraction fleshed out afterwards and I stuck to the format. Can remove if you want, but that was just the initial motivator. Now the standalone feature has been removed per request and we are unlikely to make actual "make a connector" instructional material at this point.

libs/langchain/langchain/llms/nv_aiplay.py
VKudlay: Yes! That was just now added in because I realized it was necessary when testing that edge case. The explicit error I was getting implicated SecretStr...

libs/langchain/langchain/llms/nv_aiplay.py
```diff
@@ -0,0 +1,764 @@
...
+        inputs = self.custom_preprocess(messages)
+        kwargs["labels"] = kwargs.get("labels", self.labels)
+        kwargs["stop"] = stop if stop else getattr(self.client, "stop")
+        async for response in self.get_astream(inputs, **kwargs):
+            chunk = self._get_filled_chunk(self.custom_postprocess(response))
+            yield chunk
+            if run_manager:
+                await run_manager.on_llm_new_token(chunk.text, chunk=chunk)
+
+    def custom_preprocess(self, msgs: Union[str, Sequence]) -> List[Dict[str, str]]:
```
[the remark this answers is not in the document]
VKudlay: @hinthornw Thanks! Yeah, the abstraction here settled in after a pretty annoying "follow-the-API" experience. At this point, the API for the existing models has luckily stabilized, but this is just a "centralize the components for maximum flexibility with minimal modification" effort. That way, a new model can come in with its own hard expectations (or make very weird new assumptions) and a single update was sufficient to account for it in every derivative instantiation.

On reflection, I think it would be fine to get rid of the string-to-message routine in favor of LLM positional arguments on the ContextBase/etc subclass. A similar solution could work with "system" positional arguments... but before I do that, I'd like to confirm that this would actually solve the issue you mentioned, or if there is some other considerations that need to be taken? As of right now, it does not look like the AI Playground team is planning to expose raw completion endpoints, but I would still prefer the flexibility for users to help them experiment with the playground in the context they're most comfortable with/is closest to their intended final stack. (Of note, there is a non-zero chance we might want to support vanilla completion or even more complex endpoint options in the future though, either with AI Playground or some more general efforts later on). 

What approach do you think would be the best to help support this kind of use case for now while the underlying API is still technically chat-only?

hinthornw:
Hi @VKudlay I'm going to branch this to another PR to add to its own package, now that we've completed the community split

VKudlay:
@hinthornw Sounds good! Just letting you know, I have some bandwidth to over the next few days (been swamped for the last week), so let me know how I can help out :D

hinthornw:
Main thing is all new integrations need to be in their own package so need to move this out. I'll push a draft. I still think the number of classes is really daunting and wouldn't be used + are hard to maintain

libs/langchain/langchain/llms/nv_aiplay.py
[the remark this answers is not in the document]
hinthornw: Why need positional arguments here? 

If the underlying API is chat only, then we should just support a chat model. If there is a completion or other endpoint that is exposed at some point, we can support that through a separate class. It is much harder if we try to make a fake integration now that may or may not be compatible with the future API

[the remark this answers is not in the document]
VKudlay: Honestly, fair enough. I'll take a look and see what I can do with the integrations.

hinthornw:
@VKudlay I tried to adapt it and simplify it more here: https://github.com/langchain-ai/langchain/pull/14648

Still a bit complicated I think but it gets  the job done

PR #12200, opened 2023-10-24
fixed error message and a check for processor name

lkuligin:
Replace this entire comment with:
  - **Description:** a small fix on error description / a check for processor name
  - **Issue:** the issue #11407

PR #13999, opened 2023-11-29
New model parameters and dynamic batching for VertexAIEmbeddings

vladkol:
- **Description:** VertexAIEmbeddings performance improvements
  - **Twitter handle:** @vladkol

## Improvements

- Dynamic batch size, starting from 250, lowering down to 5. Batch size varies across regions.
Some regions support larger batches, and it significantly improves performance.
When running large batches of texts in `us-central1`, performance gain can be up to 3.5x.
The dynamic batching also makes sure every batch is below 20K token limit.
- New model parameter `embeddings_type` that translates to `task_type` parameter of the API. Newer model versions support [different embeddings task types](https://cloud.google.com/vertex-ai/docs/generative-ai/embeddings/get-text-embeddings#api_changes_to_models_released_on_or_after_august_2023).


efriis:
@vladkol could you merge in the latest master and resolve conflicts?

efriis:
cc @holtskinner and @lkuligin to comment on approach!

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,16 +1,39 @@
...
+_MAX_TOKENS_PER_BATCH = 20000
+_MAX_BATCH_SIZE = 250
+_MIN_BATCH_SIZE = 5
+
 
 class VertexAIEmbeddings(_VertexAICommon, Embeddings):
     """Google Cloud VertexAI embedding models."""
 
-    model_name: str = "textembedding-gecko"
+    model_name: str = "textembedding-gecko@latest"
```
lkuligin, line 22: Have the `@latest` model being promoted to the default on the SDK level?

I'm we should do this without a long warning on LC side, since many people might be using `VertexAIEmbeddings()` with defaults, and using one model to embed a query and another one to create a vector index might be not a good idea, wdyt?

lkuligin:
LGTM. Dynamic batch size looks really cool.

lkuligin:
one more thought, probably it's a good idea to update the integration test to try it out in at least two locations with large enough batch sizes.

https://github.com/langchain-ai/langchain/blob/master/libs/langchain/tests/integration_tests/embeddings/test_vertexai.py

libs/langchain/langchain/embeddings/vertexai.py
vladkol: Agree. Let me fix that.

vladkol: Although, at some point the default model will be upgraded anyways, so anyone who didn't specify the version (e.g. `textembedding-gecko@001` will get a new version by default.

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,271 @@
...
+
+    @staticmethod
+    def _split_by_punctuation(text: str) -> List[str]:
+        """Splits a string by punctuation and whitespace characters."""
+        split_by = string.punctuation + "\t\n "
+        pattern = f"([{split_by}])"
+        # Using re.split to split the text based on the pattern
+        return [segment for segment in re.split(pattern, text) if segment]
+
+    def _prepare_batches(
```
baskaryan, line 97: would be nice to add a unit test for this and split_by_punctuation

libs/langchain/langchain/embeddings/vertexai.py
vladkol: done

libs/langchain/langchain/embeddings/vertexai.py
vladkol: done

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,271 @@
...
+        )
+
+        @retry_decorator
+        def _completion_with_retry(texts_to_process: List[str]) -> Any:
+            if self.embeddings_type:
+                from vertexai.language_models import TextEmbeddingInput
+
+                requests = [
+                    TextEmbeddingInput(text=t, task_type=self.embeddings_type)
+                    for t in texts
```
lkuligin, line 161: it looks like a typo to me, should it be `texts_to_process` here? Or am I missing anything?

P.S. I'm surprised it has not been catched by the tests, maybe it's a good idea to adjust tests too.

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,271 @@
...
+        texts_len = len(texts)
+        batch_token_len = 0
+        batches: List[List[str]] = []
+        current_batch: List[str] = []
+        if texts_len == 0:
+            return []
+        while text_index < texts_len:
+            current_text = texts[text_index]
+            current_text_token_cnt = (
+                len(VertexAIEmbeddings._split_by_punctuation(current_text)) * 2
```
lkuligin, line 115: I'm sorry but I'm a little bit confused.
Do we count tokens or characters here?

If we count tokens, why not re-use `count_tokens` method? If we count characters (and limit is in characters), maybe it's a good idea to adjust the namings.

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,271 @@
...
+            if (
+                batch_token_len + current_text_token_cnt > _MAX_TOKENS_PER_BATCH
+                or len(current_batch) == batch_size
+            ):
+                end_of_batch = True
+            else:
+                if text_index == texts_len - 1:
+                    end_of_batch = True
+                batch_token_len += current_text_token_cnt
+                current_batch.append(current_text)
```
lkuligin, line 127: it looks to me, that the last element of the batch (if batch is >1) will be always ignored. it might be a good idea to add a unit test for this.

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,271 @@
...
+                len(VertexAIEmbeddings._split_by_punctuation(current_text)) * 2
+            )
+            end_of_batch = False
+            if (
+                batch_token_len + current_text_token_cnt > _MAX_TOKENS_PER_BATCH
+                or len(current_batch) == batch_size
+            ):
+                end_of_batch = True
+            else:
+                if text_index == texts_len - 1:
```
lkuligin, line 124: nits: maybe it would be easier to read if you move this condition to the first `if` condition (4 lines above) as another `or`

libs/langchain/langchain/embeddings/vertexai.py
vladkol: I'll fixed that. The tests won't catch it as both parameters are effectively the same, always.

libs/langchain/langchain/embeddings/vertexai.py
vladkol: We count tokens to ensure that batch won't exceed 20K token limit.
Using `count-tokens` will be super expensive - it is an extra request for **every** text. This will kill the performance gain we get from batching. I think it will be even slower.
I was thinking about using one of LangChain's own tokenizers, but every single one carries external dependencies. It feels suboptimal, especially if such tokenizer is not re-used by the application.
This is why I took a conservative approach 2 tokens per word, including punctuation and whitespace characters.

libs/langchain/langchain/embeddings/vertexai.py
vladkol: There is a test for that (`test_large_batches`), and by writing it I caught exactly the bug you mentioned. The last element is not ignored now.
With the last element, `end_of_batch` becomes `True`, but right after that we also do `current_batch.append(current_text)`.

libs/langchain/langchain/embeddings/vertexai.py
vladkol: done.

libs/langchain/langchain/embeddings/vertexai.py
lkuligin: Thanks, it makes sense now. Maybe it's worth adding a comment to the codebase to explain the approach. Otherwise somebody might try to improve the implementation with the best intentions.

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,271 @@
...
+        if batch_size == 0:
+            batch_size = self.instance["batch_size"]
+        text_index = 0
+        texts_len = len(texts)
+        batch_token_len = 0
+        batches: List[List[str]] = []
+        current_batch: List[str] = []
+        if texts_len == 0:
+            return []
+        while text_index < texts_len:
```
lkuligin, line 112: it looks to me, that if the length of one element is above _MAX_TOKENS_PER_BATCH, the script will run forever.

and since it's conservative estimation, we always should have at least one element per batch (even if this element exceed the conservative limit) and let the model fail during inference with an appropriate error.

vladkol: Great catch. Thank you! Done.

libs/langchain/langchain/embeddings/vertexai.py
vladkol: Done.

libs/langchain/tests/integration_tests/embeddings/test_vertexai.py
```diff
@@ -24,6 +24,70 @@ def test_embedding_query() -> None:
     assert len(output) == 768
 
 
+def test_large_batches() -> None:
```
lkuligin, line 27: should it be a unit test?
there's an example how to add a unit test that requires a custom dependency: https://github.com/langchain-ai/langchain/pull/12200/files

integrations tests are only compiled, not run on every push, so I'd rather move unit tests to the corresponding folder
@baskaryan please, correct me if I'm wrong

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,288 @@
...
+            # necessarily be reused by the application using this class.
+            current_text_token_cnt = (
+                len(VertexAIEmbeddings._split_by_punctuation(current_text)) * 2
+            )
+            end_of_batch = False
+            if current_text_token_cnt > _MAX_TOKENS_PER_BATCH:
+                # Current text is too big even for a single batch.
+                # Such request will fail, but we still make a batch
+                # so that the app can get the error from the API.
+                current_batch.append(current_text)
```
lkuligin, line 130: I'm sorry for being too nerdy, but if the `current_batch` is not empty, should we "close it" and **also** add another batch with a single element?

vladkol: True. Done,

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,294 @@
...
+            location=location,
+            credentials=credentials,
+            request_parallelism=request_parallelism,
+            max_retries=max_retries,
+            **kwargs,
+        )
+        self.model_name = model_name
+        self.embeddings_type = embeddings_type
+
+        self.instance["batch_size"] = _MAX_BATCH_SIZE
```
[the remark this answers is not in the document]
vladkol: These values are not regulated by quotas. They mostly depend on regional capacity. 
But I agree, let's get them from kwargs when specified.

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,294 @@
...
+        self.instance["min_good_batch_size"] = _MIN_BATCH_SIZE
+        self.instance["lock"] = threading.Lock()
+        self.instance["batch_size_validated"] = False
+        self.instance["task_executor"] = ThreadPoolExecutor(
+            max_workers=request_parallelism
+        )
+
+    @staticmethod
+    def _split_by_punctuation(text: str) -> List[str]:
```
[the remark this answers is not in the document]
lkuligin: ok, that's my bad, please ignore it.

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,296 @@
...
+            # Number of tokens per a text is conservatively estimated
+            # as 2 times number of words, punctuation and whitespace characters.
+            # Using `count_tokens` API will make batching too expensive.
+            # Utilizing a tokenizer, would add a dependency that would not
+            # necessarily be reused by the application using this class.
+            current_text_token_cnt = (
+                len(VertexAIEmbeddings._split_by_punctuation(current_text)) * 2
+            )
+            end_of_batch = False
+            if current_text_token_cnt > _MAX_TOKENS_PER_BATCH:
```
lkuligin, line 125: this (and below) should probably also come from instance attributes

lkuligin (approved):

lkuligin:
LGTM, thanks!
@baskaryan please, take a look

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -1,46 +1,298 @@
...
 
+    def __init__(
+        self,
+        project: Optional[str] = None,
+        location: str = "us-central1",
+        request_parallelism: int = 5,
+        max_retries: int = 6,
+        model_name: str = "textembedding-gecko",
+        credentials: Optional[Any] = None,
+        embeddings_type: Optional[
```
baskaryan, line 61: seems unintuitive that embed_documents would use anything other than RETRIEVAL_DOCUMENT and embed_query would use anything other than RETRIEVAL_QUERY. should those use hard coded embeddings_type, and then if we want could expose a generic `embed` method that uses configured embeddings_type?

vladkol: @baskaryan I made a slight change to accommodate that.
`VertexAIEmbeddings` doesn't carry that parameter anymore. Instead, it only detects whether the embedding task type can be used - based on the endpoint name.

`embed_query` and `embed_documents` use the most suitable embeddings type for their respective task when calling a new `embed` method. 
`embed` accepts an embeddings task type name as a parameter. In all cases, the parameter is only passed to the model when it's allowed, or silently ignored otherwise.

Please take a look.

vladkol: @baskaryan Please let me know if there is anything that should be improved here. Thank you!

vladkol: @baskaryan I moved files to the new `libs/community`. Please take a look.

PR #14001, opened 2023-11-29
docs `OpenAI` platform page update

leo-gan:
Missed the OpenAI adapter reference in the OpenAI platform page
- Added this reference


hwchase17 (approved):
thanks!

PR #14004, opened 2023-11-29
Improve file system blob loader and generic loader

eyurtsev:
* Add support for passing a specific file to the file system blob loader
* Allow specifying a class parameter for the parser for the generic loader

```python

class AudioLoader(GenericLoader):
  @staticmethod
  def get_parser(**kwargs):
     return MyAudioParser(**kwargs):
```

The intent of the GenericLoader is to provide on-ramps from different sources (e.g., web, s3, file system).

An alternative is to use pipelining syntax or creating a Pipeline

```
FileSystemBlobLoader(...) | MyAudioParser
```


eyurtsev:
Actually this won't work -- need to make it possible to lazily instantiate -- one sec

libs/langchain/langchain/document_loaders/concurrent.py
```diff
@@ -75,7 +74,11 @@ def from_filesystem(
             show_progress=show_progress,
         )
         if isinstance(parser, str):
-            blob_parser = get_parser(parser)
+            if parser == "default" and cls.get_parser != GenericLoader.get_parser:
```
efriis, line 77: Would it be better to try/except NotImplementedError instead of checking method != ?

efriis (commented):
1 question, otherwise lgtm

libs/langchain/langchain/document_loaders/concurrent.py
eyurtsev: Sure that works as well I'll change

baskaryan (approved):

Issue #13662, opened 2023-11-21
Issue: How to validate Tool input arguments without raising ValidationError


deepakagrawal:
### Issue you'd like to raise.

I am creating a tool that takes a multiple input arguments (say input1, input2). I would like to validate their types and also make sure that the tool only receives input1 and input2. How do I validate this without breaking the llm chain. I would instead like to return a warning to the llm agent. something like, "The input passed were incorrect, please try again"

deepakagrawal:
@dosu-bot, can you show an example of how is this "warning" used by the llm agent?

deepakagrawal:
@dosu-bot this breaks the langchain flow due to raise ValidationError.

deepakagrawal:
@dosu-bot you are going in loop. This was your initial solution. But that did not work.

deepakagrawal:
@dosu-bot where would l call this function validate_inputs?

ahmadaii:
@deepakagrawal Were you be able to handle this? I'm facing similar validation problem :)

hmasdev:
@deepakagrawal @ahmadaii 
I also faced a similar validation problem. However, I have solved it.
So, I share a simple instant solution with you.

```bash
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import langchain
>>> langchain.__version__
'0.0.330'
>>> from functools import wraps
>>> from langchain.tools import StructuredTool
>>> from langchain import pydantic_v1 as pydantic
>>> # HERE IS A SAMPLE TOOL
>>> def joint(a: str, b: str) -> str:
...     '''joint two strings'''
...     return a + b
...
>>> joint_tool = StructuredTool.from_function(joint, name=joint.__name__, description=joint.__doc__, handle_tool_errors=True)  
>>> # PROBLEM REPRODUCTION
>>> joint_tool.run(dict(a='a', b='b'))
'ab'
>>> joint_tool.run(dict(a='a')) 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\dev_libs\langchain\libs\core\langchain_core\tools.py", line 311, in run
    parsed_input = self._parse_input(tool_input)
  File "D:\dev_libs\langchain\libs\core\langchain_core\tools.py", line 246, in _parse_input
    result = input_args.parse_obj(tool_input)
[...]

Issue #12721, opened 2023-11-01
Issue: How to implement validation on number of arguments in tools?


deepakagrawal:
### Issue you'd like to raise.

I am using the STRUCTURED_CHAT_ZERO_SHOT_REACT_DESCRIPTION agent and I am trying to implement tools that take varying number of arguments (e.g. tool1 takes a single argument and tool2 takes 3 arguments). I am using arg_schema to define the arguments and their types. I am wondering how can I return a message to the agent in case the input arguments do not satisfy the arg_schema without erroring out the service. My objective with this is that the agent will try again after modifying the "Thought". How do I do that?

PR #14007, opened 2023-11-29
core[minor]: add validation error handler to `BaseTool`

hmasdev:
- **Description:**  add a ValidationError handler as a field of [`BaseTool`](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/tools.py#L101) and add unit tests for the code change.
- **Issue:** #12721 #13662
- **Dependencies:** None
- **Tag maintainer:** 
- **Twitter handle:** @hmdev3
- **NOTE:**
  - I'm wondering if the update of document is required.


libs/core/langchain_core/tools.py
```diff
@@ -308,7 +314,23 @@ def run(
         **kwargs: Any,
     ) -> Any:
         """Run the tool."""
-        parsed_input = self._parse_input(tool_input)
+        try:
+            parsed_input = self._parse_input(tool_input)
+        except ValidationError as e:
+            if not self.handle_validation_error:
+                raise e
+            elif isinstance(self.handle_validation_error, bool):
```
eyurtsev, line 322: This should only be done on the True branch

eyurtsev:
@hinthornw What do you think?

Looks reasonable to me given that on tool error is already there.

This would handle on tool input exceptions -- which are encountered frequently with agents.

The main concern is that because this is used primarily with agents (i.e., with things that can re-interact with the tool). It feels like it belongs in the agent executor rather than on the tool itself.

hmasdev:
@eyurtsev @baskaryan 
May I resolve the conflicts? Or, is anyone working on it?

eyurtsev (approved):

baskaryan:
thank you @hmasdev!

PR #14010, opened 2023-11-29
feat: mask api_key for konko

chyroc:
for https://github.com/langchain-ai/langchain/issues/12165

### DIFF
--- a/libs/community/langchain_community/chat_models/konko.py
+++ b/libs/community/langchain_community/chat_models/konko.py
@@ -25,8 +25,8 @@ from langchain_core.language_models.chat_models import (
 )
 from langchain_core.messages import AIMessageChunk, BaseMessage
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
-from langchain_core.pydantic_v1 import Field, root_validator
-from langchain_core.utils import get_from_dict_or_env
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/chat_models/konko.py
```diff
@@ -122,23 +122,23 @@ def _default_params(self) -> Dict[str, Any]:
 
     @staticmethod
     def get_available_models(
-        konko_api_key: Optional[str] = None,
-        openai_api_key: Optional[str] = None,
+        konko_api_key: Optional[SecretStr] = None,
```
eyurtsev, line 125: This is a publicly exposed static method -- let's avoid making a breaking change here in case users are relying on using this static method directly. Instead could you keep the api keys here `str` (there's no need to mask them here since they won't get printed). If we think they'll be used in conjunction with the existing keys on set as attribtues on the model, then we can make these types: Union[str, SecretStr, None]  and support both `str` and `SecretStr` as inputs

eyurtsev (changes requested):
Looks good, but we need to fix up a potentially breaking change in get_available_models before we can merge!

eyurtsev:
Ready to be merged when tests pass

eyurtsev (approved):

chyroc:
@eyurtsev can this be merged

PR #14016, opened 2023-11-29
Improve HyDe with custom prompts and ability to supply the run_manager

yvesloy:
- **Description:** The class allows to only select between a few predefined prompts from the paper. That is not ideal, since other use cases might need a custom prompt. The changes made allow for this. To be able to monitor those, I also added functionality to supply a custom run_manager.
  - **Issue:** no issue, but a new feature,
  - **Dependencies:** none,
  - **Tag maintainer:** @hwchase17,
  - **Twitter handle:** @yvesloy


libs/langchain/langchain/chains/hyde/base.py
```diff
@@ -72,14 +74,28 @@ def from_llm(
         cls,
         llm: BaseLanguageModel,
         base_embeddings: Embeddings,
-        prompt_key: str,
+        prompt_key: Optional[str] = None,
+        custom_prompt: Optional[PromptTemplate] = None,
+        run_manager: Optional[CallbackManagerForChainRun] = None,
```
baskaryan: run_manager creation is handled in background and shouldn't be instantiated / passed around by users. you'd want to pass in `Callbacks` but that can already be down via **kwargs

yvesloy: I think it's not possible right now, because of the config: 
    class Config:
        """Configuration for this pydantic object."""

        extra = Extra.forbid
        arbitrary_types_allowed = True

baskaryan: the class inherits from Chain which has an explicit `Chain.callbacks` parameter, so i think it doesn't count as an extra

yvesloy: I didn't see that! Thanks for pointing it out

Issue #11275, opened 2023-10-02
Simple Serialization and Deserialization of Memory classes to and from dictionaries w/ System Messages included 


lameTookan:
### Feature request

I want simple methods to use on the various memory classes to enable easy serialization and de serialization of the current state of the convo, crucially including system messages as json dictionaries. 

If this is currently already included, I cannot for the life of me find it anywhere in the documentation or the source code. 

In particular there doesn't seem to be a way to add system messages directly to the memory. 

Heres some pseudo code outlining what this would look like
```python
serialized_memory = memory_instance.serialize_to_json()
new_memory = MemoryClass()
new_memory.load_from_json(serialized_memory)
```
In the new memory, the entire contents of the history(at least what has not been pruned) would be included. Including any system messages, which with the summary variants would include the convo summaries. 

### Motivation

The absence of easy serialization and deserialization methods makes it difficult to save and load conversations in JSON files or other formats. Working with dictionaries is straightforward, and they can be easily converted into various data storage formats.

### Your contribution

Possibly. I just started messing around with this module today, and I am relatively new to Python. I will be doing my best to learn more about how this library works so I can add this feature.  

With some more research and experimentation this could be something I could do.

jeffheaton:
Agree, this seems like a very useful feature, ideally using the same interface across all memory types.

D3nam:
I would like to contribute to this issue, and get more experience working on this project!

Aditya-k-23:
Hey @baskaryan,

We are working on this issue and need clarification before we provide a complete resolution.
Under the `BaseChatMessageHistory` class, we see methods `add_user_message` and `add_ai_message`, which were added as a part of [[PR #9725](https://github.com/langchain-ai/langchain/pull/9725)](https://github.com/langchain-ai/langchain/pull/9725) last month, we were wondering why a method like `add_system_message` was not added for system messages.

To make system messages a part of the chat message history as suggested by the user above, we were wondering if we should include such a method in our solution. Could you provide your thoughts on this?

Aditya-k-23:
@jeffheaton @lameTookan 
Can you please clarify which specific classes you are referring to when you say a `MemoryClass`?

lameTookan:
@Aditya-k-23 Basically all of the classes described in this section of the documentation 
https://python.langchain.com/docs/modules/memory/

avaove:
@lameTookan Hello again, couple questions:

[...]

PR #14032, opened 2023-11-29
feat: Add a feature to serialize and deserialize memory types to and from JSON

Aditya-k-23:
Description: Add a feature to serialize and deserialize the memory types into JSON format,
Issue: #11275,
Dependencies: No new dependencies,
Tag maintainer: @baskaryan, @eyurtsev, @hwchase17
Co-Authors: @D3nam, @avaove, @malharpandya


Aditya-k-23:
Our changes add the serialization and deserialization to and from the following memory types:
- `ConversationBufferWindowMemory`
- `ConversationBufferMemory`
- `ConversationSummaryBufferMemory`
- `ConversationSummaryMemory`
- `ConversationEntityMemory`
- `ConversationTokenBufferMemory`
For the following types an additional `llm` argument needs to be passed for deserialization, this is because the current BaseLLM class is not `lc_serializable`:
- `ConversationSummaryBufferMemory`
- `ConversationSummaryMemory`
- `ConversationEntityMemory`
- `ConversationTokenBufferMemory`
In the future, our implementation allows for extension to include Knowledge graphs and Vector Stores, if their `lc_serializable` methods are implemented to support both serialization and deserialization.

hwchase17 (commented):
thanks for opening this up. i dont think we want to do this quite yet. a few reasons:

1. the memory abstractions may not be the best/most stable, so we want to rethink
2. we generally make things serializable in both python and js at the same time

Curious to hear - why the desire to have them serializable?

rupertlssmith:
@hwchase17  Curious to hear - why the desire to have them serializable?

Currently building a system where each step in the conversation is produced by invoking a stateless API. So any conversation state must be passed in to the parameters used to invoke the API.

@Aditya-k-23 Why do these memory types require an llm to deserialize?

* ConversationSummaryBufferMemory
* ConversationSummaryMemory

For example, to my thinking the ConversationSummaryBufferMemory contains a list of Human and AI conversation steps, a current buffer size or token count, and maybe a summary of the conversation so far to capture the overflow from the buffer. A serialized representation could hold these 3 things, and the state could be completely recovered from them.

What we want to avoid is having to make extra calls to the llm to re-create the summary from the conversation steps when deserializing, as this is potentially costly/time consuming.

hwchase17:
i am going to close this. this isnt something we want to make serializable yet

PR #14040, opened 2023-11-29
Add Hugging Face chat wrapper

andrewrreed:
## Issue
There previously has been no easy way to make use of models hosted on Hugging Face (via Inference API or Inference Endpoints) in combination with LangChains ChatModel abstraction. 

## Description
This PR introduces a new chat_model integration that creates a wrapper around the `BaseChatModel` class that interfaces between LangChain's and the hosted LLM by leveraging [Hugging Face's Chat Templates](https://huggingface.co/docs/transformers/chat_templating).

## To do
- [x] Add wrapper class around`BaseChatModel` to interface with HF LLM integrations
- [x] Create a notebook  `docs/integrations/chat` that demonstrates its use
- [x] Add unit test

## Tag maintainer
@hwchase17

## Twitter handle
@andrewrreed


andrewrreed:
@hwchase17 When running the unit tests locally, there are 22 tests that are failing, all of which are inside these two files:
- `tests/unit_tests/callbacks/tracers/test_base_tracer.py`
- `tests/unit_tests/callbacks/tracers/test_langchain_v1.py`

However, when I run pytest on each of those files individually with `poetry run pytest --disable-socket --allow-unix-socket <test-file>` they all pass.... Not sure why this is happening?

libs/langchain/langchain/chat_models/huggingface_chat_wrapper.py
```diff
@@ -0,0 +1,165 @@
+from typing import Any, List, Optional, Union
+
+from huggingface_hub import list_inference_endpoints
+from transformers import AutoTokenizer
```
baskaryan, line 4: since these are optional dependencies they should only be imported inside functions as needed (instead of being imported at top level)

baskaryan (commented):
what's the unit test error/failure you're seeing?

libs/langchain/langchain/chat_models/huggingface_chat_wrapper.py
andrewrreed: Thanks, it looks like this was causing the test failures!

andrewrreed:
@baskaryan The unit tests were failing due to that global import of an optional dependency. Fixed that so everything is passing locally now. Let me know if there's anything else I need to add here, thanks!

libs/langchain/langchain/chat_models/huggingface.py
```diff
@@ -44,14 +41,20 @@ class ChatHuggingFace(BaseChatModel):
...
-    model_id: str = None # type: ignore
+    tokenizer: None
+    model_id: str = None  # type: ignore
 
     def __init__(self, **kwargs: Any):
         super().__init__(**kwargs)
 
-        if not isinstance(self.llm, (HuggingFaceEndpoint, HuggingFaceTextGenInference, HuggingFaceHub)):
-            raise ValueError(f"LLM type not supported for ChatHuggingFace: {type(self.llm)}")
+        from transformers import AutoTokenizer
```
andrewrreed, line 50: @A-Roucher This needs to be a local import per [this comment](https://github.com/langchain-ai/langchain/pull/14040#discussion_r1411042457) by @baskaryan 

With this change, it seems we're back to failing a few tests again :(

andrewrreed:
Thanks for your help on this @A-Roucher!

@baskaryan This PR is now passing all tests and linting checks. Please let us know if anything else is needed to get this merged!

A-Roucher:
No problem @andrewrreed, looking forward to start using this!

baskaryan:
landed in #14736

PR #14044, opened 2023-11-29
Update Prompt Format Error

hinthornw:
The number of times I try to format a string (especially in lcel) is embarrassingly high. Think this may be more actionable than the default error message.  Now I get nice helpful errors

```
KeyError: "Input to ChatPromptTemplate is missing variable 'input'.  Expected: ['input'] Received: ['dialogue']"
```

### DIFF
--- a/libs/core/langchain_core/prompts/base.py
+++ b/libs/core/langchain_core/prompts/base.py
@@ -81,13 +81,13 @@ class BasePromptTemplate(RunnableSerializable[Dict, PromptValue], ABC):
                 f" Expected: {self.input_variables}"
                 f" Received: {list(inner_input.keys())}"
             ) from e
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/prompts/base.py
```diff
@@ -70,10 +70,24 @@ def get_input_schema(
     def invoke(
         self, input: Dict, config: Optional[RunnableConfig] = None
     ) -> PromptValue:
+        def format_prompt(inner_input: Dict) -> PromptValue:
```
baskaryan, line 73: nit: does it need to be nested function?

hinthornw: Could do outside

PR #14046, opened 2023-11-29
[core/minor] Runnables: Implement a context api

nfcampos:

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -7,7 +7,7 @@ from abc import ABC, abstractmethod
 from concurrent.futures import FIRST_COMPLETED, wait
 from copy import deepcopy
 from functools import partial, wraps
-from itertools import tee
+from itertools import groupby, tee
 from operator import itemgetter
 from typing import (
     TYPE_CHECKING,
@@ -22,6 +22,7 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
+    Set,
     Tuple,
     Type,
     TypeVar,
@@ -1401,9 +1402,41 @@ class RunnableSequence(RunnableSerializable[Input, Output]):
 
     @property
     def config_specs(self) -> List[ConfigurableFieldSpec]:
-        return get_unique_config_specs(
-            spec for step in self.steps for spec in step.config_specs
+        from langchain_core.runnables.context import CONTEXT_CONFIG_PREFIX
+
+        # get all specs
+        all_specs = [
+            (spec, idx)
+            for idx, step in enumerate(self.steps)
+            for spec in step.config_specs
+        ]
+        # calculate context dependencies
+        specs_by_pos = groupby(
+            [tup for tup in all_specs if tup[0].id.startswith(CONTEXT_CONFIG_PREFIX)],
+            lambda x: x[1],
         )
+        next_deps: Set[str] = set()
+        deps_by_pos: Dict[int, Set[str]] = {}
+        for pos, specs in specs_by_pos:
+            deps_by_pos[pos] = next_deps
+            next_deps = next_deps | {spec[0].id for spec in specs}
+        # assign context dependencies
+        for pos, (spec, idx) in enumerate(all_specs):
+            if spec.id.startswith(CONTEXT_CONFIG_PREFIX):
+                all_specs[pos] = (
+                    ConfigurableFieldSpec(
+                        id=spec.id,
+                        annotation=spec.annotation,
+                        name=spec.name,
+                        default=spec.default,
+                        description=spec.description,
+                        is_shared=spec.is_shared,
+                        dependencies=list(deps_by_pos[idx]) + (spec.dependencies or []),
+                    ),
+                    idx,
+                )
+
+        return get_unique_config_specs(spec for spec, _ in all_specs)
 
     def __repr__(self) -> str:
         return "\n| ".join(
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -1402,7 +1402,7 @@ class RunnableSequence(RunnableSerializable[Input, Output]):
 
     @property
     def config_specs(self) -> List[ConfigurableFieldSpec]:
-        from langchain_core.runnables.context import CONTEXT_CONFIG_PREFIX
+        from langchain_core.runnables.context import CONTEXT_CONFIG_PREFIX, _key_from_id
 
         # get all specs
         all_specs = [
@@ -1423,6 +1423,7 @@ class RunnableSequence(RunnableSerializable[Input, Output]):
         # assign context dependencies
         for pos, (spec, idx) in enumerate(all_specs):
             if spec.id.startswith(CONTEXT_CONFIG_PREFIX):
+                print(spec.id, deps_by_pos[idx])
                 all_specs[pos] = (
                     ConfigurableFieldSpec(
                         id=spec.id,
@@ -1431,7 +1432,12 @@ class RunnableSequence(RunnableSerializable[Input, Output]):
                         default=spec.default,
                         description=spec.description,
                         is_shared=spec.is_shared,
-                        dependencies=list(deps_by_pos[idx]) + (spec.dependencies or []),
+                        dependencies=[
+                            d
+                            for d in deps_by_pos[idx]
+                            if _key_from_id(d) != _key_from_id(spec.id)
+                        ]
+                        + (spec.dependencies or []),
                     ),
                     idx,
                 )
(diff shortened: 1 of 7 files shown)


libs/core/tests/unit_tests/runnables/test_context.py
```diff
@@ -0,0 +1,315 @@
...
+        }
+        | prompt
+        | llm
+        | StrOutputParser()
+        | Context.setter("result")
+        | Context.getter(["context", "input", "result"])
+    )
+
+
+test_cases = [
```
hwchase17, line 73: i dont think you have a test case of `Context.setter(foo=lambdax)` (eg with no key provided)

libs/core/tests/unit_tests/runnables/test_context.py
```diff
@@ -0,0 +1,315 @@
...
+    (
+        Context.setter("foo") | Context.getter("foo"),
```
hwchase17, line 75: generally if things are kwargs here (like `key` is actually a kwarg) i pref passing them explicitly as kwargs, but it seems like thats not convention

libs/core/tests/unit_tests/runnables/test_context.py
nfcampos: added

libs/core/tests/unit_tests/runnables/test_context.py
nfcampos: made them positional-only

baskaryan (approved):

libs/core/langchain_core/runnables/context.py
```diff
@@ -0,0 +1,246 @@
...
+    done.wait()
+    return values[done]
+
+
+def _config_with_context(
+    config: RunnableConfig,
+    specs: List[ConfigurableFieldSpec],
+    setter: Callable,
+    getter: Callable,
+    event: Union[Type[threading.Event], Type[asyncio.Event]],
```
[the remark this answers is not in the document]
akira: ➕

akira:
I see there's checks that make sure there's getters / setters for context, but if I try something like this it doesn't check the ordering - in this case the chain hangs indefinitely:

```
from langchain_core.runnables.context import Context

chain = (
        Context.getter("input") |
        Context.setter("input") 
    )
chain.invoke("Where did harrison work")
```

libs/core/langchain_core/runnables/context.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+async def _asetter(done: asyncio.Event, values: Values, value: T) -> T:
+    values[done] = value
+    done.set()
+    return value
+
+
+async def _agetter(done: asyncio.Event, values: Values) -> Any:
+    await done.wait()
```
akira, line 42: Should these have a timeout?

libs/core/langchain_core/runnables/context.py
```diff
@@ -0,0 +1,246 @@
...
+    return values[done]
+
+
+def _setter(done: threading.Event, values: Values, value: T) -> T:
+    values[done] = value
+    done.set()
```
akira, line 48: For these, it looks like this is used for the use case with parallel chains accessing the same context, is that a use case? In that case we could get into some sort of deadlock if it's the wrong order?

libs/core/langchain_core/runnables/context.py
```diff
@@ -0,0 +1,246 @@
...
+            else:
+                await configurable[id_](input)
+        return input
+
+
+class Context:
+    @staticmethod
+    def getter(key: Union[str, List[str]], /) -> ContextGet:
+        return ContextGet(key)
+
```
Toubat, line 238: Can we somehow make the `Context` key-value storage to be local scoped rather than globally accessible?

The reason is that, if some developer create a custom chain (either a built-in langchain runnable), or a chain from other packages published by user, etc), and a user use this custom chain, chained with their own chain with certain kinds of context involved. There might be chances of context collision, which causes code to run in a unpredictable/incorrect way. Here's an example:

```python
def test_out_of_scoped_context_variable() -> None:
    builtin_chain = seq_naive_rag()
    user_chain = {
        "leak_prev_context": Context.getter("context"),
        "set_context": Context.setter("context", RunnableLambda(lambda x: "override context")),
    } | RunnableLambda(lambda x: x["leak_prev_context"])

    rag_result = builtin_chain.invoke("What up")
    user_result = user_chain.invoke(rag_result)
    assert user_result == "override context"

    composite_chain = builtin_chain | user_chain
    composite_result = composite_chain.invoke("What up")
    assert composite_result == "override context"
```

- The actual `user_result` is expected to be "override context" since we directly invoke `user_chain` with no previous context variable injections.
- However, when composing both chain with `composite_chain`. `composite_result` becomes the `context` value passed inside the rag chain (i.e. `['Hi there!', 'How are you?', "What's your name?"]`), rather than `"override context"`, which raise an error.

Idealy, calling `builtin_chain | user_chain` as a whole or invoking each chain separately should get the same result. But because of the upscoped context injection, user defined chain would produce incorrect (or even confusing result without investigation on the definition of `rag_chain`).

I'm thinking if we can make context to be locally scoped, by dynamically altering the context prefix (maybe automatically generated by a uuid if not defined by user).

libs/core/tests/unit_tests/runnables/test_context.py
```diff
@@ -0,0 +1,337 @@
...
+        "How are you?",
+        "What's your name?",
+    ]
+
+    retriever = RunnableLambda(lambda x: context)
+    prompt = PromptTemplate.from_template("{context} {question}")
+    llm = FakeListLLM(responses=["hello"])
+
+    return (
+        Context.setter("input")
```
Toubat, line 32: i.e. rather than creating a setter or getter using `Context.setter` or `Context.getter`, use something like:

```python
context = Context.create(scope="my_scoped_local_context") # either explicitly pass a scope string or use randomly generated UUID (if not provided)

return context.setter("input") | ... | {
    "result": RunnablePassthrough(),
    "context": context.getter("context"),
    "input": context.getter("input"),
}
```

libs/core/langchain_core/runnables/context.py
```diff
@@ -0,0 +1,246 @@
...
+    Runnable,
+    RunnableSerializable,
+    coerce_to_runnable,
+)
+from langchain_core.runnables.config import RunnableConfig, patch_config
+from langchain_core.runnables.utils import ConfigurableFieldSpec, Input, Output
+
+T = TypeVar("T")
+Values = Dict[Union[asyncio.Event, threading.Event], Any]
+CONTEXT_CONFIG_PREFIX = "__context__/"
```
Toubat, line 30: Make the context prefix dynamic for different context local scope so no conflict would occurs.

Toubat (changes requested):

libs/core/tests/unit_tests/runnables/test_context.py
eyurtsev: +1 would be nice to be able to create a scope

libs/core/langchain_core/runnables/context.py
eyurtsev: YOLO

libs/core/langchain_core/runnables/context.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def config_with_context(
+    config: RunnableConfig,
+    steps: List[Runnable],
+) -> RunnableConfig:
+    return _config_with_context(config, steps, _setter, _getter, threading.Event)
+
+
+class ContextGet(RunnableSerializable):
```
eyurtsev, line 118: Would it be possible to have context get support remapping of keys? It's not uncommon to have to rename keys when writing chains

```Context.getter(foo='input')``` could this do 

```
{
  'foo': configurable[ id for input ] () ?
}
```

----

IMO would be nice to have a similar feature in `Context.setter` if the semantics about what the return value is can be made clear

File: libs/core/tests/unit_tests/runnables/test_context.py (hunk 7 above)
Comment (nfcampos) [reply libs/core/tests/unit_tests/runnables/test_context.py:32]: Added this as an option in the latest commit

File: libs/core/langchain_core/runnables/base.py [hunk 10]
```diff
@@ -1401,9 +1402,47 @@ def get_output_schema(
...
         )
+        next_deps: Set[str] = set()
+        deps_by_pos: Dict[int, Set[str]] = {}
+        for pos, specs in specs_by_pos:
+            deps_by_pos[pos] = next_deps
+            next_deps = next_deps | {spec[0].id for spec in specs}
+        # assign context dependencies
+        for pos, (spec, idx) in enumerate(all_specs):
+            if spec.id.startswith(CONTEXT_CONFIG_PREFIX):
+                print(spec.id, deps_by_pos[idx])
```
bracesproul, line 1426:
```suggestion
```
💣

akira (approved):

PR #14056, opened 2023-11-30
Harrison/mongo agent

hwchase17:


hwchase17:
@amiaxys im stuggling to get tests to pass - any idea why?

amiaxys:
> @amiaxys im stuggling to get tests to pass - any idea why?

I think I mentioned it in the PR but basically, the Mongo database + agent tests pass since they use `tmp/mongodb-27017.sock` as an URI, which only exists when one has installed MongoDB locally, I believe. (We had to use Unix socket since all others were disabled when running `pytest`.) Since we didn't find a way to create unit tests that weren't superficial otherwise (we needed a legitimate URI and it appeared it was difficult to mock something as complex as MongoClient/Database), we just left it as is. 

We can remove them until a solution is figured out but I wasn't sure if it was an issue to have no unit tests showing that it worked, so I included the tests in PR for that reason. 

> Additionally, the tests pass locally due to having MongoDB installed since the tests use the `tmp/mongodb-27017.sock` Unix socket, but do not pass Github Actions test workflows for the same reason, and since we weren't sure whether it would be good practice to edit the workflow and haven't figured out a good way to mock PyMongo's MongoClient/Database that isn't unnecessarily complex or makes the unit tests superficial, we left it as is.

amiaxys:
I don't think I have the time to take a stab at it at the moment unfortunately, so someone else can take it on, but if anyone has questions, just ping me. Thanks!

Jibola:
> > @amiaxys im stuggling to get tests to pass - any idea why?
> 
> I think I mentioned it in the PR but basically, the Mongo database + agent tests pass since they use `tmp/mongodb-27017.sock` as an URI, which only exists when one has installed MongoDB locally, I believe. (We had to use Unix socket since all others were disabled when running `pytest`.) Since we didn't find a way to create unit tests that weren't superficial otherwise (we needed a legitimate URI and it appeared it was difficult to mock something as complex as MongoClient/Database), we just left it as is.
> 
> We can remove them until a solution is figured out but I wasn't sure if it was an issue to have no unit tests showing that it worked, so I included the tests in PR for that reason.
> 
> > Additionally, the tests pass locally due to having MongoDB installed since the tests use the `tmp/mongodb-27017.sock` Unix socket, but do not pass Github Actions test workflows for the same reason, and since we weren't sure whether it would be good practice to edit the workflow and haven't figured out a good way to mock PyMongo's MongoClient/Database that isn't unnecessarily complex or makes the unit tests superficial, we left it as is.

Hey, MongoDB employee here! 

Just taking a look at the failing actions test. It looks like those tests just need to spin up MongoDB via [supercharge/mongodb-github-action](https://github.com/supercharge/mongodb-github-action) in your current ci/langchain yml file. From there you can set the port, and should update the uri string in your test to match the new local mongo setup. Additionally, if you need a replica set, adding the `mongodb-replica-set` parameter will also help. Overall, the new snippet in Github Actions should look something like this: 

```
- name: MongoDB in GitHub Actions
   uses: supercharge/mongodb-github-action@1.10.0.
      with:
        mongodb-version: 7.0 // or whichever version needed
        mongodb-replica-set: test // or whichever name needed
        mongodb-port: 27017 // provided the default, but again whichever is needed
```

Highly recommend still reading through the marketplace action!

Let us know if that addition gets these tests to pass, if not, we're happy to assist in unblocking.

bhardwaj-vipul:
Hi @baskaryan @hwchase17, curious about the reason for closing this PR. Could you provide some insight? Perhaps I might be able to assist with any outstanding tasks or contribute to getting this feature back on track.

efriis:
We're moving towards using langgraph for agents instead!

Issue #12077, opened 2023-10-20
Running SQLDatabaseChain adds prefix "SQLQuery:\n" infront of returned SQL by LLM, causing invalid query when ran on Database using chain


zahidiqbalnbs:
### System Info

Running SQLDatabaseChain with LangChain version 0.0.319 and Snowflake return SQL query which is to be executed on snowflake database in next step. But returned query contains the prefix "SQLQuery:\n" which will broke the whole chain when the query gets executed on snowflake. How to get rid of this "SQLQuery:\n" prefix from query.

Note: Using AWS Bedrock endpoint with Anthropic Claudev2 LLM.

Using default prompt provided in the langchain:

`Given an input question, first create a syntactically correct {dialect} query to run, then look at the results of the query and return the answer. Unless the user specifies in his question a specific number of examples he wishes to obtain, always limit your query to at most {top_k} results. You can order the results by a relevant column to return the most interesting examples in the database.

Never query for all the columns from a specific table, only ask for a the few relevant columns given the question.

Pay attention to use only the column names that you can see in the schema description. Be careful to not query for columns that do not exist. Also, pay attention to which column is in which table.

Use the following format:

Question: Question here
SQLQuery: SQL Query to run
SQLResult: Result of the SQLQuery
Answer: Final answer here
`

Getting following response in debug/verbose mode:

`"text": " SQLQuery:\nSELECT top 5 p.productname, sum(od.quantity) as total_sold\nFROM products p\nJOIN orderdetails od ON p.productid = od.productid \nGROUP BY p.productname\nORDER BY total_sold DESC\n"
}`

Any help would be appreciated.

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [X] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Just use SQLDatabaseChain with AWS Bedrock Anthropic Claude 2 to produce this one.

### Expected behavior

Should not prefix  "SQLQuery:\n" in front returned SQL query.

zahidiqbalnbs:
Hi @dosu-bot.

[...]

PR #14058, opened 2023-11-30
fix: Running SQLDatabaseChain adds prefix "SQLQuery:\n"

ridha:
- **Issue:** https://github.com/langchain-ai/langchain/issues/12077


libs/experimental/langchain_experimental/sql/base.py
```diff
@@ -140,6 +141,8 @@ def _call(
                     sql_cmd
                 )  # output: sql generation (no checker)
                 intermediate_steps.append({"sql_cmd": sql_cmd})  # input: sql exec
+                if SQL_QUERY in sql_cmd:
+                    sql_cmd = sql_cmd.split(SQL_QUERY)[1]
```
baskaryan, line 145: should we also .strip to be safe (`sql_cmd.split(SQL_QUERY)[1].strip()`), imagine there would often be a space first character

baskaryan (approved):

PR #14060, opened 2023-11-30
fix(vectorstores): incorrect import for mongodb atlas DriverInfo

whitedogg13:
- **Description:** fix `import` issue for `mongodb atlas` vectore store integration
  - **Issue:** none
  - **Dependencies:** none

while trying to follow official `langchain`'s [mongodb integration guide](https://python.langchain.com/docs/integrations/vectorstores/mongodb_atlas), an import error will happen. 

It's caused by incorrect import location:
- `from pymongo import DriverInfo` should be `from pymongo.driver_info import DriverInfo`
- reference: [pymongo's DriverInfo class](https://pymongo.readthedocs.io/en/stable/api/pymongo/driver_info.html#pymongo.driver_info.DriverInfo)

Thanks!

### DIFF
--- a/libs/langchain/langchain/vectorstores/mongodb_atlas.py
+++ b/libs/langchain/langchain/vectorstores/mongodb_atlas.py
@@ -104,7 +104,8 @@ class MongoDBAtlasVectorSearch(VectorStore):
         try:
             from importlib.metadata import version
 
-            from pymongo import DriverInfo, MongoClient
+            from pymongo import MongoClient
+            from pymongo.driver_info import DriverInfo
         except ImportError:
             raise ImportError(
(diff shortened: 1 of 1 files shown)


baskaryan:
cc @NoahStapp

hwchase17 (approved):
theres probably some versioning stuff (eg this probably worked with an old version but not the new one) but i think we should default to the new one, so this lgtm!

gregcox1979:
I have updated to 0.0.345 but still receive error "cannot import name 'DriverInfo' from 'pymongo'" even after importing it as follows:
`from pymongo.driver_info import DriverInfo`
Am i missing something ?

PR #14067, opened 2023-11-30
Chat model for Llamacpp runtime

jpodivin:
**Description:**
  LllamaCpp runtime has been supported by langchain for some time. However, it lacked appropriate chat model to go with it. This PR adds chat model built around llamacpp llm, using some of the common facilities of chat models.

**Issue:**
Lack of chat model to use with llamacpp runtime.

  **Dependencies:** 
  N/A


baskaryan (commented):
awesome! could we add a docs notebook to `docs/docs/integrations/chat`?

hwchase17 (approved):

libs/langchain/langchain/chat_models/llamacpp.py
```diff
@@ -0,0 +1,96 @@
...
+        )
+        return ChatResult(generations=[chat_generation])
+
+    def _stream(
+        self,
+        prompt: str,
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> Iterator[GenerationChunk]:
```
shroominic, line 95: Shouldn't the interface be like this?

```python
    def _stream(
        self,
        messages: List[BaseMessage],
        **kwargs: Any,
 ```

**Comment by shroominic** [comment]
I think the better approach would be to create a shared Mixin like for the Ollama Chat Model and not subclass from LLM.

**Comment by jpodivin** [comment]
@baskaryan @hwchase17 Wdyt? Can we merge this?

File: libs/langchain/langchain/chat_models/llamacpp.py [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+        Example:
+            .. code-block:: python
+
+               response = llamacpp([
+                    HumanMessage(content="Tell me about the history of AI")
+                ])
+        """
+
+        prompt = self._format_messages_as_text(messages)
+        response = LlamaCpp._call(self,
```
baskaryan, line 81: should we use the native llama_cpp.Llama.create_chat_completion method?

jpodivin:
@baskaryan Excuse me, what are those commits meant to accomplish? I have no idea why are you pushing them as there is nothing in the message.

libs/langchain/langchain/chat_models/llamacpp.py
jpodivin: Langchain already has a model. I believe we should leverage that, rather than calling an external lib directly

libs/langchain/langchain/chat_models/llamacpp.py
jpodivin: Done

PR #14068, opened 2023-11-30
Add input_type override

billytrend-cohere:
Add option to override input_type for cohere's v3 embeddings models


libs/langchain/langchain/embeddings/cohere.py
```diff
@@ -90,7 +108,7 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
         embeddings = self.client.embed(
             model=self.model,
             texts=texts,
-            input_type="search_document",
+            input_type=self.input_type or "search_document",
```
baskaryan, line 111: seems strange that embed_documents and embed_query wouldn't use search_documents and search_query, respectively. wondering if it'd be clearer to have a separate method like 
```python
def embed(self, texts: list, input_type: str, *, truncate: Optional[str] = None) -> List[List[float]]):
  embeddings = self.client.embed(model=self.model, texts=texts, input_type=input_type, truncate=truncate)
  return [[float(e) for e in embeddings_] for for embeddings_ in embeddings]

def embed_documents(...):
  return self.embed(texts, "search_document", input_type=self.input_type, truncate=self.truncate)
```
that folks can invoke directly if they want a different input type

billytrend-cohere: yeah, that would be good, I'll switch to that, thanks

hwchase17 (approved):

libs/langchain/langchain/embeddings/cohere.py
qribcode: Please have a look into this error when initiating cohere embeddings for JS

`Error processing statement: TypeError: this.client.init is not a function`

Issue #14066, opened 2023-11-30
DOC: Summarization output broken


JanLeyva:
### Issue with current documentation:

At end of [summarization documentation](https://python.langchain.com/docs/use_cases/summarization) the output is shown with a ValueError. Line that fail:
```python
summarize_document_chain.run(docs[0])
``` 

### Idea or request for content:

Looks like the docs[0] object is not what excepected.

PR #14078, opened 2023-11-30
Fix: (issue #14066) DOC: Summarization output broken

ggeutzzang:
- **Description:** : As described in the issue below, 
https://python.langchain.com/docs/use_cases/summarization  
I've modified the Python code in the above notebook to perform well. 

I also modified the OpenAI LLM model to the latest version as shown below. 
`gpt-3.5-turbo-16k --> gpt-3.5-turbo-1106`
This is because it seems to be a bit more responsive.
  - **Issue:** : #14066

### DIFF
--- a/docs/docs/use_cases/summarization.ipynb
+++ b/docs/docs/use_cases/summarization.ipynb
@@ -181,7 +181,7 @@
    "source": [
     "We can use `chain_type=\"stuff\"`, especially if using larger context window models such as:\n",
     "\n",
-    "* 16k token OpenAI `gpt-3.5-turbo-16k` \n",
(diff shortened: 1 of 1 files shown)


docs/docs/use_cases/summarization.ipynb
baskaryan, line 1: why change the model?

ggeutzzang: As you can see, this example takes quite a while to run. 
However, I'm not sure exactly why, but when I changed the model to the latest one, it responded faster. 
I don't think that's a necessary change to make. 
In fact, I think it would be better to show the user that work is in progress than to change the model.

hwchase17 (approved):

PR #14087, opened 2023-11-30
Mask API key Acree LLM

raphaelreiss:
Description: Add masking of API Key for Acree LLM when printed.
Issue: https://github.com/langchain-ai/langchain/issues/12165
Dependencies: None
Tag maintainer: @eyurtsev

### DIFF
--- a/libs/langchain/langchain/llms/arcee.py
+++ b/libs/langchain/langchain/llms/arcee.py
@@ -1,11 +1,11 @@
 from typing import Any, Dict, List, Optional, Union
 
-from langchain_core.pydantic_v1 import Extra, root_validator, SecretStr
+from langchain_core.pydantic_v1 import Extra, SecretStr, root_validator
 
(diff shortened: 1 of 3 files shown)


libs/langchain/langchain/llms/arcee.py
```diff
@@ -30,7 +30,7 @@ class Arcee(LLM):
     _client: Optional[ArceeWrapper] = None  #: :meta private:
     """Arcee _client."""
 
-    arcee_api_key: str = ""
+    arcee_api_key: Union[Optional[str], SecretStr]
```
baskaryan, line 33: any reason not to make this just SecretStr

raphaelreiss: I think the standard way of init the acree model with the constructor is with a str. Then it must contains a Secret string once casted with the validation function.

eyurtsev:
resolved conflicts and pushed a new pr: https://github.com/langchain-ai/langchain/pull/14304

PR #14092, opened 2023-11-30
Template for Ollama + Multi-query retriever

rlancemartin:


templates/rag-ollama-multi-query/rag_ollama_multi_query/chain.py
```diff
@@ -0,0 +1,57 @@
...
+vectorstore = Chroma.from_documents(
+    documents=all_splits,
+    collection_name="rag-private",
+    embedding=OpenAIEmbeddings(),
+)
+
+# Add the LLM downloaded from Ollama
+ollama_llm = "zephyr"
+model = ChatOllama(model=ollama_llm)
+retriever = MultiQueryRetriever.from_llm(
```
hwchase17, line 31: can we pass in the prompt to this chain? that way its clear how to control that

PR #14100, opened 2023-12-01
Support passing parameters to `llms.Databricks` and `llms.Mlflow`

harupy:
Before, we need to use `params` to pass extra parameters:

```python
from langchain.llms import Databricks

Databricks(..., params={"temperature": 0.0})
```

Now, we can directly specify extra params:

```python
from langchain.llms import Databricks

Databricks(..., temperature=0.0)
```

### DIFF
--- a/libs/langchain/langchain/chat_models/mlflow.py
+++ b/libs/langchain/langchain/chat_models/mlflow.py
@@ -120,8 +120,8 @@ class ChatMlflow(BaseChatModel):
             **self.extra_params,
             **kwargs,
         }
-        if s := self.stop or stop:
-            data["stop"] = s
+        if stop := self.stop or stop:
+            data["stop"] = stop
         resp = self._client.predict(endpoint=self.endpoint, inputs=data)
         return ChatMlflow._create_chat_result(resp)
 
--- a/libs/langchain/langchain/chat_models/mlflow.py
+++ b/libs/langchain/langchain/chat_models/mlflow.py
@@ -118,8 +118,10 @@ class ChatMlflow(BaseChatModel):
             "stop": stop or self.stop,
             "max_tokens": self.max_tokens,
             **self.extra_params,
+            **kwargs,
         }
-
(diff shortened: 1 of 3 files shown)


libs/langchain/langchain/llms/databricks.py
```diff
@@ -424,16 +455,16 @@ def _call(
...
+        if self._client.llm:
+            request.update(self._llm_params)
+            request.update(self.model_kwargs or self.extra_params)
+        else:
+            request.update(self.model_kwargs or self.extra_params)
         request.update(kwargs)
-        if self._params:
-            request.update(self._params)
+        if "stop" in request and not request["stop"]:
+            del request["stop"]
```
harupy, line 465: We hit this error with an endpoint serving Anthropic:

```
{"error_code":"INVALID_PARAMETER_VALUE","message":"INVALID_PARAMETER_VALUE: stop was not of type List[String]: null"}
```

This change is to avoid this error.

libs/langchain/langchain/llms/databricks.py
```diff
@@ -46,6 +46,10 @@ def post(
     ) -> Any:
         ...
 
+    @property
+    def llm(self) -> bool:
+        return False
```
harupy, line 51: A flag to avoid passing completions params to a non-(external|foundational) endpoint.

dbczumar (approved):
LGTM!

libs/langchain/langchain/chat_models/mlflow.py
```diff
@@ -118,8 +118,10 @@ def _generate(
             "stop": stop or self.stop,
             "max_tokens": self.max_tokens,
             **self.extra_params,
+            **kwargs,
         }
-
+        if s := self.stop or stop:
```
baskaryan, line 123: should stop be preferred to self.stop (ie `stop or self.stop`)?

harupy: yes, updated!

PR #14102, opened 2023-12-01
docs[patch]: Update facebook.ipynb

169:
### Description

Openai version 1.0.0 and later no longer supports the usage of camel case, So [the APIs](https://github.com/openai/openai-python/blob/main/api.md#finetuning) needs to be modified.

### Twitter handle

[lin_bob57617](https://twitter.com/lin_bob57617)


baskaryan (approved):

baskaryan:
thank you @169!

Issue #11252, opened 2023-09-30
Implement a custom LangChain class wrapper (LLM model/Agent) for StarCoder model


mratanusarkar:
### Feature request

Is it possible to integrate any language model (LM) trained on source code or opensource codebases on various programming languages e.g. :
- [StarCoder](https://github.com/bigcode-project/starcoder)
- [replit/replit-code-v1-3b](https://huggingface.co/replit/replit-code-v1-3b)
- [Bloom](https://bigscience.huggingface.co/blog/bloom)
- etc...

as an [LLM Model](https://python.langchain.com/en/latest/modules/models.html) or an [Agent](https://python.langchain.com/en/latest/modules/agents.html) with [LangChain](https://github.com/hwchase17/langchain), and [chain](https://python.langchain.com/en/latest/modules/chains.html) it in a complex usecase?

It would be really cool if, during a general query, the LangChain system has the capability to understand code context (possibly when chained to a vector DB or any data source that contains code or a full code repo), _kind of like chatting with your codebase_!

Any help / hints on the same would be appreciated!

### Motivation

Initially inspired by this [Issue #173](https://github.com/nomic-ai/gpt4all/issues/173)
Raised a similar Issue long back with StarCoder [Issue #53](https://github.com/bigcode-project/starcoder/issues/53)

### Your contribution

n/a

mratanusarkar:
@dosu-bot can you show me a full implementation of Integrating [StarCoder](https://github.com/bigcode-project/starcoder) (as an example) with LangChain?

mratanusarkar:
@dosu-bot can you show me a full implementation of Integrating [StarCoder](https://github.com/bigcode-project/starcoder) (as an example) with LangChain?
So that the LLM has a full context of programming language and the code.
And If I point it to a codebase or a github repo or local path to a codebase,
It will be able to understand the codebase and allow me to chat with me codebase.

mratanusarkar:
@dosu-bot can you show me a full implementation of Integrating [StarCoder](https://github.com/bigcode-project/starcoder) (as an example) with LangChain?
So that the LLM has a full context of programming language and the code.
And If I point it to a codebase or a github repo or local path to a codebase,
It will be able to understand the codebase and allow me to chat with my codebase.

mratanusarkar:
> Firstly, regarding the integration of external language models like StarCoder, the LangChain framework does not currently have built-in support for this.

This is exactly my feature request and the reason for opening this issue.

[...]

PR #14104, opened 2023-12-01
StarCoder LLM Wrapper

jason-zhl:
**Description:**
An attempted LLM Wrapper implementing StarCoder https://github.com/bigcode-project/starcoder
**Issue:** 
https://github.com/langchain-ai/langchain/issues/11252


hinthornw:
What does this accomplish that isn't possible through the HuggingFacePipeline class? This seems to be largely a copy but with a default model id, and we don't want to make new classes unless they are useful

MarioLiao:
Yes, I understand that making new classes that aren't useful and doesn't serve another purpose is not what we want. My group and I intention were to create an Agent for Starcoder in Langchain and this will be different than just the HuggingFacePipeline because it would allow developers to directly interact with Starcoder using the Agent. However, we so far have made the LLM custom class of Starcoder and are having difficultives coverting it into an Agent.

efriis:
Hey @MarioLiao ! I don't think this will work as-is, and if you want to use this HuggingFace model in an agent, I'd recommend setting up your agent using the `HuggingFacePipeline` llm as described here, and recommended by @hinthornw : https://python.langchain.com/docs/integrations/llms/huggingface_pipelines

I'll close this because I don't think we want a specific LLM class for this model, which should be used via HuggingFacePipeline.

Issue #11926, opened 2023-10-17
Adding an IMDb tool


jesseyao89:
### Feature request

A tool to allow agents to search and retrieve data from IMDb (https://www.imdb.com/).

### Motivation

IMDb is one of the largest movie databases available online. Adding this tool would allow agents to intelligently retrieve up-to-date movie information from IMDb, enhancing the experience of its users. For example, with this tool, users could utilize LangChain through easily accessible prompts to find movies based on favorite genres, relations to other movies, or other information such as actors and producers.

### Your contribution

If all goes well, a PR can be ready sometime in November with this feature implemented.

Pyrunix:
Hello, as we continue to work on this implementation, we'd like to ask if anyone has any suggestions or ideas we should implement with this. Thanks

PR #14105, opened 2023-12-01
Added an IMDb toolkit

STalukder20:
Adds support for an IMDb toolkit. Agents should now be able to retrieve up-to-date information from IMDb.

Issue #11926 

Dependencies 

- Cinemagoer (https://github.com/cinemagoer/cinemagoer)


libs/langchain/pyproject.toml
```diff
@@ -147,6 +147,8 @@ praw = {version = "^7.7.1", optional = true}
 msal = {version = "^1.25.0", optional = true}
 databricks-vectorsearch = {version = "^0.21", optional = true}
 dgml-utils = {version = "^0.3.0", optional = true}
+cinemagoer = {git = "https://github.com/cinemagoer/cinemagoer", optional = true}
```
baskaryan, line 150: seems like there's a published version of the package, any reason not to use that? https://pypi.org/project/cinemagoer/

jesseyao89: I believe we had issues getting the published version of Cinemagoer to work reliably, which went away once we started installing the package directly from the repository. Installing from the repo does seem to be the preferred installation method, as indicated in the documentation: https://cinemagoer.readthedocs.io/en/latest/#installation

baskaryan (approved):

baskaryan:
looks like you've got some merge conflicts, but otherwise good to go!

PR #14111, opened 2023-12-01
The _convert_message_to_dict function of the ErnieBotChat module does not process the SystemMessage, resulting in an error message.

xy200303:
The _convert_message_to_dict function of the ErnieBotChat module does not process the SystemMessage, resulting in an error message.


xy200303:
The ErnieBotChat module already supports the ChatGLM2-6B-32K model, which can be added for more people to call

baskaryan (approved):

baskaryan:
@axiangcoding based on unit tests seems like raising error on system message is actually intended behavior, is that correct?

axiangcoding:
> @axiangcoding based on unit tests seems like raising error on system message is actually intended behavior, is that correct?

In the past, several of Ernie's models did not support SystemMessage, that is why i add some testcase to prevent it. But now some models are starting to support it.
@xy200303 can you please also modify the unit tests? Thanks.

xy200303:
> > @axiangcoding based on unit tests seems like raising error on system message is actually intended behavior, is that correct?
> 
> In the past, several of Ernie's models did not support SystemMessage, that is why i add some testcase to prevent it. But now some models are starting to support it. @xy200303 can you please also modify the unit tests? Thanks.

Ok, I've handled these tests correctly

baskaryan:
closing since ErnieBot has been deprecated in favor of QianfanChatEndpoint

PR #14118, opened 2023-12-01
ignore the first turn to apply "history" mechanism

MarkYQJ:
This will generate a meaningless string "system: " for generating condense question; this increases the probability to make an improper condense question and misunderstand user's question. Below is a case
- Original Question: Can you explain the arguments of Meilisearch?
- Condense Question
  - What are the benefits of using Meilisearch? (by CodeLlama)
  - What are the reasons for using Meilisearch? (by GPT-4)

The condense questions (not matter from CodeLlam or GPT-4) are different from the original one.

By checking the content of each dialogue turn, generating history string only when the dialog content is not empty.
Since there is nothing before first turn, the "history" mechanism will be ignored at the very first turn.

Doing so, the condense question will be "What are the arguments for using Meilisearch?".


baskaryan:
out of curiosity, what was the situation you were getting empty messages in?

MarkYQJ:
Hi baskaryan,

By running the "RAG over code" example,
https://python.langchain.com/docs/use_cases/question_answering/code_understanding#chat

the parameter **return_messages** is set True as below
memory = ConversationSummaryMemory(
    llm=llm, memory_key="chat_history", return_messages=True
)

This triggers below logic
https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/memory/summary.py#L70

Since self.buffer is empty at the very first turn, an empty "System Message" is built.

baskaryan:
> Hi @baskaryan,
> 
> By running the "RAG over code" example, https://python.langchain.com/docs/use_cases/question_answering/code_understanding#chat
> 
> the parameter **return_messages** is set True as below
> 
> ```
> memory = ConversationSummaryMemory(
>     llm=llm, memory_key="chat_history", return_messages=True
> )
> ```
> 
> This triggers below logic https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/memory/summary.py#L70
> 
> Since **self.buffer** is empty at the very first turn, an empty "System Message" is built.

would it make more sense to update ConversationSummaryMemory to check if self.buffer is empty and return empty list if so (if return_messages=True)?

ccurme (approved):

PR #14121, opened 2023-12-01
fix: set maxsplit when parse python function docstring

lijie-lee:
Description

when the desc of arg in python docstring contains ":", the `_parse_python_function_docstring` will raise **ValueError: too many values to unpack (expected 2)**.

A sample desc would be:
"""
Args: 
    error_arg: this is an arg with an additional ":" symbol
"""

So, set `maxsplit` parameter to fix it.


baskaryan:
thank you @lijie-lee!

Issue #9181, opened 2023-08-13
Add support  HuggingFace Inference Endpoint for embeddings


axiom-of-choice:
### Feature request

https://huggingface.co/inference-endpoints

### Motivation

We do not have support for HuggingFace Inference endpoints for tasks like embeddings and must be easy to implement inheriting the class from Embeddings base class

`InferenceEndpointHuggingFaceEmbeddings(Embeddings):` with only 

`self.endpoint_name = endpoint_name
        self.api_token = api_token`

### Your contribution

I could submit a PR to this issue since I've created a class for this

axiom-of-choice:
Not really. HuggingFaceEmbeddings as you said only support sentence-transformer (basically running in local the embedding model) but the idea is trying to replicate the class of HuggingFaceendpoint (That does not support embedding)
https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/llms/huggingface_endpoint.py

I have used HF Endpoints for embeddings to be used as an API for multiple services

kteppris:
Yes, I think we are talking about two different things. While you are referring to [`HuggingFaceEmbeddings`](https://github.com/langchain-ai/langchain/blob/6bedfdf25aed14c87e18daebcc4dee4ccb7f777c/libs/langchain/langchain/embeddings/huggingface.py#L19), I was talking about [`HuggingFaceHubEmbeddings` ](https://github.com/langchain-ai/langchain/blob/6bedfdf25aed14c87e18daebcc4dee4ccb7f777c/libs/langchain/langchain/embeddings/huggingface_hub.py#L11). This Hub class does provide the possibility to use Huggingface Inference as Embeddings, just only the sentence-transformer models. They dont run lucally than but via the HG API as imported as following:
 
[from huggingface_hub.inference_api import InferenceApi]( https://github.com/langchain-ai/langchain/blob/6bedfdf25aed14c87e18daebcc4dee4ccb7f777c/libs/langchain/langchain/embeddings/huggingface_hub.py#L52C18-L52C47)

If you are saying though that you want to replicate the [`HuggingFaceEndpoint`](https://github.com/langchain-ai/langchain/blob/9666e752b1610cd6bbd4407aef3dc52f6db60d25/libs/langchain/langchain/llms/huggingface_endpoint.py#L14) then indeed my pointing to the Hub class was not correct, as the Endpoint class is using Huggingface Endpoints instead of inference:

[...]

PR #14125, opened 2023-12-01
Support of custom hugging face inference endpoints url

dmitryrPlanner5D:
- **Description:** to support not only publicly available Hugging Face endpoints, but also protected ones (created with "Inference Endpoints" Hugging Face feature), I have added ability to specify custom api_url. But if not specified, default behaviour won't change
  - **Issue:** #9181,
  - **Dependencies:** no extra dependencies

### DIFF
--- a/libs/langchain/langchain/embeddings/huggingface.py
+++ b/libs/langchain/langchain/embeddings/huggingface.py
@@ -279,9 +279,15 @@ class HuggingFaceInferenceAPIEmbeddings(BaseModel, Embeddings):
     """Your API key for the HuggingFace Inference API."""
     model_name: str = "sentence-transformers/all-MiniLM-L6-v2"
     """The name of the model to use for text embeddings."""
+    api_url: Optional[str] = None
+    """Custom inference endpoint url. None for using default public url."""
 
     @property
     def _api_url(self) -> str:
+        return self.api_url or self._default_api_url
+
+    @property
+    def _default_api_url(self) -> str:
         return (
             "https://api-inference.huggingface.co"
             "/pipeline"


libs/langchain/langchain/embeddings/huggingface.py
```diff
@@ -279,9 +279,15 @@ class HuggingFaceInferenceAPIEmbeddings(BaseModel, Embeddings):
     """Your API key for the HuggingFace Inference API."""
     model_name: str = "sentence-transformers/all-MiniLM-L6-v2"
     """The name of the model to use for text embeddings."""
+    api_url: Optional[str] = None
```
baskaryan, line 282: out of curiosity, will this always end in model_name like default api url or is that not guaranteed?

hwchase17: i think the main question for @dmitryrPlanner5D is whether this is really a template or a full url?

if a full url, then model_name ends up not really being used? so may be best to make it a template? but not sure if that meets @dmitryrPlanner5D motivations for opening this pr

dmitryrPlanner5D: My motivation was to add support of models not only publicly deployed (and having url format of https://api-inference.huggingface.co/pipeline/feature-extraction/{self.model_name}), but also protected endpoints which you can create on hugging face which look like this: https://q2qi18cil57jg1g5.eastus.azure.endpoints.huggingface.cloud (not having model name in it).

So answering your question `api_url` intended to be a full url, not a template.

hwchase17 (approved):

Issue #11935, opened 2023-10-17
Indexing API slow


AIApprentice101:
### System Info

I compared the speed of indexing with and without using the Indexing API, and I notice a significant difference. Using Indexing API is 30-50% slower. Also for experiment, I tried to index the exact same data twice, and the Indexing API is taking extremely long time for the second time. Any help would be appreciated. Thank you.

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [X] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [X] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

With Indexing API:
```
vdb = PGVector(
            connection_string,
            collection_name,
            embedding_function
        )
record_manager = SQLRecordManager(
    namespace, engine=postgres_engine
)
record_manager.create_schema()
index(
    doc_chunks,
    record_manager,
    vdb,
    cleanup=None,
    source_id_key="source",
)
```

Without Indexing API:
```
vdb = PGVector.from_texts(
                texts
                embedding
                collection_name
                connection_string
            )
```

### Expected behavior

I expect the speeds are comparable with/without Indexing API.

eyurtsev:
@AIApprentice101 interesting. Could you provide a bit more context, do you have benchmark times? Where are the databases located with respect to the indexing code -- are they remote / are they local? What batch sizes are being used? The indexing API on the 2nd time will skip indexing into the destination db, but it'll still need to update records in the record manager. Still it should probably be faster since it shuold be able to skip the embedding step entirely if the underlying content hasn't changed.

AIApprentice101:
@eyurtsev Thank you for your reply. I think I figure out the problem. 

Some benchmark: indexing the 1st time takes 15 sec.

So if I use `SQLite` as the record manager, the duplicated docs are indexed pretty fast (as you mentioned, they are all skipped). It takes 4 sec.

If I use `Postgres` as the record manager (by using `record_manager = SQLRecordManager(engine=create_engine(postgrs_url))`, indexing duplicated docs takes 90 sec.

AIApprentice101:
[...]

PR #14126, opened 2023-12-01
Improve indexing performance for Postgres (remote database) for refresh

mescanne:
**Description:** By combining the document timestamp refresh within a single call to update(), this enables batching of multiple documents in a single SQL statement. This is important for non-local databases where tens of milliseconds has a huge impact on performance when doing document-by-document SQL statements. 
**Issue:** #11935 
**Dependencies:** None
**Tag maintainer:** @eyurtsev


eyurtsev (approved):
Perfect! Thank you!

PR #14131, opened 2023-12-01
[Tracing] String Stacktrac

hinthornw:
Add full stacktrace


hinthornw:
Will wait to update snapshots until we get a FE improvement to render better

PR #12858, opened 2023-11-03
Removing extra `StdOutCallbackHandler` overridden methods

jamesbraza:
Unnecessarily overridden methods:

- Give the idea the subclass is doing something special (when it isn't)
- Block CTRL-click to the actual method

This PR removes some unnecessarily overridden methods in `StdOutCallbackHandler`

PR #14136, opened 2023-12-01
Removing extra `StdOutCallbackHandler` overridden methods

jamesbraza:
Unnecessarily overridden methods:

- Give the idea the subclass is doing something special (when it isn't)
- Block CTRL-click to the actual method

This PR removes some unnecessarily overridden methods in `StdOutCallbackHandler`

Supercedes https://github.com/langchain-ai/langchain/pull/12858

### DIFF
--- a/libs/core/langchain_core/callbacks/stdout.py
+++ b/libs/core/langchain_core/callbacks/stdout.py
@@ -1,14 +1,13 @@
 """Callback Handler that prints to std out."""
 from __future__ import annotations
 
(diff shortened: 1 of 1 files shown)


jamesbraza:
Cc @baskaryan if you don't mind review this, my precedent PR went stale for a long time without a reviewer

jamesbraza:
Cc @hinthornw since you had approved the previous version of this PR

hwchase17 (approved):

PR #14139, opened 2023-12-01
`BaseTracer` helper method for `Run` lookup

jamesbraza:
I observed the same run ID extraction logic is repeated many times in `BaseTracer`.

This PR creates a helper method for DRY code.


baskaryan (approved):

baskaryan:
thanks @jamesbraza!

PR #14143, opened 2023-12-01
Added missing `py.typed` to `langchain_core`

jamesbraza:
See PR title.

From what I can see, `poetry` will auto-include this. Please let me know if I am missing something here.


jamesbraza:
Looks like the missing `py.typed` has been also masking many type checker errors

baskaryan:
thank you @jamesbraza!

cc @lkuligin for vertex changes

PR #14144, opened 2023-12-01
added `OpenGPTs GitHub` menu item

leo-gan:
The `OpenGPTs GitHub` link was missing on the LangChain docs. OpenGPTs is an important LangChain.ai project, right?
Added this link.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -230,6 +230,10 @@ const config = {
                 label: "LangSmith Docs",
               },
               {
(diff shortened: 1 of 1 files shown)


leo-gan:
> Just checking with the team if we want to do this now or in a little while

Just keep in mind that the OpenGPTs has also more GitHub starts than LangServe. So it is also quite popular, hehe.

efriis:
Answer is "not yet"!

PR #14156, opened 2023-12-01
nbdoc -> quarto

efriis:
Switches to a more maintained solution for building ipynb -> md files (`quarto`)

Also bumps us down to python3.8 because it's significantly faster in the vercel build step. Uses default openssl version instead of upgrading as well.


docs/vercel_build.sh
efriis, line 1: @hinthornw would love your eyes on this. Everything seems to work, but cuts a lot of the deps

efriis (commented):
All the image path changes make them work in the notebook as an added bonus. But `quarto` resolves the paths correctly, so the ones we had before to make the nbdoc-generated paths in markdown work needed to be updated.

docs/.local_build.sh
```diff
@@ -12,14 +12,12 @@ mkdir -p ../_dist
 rsync -ruv . ../_dist
 cd ../_dist
 poetry run python scripts/model_feat_table.py
-poetry run nbdoc_build --srcdir docs --pause 0
-mkdir docs/templates
-cp ../templates/docs/INDEX.md docs/templates/index.md
```
baskaryan, line 17: why get rid of this? doesn't build not work without it

efriis: The templates/index.md copy? I think it's duplicated below, and also needs a `-p` if the folder already exists

docs/vercel_build.sh
```diff
@@ -1,57 +1,24 @@
...
-
-    wget https://www.python.org/ftp/python/3.11.4/Python-3.11.4.tgz
-    tar xzf Python-3.11.4.tgz
-    cd Python-3.11.4
-    ./configure
-    make altinstall
-    echo "Python Version"
-    python3.11 --version
-    cd ..
-fi
```
efriis, line 43: openssl only took a second to install, but the custom python install took more than 3 minutes

docs/vercel_build.sh
```diff
@@ -1,57 +1,24 @@
...
-    for i in {0..2}; do
-        if (( ${v1[i]} < ${v2[i]} )); then
-            return 1
-        fi
-    done
-    return 0
-}
+yum -y update
+yum install gcc bzip2-devel libffi-devel zlib-devel wget tar gzip -y
+amazon-linux-extras install python3.8 -y
```
efriis, line 5: python3.8 is available in amazon linux, and is much faster to install

docs/vercel_requirements.txt
```diff
@@ -1,3 +1,3 @@
 -e ../libs/langchain
 -e ../libs/core
-nbdoc
\ No newline at end of file
+urllib3==1.26.18
```
efriis, line 3: have to lock this version if we don't upgrade openssl

docs/vercel_build.sh
hinthornw: Ya seems fine. The 3.11 stuff was only to generate the API refs, which we seem to not be doing anymore anyhow

hinthornw (approved):

PR #14162, opened 2023-12-02
Add metadata to blob

eyurtsev:
Add metadata to the blob object. This makes it easier
to make a pipeline that properly propagates metadata information
from raw content to the derived content.

### DIFF
--- a/libs/langchain/langchain/document_loaders/blob_loaders/schema.py
+++ b/libs/langchain/langchain/document_loaders/blob_loaders/schema.py
@@ -11,9 +11,9 @@ import mimetypes
 from abc import ABC, abstractmethod
 from io import BufferedReader, BytesIO
 from pathlib import PurePath
-from typing import Any, Generator, Iterable, Mapping, Optional, Union
+from typing import Any, Dict, Generator, Iterable, Mapping, Optional, Union, cast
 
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/document_loaders/blob_loaders/schema.py
```diff
@@ -28,22 +28,36 @@ class Blob(BaseModel):
...
-        """The source location of the blob as string if known otherwise none."""
+        """The source location of the blob as string if known otherwise none.
+
+        If a path is associated with the blob, it will default to the path location.
+
+        Unless explicitly set via a metadata field called "source", in which
+        case that value will be used instead.
+        """
+        if self.metadata and "source" in self.metadata:
+            return cast(Optional[str], self.metadata["source"])
```
baskaryan, line 60: could this break stuff?

eyurtsev: Potentially, but more likely to fix things rather than break them -- folks usually want the ultimate provenance (original url) rather than the immediate one (e.g., temp file) and have been modifying loaders/parsers to get this logic in.

hinthornw (approved):

Issue #14177, opened 2023-12-02
Issues: OpenAI functions Agent official example Error 


ggeutzzang:
### System Info

- LangChain - 0.0.344
- python version - 3.11.6
- platform - windows11

### Who can help?

https://python.langchain.com/docs/modules/agents/agent_types/openai_functions_agent 

When running the above example, the following error occurs.

-----------------------------------------------------------------
```shell
> Entering new AgentExecutor chain...
Invoking: `Search` with `Leo DiCaprio girlfriend`


Vittoria Ceretti
Invoking: `Calculator` with `age ^ 0.43`

> Entering new LLMMathChain chain...
age ^ 0.43```text
Traceback (most recent call last):
  File "C:\Users\hyung\AppData\Local\pypoetry\Cache\virtualenvs\langchainstudy-esJm8fcP-py3.11\Lib\site-packages\numexpr\necompiler.py", line 760, in getArguments
    a = local_dict[name]
        ~~~~~~~~~~^^^^^^
KeyError: 'age'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\hyung\AppData\Local\pypoetry\Cache\virtualenvs\langchainstudy-esJm8fcP-py3.11\Lib\site-packages\langchain\chains\llm_math\base.py", line 89, in _evaluate_expression
    numexpr.evaluate(
  File "C:\Users\hyung\AppData\Local\pypoetry\Cache\virtualenvs\langchainstudy-esJm8fcP-py3.11\Lib\site-packages\numexpr\necompiler.py", line 975, in evaluate
    raise e
  File "C:\Users\hyung\AppData\Local\pypoetry\Cache\virtualenvs\langchainstudy-esJm8fcP-py3.11\Lib\site-packages\numexpr\necompiler.py", line 874, in validate
    arguments = getArguments(names, local_dict, global_dict, _frame_depth=_frame_depth)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\hyung\AppData\Local\pypoetry\Cache\virtualenvs\langchainstudy-esJm8fcP-py3.11\Lib\site-packages\numexpr\necompiler.py", line 762, in getArguments
    a = global_dict[name]
        ~~~~~~~~~~~^^^^^^
KeyError: 'age'

During handling of the above exception, another exception occurred:

[...]

PR #14178, opened 2023-12-02
Fix: (issue #14177) OpenAI functions Agent official example Error

ggeutzzang:
- **Issue:** #14177


hwchase17 (changes requested):
mmmm how does this fix that error?

that error occurs because we are passing a string to the math function. i dont think this touches that at all?

PR #14201, opened 2023-12-03
Prefer byte store interface for Upstash BaseStore to match other Redis

jacoblee93:
If we are not going to make the existing Docstore class also implement `BaseStore[str, Document]`, IMO all base store implementations should always be `[str, bytes]` so that they are more interchangeable.

CC @rlancemartin @eyurtsev

### DIFF
--- a/libs/langchain/langchain/storage/upstash_redis.py
+++ b/libs/langchain/langchain/storage/upstash_redis.py
@@ -1,9 +1,10 @@
 from typing import Any, Iterator, List, Optional, Sequence, Tuple, cast
 
+from langchain_core._api.deprecation import deprecated
 from langchain_core.stores import BaseStore
 
 
-class UpstashRedisStore(BaseStore[str, str]):
+class _UpstashRedisStore(BaseStore[str, str]):
     """BaseStore implementation using Upstash Redis as the underlying store."""
 
     def __init__(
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/storage/upstash_redis.py
```diff
@@ -117,3 +118,57 @@ def yield_keys(self, *, prefix: Optional[str] = None) -> Iterator[str]:
...
+        namespace: Optional[str] = None,
+    ) -> None:
+        self.underlying_store = _UpstashRedisStore(
+            client=client, url=url, token=token, ttl=ttl, namespace=namespace
+        )
+
+    def mget(self, keys: Sequence[str]) -> List[Optional[bytes]]:
+        """Get the values associated with the given keys."""
+        return [
+            value.encode("utf-8") if value is not None else None
```
eyurtsev, line 155: I can't figure out how this ends up being a string. Redis works with bytes not strings... is upstash redis doing some additional encoding/decoding?

libs/langchain/langchain/storage/upstash_redis.py
```diff
@@ -117,3 +118,57 @@ def yield_keys(self, *, prefix: Optional[str] = None) -> Iterator[str]:
                     yield relative_key
                 else:
                     yield key
+
+
+@deprecated("0.0.335", alternative="UpstashRedisByteStore")
```
eyurtsev, line 123: one nit: The new code seems to involve overhead encoding/decodings, feels a bit strange that the new alternative is less optimized from that point of view?

eyurtsev (approved):
Where does the extra encoding/decoding step appear?

libs/langchain/langchain/storage/upstash_redis.py
jacoblee93: Would you want to keep both then? IMO it's worth it to simplify/standardize how devs use it and I doubt anybody is at the moment since you have to write your own encoding and we don't document it

libs/langchain/langchain/storage/upstash_redis.py
jacoblee93: I guess so? The underlying implementation returns a string 🤷

PR #14202, opened 2023-12-03
Allow base_store to be used directly with MultiVectorRetriever

jacoblee93:
Allow users to pass a generic `BaseStore[str, bytes]` to MultiVectorRetriever, removing the need to use the `create_kv_docstore` method. This encoding will now happen internally.

@rlancemartin @eyurtsev


libs/langchain/langchain/retrievers/multi_vector.py
```diff
@@ -27,12 +28,40 @@ class MultiVectorRetriever(BaseRetriever):
...
     """Type of search to perform (similarity / mmr)"""
 
+    def __init__(
+        self,
+        *,
+        vectorstore: VectorStore,
+        docstore: Optional[BaseStore[str, Document]] = None,
+        base_store: Optional[BaseStore[str, bytes]] = None,
+        id_key: str = "doc_id",
+        search_kwargs: dict = {},
```
eyurtsev, line 44: Use None as the default as dict is mutable

```suggestion
        search_kwargs: Optional[dict] = None,
```

Then downstream

```
search_kwargs = search_kwargs if search_kwargs is not None else {}
```

libs/langchain/langchain/retrievers/multi_vector.py
```diff
@@ -27,12 +28,40 @@ class MultiVectorRetriever(BaseRetriever):
...
+        search_type: SearchType = SearchType.similarity,
+    ):
+        if base_store is not None:
+            docstore = create_kv_docstore(base_store)
+        elif docstore is not None:
+            warnings.warn(
```
eyurtsev, line 50: Let's not deprecate this. We don't know how users created the implementation (it may not have been through create_kv_docstore) -- i.e., could be through inheritance of basestore

libs/langchain/langchain/retrievers/multi_vector.py
```diff
@@ -27,12 +28,40 @@ class MultiVectorRetriever(BaseRetriever):
...
-    id_key: str = "doc_id"
-    search_kwargs: dict = Field(default_factory=dict)
+    id_key: str
+    search_kwargs: dict
     """Keyword arguments to pass to the search function."""
-    search_type: SearchType = SearchType.similarity
+    search_type: SearchType
     """Type of search to perform (similarity / mmr)"""
 
+    def __init__(
```
eyurtsev, line 37: This is a pydantic object. So you use a pre-validator instead of using the __init__. The init might be a bit easier though since it contains `base_store` parameter

libs/langchain/langchain/retrievers/multi_vector.py
```diff
@@ -27,12 +28,40 @@ class MultiVectorRetriever(BaseRetriever):
...
+        *,
+        vectorstore: VectorStore,
+        docstore: Optional[BaseStore[str, Document]] = None,
+        base_store: Optional[BaseStore[str, bytes]] = None,
+        id_key: str = "doc_id",
+        search_kwargs: dict = {},
+        search_type: SearchType = SearchType.similarity,
+    ):
+        if base_store is not None:
+            docstore = create_kv_docstore(base_store)
```
eyurtsev, line 48: I wonder whether we should update the implementation in `create_kv_docstore` to inspect the type signature associated with base_store and give a warning to users if they pass `BaseStore[str, str]` etc.

jacoblee93: Can we do that? I tried and couldn't figure out how to differentiate - `isinstance` didn't work

libs/langchain/langchain/retrievers/multi_vector.py
jacoblee93: We haven't documented at all so I'd be surprised if anyone figured it out but yeah I can remove

libs/langchain/langchain/retrievers/multi_vector.py
jacoblee93: Ok, will try to switch it

PR #14205, opened 2023-12-04
Adds "NIN" metadata filter for pgvector to all checking for set absence

jeffpezzone:
This PR adds support for metadata filters of the form:

`{"filter": {"key": { "NIN" : ["list", "of", "values"]}}}`

"IN" is already supported, so this is a quick & related update to add "NIN"

### DIFF
--- a/libs/langchain/langchain/vectorstores/pgvector.py
+++ b/libs/langchain/langchain/vectorstores/pgvector.py
@@ -434,16 +434,22 @@ class PGVector(VectorStore):
 
             if filter is not None:
                 filter_clauses = []
+                IN, NIN = "in", "nin"
                 for key, value in filter.items():
-                    IN = "in"
-                    if isinstance(value, dict) and IN in map(str.lower, value):
+                    if isinstance(value, dict):
                         value_case_insensitive = {
                             k.lower(): v for k, v in value.items()
                         }
-                        filter_by_metadata = self.EmbeddingStore.cmetadata[
-                            key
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/vectorstores/pgvector.py
```diff
@@ -434,16 +434,22 @@ def __query_collection(
...
-                        ].astext.in_(value_case_insensitive[IN])
-                        filter_clauses.append(filter_by_metadata)
+                        if IN in map(str.lower, value):
+                            filter_by_metadata = self.EmbeddingStore.cmetadata[
+                                key
+                            ].astext.in_(value_case_insensitive[IN])
+                        elif NIN in map(str.lower, value):
+                            filter_by_metadata = self.EmbeddingStore.cmetadata[
+                                key
+                            ].astext.not_in(value_case_insensitive[NIN])
```
hwchase17, line 450: i think you need `else: filter_by_metadata = None`? otherwise it doesnt get set as None?

jeffpezzone: I was getting an exception previously when I simply had "if filter_by_metadata:". I think filter_by_metadata should be None if it doesn't hit either one of those conditionals. But I can pop in that else statement today as it doesn't hurt to have it

jeffpezzone: @hwchase17 done :) thanks for the comment dude!

PR #14206, opened 2023-12-04
Modify LLMs/Anyscale work with OpenAI API v1

kylehh:
- **Description:** 
1. Modify LLMs/Anyscale to work with OAI v1
2. Get rid of openai_ prefixed variables in Chat_model/ChatAnyscale
3. Modify `anyscale_api_base` to `anyscale_base_url` to follow OAI name convention (reverted)


libs/langchain/langchain/chat_models/anyscale.py
```diff
@@ -99,26 +99,21 @@ def get_available_models(
...
             )
         )
-        values["openai_api_base"] = get_from_dict_or_env(
+        values["anyscale_base_url"] = get_from_dict_or_env(
             values,
-            "anyscale_api_base",
-            "ANYSCALE_API_BASE",
-            default=DEFAULT_API_BASE,
+            "anyscale_base_url",
+            "ANYSCALE_BASE_URL",
```
efriis, line 115: could you revert this name change? This will be a breaking change to anyone currently using this class.

libs/langchain/langchain/chat_models/anyscale.py
```diff
@@ -137,8 +132,8 @@ def validate_environment_override(cls, values: dict) -> dict:
         try:
             if is_openai_v1():
                 client_params = {
-                    "api_key": values["openai_api_key"],
-                    "base_url": values["openai_api_base"],
+                    "api_key": values["anyscale_api_key"].get_secret_value(),
```
efriis, line 135: Looks like this might correspond with some changes in `Anyscale` that won't be available here in `ChatAnyscale`

efriis (commented):
Hey @kylehh ! I don't think this will work as-is. Could you look through the CI failures as well as some of the comments to fix?

libs/langchain/langchain/chat_models/anyscale.py
kylehh: agreed. revert it back

libs/langchain/langchain/chat_models/anyscale.py
kylehh: The point is to remove `openai_` prefixed variables in `anyscale.py`. 
And `anyscale_api_key` is SecreStr so we need to call `.get_secret_values()` explicitly to get the value.

kylehh: The previous code is hacky ( line 105-109). `openai_api_key` are still simple strings and used later in the code. `anyscale_api_key` is secretstr but not used. 
Now the new version fixed it

efriis (commented):
Does this work for you locally? I get `invalid model id` for any of the models, I believe from the OpenAI sdk, so this might not achieve what it intends.

libs/langchain/langchain/chat_models/anyscale.py
```diff
@@ -161,20 +156,21 @@ def validate_environment_override(cls, values: dict) -> dict:
...
-        available_models = cls.get_available_models(
-            values["openai_api_key"],
-            values["openai_api_base"],
-        )
-
-        if model_name not in available_models:
-            raise ValueError(
-                f"Model name {model_name} not found in available models: "
-                f"{available_models}.",
+        if model_name != "test":
```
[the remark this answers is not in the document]
kylehh: Agree. Checking model available here would need api_key and can't pass the unit test ( that's why I added the test condition). Will fix it

efriis:
Using Anyscale or ChatAnyscale in this PR with OpenAI v1

efriis:
to reproduce:
```
from langchain.chat_models import ChatAnyscale
ChatAnyscale(anyscale_api_key='xx', model_id='meta-llama/Llama-2-70b-chat-hf').invoke('hello there')
```

I get different errors on OpenAI v0 and v1, but neither of them seem to work:
on `openai==0.28.1` I get `openai.error.InvalidRequestError: invalid model ID` 

on `openai==1.3.7` I get `Completions.create() got an unexpected keyword argument 'model_id'`.

efriis:
Appreciate it! Retrying

efriis:
This fixed with `openai==1.3.7`, but I still get the same invalid model id error message on `openai==0.28.1`. Tried it with the same mistral model as your screenshot as well.

kylehh:
Added fix for it. baiscally we need pass values to `open_api_` from `anyscale_api_` for OAI<v1 case

efriis:
sounds good. This won't go in until next week because of some ongoing refactoring in these folders. Will follow up next week, and may require some small changes to address merge conflicts.

efriis:
@kylehh sorry for the late pin! That refactor is in, so if you want to move this change over to community, that would be great!

kylehh:
@efriis Code refactored and please review Thanks!

jeffxtang:
@efriis @kylehh is it possible to get this PR merged this week? Would love to try this in an official LC release. Thanks!

libs/community/langchain_community/llms/anyscale.py
```diff
@@ -132,10 +165,306 @@ def _identifying_params(self) -> Mapping[str, Any]:
...
+        return self.create_llm_result(
+            choices,
+            prompts,
+            params,
+            token_usage,
+            system_fingerprint=system_fingerprint,
+        )
+
+
+class AnyscaleChat(BaseOpenAI):
```
efriis, line 326: Let's skip this - assuming this is mimicking the deprecated OpenAIChat implementation.

libs/core/langchain_core/load/mapping.py
```diff
@@ -220,6 +220,12 @@
         "base",
         "RunnableParallel",
     ),
+    ("langchain", "chat_models", "openai", "ChatAnyscale"): (
```
efriis, line 223: fix

libs/langchain/langchain/__init__.py
```diff
@@ -134,6 +134,12 @@ def __getattr__(name: str) -> Any:
         _warn_on_import(name, replacement="langchain.llms.Anthropic")
 
         return Anthropic
+    elif name == "Anyscale":
```
efriis, line 137: remove

libs/community/langchain_community/chat_models/anyscale.py
```diff
@@ -43,7 +42,7 @@ class ChatAnyscale(ChatOpenAI):
     Example:
         .. code-block:: python
 
-            from langchain_community.chat_models import ChatAnyscale
+            from langchain.chat_models import ChatAnyscale
```
efriis, line 45: revert

libs/community/langchain_community/chat_models/anyscale.py
```diff
@@ -56,11 +55,7 @@ def _llm_type(self) -> str:
     def lc_secrets(self) -> Dict[str, str]:
         return {"anyscale_api_key": "ANYSCALE_API_KEY"}
 
-    @classmethod
```
efriis, line 59: revert

libs/community/langchain_community/chat_models/anyscale.py
```diff
@@ -102,22 +97,17 @@ def get_available_models(
 
         return {model["id"] for model in models_response.json()["data"]}
 
-    @root_validator(pre=True)
-    def validate_environment_override(cls, values: dict) -> dict:
+    @root_validator()
+    def validate_environment(cls, values: dict) -> dict:
         """Validate that api key and python package exists in environment."""
-        values["openai_api_key"] = get_from_dict_or_env(
```
efriis, line 108: confirm not setting this doesn't throw an error

libs/community/langchain_community/chat_models/anyscale.py
```diff
@@ -102,22 +101,17 @@ def get_available_models(
 
         return {model["id"] for model in models_response.json()["data"]}
 
-    @root_validator(pre=True)
-    def validate_environment_override(cls, values: dict) -> dict:
+    @root_validator()
+    def validate_environment(cls, values: dict) -> dict:
```
efriis, line 105: why was this necessary?

libs/community/langchain_community/chat_models/anyscale.py
```diff
@@ -164,20 +160,21 @@ def validate_environment_override(cls, values: dict) -> dict:
...
-        available_models = cls.get_available_models(
-            values["openai_api_key"],
-            values["openai_api_base"],
-        )
-
-        if model_name not in available_models:
-            raise ValueError(
-                f"Model name {model_name} not found in available models: "
-                f"{available_models}.",
+        if model_name != "test":
```
efriis, line 163: is this conditional something we want to keep?

libs/community/langchain_community/llms/anyscale.py
```diff
@@ -84,30 +89,58 @@ def send_query(llm, text):
...
     prefix_messages: List = Field(default_factory=list)
 
-    @classmethod
-    def is_lc_serializable(cls) -> bool:
-        return False
+    def __new__(cls, **data: Any) -> BaseOpenAI:  # type: ignore
+        """Initialize the OpenAI object."""
+        model_name = data.get("model_name", "")
+        if model_name not in COMPLETION_MODELS:
+            return AnyscaleChat(**data)
```
efriis, line 102: what is this case for? Should we just throw an error and tell users to use `ChatAnyscale` here instead?

libs/community/langchain_community/chat_models/anyscale.py
efriis: (the change in the validator type)

efriis:
Hey @kylehh ! I changed a lot, so I'd love your take on some of the unresolved comments.

Overall, I like the changes related to the openai client, but there were a bunch of others that I don't think we want. 

Let me know what you think.

libs/community/langchain_community/chat_models/anyscale.py
kylehh: I recalled that we needed this change, otherwise it would complain no `openai_api_token` environment set

libs/community/langchain_community/chat_models/anyscale.py
kylehh: added this condition for unit test. But we can remove

libs/community/langchain_community/llms/anyscale.py
kylehh: Hi Erick, just got back from vacation. Now I have a second thought on it. Yes. I think i should make this client simpler by only accept completion models, and remind ppl to use chat_model client for chat models.
Let's rework and test out your changes as well

libs/community/langchain_community/chat_models/anyscale.py
kylehh: Just tested and confirmed. with pre=True, you will get following validation error

 "Did not find openai_api_key, please add an environment variable `OPENAI_API_KEY` which contains it, or pass `openai_api_key` as a named parameter. (type=value_error)"

efriis: My intuition is that this will still fail if you don't have `OPENAI_API_KEY` in your environment then. I think this is the reason this implementation used to store the api key in `openai_api_key` instead of `anyscale_api_key`. It may be worth it to leave that as-is.

libs/community/langchain_community/chat_models/anyscale.py
```diff
@@ -152,6 +146,8 @@ def validate_environment_override(cls, values: dict) -> dict:
                 }
                 values["client"] = openai.OpenAI(**client_params).chat.completions
             else:
+                values["openai_api_base"] = values["anyscale_api_base"]
+                values["openai_api_key"] = values["anyscale_api_key"].get_secret_value()
```
efriis, line 150: this is changing to a SecretStr as well soon too

libs/community/langchain_community/chat_models/anyscale.py
kylehh: User should NOT use OPENAI_API_KEY environment var when using Anyscale, but only need to supply ANYSCALE_API_KEY. This change is for this purpose. Otherwise code would fail even if you have ANYSCALE_API_KEY but missing OPENAI_API_KEY

libs/community/langchain_community/chat_models/anyscale.py
kylehh: Sorry, I don't understand this comment. In `openai.py`, the `values["openai_api_key"]` is assigned as string as well

kylehh:
Hi @efriis Can this PR be merged?

libs/community/langchain_community/chat_models/anyscale.py
kylehh: removed

efriis (approved):

Issue #13939, opened 2023-11-28
To support open v1 SDK proxy setting


shadowlinyf:
### System Info

langchain 0.0.340

### Who can help?

@hwchase17 @agola11 

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [X] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

After openai v1 sdk the previous method openai.proxy = {xxx} is no longer supported. But langchain is still using it.
It seems openai v1 sdk only support to set proxy on client level. see this issue on open ai sdk repo https://github.com/openai/openai-python/issues/825#issuecomment-1826047567
 

### Expected behavior

Should implement the way openai sdk required to set proxy

PR #14212, opened 2023-12-04
Support `openai_proxy` compatible with openai>1.0.0

169:
### **Description:** 

`openai` v1 no longer supports the method of setting `openai.proxy`. If using `OPENAI_PRPXY`, the user needs to initialize an http_client.

 ### **Issue:**

https://github.com/langchain-ai/langchain/issues/13939


169:
@baskaryan, @eyurtsev, @hwchase17 PTAL, some countries can access OPENAI API only by using a proxy. Now this function is invalid.

169:
@baskaryan, @eyurtsev, @hwchase17 I have changed it, put the code into `langchain_community`, PTAL

Issue #14213, opened 2023-12-04
BiliBiliLoader Credential No Sessdata error


Jiaaming:
### System Info

Mac os
Python3.9

### Who can help?

@Jiaaming
I'm working on this issue.

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

when you run the BiliBiliLoader example from the official [docs](https://python.langchain.com/docs/integrations/document_loaders/bilibili)
```
loader = BiliBiliLoader(
        [
            "https://www.bilibili.com/video/BV1g84y1R7oE/",
        ]
    )
docs = loader.load()
print(docs)
```
will get
```
bilibili_api.exceptions.CredentialNoSessdataException.CredentialNoSessdataException: Credential 类未提供 sessdata 或者为空。

Process finished with exit code 1
```

The is because the original [bilibili_api](https://nemo2011.github.io/bilibili-api/#/get-credential) require a Credential to fetch the info from the video

### Expected behavior

Should return a Document object
```
[Document(page_content="Video Title:...,description:....)]
```

Jiaaming:
Already open a pull request [here](https://github.com/langchain-ai/langchain/pull/14214)

Shengguang-Zhou:
It's still not been solved

fenghuoliansanyue:
Modify the following file:
\Lib\site-packages\langchain_community\document_loaders\bilibili.py

Add the following in the call library function area:
```
from bilibili_api import Credential
sessdata = ""
bili_jct = ""
buvid3 = ""
credential = Credential(sessdata=sessdata, bili_jct=bili_jct, buvid3=buvid3)
```
See this link below for more information:
https://geekdaxue.co/read/bilibili-api-docs/docs-get-credential.md
then,find the following code:
`v = video.Video(bvid=bvid.group())`
change to:
`v = video.Video(bvid=bvid.group(), credential=credential)`
then,it will be ok!
**Do not disclose the above information to others!!!**

PR #14214, opened 2023-12-04
Fix issue(#14213): BiliBiliLoader Credential No Sessdata error 

Jiaaming:
- **Description:** Fix the BiliBiliLoader that can receive cookie parameters; Add example in docs
  - **Issue:** [#14213]

### DIFF
--- a/libs/partners/anthropic/README.md
+++ b/libs/partners/anthropic/README.md
@@ -1,1 +1,34 @@
 # langchain-anthropic
+
+This package contains the LangChain integration for Anthropic's generative models.
+
+## Installation
+
+`pip install -U langchain-anthropic`
+
+## Chat Models
+
+| API Model Name     | Model Family   |
+| ------------------ | -------------- |
+| claude-instant-1.2 | Claude Instant |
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/document_loaders/bilibili.py
```diff
@@ -4,80 +4,102 @@
...
 class BiliBiliLoader(BaseLoader):
-    """Load `BiliBili` video transcripts."""
+    """
+    Loader for fetching transcripts from BiliBili videos.
+    """
 
-    def __init__(self, video_urls: List[str]):
-        """Initialize with bilibili url.
+    def __init__(
+        self, video_urls: List[str], sessdata: str, bili_jct: str, buvid3: str
```
hwchase17, line 23: is there a way to do this in a backwards comptabile way so this doesnt break? it seems like if this worked before, these should be optional anwyays?

Jiaaming: Thank you for the advice. I change it and it passed all the test. Should be worked now.

PR #9850, opened 2023-08-28
Virtual column name for self-querying retrievers

mpskex:
Self query retrievers are powerful and it can be even stronger than before. Taking MyScale as an example, sometimes the user may add complex SQL function to a column name. So we add a `VirtualColumnName` for self-query retriever to reduce token usage with extra functionality.

1. a function that dynamically creates function mapped column names, which becomes handy when the user want to compare between current time / location or other attributes along with a column.
2. complex function call in plain string can now be replaced with a shorter nickname in prompt, which saves number of token for very long function names.
3. refined the logic in MyScale translator to meet this standard.

This change will not affect other self query retrievers as it preserves the original plain string interface under comparison. The actual difference happens under `QueryTransformer` in `query_construct.parser` for `Lark` parser.

`Comparator.attribute` can now be either a string or a `VirtualColumnName`. I have defined the default behaviour which stops the user to use `VirtualColumnName` under other vectorstores than MyScale.

This functionality will boost the self-query retrievers as a bridge between text and simple SQL query. And we believe this will help users to expand their usage to this retriever. Other SQL vector database is theoretically compatible to this new feature as well.

Hope you find this interesting.
@hwchase17  @baskaryan 

Our contact on twitter: @MyScaleDB @mpsk_liu

PR #14217, opened 2023-12-04
[Revised] Virtual column name for self-querying retrievers

mpskex:
Self query retrievers are powerful and it can be even stronger than before. Taking MyScale as an example, sometimes the user may add complex SQL function to a column name. So we add a `VirtualColumnName` for self-query retriever to reduce token usage with extra functionality.

1. a function that dynamically creates function mapped column names, which becomes handy when the user want to compare between current time / location or other attributes along with a column.
2. complex function call in plain string can now be replaced with a shorter nickname in prompt, which saves number of token for very long function names.
3. refined the logic in MyScale translator to meet this standard.

This change will not affect other self query retrievers as it preserves the original plain string interface under comparison. ~~The actual difference happens under `QueryTransformer` in `query_construct.parser` for `Lark` parser.~~

`Comparator.attribute` can now be either a string or a `VirtualColumnName`. ~~I have defined the default behaviour which stops the user to use `VirtualColumnName` under other vectorstores than MyScale.~~ Parser for `VirtualColumnName`s are chained with original structured query parser, using LCEL.

This functionality will boost the self-query retrievers as a bridge between text and simple SQL query. And we believe this will help users to expand their usage to this retriever. Other SQL vector database is theoretically compatible to this new feature as well.

Hope you find this interesting.
@hwchase17  @baskaryan @efriis 

Our contact on twitter: @MyScaleDB @mpsk_liu

=========

Updates:

* Separate from `StructuredQueryOutputParser` and created new `VirtualColumnParser`
* Rewritten with LCEL
* Won't hurt any inheritance / API signatures


efriis (commented):
Hey @mpskex - same question as in #9850 - is there a way to implement this in a way that doesn't change any of the general query construction business logic? Would be better if this format of a VirtualColumnName only existed in `retrievers/self_query/myscale.py`.

mpskex:
> Hey @mpskex - same question as in #9850 - is there a way to implement this in a way that doesn't change any of the general query construction business logic? Would be better if this format of a VirtualColumnName only existed in `retrievers/self_query/myscale.py`.

Sorry I just don't get it quite much. There are few questions I would like to seek for help in order to revise according your advice. 

1. Do you have any idea how to separate the `AttributeInfo`, where the user provides attributes with `VirtualColumnName`, from the output parser? Actually the self querying retrievers takes structured query parser as an attribute so there is no way to by pass this virtual attribute to the upper class. 
2. Do you have any idea how to migrate this user defined attribute prompt mapping into a translator. We are all familiar with the relationship between translator and the query parser. They are separated and the only connection is IR. Do you mean I have to change the IR in order to create such a "vendor specific" logic?

This PR is creating a middleware that optimize / maps prompts for columns with the real column name, by implementing a parser to translate some prompted column name if needed. I simply didn't get why we refusing this PR. Are there any concern on the changes we made here? I would be very happy to change if I do know a reason for this.

Thanks

mpskex:
Again, please let me clarify here. I recognize this feature as a common feature, instead of a business specific one, especially to SQL vector databases. PGVector and others can also make use of it. Also, it could be a positive thing for Pinecone and Chroma, if you ask me. I think it's a good way to provide our customer with more possibilities don't you think? 

We treat LangChain as a great interface where we push AI towards customers. That's why we have so many tools and frameworks here in LangChain. They just helped our customers to get a fast implementation to their business. Correct me if I am wrong.

Issue #14120, opened 2023-12-01
Expose Amazon Bedrock amazon-bedrock-invocationMetrics in response


ben-gineer:
### Feature request

Amazon Bedrock now provides some useful metrics when it has finished:

`"completion":"Response from Bedrock...","stop_reason":"stop_sequence","stop":"\n\nHuman:","amazon-bedrock-invocationMetrics":{"inputTokenCount":16,"outputTokenCount":337,"invocationLatency":4151,"firstByteLatency":136}}`

It would be good to expose these via LangChain, similar to how OpenAI API provides them.

### Motivation

To make it easier to calculate the cost of Amazon Bedrock queries.

### Your contribution

I'm happy to test any solution. I'd look to existing developers to guide us how best to expose them.

shabie:
Did you have something like this in mind https://python.langchain.com/docs/modules/model_io/llms/token_usage_tracking?

ben-gineer:
Yep. Something like that.

thiagotps:
This issue is important. Langchain api should not thrown out useful bedrock data.

PR #14219, opened 2023-12-04
add bedrock callback for token counting

shabie:
- **Description:** Bedrock API now returns [token usage](https://docs.aws.amazon.com/bedrock/latest/userguide/monitoring-cw.html) details across all hosted models which can be used to collect statistics on token usage. This can be used to implement a similar [token usage callback](https://python.langchain.com/docs/modules/model_io/llms/token_usage_tracking) as OpenAI.
  - **Issue:** #14120
  - **Dependencies:** None
  - **Tag maintainer:** @3coins

Token counting (as for OpenAI's callback) only works for the non-stream variant currently and hence the warning has been included.


shabie:
@eyurtsev @3coins could you kindly review?

docs/docs/modules/model_io/llms/token_usage_tracking.ipynb
```diff
@@ -158,13 +168,95 @@
...
+   "source": [
+    "with get_bedrock_token_count_callback() as cb:\n",
+    "    text = \"\\n\\nHuman: What is the square root of 4?\\n\\nAssistant: \"\n",
+    "    answer = llm(text)\n",
+    "    print(answer)  # check if invocations return same response\n",
+    "    answer = llm(text)\n",
+    "    print(answer)  # check if invocations return same response\n",
+    "\n",
+    "    total_tokens = cb.total_tokens\n",
+    "    print(cb.total_tokens, cb.prompt_tokens, cb.completion_tokens)\n",
```
3coins, line 250: Currently, this is just printing numbers which doesn't have any info on the counts. Updating this to show the counts and their purpose would be useful.
```suggestion
    "    print(cb)\n",
```

libs/langchain/langchain/llms/bedrock.py
```diff
@@ -229,6 +236,16 @@ def validate_environment(cls, values: Dict) -> Dict:
...
         return values
 
+    @validator("streaming")
+    def check_flag(cls, value):
+        if value:
+            warnings.warn(
+                "The 'streaming' attribute is set to True. Token usage statistics will "
+                "not be available.",
+                BedrockStreamingWarning,
+            )
```
3coins, line 246: I am not sure if a warning is warranted here, since there are going to be a lot of users who will be using streaming, and this will also not be an error from the user to set this flag. If we really want to do something, it might be useful to just use the logger and log this info. It might also be useful to just note this in the docs for the callback.

libs/langchain/langchain/llms/bedrock.py
```diff
@@ -441,3 +473,7 @@ def get_token_ids(self, text: str) -> List[int]:
             return get_token_ids_anthropic(text)
         else:
             return super().get_token_ids(text)
+
+
+class BedrockStreamingWarning(Warning):
+    pass
```
3coins, line 479: See my comment above, this might not be needed.

libs/langchain/tests/unit_tests/test_bedrock_info.py
```diff
@@ -0,0 +1,31 @@
+import pytest
```
3coins, line 1: Thanks for adding a unit test. 👍

3coins (commented):
@shabie
Thanks for making these changes. Left some minor comments; looks good otherwise. You might also want to rebase on the master, and fix the lint errors.

libs/langchain/langchain/llms/bedrock.py
ben-gineer: Is there a reason we can't expose this via streaming? The response metrics are included on the end token.

baskaryan:
apologies for delay @shabie, looks like there's some merge conflicts

ben-gineer:
@shabie - did you see my comment about adding support to the streaming version too? Bedrock provides the metrics on both (in response metadata on non-streaming, and as metadata on the final stop action when streaming).

shabie:
> @shabie - did you see my comment about adding support to the streaming version too? Bedrock provides the metrics on both (in response metadata on non-streaming, and as metadata on the final stop action when streaming).

Yeah. I wanted to do it in two steps. First for batch then for streaming. Cuz otherwise the PR would get probably even bigger. I'll have a second look while I fix these merge conflicts.

shabie:
I think this can be reviewed. If I am missing something please let me know. It has been a while @3coins @baskaryan @ben-gineer

ccurme:
There's a callback for Bedrock implemented: https://python.langchain.com/v0.2/docs/how_to/chat_token_usage_tracking/#bedrock-anthropic

Apologies as I believe this PR was earlier.

Bedrock models also at this point return token usage via the `.usage_metadata` attribute on `AIMessage` (including in streaming).

Further improvements should be submitted to the `langchain-aws` repo: https://github.com/langchain-ai/langchain-aws

PR #14221, opened 2023-12-04
Feature: GitLab url from ENV

balaba-max:


hwchase17 (approved):

libs/langchain/langchain/utilities/gitlab.py
```diff
@@ -38,6 +38,10 @@ class Config:
     @root_validator()
     def validate_environment(cls, values: Dict) -> Dict:
         """Validate that api key and python package exists in environment."""
+
+        gitlab_url = get_from_dict_or_env(
+            values, "gitlab_url", "GITLAB_URL", default="https://gitlab.com"
```
efriis, line 43: default to None and let sdk populate default

PR #14222, opened 2023-12-04
feat: Increased compatibility with new and old versions for dalle

majiayu000:
- **Description:** Increased compatibility with all versions openai for dalle, 

This pr add support for openai version from 0 ~ 1.3.

### DIFF
--- a/libs/langchain/langchain/utilities/dalle_image_generator.py
+++ b/libs/langchain/langchain/utilities/dalle_image_generator.py
@@ -1,10 +1,16 @@
 """Utility that calls OpenAI's Dall-E Image Generator."""
-from typing import Any, Dict, Optional
+import os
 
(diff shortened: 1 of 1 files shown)


hwchase17 (commented):
we probably dont want to make this Serializable = True yet (we are being pretty selective in how we do that)

Could we remove that from this PR? I like the idea of making it more OpenAI compatible

majiayu000:
> we probably dont want to make this Serializable = True yet (we are being pretty selective in how we do that)
> 
> Could we remove that from this PR? I like the idea of making it more OpenAI compatible


Thanks for the feedback. I delete the serializable function, make it like other tools.

hwchase17 (commented):
we can also delete lc_secrets and lc_attributes

majiayu000:
we finally delete all lc functions.

hwchase17 (approved):
thanks!

PR #14230, opened 2023-12-04
Fix `config` arg detection for wrapped lambdarunnable

Vinno97:
**Description:**
When a RunnableLambda only receives a synchronous callback, this callback is wrapped into an async one since #13408. However, this wrapping with `(*args, **kwargs)` causes the `accepts_config` check at [/libs/core/langchain_core/runnables/config.py#L342](https://github.com/langchain-ai/langchain/blob/ee94ef55ee6ab064da08340817955f821dfa6261/libs/core/langchain_core/runnables/config.py#L342) to fail, as this checks for the presence of a "config" argument in the method signature. 

Adding a `functools.wraps` around it, resolves it.

**Issue:** 
None, writing this PR already took longer than the fix ;)

**Tag maintainer:** 
@nbondoux @efriis (participants of the original PR)

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -6,7 +6,7 @@ import threading
 from abc import ABC, abstractmethod
 from concurrent.futures import FIRST_COMPLETED, wait
 from copy import deepcopy
-from functools import partial
+from functools import partial, wraps
 from itertools import tee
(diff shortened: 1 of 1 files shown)


efriis (approved):
Great catch @Vinno97 ! Let me chat with the team if we want to fix this here, or do something different with `accepts_config` before merging.

PR #14236, opened 2023-12-04
Fixed OpenAIFunctionsAgent not returning when receiving AgentFinish

lh0x00:
**Description:** The way the condition is checked in the `return_stopped_response` function of `OpenAIAgent` may not be correct, when the value returned is `AgentFinish` from the tools it does not work properly.

Thanks for review, @baskaryan, @eyurtsev, @hwchase17.


hwchase17 (approved):
thanks

PR #14270, opened 2023-12-05
Handling inline references to .ipynb files

169:
### Description

I found that [the OpenAI Adapter document](https://python.langchain.com/docs/integrations/adapters/openai) cannot correctly jump to other internal links. Clicking on the link will download it as an ipynb file.

If it works normally on both jupyter notebook and doc websites, additional processing needs to be done, that is, replacing the link before rendering.

I added 2 scripts and tested `make docs_build` and `./vercel_build.sh`. In both cases, the link can be generated correctly without affecting the original ipynb file content.

@efriis Please review this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -179,3 +179,4 @@ docs/docs/node_modules
 docs/docs/yarn.lock
 _dist
 docs/docs/templates
+docs/.inline_ipynb.json
--- a/docs/.local_build.sh
+++ b/docs/.local_build.sh
@@ -12,6 +12,7 @@ mkdir -p ../_dist
 rsync -ruv --exclude node_modules . ../_dist
 cd ../_dist
 poetry run python scripts/model_feat_table.py
+poetry run python scripts/replace_inline_ipynb_filename.py
 cp ../cookbook/README.md src/pages/cookbook.mdx
 cp ../.github/CONTRIBUTING.md docs/contributing.md
 mkdir -p docs/templates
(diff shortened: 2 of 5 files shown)


efriis:
Hey @169 ! Actually going to go with #14325 for now. I'm still iterating on our docs build in pursuit of making everything mdxv2 compatible for docusaurus 3, and the fewer custom scripts we're reliant on through that, the less likely the docs will break again :).

Great shout though! Found a few actually broken links in addition to the ipynb rewrite ones through exploring `ipynb` files for it.

If you find other broken (or downloading) links, please let me know!

169:
@efriis OK Before, I also considered adding quarto [configuration file](https://quarto.org/docs/projects/scripts.html#pre-and-post-render) to solve the problem, namely `pre-render` and `post-render`, but in fact they also use additional scripts, and it is not easy to determine the source when a problem occurs. It is better to add it directly to in a bash script.

169:
I did see another format problem, that is, the format displayed by docusaurus admonitions is incorrect. 

For [example](https://python.langchain.com/docs/expression_language/how_to/message_history#example-messages-input-dict-output):
Screenshot 2023-12-06 at 10 20 01 AM

In the tips, the markdown url is not displayed accurately, there are several such links, but they are all displayed incorrectly. but it seems to be a bug in docusaurus.

efriis:
Great catch! @bracesproul started work in #14349 it looks like

169:
@bracesproul @efriis They don't look as expected:
Screenshot 2023-12-07 at 8 13 04 AM

PR #14273, opened 2023-12-05
docs: Mendable AI Floating Button

nickscamara:
Hey yall,

As discussed, here is the Mendable AI chat triggered by a small floating button on the right of the screen. That way users can continue to get the value out of the assistant while you experiment. The implementation is very similar to what we have done with Snapchat, Llamaindex, AutoGPT.

Let me know if you have any questions and feedback, happy to iterate! If this looks good, we will PR in the langchain js docs as well.

Thank you!


efriis:
Responded in slack!

Issue #12250, opened 2023-10-25
Vertex AI Models Throws AttributeError: After Langchain 0.0295 Version


rsandeesh:
### System Info

1

I have been working with LangChain for some time, and after updating my LangChain module to version 0.0.295 or higher, my Vertex AI models throw an error in my application. The error is

  File "E:\Office\2023\AI-Platform\kaya-aia-general-bot\venv\Lib\site-packages\langchain\llms\vertexai.py", line 301, in _generate
    generations.append([_response_to_generation(r) for r in res.candidates])
                                                            ^^^^^^^^^^^^^^
AttributeError: 'TextGenerationResponse' object has no attribute 'candidates'

I was expecting my previous implementation to work seamlessly. This is how I initialized the chain with the vertex model.

case 'public_general_palm':
llm = ChatVertexAI(model_name="chat-bison", temperature=0)
chain = self.__initialize_basic_chain(llm, self.chain_memory)
return chain(request.msg, return_only_outputs=True)
This is the __initialize_basic_chain method.

def __initialize_basic_chain(self, llm, memory):
   return ConversationChain(llm=llm, memory=memory)
However, I have been unable to find a solution for this error. Kindly assist in resolving this issue to ensure the smooth operation of the application.

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [X] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [X] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Update langchain module 0.315 and higher. 

1. Initialize llm with the ChatVertexAI(model_name="chat-bison", temperature=0). 
2. Pass into a basic chain or an agent with the input messages

### Expected behavior

Throws an internal module error called 

 File "E:\Office\2023\AI-Platform\kaya-aia-general-bot\venv\Lib\site-packages\langchain\llms\vertexai.py", line 301, in _generate
    generations.append([_response_to_generation(r) for r in res.candidates])
                                                            ^^^^^^^^^^^^^^
AttributeError: 'TextGenerationResponse' object has no attribute 'candidates'

PR #14290, opened 2023-12-05
fix error issue 12250

daniellefranca96:
Description: Correction of bug that make the predict method failed on VertexAI LLM.
Issue: #12250 
Dependencies: none.
Tag maintainer: @hwchase17


hwchase17 (commented):
mmm but we explicitly have candidate_count as something in _default_params. We should seemingly remove that as well right?

daniellefranca96:
@hwchase17 i removed the paramter n from VertexAI and ChatVertexAI

PR #14294, opened 2023-12-05
Update docs

169:
### Description

Fixed 3 doc  issues:

1. `ConfigurableField ` needs to be imported in `docs/docs/expression_language/how_to/configure.ipynb`
2. use `error` instead of `RateLimitError()` in `docs/docs/expression_language/how_to/fallbacks.ipynb`
3. I think it might be better to output the fixed json data(when I looked at this example, I didn't understand its purpose at first, but then I suddenly realized):
Screenshot 2023-12-05 at 10 34 13 PM


hwchase17 (approved):
thanks!

PR #14297, opened 2023-12-05
Supporting LLM Models over REST endpoints

daichi-m:
**Description** 

With multiple LLMs being developed by various organizations, and in-house LLMs developed for internal consumption, we at Walmart want to expose all these LLMs over simple REST endpoint. This enables each team to independently develop their LLMs and their API signatures - which can then be easily exposed to users via LangChain. This PR adds the support to serve LLMs over REST endpoints. 

For the initial PR, this is a simple implementation of Sync HTTP REST call. Based on approval and interests, we would be enhancing this further with support for Async HTTP and Streaming HTTP as well in future PRs.

**Dependencies**
This is built on top of `requests`. If `httpx` is available on the PYTHONPATH, then httpx is preferred over requests.

**Twitter Handles**
@daichi_majumdar

### DIFF
--- a/docs/docs/integrations/llms/llm_over_rest.ipynb
+++ b/docs/docs/integrations/llms/llm_over_rest.ipynb
@@ -1,0 +1,148 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "169eda47-d480-4311-9556-74e900e892a7",
+   "metadata": {},
+   "source": [
+    "# LLM Over REST\n",
+    "\n",
+    "This notebook explains how to use the LLM Over REST integration with LangChain. There is no additional dependencies that you would need to install. It uses `httpx` under the hood, if it is available, otherwise falling back to `requests`.\n",
+    "\n",
+    "\n",
+    "## Configuration Parameters\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


hwchase17:
how general is this? compared to being specific to how Walmart deploys LLMs internally?

My concern would be this makes some assumption about how you deploy LLMs that are not generalizable outside.

An alternative would be to maintain this as its own package, since it should just rely on langchain-core. this is actually how we will be moving most integrations in the future. if this is interesting, we're happy to help advise on this!

daichi-m:
@hwchase17 I would argue REST is in fact, more generalized of all because this same class can be used to call OpenAI endpoints or LLMs served through HuggingFace TGI or vLLM. 

The design of the solution does not assume how the model is deployed within Walmart but makes a conscious choice to support any LLM API that is deployed as a REST endpoint without any assumption of the inference server. And since this is built on top of `requests` - there would be no need for the user to install and manage any additional dependencies. 

With this change, we can allow users to use LangChain with LLMs deployed on a vanilla Flask server, or any other inference server that is not yet integrated with LangChain.

daichi-m:
@hwchase17 Do you think we can move forward with this?

ccurme:
It's unclear to me that this is sufficiently general. For instance we set
```python
json_body: Dict[str, Any] = {"prompt": prompt}
```
APIs would need to conform to this interface, right? And incorporating additional features may become unwieldy.

Closing for now, if there's interest we could move to `langchain-experimental`.

PR #14306, opened 2023-12-05
ci matrix

baskaryan:


.github/scripts/check_diff.py
```diff
@@ -0,0 +1,27 @@
...
+    "libs/experimental",
+    "libs/partners/openai",
+}
+
+if __name__ == "__main__":
+    files = sys.argv[1:]
+    dirs_to_run = set()
+
+    for file in files:
+        if any(dir_ in file for dir_ in (".github/workflows", ".github/tools", ".github/actions", "libs/core")):
```
efriis, line 17: might be better to make this a negative case just in case we introduce some new folder structure in the future, probably want to run too much rather than too little CI

.github/workflows/check_diffs.yml
```diff
@@ -0,0 +1,52 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - id: files
+        uses: Ana06/get-changed-files@v2.2.0
+      - run: |
+          for changed_file in ${{ steps.files.outputs.all }}; do
```
efriis, line 36: assume this was debug?

.github/workflows/check_diffs.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - id: files
+        uses: Ana06/get-changed-files@v2.2.0
```
efriis, line 34: Can also do this with a `git diff --name-only` based on the base.sha for PRs.

What does this do in the `workflow_dispatch` case where there isn't a reference PR?

.github/workflows/check_diffs.yml
```diff
@@ -0,0 +1,52 @@
...
+          done
+      - id: set-matrix
+        run: echo "dirs-to-run=$(python .github/scripts/check_diff.py ${{ steps.files.outputs.all }})" >> $GITHUB_OUTPUT
+    outputs:
+      dirs-to-run: ${{ steps.set-matrix.outputs.dirs-to-run }}
+  ci:
+    needs: [ build ]
+    strategy:
+      matrix:
+        working-directory: ${{ fromJson(needs.build.outputs.dirs-to-run) }}
```
efriis, line 47: this is rad

efriis (commented):
v cool

.github/workflows/check_diffs.yml
baskaryan: probably don't need to support workflow_dispatch tbh?

efriis: Works for me

.github/workflows/_all_ci.yml
```diff
@@ -1,21 +1,14 @@
...
-      - "libs/langchain/**"
-      - "libs/core/**"
-  workflow_dispatch: # Allows to trigger the workflow manually in GitHub UI
+  workflow_call:
+    inputs:
+      working-directory:
+        required: true
+        type: string
+        description: "From which folder this pipeline executes"
+
```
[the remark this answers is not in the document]
baskaryan: but then we've gotta update each time we add a package (goal is to have to do less of that)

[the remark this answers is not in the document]
efriis: Fair but feels useful to be able to "run CI on this lib on this branch" in a pinch. Maybe just `type: string` would be good and can fill it out?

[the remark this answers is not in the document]
baskaryan: oh oh didn't realize we were adding a workflow-dispatch option, thought you were changing the input type in workflow-call

.github/scripts/check_diff.py
```diff
@@ -0,0 +1,45 @@
...
+                ".github/scripts/check_diff.py",
+            )
+        ):
+            dirs_to_run = ALL_DIRS
+            break
+        elif "libs/community" in file:
+            dirs_to_run.update(
+                ("libs/community", "libs/langchain", "libs/experimental")
+            )
+        elif "libs/partners" in file:
```
hwchase17, line 31: why run langchain and experimental for these updates?

.github/scripts/check_diff.py
```diff
@@ -0,0 +1,45 @@
...
+            partner_dir = file.split("/")[2]
+            dirs_to_run.update(
+                (f"libs/partners/{partner_dir}", "libs/langchain", "libs/experimental")
+            )
+        elif "libs/langchain" in file:
+            dirs_to_run.update(("libs/langchain", "libs/experimental"))
+        elif "libs/experimental" in file:
+            dirs_to_run.add("libs/experimental")
+        elif file.startswith("libs/"):
```
hwchase17, line 40: what is this? wont this run everything on any changes in libs?

baskaryan: it's an elif, so only if we didn't match any other conditions. basically catches new libraries in case we forget to update this (better to run too much than too little CI when adding new stuff)

PR #14308, opened 2023-12-05
Allow not enforcing function usage when a single function is passed to openai function executable

karimassi:
- **Description:** allows not enforcing function usage when a single function is passed to an openAI function executable (or corresponding legacy chain). This is a desired feature in the case where the model does not have enough information to call a function, and needs to get back to the user. 
  - **Issue:** N/A
  - **Dependencies:** N/A
  - **Tag maintainer:** N/A


hwchase17 (approved):
thanks!

PR #14315, opened 2023-12-05
docs: `integrations/providers/` update

leo-gan:
- added missed provider files (from `integrations/Callbacks`
- updated notebooks: added links; updated into consistent formats


hwchase17 (approved):
thanks!

PR #14325, opened 2023-12-06
docs[patch]: fix ipynb links

efriis:
Keeping it simple for now.

Still iterating on our docs build in pursuit of making everything mdxv2 compatible for docusaurus 3, and the fewer custom scripts we're reliant on through that, the less likely the docs will break again.

Other things to consider in future:

Quarto rewriting in ipynbs: https://quarto.org/docs/extensions/nbfilter.html (but this won't do md/mdx files)

Docusaurus plugins for rewriting these paths

### DIFF
--- a/docs/docs/integrations/adapters/openai-old.ipynb
+++ b/docs/docs/integrations/adapters/openai-old.ipynb
@@ -7,7 +7,7 @@
    "source": [
     "# OpenAI Adapter(Old)\n",
     "\n",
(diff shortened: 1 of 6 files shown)


efriis:
Hey @169 ! Just playing with the different ways quarto compiles links while reviewing your PR. We have a mix of direct (ipynb) and web (working) links throughout our docs, and I'm considering what would work best!

PR #14334, opened 2023-12-06
Exclude `max_tokens` from request if it's None

harupy:
We found a request with `max_tokens=None` results in the following error in Anthropic:

```
HTTPError: 400 Client Error: Bad Request for url: https://oregon.staging.cloud.databricks.com/serving-endpoints/corey-anthropic/invocations. 
Response text: {"error_code":"INVALID_PARAMETER_VALUE","message":"INVALID_PARAMETER_VALUE: max_tokens was not of type Integer: null"}
```

This PR excludes `max_tokens` if it's None.

### DIFF
--- a/libs/langchain/langchain/llms/databricks.py
+++ b/libs/langchain/langchain/llms/databricks.py
@@ -334,7 +334,7 @@ class Databricks(LLM):
 
     @property
     def _llm_params(self) -> Dict[str, Any]:
-        params = {
+        params: Dict[str, Any] = {
             "temperature": self.temperature,
             "n": self.n,
         }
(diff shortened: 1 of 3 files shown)


libs/langchain/langchain/llms/databricks.py
```diff
@@ -434,6 +434,12 @@ def _default_params(self) -> Dict[str, Any]:
             # "transform_output_fn": self.transform_output_fn,
         }
 
+        if self.stop:
+            params["stop"] = self.stop
+        if self.max_tokens is not None:
+            params["max_tokens"] = self.max_tokens
+        return params
```
harupy, line 441:
```suggestion
```

hwchase17 (approved):
thanks, seems reasonable!

harupy:
Thanks!

PR #14351, opened 2023-12-06
Update RunnableWithMessageHistory

eyurtsev:
This PR updates RunnableWithMessage history to support user specific configuration for the factory.

It extends support to passing multiple named arguments into the factory if the
factory takes more than a single argument.


libs/core/langchain_core/runnables/history.py
```diff
@@ -28,8 +25,36 @@
...
 
+import inspect
+from typing import Callable, Dict, Union
+
 MessagesOrDictWithMessages = Union[Sequence["BaseMessage"], Dict[str, Any]]
-GetSessionHistoryCallable = Callable[..., BaseChatMessageHistory]
+GetSessionHistoryCallable = Union[Callable[..., BaseChatMessageHistory]]
+
+
+def check_callable_argument_type(callable_: GetSessionHistoryCallable) -> str:
```
eyurtsev, line 35: err this isn't needed

libs/core/langchain_core/runnables/history.py
```diff
@@ -28,8 +25,11 @@
     from langchain_core.runnables.config import RunnableConfig
     from langchain_core.tracers.schemas import Run
 
+import inspect
+from typing import Callable, Dict, Union
+
 MessagesOrDictWithMessages = Union[Sequence["BaseMessage"], Dict[str, Any]]
-GetSessionHistoryCallable = Callable[..., BaseChatMessageHistory]
+GetSessionHistoryCallable = Union[Callable[..., BaseChatMessageHistory]]
```
efriis, line 32: Does this do anything?

libs/core/langchain_core/runnables/history.py
```diff
@@ -135,6 +139,8 @@ def get_session_history(
                 as output.
             history_messages_key: Must be specified if the base runnable accepts a dict
                 as input and expects a separate key for historical messages.
+            custom_config_specs: Configure fields that should be passed to the chat
```
efriis, line 142: name mismatch

libs/core/langchain_core/runnables/history.py
```diff
@@ -96,6 +99,7 @@ def __init__(
         input_messages_key: Optional[str] = None,
         output_messages_key: Optional[str] = None,
         history_messages_key: Optional[str] = None,
+        session_history_config_specs: Optional[Sequence[ConfigurableFieldSpec]] = None,
```
efriis, line 102:
```suggestion
        history_factory_config: Optional[Sequence[ConfigurableFieldSpec]] = None,
```

efriis: Don't love either name - both a bit wordy and difficult to figure out what it is at first glance. Maybe `session_history_config`? Maybe history is redundant because it's chat history

efriis (commented):
Will look more in the office and why it needs to be a special type of config

libs/core/langchain_core/runnables/history.py
nfcampos: A union of a single element I think is actually invalid

eyurtsev:
Linter +10 Eugene +1

nfcampos (approved):

PR #14353, opened 2023-12-06
Update readthedocs_documentation.ipynb

Prish10:
Have added latest implementation of ReadTheDocsLoader that I tried out on Langchain docs. 
The important parameter here is "encoding". Many users face the issue of 'UnicodeDecodeError':'character maps to Undefined. So, they need to change encoding according to their html files to utilize full power of ReadTheDocsLoader.


docs/docs/integrations/document_loaders/readthedocs_documentation.ipynb
baskaryan, line 1: seems to be an error with the notebook

Issue #3970, opened 2023-05-02
load_qa_chain with map_rerank by local huggingface model 


flaviadeutsch:
I use the huggingface model locally and run the following code:

```
chain = load_qa_chain(llm=chatglm, chain_type="map_rerank", return_intermediate_steps=True, prompt=PROMPT)
chain({"input_documents": search_docs_Documents, "question": query}, return_only_outputs=True)
```
The error is as follows：

```
─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /tmp/ipykernel_274378/983731820.py:2 in <module>                                                 │
│                                                                                                  │
│ [Errno 2] No such file or directory: '/tmp/ipykernel_274378/983731820.py'                        │
│                                                                                                  │
│ /tmp/ipykernel_274378/14951549.py:11 in answer_docs                                              │
│                                                                                                  │
│ [Errno 2] No such file or directory: '/tmp/ipykernel_274378/14951549.py'                         │
│                                                                                                  │
│ /home/hysz/anaconda3/envs/chatglm/lib/python3.10/site-packages/langchain/chains/base.py:116 in   │
│ __call__                                                                                         │
│                                                                                                  │
│   113 │   │   │   outputs = self._call(inputs)                                                   │
│   114 │   │   except (KeyboardInterrupt, Exception) as e:                                        │
│   115 │   │   │   self.callback_manager.on_chain_error(e, verbose=self.verbose)                  │
│ ❱ 116 │   │   │   raise e                                                                        │
│   117 │   │   self.callback_manager.on_chain_end(outputs, verbose=self.verbose)                  │
│   118 │   │   return self.prep_outputs(inputs, outputs, return_only_outputs)                     │
│   119                                                                                            │
│                                                                                                  │
│ /home/hysz/anaconda3/envs/chatglm/lib/python3.10/site-packages/langchain/chains/base.py:113 in   │
│ __call__                                                                                         │
│                                                                                                  │
│   110 │   │   │   verbose=self.verbose,                                                          │
│   111 │   │   )                                                                                  │
│   112 │   │   try:                                                                               │
│ ❱ 113 │   │   │   outputs = self._call(inputs)                                                   │
│   114 │   │   except (KeyboardInterrupt, Exception) as e:                                        │
│   115 │   │   │   self.callback_manager.on_chain_error(e, verbose=self.verbose)                  │
│   116 │   │   │   raise e                                                                        │
│                                                                                                  │
[...]

PR #14356, opened 2023-12-06
prevent regexparser output for map_rerank from failing

Siddhant-Ray:
Description: The Map Rerank chain fails if, for any of the intermediate outputs, a score is not generated. This happens because the regex parser is an exact match on the "Score" keyword. No exception handling is done if this score is not generated and the pipeline fails. This is cumbersome to fix manually for a large set of queries.

Fix: Add an exception case that if a score is not generated, add the "Score : 0" for that case. Logically this means the same as a score having no meaning.

This fix resolves issue: #3970


Siddhant-Ray:
@baskaryan Could you take a look when possible?

hwchase17:
this fix is not general enough to add to the llmChain class

PR #14375, opened 2023-12-07
[Feature] RecursiveURLLoader: Custom PreventOutside

hinthornw:
For websites where you start at a .html file, you aren't able to load sibling paths. If we allow prevent_outside to be another url, you can do things like:
start at `https://api.python.langchain.com/en/latest/core_api_reference.html` and prevent_outside to be `https://api.python.langchain.com/en/latest/`


baskaryan:
added a base_url param that allows this!

PR #14384, opened 2023-12-07
Update minimax.py

ethynic:
MiniMaxChat class _generate method shoud return a ChatResult object not str


libs/langchain/langchain/chat_models/minimax.py
```diff
@@ -79,9 +79,11 @@ def _generate(
         payload = self._default_params
         payload["messages"] = history
         text = self._client.post(payload)
+        generations = ChatGeneration(message=AIMessage(content=text))
+        return ChatResult(generations=[generations])
 
         # This is required since the stop are not enforced by the model parameters
-        return text if stop is None else enforce_stop_tokens(text, stop)
+        # return text if stop is None else enforce_stop_tokens(text, stop)
```
hwchase17, line 86: dont we need to still enforce stop tokens?

baskaryan (approved):

Issue #14338, opened 2023-12-06
api_base parameter not properly propagated in client calls in ChatLiteLLM


DavidLMS:
### System Info

- LangChain version: 0.0.346
- Platform: Mac mini M1 16GB - macOS Sonoma 14.0
- Python version: 3.11
- LiteLLM version: 1.10.6

### Who can help?

@hwchase17
@agola11

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

```
from langchain.chat_models import ChatLiteLLM

# Code for initializing the ChatLiteLLM instance
chat_model = ChatLiteLLM(api_base="https://custom.endpoints.huggingface.cloud", model="huggingface/Intel/neural-chat-7b-v3-1")

# Make a call to LiteLLM
text = "What would be a good company name for a company that makes colorful socks?"
messages = [HumanMessage(content=text)]

print(chat_model(messages).content)
```

Error:

```
Traceback (most recent call last):
  File "/opt/homebrew/lib/python3.11/site-packages/litellm/utils.py", line 4919, in handle_huggingface_chunk
    raise ValueError(chunk)
ValueError: {"error":"The model Intel/neural-chat-7b-v3-1 is too large to be loaded automatically (14GB > 10GB). Please use Spaces (https://huggingface.co/spaces) or Inference Endpoints (https://huggingface.co/inference-endpoints)."}
```

Is the same error if:
`chat_model = ChatLiteLLM(model="huggingface/Intel/neural-chat-7b-v3-1")`

So api_base parameter not properly propagated in client calls in ChatLiteLLM.

### Expected behavior

I would expect the ChatLiteLLM instance to correctly utilize the api_base parameter when making requests to the LiteLLM client. This should enable using models larger than the default size limit without encountering the error message about model size limits.

[...]

PR #14393, opened 2023-12-07
add api_base to _client_params

DavidLMS:
- **Description:** This PR adds `api_base` to `_client_params` in the `chat_model` of LiteLLM to ensure it's included in API calls. Previously, `api_base` was set on the client but was not included in the parameters passed to the completion function. This change ensures that `api_base` is correctly passed to all API calls.
  - **Issue:** #14338
  - **Tag maintainer:** @hwchase17 @agola11
  - **Twitter handle:** @LMS_David_RS


DavidLMS:
Changed to community version in #14644.

PR #14413, opened 2023-12-07
Update prompts.py

Sebas102507:


hwchase17:
this seemingly deletes all files? not sure if a mistake

PR #14414, opened 2023-12-07
core[patch], langchain[patch], experimental[patch]: import CI

efriis:


libs/core/Makefile
```diff
@@ -27,7 +31,7 @@ lint_diff format_diff: PYTHON_FILES=$(shell git diff --relative=libs/experimenta
 
 lint lint_diff:
 	./scripts/check_pydantic.sh .
-	./scripts/check_imports.sh
+	./scripts/lint_imports.sh
```
efriis, line 34: renamed this to not confuse with `make check_imports`

.github/workflows/_pydantic_compatibility.yml
efriis, line 1: can rename these to reflect they do more than pydantic compatibility

libs/langchain/langchain/vectorstores/_pgvector_data_models.py
efriis, line 1: lmk if this is overkill because `_` file. Feel like it's better to be too strict than too loose though.

libs/langchain/langchain/utilities/github.py
efriis, line 1: Look at CI finding actual problems :)

Issue #14419, opened 2023-12-07
Dependency Conflict between litellm and tiktoken


PeterNotFound114:
### System Info

Langchain Version: 0.0.348
Output from `poetry env info`
**Virtualenv**
Python:         3.9.12
Implementation: CPython
Path:           /Users/peternf/Desktop/langchain/libs/langchain/.venv
Executable:     /Users/peternf/Desktop/langchain/libs/langchain/.venv/bin/python
Valid:          True

**System**
Platform:   darwin
OS:         posix
Python:     3.9.12
Path:       /Users/peternf/opt/miniconda3
Executable: /Users/peternf/opt/miniconda3/bin/python3.9

### Who can help?

@eyurtsev 

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reason to Install litellm
@xavier-xia-99 and I are collaborating on finishing the tasks of #12165 for _libs/langchain/langchain/chat_models/litellm.py_. We have finished the implementation and is adding unit tests to verify our logic when we discovered that we need to include litellm for the unit tests, or `make test` will just skip our newly added unit tests. Yet as we try to add the optional dependency litellm according to the instructions provided in langchain/.github/CONTRIBUTING.md, it shows the following dependency conflict. Hence, we will appreciate any help in addressing this conflict or running the CI. Thank you!

### Steps to Reproduce the Behavior

1. Run `poetry add --optional litellm`
2. Error Message: <img width="720" alt="Screenshot 2023-12-07 at 5 50 43 PM" src="https://github.com/langchain-ai/langchain/assets/98713019/3e0d5730-2134-4e11-b901-fbda927bc796">  

### Expected behavior

Output of litellm downloads in progress and successful installation of relevant packages.

PR #14420, opened 2023-12-07
feat: masking api key for litellm 

xavier-xia-99:
Description: Masking API key for LiteLLM Chat Model to protect user secrets.
Issue: https://github.com/langchain-ai/langchain/issues/12165
Dependencies: Litellm and LangChain 
Tag maintainer: @eyurtsev 

`make test` skips current unit_test and we are unable to install the dependency `litellm` due to a conflict with `tiktoken` See #14419

### DIFF
--- a/libs/langchain/langchain/chat_models/litellm.py
+++ b/libs/langchain/langchain/chat_models/litellm.py
@@ -51,11 +51,13 @@ from langchain.utils import get_from_dict_or_env
 
 logger = logging.getLogger(__name__)
 
+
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/chat_models/litellm.py
baskaryan, line 1: out of curiosity, do these api keys even need to be passed in? doesn't seem like they're used anywhere?

hwchase17:
closing due to merge conflicts

Issue #11592, opened 2023-10-10
Return ids for the document returned from the Similarity Search.


meet1919:
### Feature request

I want to get the ids of the document returned when performing `similarity_search()` or `similarity_search_with_score()`. The id should be present in the metadata = {"id": id}

  

### Motivation

Want to update the metadata of the documents that are returned in the similarity search. This update can only be done if the the documents returned has the id in its metadata. When adding the documents to the vectordb, I am not adding the ids, as they are automatically generated if not passed. 

### Your contribution

No contributions but below is the changes that can be made:

```
def _results_to_docs_and_scores(results: Any) -> List[Tuple[Document, float]]:
    return [
        # TODO: Chroma can do batch querying,
        # we shouldn't hard code to the 1st result
        (Document(page_content=result[0], metadata=(result[1] | {'id': result[3]}) or {}), result[2])
        for result in zip(
            results["documents"][0],
            results["metadatas"][0],
            results["distances"][0],
            results["ids"][0]
        )
    ]

```

FelixOpolka:
This would indeed be very useful!

PR #14423, opened 2023-12-08
feat: patch chroma.similarity_search_with_score to return vector id

timothycarambat:
**Description:** In chroma, return the vector_ids for associated vectors when using the `similarity_search_with_score` and `similarity_search_by_vector_with_relevance_scores` so that turning around to manage the vector (update/delete) is instantly available. Right now, there is no easy way to get similarity and get the associated vector id for the results. Given the id on creation can be created in-library or _optionally_ set by the user we should greedily return the id so the user can always manage their vectors easily.
**Issue:** #11592,
**Dependencies:** N/A
**Tag maintainer:** N/A,
**Twitter handle:** @mintplexlabs

### DIFF
--- a/libs/langchain/langchain/vectorstores/chroma.py
+++ b/libs/langchain/langchain/vectorstores/chroma.py
@@ -36,15 +36,16 @@ def _results_to_docs(results: Any) -> List[Document]:
     return [doc for doc, _ in _results_to_docs_and_scores(results)]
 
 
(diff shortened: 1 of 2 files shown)


hwchase17 (commented):
im a bit hesitant to modify public methods signatures (would be backwards breaking)

could we include vector id in document metadata? or is that too hacky?

PR #14424, opened 2023-12-08
GITLAB_URL should take default https://gitlab.com instead of error

manjunathshiva:
The fix #14221 has broken default gitlab url which is forcing the users to specify GITLAB_URL  for default one.  With this fix if GITLAB_URL  is not set, the default gitlab url  will be taken.


manjunathshiva:
Due to movement of logic to different package. Opened #14638 for this fix.

Issue #14403, opened 2023-12-07
Issue: ElasticsearchStore with ssl_verify = {'verify_certs':False} does not work


Civel-1:
### Issue you'd like to raise.

I am using langchain 0.0.316 an trying to create a ElasticsearchStore to do some similarity_search

However, whenever I try to create it (using from_documents or not) I get the following error : 

`raise ValueError("check_hostname requires server_hostname")`

This is a SSL error, and I suspect it to be the problem as I cannot use SSL. Somewhere else in the project, I cannot with Python to Elastic using verify_certs = False and everything works perfectly.

Thus I tried to create ElasticsearchStore with the following arguments :

```
db = ElasticsearchStore(texts,
                                      es_url='...',
                                      index_name = '...',
                                      embedding,
                                      ssl_verify = {'verify_certs'=False'})
```

But I still have the error and nothing has changed. 
How can I make langchain to initialize Elastic without checking for ssl certificates ?

Civel-1:
Are you sure that SSL certificates to false are not supported yet ? 

Looking at this piece of code from the langchain tests files : 
```
    def test_similarity_search_with_ssl_verify(self, elasticsearch_url: str) -> None:
        """Test end to end construction and search with ssl verify."""
        ssl_verify = {
            "verify_certs": True,
            "basic_auth": ("ES_USER", "ES_PASSWORD"),
            "ca_certs": "ES_CA_CERTS_PATH",
        }
        texts = ["foo", "bar", "baz"]
        docsearch = ElasticVectorSearch.from_texts(
            texts,
            FakeEmbeddings(),
            elasticsearch_url=elasticsearch_url,
            ssl_verify=ssl_verify,
        )
        output = docsearch.similarity_search("foo", k=1)
        assert output == [Document(page_content="foo")]
```
If looks like verify_certs is a supported parameter ? Or is it just supported when True and not False ?

manel00:
Hello, is there a way to verify the certs with the param already? for me it is not working with it still

Civel-1:
I don't think so,.
For me, creating a elastic connection first initialized with verify_certs to false then use it as a es_connection param when creating the ElasticsearchStore did the trick.

RyanAlexton:
> I found a way to work around this. You can instantiate a ElasticsearchStore by using an existing elastic connection.
> 
> ```
> elastic_client = Elasticsearch(..., verify_certs=False)
> db = ElasticsearchStore(texts,
>                                       index_name = '...',
[...]

PR #14435, opened 2023-12-08
Improvement: Allow passing parameters to the underlying es_client. Closes: #14403

169:
### Description

In https://github.com/langchain-ai/langchain/issues/14403, the user mentioned that he hopes not to verify ssl and needs to pass more parameters

I found that the `Elasticsearch` class [has very many parameters](https://github.com/elastic/elasticsearch-py/blob/98f2af2134dfc3f2b2054b87a4056e2508b9b8db/elasticsearch/_sync/client/__init__.py#L131-L191
):

Screenshot 2023-12-08 at 4 24 39 PM

In order to adapt to more situations, I want to add the kwargs parameter so that users can enter more `Elasticsearch` parameters. Like [redis](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/vectorstores/redis/base.py#L253), [tair](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/vectorstores/tair.py#L32), [myscale](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/vectorstores/myscale.py#L112) and so on.

### DIFF
--- a/libs/community/langchain_community/vectorstores/elasticsearch.py
+++ b/libs/community/langchain_community/vectorstores/elasticsearch.py
@@ -256,7 +256,7 @@ class ExactRetrievalStrategy(BaseRetrievalStrategy):
         elif similarity is DistanceStrategy.DOT_PRODUCT:
             similarityAlgo = f"""
             double value = dotProduct(params.query_vector, '{vector_query_field}');
-            return sigmoid(1, Math.E, -value); 
+            return sigmoid(1, Math.E, -value);
             """
         else:
             raise ValueError(f"Similarity {similarity} not supported.")
@@ -512,6 +512,7 @@ class ElasticsearchStore(VectorStore):
             ]
         ] = None,
         strategy: BaseRetrievalStrategy = ApproxRetrievalStrategy(),
+        es_params: Optional[Dict[str, Any]] = None,
     ):
         self.embedding = embedding
         self.index_name = index_name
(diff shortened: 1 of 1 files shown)


eyurtsev:
cc @joemcelroy 

Should we have this be a dict rather than kwargs?

libs/langchain/langchain/vectorstores/elasticsearch.py
```diff
@@ -509,6 +509,7 @@ def __init__(
             ]
         ] = None,
         strategy: BaseRetrievalStrategy = ApproxRetrievalStrategy(),
+        **kwargs: Any,
```
eyurtsev, line 512: I'm a bit worried about using **kwargs in such a large initializer; perhaps we could accept a dict to make it more targeted?

es_params: Optional[Dict[str, Any]] = None

and unpack at destination with **es_params

169: Fixed: https://github.com/langchain-ai/langchain/pull/14435/commits/95e5b51515f9a0e80a61f59bd3630c4727c7c748

169:
@baskaryan, @eyurtsev, @hwchase17 I have changed it to the code in `langchain_community`, PTAL

joemcelroy (approved):
thanks for update!

Would be good to also document this. I will if you aren't able to :)

hwchase17 (approved):

PR #14446, opened 2023-12-08
Fix: redundant casting to SecretStr API keys

### LINKED ISSUE #14445: Issue: Flawed implementations of SecretStr for API keys
### Issue you'd like to raise.

During the recent initiative to secure API keys with `SecretStr` (https://github.com/langchain-ai/langchain/issues/12165), some implementations and their corresponding tests were implemented with some flaws. More specifically, they were not really masking the API key.

For instsance, in `libs/langchain/langchain/chat_models/javelin_ai_gateway.py` we have:
```
    @property
    def _default_params(self) -> Dict[str, Any]:
        params: Dict[str, Any] = {
            "gateway_uri": self.gateway_uri,
            "javelin_api_key": cast(SecretStr, self.javelin_api_key).get_secret_value(),
            "route": self.route,
            **(self.params.dict() if self.params else {}),
        }
        return params
```
In the above snippet, `self.javelin_api_key` is cast to `SecretStr`, and then immediately `.get_secret_value()` is invoked, essentially retrieving the original string. Note that Javelin chat lacks unit tests.

It's worth noting that this pattern is repeated in tests, such as in `libs/langchain/tests/integration_tests/chat_models/test_baiduqianfan.py`:
```
def test_uses_actual_secret_value_from_secret_str() -> None:
    """Test that actual secret is retrieved using `.get_secret_value()`."""
    chat = QianfanChatEndpoint(
        qianfan_ak="test-api-key",
        qianfan_sk="test-secret-key",
    )
    assert cast(SecretStr, chat.qianfan_ak).get_secret_value() == "test-api-key"
    assert cast(SecretStr, chat.qianfan_sk).get_secret_value() == "test-secret-key"
```

The point of the test would be to assert that the API key is indeed a secret, and not just cast it back and forth.
Let me point out that the test suite for baiduqianfan chat does indeed catch whether the API key is indeed masked with a `SecretStr` by capturing the stdout.

@eyurtsev @hwchase17 

### Suggestion:

PR to fix the issues

rancomp:
- **Description:** This PR serves as standardization of the masking and testing.
  - Some API keys were not really masked when stored in the `values` dict. They should be masked and cast back to string only during the `POST` request. I fixed 1 occurance of this.
  - Some tests were not covering this flaw because they were repeating the back-and-forth pattern of casting to `SecretStr` and invoking `.get_secret_value()` immediately. This pattern is confusing, and in order to clear this out I added an `extract_secret_str : SecretStr | Str | None -> str` method.
  - The code-base was filled with `secretstr, secret_str, secret_string` names. I replaced all of them with `secretstr`.
  - **Issue:** fixes https://github.com/langchain-ai/langchain/issues/14445, related https://github.com/langchain-ai/langchain/issues/12165
  - **Dependencies:** None
  - **Tag maintainer:** @eyurtsev @hwchase17


efriis (commented):
Appreciate it @rancomp ! I'll actually recommend picking this up again next week. We're in the middle of a big package organizing effort (context [here](https://github.com/langchain-ai/langchain/discussions/14243)) that may cause merge conflicts with this that will have to be resolved later next week.

Great change though! Just going to need a few more days to merge it in :)

rancomp:
Hey @efriis ! Thanks, I was actually not aware of that big move. I'll revisit this PR after the merge is complete.

efriis:
As you can see from the merge conflict list, it's in 😅 

Feel free to ping when this PR is updated!

rancomp:
Hey @efriis , can you check it out now? There's another thing I wanted to do which is clean up the tests, but I think it's better to separate it to another PR.

hwchase17:
is it possible to keep the name of the function as-is in master for this pr? two reasons:

1. makes it tough to see what logic is actually changing
2. with that change in naming, is not backwards compatible, so we'd need to make sure to bump versions across langchain/langchain-community/langchain-core at the same time which is tricky

also, in general trying to NOT make backwards incompatible changes to langchain-core, which this is

rancomp:
> is it possible to keep the name of the function as-is in master for this pr? two reasons:
> 
> 1. makes it tough to see what logic is actually changing
> 2. with that change in naming, is not backwards compatible, so we'd need to make sure to bump versions across langchain/langchain-community/langchain-core at the same time which is tricky
> 
> also, in general trying to NOT make backwards incompatible changes to langchain-core, which this is

hey. yeah definitely agree with you.

rancomp:
@hwchase17 @efriis I have a small change I'm proposing within `langchain-core` over the `convert_to_secret_str` method.  The change I'm proposing allows this to accept a `None` and treat it as an empty string `""`. This makes it 1) consistent with the behavior of the `SecretStr` type from pydantic, 2) more compatible with the `api_key` types, which are usually typed as `Optional[str]`, and defaulting to `None`.

efriis:
Ah I think this actually bypasses an important part of type checking, where some classes may want to require that the passed in string is not None.

If you want to fix the case where it wasn't actually storing a masked SecretStr separately though, that would be great!

rancomp:
> Ah I think this actually bypasses an important part of type checking, where some classes may want to require that the passed in string is not None.
> 
> If you want to fix the case where it wasn't actually storing a masked SecretStr separately though, that would be great!

I think there's confusion here. This PR:
- Still allows for the API-keys to be `None`
- Changes the way the SecretStr is extracted: Instead of casting `None` to `SecretStr` and then extracting the secret, the secret is extracted for `Union[SecretStr, str, None]` type all together through a single function.
- I made sure that all the classes use the `convert_to_secretstr` function (for standardization).

What I can also change is that the API-key type hint is changed to `Union[SecretStr, str, None]` instead of `Optional[str]` or `Optional[SecretStr]`.

What I can do is allow for the API-keys to be passed as `SecretStr` and not just `str`.

PR #14451, opened 2023-12-08
Corrected the argument alias from 'base_store' to 'docstore'

xkra:
- **Description:** The sample code in the documentation about MultiVectorRetriever throws an error. The reason is an incorrect reference to an argument alias (`base_store` instead of `docstore`) The definition of the MultiVectorRetriever class shows that the correct argument alias is `docstore`. This fixes the error.,


eyurtsev (approved):

efriis:
Actually would love to move people to using a `ByteStore` here instead (allows for swappability with persistent ones too), so I'll close this and put up a PR fixing it in code! Looks like there was a bug in ByteStore too.

#14474

PR #14454, opened 2023-12-08
Added option to store raw html in metadata when transforming

AndreasThinks:
- **Description:** Adds a store_raw_html method to the Html2Text transformer, allowing quick storage of the raw_html for future use
  - **Issue:** NA
  - **Dependencies:** NA

### DIFF
--- a/docs/docs/integrations/document_transformers/html2text.ipynb
+++ b/docs/docs/integrations/document_transformers/html2text.ipynb
@@ -61,8 +61,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "urls = [\"https://www.espn.com\", \"https://lilianweng.github.io/posts/2023-06-23-agent/\"]\n",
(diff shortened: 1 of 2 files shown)


efriis:
The raw html *should* already exist in the documents you loaded from AsyncHtmlLoader. If you want to store the page content in your metadata, I would recommend just doing something like the following

```
from langchain.document_transformers import Html2TextTransformer
from langchain.document_loaders import AsyncHtmlLoader

urls = ["https://www.espn.com", "https://lilianweng.github.io/posts/2023-06-23-agent/"]
loader = AsyncHtmlLoader(urls)
docs = loader.load()

html2text = Html2TextTransformer()
docs_transformed = html2text.transform_documents(docs)

for d, dt in zip(docs, docs_transformed):
    dt.metadata["original_page"] = d.page_content
```

PR #14457, opened 2023-12-08
rail_parser: Simplify logic

irgolic:
- **Description:** Changes GuardrailsOutputParser to be a simple wrapper of the Guard object
  - **Issue:** none
  - **Dependencies:** none
  - **Tag maintainer:** for a quicker response, tag the relevant maintainer (see below),
  - **Twitter handle:** none

### DIFF
--- a/libs/langchain/langchain/output_parsers/rail_parser.py
+++ b/libs/langchain/langchain/output_parsers/rail_parser.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Dict
 
 from langchain_core.output_parsers import BaseOutputParser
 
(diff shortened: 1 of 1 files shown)


efriis (commented):
I believe this will be a breaking change because it no longer maps `api` to `llm_api`. https://docs.guardrailsai.com/api_reference/guard/#guardrails.guard.Guard.__call__

I'm tempted to keep this as-is and close the PR unless there's a strong reason to remove documentation for `api` for some reason!

If this is a change on the Guardrails side, it might be better to warnings.warn a deprecation error re: switching `api` to `llm_api` if the goal is to get people to switch the param they use :)

efriis:
Happy to chat here and reopen if it makes sense!

irgolic:
Hey @efriis, nice to meet you, and thanks for reviewing this. I understand and agree with your rationale behind closing the proposed change.

Without making a backwards-incompatible change, we'd like to consolidate the GuardrailsOutputParser interface to just `from_guard`, that takes an already created `Guard` object as an argument. Guardrails' interface is still relatively fluid as it's in 0.x, we hope this will make it simpler to use different versions of Guardrails with Langchain.

I've updated the branch with this new classmethod, and deprecation warnings for the others. I don't think the PR is updating because it's closed though. Could you please reopen and review it? Thanks!

irgolic:
Hey, I've made a few more changes and opened a new pull request here: https://github.com/langchain-ai/langchain/pull/14657

PR #14464, opened 2023-12-08
When using OpenAI Function Calling, change default model in ChatOpenAI

ggeutzzang:
- **Description:** : After reviewing the OpenAI Function Calling official guide at https://platform.openai.com/docs/guides/function-calling,  the following information was noted:

> "The latest models (gpt-3.5-turbo-1106 and gpt-4-1106-preview) have been trained to both detect when a function should be called (depending on the input) and to respond with JSON that adheres to the function signature more closely than previous models. With this capability also comes potential risks. We strongly recommend building in user confirmation flows before taking actions that impact the world on behalf of users (sending an email, posting something online, making a purchase, etc)."

Therefore, I have updated the default model in ChatOpenAI from 'gpt-3.5-turbo' to 'gpt-3.5-turbo-1106' only when function calling is used and no specific model is designated by the user.

### DIFF
--- a/libs/langchain/langchain/chat_models/openai.py
+++ b/libs/langchain/langchain/chat_models/openai.py
@@ -240,6 +240,10 @@ class ChatOpenAI(BaseChatModel):
         """Build extra kwargs from additional params that were passed in."""
         all_required_field_names = get_pydantic_field_names(cls)
         extra = values.get("model_kwargs", {})
+
+        if "model" not in values and {"tools", "functions"} & values.keys():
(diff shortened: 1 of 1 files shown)


efriis:
Appreciate it @ggeutzzang , and I'd recommend using these models in your own code.

We're moving away from making prescriptive model decisions in the library, so instead of this, it may be better to just update the openai functions docs to use these models: https://python.langchain.com/docs/modules/chains/how_to/openai_functions

Will close this, and feel free to tag me in a docs update if you do that!

PR #14466, opened 2023-12-08
SmartLLMChain Output Key Customization

anishnag:
**Description**
The `SmartLLMChain` was was fixed to output key "resolution". Unfortunately, this prevents the ability to use multiple `SmartLLMChain` in a `SequentialChain` because of colliding output keys. This change simply gives the option the customize the output key to allow for sequential chaining. The default behavior is the same as the current behavior.

Now, it's possible to do the following:
```
from langchain.chat_models import ChatOpenAI
from langchain_experimental.smart_llm import SmartLLMChain
from langchain.chains import SequentialChain

joke_prompt = PromptTemplate(
    input_variables=["content"],
    template="Tell me a joke about {content}.",
)
review_prompt = PromptTemplate(
    input_variables=["scale", "joke"],
    template="Rate the following joke from 1 to {scale}: {joke}"
)

llm = ChatOpenAI(temperature=0.9, model_name="gpt-4-32k")
joke_chain = SmartLLMChain(llm=llm, prompt=joke_prompt, output_key="joke")
review_chain = SmartLLMChain(llm=llm, prompt=review_prompt, output_key="review")

chain = SequentialChain(
    chains=[joke_chain, review_chain],
    input_variables=["content", "scale"],
    output_variables=["review"],
    verbose=True
)
response = chain.run({"content": "chickens", "scale": "10"})
print(response)
```

### DIFF
--- a/libs/experimental/langchain_experimental/smart_llm/base.py
+++ b/libs/experimental/langchain_experimental/smart_llm/base.py
@@ -170,7 +170,7 @@ class SmartLLMChain(Chain):
         self.history.critique = critique
         resolution = self._resolve(stop, run_manager)
         if self.return_intermediate_steps:
-            return {"ideas": ideas, "critique": critique, self.ouput_key: resolution}
+            return {"ideas": ideas, "critique": critique, self.output_key: resolution}
         return {self.output_key: resolution}
 
     def _get_text_from_llm_result(self, result: LLMResult, step: str) -> str:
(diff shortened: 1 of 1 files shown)


efriis (approved):
Thanks! Would also recommend checking out LCEL (RunnableSequence instead of SequentialChain), where passing data through different keys is more flexible by design :) https://python.langchain.com/docs/expression_language/how_to/passthrough

PR #14471, opened 2023-12-08
docs cleanup

leo-gan:
Fixed inconsistencies; added links and descriptions


efriis (commented):
Hey @leo-gan ! Just noticed the explanation paragraph is pulled from Wikipedia. Could you cite that source?

Something simple like an addition link to the wikipedia page like `(Source)` at the end should suffice, just to be clear that the paragraph comes from there.

leo-gan:
@efriis Sure. I've added it.

efriis (approved):

PR #14479, opened 2023-12-09
fix(embeddings): define task using parameters instead of json payload for HuggingFaceHubEmbeddings

randyphoa:
**Description:** This pull request explicitly defines the task of an embedding feature extraction. The current issue is the default task for models like `sentence-transformers/all-mpnet-base-v2`, is not `feature-extraction` and when running `HuggingFaceHubEmbeddings` causes an error `SentenceSimilarityInputsCheck expected dict not list`.
 
  **Issue:** By explicitly defining the task as `feature-extraction` as per [Inference documentation](https://huggingface.co/docs/huggingface_hub/v0.19.3/en/package_reference/inference_client#huggingface_hub.AsyncInferenceClient.post), it ensures `HuggingFaceHubEmbeddings` is using the right task. Passing the task as a json payload does not work.

Here's the problem,
```
embeddings = HuggingFaceHubEmbeddings(
    model="sentence-transformers/all-mpnet-base-v2",
    task="feature-extraction",
    huggingfacehub_api_token=os.getenv("HF_TOKEN"),
)
embeddings.embed_query("Hello Langchain!")
```

Error:

```
Bad request:
SentenceSimilarityInputsCheck expected dict not list: `__root__` in `parameters`
```

From [huggingface_hub.py](https://github.com/langchain-ai/langchain/blob/d9bfdc95ea9e37b88cdfe7c2baef51ad52713463/libs/langchain/langchain/embeddings/huggingface_hub.py#L95), a simplified problem and fix.

Current, task passed as json payload:
```
client.post(json={"inputs": text, "task": "feature-extraction"})
```

Fix, task passed explicitly:
```
client.post(json={"inputs": text}, task="feature-extraction")
```

The final fix is in the pull request.
```
responses = self.client.post(
    json={"inputs": texts, "parameters": _model_kwargs},
    task=self.task
)
```

 
 **Dependencies:** None
 **Tag maintainer:** @hwchase17
 **Twitter handle:** @randyphoa

### DIFF
--- a/libs/langchain/langchain/embeddings/huggingface_hub.py
+++ b/libs/langchain/langchain/embeddings/huggingface_hub.py
@@ -93,7 +93,8 @@ class HuggingFaceHubEmbeddings(BaseModel, Embeddings):
         texts = [text.replace("\n", " ") for text in texts]
         _model_kwargs = self.model_kwargs or {}
         responses = self.client.post(
-            json={"inputs": texts, "parameters": _model_kwargs, "task": self.task}
+            json={"inputs": texts, "parameters": _model_kwargs},
+            task=self.task
         )
         return json.loads(responses.decode())
 


hwchase17:
fixed in separate pr! thanks for catching tho

PR #14481, opened 2023-12-09
feat: adding paygo api support

santiagxf:
- **Description:** Introducing support for LLMs and Chat models running in Azure AI studio and Azure ML using the new deployment mode pay-as-you-go (model as a service).
- **Issue:** NA
- **Dependencies:** None.
- **Tag maintainer:** @prakharg-msft @gdyre 
- **Twitter handle:** @santiagofacundo

Examples added:
* [docs/docs/integrations/llms/azure_ml.ipynb](https://github.com/santiagxf/langchain/blob/santiaxf/azureml-endpoints-paygo/docs/docs/integrations/llms/azure_ml.ipynb)
* [docs/docs/integrations/chat/azureml_chat_endpoint.ipynb](https://github.com/santiagxf/langchain/blob/santiaxf/azureml-endpoints-paygo/docs/docs/integrations/chat/azureml_chat_endpoint.ipynb)


libs/langchain/langchain/llms/azureml_endpoint.py
```diff
@@ -233,6 +280,49 @@ class AzureMLOnlineEndpoint(LLM, BaseModel):
     model_kwargs: Optional[dict] = None
     """Keyword arguments to pass to the model."""
 
+    @root_validator()
+    @classmethod
+    def validate_endpoint_api_type(cls, values: Dict) -> AzureMLBaseEndpoint:
```
gdyre: I am wondering, is there any way that we should be checking that the url matches the class? As in a url with "v1/chat/completions" should only be used in the AzureMLChatOnlineEndpoint class and vice versa?

santiagxf: @gdyre it's a good question. I was wondering the same thing. We can do that check for MaaS since URLs are different. We can't for MaaP. I don't see any point on using a chat API for a completions or vice versa. We can also warn the user instead of failing. Which one should we follow?

libs/langchain/langchain/llms/azureml_endpoint.py
```diff
@@ -101,15 +112,22 @@ def escape_special_characters(prompt: str) -> str:
...
         return prompt
 
+    @property
+    @abstractmethod
+    def supported_api_types(self) -> List[AzureMLEndpointApiType]:
+        """Supported APIs for the given formatter"""
+
     @abstractmethod
-    def format_request_payload(self, prompt: str, model_kwargs: Dict) -> bytes:
+    def format_request_payload(
```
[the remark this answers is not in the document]
santiagxf: Great point! I forgot people may have already created derivate classes from this one. Adding that change!

santiagxf:
PR abandoned as we moved to `langchain_community` on PR #14560

PR #14484, opened 2023-12-09
fix dashvector endpoint params error

Funkeke:


hwchase17 (approved):
i added a default so that this isn't a backwards incompatible change

Issue #5210, opened 2023-05-24
GPT4All chat error with async calls


PiotrPmr:
Hi, I believe this issue is related to this one: #1372 

I'm using GPT4All integration and get the following error after running `ConversationalRetrievalChain` with `AsyncCallbackManager`:
`ERROR:root:Async generation not implemented for this LLM.`
Changing to `CallbackManager` does not fix anything.

The issue is model-agnostic, i.e., I have used _ggml-gpt4all-j-v1.3-groovy.bin_ and _ggml-mpt-7b-base.bin_. The LangChain version I'm using is `0.0.179`. Any ideas how this can be potentially solved or should we just wait for a new release fixing it?

### Suggestion:

Release a fix, similar as in #1372

poojatambe:
@PiotrPmr  Hi,
Any solution found for this issue?
Getting ```NotImplementedError: Async generation not implemented for this LLM``` error for langchain version 0.0.186 with gpt4all model.

ncfx:
I'm having the same issue with another LlamaCpp LLM as well as a HuggingFaceHub LLM. I'm using LLMChain. Hoping someone can fix this!

khaledadrani:
having the same issue, +1

khaledadrani:
I managed to make a fix, and will be making a pr soon

khaledadrani:
I am currently checking with the requirements for my pr to be ready for reviewing (format, linting, testing)
I will finish them ASAP and this is my first contribution ever in an open source project :)

khaledadrani:
Hello again, I read in the contribution document that is it possible to add a jupyter notebook example  https://github.com/hwchase17/langchain/blob/master/.github/CONTRIBUTING.md, however, I am unable to find any notebooks in the repository? Can someone tell me where should I put the example notebook? Thanks!

chrisedington:
Surprised there isn't more action on this issue, would be great to see it merged. Thanks for the efforts @khaledadrani

khaledadrani:
I think someone made an implementation already, but did not report to this issue? Can anyone confirm that it works? (it happened while rebasing my fork, I found almost the same implementation)

diegovazquez:
A workaround for using ConversationalRetrievalChain with llamaCpp is implemet _acall function. This is not tested extensively.

```python

from langchain.llms import LlamaCpp
from typing import Any, Dict, List, Generator, Optional
from langchain.callbacks.manager import AsyncCallbackManagerForLLMRun

[...]

PR #14495, opened 2023-12-09
feature: implement _acall method for GPT4All(LLM)

khaledadrani:
- **Description:** Implement the _acall method of the class langchain.llms.gpt4all.GPT4All  
 - **Issue:** GPT4All chat error with async calls #5210 
 - **Dependencies:** no dependency
 - **Twitter handle:** None

Okey, I tested locally. 

I ran "make format". No issue.

I ran "make lint". I found this issue [How to fix error "cannot import name 'ModelMetaclass' from 'pydantic.main'"](https://stackoverflow.com/questions/76647518/how-to-fix-error-cannot-import-name-modelmetaclass-from-pydantic-main). I followed the answer pip install lightning==2.0.1 and the command ran without problem. I believe this is not related to the PR but I had to clarify this.

I ran "make test". No issue. 

This is more of a fix for the existing default implementation of the _acall method inherited by the GPT4All class.

This is my first contribution ever in the open source world, so let me know if there is anything wrong. 

1. Let me know if there is a need for adding unit tests and integration tests for LLMs or chains or callbacks. I read the existing tests and found out that they cover them in a generic manner.

2. I have a working example that showcases how to effectively use an asynchronous callback handler with the GPT4All and RetrievalQA. I did not find the directory docs/extra, should I put it in cookbooks?


libs/langchain/langchain/llms/gpt4all.py
```diff
@@ -209,3 +212,33 @@ def _call(
...
         return text
+
+    async def _acall(
+        self,
+        prompt: str,
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[AsyncCallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> str:
+        r"""Asynchronous Call out to GPT4All's generate method.
```
hwchase17, line 223:
```suggestion
        """Asynchronous Call out to GPT4All's generate method.
```

libs/langchain/langchain/llms/gpt4all.py
```diff
@@ -209,3 +212,33 @@ def _call(
...
+        Returns:
+            The string generated by the model.
+
+        """
+        text_callback = None
+        if run_manager:
+            text_callback = partial(run_manager.on_llm_new_token, verbose=self.verbose)
+
+        text = ""
+        for token in self.client.generate(prompt, **self._default_params()):
```
hwchase17, line 238: this probably wants to call agenerate right?

khaledadrani: > this probably wants to call agenerate right?

agenerate is not found for GPT4All?

AttributeError: 'GPT4All' object has no attribute 'agenerate'

What I meant to do, is using the streaming of the gpt4all model while awaiting the callback from the async callback handler

khaledadrani:
@hwchase17 the logic of gpt4all was changed to langchain-community, right?

khaledadrani:
@hwchase17 could you please review again?

libs/community/langchain_community/llms/gpt4all.py
```diff
@@ -209,3 +212,38 @@ def _call(
...
+
+                prompt = "Once upon a time, "
+                response = model(prompt, n_predict=55)
+        """
+        text_callback = None
+        if run_manager:
+            text_callback = partial(run_manager.on_llm_new_token, verbose=self.verbose)
+
+        text = ""
+        for token in self.client.generate(prompt, **self._default_params()):
```
baskaryan, line 243: just making sure, the client doesn't have any async generation method does it?

khaledadrani: I am assuming you are referring to agenerate. I already made a previous comment about this.

agenerate is not found for GPT4All?

AttributeError: 'GPT4All' object has no attribute 'agenerate'

I will double check now, maybe some updates occured.

khaledadrani: I checked now, the same issue if I try to replace generate by agenerate "AttributeError: 'GPT4All' object has no attribute 'agenerate'"

khaledadrani: @baskaryan I am working on a cookbook too, I will add it to show how this feature works. 
However, I have a question:
if I use this:
```
from langchain.llms.gpt4all import GPT4All

model_path = ".../mistral-7b-openorca.Q4_0.gguf"
model = GPT4All(model=model_path,
               callbacks=streaming_callbacks,
               verbose=False)
await model._acall(prompt="tell me about Python", n_predict=55)
# returns "\n" only
```
But if I integrated it in the RetrievalQA, it works with acall:

```
qa = RetrievalQA.from_chain_type(llm=model, chain_type="stuff", retriever=vector_retriever_client,
                                              return_source_documents=True)
model_response = await qa.acall(
                inputs={"query": "what is Python?"}
            )
```
I am still working in the notebook and I did not understand this behavior. This is also the exact situation I encountered in a professional project that drove me to make this PR.

khaledadrani:
I can see this warning
```
LangChainDeprecationWarning: The function `acall` was deprecated in LangChain 0.1.0 and will be removed in 0.2.0. Use ainvoke instead.
  warn_deprecated(
```
I am implementing acall. should I change the implementation to use ainvoke?

libs/community/langchain_community/llms/gpt4all.py
```diff
@@ -211,3 +214,40 @@ def _call(
         if stop is not None:
             text = enforce_stop_tokens(text, stop)
         return text
+
+    async def _acall(
```
baskaryan, line 218: doesn't LLM._acall have a default implementation that does the same thing https://github.com/langchain-ai/langchain/blob/304f3f5fc1517adf54f9d022f13e8209749b00c1/libs/core/langchain_core/language_models/llms.py#L1109

cc @nfcampos

khaledadrani: Okey, that is true. It has that method implemented. I tested it also and it works the same as this one.
My PR is linked to this issue https://github.com/langchain-ai/langchain/issues/5210, if the issue is solved already, then close both the issue and this PR.

khaledadrani: Was the current implementation for GPT4All tested with websockets for example (to test the asynchronous behavior works as intended?) (I will double check this ASAP)

khaledadrani:
user @charlord confirmed that the current ainvoke works. So this issue can be closed https://github.com/langchain-ai/langchain/issues/5210

baskaryan:
> user @charlod confirmed that the current ainvoke works. So this issue can be closed #5210

closing per ^

PR #14500, opened 2023-12-10
Validate_template modification.

venkatsubash2003:
I have made changes to the validate_template Document.
The validate_template is now set to False by default while creating a PromptTemplate. So, I have made sure that those changes are now being reflected in this new validate.ipynb file where I have written codes by using both the cases, that are validate_template = True and validate_template = True.


venkatsubash2003:
I have made some changes to index.ipynb file in the prompt_templates. 
In the cell number 19, I have added the parameter validate_template = True for validation.
For additional validation, specify input_variables explicitly and set the parameter validate_template to True. These variables will be compared against the variables present in the template string during instantiation, raising an exception if there is a mismatch. For example:

venkatsubash2003:
> I have made changes to the validate_template Document. The validate_template is now set to False by default while creating a PromptTemplate. So, I have made sure that those changes are now being reflected in this new validate.ipynb file where I have written codes by using both the cases, that are validate_template = True and validate_template = False

PR #14524, opened 2023-12-11
Add the collection_description parameter to Milvus

mokeyish:
Because Milvus' collection_name doesn't support UFT8 characters in other languages, I want the `collection_descriotion`.


hwchase17 (commented):
we should do this in a backwards compatible way, eg default to None or empty string (depending on what the default it)

mokeyish:
@hwchase17 I've updated collection_description default to `""`.

hwchase17:
> @hwchase17 I've updated collection_description default to `""`.

is that what the default is in the client?

mokeyish:
> > @hwchase17 I've updated collection_description default to `""`.
> 
> is that what the default is in the client?

Yes

hwchase17 (approved):

PR #14535, opened 2023-12-11
remove optional from ids param

### LINKED ISSUE #14409: FAISS db.delete() says `ids` is required even when it is Optional
### System Info

LangChain version : Latest

```python
    def delete(self, ids: Optional[List[str]] = None, **kwargs: Any) -> Optional[bool]:
```

[This check](https://github.com/langchain-ai/langchain/blob/54040b00a4a05e81964a1a7f7edbf0b830d4395c/libs/langchain/langchain/vectorstores/faiss.py#L798) causes the issue.

### Who can help?

@eyurtsev

### Information

- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Output Parsers
- [ ] Document Loaders
- [X] Vector Stores / Retrievers
- [ ] Memory
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

```python
db = FAISS.from_documents(text_pages, embeddings)
db.delete()
```

Error

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[51], line 1
----> 1 db.delete()

File /opt/conda/lib/python3.10/site-packages/langchain/vectorstores/faiss.py:799, in FAISS.delete(self, ids, **kwargs)
    789 """Delete by ID. These are the IDs in the vectorstore.
    790 
    791 Args:
   (...)
    796     False otherwise, None if not implemented.
    797 """
    798 if ids is None:
--> 799     raise ValueError("No ids provided to delete.")
    800 missing_ids = set(ids).difference(self.index_to_docstore_id.values())
    801 if missing_ids:

ValueError: No ids provided to delete.
```

### Expected behavior

The index should be deleted without needing to pass an index `id`.

umair313:
delete() requires ids to remove documents from the db. 
Remove the optional type hinting for ids param.

fixed #14409

@efriis 
@hwchase17 
@baskaryan 

Do you think we can add a separate function which remove's all documents without ids?

### DIFF
--- a/libs/langchain/langchain/vectorstores/faiss.py
+++ b/libs/langchain/langchain/vectorstores/faiss.py
@@ -785,7 +785,7 @@ class FAISS(VectorStore):
         )
         return docs
 
-    def delete(self, ids: Optional[List[str]] = None, **kwargs: Any) -> Optional[bool]:
+    def delete(self, ids: List[str] = None, **kwargs: Any) -> Optional[bool]:
         """Delete by ID. These are the IDs in the vectorstore.
 
         Args:


hwchase17:
needed to be this way for backwards compatibility

Issue #13637, opened 2023-11-21
Add progress bar to GooglePalmEmbeddings


ugm2:
### Feature request

Add a progress bar to `GooglePalmEmbeddings.embed_documents()` function. [tqdm](https://github.com/tqdm/tqdm) would work just fine.

In my opinion, all embedders should have a progress bar.

### Motivation

When processing embeddings the user should have an idea of how much time is going to take to embed the data. While using GooglePalmEmbeddings, which are not the fastest, I couldn't see a progress bar and that was frustrating me because I had no idea if it was even accessing GooglePalm correctly or how much time it was going to take.

### Your contribution

```python
from __future__ import annotations

import logging
from typing import Any, Callable, Dict, List, Optional
from tqdm import tqdm

from langchain_core.pydantic_v1 import BaseModel, root_validator
from langchain_core.schema.embeddings import Embeddings
from tenacity import (
    before_sleep_log,
    retry,
    retry_if_exception_type,
    stop_after_attempt,
    wait_exponential,
)

from langchain.utils import get_from_dict_or_env

logger = logging.getLogger(__name__)

...

class GooglePalmEmbeddings(BaseModel, Embeddings):
    """Google's PaLM Embeddings APIs."""

    client: Any
    google_api_key: Optional[str]
    model_name: str = "models/embedding-gecko-001"
    """Model name to use."""

    @root_validator()
    def validate_environment(cls, values: Dict) -> Dict:
        """Validate api key, python package exists."""
        google_api_key = get_from_dict_or_env(
            values, "google_api_key", "GOOGLE_API_KEY"
        )
        try:
            import google.generativeai as genai

            genai.configure(api_key=google_api_key)
        except ImportError:
            raise ImportError("Could not import google.generativeai python package.")

        values["client"] = genai

        return values

    def embed_documents(self, texts: List[str]) -> List[List[float]]:
        return [self.embed_query(text) for text in tqdm(texts)]

    def embed_query(self, text: str) -> List[float]:
        """Embed query text."""
        embedding = embed_with_retry(self, self.model_name, text)
        return embedding["embedding"]
```

PR #14542, opened 2023-12-11
Adds progress bar to VertexAIEmbeddings

ugm2:
- **Description:** Adds progress bar to VertexAIEmbeddings 
  - **Issue:** related issue https://github.com/langchain-ai/langchain/issues/13637

Signed-off-by: ugm2 <unaigaraymaestre@gmail.com>


libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -6,11 +7,14 @@
...
 from langchain.utilities.vertexai import raise_vertex_import_error
 
+logger = logging.getLogger(__name__)
+
 
 class VertexAIEmbeddings(_VertexAICommon, Embeddings):
     """Google Cloud VertexAI embedding models."""
 
     model_name: str = "textembedding-gecko"
+    show_progress_bar: bool = False
```
leo-gan, line 17: @eyurtsev @baskaryan  Is it possible to add it to the base Embeddings? It looks like a useful generic capability.

ugm2: It would be nice, but the base Embeddings class is just an interface. And adding the attribute doesn't necessarily add the capability itself, because then you need to implement it on the embed_documents function

libs/langchain/langchain/embeddings/vertexai.py
```diff
@@ -36,11 +40,32 @@ def embed_documents(
...
+            except ImportError:
+                logger.warning(
+                    "Unable to show progress bar because tqdm could not be imported. "
+                    "Please install with `pip install tqdm`."
+                )
+                progress_bar = None
+            else:
+                progress_bar = tqdm(total=len(texts), desc="VertexAIEmbeddings")
+        else:
+            progress_bar = None
```
hwchase17, line 55: I think theres a clear way to do this where you dont have use the progress-bar update and close below
```suggestion
        if self.show_progress_bar:
            try:
                from tqdm import tqdm
            except ImportError:
                logger.warning(
                    "Unable to show progress bar because tqdm could not be imported. "
                    "Please install with `pip install tqdm`."
                )
                _iterator = range(0, len(texts), batch_size)
            else:
                _iterator = tqdm(range(0, len(texts), batch_size), total=len(texts), desc="VertexAIEmbeddings")
        else:
            _iterator = range(0, len(texts), batch_size)
```

ugm2: @hwchase17 
I'm using the progress-bar update so people know the total number of docs. To me, maybe just me, that's important. The tricky part is that we are using a batch size, so that's why I'm manually updating the progress bar.

In your approach, you are creating a `range(0, len(texts), batch_size)` but then setting the total to `len(texts)`. With that the progress bar will increase 1 by 1, but internally it's progressing at batch size, which will make the progress bar not get filled completely.

Here you have a python script that you can execute on your shell to prove my point:

```python
from tqdm import tqdm
import time

texts = [1,2,3,4,5,6,7,8,9,10]
batch_size = 2
_iterator = tqdm(range(0, len(texts), batch_size), total=len(texts), desc="VertexAIEmbeddings")

for a in _iterator:
    time.sleep(1)
```
VertexAIEmbeddings:  50%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▌                                                                                                                    | 5/10 [00:06<00:06,  1.30s/it]

ugm2:
Hi @hwchase17 ! I'm sure you guys are all swamped, but just wondering what's the status on this 🙂 Thanks!!

hwchase17:
sorry for delay @ugm2 - this was moved to community package and as such theres some merge conflicts

ugm2:
@hwchase17 hi! I see you guys have changed to use langchain_community. So I adapted the code I had added to what's currently in vertexai.py. You can review now

efriis (approved):

PR #14543, opened 2023-12-11
Fix duplication of query embedding for OpenSearch MMR

justinas-kazanavicius:
- **Description:** Currently, any time query is being embedded, it is being embedded twice. This is unecessary and expensive. This PR addresses this by embedding the query only do once and passing the existing embedding to _raw_similarity_search_with_score.
  - **Issue:** Haven't created one,
  - **Dependencies:** None

### DIFF
--- a/libs/langchain/langchain/vectorstores/opensearch_vector_search.py
+++ b/libs/langchain/langchain/vectorstores/opensearch_vector_search.py
@@ -555,7 +555,7 @@ class OpenSearchVectorSearch(VectorStore):
         return documents_with_scores
 
     def _raw_similarity_search_with_score(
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/vectorstores/opensearch_vector_search.py
```diff
@@ -555,7 +555,7 @@ def similarity_search_with_score(
         return documents_with_scores
 
     def _raw_similarity_search_with_score(
```
hwchase17, line 557: since this is only used in two places, would it make sense to update the other place to also pass in embeddings? that way we could make the method signature require fewer args and be tighter

PR #14551, opened 2023-12-11
docs[patch]: fix star bullets

efriis:


efriis:
Problem is newlines not stars

PR #14554, opened 2023-12-11
Simplify Huggingface Object Parsing

Amyh102:
**Description:** Simplifies [previous PR](https://github.com/langchain-ai/langchain/pull/13864) by removing a helper function


hwchase17 (commented):
this actually changes the results

if its already a string, instead of being `"hi"` its now `'"hi"'`

im guessing thats incorrect and we want actually keep this as is? I'm going to close this for now, feel free to reopen if wrong

Issue #14550, opened 2023-12-11
GithubAPIWrapper throws list index out of range error


usamasaleem1:
### System Info

Name: langchain
Version: 0.0.348

Name: PyGithub
Version: 2.1.1

### Who can help?
@hwchase17 
@agola11 
``` 
Traceback (most recent call last):
  File "/Users/mac/Dev Projects/Chainlit_qa/test.py", line 12, in <module>
    github = GitHubAPIWrapper(
             ^^^^^^^^^^^^^^^^^
  File "/Users/mac/Dev Projects/Chainlit_qa/myenv/lib/python3.11/site-packages/pydantic/v1/main.py", line 339, in __init__
    values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mac/Dev Projects/Chainlit_qa/myenv/lib/python3.11/site-packages/pydantic/v1/main.py", line 1102, in validate_model
    values = validator(cls_, values)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mac/Dev Projects/Chainlit_qa/myenv/lib/python3.11/site-packages/langchain/utilities/github.py", line 69, in validate_environment
    installation = gi.get_installations()[0]
                   ~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/Users/mac/Dev Projects/Chainlit_qa/myenv/lib/python3.11/site-packages/github/PaginatedList.py", line 62, in __getitem__
    return self.__elements[index]
           ~~~~~~~~~~~~~~~^^^^^^^
IndexError: list index out of range
```

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [X] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Run basic python file:

```
from langchain.agents.agent_toolkits.github.toolkit import GitHubToolkit
from langchain.utilities.github import GitHubAPIWrapper

github = GitHubAPIWrapper()
[...]

PR #14555, opened 2023-12-11
Helpefull Message for GitHubAPIWrapper

keenborder786:
Very simple change in relation to the issue #14550


hwchase17:
probably need to update this to point against the file in langchain_community. also seems like the person in the issue ran into a separate issue?

also, i read this error message you added and I don't 100% understand, could be good to make it even more clear

keenborder786:
@hwchase17 okay

PR #14556, opened 2023-12-11
[Template] OAI Functions Agent Gmail Example

hinthornw:
Imports from community and core rather than langchain wherever needed.


hinthornw:
Resolved in #14567

PR #14557, opened 2023-12-11
added new generative models on VertexAI

lkuligin:
- **Description:** added new models on VertexAI
  - **Twitter handle:** @lkuligin


libs/community/langchain_community/llms/vertexai.py
```diff
@@ -1,22 +1,17 @@
...
-    ClassVar,
-    Dict,
-    Iterator,
-    List,
-    Optional,
-)
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, Iterator, List, Optional, Union
 
-from langchain_core.callbacks import (
+from langchain.callbacks.manager import (
```
baskaryan, line 6: should import from core lines 6-12
```python
from langchain_core.callbacks import AsyncCallbackManagerForLLMRun ,CallbackManagerForLLMRun
from langchain_core.language_models import BaseLLM
```

lkuligin: thanks, missed it!

libs/community/langchain_community/llms/vertexai.py
```diff
@@ -288,6 +304,19 @@ def get_num_tokens(self, text: str) -> int:
...
 
+    def _response_to_generation(
+        self, response: TextGenerationResponse
+    ) -> GenerationChunk:
+        """Converts a stream response to a generation chunk."""
+        try:
+            generation_info = {
+                "is_blocked": response.is_blocked,
+                "safety_attributes": response.safety_attributes,
+            }
```
holtskinner, line 315: Is there a class to create this instead of a dict?

libs/community/langchain_community/llms/vertexai.py
```diff
@@ -309,11 +338,89 @@ def _generate(
...
+                vertex_messages.append(content["text"])
+            elif content["type"] == "image_url":
+                path = content["image_url"]["url"]
+                if path.startswith("gs://"):
+                    vertex_messages.append(
+                        _load_image_from_gcs(path=path, project=self.project)
+                    )
+                else:
+                    vertex_messages.append(Image.load_from_file(path))
+            else:
```
artdent, line 420: It would be good to find a way to allow passing in an image as `bytes` instead of only allowing loading from a file or from GCS. Unfortunately I can't think of a good way to distinguish that case while staying strictly compatible with the OpenAI message format. If you're willing to extend the format, perhaps you could add a case like:

```
elif content["type"] == "image":
  vertex_messages.append(Image.from_bytes(content["image"])
```

(or I can send this as a followup PR)

lkuligin: done

libs/community/langchain_community/llms/vertexai.py
lkuligin: I don't know one, it looks to me it's vertex-specific.

holtskinner (approved):

libs/community/langchain_community/utilities/vertexai.py
```diff
@@ -105,3 +106,19 @@ def get_client_info(module: Optional[str] = None) -> "ClientInfo":
...
+def load_image_from_gcs(path: str, project: Optional[str] = None) -> "Image":
+    """Loads im Image from GCS."""
+    try:
+        from google.cloud import storage
+    except ImportError:
+        raise ImportError("Could not import google-cloud-storage python package.")
+    from vertexai.preview.generative_models import Image
+
+    gcs_client = storage.Client(project=project)
+    pieces = path.split("/")
```
holtskinner, line 120: It would probably be good to create utility functions that can split and merge GCS paths using Regex (or use one if it already exists).

There are some in Document AI Toolbox for a reference, probably doesn't make sense to import it just for these functions.

https://github.com/googleapis/python-documentai-toolbox/blob/7d30aa5000c0b42da45b9de7edf6ee75362f8e5f/google/cloud/documentai_toolbox/utilities/gcs_utilities.py#L152

hinthornw (approved):
Thanks for this! I've gotta keep the export so I'll add one commit in a separate PR

PR #14566, opened 2023-12-11
infra: import CI speed

efriis:
Was taking 10 mins. Now a few seconds.


libs/community/Makefile
```diff
@@ -39,7 +37,7 @@ lint_tests: MYPY_CACHE=.mypy_cache_test
 
 lint lint_diff lint_package lint_tests:
 	./scripts/check_pydantic.sh .
-	./scripts/check_imports.sh
+	./scripts/lint_imports.sh
```
baskaryan, line 40: was check_imports.sh supposed to be renamed to lint_imports

baskaryan: oh there's duplicate scripts

baskaryan (approved):

libs/community/Makefile
efriis: Ye I deleted the old name. I renamed them last week to not be confused with the import check one

PR #14568, opened 2023-12-11
fix: to rag-semi-structured template

shauryr:
**Description:** 

Fixes to rag-semi-structured template.

- Added required libraries
- pdfminer was causing issues when installing with pip. pdfminer.six works best
- Changed the pdf name for demo from llama2 to llava


hwchase17 (approved):
thanks!

PR #14571, opened 2023-12-11
cli[patch]: integration template

efriis:
Adds a `langchain integration new --name integration-name` command to the cli.

Should be run from `libs/partners`


libs/cli/langchain_cli/integration_template/integration_template/llms.py
```diff
@@ -0,0 +1,73 @@
...
+
+from langchain_core.callbacks import (
+    AsyncCallbackManagerForLLMRun,
+    CallbackManagerForLLMRun,
+)
+from langchain_core.language_models import BaseLLM
+from langchain_core.outputs import GenerationChunk, LLMResult
+
+
+class Integration(BaseLLM):
```
leo-gan, line 19: Can we use the `<Name>LLM` name convention? If so, it would be `IntegrationLLM`

efriis: Updating! This would be moving forward, and we'll probably tolerate people naming a "default" thing. e.g. the `Chroma` or `FAISS` vector stores, or the LLMs you referenced.

leo-gan: Thanks!!!!

libs/cli/langchain_cli/integration_template/tests/integration_tests/test_chat_models.py
```diff
@@ -0,0 +1,63 @@
+"""Test Chat__ModuleName__ chat model."""
+from __module_name__.chat_models import Chat__ModuleName__
+
+
+def test_integration_stream() -> None:
```
baskaryan, line 5: ik this is my fault but should we just call this test_stream

PR #14574, opened 2023-12-12
Rename LLMonitor Callback Handler

hughcrt:
Hi,

We are renaming our product from LLMonitor to Lunary.ai.

This PR changes all references to LLMonitor throughout the code to Lunary.

Thanks!


hwchase17 (commented):
Can we do this so that it's (1) backwards compatible, (2) doesn't change langchain

Practically, this would mean keeping around LLMMonitor as an alias (`LLMonitorCallbackHandler=LunaryCallbackHandler`) and then not updating langchain

This means that to import LunaryCallbackHandler people would have to import from `langchain_community` (this is the desired way going forward)

An alternative would be to create a dedicated `langchain-lunary` integration package - lmk if interesting and we can pursue that

PR #14587, opened 2023-12-12
fix a bug in RedisNum filter againt value 0

dandanwei:
- **Description:** There is a bug in RedisNum filter that filter towards value 0 will be parsed as "*". This is a fix to it.
  - **Issue:** NA
  - **Dependencies:** NA
  - **Tag maintainer:** NA
  - **Twitter handle:** NA


baskaryan:
cc @Spartee

hwchase17 (approved):

PR #14614, opened 2023-12-12
changed default for VertexAIEmbeddings

lkuligin:
  - **Description:** @kurtisvg has raised a point that it's a good idea to have a fixed version for embeddings (since otherwise a user might run a query with one version vs a vectorstore where another version was used). 
  In order to avoid breaking changes, I'd suggest to give users a warning, and make a `model_name` a required argument in 1.5 months.

### DIFF
--- a/libs/community/langchain_community/embeddings/vertexai.py
+++ b/libs/community/langchain_community/embeddings/vertexai.py
@@ -29,21 +29,28 @@ class VertexAIEmbeddings(_VertexAICommon, Embeddings):
     def validate_environment(cls, values: Dict) -> Dict:
         """Validates that the python package exists in environment."""
         cls._try_init_vertexai(values)
+        if values["model_name"] == "textembedding-gecko-default":
+            logger.warning(
+                "Model_name will become a required arg for VertexAIEmbeddings "
+                "starting from Feb-01-2024. Currently the default is set to "
+                "textembedding-gecko@001"
+            )
+            values["model_name"] = "textembedding-gecko@001"
         try:
             from vertexai.language_models import TextEmbeddingModel
-
-            values["client"] = TextEmbeddingModel.from_pretrained(values["model_name"])
         except ImportError:
             raise_vertex_import_error()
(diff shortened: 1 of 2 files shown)


kurtisvg (approved):

libs/community/langchain_community/embeddings/vertexai.py
```diff
@@ -20,6 +24,13 @@ def validate_environment(cls, values: Dict) -> Dict:
             from vertexai.language_models import TextEmbeddingModel
         except ImportError:
             raise_vertex_import_error()
+        if values["model_name"] == "textembedding-gecko-default":
+            logger.warning(
+                "Model_name will become a required arg for VertexAIEmbeddings "
+                "starting from Feb-01-2024. Currently the default is set to "
+                "textembedding-gecko@001"
+            )
+            values["model_name"] = "textembedding-gecko@001"
```
baskaryan, line 33: what's the difference between textembedding-gecko and textembedding-gecko@001? does the former track latest version

lkuligin: `textembedding-gecko@001` points to the specific version while `textembedding-gecko` to the default one that can be changed.

baskaryan: if google is exposing a default version in the API is it wrong to default to that here, too?

kurtisvg: The Embeddings API actually uses the model name as a path parameter, which makes it a required input: https://cloud.google.com/vertex-ai/docs/generative-ai/embeddings/get-text-embeddings#supported_models

We want to make the model name required input here as well, so we are starting with this small change to pin the version and warn of the upcoming break. Pinning the version is also consistent with the best practices called out in that documentation and how OpenAIs model number is pinned

vladkol:
+1 to this. Pinning a version of the model make results predictable and repeatable.

baskaryan:
makes sense, thank you all!

PR #14629, opened 2023-12-12
core[patch]: Fix runnable with message history

baskaryan:
Fix bug shown in #14458. Namely, that saving inputs to history fails when the input to base runnable is a list of messages

### DIFF
--- a/libs/core/langchain_core/runnables/history.py
+++ b/libs/core/langchain_core/runnables/history.py
@@ -1,13 +1,17 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
+    Dict,
     List,
     Optional,
     Sequence,
     Type,
+    Union,
 )
 
 from langchain_core.chat_history import BaseChatMessageHistory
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/runnables/history.py
```diff
@@ -350,6 +343,12 @@ def _exit_history(self, run: Run, config: RunnableConfig) -> None:
         input_val = inputs[self.input_messages_key or "input"]
         input_messages = self._get_input_messages(input_val)
 
+        # If historic messages were prepended to the input messages, remove them to
+        # avoid adding duplicate messages to history.
+        if not self.history_messages_key:
```
baskaryan, line 348: functional change

nfcampos (approved):

libs/core/langchain_core/runnables/history.py
```diff
@@ -350,6 +343,12 @@ def _exit_history(self, run: Run, config: RunnableConfig) -> None:
...
+            historic_messages = config["configurable"]["message_history"].messages
```
eyurtsev, line 349: Is this an extra network request?

eyurtsev (approved):

libs/core/langchain_core/runnables/history.py
baskaryan: yea. you're thinking we could store the messages in the config when we initially pull them in _enter_history?

eyurtsev: If possible or maybe there's some other way to do this?

I think we should merge as is for now for correctness, and proceed with optimization at some point later if possible.

We also have multiple network request with add_message which isn't amazing, but not end of world here since it's low throughput

baskaryan: will land bug fix for now and follow up with improvements

PR #14630, opened 2023-12-12
infra: skip extended testing for partner packages

efriis:
Tested by merging into #14627


.github/scripts/check_diff.py
```diff
@@ -23,7 +23,7 @@
                 ".github/scripts/check_diff.py",
             )
         ):
-            dirs_to_run = ALL_DIRS
+            dirs_to_run.update(ALL_DIRS)
```
baskaryan, line 26: probably shouldn't call this ALL_DIRS if it's not longer all dirs

also should it still be breaking?

PR #14636, opened 2023-12-13
Milvus allows to store metadata as json field

mokeyish:
Because Milvus doesn't support nullable fields, but document metadata is very rich, so it makes more sense to store it as json.

https://github.com/milvus-io/pymilvus/issues/1705#issuecomment-1731112372


fzliu:
@mokeyish thanks for the PR! lgtm

hwchase17 (approved):

PR #14648, opened 2023-12-13
[Integration] NVIDIA AI Playground

hinthornw:
Tried to simplify this pr as much as possible (https://github.com/langchain-ai/langchain/pull/13996/files). The base class is still more complicated than it needs to be, but I think it's OK as an integration package.


VKudlay:
@hinthornw Heyo! Checked over it. Honestly, looks great! If it helps streamline things and you've already done extensive testing, I'm personally happy with it. Let me toss it out to a few people and see if anybody else can put an eye on it. I'm still a bit confused about the partner stuff (would love some clarification so we can circulate internally), but it seems to be the new norm so ok.

I'm gonna run some tests and integrate the current version into my workflows :D

hinthornw:
@VKudlay we're still working on better communicating it, but basically to better maintain the growing ecosystem, we're adding new integrations as separate packages that all depend on `langchain_core`, that way everything is compatible, and we'll still track it in the monorepo, but we can then improve the experience for communicating and managing dependencies, etc.

VKudlay:
Sounds good to me! Yeah, I read [the blogpost](https://blog.langchain.dev/the-new-langchain-architecture-langchain-core-v0-1-langchain-community-and-a-path-to-langchain-v0-1/) and can more-or-less see the vision.

It looks like at least one of our teams has already tried the new stuff, so looks like the changes are cool! I'm still gonna try it out as-is, but is it all good to go in its current state when the relevant people give the ok? Or are there any outstanding issues or things you want a second pair of eyes over? :D

hinthornw:
I think it's pretty good - want one more pair of eyes to review it before landing and cutting a release.

VKudlay:
@hinthornw Awesome! I'll be sure to communicate that. And it works great so far! Let me also say, that notebook update is awesome! I feel like you pulled in all the right examples and even tossed in the new models 😄

hinthornw:
https://pypi.org/project/langchain-nvidia-aiplay/

libs/partners/nvidia-aiplay/langchain_nvidia_aiplay/_common.py
```diff
@@ -0,0 +1,706 @@
...
+            # it's a Chesterton's fence I'm unwilling to touch
+            return dict(role="user", content=msg)
+        if isinstance(msg, dict):
+            if msg.get("content", None) is None:
+                raise ValueError(f"Message {msg} has no content")
+            return msg
+        raise ValueError(f"Unknown message received: {msg} of type {type(msg)}")
+
+
+class NVAIPlayBaseModel(_NVAIPlayClient):
```
baskaryan, line 579: why implement here and not on chat model class?

hinthornw: agreed

PR #14655, opened 2023-12-13
Add Support for direct usage of llm LLamaCpp in LlamaCppEmbeddings

bfortunato:
- **Description:** Add support for using a llm LLamaCpp directly by introducing a new parameter, `model`, to the `LlamaCppEmbeddings` constructor. This allows users to provide a llm LLamaCpp instance directly, in addition to the existing method of specifying the model path.
- **Issue:** N/A
- **Dependencies:** N/A
- **Tag maintainer:** @bfortunato

### DIFF
--- a/libs/community/langchain_community/embeddings/llamacpp.py
+++ b/libs/community/langchain_community/embeddings/llamacpp.py
@@ -2,6 +2,7 @@ from typing import Any, Dict, List, Optional
 
 from langchain_core.embeddings import Embeddings
 from langchain_core.pydantic_v1 import BaseModel, Extra, Field, root_validator
+from langchain.llms.llamacpp import LlamaCpp
 
 
 class LlamaCppEmbeddings(BaseModel, Embeddings):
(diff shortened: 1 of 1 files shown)


daniilpastukhov:
Any update on this please?

bfortunato:
I don’t know. I’m available if I can do something to go ahead.

baskaryan (commented):
could you already initialize as
```python
model = LlamaCpp(...)
embeddings = LlamaCppEmbeddings(client=model.client, model_path=model.model_path)
```
in which case main improvement would be to update validate_environment to check if a client was already passed in, so that it doesn't reload the client for no reason?

baskaryan:
> could you already initialize as
> 
> ```python
> model = LlamaCpp(...)
> embeddings = LlamaCppEmbeddings(client=model.client, model_path=model.model_path)
> ```
> 
> in which case main improvement would be to update validate_environment to check if a client was already passed in, so that it doesn't reload the client for no reason?

closing per ^, let me know if you disagree

PR #14675, opened 2023-12-13
docs: build partner api refs

baskaryan:


docs/api_reference/requirements.txt
```diff
@@ -2,6 +2,7 @@
 -e libs/langchain
 -e libs/core
 -e libs/community
+-e libs/partners/google-genai
```
efriis, line 5: could also do this as a separate CI step that lists partner packages to avoid maintaining this

efriis: nvm this is for sphinx

efriis (approved):

PR #14680, opened 2023-12-13
remove unused imports

hwchase17:


libs/langchain/langchain/chat_models/litellm.py
```diff
@@ -1,17 +1,7 @@
 from langchain_community.chat_models.litellm import (
     ChatLiteLLM,
-    ChatLiteLLMException,
```
hinthornw, line 3: We're pretty sure no one is using this in things like `with_retries` right?

hellonlp:
In the embedding module, I recommend using [hellonlp/promcse](https://huggingface.co/hellonlp/promcse-bert-base-zh), which is better than bge.

PR #14684, opened 2023-12-13
docs[patch]: fix bullet points

efriis:
- docs fixes
- escape
- bullets


docs/docs/expression_language/how_to/functions.ipynb
efriis, line 1: unrelated fix - should be a `raw` cell not a `markdown` cell

PR #14685, opened 2023-12-13
infra: individual template CI

efriis:
don't merge yet - requires changes to every template.


libs/cli/langchain_cli/package_template/pyproject.toml
```diff
@@ -7,11 +7,46 @@ readme = "README.md"
...
+[tool.poetry.group.test]
+optional = true
+
+[tool.poetry.group.test.dependencies]
+pytest = "^7.3.0"
+freezegun = "^1.2.2"
+pytest-mock  = "^3.10.0"
+syrupy = "^4.0.2"
+pytest-watcher = "^0.3.4"
+pytest-asyncio = "^0.21.1"
```
baskaryan, line 21: do we need all these?

libs/cli/langchain_cli/package_template/pyproject.toml
```diff
@@ -7,11 +7,46 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
-langchain = ">=0.0.313, <0.1"
-openai = "^0.28.1"
+langchain = ">=0.0.350, <0.2"
```
baskaryan, line 10: why 0.2 instead of 0.1

libs/cli/langchain_cli/package_template/pyproject.toml
efriis: We don't even need the test group yet but would love to avoid a second round of editing if we add in coming weeks to match standard unit tests

libs/cli/langchain_cli/package_template/pyproject.toml
efriis: Given current plan thought it made sense to just update now but happy to adjust

efriis: Reverting

libs/cli/langchain_cli/package_template/pyproject.toml
efriis: Removing

efriis:
Gonna do this in a way that doesnt require any changes to any templates (just CI) to keep it simpler

PR #14688, opened 2023-12-13
infra: add integration test workflow

baskaryan:


.github/workflows/_integration_test.yml
```diff
@@ -0,0 +1,57 @@
...
+      run:
+        working-directory: ${{ inputs.working-directory }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version:
+          - "3.8"
+          - "3.9"
+          - "3.10"
+          - "3.11"
```
efriis, line 25: maybe just 3.8 and 3.11? We already hit rate limits on some providers

efriis (commented):
should this instead be a rewrite of `scheduled_test.yml`? I think setting up some of the integration creds is more involved than env vars (e.g. gcloud, aws)

baskaryan:
> should this instead be a rewrite of `scheduled_test.yml`? I think setting up some of the integration creds is more involved than env vars (e.g. gcloud, aws)

this'll be partners package specific (and atm only partners package is google-genai which just needs api key)

but i did forget to inherit secrets in general, will add!

efriis (approved):

PR #14689, opened 2023-12-13
python-lint

joshuasundance-swca:
# Description: _python-lint_

This agent writes Python code that is formatted and linted using `black`, `ruff`, and `mypy`, but does not execute the code. It writes the code to a temporary file and then runs the linters. Once these checks pass, the code is returned.

# Dependencies

- black
- ruff
- mypy

# Demo

The functionality can be seen here: https://huggingface.co/spaces/joshuasundance/langchain-streamlit-demo


templates/python-lint/README.md
```diff
@@ -0,0 +1,69 @@
+# python-lint
+
+This agent writes Python code that is formatted and linted using `black`, `ruff`, and `mypy`. 
+
+It does not execute any code.
```
jamesbraza, line 5: Hi @joshuasundance-swca this is really cool! Nice work 👍 

One comment is, would you be willing to add a small blurb of the difference with:
- Using this agent
- Writing code then running the tools

I guess I am wondering your thoughts on when to use this agent vs when to just write code and use the tools from a command line

joshuasundance-swca: Thanks @jamesbraza! :) 

I think this agent will be useful for code generation tasks. Since the agent is instructed to run checks before submitting the finished code, and has an opportunity to make changes in response to failed checks, the code should end up being more consistent and reliable. When I write code, I use these tools in `pre-commit`. Now, when I ask the llm to write code, I expect it to pass the same checks. :)

The reason this agent doesn't actually run the code is because that may require greater security considerations and because it can be handled downstream.

jamesbraza: Okay gotchu, thanks and I agree with your thought process. Cheers!

hwchase17 (approved):
this is awesome! thanks!

PR #14691, opened 2023-12-14
cli[patch]: integration template nits

baskaryan:


libs/cli/langchain_cli/integration_template/Makefile
```diff
@@ -5,13 +5,16 @@ all: help
...
+INTEGRATION_TEST_FILE ?= tests/integration_tests/
 
 test:
 	poetry run pytest $(TEST_FILE)
 
 tests:
 	poetry run pytest $(TEST_FILE)
 
+integration_tests:
+	poetry run pytest $(INTEGRATION_TEST_FILE)
```
efriis, line 17: Given these are new, could we
- get rid of `tests` (duplicate of `test`)
- keep one var but set it to the right path
```suggestion
integration_test: TEST_FILE = tests/integration_tests/

test integration_test:
	poetry run pytest $(TEST_FILE)

integration_tests:
	poetry run pytest $(INTEGRATION_TEST_FILE)
```

libs/cli/langchain_cli/integration_template/Makefile
```diff
@@ -30,7 +33,7 @@ lint lint_diff lint_package lint_tests:
 	poetry run ruff .
 	poetry run ruff format $(PYTHON_FILES) --diff
 	poetry run ruff --select I $(PYTHON_FILES)
-	mkdir $(MYPY_CACHE); poetry run mypy $(PYTHON_FILES) --cache-dir $(MYPY_CACHE)
+	mkdir -p $(MYPY_CACHE); poetry run mypy $(PYTHON_FILES) --cache-dir $(MYPY_CACHE)
```
efriis, line 36:
```suggestion
	poetry run mypy $(PYTHON_FILES) --cache-dir $(MYPY_CACHE)
```
I'm fairly certain this is supposed to work

libs/cli/langchain_cli/integration_template/integration_template/chat_models.py
```diff
@@ -14,50 +15,52 @@ class Chat__ModuleName__(BaseChatModel):
...
-        stop: Optional[List[str]] = None,
-        run_manager: Optional[AsyncCallbackManagerForLLMRun] = None,
-        **kwargs: Any,
+            self,
+            messages: List[BaseMessage],
+            stop: Optional[List[str]] = None,
+            run_manager: Optional[AsyncCallbackManagerForLLMRun] = None,
+            **kwargs: Any,
     ) -> AsyncIterator[ChatGenerationChunk]:
-        yield ChatGenerationChunk(
```
efriis, line 45: I think removing these will cause mypy to fail. Also feels like a good template to help people mimic?

libs/cli/langchain_cli/integration_template/integration_template/chat_models.py
```diff
@@ -14,50 +15,52 @@ class Chat__ModuleName__(BaseChatModel):
...
-    def _generate(
-        self,
-        messages: List[BaseMessage],
-        stop: Optional[List[str]] = None,
-        run_manager: Optional[CallbackManagerForLLMRun] = None,
-        **kwargs: Any,
-    ) -> ChatResult:
         raise NotImplementedError
 
+    # TODO: Implement if __model_name__ supports async generation. Otherwise delete
```
efriis, line 62: I like this - even if we have the `yield` stub

libs/cli/langchain_cli/integration_template/integration_template/llms.py
```diff
@@ -62,12 +62,13 @@ def _stream(
...
+    # TODO: Implement if __model_name__ supports async streaming. Otherwise delete
+    # method.
     async def _astream(
         self,
         prompt: str,
         stop: Optional[List[str]] = None,
         run_manager: Optional[AsyncCallbackManagerForLLMRun] = None,
         **kwargs: Any,
     ) -> AsyncIterator[GenerationChunk]:
-        yield GenerationChunk(text="Yield chunks")
```
efriis, line 72: same- I think this will cause mypy to fail

libs/cli/langchain_cli/integration_template/pyproject.toml
```diff
@@ -7,25 +7,21 @@ readme = "README.md"
...
+langchain-core = "^0.1"
 
 [tool.poetry.group.test]
 optional = true
 
 [tool.poetry.group.test.dependencies]
-pytest = "^7.3.0"
-freezegun = "^1.2.2"
-pytest-mock  = "^3.10.0"
-syrupy = "^4.0.2"
```
efriis, line 19: I think we need `syrupy` at a minimum for snapshot flag. Others maybe not?

libs/cli/langchain_cli/integration_template/pyproject.toml
```diff
@@ -36,10 +32,10 @@ optional = true
 optional = true
 
 [tool.poetry.group.lint.dependencies]
-ruff = "^0.1.5"
+ruff = "^0.1.8"
 
 [tool.poetry.group.typing.dependencies]
-mypy = "^0.991"
+mypy = "^1.7.1"
```
efriis, line 38: ooo maybe the yield isn't necessary anymore with this

libs/cli/langchain_cli/integration_template/integration_template/llms.py
efriis: mypy has fixed this

libs/cli/langchain_cli/integration_template/integration_template/chat_models.py
efriis: Thoughts on template to mimic?

libs/cli/langchain_cli/integration_template/Makefile
baskaryan: how's about 
```
integration_test integration_tests: TEST_FILE = tests/integration_tests/

test tests integration_test integration_tests:
    poetry run pytest $(TEST_FILE)
```

libs/cli/langchain_cli/integration_template/integration_template/chat_models.py
baskaryan: but the actual implementation will only yield in for loop, this could confuse?

libs/cli/langchain_cli/integration_template/Makefile
efriis: Works for me. Just curious why the plural plurality

libs/cli/langchain_cli/integration_template/pyproject.toml
baskaryan: lets remove snapshot flag if not used anywhere

libs/cli/langchain_cli/integration_template/integration_template/chat_models.py
efriis: Could switch to 
```
for chunk in ["Yield chunks", " like this!"]:
  yield ChatGenerationChunk(...)
```

libs/cli/langchain_cli/integration_template/Makefile
baskaryan: we've got one CI that uses test and one that uses integration_tests and having one singular and one plural seems worse than supporting both for both

libs/cli/langchain_cli/integration_template/integration_template/chat_models.py
baskaryan: why have fake implementation for this method and not others

efriis (approved):

PR #14692, opened 2023-12-14
infra: use latest ruff and mypy in core

baskaryan:


Undertone0809:
Does it seem like it has changed the mypy rules?

libs/core/langchain_core/tracers/base.py
```diff
@@ -336,7 +336,7 @@ def on_tool_start(
             child_runs=[],
             run_type="tool",
             tags=tags or [],
-            name=name,
+            name=name,  # type: ignore[arg-type]  # TODO: fix typing
```
eyurtsev, line 339: due to None?

libs/core/langchain_core/tracers/langchain_v1.py
```diff
@@ -60,29 +61,30 @@ def _convert_to_v1_run(self, run: Run) -> Union[LLMRun, ChainRun, ToolRun]:
             else:
                 raise ValueError("No prompts found in LLM run inputs")
             return LLMRun(
-                uuid=str(run.id) if run.id else None,
+                uuid=str(run.id) if run.id else "",
```
eyurtsev, line 64: Is this safe? Is downstream code ever assuming this is a None?

libs/core/langchain_core/tracers/base.py
baskaryan: yep

libs/core/langchain_core/tracers/langchain_v1.py
baskaryan: yea should maybe type ignore instead. cc @hinthornw any idea if None is expected downstream?

hinthornw: We should just delete this file entirely. It's not used anymore. @agola11 had a PR out to remove at some point that must have never been merged

PR #14720, opened 2023-12-14
[Hub|tracing] Tag hub prompts

hinthornw:
If you're using the hub, you'll likely be interested in tracking the commit/object when tracing. This PR adds it to the config


efriis (approved):
A unit test could be good for this one

PR #14729, opened 2023-12-14
proposal

efriis:


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -440,47 +440,29 @@ async def agenerate(
...
-        output = LLMResult(generations=generations, llm_output=llm_output)
-        await asyncio.gather(
-            *[
-                run_manager.on_llm_end(flattened_output)
-                for run_manager, flattened_output in zip(
-                    run_managers, flattened_outputs
+        # report results and errors
+        if run_managers:
+            jobs = [
+                run_manager.on_llm_error(res, response=LLMResult(generations=[]))
```
baskaryan, line 446: should be await i believe

efriis: I think because we await on the gather below this allows it to parallelize

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -440,47 +440,29 @@ async def agenerate(
...
-        generations = [res.generations for res in cast(List[ChatResult], results)]
-        output = LLMResult(generations=generations, llm_output=llm_output)
-        await asyncio.gather(
-            *[
-                run_manager.on_llm_end(flattened_output)
-                for run_manager, flattened_output in zip(
-                    run_managers, flattened_outputs
+        # report results and errors
+        if run_managers:
+            jobs = [
```
efriis, line 445: confirm with @nfcampos that reporting results and exceptions together like this isn't a breaking change

efriis:
Got 👍 from nuno - just need to implement for sync too in upstream pr

PR #13976, opened 2023-11-28
Nvidia Triton + TRT-LLM Connector

jdye64:
**Description**:
This PR includes a connector for using the [Nvidia TensorRT LLM](https://developer.nvidia.com/tensorrt) which allows for accelerated LLM optimization and performance.

**Issue**: This closes #13975 

**Dependencies**:
triton-client library was added to pyproject.toml

PR #14733, opened 2023-12-14
[Partner] NVIDIA TRT Package

hinthornw:
Simplify #13976 and add as a separate package.

@jdye64 I've got some questions and need your help to validate that this works. Will add some comments below.

- [] Add README
- [X] Add doc notebook
- [X] Add simple LLM integration


libs/partners/nvidia-trt/tests/integration_tests/test_llms.py
```diff
@@ -0,0 +1,64 @@
+"""Test TritonTensorRTLLM llm."""
+# TODO: These will not run on CPU...
+from langchain_nvidia_trt.llms import TritonTensorRTLLM
+
+
+def test_stream() -> None:
```
hinthornw, line 6: I'll have to add skip marks here since we can't run..

libs/partners/nvidia-trt/langchain_nvidia_trt/llms.py
```diff
@@ -0,0 +1,405 @@
...
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> LLMResult:
+        invocation_params = self._get_invocation_params(**kwargs)
+        # (WFH) TODO: It looks like we were doing this already but...?
+        self._load_model(self.model_name)
+        # stop_words = stop if stop is not None else self.stop
+        generations = []
+        # TODO: We should handle the native batching instead.
```
hinthornw, line 153: @jdye64, I'm mainly following your structure here: I'm assuming TRT natively handles batching. Got any pointers?

libs/partners/nvidia-trt/langchain_nvidia_trt/llms.py
```diff
@@ -0,0 +1,405 @@
...
+        for prompt in prompts:
+            invoc_params = {**invocation_params, "prompt": [[prompt]]}
+            # request_id = str(random.randint(1, 9999999))  # nosec
```
hinthornw, line 156: @jdye64 it doesn't look like `client.infer` requires/accepts a request_id. Could you confirm that's only used for streaming?

libs/partners/nvidia-trt/langchain_nvidia_trt/llms.py
```diff
@@ -0,0 +1,405 @@
...
+        self,
+        prompts: List[str],
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> LLMResult:
+        invocation_params = self._get_invocation_params(**kwargs)
+        # (WFH) TODO: It looks like we were doing this already but...?
+        self._load_model(self.model_name)
+        # stop_words = stop if stop is not None else self.stop
```
hinthornw, line 151: For non-streaming invocations, I'm assuming we'll just strip off the suffix manually based on the stop words, but could you confirm this?

libs/partners/nvidia-trt/langchain_nvidia_trt/llms.py
```diff
@@ -0,0 +1,405 @@
...
+            self._prepare_tensor("beam_width", beam_width_array),
+            self._prepare_tensor("stream", streaming_data),
+        ]
+        return inputs
+
+    def _trim_batch_response(self, result_str: str) -> str:
+        """Trim batch response by removing prompt and extra generated text."""
+        # extract the generated part of the prompt
+        # TODO: This assumes llama-style prompting...
+        split = result_str.split("[/INST]", 1)
```
hinthornw, line 289: @jdye64 , this is copied over from the other PR. It assumes llama-style prompting, which seems like a bad assumption. Could you confirm it's OK for me instead to remove the input prefix? Or will that be mangled in the response / unreliable?

I'm thinking basiclaly
```
self. _trim_batch_response(result_str, prompt_str)
...
```

libs/partners/nvidia-trt/langchain_nvidia_trt/llms.py
```diff
@@ -0,0 +1,404 @@
...
+        if error:
+            result_queue.put(error)
+        else:
+            response_raw: dict = result.get_response(as_json=True)
+            # TODO: Check the response is a map rather than a string
+            if "outputs" in response_raw:
+                # the very last response might have no output, just the final flag
+                response = self._process_result(response_raw)
+                if force_batch:
+                    response = self._trim_batch_response(response)
```
hinthornw, line 355: For the streaming case, is it returning only new tokens? Or is it returning the full sequence on each iteration? If the former, this trim method will break (it looks for the input prompt/split token)

libs/partners/nvidia-trt/langchain_nvidia_trt/llms.py
hinthornw: @jdye64 a followup:  if we're expecting this type of chat templating, should(n't) we be creating a chat model rather than an LLM interface? That way we can handle the structure on behalf of the user, rather than expecting them to do their own string formatting. That would also make it a lot more interoperable with the rest of langchain

libs/partners/nvidia-trt/langchain_nvidia_trt/llms.py
jdye64: Correct. The `request_id` is used to manage GRPC streams on the Triton server end so certain API calls require it. I have refactored and removed the call to `client.infer` however since we are always using the GRPC client and not HTTP client now it is safer to just always use a stream and batch locally if needed. This is required because certain models and configurations on the Triton GRCP server side might cause calls to just `infer` to fail and instead need `async_stream_infer` calls.

hinthornw: helpful - thanks!

PR #14734, opened 2023-12-14
update VoyageEmbeddings to support larger batch size

thomas0809:
- **Description:** update VoyageEmbeddings to support larger batch size
  - **Issue:** 
  - **Dependencies:** require `pip install voyageai` to use VoyageEmbeddings
  - **Tag maintainer:** 
  - **Twitter handle:**

### DIFF
--- a/libs/partners/google-genai/langchain_google_genai/chat_models.py
+++ b/libs/partners/google-genai/langchain_google_genai/chat_models.py
@@ -21,6 +21,8 @@ from typing import (
 )
 from urllib.parse import urlparse
 
+import google.api_core
+
 # TODO: remove ignore once the google package is published with types
 import google.generativeai as genai  # type: ignore[import]
 import requests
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/embeddings/voyageai.py
```diff
@@ -1,63 +1,20 @@
...
-    stop_after_attempt,
-    wait_exponential,
-)
+from langchain_core.pydantic_v1 import BaseModel, Extra, root_validator
+from langchain_core.utils import get_from_dict_or_env
 
 logger = logging.getLogger(__name__)
 
 
-def _create_retry_decorator(embeddings: VoyageEmbeddings) -> Callable[[Any], Any]:
```
hinthornw, line 30: We unfortunately right now have to keep these values exported due to the dep between community and langchain. Could set as `None` for now. We'll go through and remove soon

thomas0809: I removed the dependencies in libs/langchain/langchain/embeddings/voyageai.py. Are those what you meant?

hinthornw: since these are different packages, there will be a couple versions of langchain that still try and fail to import from the community page

PR #14736, opened 2023-12-14
Jacob/add hf chat wrapper

jacoblee93:
Builds on #14040 with community refactor merged and notebook updated.

Note that with this refactor, models will be imported from `langchain_community.chat_models.huggingface` rather than the main `langchain` repo.

CC @baskaryan @andrewrreed


A-Roucher:
Thanks a lot @jacoblee93 and @baskaryan, very excited about merging this!
Out of curiosity, why do you prefer to import models from `langchain_community.chat_models.huggingface` rather than `langchain.llms` ?

jacoblee93:
We have moved all community exports to be from the integration-focused `langchain_community` in order to make the main `langchain` package more focused.

You can read a bit more about it here:

https://github.com/langchain-ai/langchain#-what-is-langchain

It may make sense to split out into a separate `langchain-huggingface` package in the future, but that's out of scope in this PR!

PR #14740, opened 2023-12-15
Implement similarity_score_threshold for MongoDB Vector Store

NoahStapp:
Adds the option for `similarity_score_threshold` when using `MongoDBAtlasVectorSearch` as a vector store retriever.

Example use:

```
vector_search = MongoDBAtlasVectorSearch.from_documents(...)

qa_retriever = vector_search.as_retriever(
    search_type="similarity_score_threshold",
    search_kwargs={
        "score_threshold": 0.5,
    }
)

qa = RetrievalQA.from_chain_type(
	llm=OpenAI(), 
	chain_type="stuff", 
	retriever=qa_retriever,
)

docs = qa({"query": "..."})
```

I've tested this feature locally, using a MongoDB Atlas Cluster with a vector search index.

### DIFF
--- a/libs/community/langchain_community/vectorstores/mongodb_atlas.py
+++ b/libs/community/langchain_community/vectorstores/mongodb_atlas.py
@@ -11,7 +11,8 @@ from typing import (
     Optional,
     Tuple,
     TypeVar,
-    Union, Callable,
+    Union,
+    Callable,
 )
 
 import numpy as np
@@ -60,7 +61,7 @@ class MongoDBAtlasVectorSearch(VectorStore):
         index_name: str = "default",
         text_key: str = "text",
         embedding_key: str = "embedding",
-        relevance_score_fn: str = "cosine"
+        relevance_score_fn: str = "cosine",
     ):
         """
         Args:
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/mongodb_atlas.py
```diff
@@ -198,7 +215,6 @@ def _similarity_search_with_score(
     def similarity_search_with_score(
         self,
         query: str,
-        *,
```
baskaryan, line 201: why remove

NoahStapp: `VectorStore.similarity_search_with_relevance_scores` calls `self.similarity_search_with_score(query, k, **kwargs)`, which does not explicitly pass `k` as a keyword argument. This fails since the old implementation of `MongoDBAtlasVectorSearch` explicitly requires `k` to be a keyword argument. Removing this requirement makes this method consistent in signature with other vector store implementations.

baskaryan (approved):

Vishnu-add:
similarity_score_threshold isn't working for MongoDB

PR #14742, opened 2023-12-15
infra: fix pre-release integration test and add unit test

baskaryan:


.github/workflows/_release.yml
```diff
@@ -152,6 +152,10 @@ jobs:
             --extra-index-url https://test.pypi.org/simple/ \
             "$PKG_NAME==$VERSION"
 
+      - name: Run unit tests
```
efriis, line 155: Won't this need some of the poetry test deps?

baskaryan: those already installed in L140

PR #14746, opened 2023-12-15
Fix OAI Tool Message

hinthornw:
See format here: https://platform.openai.com/docs/guides/function-calling/parallel-function-calling

It expects a "name" argument, which we aren't providing by default.

image

Alternative is to add the 'name' field directly to the message if people prefer.


libs/langchain/langchain/agents/format_scratchpad/openai_tools.py
```diff
@@ -31,6 +31,7 @@ def _create_tool_message(
     return ToolMessage(
         tool_call_id=agent_action.tool_call_id,
         content=content,
+        additional_kwargs={"name": agent_action.tool},
```
hinthornw, line 34: This is the main fix

eyurtsev (approved):

PR #14759, opened 2023-12-15
[FEATURE] Add Aphrodite Engine support

AlpinDale:
This PR adds support for PygmalionAI's [Aphrodite Engine](https://github.com/PygmalionAI/aphrodite-engine), based on vLLM's attention mechanism. At the moment, this PR does not include support for the API servers, but they will be added in a later PR.

The only dependency as of now is `aphrodite-engine==0.4.2`. We pin the version to prevent breakage due to changes in the aphrodite-engine library.


baskaryan:
looks great, thanks @AlpinDale!

Issue #14764, opened 2023-12-15
KeyError 'role' in OpenAI Adapter


Sypherd:
### System Info

Python 3.10
LangChain 0.0.348

### Who can help?

@hwchase17 @hin

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [X] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

```
from langchain.chat_models import ChatOpenAI
from langchain.memory import ChatMessageHistory

history = ChatMessageHistory()
history.add_user_message("Tell me anything in ten words or less.")

params = {
    "model": "gpt-4-1106-preview",
    "temperature": 0.5,
    "max_tokens": 1000,
}
llm = ChatOpenAI(**params)
print(llm(history.messages))
```
This will **occasionally** yield the following error:
```
File "/.../.venv/lib/python3.10/site-packages/langchain/adapters/openai.py", line 74, in convert_dict_to_message
    role = _dict["role"]
KeyError: 'role'
```

### Expected behavior

* Safe lookups ensure the response is passed back even if the role isn't attached

Sypherd:
Here's a PR to close this issue: https://github.com/langchain-ai/langchain/pull/14765/

Sypherd:
I believe the above fix resolved this issue.

PR #14765, opened 2023-12-15
Add safe lookup to OpenAI response adapter

Sypherd:
## Description
Similar to https://github.com/langchain-ai/langchain/issues/5861, I've experienced `KeyError`s resulting from unsafe lookups in the `convert_dict_to_message` function in [this file](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/adapters/openai.py). While that issue focused on `KeyError 'content'`, I've opened another issue (#14764) about how the problem still exists in the same function but with `KeyError 'role'`. The fix for #5861 only added a safe lookup to the specific line that was giving them trouble.. This PR fixes the unsafe lookup in the rest of the function but the problem still exists across the repo.

## Issues
* #14764
* #5861 

## Dependencies
* None

## Checklist
[x] make format
[x] make lint
[ ] make test - Results in `make: *** No rule to make target 'test'.  Stop.`

## Maintainers
* @hinthornw

### DIFF
--- a/libs/community/langchain_community/adapters/openai.py
+++ b/libs/community/langchain_community/adapters/openai.py
@@ -71,27 +71,30 @@ def convert_dict_to_message(_dict: Mapping[str, Any]) -> BaseMessage:
     Returns:
         The LangChain message.
     """
-    role = _dict["role"]
+    role = _dict.get("role")
     if role == "user":
-        return HumanMessage(content=_dict["content"])
+        return HumanMessage(content=_dict.get("content"))
     elif role == "assistant":
         # Fix for azure
         # Also OpenAI returns None for tool invocations
-        content = _dict.get("content", "") or ""
+        content = _dict.get("content", "")
         additional_kwargs: Dict = {}
-        if _dict.get("function_call"):
-            additional_kwargs["function_call"] = dict(_dict["function_call"])
-        if _dict.get("tool_calls"):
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/adapters/openai.py
```diff
@@ -71,34 +71,34 @@ def convert_dict_to_message(_dict: Mapping[str, Any]) -> BaseMessage:
...
-        return SystemMessage(content=_dict["content"])
+        return SystemMessage(content=_dict.get("content", ""))
     elif role == "function":
-        return FunctionMessage(content=_dict["content"], name=_dict["name"])
+        return FunctionMessage(content=_dict.get("content", ""), name=_dict.get("name"))
     elif role == "tool":
         additional_kwargs = {}
         if "name" in _dict:
-            additional_kwargs["name"] = _dict["name"]
+            additional_kwargs["name"] = _dict.get("name")
```
hinthornw, line 94: this one isn't actually needed

hinthornw (approved):

libs/community/langchain_community/adapters/openai.py
Sypherd: True! Because we check if "name" is in the dict.

Sypherd:
@hinthornw is any additional action required on my part to merge?

PR #14771, opened 2023-12-15
docs `microsoft` pages sort order fix

leo-gan:
`integrations/document_loaders/` `Excel` and `OneNote` pages in the navbar were in the wrong sort order. It is because the file names are not equal to the page titles.
- renamed `excel` and `onenote` file names


baskaryan (commented):
preference for keeping docs file names / structure symmetric with package files names. if we care about order we can change docs title?

leo-gan:
> preference for keeping docs file names / structure symmetric with package files names. if we care about order we can change docs title?

We don't have this structure now:
- `microsoft_onedrive`  - OneDriveLoader
- `microsoft_word` - UnstructuredWordDocumentLoader
- `onenote` - OneNoteLoader
etc, 
I'm just trying to do some order in the ToC. That's my goal. To make the document consitent.
I'm placing most of the Office products under this naming convention `Microsoft <name>` 
Otherwise, we have those menu items for Office in one place: OneDrive, PowerPoint, SharePoint, Word. But Excel and OneNote are somewhere else in ToC and out of order.

hwchase17 (approved):

PR #14772, opened 2023-12-15
docs `alibaba cloud`

leo-gan:
The [provider page](https://python.langchain.com/docs/integrations/providers/alibabacloud_opensearch) holds the vector store information. The [Chat example](https://python.langchain.com/docs/integrations/chat/pai_eas_chat_endpoint) was incorrectly sorted in the navbar because of the wrong file name.
- Recreated a provide page
- Added missed links and descriptions
- Compound information about vector store from two pages into one
- Fixed file name


efriis (approved):

docs/vercel.json
efriis, line 1: should there also be a redirect for the provider page?

leo-gan: Great catch! Thank you!!!

efriis (approved):

PR #14786, opened 2023-12-16
docs[patch]: fix zoom

efriis:
not sure why quarto is removing divs

### DIFF
--- a/docs/src/theme/Columns.js
+++ b/docs/src/theme/Columns.js
@@ -10,7 +10,7 @@ export function ColumnContainer({children}) {
 
 export function Column({children}) {
     return (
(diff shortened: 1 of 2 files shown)


docs/src/theme/Columns.js
```diff
@@ -10,7 +10,7 @@ export function ColumnContainer({children}) {
 
 export function Column({children}) {
     return (
-        <div style={{ flex: "1 0 300px", padding: "10px", overflowX: "clip" }}>
+        <div style={{ flex: "1 0 300px", padding: "10px", overflowX: "clip", zoom: '80%' }}>
```
efriis, line 13: add it back here

baskaryan (approved):

Issue #13723, opened 2023-11-22
RunnableBranch doesnt stream correctly


MateusZ36:
### System Info

LangChain Version: 0.0.339
Platform: Windows 10
Python Version: 3.10.10

### Who can help?

@agol

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [X] Chains
- [X] Callbacks/Tracing
- [X] Async

### Reproduction

I've implemented this `RunnableBranch` example: https://python.langchain.com/docs/expression_language/how_to/routing
using this `MyCallbackHandler` example: https://python.langchain.com/docs/modules/agents/how_to/streaming_stdout_final_only

For streamming I've used this `streamming` example: https://python.langchain.com/docs/modules/model_io/chat/streaming

the result becomes something like this:
```
##
#Other#
##
(Verse 1)... <the output of the question as a whole, not separated by tokens, generated all at once>
```

### Expected behavior

every single token be surrounded by ##, like this: 
```
#token1#
#token2#
...
```

MateusZ36:
I'm using `RunnableBranch `to decide via lambda which chain to run, it is choosing correctly but it doesnt stream that chain untill the last token is generated, which just sends a wall of text to the end user instead of slowly sending it token by token, if that's the expected result, what alternative do I have for this use case?

daolmedo:
Same problem when routing runnables! Reported here too: https://github.com/langchain-ai/langchain/issues/11998

qtangs:
> Same problem when routing runnables! Reported here too: #11998

Can try this subclass `StreamingRunnableBranch` which adapts the logic for `RunnableBranch.invoke`:

```
from typing import Any, Iterator

from langchain.load import dumpd
from langchain.schema.runnable import RunnableBranch
from langchain.schema.runnable import RunnableConfig
from langchain.schema.runnable.config import ensure_config
from langchain.schema.runnable.config import get_callback_manager_for_config
from langchain.schema.runnable.config import patch_config
from langchain.schema.runnable.utils import Input
from langchain.schema.runnable.utils import Output


[...]

PR #14805, opened 2023-12-16
core(minor): Implement stream and astream for RunnableBranch

qtangs:
**Description:** Implement stream and astream methods for RunnableBranch
  - **Issue:** https://github.com/langchain-ai/langchain/issues/13723
  - **Twitter handle:** https://twitter.com/qtangs

### DIFF
--- a/libs/core/langchain_core/runnables/branch.py
+++ b/libs/core/langchain_core/runnables/branch.py
@@ -25,6 +25,7 @@ from langchain_core.runnables.base import (
 from langchain_core.runnables.config import (
     RunnableConfig,
     ensure_config,
+    get_async_callback_manager_for_config,
     get_callback_manager_for_config,
     patch_config,
 )
@@ -214,7 +215,7 @@ class RunnableBranch(RunnableSerializable[Input, Output]):
                     ),
                     **kwargs,
                 )
-        except Exception as e:
+        except BaseException as e:
             run_manager.on_chain_error(e)
             raise
         run_manager.on_chain_end(dumpd(output))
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/runnables/branch.py
```diff
@@ -264,3 +266,117 @@ async def ainvoke(
...
+                    for output in runnable.stream(
+                        input,
+                        config=patch_config(
+                            config,
+                            callbacks=run_manager.get_child(tag=f"branch:{idx + 1}"),
+                        ),
+                        **kwargs,
+                    ):
+                        yield output
+                        outputs.append(output)
```
nfcampos, line 309: we try to ensure the output is the same when called with or without streaming. to achieve that we combine streamed chunks with `+` instead of accumulating in a list

nfcampos: See here for an example https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/runnables/base.py#L2318

libs/core/langchain_core/runnables/branch.py
```diff
@@ -264,3 +266,117 @@ async def ainvoke(
...
+                    async for output in runnable.astream(
+                        input,
+                        config=patch_config(
+                            config,
+                            callbacks=run_manager.get_child(tag=f"branch:{idx + 1}"),
+                        ),
+                        **kwargs,
+                    ):
+                        yield output
+                        outputs.append(output)
```
nfcampos, line 366: Same as above

libs/core/langchain_core/runnables/branch.py
```diff
@@ -264,3 +266,117 @@ async def ainvoke(
...
+    async def astream(
+        self,
+        input: Input,
+        config: Optional[RunnableConfig] = None,
+        **kwargs: Optional[Any],
+    ) -> AsyncIterator[Output]:
+        """First evaluates the condition,
+        then delegate to true or false branch."""
+        config = ensure_config(config)
+        callback_manager = get_callback_manager_for_config(config)
```
nfcampos, line 336: this should use `get_async_callback_manager_for_config` instead (ainvoke above is wrong and needs to be fixed)

eyurtsev:
@nfcampos / @qtangs 

Made a few minor changes:

* rely on `+` for aggregation of final output -> changes expected results in unit tests also for the outputs
* changing all code to use Base Exception
* async callbacks in both ainvoke and astream

nfcampos (approved):

PR #14814, opened 2023-12-17
feat: Qdrant sparse vector retriever

Anush008:
## Description

This PR intends to add support for Qdrant's new [sparse vector retrieval](https://qdrant.tech/articles/sparse-vectors/) by introducing a new retriever class, `QdrantSparseVectorRetriever`. 

Necessary usage docs and integration tests have been added for the retriever.


baskaryan (commented):
could it make sense to add this functionality directly to the Qdrant vector store? Imagining someone could pass in a sparse encoder to the vector store

Anush008:
We did have a thought about this. But that would require a major refactor of the vectorstore implementation and would be challenging to maintain backward compatibility.
As a separate retriever, that won't be a problem and we could also support hybrid searches using the sparse and dense vectors via the [ensemble](https://python.langchain.com/docs/modules/data_connection/retrievers/ensemble) retriever.

Anush008:
We did have a thought about this. But that would require a major refactor of the vectorstore implementation and would be challenging to maintain backward compatibility.
As a separate retriever, that won't be a problem and we can support hybrid searches using the sparse and dense vectors via the [ensemble](https://python.langchain.com/docs/modules/data_connection/retrievers/ensemble) retriever.

Anush008:
We did have a thought about this. But that would require a major refactor of the vectorstore implementation and would be challenging to maintain backward compatibility.
As a separate retriever, that won't be a problem and we can support hybrid searches uaing the [ensemble](https://python.langchain.com/docs/modules/data_connection/retrievers/ensemble) retriever.

Anush008:
We did have a thought about this. But that would require a major refactor of the vectorstore implementation and would be challenging to maintain backward compatibility.
As a separate retriever, that won't be a problem and we can support hybrid searches using the [ensemble](https://python.langchain.com/docs/modules/data_connection/retrievers/ensemble) retriever.

Anush008:
We did have a thought about this. But it would require a major refactor of the vectorstore implementation and we could run into backward compatibility issues.
As a separate retriever, that won't be a problem and we can support hybrid searches using the [ensemble](https://python.langchain.com/docs/modules/data_connection/retrievers/ensemble) retriever.

hwchase17 (commented):
lgtm, except we should NOT re-expose this in langchain. langchain should not change, only langchain-community. otherwise it will cause compatibility issues

Anush008:
I've removed the re-export.

PR #14817, opened 2023-12-17
Fix failing serpapi response processing for Google Maps API

LoopKarma:
**Description:** Fix for processing for serpapi response for Google Maps API 
**Issue:** Due to the fact corresponding [api](https://serpapi.com/google-maps-api) returns 'local_results' as list, and old version requested `res["local_results"].keys()` of the list. As the result we got exception: ```AttributeError: 'list' object has no attribute 'keys'```.

Way to reproduce wrong behaviour:
```
    params = {
        "engine": "google_maps",
        "type": "search",
        "google_domain": "google.de",
        "ll": "@51.1917,10.525,14z",
        "hl": "de",
        "gl": "de",
    }
    search = SerpAPIWrapper(params=params)
    results = search.run("cafe")
```

### DIFF
--- a/libs/community/langchain_community/utilities/serpapi.py
+++ b/libs/community/langchain_community/utilities/serpapi.py
@@ -211,8 +211,8 @@ class SerpAPIWrapper(BaseModel):
 
         if "buying_guide" in res.keys():
             snippets.append(res["buying_guide"])
-        if "local_results" in res.keys() and "places" in res["local_results"].keys():
-            snippets.append(res["local_results"]["places"])
+        if "local_results" in res.keys() and isinstance(res["local_results"], list):
+            snippets += res["local_results"]
 
         if len(snippets) > 0:
             return str(snippets)
--- a/libs/community/langchain_community/utilities/serpapi.py
+++ b/libs/community/langchain_community/utilities/serpapi.py
@@ -211,7 +211,7 @@ class SerpAPIWrapper(BaseModel):
 
         if "buying_guide" in res.keys():
             snippets.append(res["buying_guide"])
-        if "local_results" in res.keys() and isinstance(res["local_results"], list):
+        if "local_results" in res and isinstance(res["local_results"], list):
             snippets += res["local_results"]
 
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/utilities/serpapi.py
```diff
@@ -211,8 +211,8 @@ def _process_response(res: dict) -> str:
 
         if "buying_guide" in res.keys():
             snippets.append(res["buying_guide"])
-        if "local_results" in res.keys() and "places" in res["local_results"].keys():
-            snippets.append(res["local_results"]["places"])
+        if "local_results" in res.keys() and isinstance(res["local_results"], list):
```
rancomp, line 214: I have two remarks:

1. Not directly related to your PR's purpose, but there's no need to invoke `.keys()`. You can just iterate over the dictionary to access the keys.

```suggestion
        if "local_results" in res and isinstance(res["local_results"], list):
```
2. IIUC, the problem was that `res["local_results"]` is indeed a `List`, and has no `.keys()` attribute. Shouldn't the if-statement still check the expression `"places" in  res["local_results"]`?

LoopKarma: Thank you for taking a look @rancomp 
1. good point about removing `.keys` - I wanted to keep it consistent and short though 😉
2. nope, there are no 'places' key in the `res["local_results"]`. We can directly return content of the `res["local_results"]`. I see this in [docs](https://serpapi.com/google-maps-api) and in real api calls

hwchase17: 2. it seems like the type of this key can differ depending on engine: https://serpapi.com/yahoo-local-pack

maybe, we could check type? if list then do this. if dict, then check for places

LoopKarma: good point, thank you @hwchase17 👍
added handling of the yahoo local pack case.

hwchase17 (approved):

PR #14826, opened 2023-12-18
Add logprobs in gen output

hinthornw:
Now that it's supported again for OAI chat models .

Shame this wouldn't include it in the `.invoke()` output though (it's not included in the message itself). Would need to do a follow-up for that to be the case


baskaryan (approved):

sbarman-mi9:
Can you please push this update? Also, it needs to be added to _stream() and _astream()? @baskaryan 

image

PR #14830, opened 2023-12-18
[Bugfix] Ensure tool output is a str, for OAI Assistant

hinthornw:
Tool outputs have to be strings apparently. Ensure they are formatted correctly before passing as intermediate steps.
 

```
BadRequestError: Error code: 400 - {'error': {'message': '1 validation error for Request\nbody -> tool_outputs -> 0 -> output\n  str type expected (type=type_error.str)', 'type': 'invalid_request_error', 'param': None, 'code': None}}
```


baskaryan (approved):

baskaryan:
maybe worth raising warning if outputs aren't a string?

hinthornw:
Hmmm I think it's ok if not?

PR #14838, opened 2023-12-18
added jaguar vector store

fserv:
Description: A new vector store Jaguar is being added. Class, test scripts, and documentation is added.
Issue: None -- This is the first PR contributing to LangChain
Dependencies: This depends on "pip install -U jaguardb-http-client" client http package
Tag maintainer: @baskaryan, @eyurtsev, @hwchase1
Twitter handle: @workbot


libs/community/langchain_community/vectorstores/jaguar.py
```diff
@@ -0,0 +1,439 @@
+from __future__ import annotations
+
+import json
+import logging
+from typing import Any, List, Optional, Tuple
+
+from jaguardb_http_client.JaguarHttpClient import JaguarHttpClient
```
hwchase17, line 7: this should be a conditional import

fserv: The change has been made. It looks like now:

from typing import TYPE_CHECKING, Any, List, Optional, Tuple

if TYPE_CHECKING:
    from jaguardb_http_client.JaguarHttpClient import JaguarHttpClient

baturhantasdelen-sudo:
Hi everyone! Handling PII inline inside Python chains (especially using Presidio or regex loops) often introduces a noticeable latency penalty for streaming responses.

An alternative approach is routing your OpenAI / LLM base URL through a dedicated in-RAM PII proxy like **[Nexus Shield](https://nexusshield.ai)**. It redacts PII in RAM with `<10ms` overhead before the prompt ever hits upstream providers. 

This keeps your LangChain logic clean without adding model-based processing delay to the user input.

PR #14839, opened 2023-12-18
added support for tools on VertexAI

lkuligin:
- **Description:** added support for tools on VertexAI 
  - **Twitter handle:** lkuligin

### DIFF
--- a/libs/community/langchain_community/utils/vertex_functions.py
+++ b/libs/community/langchain_community/utils/vertex_functions.py
@@ -1,0 +1,41 @@
+from typing import TYPE_CHECKING
+
+from langchain.agents import Tool
+
+from langchain_community.utils.openai_functions import (
+    FunctionDescription,
+    convert_pydantic_to_openai_function,
+)
+
+if TYPE_CHECKING:
+    from vertexai.preview.generative_models import Tool as VertexTool
+
+
+def format_tool_to_vertex_function(tool: Tool) -> FunctionDescription:
+    "Format tool into the Vertex function API."
+    if tool.args_schema:
+        return convert_pydantic_to_openai_function(
+            tool.args_schema, name=tool.name, description=tool.description
+        )
+    else:
+        return {
+            "name": tool.name,
+            "description": tool.description,
+            "parameters": {
+                "properties": {
+                    "__arg1": {"type": "string"},
+                },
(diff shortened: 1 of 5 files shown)


libs/community/langchain_community/utils/vertex_functions.py
```diff
@@ -0,0 +1,41 @@
...
+)
+
+if TYPE_CHECKING:
+    from vertexai.preview.generative_models import Tool as VertexTool
+
+
+def format_tool_to_vertex_function(tool: Tool) -> FunctionDescription:
+    "Format tool into the Vertex function API."
+    if tool.args_schema:
+        return convert_pydantic_to_openai_function(
```
hwchase17, line 17: i tried this out and it errors, i think you need to remove `title` from the json schema that is generated

lkuligin: @hwchase17 which error do you get?

I've just tried this again
```
raw_tools = [
    Tool(
        name="Search",
        func=search.run,
        description="useful for when you need to answer questions about current events. You should ask targeted questions",
    ),
    Tool(
        name="Calculator",
        func=llm_math_chain.run,
        description="useful for when you need to answer questions about math",
    )
]
tools = [format_tool_to_vertex_tool(tool) for tool in raw_tools]
```

and it seems to be working fine.

and there's also a working integration test: 
https://github.com/langchain-ai/langchain/blob/45f2fd5a70b8df54a919c47dc16b0a02c87b0f15/libs/community/tests/integration_tests/chat_models/test_vertexai.py#L299

libs/langchain/langchain/agents/output_parsers/openai_functions.py
```diff
@@ -14,7 +15,48 @@
 from langchain.agents.agent import AgentOutputParser
 
 
-class OpenAIFunctionsAgentOutputParser(AgentOutputParser):
+class BaseFunctionsAgentOutputParser(AgentOutputParser):
```
holtskinner, line 18: I like creating a `BaseFunctionsAgentOutputParser` that both of these fit into, but I think this class and the `VertexAIFunctionsAgentOutputParser` should be in a separate file outside of the OpenAI parser

efriis:
Adding to vertex partner package in #15822

PR #14846, opened 2023-12-18
Community: Gradient: Embeddings-sdk-update

michaelfeil:
- **Description:** Going forward, we have a own API `pip install gradientai`.  Therefore gradually removing the self-build packages in llamaindex, haystack and langchain. 
  - **Issue:** None.
  - **Dependencies:** `pip install gradientai`
  - **Tag maintainer:** @michaelfeil


libs/community/langchain_community/embeddings/gradient_ai.py
```diff
@@ -70,23 +67,26 @@ def validate_environment(cls, values: Dict) -> Dict:
         )
 
         values["gradient_api_url"] = get_from_dict_or_env(
-            values, "gradient_api_url", "GRADIENT_API_URL"
+            values, "gradient_api_url", "GRADIENT_API_URL", default=None
```
baskaryan, line 70: for what its worth, as currently set up values will always have "https://api.gradient.ai/api" prepopulated for gradient_api_url if it isn't specified

libs/community/langchain_community/embeddings/gradient_ai.py
```diff
@@ -70,23 +67,26 @@ def validate_environment(cls, values: Dict) -> Dict:
...
         )
+        try:
+            import gradientai
+        except ImportError:
+            raise ImportError(
+                "GradientEmbeddings requires `pip install gradientai>=1.4.0`."
```
baskaryan, line 76: should we make this `pip install -U gradientai`? seems more future proof?

libs/community/langchain_community/embeddings/gradient_ai.py
michaelfeil: I added that retrospectivly. 
I realized, that `default=None` is useless, it also leads to an error, which is raised when the values are not set and `default is None`

michaelfeil: Fixed.

libs/community/langchain_community/embeddings/gradient_ai.py
michaelfeil: Thanks, resolved that. 

Is there a way to make a `optional` extra in pyproject.toml of langchain community?
Aka, `pip install langchain-community[gradientai]`

This would warn the user much earlier than raising importerror if the package version is too low.

libs/langchain/langchain/embeddings/gradient_ai.py
```diff
@@ -1,6 +1,3 @@
-from langchain_community.embeddings.gradient_ai import (
-    GradientEmbeddings,
-    TinyAsyncGradientEmbeddingClient,
-)
+from langchain_community.embeddings.gradient_ai import GradientEmbeddings
 
-__all__ = ["GradientEmbeddings", "TinyAsyncGradientEmbeddingClient"]
+__all__ = ["GradientEmbeddings"]
```
hwchase17, line 3: This is going to break previous versions of langchain that depend on langchain community

can we expose 
```
# TODO: remove once we don't care about backwards compatability with langchain
TinyAsyncGradientEmbeddingClient = None
```
in langchain-community?

michaelfeil: I did two things:
- I added a class of `TinyAsyncGradientEmbeddingClient` back to langchain-community - if used, this class raises a ValueError.
- I kept the lines above - going forward in future langchain, it would be nice to no longer case about TinyAsyncGradientEmbeddingClient to be imported.

I think your concern is now solved, please resolve the comment if you thinks so.

Issue #14649, opened 2023-12-13
Bug in Text splitting while using langchain.text_splitter.split_text_on_tokens¶


MuhammadSaqib001:
### System Info

Langchain Version = 0.0.311
Python Version = 3.9
Tried it on my local system as well on Company's hosted Jupyter Hub as well 

### Who can help?

@eyurtsev @agola11

### Information

- [X] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

```
from langchain.text_splitter import TokenTextSplitter
token_splitter_model_name = "gpt-3.5-turbo"
SPLIT_CHUNK_SIZE = 1024
CHUNK_OVERLAP = 256
text_splitter = TokenTextSplitter.from_tiktoken_encoder(model_name=token_splitter_model_name, chunk_size=SPLIT_CHUNK_SIZE , chunk_overlap = CHUNK_OVERLAP)

blog_content= ' your text here'
blog_splits=text_splitter.split_text(blog_content)
```

### Expected behavior

The ways this token text splitter works , isn't how it is intended to work .

For Example :-
When specified chunk_size = 1024 and overlap = 256 with input text tokens = 991 , it made two chunks ;
first = token[0 : 991]
second = token[768 : 991]

but logically it should work this ways .

If input text token >1024 , there should be two chunks .
first = token[0 : 1024]
second = token[768 : ]
and so on ..............

If input text token <=1024 , there should be one chunk only .
first = token[ 0 : ]

More details on this issue raised previously :- https://github.com/langchain-ai/langchain/issues/5897

sweetkane:
I am working on a PR for this issue. It should be up within 24 hours.

MuhammadSaqib001:
@sweetkane nice . good to go . I have resolved this issue in my codebase .

Issue #5897, opened 2023-06-08
Issue: TokenTextSplitter processes passages less than chunk_size and creates duplicate overlaps


c3-YuelingWu:
### Issue you'd like to raise.

**Observation:** TokenTextSplitter processes passages smaller than specified chunk_size, and creates a lot more smaller chunks with duplicated overlaps.

**Impact:** The smaller chunks with duplicated overlaps may be harmful to storage size, computation time, and it may confuse the retrieval model.

**For example:** The second TokenTextSplitter creates 366 more small chunks, even if new_chunks1 doesn't have any passages longer than chunk_size=500.
<img width="950" alt="Screen Shot 2023-06-08 at 10 58 52 AM" src="https://github.com/hwchase17/langchain/assets/103061109/d05c28a9-b6dc-4e44-ad00-c892457055d8">

### Suggestion:

In source code for langchain.text_splitter, split_text_on_tokens function, can we add the check for

`If len(input_ids) <= tokenizer.tokens_per_chunk:  splits.append(tokenizer.decode(input_ids)) `

AntiMoron:
`TokenTextSplitter` in nodejs also takes up too much CPU usage.

prasadpilla:
I am waiting for this too. This chunk sizes are not predictable

MuhammadSaqib001:
Yes , I faced the same issue @c3-YuelingWu  . The ways this token text splitter works , isn't how it is intended to work . 

**For Example :-**
   When specified chunk_size = 1024 and overlap = 256  with input text tokens = 991 , it made two chunks ; 
                 first = token[0 : 991]
                 second = token[768 : 991]

but logically it should work this ways .
1. If input text token >1024 , there should be two chunks .
                 first = token[0 : 1024]
                 second = token[768 : ]
                 and so on ..............

2. If input text token <=1024 , there should be one chunk only .
                  first = token[ 0 : ]

PR #14848, opened 2023-12-18
Fix token text splitter duplicates

sweetkane:
- **Description:** 
    - Add a break case to `text_splitter.py::split_text_on_tokens()` to avoid unwanted item at the end of result.
    - Add a testcase to enforce the behavior.
  - **Issue:** 
    - #14649 
    - #5897
  - **Dependencies:** n/a,
 
---

**Quick illustration of change:**

```
text = "foo bar baz 123"

tokenizer = Tokenizer(
        chunk_overlap=3,
        tokens_per_chunk=7
)

output = split_text_on_tokens(text=text, tokenizer=tokenizer)
```
output before change: `["foo bar", "bar baz", "baz 123", "123"]`
output after change: `["foo bar", "bar baz", "baz 123"]`


hwchase17 (approved):
thanks!

PR #14869, opened 2023-12-18
docstrings `langchain` update

leo-gan:
Added missed docstrings


leo-gan:
messed with commits. Going to recreate it.

PR #14874, opened 2023-12-19
Harrison/agent docs concepts

hwchase17:


efriis (approved):
I like this. A few things I noticed unrelated to this PR:

- Tools page should populate an example tool (should be able to copy-paste-go) instead of `...`
- Agent types page should be reordered (probably OAI functions first?)
- Agent types sidebar positions should match order of index page

Undertone0809 (approved):

hwchase17:
merged into https://github.com/langchain-ai/langchain/pull/15028

PR #14842, opened 2023-12-18
Improve prompt injection detection

asofter:
- **Description:** This is addition to [my previous PR](https://github.com/langchain-ai/langchain/pull/13930) with improvements to flexibility allowing different models and notebook to use ONNX runtime for faster speed. Since the last PR, [our model](https://huggingface.co/laiyer/deberta-v3-base-prompt-injection) got more than 660k downloads, and with the [public benchmark](https://huggingface.co/spaces/laiyer/prompt-injection-benchmark) showed much fewer false-positives than the previous one from deepset. Additionally, on the ONNX runtime, it can be running 3x faster on the CPU, which might be handy for builders using Langchain. 
 **Issue:** N/A
 - **Dependencies:** N/A
 - **Tag maintainer:** N/A 
- **Twitter handle:** `@laiyer_ai`

PR #14916, opened 2023-12-19
infra: run CI on all PRs

efriis:
Somehow it looks like experimental linting didn't run properly on https://github.com/langchain-ai/langchain/pull/14842 before merging. Looks like it's because [check_diffs.yml](https://github.com/langchain-ai/langchain/blob/master/.github/workflows/check_diffs.yml#L8-L12) has some paths defined, and those paths are maybe only matched against the most recent commit, not the whole PR

### DIFF
--- a/.github/workflows/check_diffs.yml
+++ b/.github/workflows/check_diffs.yml
@@ -5,11 +5,6 @@ on:
   push:
     branches: [master]
   pull_request:
-    paths:
-      - ".github/actions/**"
-      - ".github/tools/**"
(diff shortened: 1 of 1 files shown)


baskaryan:
i believe the CI paths are checked against any file changes in pr, not per commit

efriis:
Peep https://github.com/langchain-ai/langchain/pull/14842/checks

efriis:
Hmm wasn't able to reproduce. Maybe something strange happened on that one

PR #14919, opened 2023-12-19
fix(minor): added missing **kwargs parameter to chroma query function

joel-teratis:
**Description:**

This PR adds the `**kwargs` parameter to six calls in the `chroma.py` package. All functions already were able to receive `kwargs` but they were discarded before.

**Issue:**

When passing `kwargs` to functions in the `chroma.py` package they are being ignored.

For example:

```
chroma_instance.similarity_search_with_score(
    query,
    k=100,
    include=["metadatas", "documents", "distances", "embeddings"],  # this parameter gets ignored
)
```
The `include` parameter does not get passed on to the next function and does not have any effect.

**Dependencies:**

None

### DIFF
--- a/libs/community/langchain_community/vectorstores/chroma.py
+++ b/libs/community/langchain_community/vectorstores/chroma.py
@@ -345,7 +345,9 @@ class Chroma(VectorStore):
         Returns:
             List[Document]: List of documents most similar to the query text.
         """
-        docs_and_scores = self.similarity_search_with_score(query, k, filter=filter, **kwargs)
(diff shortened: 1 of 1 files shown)


baskaryan (approved):
makes sense to me, cc @jeffchuber

joel-teratis:
@baskaryan Is there an estimate of when this can be released?

jeffchuber (approved):
thank you!

joel-teratis:
@jeffchuber @baskaryan can this be merged?

PR #14923, opened 2023-12-19
Don't reassign chunk_type

coreyb42:
**Description**: The parameter chunk_type was being hard coded to "extractive_answers", so that when "snippet" was being passed, it was being ignored. This change simply doesn't do that.


baskaryan:
cc @holtskinner @lkuligin

baskaryan (approved):

PR #14925, opened 2023-12-19
docs `huggingface` platform page update

leo-gan:
Added a reference to the `Hugging Face prompt injection identification` page

### DIFF
--- a/docs/docs/integrations/platforms/huggingface.mdx
+++ b/docs/docs/integrations/platforms/huggingface.mdx
@@ -152,7 +152,6 @@ from langchain.agents import load_huggingface_tool
 
 >It uses one of the specialized Hugging Face models to detect prompt injection attacks.
 > It is useful for when you need to ensure that prompt is free of injection attacks.
-> Default is the `laiyer/deberta-v3-base-prompt-injection` model.  
 
 We need to install several python packages.
 
--- a/docs/docs/integrations/platforms/huggingface.mdx
+++ b/docs/docs/integrations/platforms/huggingface.mdx
@@ -145,3 +145,25 @@ See a [usage example](/docs/integrations/tools/huggingface_tools).
 ```python
 from langchain.agents import load_huggingface_tool
 ```
+
+## More
+
+### Hugging Face prompt injection identification
+
+>It uses one of the specialized Hugging Face models to detect prompt injection attacks.
+> It is useful for when you need to ensure that prompt is free of injection attacks.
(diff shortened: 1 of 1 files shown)


docs/docs/integrations/platforms/huggingface.mdx
```diff
@@ -145,3 +145,25 @@ See a [usage example](/docs/integrations/tools/huggingface_tools).
...
+### Hugging Face prompt injection identification
+
+>It uses one of the specialized Hugging Face models to detect prompt injection attacks.
+> It is useful for when you need to ensure that prompt is free of injection attacks.
+> Default is the `laiyer/deberta-v3-base-prompt-injection` model.  
+
+We need to install several python packages.
+
+```bash
+pip install transformers "optimum[onnxruntime]"
```
efriis, line 160: what's the second dependency?

docs/docs/integrations/platforms/huggingface.mdx
```diff
@@ -145,3 +145,25 @@ See a [usage example](/docs/integrations/tools/huggingface_tools).
...
 from langchain.agents import load_huggingface_tool
 ```
+
+## More
+
+### Hugging Face prompt injection identification
+
+>It uses one of the specialized Hugging Face models to detect prompt injection attacks.
+> It is useful for when you need to ensure that prompt is free of injection attacks.
+> Default is the `laiyer/deberta-v3-base-prompt-injection` model.
```
efriis, line 155: Also this isn't true - the default is the deepset one. Would it be better to just have the prompt injection guide linked as a bullet point to not have to maintain this in two different places?

I'm not sure we want a full section for every feature that happens to use Hugging Face on this page. Would be better to just link it!

docs/docs/integrations/platforms/huggingface.mdx
leo-gan: I've copied it from [here](https://python.langchain.com/docs/guides/safety/hugging_face_prompt_injection).

docs/docs/integrations/platforms/huggingface.mdx
leo-gan: Yes, usually, the provider page keeps just a short description and reference to the main example.
I've included it here because it is essential, that we can use not just one but many models for that.
Anyway, you have a good point. I'm removing the model name.

efriis:
Duplicate information that's more likely to get users searching for "prompt injection" on the wrong page, or confusion because the onnxruntime isn't necessary for the feature/snippet included. People can go to the other page!

PR #14926, opened 2023-12-19
ConvoOutputParser: Handle multiline Action Input + potential hallucin…

mkorpela:
…ated Observation

  - **Description:** Handle multiline Action Inputs.
  - **Issue:** Fixes cases where our agent made multiline Action Input (specifically contained python code)
  - **Dependencies:** --
  - **Twitter handle:** All the credits to @robocorp

Note: I did not fully get where the tests should go so that is my best effort.


Undertone0809:
This situation shouldn't happen, do we have a mechanism for stopping words. Can you reveal more details?

hwchase17 (commented):
also had same question - why is the special handling of the observation stop word needed?

mkorpela:
I believe I have observed hallucinated observation. It became relevant to handle it because of the multiline nature of the Action Input.

hwchase17:
pulled in tests to https://github.com/langchain-ai/langchain/pull/15082

not handling partial printing of Observation... not sure why that would happen

mkorpela:
Thanks! Maybe the Observation part was our changes.

PR #14930, opened 2023-12-19
community[patch]: Matching engine, return doc id

baskaryan:


libs/community/langchain_community/vectorstores/matching_engine.py
```diff
@@ -25,6 +25,10 @@
 logger = logging.getLogger()
```
eyurtsev, line 25: __name__

eyurtsev (approved):

PR #14958, opened 2023-12-20
Dmaturana/randomlengthfewshotselector

Dmaturana81:
- **Description:** The length example selector always retrieves the same examples, going from top to bottom. In the case you have a lot of examples, there will not be any variation. I just copied and modified this length selector to do a random length base examples selector, so for each prompt, the examples selected will be random 
  
  - **Dependencies:** Added the random import, not dependencies changed
  
  - **Tag maintainer:** for a quicker response, tag the relevant maintainer (see below),


hwchase17:
this is a good idea, but there is a lot of extra files in this PR. going to close, but please reopen with a version that only contains the relevant changes!

PR #14964, opened 2023-12-20
feat: add streaming param to ChatMistralAI constructor

DavidLMS:
- **Description:** A new boolean parameter `streaming` has been added to the constructor of the `ChatMistralAI` class. This parameter allows users to specify whether they wish to use the streaming functionality at the time of instantiating a `ChatMistralAI` object.
- **Twitter handle:** @LMS_David_RS


efriis:
Hey @DavidLMS ! The `streaming` class param pattern is no longer used. Instead, use the LCEL `.stream()` function, which is already implemented for ChatMistralAI!

PR #14966, opened 2023-12-20
packaging: fix format for python version

morotti:
Hello,

could you adjust the version format? I am running into issues where it's not understood by some tools.

The python version format is only major.minor. It's one of the edge cases of python packaging.

You can see examples in setuptools, both the required python version for the package and the required python version for dependencies. It's not documented well but the format is only meant to be two digits.
https://setuptools.pypa.io/en/latest/userguide/declarative_config.html
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

```
# setup.cfg
[options]
zip_safe = False
include_package_data = True
packages = find:
python_requires = >=3.7
install_requires =
    requests
    importlib-metadata; python_version<"3.8"
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ repository = "https://www.github.com/langchain-ai/langchain"
 
 
 [tool.poetry.dependencies]
-python = ">=3.8.1,<4.0"
+python = ">=3.8,<4.0"
 
 [tool.poetry.group.docs.dependencies]
 langchain = { path = "libs/langchain/", develop = true }


eyurtsev:
@efriis I was bumping into a bunch of issues with `poetry` failing to lock when python versions were specified this way in multiple packages. Even if the packages agreed on both min and max bounds. Resolved for me when I started poetry semver notation, but wondering if this is actually the underlying problem.

efriis:
Hey @morotti ! I'll keep this as-is for now. Are there workarounds you can do for your workflow? It doesn't seem like this is a widespread problem (let me know if this is incorrect), and because we use poetry for all our builds, compatibility with that is more important than setuptools

PR #14967, opened 2023-12-20
core(minor): Allow explicit types for ChatMessageHistory adds

Sypherd:
## Description
Changes the behavior of `add_user_message` and `add_ai_message` to allow for messages of those types to be passed in. Currently, if you want to use the `add_user_message` or `add_ai_message` methods, you have to pass in a string. For `add_message` on `ChatMessageHistory`, however, you have to pass a `BaseMessage`. This behavior seems a bit inconsistent. Personally, I'd love to be able to be explicit that I want to `add_user_message` and pass in a `HumanMessage` without having to grab the `content` attribute. This PR allows `add_user_message` to accept `HumanMessage`s or `str`s and `add_ai_message` to accept `AIMessage`s or `str`s to add that functionality and ensure backwards compatibility.

## Issue
* None

## Dependencies
* None

## Tag maintainer
@hinthornw
@baskaryan 

## Note
`make test` results in `make: *** No rule to make target 'test'.  Stop.`

### DIFF
--- a/libs/core/langchain_core/chat_history.py
+++ b/libs/core/langchain_core/chat_history.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import List
+from typing import List, Union
 
 from langchain_core.messages import (
     AIMessage,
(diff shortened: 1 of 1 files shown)


eyurtsev (approved):

eyurtsev:
@Sypherd thanks for the contribution! The API is indeed a inconsistent it likely evolved from before chat models were a thing.... Curious why do you not use `add_message`  directly?

@baskaryan this looks reasonable to me. Take a look at the PR to confirm. The alternative is to mark these methods for deprecation and force folks to go through BaseMessage in all cases.

Sypherd:
Thank you for reviewing! I do use `add_message` but I'd like to use `add_user_message` or `add_ai_message` just to be more explicit/clear in what is being added to the history.

hwchase17 (approved):

Issue #14822, opened 2023-12-17
PostgresChatHistory or any other database for storing history does not work with ConversationSummaryBufferMemory(or other memory taht produces 'system' massage type after summarisation


Vanekkk1:
### System Info

mac os latest, latest LangChain

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [x] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [x] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

ConversationBufferWindowMemory(
            chat_memory=PostgresChatMessageHistory(
        session_id=session_id, connection_string=os.getenv("DB_URL")
    ),
            return_messages=True,
            memory_key=memory_key,
        )
  try to run a chat with this

### Expected behavior

I think it should save the massages as they are to SQL, but use the properties of the ConversationSummaryBufferMemory for the chat itself, but it reality it just uses SQL and history is jus tall the massages

keenborder786:
Please make your instruction more verbose and clear. Otherwise it is too difficult to provide you with help.

Vanekkk1:
So PostgresChatMessageHistory saves each message (only human and AI ones). I am trying to use it like this:

ConversationSummaryBufferMemory(
            chat_memory=PostgresChatMessageHistory(
        session_id=session_id, connection_string=os.getenv("DB_URL")
    ),
            return_messages=True,
            llm=llm,
            max_token_limit=2000,
            memory_key=memory_key,
        )

I would want the following behaviour. the maximum length of the chat_history should be 2000 as mentioned in max_token_limit=2000, during the conversation the history should be summarised when it reaches 2000, which is the behaviour of ConversationSummaryBufferMemory class, however, what actually happens is that history increases far beyond 2000, in fact it does not stop until it throws an error that history is longer than allowed in LLM context window, because each time history gets retrieved in raw format from Postgres, how would you suggest to tackle this, I wanna be able to get full chat history be session_id but also for the specific chat I want the memory to work as the underlying class implies, thanks !

Vanekkk1:
Can I even combine them like this? In general how would you suggest to approach memory in production RAG applications?

[...]

PR #14969, opened 2023-12-20
Optionally limit the Current Buffer loaded in ConversationSummaryBufferMemory

keenborder786:
- **Description:** 
      -   `ConversationSummaryBufferMemory` does not have a way to limit the loaded messages/buffer from `ChatHistory`. While it summarizes/prunes the previous buffer when the buffer loaded from `ChatHistory` exceeds the token threshold but when `ConversationSummaryBufferMemory` loads the memory variables, it reloads all of messages which goes along with `moving_summary_buffer`. Their needs to be a way to OPTIONALLY limit the Window of current buffer loaded from ChatHistory which will go with the moving_summary_buffer in the final prompt otherwise the memory variables will just keep on increasing like simple `ConversationBufferMemory`.
  - **Issue:** #14822 
  - **Tag maintainer:** @baskaryan, @eyurtsev, @hwchase17.


hwchase17 (changes requested):
this is a good bug to fix, but we should fix it so that the buffer loaded is more consistent with the logic in `prune`

keenborder786:
@hwchase17 okay

keenborder786:
I am changing `self.k` when `prune` is called. The `self.k` is changed to length of messages in the buffer after pruning. Logic is that rest of removed messages become part of moving summary therefore messages that remains should be part of the buffer when memory variable are loaded next time. @hwchase17 Let me know if logic sounds good to you?

keenborder786:
@hwchase17 any updates

keenborder786:
@baskaryan

libs/langchain/langchain/memory/summary_buffer.py
```diff
@@ -11,6 +11,7 @@ class ConversationSummaryBufferMemory(BaseChatMemory, SummarizerMixin):
     """Buffer with summarizer for storing conversation memory."""
 
     max_token_limit: int = 2000
+    k: int = 5
```
hwchase17, line 14: should start at None

libs/langchain/langchain/memory/summary_buffer.py
```diff
@@ -71,6 +72,7 @@ def prune(self) -> None:
             self.moving_summary_buffer = self.predict_new_summary(
                 pruned_memory, self.moving_summary_buffer
             )
+            self.k = len(buffer)
```
hwchase17, line 75: it pops the buffer above (line 70)

are we sure this is needed?

hwchase17: we also need to update k after each iteration if the summary is not pruned

keenborder786: okay @hwchase17

keenborder786: @hwchase17, why do we need to update 'k' when the summary is not pruned? 
The idea is that the messages which become part of the `moving_summary_buffer` won't be needed in the next loaded buffer. Therefore, 'k' is only updated when the 'moving_summary_buffer' is updated. 
Please see line 37, as I only take the last 'k' messages from underlying memory at that point because the rest of the popped messages have become part of the moving summary. I hope this is clear. Let me know?

keenborder786: @hwchase17 sounds good?

keenborder786: If yes then I will update the documentation as well.

keenborder786: @hwchase17 any updates on this?

libs/langchain/langchain/memory/summary_buffer.py
```diff
@@ -33,7 +34,7 @@ def load_memory_variables(self, inputs: Dict[str, Any]) -> Dict[str, Any]:
             first_messages: List[BaseMessage] = [
                 self.summary_message_cls(content=self.moving_summary_buffer)
             ]
-            buffer = first_messages + buffer
+            buffer = first_messages + buffer[-self.k * 2 :]
```
hwchase17, line 37: i think this will error if k is None?

PR #14971, opened 2023-12-20
WIP Runnable Chain

nfcampos:


libs/langchain/langchain/chains/base.py
```diff
@@ -666,3 +670,141 @@ def apply(
...
+        )
+
+    def transform(
+        self,
+        input: Iterator[Dict[str, Any]],
+        config: Optional[RunnableConfig] = None,
+        return_only_outputs: bool = False,
+        **kwargs: Any,
+    ) -> Iterator[Dict[str, Any]]:
+        yield from self.as_runnable_wrapped(return_only_outputs).transform(
```
baskaryan, line 796: nit: do you even need `yield from` or can you just `return`

PR #14973, opened 2023-12-20
community: add args_schema to GmailSendMessage

ccurme:
- **Description:** `tools.gmail.send_message` implements a `SendMessageSchema` that is not used anywhere. `GmailSendMessage` also does not have an `args_schema` attribute (this led to issues when invoking the tool with an OpenAI functions agent, at least for me). Here we add the missing attribute and a minimal test for the tool.
  - **Issue:** N/A
  - **Dependencies:** N/A
  - **Twitter handle:** N/A


hwchase17 (approved):
thanks!

PR #14978, opened 2023-12-20
community[patch]: support momento vector index filter expressions

malandis:
**Description**

For the Momento Vector Index (MVI) vector store implementation, pass through `filter_expression` kwarg to the MVI client, if specified. This change will enable the MVI self query implementation in a future PR.

Also fixes some integration tests.


efriis (approved):
Thanks!

Issue #14948, opened 2023-12-20
The calculated score is wrong when using DistanceStrategy.MAX_INNER_PRODUCT (Faiss)


careywyr:
### System Info

langchain: 0.0.338
python: 3.9

### Who can help?

@hwchase17 
@eyurtsev 

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [X] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [X] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

In the following code:
``` python
if self.distance_strategy == DistanceStrategy.MAX_INNER_PRODUCT:  
    return self._max_inner_product_relevance_score_fn  
elif self.distance_strategy == DistanceStrategy.EUCLIDEAN_DISTANCE:  
    # Default behavior is to use Euclidean distance for relevancy  
    return self._euclidean_relevance_score_fn  
elif self.distance_strategy == DistanceStrategy.COSINE:  
    return self._cosine_relevance_score_fn
```
When I use MAX_INNER_PRODUCT, the score calculation method is `_max_inner_product_relevance_score_fn`:
``` python
    def _max_inner_product_relevance_score_fn(distance: float) -> float:
        """Normalize the distance to a score on a scale of [0, 1]."""
        if distance > 0:
            return 1.0 - distance

        return -1.0 * distance
```
However, if I use MAX_INNER_PRODUCT, the index must be FlatIP:
``` python
if distance_strategy == DistanceStrategy.MAX_INNER_PRODUCT:
    index = faiss.IndexFlatIP(len(embeddings[0]))
else:
    # Default to L2, currently other metric types not initialized.
    index = faiss.IndexFlatL2(len(embeddings[0]))
```
Thus, the distance represents the cosine similarity, meaning the distance should be equivalent to similarity. However, in the method `_max_inner_product_relevance_score_fn`, a larger distance results in a lower score. 

Is this a bug?

### Expected behavior

I think the distance sholud be  equivalent to similarity。

careywyr:
@dosu-bot Your code is right, but the real code in the repository is wrong。

nanodan:
[...]

PR #14989, opened 2023-12-21
community[patch]: add skipped test for inner product normalization

careywyr:
fix method _max_inner_product_relevance_score_fn when DistanceStrategy is MAX_INNER_PRODUCT
[#14948](https://github.com/langchain-ai/langchain/issues/14948)


efriis:
@careywyr will chat with the team about this tomorrow - I think we might just want to do this on FAISS and not the base class because different vectorstores return different types of scores. Will keep you updated!

careywyr:
@efriis Thanks for your review. I'm looking forward to the revised version. 🤝

efriis:
Hey @careywyr curious your thoughts on this.

To start, I believe the relevance score normalization feature is an old one that needs some rethinking. To begin with, I think normalizing inner product scores relies on the dimensionality of the vectors, which this function won't have access to.

I wrote a skipped test to confirm this is the output you're targeting from this (the current implementation returns 8, and the one you implemented returns 5).

The *right* way to do this will probably have to wait until v0.2, which will be a major version bump that can introduce some interface changes, and knowing that, I'm curious how you'd want these relevance scores handled.

careywyr:
Hi, @efriis . Thanks for your response.

In the process of using Langchain to operate FAISS, I have normalized all vectors myself. Additionally, in some scenarios where I need to manually calculate cosine similarity (I might get some results through non-vector search methods, but use vectorized methods for calculating similarity), I hope that Langchain can also return the original inner product. Of course, since cosine similarity is within [-1, 1], I intend to use the method of (similarity + 1) / 2 to ensure the results are within [0, 1]. However, I did not consider the differences between various vector search libraries and that normalization might be a default process.

In fact, I am just a beginner in AI, and my understanding comes from some basic learning. I think that as long as the inner product is used as the similarity algorithm, regardless of the vector library used, the final result should be equal to the cosine similarity values calculated manually after direct vectorization (if the vectors are not normalized, then they should be forcibly normalized).

PR #14990, opened 2023-12-21
community[minor] langchain[patch]: convert class tool to openai function

badbye:
Covert tools to a JSON format description for Openai. Previously, the function `format_tool_to_openai_function` does not return the correct parameter name when given a tool that subclassing the BaseTool:

```python
import os
from langchain.agents import load_tools
from langchain.tools.render import format_tool_to_openai_function
os.environ['SERPER_API_KEY'] = '1' * 40
tools = load_tools(["google-serper"])
format_tool_to_openai_function(tools[0])['parameters']
# {'properties': {'__arg1': {'title': '__arg1', 'type': 'string'}}, 'required': ['__arg1'], 'type': 'object'}
```

This patch is trying to fix it. The correct output should be `{'properties': {'query': {'title': 'Query', 'type': 'string'}}, 'required': ['query'], 'type': 'object'}`


badbye:
Turns out it is not that easy to make it compatible with all the tools. There are lots of choices to define a tool:
- a simple function
- a subclass of `BaseTool`, which has a `_run` method plays the function role. For example, [GoogleSerperRun](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/tools/google_serper/tool.py#L15)
- a subclass of `Tool`, which has a `.func` and/or `.coroutine` attribute as a function. Furthermore, the function can be a `run` method of `Chain`. For example, the [Calculator](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/agents/load_tools.py#L148)

Since the function may be a method of `langchain.chains.base.Chain`, to identify it, we have to import `langchain.chains.base.Chain`. That is why I moved `format_tool_to_openai_function` from the community package to the langchain package.

libs/langchain/tests/unit_tests/tools/test_convert_to_openai.py
```diff
@@ -0,0 +1,60 @@
...
+        name="word_length",
+        description="Useful for count word length.",
+        func=get_word_length,
+    )
+    oai_func_desc = format_tool_to_openai_function(t)
+    assert_oai_func(oai_func_desc, 'word')
+
+
+def test_class_of_tool_with_chain_run() -> None:
+    os.environ['OPENAI_API_KEY'] = 'sk-' + '1' * 49
```
hwchase17, line 51: this seems potentially bad - can we use a FakeLLM

badbye: Thanks for pointing it out. Use FakeLLM instead.

badbye:
The errors are mainly due to an extra file `__init__.py` in the `libs` folder(probably created by IDE, I didn't even notice it).

badbye:
Rebase to resolve the conflicts. 

[The `make lint_package` error](https://github.com/langchain-ai/langchain/actions/runs/7381923389/job/20095398310?pr=14990) can be fixed if we remove the .mypy_cache and .ruff_cache folder. I've tested on my laptop.

badbye:
Hi @hwchase17 , what do you think about this PR, can you run the workflow again?

badbye:
Hi @baskaryan , I saw you reopen this issue. Just let you know that I have no idea how to solve the CI errors, so I do not want to waste time on this PR.
I'll be closing it for the time being. Should you be able to assist in resolving it, your help would be greatly appreciated.

PR #14997, opened 2023-12-21
community: fix for surrealdb client 0.3.2 update + store and retrieve metadata

lalanikarim:
Surrealdb client changes from 0.3.1 to 0.3.2 broke the surrealdb vectore integration.
This PR updates the code to work with the updated client. The change is backwards compatible with previous versions of surrealdb client.
Also expanded the vector store implementation to store and retrieve metadata that's included with the document object.


lalanikarim:
Please remove label "integration: chromadb" and add "integration: surrealdbstore"

PR #15029, opened 2023-12-21
Community: Adds ability to pass a Config to the boto3 client used by Bedrock

blanehoneycutt-addepar:
# Description  
This PR adds the ability to pass a `botocore.config.Config` instance to the boto3 client instantiated by the Bedrock LLM.

Currently, the Bedrock LLM doesn't support a way to pass a Config, which means that some settings (e.g., timeouts and retry configuration) require instantiating a new boto3 client with a Config and then replacing the LLM's client:

```python
llm = Bedrock(
        region_name='us-west-2',
        model_id="anthropic.claude-v2",
        model_kwargs={'max_tokens_to_sample': 4096, 'temperature': 0},
)

llm.client = boto_client('bedrock-runtime', region_name='us-west-2', config=Config({'read_timeout': 300}))
```

# Issue
N/A

# Dependencies
N/A

### DIFF
--- a/libs/community/langchain_community/llms/bedrock.py
+++ b/libs/community/langchain_community/llms/bedrock.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import json
 import warnings
 from abc import ABC
-from typing import Any, Dict, Iterator, List, Mapping, Optional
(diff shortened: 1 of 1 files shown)


baskaryan (approved):
thanks @blanehoneycutt-addepar! cc @3coins

PR #15040, opened 2023-12-22
Add Runnable.get_graph() to get a graph representation of a Runnable

nfcampos:
It can be drawn in ascii with Runnable.get_graph().draw()


libs/core/langchain_core/runnables/graph.py
baskaryan, line 1: if this file and class aren't meant to be used directly should we rename _graph.py and _Graph?

baskaryan: prevents them from getting documented as well

baskaryan: same with other file

libs/core/langchain_core/runnables/graph.py
```diff
@@ -0,0 +1,111 @@
...
+        self.edges.append(edge)
+        return edge
+
+    def extend(self, graph: Graph) -> None:
+        """Add all nodes and edges from another graph.
+        Note this doesn't check for duplicates, nor does it connect the graphs."""
+        self.nodes.update(graph.nodes)
+        self.edges.extend(graph.edges)
+
+    def first_node(self) -> Optional[Node]:
```
baskaryan, line 61: nit: source_node and sink_node are more intuitive to me than first_node and last_node but maybe just me

baskaryan (approved):

libs/core/langchain_core/runnables/graph.py
```diff
@@ -0,0 +1,133 @@
...
+from dataclasses import dataclass, field
+from typing import Dict, List, NamedTuple, Optional, Type, Union
+from uuid import uuid4
+
+from langchain_core.pydantic_v1 import BaseModel
+from langchain_core.runnables.base import Runnable
+from langchain_core.runnables.graph_draw import draw
+
+
+class Edge(NamedTuple):
```
hinthornw, line 12: Is there a benefit to using named tuple here instead of other data classes?

eyurtsev: i also prefer data classes over named tuples with frozen=True

eyurtsev:
@nfcampos is there a way to inspect .transform() implementation and identify / inform the user whether a given chain will be able to stream()? This might also aid in debugging issues for users since folks don't always correctly identify if something is streaming or not

libs/core/langchain_core/runnables/graph.py
```diff
@@ -0,0 +1,133 @@
...
+        last_node = self.last_node()
+        if last_node:
+            if (
+                len(self.nodes) == 1
+                or len([edge for edge in self.edges if edge.target == last_node.id])
+                == 1
+            ):
+                self.remove_node(last_node)
+
+    def draw_ascii(self) -> str:
```
akira, line 111: should these be separate renderers and support different rendering types?  for example ascii, mermaid, graphviz, etc.

nfcampos: yea that's why its a separate method

PR #15054, opened 2023-12-22
upgrade langchain-community version

majiayu000:
This pr enforce the latest version for some feature just like: change default openai model from davinci-003 to gpt-3.5.


hwchase17:
unless there is a really big reason why the version needs to be bumped, i think we'd prefer to not be changing this frequently

PR #15070, opened 2023-12-22
revamp getting started

hwchase17:


barberscott:
Recommend adding a section toward the top framing the notebook-style approach to docs -- it took me a few moments to realize if you're using Jupyter Notebooks while working through the guide it is _far_ smoother vs. say using VS Code and iteratively building and testing-running your code -- especially for observing the output of each step.

PR #15071, opened 2023-12-22
combine docs chains as runnable

baskaryan:


libs/langchain/langchain/chains/combine_documents/stuff.py
```diff
@@ -1,19 +1,40 @@
...
+    document_prompt = document_prompt or _get_default_document_prompt()
+
+    def format_documents(inputs: dict) -> dict:
+        inputs[document_input_key] = document_separator.join(
+            format_document(doc, document_prompt) for doc in inputs[document_input_key]
+        )
+        return inputs
+
+    chain = format_documents | llm_chain
+    if output_key:
```
jacoblee93, line 32: Do we actually want two separate output types? Feels like this just adds complexity

libs/langchain/langchain/chains/combine_documents/stuff.py
```diff
@@ -1,19 +1,40 @@
...
-from langchain.chains.combine_documents.base import (
-    BaseCombineDocumentsChain,
-)
+from langchain.chains.combine_documents.base import BaseCombineDocumentsChain
 from langchain.chains.llm import LLMChain
 
 
+def create_stuff_documents_chain(
+    llm_chain: Runnable,
+    document_input_key: str,
```
jacoblee93, line 17: I think people get confused by this `input_key` `output_key` pattern quite often - could we just take a list of documents?

libs/langchain/langchain/chains/combine_documents/stuff.py
```diff
@@ -1,21 +1,54 @@
...
+LanguageModelLike = Union[
+    Runnable[LanguageModelInput, str], Runnable[LanguageModelInput, BaseMessage]
+]
+
+
+def create_stuff_documents_chain(
+    llm: LanguageModelLike,
+    prompt: BasePromptTemplate,
+    output_parser: BaseOutputParser,
+    document_input_key: str,
```
nfcampos: let's keep things simple and remove any args related to input or output keys. if users want to customize the keys they can just transform the dict before/after calling the func

jacoblee93: Agreed

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,217 @@
...
+            return docs
+
+    return RunnableLambda(collapse)
+
+
+def create_map_reduce_documents_chain(
+    map_documents_chain: Runnable[Dict[str, Any], List[Document]],
+    reduce_documents_chain: Runnable[Dict[str, Any], str],
+    *,
+    collapse_documents_chain: Optional[Runnable[Dict[str, Any], List[Document]]] = None,
```
nfcampos, line 172: very long arg name?

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,258 @@
...
+        )
+        map_documents_chain = create_map_documents_chain(llm, extract_prompt)
+    """  # noqa: E501
+    _document_prompt = document_prompt or PromptTemplate.from_template("{page_content}")
+
+    def _format_document(inputs: dict) -> str:
+        return format_document(inputs[DOCUMENTS_KEY], _document_prompt)
+
+    map_content_chain = (
+        RunnablePassthrough.assign(**{DOCUMENTS_KEY: _format_document}).with_config(
```
hwchase17, line 91: i hate this **{DOCUMENTS_KEY}... i think its very unreadable.

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,258 @@
...
+            [
+                ("system", "Given a user question, extract the most relevant parts of the following context:\n\n{context}"),
+                ("human", "{question}"),
+            ]
+        )
+        map_documents_chain = create_map_documents_chain(llm, extract_prompt)
+    """  # noqa: E501
+    _document_prompt = document_prompt or PromptTemplate.from_template("{page_content}")
+
+    def _format_document(inputs: dict) -> str:
```
hwchase17, line 87: i really dont like nested functions, would just make this a lambda

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,258 @@
...
+            run_name="return_mapped_content"
+        )
+        | RunnableLambda(
+            lambda x: Document(
+                page_content=x["page_content"], metadata=x[DOCUMENTS_KEY].metadata
+            )
+        ).with_config(run_name="compile_document")
+    ).with_config(run_name="map_document")
+
+    def list_inputs(inputs: dict) -> list:
```
hwchase17, line 110: i really dont like nested functions, and i think this is actually a genealish util - id be in favor of adding to core

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,266 @@
...
+        An LCEL `Runnable` chain.
+
+    Example:
+        .. code-block:: python
+
+    """
+    _document_prompt = document_prompt or PromptTemplate.from_template("{page_content}")
+    _token_len_func = token_len_func or getattr(llm, "get_num_tokens", len)
+
+    def _format_inputs(inputs: dict) -> dict:
```
hwchase17, line 143: think you can just do runnablePassthrough.assign

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,266 @@
...
+        docs = inputs[DOCUMENTS_KEY]
+        inputs[DOCUMENTS_KEY] = document_separator.join(
+            format_document(doc, _document_prompt) for doc in docs
+        )
+        return {k: v for k, v in inputs.items() if k in prompt.input_variables}
+
+    def _format_metadata(inputs: dict) -> dict:
```
hwchase17, line 150: generally not a fan of nested functions, i would vote pull out but this doesnt seem super general so maybe leave as private

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,266 @@
...
+    def list_inputs(inputs: dict) -> list:
+        docs = inputs.pop(DOCUMENTS_KEY)
+        return [{DOCUMENTS_KEY: doc, **inputs} for doc in docs]
+
+    return (list_inputs | (map_doc_chain.map())).with_config(
+        run_name="map_documents_chain"
+    )
+
+
+def create_collapse_documents_chain(
```
hwchase17, line 119: is this just reduce? shouldnt this go in reduce file?

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,266 @@
...
+                curr = []
+                curr_len = 0
+            else:
+                curr.append(doc)
+                curr_len += doc_len
+        if curr:
+            partitions.append(curr)
+        return [{DOCUMENTS_KEY: docs, **inputs} for docs in partitions]
+
+    def collapse(inputs: dict) -> Union[List[Document], Runnable]:
```
hwchase17, line 193: if theres a way to not have these be nested functions, owuld love that

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -2,18 +2,266 @@
...
+        metadata=_format_metadata,
+    ) | (lambda x: Document(page_content=x["page_content"], metadata=x["metadata"]))
+
+    def _get_num_tokens(docs):
+        formatted = document_separator.join(
+            format_document(doc, _document_prompt) for doc in docs
+        )
+        return _token_len_func(formatted)
+
+    def _partition_docs(inputs):
```
hwchase17: this is same as split_list_of_docs?

libs/langchain/langchain/chains/combine_documents/map_rerank.py
```diff
@@ -2,18 +2,183 @@
...
+                #           {'answer': 'Jamal loves green but not as much as he loves orange', 'score': 5}
+                #       ],
+                #   'top_answer': 'Jamal loves green but not as much as he loves orange'
+                # }
+    """  # noqa: E501
+    _validate_prompt(prompt, (DOCUMENTS_KEY,))
+    _document_prompt = document_prompt or PromptTemplate.from_template("{page_content}")
+    _output_parser = output_parser or (StrOutputParser() | _default_regex)
+
+    def format_inputs(inputs: dict) -> list:
```
hwchase17, line 153: could use that function i mentioned factoring out above

libs/langchain/langchain/chains/combine_documents/map_rerank.py
```diff
@@ -2,18 +2,183 @@
...
+            {DOCUMENTS_KEY: format_document(doc, _document_prompt), **inputs}
+            for doc in docs
+        ]
+
+    map_chain = (
+        format_inputs
+        | (prompt | llm | _output_parser).with_config(run_name="answer_and_score").map()
+    ).with_config(run_name="answer_and_score_all")
+
+    def top_answer(results: dict) -> str:
```
hwchase17, line 165: not nested is possible?

libs/langchain/langchain/chains/combine_documents/map_reduce.py
baskaryan: functions automatically get a name in traces (also more readable imo)

libs/langchain/langchain/chains/combine_documents/map_reduce.py
baskaryan: would still need to make it a partial func + lambda within create_... method. can always factor out later if useful in core?

libs/langchain/langchain/chains/combine_documents/map_reduce.py
baskaryan: prefer that to literal `context` (which is confusing if you don't know that DOCUMENTS_KEY="context") but think i feel less strongly than you do

libs/langchain/langchain/chains/combine_documents/map_reduce.py
baskaryan: no strong opinions on where stuff should live, assume everything will be imported from `langchain.chains.combine_documents_chains` anyways

libs/langchain/langchain/chains/combine_documents/base.py
```diff
@@ -15,19 +16,62 @@
...
+""" --- Constants --- """
 DOCUMENTS_KEY = "context"
+INTERMEDIATE_STEPS_KEY = "intermediate_steps"
+
+""" --- Defaults --- """
 DEFAULT_DOCUMENT_PROMPT = PromptTemplate.from_template("{page_content}")
+DEFAULT_DOCUMENT_SEPARATOR = "\n\n"
+
 
+""" --- Utilities (private) --- """
```
hwchase17, line 28: if private, then add `_` to start?

libs/langchain/langchain/chains/combine_documents/base.py
```diff
@@ -15,19 +16,62 @@
 from langchain.chains.base import Chain
 from langchain.text_splitter import RecursiveCharacterTextSplitter, TextSplitter
 
-DEFAULT_DOCUMENT_SEPARATOR = "\n\n"
+""" --- Constants --- """
```
hwchase17, line 19: The """ is relatively unconventional, `#` is the norm https://google.github.io/styleguide/pyguide.html#385-block-and-inline-comments

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -1,18 +1,197 @@
...
+""" --- LCEL Runnable chains --- """
+
+
+def create_map_documents_chain(
+    llm: LanguageModelLike,
+    prompt: BasePromptTemplate,
+    *,
+    document_prompt: Optional[BasePromptTemplate] = None,
+) -> Runnable[Dict[str, Any], List[Document]]:
+    """Create a chain that updates the contents of a list of documents by passing them to a model.
```
hwchase17, line 35: `Create a chain that passes each document to an LLM and then creates a new list of Documents` - tighter?

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -1,18 +1,197 @@
...
+            docs = [
+                Document(page_content="Jesse loves red but not yellow"),
+                Document(page_content = "Jamal loves green but not as much as he loves orange")
+            ]
+
+            map_documents_chain.invoke({"context": docs, "question": "Who loves green?"})
+    """  # noqa: E501
+    validate_prompt(prompt, (DOCUMENTS_KEY,))
+    _document_prompt = document_prompt or PromptTemplate.from_template("{page_content}")
+
```
hwchase17, line 84: some comments in code would be nice

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -1,18 +1,197 @@
...
+    _format = partial(format_document_inputs, document_prompt=_document_prompt)
+    map_content_chain = _format | prompt | llm | StrOutputParser()
+    map_content_chain = map_content_chain.with_config(run_name="map_content")
+
+    assign_page_content: Runnable = RunnablePassthrough.assign(
+        page_content=map_content_chain
+    )
+    assign_page_content = assign_page_content.with_config(
```
hwchase17, line 92: can just add this to the expression above?

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -1,18 +1,197 @@
...
+        run_name="assign_page_content"
+    )
+    map_doc_chain = assign_page_content | _compile_document
+    map_docs_chain = map_doc_chain.with_config(run_name="map_document").map()
```
hwchase17, line 96: i'd do the map when you use it in the expression, it doesnt really make the chain

hwchase17 (commented):
comments with all the LCEL would be great - if we are asking people to look at source code and modify, needs to be understandable

libs/langchain/langchain/chains/combine_documents/base.py
baskaryan: it's imported in other files just not exposed in combine_documents.init. should still prefix with `_` in that case?

libs/langchain/langchain/chains/combine_documents/map_rerank.py
```diff
@@ -2,17 +2,185 @@
...
+    )
+
+    answer_chain = prompt | llm | _output_parser
+    answer_chain = answer_chain.with_config(run_name="answer_and_score")
+    map_chain = (format_as_list | answer_chain.map()).with_config(
+        run_name="answer_and_score_all"
+    )
+
+    assign_all_answers: Runnable = RunnableParallel(all_answers=map_chain)
+    assign_all_answers = assign_all_answers.with_config(run_name="assign_all_answers")
```
hwchase17, line 141: same line as above?

libs/langchain/langchain/chains/combine_documents/reduce.py
```diff
@@ -2,13 +2,166 @@
...
+    token_len_func: Optional[Callable[[str], int]] = None,
+) -> Runnable[Dict[str, Any], List[Document]]:
+    """Create
+
+    Args:
+        llm: Language model to use for collapsing document contents.
+        prompt: Prompt to use for collapsing document contents. Must accept "context" as
+            one of the input variables. The formatted documents will be passed in as
+            "context".
+        token_max: The maximum cumulative token length that the list of Documents can
```
hwchase17, line 49: no default?

libs/langchain/langchain/chains/combine_documents/reduce.py
```diff
@@ -2,13 +2,166 @@
...
+            from langchain.chains.combine_documents import create_collapse_documents_chain,
+
+            llm = ChatOpenAI(model="gpt-3.5-turbo")
+            extract_prompt = ChatPromptTemplate.from_template(
+                [
+                    ("system", "Given a user question, extract the most relevant parts of the following context:\n\n{context}"),
+                    ("human", "{question}"),
+                ]
+            )
+            collapse_documents_chain = create_collapse_documents_chain(llm, extract_prompt, token_max=4000)
```
hwchase17, line 86: show example of how to use?

libs/langchain/langchain/chains/combine_documents/reduce.py
```diff
@@ -2,13 +2,166 @@
...
+    BaseCombineDocumentsChain,
+    format_document_inputs,
+    format_documents,
+    validate_prompt,
+)
+
+""" --- LCEL Runnable chains --- """
+
+
+def create_collapse_documents_chain(
```
hwchase17, line 33: confusing that `create_collapse_documents_chain` is in reduce.py. move to new file `collapse.py`?

libs/langchain/langchain/chains/combine_documents/map_rerank.py
baskaryan: done for readability (turns into 4 lines if you do it all in one assignment)

libs/langchain/langchain/chains/combine_documents/reduce.py
baskaryan: seems like the kind of thing you'd almost always want to set yourself (since it depends on model), so default would obfuscate more than anything?

libs/langchain/langchain/chains/combine_documents/reduce.py
baskaryan: should the legacy collapse functionality also go there (`collapse_docs`, `split_list_of_docs`)?

libs/core/langchain_core/runnables/base.py
```diff
@@ -753,7 +753,7 @@ def with_config(
         config: Optional[RunnableConfig] = None,
         # Sadly Unpack is not well supported by mypy so this will have to be untyped
         **kwargs: Any,
-    ) -> Runnable[Input, Output]:
+    ) -> RunnableBinding[Input, Output]:
```
nfcampos, line 756: why?

baskaryan: was causing typing issues (if somethings RunnableSerializable and then you update with `with_config` it was becoming just Runnable)

baskaryan: also isn't it best practice to type outputs as specifically as possible

nfcampos: then let's make it return RunnableSerializable. The idea with returning a looser type is to allow subclasses to override it (eg because they handle config themselves

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -1,18 +1,194 @@
...
+
+            docs = [
+                Document(page_content="Jesse loves red but not yellow"),
+                Document(page_content = "Jamal loves green but not as much as he loves orange")
+            ]
+
+            map_documents_chain.invoke({"context": docs, "question": "Who loves green?"})
+    """  # noqa: E501
+    validate_prompt(prompt, (DOCUMENTS_KEY,))
+    _document_prompt = document_prompt or PromptTemplate.from_template("{page_content}")
```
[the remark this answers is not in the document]
baskaryan: mypy complains later on

libs/langchain/langchain/chains/combine_documents/map_reduce.py
```diff
@@ -1,18 +1,194 @@
...
+    assign_page_content: Runnable = RunnablePassthrough.assign(
+        page_content=map_content_chain
+    )
+    assign_page_content = assign_page_content.with_name("assign_page_content")
+    map_doc_chain = (assign_page_content | _compile_document).with_name("map_document")
+
+    format_as_list = partial(
+        format_document_inputs_as_list, document_prompt=document_prompt
+    )
+    return (format_as_list | (map_doc_chain.map())).with_name("map_documents_chain")
```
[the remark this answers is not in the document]
baskaryan: otherwise whole sequence was being mapped iirc

PR #15079, opened 2023-12-22
core:  update json output parser

apisani1:
- **Description:** changed json.py to handle additional cases of partial json string to be parsed,  basically by dropping the last character in the string until a valid json string is found or the string is empty. Also added additional test cases.
  
  - **Issue:** function parse_partial_json could not parse cases where the key is present but the value is not.

### DIFF
--- a/libs/core/langchain_core/output_parsers/json.py
+++ b/libs/core/langchain_core/output_parsers/json.py
@@ -102,9 +102,21 @@ def parse_partial_json(s: str, *, strict: bool = False) -> Any:
     if is_inside_string:
         new_s += '"'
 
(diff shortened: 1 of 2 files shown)


hwchase17:
i like the idea, cc @nfcampos 
i think this pr needs to be updated to merge in some changes we made

apisani1:
I rebased it again but when  I merged my proposal it failed one test: "test_raises_error - Failed: DID NOT RAISE <class 'Exception'>"

nfcampos:
@hwchase17 @apisani1 rebased and updated to fix test

nfcampos (approved):

PR #15081, opened 2023-12-22
Random length base example selector

Dmaturana81:
Description: The length example selector always retrieves the same examples, going from top to bottom. In the case you have a lot of examples, there will not be any variation. I just copied and modified this length selector to do a random length base examples selector, so for each prompt, the examples selected will be random

    Dependencies: Added the random import, not dependencies changed

    Tag maintainer: for a quicker response, tag the relevant maintainer (see below),


baskaryan (commented):
could we add some simple unit tests?

hwchase17:
going to close this. it needs tests at a minimum. we are also going to think more deeply about example selectors soon, we will reference this then

PR #15082, opened 2023-12-22
Update regex in output parser

sharrajesh:
The regex used to match "Action" and "Action Input" in the output parser has been updated. Previously, the regex did not correctly handle multi-line inputs for "Action Input". The updated code uses the 're.DOTALL' flag to ensure multi-line inputs are correctly captured.


hwchase17:
pulled in some tests from https://github.com/langchain-ai/langchain/pull/14926

PR #15083, opened 2023-12-22
Improve graph repr for runnable passthrough and itemgetter

nfcampos:


libs/core/langchain_core/runnables/base.py
```diff
@@ -2024,7 +2024,7 @@ def get_output_schema(
     ) -> Type[BaseModel]:
         # This is correct, but pydantic typings/mypy don't think so.
         return create_model(  # type: ignore[call-overload]
-            "RunnableParallelOutput",
+            "RunnableMapOutput",
```
baskaryan, line 2027: double checking, no backwards compat concerns here right?

baskaryan (approved):

libs/core/langchain_core/runnables/base.py
nfcampos: nah just names

PR #15103, opened 2023-12-23
community: normalize bedrock embeddings

dmenini:
In this PR I added a post-processing function to normalize the embeddings. This happens only if the new `normalize` flag is `True`.


hwchase17 (commented):
should be added to embed_document as well?

dmenini:
@hwchase17 Could this PR be merged? Thanks for the review!

nicoloboschi (commented):
what about adding a unit test that the normalization happens as expected ?

hwchase17 (approved):

Issue #15086, opened 2023-12-23
DOC: Documentation Update Needed for 'Return Source Documents' Functionality 


wu-changxing:
#### Issue Description
- **Overview**: The current documentation for the 'Return Source Documents' functionality seems to be outdated or incorrect. The provided code snippet results in errors when executed.

https://python.langchain.com/docs/integrations/providers/vectara/vectara_chat#conversationalretrievalchain-with-question-answering-with-sources

- **Details**: 
    - The current code:
        ```python
        chat_history = []
        query = "What did the president say about Ketanji Brown Jackson"
        result = bot({"question": query, "chat_history": chat_history})
        ```
      produces the following error:
        ```
        /python3.9/site-packages/langchain/memory/chat_memory.py", line 29, in _get_input_output
        raise ValueError(f"One output key expected, got {outputs.keys()}")
        ValueError: One output key expected, got dict_keys(['answer', 'source_documents'])
        ```
- **Examples**: This issue occurs when trying to use the 'Return Source Documents' as outlined in the current documentation.

#### Additional Information
- **Related Issue**: This documentation update is related to the issue raised in https://github.com/langchain-ai/langchain/issues/2256.

#### Suggested Fix
- Update the documentation with the correct code snippet:
    ```python
    memory = ConversationBufferMemory(memory_key="chat_history", input_key='question', output_key='answer', return_messages=True)
    bot = ConversationalRetrievalChain.from_llm(llm, retriever, memory=memory, return_source_documents=True)
    result = bot({"question": query,})
    
    ```
- This revision correctly handles the output and does not produce the aforementioned error.

cyai:
#15106 This might help

PR #15106, opened 2023-12-23
docs: updated document for 'Return Source Documents' Functionality

cyai:
- **Description:** updated the outdated code in the document that was generating the error, 
  - **Issue:** #15086 ,
  - **Dependencies:** N/A,
  - **Twitter handle:** [@vardhaman722](https://twitter.com/vardhaman722)


hwchase17 (approved):

hwchase17:
even easier - and better - fix would be to just call `bot.invoke({})`

PR #15108, opened 2023-12-23
refactored `CacheBackedEmbeddings`, `langchain.storage`

leo-gan:
Refactored `CacheBackedEmbeddings`. Moved it from `langchain` to `langchain_core` (by @eyurtsev recommendation). As a result, [`langchain.embeddings` namespace completely removed](https://api.python.langchain.com/en/latest/langchain_api_reference.html#module-langchain.embeddings).
There was a dependency on `BaseStore`. So, I have also moved `langchain.storage` core artifacts to `langchain_core.storage`. [`langchain.storage` also](https://api.python.langchain.com/en/latest/langchain_api_reference.html#module-langchain.storage) completely removed.


leo-gan:
@baskaryan  Any comments?

leo-gan:
@baskaryan Any comments?

leo-gan:
@baskaryan Any comments?

efriis:
@eyurtsev not sure these should be in core. I think these are better in `langchain` for now as implementations (esp things like File store). Happy to discuss/reopen if you strongly disagree.

leo-gan:
@efriis The main reason for this change is the current `langchain.embeddings` in [API Ref](https://api.python.langchain.com/en/latest/langchain_api_reference.html#module-langchain.embeddings) looks too tiny. We keep the full namespace for a single class. That's a waste. 
A full namespace for 1-3 entities — that's not a good design. That's a direct way to spaghetti code.

PR #15111, opened 2023-12-23
community:Lazy load wikipedia dump file

cjaniake:
**Description:** the MWDumpLoader implementation currently does not support the lazy_load method, and the files are usually very large. We are proposing refactoring the load function, extracting two private functions with the functionality of loading the dump file and parsing a single page, to reuse the code in the lazy_load implementation.


hwchase17 (commented):
looks good, but can we have the load function just call lazy load? that way theres no duplicated code there

hwchase17 (approved):
thanks!

PR #14025, opened 2023-11-29
Update text_splitter.py

govinda18:
Description: Fixes a minor bug to make HTMLHeaderTextSplitter work.
Issue: https://github.com/langchain-ai/langchain/issues/14024
Dependencies: none,

PR #15113, opened 2023-12-24
refactored `text_splitter`

leo-gan:
All text splitters are in a single file. It is hard to extract a base classes to `langchain_core`.
- split code to several files
- added interface classes `BaseTextSplitter`, `BaseTextToDocumentsSplitter` and moved them into `core`

This allows for adding community splitters.


leo-gan:
@govinda18 Could you, please, check the error in the `test_html_header_text_splitter` unit test? Is it something with `lxml` library? Thanks!

govinda18:
Sorry for the delayed response as I was out over the last couple of weeks. @baskaryan may be able to help here as he added that test in this PR - https://github.com/langchain-ai/langchain/pull/14025

leo-gan:
@baskaryan Can you, please, check the error in the test_html_header_text_splitter unit test? Is it something with lxml library? Thanks!

leo-gan:
@baskaryan  any comments?

libs/langchain/langchain/text_splitter.py
efriis, line 1: only export the text splitters. No need to export the TS generic and that kind of thing

docs/docs/integrations/providers/spacy.mdx
```diff
@@ -16,7 +16,8 @@ pip install spacy
 See a [usage example](/docs/modules/data_connection/document_transformers/split_by_token#spacy).
 
 ```python
-from langchain.text_splitter import SpacyTextSplitter
+
+from langchain.text_splitters.spacy import SpacyTextSplitter
```
efriis, line 20: Old import path still works

docs/docs/integrations/providers/elasticsearch.mdx
```diff
@@ -34,7 +34,7 @@ The vector store is a simple wrapper around Elasticsearch. It provides a simple
 from langchain_community.vectorstores import ElasticsearchStore
 
 from langchain_community.document_loaders import TextLoader
-from langchain.text_splitter import CharacterTextSplitter
+from langchain.text_splitter.character import CharacterTextSplitter
```
efriis, line 37: Old import path still works

libs/langchain/langchain/text_splitter.py
efriis: Also @baskaryan do we have a deprecation format for moves like this text_splitter -> text_splitters

libs/langchain/langchain/text_splitters/__init__.py
efriis, line 1: Same comment - no need to re-export non-text splitters

libs/langchain/tests/integration_tests/test_text_splitter.py
```diff
@@ -5,8 +5,8 @@
 from langchain.text_splitter import (
     CharacterTextSplitter,
     SentenceTransformersTokenTextSplitter,
-    TokenTextSplitter,
 )
+from langchain.text_splitters.token import TokenTextSplitter
```
efriis, line 9: use old import path

libs/langchain/tests/unit_tests/test_text_splitter.py
```diff
@@ -1053,16 +1054,16 @@ def test_md_header_text_splitter_preserve_headers_1() -> None:
     ]
     markdown_splitter = MarkdownHeaderTextSplitter(
         headers_to_split_on=headers_to_split_on,
-        strip_headers=False,
+        return_each_line=False,
```
efriis, line 1057: is this a breaking change?

efriis: If so please revert

libs/langchain/tests/unit_tests/test_text_splitter.py
```diff
@@ -1091,27 +1092,22 @@ def test_md_header_text_splitter_preserve_headers_2() -> None:
...
     )
     output = markdown_splitter.split_text(markdown_document)
     expected_output = [
         Document(
-            page_content="# Foo  \n## Bar  \nHi this is Jim  \nHi this is Joe",
```
efriis, line 1099: let's leave all these tests as-is. I'm a little suspicious this PR has breaking changes because of these test changes.

efriis (commented):
Some comments. Broadly:

- let's leave unit-tests as-is, and only add unit tests for new things if you want to test them. This should not be a breaking change. If you want to do a follow-up PR updating import paths separately to be from `text_splitters` instead of `text_splitter`, that's ok. But for this PR, would love to make sure all the `text_splitter` import paths still work.

leo-gan:
@efriis I've rolled back all changes in tests and in .mdx files.

baskaryan:
text splitters have been moved to their own package #18346

leo-gan:
@baskaryan Thanks! A separate package makes a perfect sense.

PR #15119, opened 2023-12-24
Refactor: use SecretStr for Predibase llms

chyroc:


libs/community/langchain_community/llms/predibase.py
```diff
@@ -13,7 +13,7 @@ class Predibase(LLM):
     """
 
     model: str
-    predibase_api_key: str
+    predibase_api_key: SecretStr
```
hwchase17, line 16: is this backwards compatible?

hwchase17: tests suggest it is, lgtm!

hwchase17 (approved):

PR #15142, opened 2023-12-25
Feat: add TinyImage tool to compress images

chyroc:
@hwchase17

### DIFF
--- a/libs/community/langchain_community/tools/tiny_image/tool.py
+++ b/libs/community/langchain_community/tools/tiny_image/tool.py
@@ -1,0 +1,36 @@
+"""Tool for the TinyImage API."""
+
+from typing import Optional
+
+from langchain_core.callbacks import CallbackManagerForToolRun
+from langchain_core.tools import BaseTool
+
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/tools/tiny_image/tool.py
```diff
@@ -0,0 +1,36 @@
...
+
+from typing import Optional
+
+from langchain_core.callbacks import CallbackManagerForToolRun
+from langchain_core.tools import BaseTool
+
+from langchain_community.utilities.tiny_image import TinyImageAPIWrapper
+
+
+class WikipediaQueryRun(BaseTool):
```
hwchase17, line 11: name

hwchase17 (commented):
how do you imagine this being used?

chyroc:
> how do you imagine this being used?

Compress the image and send it to the user as a thumbnail instead of sending the original image to the user. When the user clicks, the original image is displayed.

hwchase17:
tools are largely meant to be called by LLMs, and to me this seems like functionality someone would want to do automatically rather than leave to an LLM to call

PR #15152, opened 2023-12-25
refactored `BaseChatMemory`, `ChatMessageHistory`, `get_prompt_input_key`

leo-gan:
Moved `BaseChatMemory`, `ChatMessageHistory`, `get_prompt_input_key` into `langchain_core`
`BaseChatMemory`, `ChatMessageHistory` are base classes. The `get_prompt_input_key` was moved because of dependency.
Also fixed an error in imports in `langchain_experimental/autonomous_agents/autogpt/memory.py`


hwchase17 (commented):
we probably dont want to put the BaseMemory class here (not sure its production ready, will likely deprecate)

Why do we want ChatMessageHistory here? I'm fine leaving it community or Langchain

leo-gan:
Was reimplemented in  #15437

PR #15166, opened 2023-12-26
core: add .images + .message property to BaseMessage

shroominic:
**Description:** 
Adding helper methods for improving implementation speed and code clarity.

**Issue:**
I write a lot of boiler plate and my code feels chaotic since the new vision message schemas.


hwchase17:
i am going to close this. we are actively thinking a lot about messages and i dont think this has popped up as a desired interface

PR #15212, opened 2023-12-27
community: avoid KeyError when language not in LANGUAGE_SEGMENTERS

dzlab:
**Description:**

Handle unsupported languages in same way as when none is provided 
 
**Issue:**

The following line will throw a KeyError if the language is not supported.
```python
self.Segmenter = LANGUAGE_SEGMENTERS[language]
```
E.g. when using `Language.CPP` we would get `KeyError: <Language.CPP: 'cpp'>`


hwchase17 (commented):
wouldnt you want to error though? shouldnt everything in LANGUAGE_EXTENSIONS be in LANGUAGE_SEGMENTERS?

hwchase17 (commented):
or alternatively, if you're passing in a Language that isnt supported wouldn't you want to know so that you dont assume its actually doing anything?

dzlab:
@hwchase17 the problem is the current default error message with `KeyError` seems more like a bug, the user may need to look at the code to understand what's going on. My initial change was to provide a better error message like this
```python
if language not in LANGUAGE_SEGMENTERS:
  raise Exception(f"No parser available for {language}")
```
but then saw that when language is `None` the current implementation returns a document with the whole source so I thought we should do the same when language is not supported yet with
```python
if language is None or language not in LANGUAGE_SEGMENTERS
```
Let me know which approach is more preferable and I will update the PR accordingly. Or if the current behavior is good enough then I can just close this PR.

dzlab:
@hwchase17 between if I had to add a parser for C++ will it be preferable to use `clang` or `pycparser` or maybe something else?

hwchase17:
> @hwchase17 the problem is the current default error message with `KeyError` seems more like a bug, the user may need to look at the code to understand what's going on. My initial change was to provide a better error message like this
> 
> ```python
> if language not in LANGUAGE_SEGMENTERS:
>   raise Exception(f"No parser available for {language}")
> ```
> 
> but then saw that when language is `None` the current implementation returns a document with the whole source so I thought we should do the same when language is not supported yet with
> 
> ```python
> if language is None or language not in LANGUAGE_SEGMENTERS
> ```
> 
> Let me know which approach is more preferable and I will update the PR accordingly. Or if the current behavior is good enough then I can just close this PR.

I'd prefer the first - would rather loudly error than silently fail in this case. thansk!

dzlab:
@hwchase17 thanks for the feedback, I've updated the PR accordingly.

hwchase17 (commented):
can we do this check in __init__? that way it fail faster

from my understanding, `language` here is set in two ways, either self.language or LANGUAGE_EXTENSIONS. everything in LANGUAGE_EXTENSIONS is in LANGUAGE_SEGMENTERS, so the mismatch only occurs when a self.language is specified poorly, which we can catch in __init__. please lmk if im misunderstanding

dzlab:
@hwchase17 I do agree, the issue may happen if the language was set in `__init__`, I've update the PR.

hwchase17 (approved):

PR #15216, opened 2023-12-27
Add `rag-pinecone-cohere-embed-rerank` template

samuelpath:
- **Description:** Following a private discussion with @rlancemartin and @SandraKublik, this PR adds a new template using:
  - Cohere Embed v3
  - Cohere Chat
  - Cohere Re-Rank

This template is very similar to the [Rag Pinecone Rerank template](https://github.com/langchain-ai/langchain/blob/master/templates/rag-pinecone-rerank) which uses OpenAI's API instead for the embeddings and chat endpoints.

I have tested it locally and it works as expected.

If this PR gets merged, we may want to rename the Rag Pinecone Rerank template to specify that OpenAI is used for the embeddings and chat (for instance to something like: `rag-pinecone-openai-embed-rerank`).

- **Issue:** N/A
- **Dependencies:** No dependencies added for this change.
- **Twitter handle:** [@smlpth](https://twitter.com/smlpth)

PR #15219, opened 2023-12-27
Add `rag-conversation-cohere` template

samuelpath:
- **Description:** Following a private discussion with @rlancemartin and @SandraKublik, this PR adds a new template for RAG Conversation using Cohere Embed v3 and Cohere Chat.

This template is very similar to the [Rag Conversation template](https://github.com/langchain-ai/langchain/tree/master/templates/rag-conversation) which uses OpenAI's API instead for the embeddings and chat endpoints.

I have tested it locally and it works as expected.

If this PR gets merged, we may want to rename the Rag Conversation template to specify that OpenAI is used for the embeddings and chat (for instance to something like: `rag-conversation-openai`).

- **Issue:** N/A
- **Dependencies:** No dependencies added for this change.
- **Twitter handle:** [@smlpth](https://twitter.com/smlpth)

### DIFF
--- a/templates/docs/INDEX.md
+++ b/templates/docs/INDEX.md
@@ -6,12 +6,11 @@ Highlighting a few different categories of templates
 
 These are some of the more popular templates to get started with.
 
-- [Retrieval Augmented Generation Chatbot](../rag-conversation): Build a chatbot over your data. Defaults to OpenAI and Pinecone.
+- [Retrieval Augmented Generation Chatbot](../rag-conversation): Build a chatbot over your data. Defaults to OpenAI (with a toggle switch to Cohere) and Pinecone.
 - [Extraction with OpenAI Functions](../extraction-openai-functions): Do extraction of structured data from unstructured data. Uses OpenAI function calling.
 - [Local Retrieval Augmented Generation](../rag-chroma-private): Build a chatbot over your data. Uses only local tooling: Ollama, GPT4all, Chroma.
 - [OpenAI Functions Agent](../openai-functions-agent): Build a chatbot that can take actions. Uses OpenAI function calling and Tavily.
 - [XML Agent](../xml-agent): Build a chatbot that can take actions. Uses Anthropic and You.com.
-
 
 ## 📥 Advanced Retrieval
 
--- a/templates/rag-conversation/README.md
+++ b/templates/rag-conversation/README.md
@@ -1,7 +1,10 @@
 
 # rag-conversation
 
-This template is used for [conversational](https://python.langchain.com/docs/expression_language/cookbook/retrieval#conversational-retrieval-chain) [retrieval](https://python.langchain.com/docs/use_cases/question_answering/), which is one of the most popular LLM use-cases. 
+This template is used for [conversational](https://python.langchain.com/docs/expression_language/cookbook/retrieval#conversational-retrieval-chain) [retrieval](https://python.langchain.com/docs/use_cases/question_answering/), which is one of the most popular LLM use-cases.
+
+It uses OpenAI for both the embeddings and the chat by default.
+You can also use Cohere for the embeddings and the chat by setting the environment variable `EMBEDDING_TYPE` to `Cohere`.
 
 It passes both a conversation history and retrieved documents into an LLM for synthesis.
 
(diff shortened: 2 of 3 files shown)


hwchase17:
whats the difference between this and https://github.com/langchain-ai/langchain/pull/15216?

samuelpath:
Thanks for your comment @hwchase17.

This current PR (https://github.com/langchain-ai/langchain/pull/15219) is similar to the [Rag Conversation template](https://github.com/langchain-ai/langchain/tree/master/templates/rag-conversation): used for conversational retrieval, pasing both a conversation history and retrieved documents into an LLM for synthesis.

The other PR (https://github.com/langchain-ai/langchain/pull/15216) is similar to the [Rag Pinecone Rerank template](https://github.com/langchain-ai/langchain/blob/master/templates/rag-pinecone-rerank): performing RAG using Pinecone  with re-ranking on returned documents.

In both cases, I start by copying the template built around OpenAI then adapt it to make it work using the Cohere Embed v3 model.

Does that make more sense to you?

rlancemartin:
> This current PR (#15219) is similar to the [Rag Conversation template](https://github.com/langchain-ai/langchain/tree/master/templates/rag-conversation): used for conversational retrieval, passing both a conversation history and retrieved documents into an LLM for synthesis.

Got it, so this is Conversational Retrieval using Cohere embd and chat model.

efriis:
Hey @samuelpath ! I think this would be better modifying the `rag-conversation` template to have configurable alternatives for cohere models instead of a completely separate template. The idea of templates is people can modify them to suit their use case, and that format would let people experiment with both in the same surrounding app before deleting all but one!

samuelpath:
Thanks @efriis, your suggestion makes total sense. Actually I wasn't satisfied with the amount of duplication in my PR.

I've implemented the toggle switch using an environment variable called `EMBEDDING_TYPE` that would default to `OpenAI` but that could be overriden to `Cohere`.

Is that what you had in mind?

I also updated the templates index README and the rag-conversation README in light of this change.

efriis:
Hey @samuelpath ! Actually was thinking more of using a `configurable_alternative` like this one: https://templates.langchain.com/?integration_name=pirate-speak-configurable

Let me know if you need any help/guidance modifying!

samuelpath:
Thank you @efriis for your feedback.

I tried to implement your suggestion using `configurable_alternative` like in [pirate-speak-configurable](https://templates.langchain.com/?integration_name=pirate-speak-configurable), but it turned out to be trickier than expected.

If I use the code below:

```python
chat_model = ChatOpenAI().configurable_alternatives(
    ConfigurableField(id="llm_provider"),
    default_key="openai",
    cohere=ChatCohere,
)

embedding_model = OpenAIEmbeddings().configurable_alternatives(
    ConfigurableField(id="llm_provider"),
    default_key="openai",
    cohere=CohereEmbeddings,
)
```

The `chat_model` is properly initialized but during the `embedding_model` initialization, I get the following error:

```sh
AttributeError: 'OpenAIEmbeddings' object has no attribute 'configurable_alternatives'
```

I see that `configurable_alternatives` is defined in the the `RunnableSerializable` class [here](https://github.com/langchain-ai/langchain/blob/a989f82027d42e893a60b8ba56da703ab22a22ab/libs/core/langchain_core/runnables/base.py#L1634-L1672) so I understand that a class needs to have `RunnableSerializable` in its inheritance chain in order to be able to use it.

For `ChatOpenAI`, this is straightforward with this inheritance chain:

```
RunnableSerializable
    ↓
BaseLanguageModel
    ↓
BaseChatModel
    ↓
ChatOpenAI
```

But for `OpenAIEmbeddings`, I see that it is inheriting from `BaseModel` and `Embeddings`:

https://github.com/langchain-ai/langchain/blob/85a4594ed7dbde9d257e230d5d789b68c42421cf/libs/community/langchain_community/embeddings/openai.py#L146

https://github.com/langchain-ai/langchain/blob/85a4594ed7dbde9d257e230d5d789b68c42421cf/libs/community/langchain_community/embeddings/openai.py#L23-L24

I'm not quite sure what `BaseModel` refers to since when I look for `class BaseModel(` in the entire repo, I don't see any relevant implementation:

image

As for the `Embeddings` class, it's only an abstract interface:

https://github.com/langchain-ai/langchain/blob/eb5e250188b1ab26f4d57d20245ddf45b8f5cdff/libs/core/langchain_core/embeddings.py#L7-L8

So I see that `OpenAIEmbeddings` is not inheriting from `RunnableSerializable` and therefore does not have access to `configurable_alternatives`.

So here my question becomes: what would be the best way to make it inherit from `RunnableSerializable` so that it has access to `configurable_alternatives`?

Now I also see that most templates are using chat models and embeddings from `langchain_community` which have been deprecated since `v0.1.0` 3 weeks ago:

https://github.com/langchain-ai/langchain/blob/85a4594ed7dbde9d257e230d5d789b68c42421cf/libs/community/langchain_community/chat_models/openai.py#L147-L150

https://github.com/langchain-ai/langchain/blob/85a4594ed7dbde9d257e230d5d789b68c42421cf/libs/community/langchain_community/embeddings/openai.py#L141-L146

So I imagine that the solution is **not** to refactor the deprecated `OpenAIEmbeddings` to make it configurable, but rather to think about how to make the new [OpenAIEmbeddings](https://github.com/langchain-ai/langchain/blob/61e876aad8353891a3d68ae32286e30d82ccb1ae/libs/partners/openai/langchain_openai/embeddings/base.py) class inherit from `RunnableSerializable`.

This doesn't look straightforward to me… Do you have some advice as to how I could do it?

Thank you for your help!

Issue #15221, opened 2023-12-27
Enable manual override of vector search query for controlled RAG


devinbost:
### Feature request

I need a mechanism to allow more control over the ANN search performed for a given RAG chain. Consider the initial example:
```
retriever = vectorstore.as_retriever()
template = """You're a helpful assistant who is great at code generation. Don't give me any explanation or summary. I'll give you some examples that may or may not be relevant, and I want you to use the examples to write code that solves the provided problem. Return only the code that solves the problem.

PROBLEM:
{problem}

EXAMPLES:
{context}

ANSWER: 
"""

prompt = ChatPromptTemplate.from_template(template)

model = ChatOpenAI()
chain = (
    {"context": retriever, "problem": RunnablePassthrough()}
    | prompt
    | model
    | StrOutputParser()
)
chain.invoke("Generate an example Python method that uses LCEL to write a CQL query")
```
This approach assumes that the question will be used for creating the embedding. However, consider something like this:

```
retriever = vectorstore.as_retriever(ann_query="LCEL, AstraDB, CQL")
```
In this situation, when the retriever is invoked to embed the query, instead of performing vector search on the embedding of the very wordy

> "Generate an example Python method that uses LCEL to write a CQL query" 

I want vector search to perform ANN on:

> "LCEL, AstraDB, CQL"

so that I have a greater likelihood of having the right docs stuffed into the prompt for the LLM to solve the problem, which was:
> Generate an example Python method that uses LCEL to write a CQL query

### Motivation

RAG results can be poor when the human input is very wordy or contains more info (for the LLM) than we want the vector store to search for. We need a mechanism to allow separation between the vector search query and the LLM query. 

### Your contribution

I will create a PR.

PR #15223, opened 2023-12-27
core + community: Added ANN override to VectorStore [new feature]

devinbost:
Addresses: https://github.com/langchain-ai/langchain/issues/15221
(Most of the context of this PR exists on that issue.)

Adds this method to VectorStore:
```
    @abstractmethod
    def similarity_search_with_override(
        self, query: str, search_override: str, k: int = 4, **kwargs: Any
    ) -> List[Document]:
        """Return docs most similar to query."""
```

Updates _get_relevant_documents() on VectorStore to include search_override, like this:
```
        elif self.search_type == "similarity_with_override":
            docs = self.vectorstore.similarity_search_with_override(query, self.search_override, **self.search_kwargs)
```
Also adds the corresponding method to AstraDB (as an example):

```
    def similarity_search_with_override(
            self,
            query: str,
            search_override: str,
            k: int = 4,
            filter: Optional[Dict[str, str]] = None,
            **kwargs: Any,
    ) -> List[Document]:
        embedding_vector = self.embedding.embed_query(search_override)
        return self.similarity_search_by_vector(
            embedding_vector,
            k,
            filter=filter,
        )
```

When I try to run the tests from root, I get:

```
% poetry install --with test
Installing dependencies from lock file

No dependencies to install or update

% make test
make: *** No rule to make target `test'.  Stop.
```
When I try to run the tests from core, I get:
```
% make test
poetry run pytest tests/unit_tests/
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --snapshot-warn-unused
  inifile: /Users/devin.bost/proj/repos/langchain/libs/core/pyproject.toml
  rootdir: /Users/devin.bost/proj/repos/langchain/libs/core

make: *** [test] Error 4
```

I tried the steps in the [guide](https://python.langchain.com/docs/contributing/testing), but I'm still not getting tests to run. Some guidance would be much appreciated. 

I'd also appreciate some guidance on how I might test this new functionality in LCEL.

### DIFF
--- a/libs/community/langchain_community/vectorstores/astradb.py
+++ b/libs/community/langchain_community/vectorstores/astradb.py
@@ -128,20 +128,20 @@ class AstraDB(VectorStore):
             return {f"metadata.{mdk}": mdv for mdk, mdv in filter_dict.items()}
 
     def __init__(
-        self,
-        *,
-        embedding: Embeddings,
-        collection_name: str,
-        token: Optional[str] = None,
-        api_endpoint: Optional[str] = None,
-        astra_db_client: Optional[Any] = None,  # 'astrapy.db.AstraDB' if passed
-        namespace: Optional[str] = None,
-        metric: Optional[str] = None,
-        batch_size: Optional[int] = None,
-        bulk_insert_batch_concurrency: Optional[int] = None,
-        bulk_insert_overwrite_concurrency: Optional[int] = None,
-        bulk_delete_concurrency: Optional[int] = None,
-        pre_delete_collection: bool = False,
+            self,
+            *,
+            embedding: Embeddings,
+            collection_name: str,
+            token: Optional[str] = None,
+            api_endpoint: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


hwchase17:
not sure i fully understand the motivation. in what context would you want to hardcode the retrieval string? doesn't that defeat the purpose of doing retrieval?

devinbost:
@hwchase17 My goal was to decouple the ANN search query from the input query, but it looks like LCEL can enable that as well. Am I understanding correctly from the following example that it's possible to send one string to vector search and a different string to the LLM? https://python.langchain.com/docs/expression_language/how_to/map#using-itemgetter-as-shorthand

devinbost:
Looks like the RunnableBranch will satisfy my requirement, so I'm closing this PR.

I'm quite impressed at how deeply you thought through the design of LCEL.

PR #15226, opened 2023-12-27
Implement nicer runnable seq constructor, Propagate name through Runn…

nfcampos:
…ableBinding


libs/core/langchain_core/runnables/base.py
```diff
@@ -1448,6 +1465,29 @@ def mul_two(x: int) -> int:
     last: Runnable[Any, Output]
     """The last runnable in the sequence."""
 
+    def __init__(self, *steps: RunnableLike, name: Optional[str] = None) -> None:
```
baskaryan, line 1468: should we still accept **kwargs and parse them for backwards compatibility

nfcampos: Done

baskaryan (approved):

PR #15229, opened 2023-12-27
Add .pick and .assign methods to Runnable

nfcampos:


eyurtsev:
Is there a way to pick and rename?

baskaryan (commented):
are we sure
 ```python
x.assign(foo=bar).pick(["foo", "baz"]) | y  | z
```
 is clearer than 
 ```python
x | RunnableAssign(foo=bar) | RunnablePick(["foo", "baz"]) | y | z
```
? kinda nice only having one chaining operator (`|`) imo. but if we're going the route of having instance methods that are also chaining operators should we also add `.append()` and `.extend()` so that if someone wants they can just use that syntax?
```python
x.assign(foo=bar).pick(["foo", "baz"]).extend([y, z])
```

libs/core/langchain_core/runnables/base.py
```diff
@@ -396,6 +396,30 @@ def __ror__(
         """Compose this runnable with another object to create a RunnableSequence."""
         return RunnableSequence(first=coerce_to_runnable(other), last=self)
 
+    def pick(self, keys: Union[str, List[str]]) -> Runnable[Any, Any]:
```
baskaryan, line 399: what if this was `*keys: str`?

baskaryan:
> Is there a way to pick and rename?

maybe better to add a separate `.rename(foo="bar", baz="buz", ...)` and let user chain pick and rename themselves?

libs/core/langchain_core/runnables/base.py
nfcampos: A single key gets you the value itself whereas a list of key(s) gets you a dict with those keys

baskaryan: could do `(self, *keys: str, key: Optional[str]=None)`?

baskaryan: separately should we support `name: Optional[str]=None` and pass that to RunnableAssign?

baskaryan: or just pass through **kwargs

PR #15231, opened 2023-12-27
refactored `output_parsers`

leo-gan:
Several output parsers belong to the `langchain_community` package.
- Created `langchain_community.output_parsers` namespace
- Moved several parsers to community package


libs/community/langchain_community/output_parsers/openai_functions.py
```diff
@@ -0,0 +1,174 @@
+import copy
```
hwchase17, line 1: we want to keep this in LangChain

openai is becoming a standard-ish interface that more than one provider is implementing, so we will likely keep utils for openai in langchain

libs/community/langchain_community/output_parsers/openai_tools.py
```diff
@@ -0,0 +1,61 @@
+import copy
```
hwchase17, line 1: same reason as above, we want to keep this in langchain

libs/community/langchain_community/output_parsers/pandas_dataframe.py
```diff
@@ -0,0 +1,176 @@
+import re
```
hwchase17, line 1: lets keep this in langchain for now.

leo-gan: Nah. I've made a research before moving pandas parser. We have the pandas data_loader in community package and nothing in core. The only pandas dependent piece in langchain was this parser. That's why.

libs/community/langchain_community/output_parsers/openai_functions.py
leo-gan: It is a stab in the whole idea of the `community` package. Meta, Alibaba, Baidu, Google, Antrophic would not be happy by this, no? LangChain has 5M downloads/month, it is a lot of leverage that should not be easily passed to one partner company.
The second reason is a more important one. There will be another better model from a different company in several months. "Standard" should be used for something more stable.

hwchase17: their interface is adopted by many other model providers - there are many "open ai drop ins", the meta/anthropic/etc interface is not

in the future (langchain 0.9.0 or whatnot) we can update what interfaces are the standard, but for now, openai function calling is a standard

leo-gan: Yeah, OpenAI functions are almost a standard. Not sure about OpenAI tools.

leo-gan:
@hwchase17 The OpenAI and pandas parsers are rolled back :)

hwchase17 (commented):
thanks. @baskaryan this lgtm, but we'll need to remember to bump the langchain-community version

libs/langchain/langchain/output_parsers/pandas_dataframe.py
baskaryan, line 1: should this file be in community?

leo-gan: @baskaryan As per @hwchase17 it could be in [langchain for now](https://github.com/langchain-ai/langchain/pull/15231#discussion_r1437261397).

leo-gan:
It is conflicting with too many new updates. I'll redo it from cratch

PR #15238, opened 2023-12-28
Retrieval Docs Revamp

hwchase17:


docs/docs/modules/data_connection/document_transformers/recursive_text_splitter.ipynb
baskaryan, line 1: unrelated to this pr but why don't we try to split on ".?!" before " "

hwchase17: good q, we could

Issue #15234, opened 2023-12-27
MlflowEmbeddings: input_type argument is missing, required by Cohere embeddings models


elucherini:
### System Info

python = "3.11"
langchain = "0.0.352"
cohere = "4.39"
mlflow = {extras = ["genai"], version = "2.9.2"}

### Who can help?

@harupy
@hwchase17
@agola11

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [X] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

I followed the official example for embeddings here, except that I am using Cohere instead of OpenAI: https://python.langchain.com/docs/integrations/providers/mlflow

<details>
<summary>Click for specific steps</summary>

More specifically, I installed mlflow genai and set my `COHERE_API_KEY` environment variable:
```bash
pip install 'mlflow[genai]'
export COHERE_API_KEY=...
```

I created `config.yaml` like so:
```yaml
endpoints:
  - name: completions
    endpoint_type: llm/v1/completions
    model:
      provider: cohere
      name: command
      config:
        cohere_api_key: $COHERE_API_KEY


  - name: embeddings
    endpoint_type: llm/v1/embeddings
    model:
      provider: cohere
      name: embed-english-light-v3.0
      config:
        cohere_api_key: $COHERE_API_KEY
```

I started the mlflow deployments server:
```bash
mlflow deployments start-server --config-path config.yaml
```

<details>
<summary>The server started as expected</summary>

```
xxx/python3.11/site-packages/pydantic/_internal/_config.py:321: UserWarning: Valid config keys have changed in V2:
* 'schema_extra' has been renamed to 'json_schema_extra'
  warnings.warn(message, UserWarning)
[2023-12-27 13:53:18 -0800] [22480] [INFO] Starting gunicorn 21.2.0
[2023-12-27 13:53:18 -0800] [22480] [INFO] Listening at: http://127.0.0.1:5000 (22480)
[2023-12-27 13:53:18 -0800] [22480] [INFO] Using worker: uvicorn.workers.UvicornWorker
[...]

PR #15242, opened 2023-12-28
community: allow additional kwargs in MlflowEmbeddings for compatibility with Cohere API

elucherini:
- **Description:** add support for kwargs in`MlflowEmbeddings` `embed_document()` and `embed_query()` so that all the arguments required by Cohere API (and others?) can be passed down to the server.
  - **Issue:** #15234 
  - **Dependencies:** MLflow with MLflow Deployments (`pip install mlflow[genai]`)

**Tests**
Now this code [adapted from the docs](https://python.langchain.com/docs/integrations/providers/mlflow#embeddings-example) for the Cohere API works locally.

```python
"""
Setup
-----
export COHERE_API_KEY=...
mlflow deployments start-server --config-path examples/deployments/cohere/config.yaml

Run
---
python /path/to/this/file.py
"""
embeddings = MlflowCohereEmbeddings(target_uri="http://127.0.0.1:5000", endpoint="embeddings")
print(embeddings.embed_query("hello")[:3])
print(embeddings.embed_documents(["hello", "world"])[0][:3])
```

Output
```
[0.060455322, 0.028793335, -0.025848389]
[0.031707764, 0.021057129, -0.009361267]
```


hwchase17 (commented):
i think prefereable to store as part of the class (so that we do not change the method signature)

elucherini:
I just changed the code to this, let me know what you think:

```python

class MlflowEmbeddings(Embeddings, BaseModel):
    ...
    """Parameters to use for queries."""
    query_params: Dict[str, str] = {}
    """Parameters to use for documents."""
    documents_params: Dict[str, str] = {}

    ...

    def embed(self, texts: List[str], params: Dict[str, str]) -> List[List[float]]:
        # Mostly the same as the old embed_documents
        embeddings: List[List[float]] = []
        for txt in _chunk(texts, 20):
            # Add parameters to inputs
            resp = self._client.predict(endpoint=self.endpoint, inputs={"input": txt, **params})
            embeddings.extend(r["embedding"] for r in resp["data"])
        return embeddings

    def embed_documents(self, texts: List[str]) -> List[List[float]]:
        # Pass in documents_params
        return self.embed(texts, params=self.documents_params)

    def embed_query(self, text: str) -> List[float]:
        # Pass in query_params
        return self.embed([text], params=self.query_params)[0]
```

I also added a class with the specific parameters needed by Cohere:

```python
class MlflowCohereEmbeddings(MlflowEmbeddings):
    query_params: Dict[str, str] = {"input_type": "search_query"}
    document_params: Dict[str, str] = {"input_type": "search_document"}
```

This could expand to other frameworks, whether the additional parameters are required (like in this case) or optional.

libs/langchain/langchain/embeddings/mlflow.py
```diff
@@ -1,3 +1,6 @@
-from langchain_community.embeddings.mlflow import MlflowEmbeddings
+from langchain_community.embeddings.mlflow import (
+    MlflowCohereEmbeddings,
+    MlflowEmbeddings,
+)
 
-__all__ = ["MlflowEmbeddings"]
+__all__ = ["MlflowEmbeddings", "MlflowCohereEmbeddings"]
```
hwchase17, line 6: should not update code in `langchain` package

elucherini: Fixed, only updating `community`

elucherini:
@hwchase17 gentle reminder

baskaryan (approved):

PR #15265, opened 2023-12-28
community:  Add ChatGLM3

169:
Add [ChatGLM3](https://github.com/THUDM/ChatGLM3) and updated [chatglm.ipynb](https://python.langchain.com/docs/integrations/llms/chatglm)


hwchase17 (commented):
langchain package should NOT be updated (only langchain-community)

169:
@hwchase17 Actually, I don't want to change the package dependencies either, but since `langchain` relies on modules from` langchain-community`, it indirectly necessitates the addition of `httpx`.

I'm wondering if the `langchain` project plans to use `httpx` as a replacement for `requests` in the future? If I use `requests`, it actually won't require additional modifications to the dependencies. I switched to httpx because the https://github.com/openai/openai-python uses it, and I feel like `httpx` is the trend for the future.

hwchase17:
> @hwchase17 Actually, I don't want to change the package dependencies either, but since `langchain` relies on modules from` langchain-community`, it indirectly necessitates the addition of `httpx`.
> 
> I'm wondering if the `langchain` project plans to use `httpx` as a replacement for `requests` in the future? If I use `requests`, it actually won't require additional modifications to the dependencies. I switched to httpx because the https://github.com/openai/openai-python uses it, and I feel like `httpx` is the trend for the future.

langchian will pick up the dependency via langchain-community - as long as its listed there, its fine

Also, we have not discussed making httpx the default or not. Since we have no current plans to, its probably best to add it as an **optional** dependency in langchain-community (not a required one)

169:
@hwchase17 I changed it to only use `httpx` as an optional dependency of `langchain-community`, but it still seems to affect the tests in `langchain`. Is my usage method wrong?

Screenshot 2024-01-02 at 1 34 48 PM


And the added `httpx` is not an SDK, I feel that it cannot be imported inline, similar to the following usage:

```
try:
             import httpx
             # do some stuff
         except ImportError:
             raise RuntimeError(
                 "Could not import httpx package. "
                 "Please install it via 'pip install httpx'"
             )
```

baskaryan (approved):

PR #15267, opened 2023-12-28
community: Support openai_proxy compatible with openai>1.0.0

169:
### **Description:** 

`openai` v1 no longer supports the method of setting `openai.proxy`. If using `OPENAI_PRPXY`, the users need to initialize an `http_client`.

 ### **Issue:**

https://github.com/langchain-ai/langchain/issues/13939


169:
@hwchase17  This feature is crucial for Chinese users. It seems to have similar issues to what's found in https://github.com/langchain-ai/langchain/pull/15265. However, in that PR, `httpx` can be an optional dependency, while in this current PR, `httpx` appears to be a necessary dependency. This is because OpenAI's client implementation is based on httpx, thus `http_client` or `async_http_client` needs to utilize `httpx`, and requests won't suffice. Do you have any ideas for a solution?

rigazilla:
we would also like to have httpx for http2 in #21910

baskaryan:
closing, feature has been implemented

PR #15277, opened 2023-12-28
[core, langchain] modelio code improvements

hwchase17:


libs/core/langchain_core/output_parsers/__init__.py
```diff
@@ -3,7 +3,7 @@
     BaseLLMOutputParser,
     BaseOutputParser,
 )
-from langchain_core.output_parsers.json import SimpleJsonOutputParser
+from langchain_core.output_parsers.json import JsonOutputParser, SimpleJsonOutputParser
```
nfcampos, line 6: quick start guide for parsers needs to be updated for new name

nfcampos (approved):

PR #15279, opened 2023-12-28
docs: add use cases index

baskaryan:


hwchase17 (approved):
while we're at it, maybe swap RAG and put it first? can also do when improving those docs

PR #15308, opened 2023-12-29
Docs: add param comment for `tracing_v2_enabled`

chyroc:


libs/core/langchain_core/tracers/schemas.py
```diff
@@ -109,7 +109,7 @@ class Run(BaseRunV2):
     events: List[Dict[str, Any]] = Field(default_factory=list)
 
     @root_validator(pre=True)
-    def assign_name(cls, values: dict) -> dict:
+    def assign_name(self, values: dict) -> dict:
```
hwchase17, line 112:
```suggestion
    def assign_name(cls, values: dict) -> dict:
```

hwchase17 (approved):

PR #15311, opened 2023-12-29
Feat: support RunnableBranch constructed with `|`

chyroc:
```python
from langchain_core.runnables import RunnableBranch

branch = RunnableBranch(
    (lambda x: isinstance(x, str), lambda x: x.upper()),
    (lambda x: isinstance(x, int), lambda x: x + 1),
    (lambda x: isinstance(x, float), lambda x: x * 2),
    lambda x: "goodbye",
  )

branch.invoke("hello") # "HELLO"
branch.invoke(None) # "goodbye"

# A RunnableBranch constructed using the `|` operator
branch = branch | (lambda x: isinstance(x, bool), lambda x: f'{x}')
branch.invoke(true) # "true"
```

### DIFF
--- a/libs/core/langchain_core/runnables/branch.py
+++ b/libs/core/langchain_core/runnables/branch.py
@@ -63,6 +63,10 @@ class RunnableBranch(RunnableSerializable[Input, Output]):
 
             branch.invoke("hello") # "HELLO"
             branch.invoke(None) # "goodbye"
+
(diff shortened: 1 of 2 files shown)


hwchase17:
we already use this operator generically to pipe thing together

im all for easier syntax to make this more clear/easier, but dont think this is the best way

PR #15330, opened 2023-12-29
Community: Newlines before bullets in IPYNB files (Vectara)

ofermend:
- **Description:** updated all Vectara IPYNB files so that bullets look okay in docs (added newline)
  - **Twitter handle:** @ofermend


hwchase17 (approved):
thanks!

Issue #15331, opened 2023-12-29
Self query retriever with Vector Store type <class 'langchain_community.vectorstores.faiss.FAISS'> not supported.


XariZaru:
### System Info

I've been trying to create a self query retriever so that I can look at metadata field info. This issue comes up. Should I be using another vector store to make this work? I can only really work with FAISS. I cannot use ChromaDB since my Python environment is limited to a previous version.

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [X] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [X] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

1. Create a FAISS Vectorstore DB
2. Create a metadata_field_info object and pass it to a SelfQuery object
3. Create LLM with this retriever

```python
embedding_function = OpenAIEmbeddings(openai_api_key=OPENAI_API_KEY, model="text-embedding-ada-002")
    
    db = FAISS.load_local(input_dir + "/" + "storage/deploy/faiss-db", embedding_function)#, distance_strategy="COSINE", normalize_L2 = True)
    # retriever = KNNRetriever(vectorstore=db)
    
    llm = ChatOpenAI(
        temperature=0,
        verbose=False,
        openai_api_key=key,
        model_name="gpt-3.5-turbo"
    )       

    metadata_field_info = [
        AttributeInfo(
            name="source",
            description="The document this chunk is from.",
            type="string",
        ),
        AttributeInfo(
            name="origin",
            description="The origin the document came from. Comes from either scraped websites like TheKinection.org, Kinecta.org or database files like Bancworks. Bancworks is the higher priority.",
            type="string",
        ),
        AttributeInfo(
            name="date_day",
            description="The day the document was uploaded.",
            type="integer",
        ),
        AttributeInfo(
            name="date_month",
            description="The month the document was uploaded.",
            type="integer",
        ),
        AttributeInfo(
            name="date_year",
            description="The year the document was uploaded.",
            type="integer",
        ),
    ]

    # retriever = db.as_retriever(search_type="similarity", search_kwargs={'k': 6}, metadata_field_info=metadata_field_info)

[...]

PR #15337, opened 2023-12-30
Langchain: FAISS Translator for Self Query Retriver

keenborder786:
**Description:** Issue #15331 pointed out that there was no support for FAISS for `SelfQueryRetriever`. Therefore, I decided to add the desired support. I understand that FAISS does not have an internal query language like other vector stores (e.g., Chroma), but I noticed that the current implementation of the `FAISS` VectorStore Class in Langchain allows the user to filter documents based on given metadata attributes. Therefore, the translator simply translates the structured query according to the logic of filtering defined in `similarity_search_with_score_by_vector` and `max_marginal_relevance_search_with_score_by_vector`, which filters the documents for the specific values present in metadata keys.

**Issue:** #15331

PS: I have not added the docs for this one. Only once you see the value in adding this translator, I will add the docs.


keenborder786:
@hwchase17

keenborder786:
@baskaryan

hwchase17 (commented):
i def see the value

will defer to @baskaryan on correct way to implement

CanIbeyourdog:
Thanks for your code! 
  When I use FAISS for Self-Query Retriver of official demo(https://python.langchain.com/docs/modules/data_connection/retrievers/self_query),I got a blank list.But Chroma works.

PR #15338, opened 2023-12-30
<langchain>: add checks for qaSourcesChain to check if the returned answer referen…

murdadesmaeeli:
…ces the source from which it returning an answer


hwchase17 (commented):
we probably don't want to add this to the built in chain for a few reasons

1. this adds another call the llm, not sure everyone wants this
2. we are moving away from built in chains, in favor of LCEL and other documentation

I would suggest opening a PR to add a section for "citing sources" here: https://python.langchain.com/docs/use_cases/question_answering/

hwchase17:
closing - see PR comments for suggestion of how to reopen

mehrdad-es:
@hwchase17 thanks, I think going away from built in chains is the right approach as there would be too many to cater to. I'll look at PR options.

PR #15351, opened 2023-12-31
Calculate trace_id and dotted_order client side

agola11:


libs/core/langchain_core/tracers/schemas.py
```diff
@@ -16,6 +17,8 @@ class Run(BaseRunV2):
     child_runs: List[Run] = Field(default_factory=list)
     tags: Optional[List[str]] = Field(default_factory=list)
     events: List[Dict[str, Any]] = Field(default_factory=list)
+    trace_id: Optional[UUID] = None
```
nfcampos, line 20: Why optional

agola11: because these fields are set after the object is created with the current logic, in `_start_trace`. I can change that but it would require shifting around more code

PR #15352, opened 2023-12-31
Langchain: Fix quickstart doc code not working

AhmedHathout:
The quickstart doc is missing a few but very simple things that without them, the code does not work. This PR fixes that by 
- Adding commands to install `tiktoken` and `langchainhub`
- Adds a comma between 2 parameters for one of the methods


hwchase17 (approved):
thanks!

Issue #15360, opened 2023-12-31
SQLDatabase returns result without column names.


dudu-upstream:
### System Info

Python 3.9
LangChain 0.0.339

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [X] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

I created new instance of SQLDatabase and executed SQL queries using the 'run' method.
The problem is that when I execute a simple SELECT query the result returns without the column names, only the values, and therefore I can't tell which value relates to which column.

I saw in the implementation the following code in the 'run' method:
```
result = self._execute(command, fetch)
        # Convert columns values to string to avoid issues with sqlalchemy
        # truncating text
        res = [
            tuple(truncate_word(c, length=self._max_string_length) for c in r.values())
            for r in result
        ]
```

and I'm wondering if this data manipulation is done by intention or maybe it's a bug.

### Expected behavior

For example:
executes the following SQL: select id, count(*) as num_count from some_table
expected result: [{'id': 383, 'num_count': 10}]
actual result: [(383, 10)]

PR #15361, opened 2023-12-31
SQLDatabase drop the column names in the result.

dudub12:
Fix for the following bug:
https://github.com/langchain-ai/langchain/issues/15360


hwchase17 (commented):
if returning lots of rows, this may take up a lot of tokens, which is perhaps why this was omitted? maybe add a flag to enable this?

dudu-upstream:
> if returning lots of rows, this may take up a lot of tokens, which is perhaps why this was omitted? maybe add a flag to enable this?

But it's not something that involve LLM, it's just a simple SQL response.. sure I can add a flag for that for backward compatibility.

hwchase17:
> > if returning lots of rows, this may take up a lot of tokens, which is perhaps why this was omitted? maybe add a flag to enable this?
> 
> But it's not something that involve LLM, it's just a simple SQL response.. sure I can add a flag for that for backward compatibility.

yeah but practically the outputs are usually piped to an llm

dudu-upstream:
> > > if returning lots of rows, this may take up a lot of tokens, which is perhaps why this was omitted? maybe add a flag to enable this?
> > 
> > 
> > But it's not something that involve LLM, it's just a simple SQL response.. sure I can add a flag for that for backward compatibility.
> 
> yeah but practically the outputs are usually piped to an llm

You are right but in that way the LLM don't know which value refers to which column some it may return incorrect results.
Anyway I will add the flag with a default value to give the ability to control that behavior.

hwchase17 (approved):

PR #15363, opened 2023-12-31
Langchain: Update Vectara templates

ofermend:
fixed multi-query template for Vectara
added self-query template for Vectara

Also added prompt_name parameter to summarization

CC @efriis 
 **Twitter handle:** @ofermend


templates/rag-vectara-selfquery/rag_vectara_selfquery/chain.py
```diff
@@ -0,0 +1,71 @@
...
+    vectara = Vectara()
+
+    # Set up a retriever to query your vectara service with self-querying capabilities
+    retriever = SelfQueryRetriever.from_llm(
+        llm, vectara, document_contents, metadata_field_info, verbose=True
+    )
+
+    chain = (
+        RunnableParallel({"context": retriever, "question": RunnablePassthrough()})
+        | (lambda res: res[-1])
```
efriis, line 65: Bug

efriis (commented):
Hey @ofermend ! The multiquery changes look great.

I don't think the self query one has been tested. I think the chain is missing a prompt and LLM call potentially, and let me know if that's incorrect! I run into a KeyError on `-1` when running it because the object passed into the lambda is a dict.

templates/rag-vectara-selfquery/pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+authors = [
+    "Ofer Mendelevitch <ofer@vectara.com>",
+]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = ">=3.8.1,<4.0"
+langchain = ">=0.0.325"
+openai = "<2"
+tiktoken = "^0.5.1"
```
efriis, line 14: also missing `lark`

templates/rag-vectara-selfquery/rag_vectara_selfquery/chain.py
ofermend: yes, you're right. 
I've spent some time looking at this more closely and I think it's more complex than I had originally thought. Not this bug specifically, I mean the whole self-query RAG. I wanted to make it work such that it uses Vectara summarizer, but it seems quite complex in the current setting, as SelfQueryRetriever seems to be built around retreiving documents and then using a separate LLM to do the final summarization.
1) I could probably still make this template follow this path - use Vectara as purely a document retriever (without summarization) - what other example is best to follow in this case?
2) Potentially later we can refactor this to make SelfQuery also support accepting a summarization component. 
WDYT?

ofermend: @efriis due to not having too much bandwidth next few weeks to make this work, let me just clean up (remove) the SQ stuff and will just leave the good fixes for MQ. We can tackle the SQ at some later time. SG?

efriis: Sounds great!

ofermend:
Updated w/o SelfQuery

ofermend:
@efriis - update as discussed. Pls let me know if anything else missing.

baskaryan (approved):

PR #15368, opened 2023-12-31
refactor `tools`

leo-gan:
The `langchain` [still holds several artifacts](https://api.python.langchain.com/en/latest/langchain_api_reference.html#module-langchain.tools) that belongs to `core`. If they moved then `langchain.tools` would be split between `core` and `community` (no `langchain` artifacts), which is nice. It effectively removes the `langchain.tools` namespace.
- moved a class and functions to `core`

### DIFF
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -1,4 +1,3 @@
-"""Base implementation for tools or skills."""
 from __future__ import annotations
 
 import inspect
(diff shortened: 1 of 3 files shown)


hwchase17 (commented):
i think im fine with this. @baskaryan ?

leo-gan:
@hwchase17 Any new comments?

efriis:
Think these make more sense where they are already

Issue #15269, opened 2023-12-28
[mistralai]: Don´t support stop sequence


DavidLMS:
### System Info

Python: 3.11
Langchain: 0.0.352
mistralai: 0.0.8

### Who can help?

@efriis 

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [X] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

If the ChatMistralAI model is used for an agent or similar, an error appears because the official Mistral API does not currently support the stop parameter (as other APIs such as OpenAI do).

### Expected behavior

Although this is something that should be fixed by Mistral in its official API, one of the following options should be done:
- Warn the user that this model cannot be used with a stop sequence before breaking execution due to the error.
- Implement an own solution for the stop sequence in the same package and do not send that parameter to the official client call.

niels-garve:
Hi @DavidLMS and @efriis,

I fixed that in my pull request: https://github.com/langchain-ai/langchain/pull/15375. What do you think of my suggested solution?

Thanks and cheers, Niels

PR #15375, opened 2024-01-01
mistralai[patch]: warn on stop token, fix on_llm_new_token

niels-garve:
## Description
First, thanks for adding Mistral AI support to LangChain. I started to use it in my application and found that async LLM calls — both for streaming and generation — led to `Unclosed connector` warnings when executed twice or more. In `acompletion_with_retry` the Mistral AI client gets instantiated over and over again without being closed.

To achieve this for both generation and streaming, I've introduced a client manager. The `ClientManager` class inherits from `AsyncCallbackHabdler` and is therefore ready to detect the end of a stream to close the client. And it's a container for client handling in case of not streamed generation.

Besides this feature, I fixed two little bugs:
1. The `stop` parameter is not yet implemented in the API, or am I missing something? I thought it makes sense to keep `_create_message_dicts` and log a warning because support may be added in the future?
2. I completed the call to `on_llm_new_token`, since the library I am currently using relies on it.

What do you think? Looking forward to your feedback.

## Issue
* https://github.com/langchain-ai/langchain/issues/15269

## Dependencies
No dependencies required for this change

## Integration Tests
Executing integration tests on the `master` branch shows multiple `Unclosed connector` warnings. **These are gone when integration tests are executed on this feature branch**. So, IMHO, there is no need to add extra tests here.

✅ `MISTRAL_API_KEY=XYZ make integration_tests` passes

## Misc
**Twitter handle:** @aisneaks

✅ `make format` passes
✅ `make lint` passes
✅ `make test` passes


niels-garve:
@baskaryan

efriis:
Great catch! However, I think the correct solution will be to maintain the `client` on the llm class. I'll put up a PR of that - I think it will be slightly simpler!

efriis:
#15786 for the async issue

Will look at fixing the other two bugs here and re-release langchain-mistralai. Appreciate your contribution @niels-garve !

efriis (approved):

efriis:
Oops stacked prs wrong. Dw you'll show as a coauthor on #15787 !

niels-garve:
Thanks @efriis ! If there's anything, please let me know. No worries about the co-author, it was a team effort

PR #15398, opened 2024-01-02
langchain: convert json to minified json to save prompt token cost

mengjincn:
- **Description:** convert json to minified json to save prompt token cost,
  - **Issue:** no issue,
  - **Dependencies:** not change any dependencies,


hwchase17 (commented):
not sure we want to make this change generically. would likely affect lots of peoples prompts. happy to make this a parameter one can toggle

PR #15427, opened 2024-01-02
core:adds tests for partial_variables

dkajtoch:
**Description:** Added small tests to test partial_variables in PromptTemplate. It was missing.


hwchase17 (approved):
thanks!

PR #15432, opened 2024-01-02
refactor `utils`

leo-gan:
The `langchain` [still holds several artifacts](https://api.python.langchain.com/en/latest/langchain_api_reference.html#module-langchain.utils) that belongs to `community`. If they moved then `langchain.utils` namespace would be  removed completely.
- moved `ernie_functions` artifacts to `community`


hwchase17 (commented):
thansk!

hwchase17 (approved):

PR #15437, opened 2024-01-03
refactored `memory`

leo-gan:
Created `langchain_community.memory`. Moved there two classes from `langchain.memory`

This movement triggered another changes:

1.
There is a lint rule that `langchain_community` code should not depend on `langchain`. Below, I show these dependencies:

```
`ZepMemory`
  --> langchain.memory.buffer.`ConversationBufferMemory`
       --> langchain.memory.chat_memory.`BaseChatMemory`
            --> langchain_community.chat_message_histories.in_memory.`ChatMessageHistory`
       --> langchain.memory.utils.`get_prompt_input_key`  # the same as in MotorheadMemory
      
`MotorheadMemory`
   --> langchain.memory.chat_memory.BaseChatMemory   # then the same dependency as in ZepMemory
        --> langchain_community.chat_message_histories.in_memory.ChatMessageHistory
   --> langchain.memory.utils.get_prompt_input_key

So, these classes and a function were moved into `core` because of the above dependencies:
- `ConversationBufferMemory`
- `BaseChatMemory`
- `ChatMessageHistory`
- `get_prompt_input_key`
```


leo-gan:
FYI @baskaryan

efriis (commented):
Let's leave the memory classes where they are, and if that bug exists without the moves, let's just fix that here!

leo-gan:
@efriis Fixed the import bug in the separate #17078. It is not mixed with this PR anymore.

libs/core/langchain_core/memory.py
```diff
@@ -81,3 +89,158 @@ def clear(self) -> None:
...
+        self.add_messages(messages)
+
+    def clear(self) -> None:
+        self.messages = []
+
+    async def aclear(self) -> None:
+        self.clear()
+
+
+class BaseChatMemory(BaseMemory, ABC):
```
eyurtsev, line 120: @leo-gan we don't have a clear road map for the old memory abstractions yet, so I'm not sure what packages make sense for the abstractions

The `BaseMemory`  abstraction is problematic -- overly complex and incorrect interface. We'll likely end up deprecating it longer-term.

It could potentially go in langchain_community, but not sure what we'd want to do with `ConversationBufferMemory`.

Overall dependency goals are:

langchain-core: base abstractions and core implementations
langchain: implementations of common utilities (e.g., token counting, generic agent implements etc)
community / partner packages: specific implementations of the abstractions

So:
* langchain-core without dependencies
* langchain depends on langchain-core, it does not depend on langchain-community
* langchain-community depends on langchain-core and langchain
* partner packaegs depending on langchain-core (and maybe langchain)

---

cc @efriis  / @baskaryan

Issue #15006, opened 2023-12-21
ConversationalRetrievalChain not working with stream callback handler for ChatGoogleGenerativeAI


hasansustcse13:
### System Info

aiohttp==3.9.1
aiosignal==1.3.1
annotated-types==0.6.0
anyio==4.2.0
attrs==23.1.0
cachetools==5.3.2
certifi==2023.11.17
charset-normalizer==3.3.2
colorama==0.4.6
dataclasses-json==0.6.3
distro==1.8.0
frozenlist==1.4.1
google-ai-generativelanguage==0.4.0
google-api-core==2.15.0
google-auth==2.25.2
google-generativeai==0.3.2
googleapis-common-protos==1.62.0
greenlet==3.0.2
grpcio==1.60.0
grpcio-status==1.60.0
h11==0.14.0
httpcore==1.0.2
httpx==0.26.0
idna==3.6
jsonpatch==1.33
jsonpointer==2.4
langchain==0.0.352
langchain-community==0.0.5
langchain-core==0.1.2
langchain-google-genai==0.0.5
langsmith==0.0.72
marshmallow==3.20.1
multidict==6.0.4
mypy-extensions==1.0.0
numpy==1.26.2
openai==1.6.0
packaging==23.2
Pillow==10.1.0
proto-plus==1.23.0
protobuf==4.25.1
pyasn1==0.5.1
pyasn1-modules==0.3.0
pydantic==2.5.2
pydantic_core==2.14.5
PyYAML==6.0.1
requests==2.31.0
rsa==4.9
sniffio==1.3.0
SQLAlchemy==2.0.23
tenacity==8.2.3
tqdm==4.66.1
typing-inspect==0.9.0
typing_extensions==4.9.0
urllib3==2.1.0
yarl==1.9.4

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

```
class StreamingLLMCallbackHandler(BaseCallbackHandler):
    """Callback handler for streaming LLM responses to a queue."""

[...]

PR #15459, opened 2024-01-03
Fixed streaming support using callback handler for Google genai

hasansustcse13:
- **Description:** Fixed streaming support using callback handler for Google genai. Callback handler not working with the `ChatGoogleGenerativeAI` which has been fixed in this PR. 
  - **Issue:** langchain-ai/langchain-google#120 langchain-ai/langchain#15006

### DIFF
--- a/libs/partners/google-genai/langchain_google_genai/chat_models.py
+++ b/libs/partners/google-genai/langchain_google_genai/chat_models.py
@@ -30,8 +30,8 @@ from langchain_core.callbacks.manager import (
 )
 from langchain_core.language_models.chat_models import (
     BaseChatModel,
-    generate_from_stream,
-    agenerate_from_stream
+    agenerate_from_stream,
+    generate_from_stream
 )
 from langchain_core.messages import (
     AIMessage,
--- a/libs/partners/google-genai/langchain_google_genai/chat_models.py
+++ b/libs/partners/google-genai/langchain_google_genai/chat_models.py
@@ -28,7 +28,7 @@ from langchain_core.callbacks.manager import (
     AsyncCallbackManagerForLLMRun,
     CallbackManagerForLLMRun,
 )
-from langchain_core.language_models.chat_models import BaseChatModel
+from langchain_core.language_models.chat_models import BaseChatModel, generate_from_stream, agenerate_from_stream
 from langchain_core.messages import (
     AIMessage,
     AIMessageChunk,
@@ -466,6 +466,8 @@ Supported examples:
     
     Gemini does not support system messages; any unsupported messages will 
     raise an error."""
+    streaming: bool = False
+    """Whether to stream the results or not."""
 
     class Config:
         allow_population_by_field_name = True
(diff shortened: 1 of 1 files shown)


hasansustcse13:
@efriis Could you please look into this PR? Our current project has a large dependency on it.

efriis:
Hey @hasansustcse13 ! This isn't the interface to follow anymore. Instead, you should stream using `.stream()` instead!

Let me know if you have any other questions.

Erick

hasansustcse13:
@efriis Streaming doesn't work if I don't use **LangChain Expression Language (LCEL)**. In our project, we have used streaming using a Callback handler(on_llm_new_token) and different kinds of built-in chains.  As Gemini is not implanted the same as OpenAI, it does not work with a callback handler (on_llm_new_token). To need to support streaming in Gemini along with other LLM we have to convert all the code to LCEL. I believe this will be faced by other users also. That's why I have made the Change in the PR. Is there any other way to support streaming without LCEL and with a minimal changes? 

```
from langchain.chains import LLMChain

prompt_template = "Tell me details about the Company {name} with 4 bullet point?"


def gemini_stream():
    llm = ChatGoogleGenerativeAI(model="gemini-pro", google_api_key='AIzaSyliRCeTRontPQ')
    llm_chain = LLMChain(llm=llm, prompt=PromptTemplate.from_template(prompt_template), verbose=True)
    for content in llm_chain.stream({'name': "Google"}):
        print(content)


def gemini_stream_with_lang_expression():
    llm = ChatGoogleGenerativeAI(model="gemini-pro", google_api_key='AIzaSyCuQ')
    prompt = PromptTemplate.from_template(prompt_template)
    for content in chain.stream({'name': "Google"}):
        print(content)


if __name__ == '__main__':
    # This is not working though I have use .stream() method. It gives the output after finising the completetion completely.
    gemini_stream()

    # This is working
    # gemini_stream_with_lang_expression()

```

andrei-rusu:
I've also bumped into this problem, and I don't understand why this PR was closed. I was trying to set up a streaming agent as in [this example](https://github.com/langchain-ai/streamlit-agent/blob/main/streamlit_agent/basic_streaming.py) with Gemini, but ``on_llm_new_token`` indeed does not fire... I thought using a callback handler was the way to achieve per-token streaming for displaying (instead of per-chunk).

If there is another way to achieve this now, I would like to know (And probably the streamlit agent examples need to be rewritten accordingly).

pseudotensor:
@efriis I also agree with @hasansustcse13 that you are probably confused.  One does not just invoke llm.stream() directly, but it pass an llm with callback into something like load_qa_chain, then streaming is broken for gemini but not other models.

atuncer:
Is there a version that I can import via pip

rushikeshvizerto7:
how to use map reduce chain type with gemini models using chatgooglegenerativeai and also get streaming responses at the same time using custom callback handler where I want token level streaming by appending in a queue

RushikeshKothawade07:
how to use map reduce chain type with gemini models using chatgooglegenerativeai and also get streaming responses at the same time using custom callback handler where I want token level streaming by appending in a queue

efriis:
try `ChatGoogleGenerativeAI(...).bind(stream=True)`

PR #15460, opened 2024-01-03
Feat: add url support for UnstructuredHTMLLoader

chyroc:
usage:

```python
from langchain.document_loaders import UnstructuredHTMLLoader

source_url = "https://www.sec.gov/ix?doc=/Archives/edgar/data/40987/000004098720000010/gpc-12312019x10k.htm"
loader = UnstructuredHTMLLoader(source_url)
documents = loader.load()
print(documents)

```

### DIFF
--- a/libs/community/langchain_community/document_loaders/html.py
+++ b/libs/community/langchain_community/document_loaders/html.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Any, List, Union
 
 from langchain_community.document_loaders.unstructured import UnstructuredFileLoader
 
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/document_loaders/html.py
```diff
@@ -27,6 +27,35 @@ class UnstructuredHTMLLoader(UnstructuredFileLoader):
...
+        self,
+        file_path: Union[str, List[str]] = None,
+        **kwargs: Any,
+    ):
+        def _url_to_file(url: str) -> str:
+            import tempfile
+
+            import requests
+
+            r = requests.get(url)
```
baskaryan, line 40: we should try to avoid making requests in initialization, could we add this functionality to `load`?

PR #15461, opened 2024-01-03
Community: Support different Pinecone initializations depending on the version

DosticJelena:
**Description**: Adding support for different initialization implementations depending on the Pinecone client version.
**Issue**: N/A
**Dependencies**: N/A
**Twitter handle**: N/A

Tested by installing different versions of the Pinecone client and trying to create an index (both regular and serverless for the new version) from documents, and using the `Pinecone` vector store to query the index. The process ran smoothly and as expected in both cases.

### DIFF
--- a/libs/community/langchain_community/vectorstores/pinecone.py
+++ b/libs/community/langchain_community/vectorstores/pinecone.py
@@ -1,15 +1,27 @@
 from __future__ import annotations
 
 import logging
+import os
 import uuid
 import warnings
-from typing import TYPE_CHECKING, Any, Callable, Iterable, List, Optional, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Iterable,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 
 import numpy as np
 from langchain_core.documents import Document
 from langchain_core.embeddings import Embeddings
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/pinecone.py
```diff
@@ -359,11 +370,24 @@ def get_pinecone_index(
                 "Please install it with `pip install pinecone-client`."
             )
 
-        indexes = pinecone.list_indexes()  # checks if provided index exists
+        pinecone_client_version = get_distribution("pinecone-client").version
```
efriis, line 373: would it be better to rely on `pinecone.__version__` instead? I think it isn't in the v3 release candidates, but might be good to add for consistency?

I believe `pkg_resources` is deprecated.

jamescalam: @efriis we're getting `__version__` added back into `pinecone-client` — will get back to you once implemented

efriis: Amazing. Thanks! Think that'll be better :)

efriis:
Hey team - I'll implement this in #15717 due to the urgency. I don't have edit rights to this branch 

https://python.langchain.com/docs/contributing/faq#how-do-i-allow-maintainers-to-edit-my-pr

PR #15464, opened 2024-01-03
Refactor: use SecretStr for zapier tool

chyroc:


libs/community/langchain_community/utilities/zapier.py
```diff
@@ -176,8 +181,8 @@ def list(self) -> List[Dict]:
         https://nla.zapier.com/docs/using-the-api#ai-guessing)
         """
         session = self._get_session()
+        response = session.get(self.zapier_nla_api_base + "exposed/")
```
baskaryan, line 184: why move this outside the try?

ccurme:
Closing as author is unresponsive but will re-open if there's desire to keep iterating.

PR #15468, opened 2024-01-03
Patch: improve openai.ChatCompletion getter

chyroc:


ccurme:
This needs to be updated as `openai.ChatCompletion` is no longer supported in openai v1. Will re-open if needed.

PR #15489, opened 2024-01-03
community: changing the .dict() to .model_dump() since it is deprecated

henriquevedoveli:
- **Description:** The method `.dict()` was replaced by `.model_dump` since its deprecated in Pydantic V2 and will be removed in Pydantic V.3, 
  - **Issue:** None,
  - **Dependencies:** None,
  - **Twitter handle:** @hhvedoveli


hwchase17:
does this still work for pydantic v1?
we've also split out openai into its own package - langchain_openai. the code in community is deprecated and left around because some classes depend on it. can we update this pr to make this change in the langchain_openai package

hwchase17:
closing, should be reopened against langchain-openai package

PR #15499, opened 2024-01-03
core[minor], langchain[minor]: deprecate Chain and BaseLanguageModel …

baskaryan:
…methods

### DIFF
--- a/libs/core/langchain_core/language_models/base.py
+++ b/libs/core/langchain_core/language_models/base.py
@@ -61,17 +61,6 @@ class BaseLanguageModel(
     """Abstract base class for interfacing with language models.
 
     All language model wrappers inherit from BaseLanguageModel.
-
-    Exposes three main methods:
-    - generate_prompt: generate language model outputs for a sequence of prompt
-        values. A prompt value is a model input that can be converted to any language
-        model input format (string or messages).
-    - predict: pass in a single string to a language model and return a string
-        prediction.
-    - predict_messages: pass in a sequence of BaseMessages (corresponding to a single
-        model call) to a language model and return a BaseMessage prediction.
-
-    Each of these has an equivalent asynchronous method.
     """
 
     @property
--- a/libs/core/langchain_core/language_models/base.py
+++ b/libs/core/langchain_core/language_models/base.py
@@ -166,7 +166,7 @@ class BaseLanguageModel(
     def predict(
         self, text: str, *, stop: Optional[Sequence[str]] = None, **kwargs: Any
     ) -> str:
-        """[DEPRECATED] Pass a single string input to the model and return a string.
+        """Pass a single string input to the model and return a string.
 
          Use this method when passing in raw text. If you want to pass in specific
             types of chat messages, use predict_messages.
@@ -191,7 +191,7 @@ class BaseLanguageModel(
         stop: Optional[Sequence[str]] = None,
         **kwargs: Any,
     ) -> BaseMessage:
-        """[DEPRECATED] Pass a message sequence to the model and return a message.
+        """Pass a message sequence to the model and return a message.
 
         Use this method when passing in chat messages. If you want to pass in raw text,
             use predict.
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -663,6 +711,7 @@ async def _call_async(
         else:
             raise ValueError("Unexpected generation type")
 
+    @deprecated("0.1.0", alternative="ainvoke")
```
nfcampos, line 714: should be invoke

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -632,6 +679,7 @@ def _astream(
     ) -> AsyncIterator[ChatGenerationChunk]:
         raise NotImplementedError()
 
+    @deprecated("0.1.0", alternative="invoke")
     def __call__(
```
nfcampos, line 683: shouldn't we be deprecating acall too?

libs/langchain/langchain/chains/base.py
```diff
@@ -358,38 +399,14 @@ async def acall(
...
-            raise e
-        await run_manager.on_chain_end(outputs)
-        final_outputs: Dict[str, Any] = self.prep_outputs(
-            inputs, outputs, return_only_outputs
-        )
-        if include_run_info:
-            final_outputs[RUN_KEY] = RunInfo(run_id=run_manager.run_id)
-        return final_outputs
 
     def prep_outputs(
```
nfcampos, line 411: Make private?

libs/langchain/langchain/chains/base.py
```diff
@@ -528,6 +546,7 @@ def run(
                 f" but not both. Got args: {args} and kwargs: {kwargs}."
             )
 
+    @deprecated("0.1.0", alternative="invoke")
```
nfcampos, line 549: ainvoke

libs/core/langchain_core/language_models/base.py
```diff
@@ -160,11 +161,12 @@ async def agenerate_prompt(
                 prompt and additional model provider-specific output.
         """
 
+    @deprecated("0.1.0", alternative="invoke")
     @abstractmethod
     def predict(
         self, text: str, *, stop: Optional[Sequence[str]] = None, **kwargs: Any
     ) -> str:
-        """Pass a single string input to the model and return a string prediction.
+        """[DEPRECATED] Pass a single string input to the model and return a string.
```
eyurtsev, line 169: No need to do that -- it's already done by the decorator

https://github.com/langchain-ai/langchain/blob/26f84b74d0f7dc4d2211a1a62d47eec36cb1d726/libs/core/langchain_core/_api/deprecation.py#L216-L216

libs/core/langchain_core/language_models/chat_models.py
baskaryan: dont think BaseChatModel implements acall

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -632,6 +679,7 @@ def _astream(
     ) -> AsyncIterator[ChatGenerationChunk]:
         raise NotImplementedError()
 
+    @deprecated("0.1.0", alternative="invoke")
```
eyurtsev, line 682: This means deprecated since 0.1.0

baskaryan: yea was imagining we wouldn't land this till 0.1.0

libs/core/langchain_core/language_models/base.py
baskaryan: 😮 ‼️

libs/langchain/langchain/chains/base.py
```diff
@@ -289,39 +353,21 @@ def __call__(
...
-            self.verbose,
-            tags,
-            self.tags,
-            metadata,
-            self.metadata,
-        )
-        new_arg_supported = inspect.signature(self._call).parameters.get("run_manager")
-        run_manager = callback_manager.on_chain_start(
-            dumpd(self),
+        config = {
```
hwchase17, line 356: why are we changing this? are we sure this doesnt have any side effects? eg with callbacks

baskaryan: so that we can deprecate it. couldn't deprecate before bc invoke used it, now we flipped dependency

shouldn't change anything unless i made a mistake, can test some more

PR #15514, opened 2024-01-04
langchain: Introducing Google Semantic Retriever and Attributed Question and Answering (AQA)

pikalaw:
**Description:**

Introducing two new features from Google's Generative Language API.

* Bundled embeddings calculation and vector storage for enabling semantic retrieval.
* State-of-the-art fine-tuned model for answering questions grounded in passages.

**Dependencies:**

The Python package [google-ai-generativelanguage](https://ai.google.dev/api/python/google/ai/generativelanguage)

**Twitter handle:**

https://twitter.com/googledevs


efriis (commented):
Hey @pikalaw ! I haven't looked deeply at the code yet, and I'll do so when it's in the right directories! Could you move the implementation around to match one of the two [integration types](https://python.langchain.com/docs/contributing/integrations)? Happy to help answer any questions as well as you build the integration!

libs/langchain/langchain/llms/google/generativeai/aqa_model.py
```diff
@@ -0,0 +1,42 @@
+from typing import Any, List, Optional
+
+import google.ai.generativelanguage as genai
```
efriis, line 3: I would recommend having a different alias than `genai` for the generative language package (maybe `genlang` or something). Just so people don't get confused about it being different from the google.generativeai package.

efriis:
Discussed in email - should move to https://github.com/langchain-ai/langchain-google

PR #15515, opened 2024-01-04
added interfaces as `core.schemas`

leo-gan:
Defined interfaces for base classes.
The idea is NOT to move Base classes at all. The idea is to extract pure interfaces and use them to help with integrations. The mandatory fields and functions should be clearly defined. The optional convenience functions should not be part of the Interface.
Some of the current Base classes are overloaded with helper functions, and it is not clear what is mandatory and what is optional without going deep into the implementations of Base classes. As time goes on, more and more convenience functions are added, and integration becomes more unnecessarily complex.
CC @baskaryan

### DIFF
--- a/libs/core/langchain_core/schema/chat_message_history.py
+++ b/libs/core/langchain_core/schema/chat_message_history.py
@@ -1,0 +1,70 @@
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, List, Union
+
+if TYPE_CHECKING:
+    from langchain_core.messages import (
+        AIMessage,
+        BaseMessage,
+        HumanMessage,
+    )
+
+
+class ChatMessageHistoryInterface(ABC):
+    """Interface for storing chat message history.
+
+    See `ChatMessageHistory` for default implementation.
+    The default implementation is the BaseChatMessageHistory class.
+
+    Example:
+        .. code-block:: python
+
+            class FileChatMessageHistory(BaseChatMessageHistory):
+                storage_path:  str
+                session_id: str
+
+               @property
+               def messages(self):
+                   with open(os.path.join(storage_path, session_id), 'r:utf-8') as f:
+                       messages = json.loads(f.read())
+                    return messages_from_dict(messages)
+
+               def add_message(self, message: BaseMessage) -> None:
+                   messages = self.messages.append(_message_to_dict(message))
+                   with open(os.path.join(storage_path, session_id), 'w') as f:
+                       json.dump(f, messages)
+
+               def clear(self):
+                   with open(os.path.join(storage_path, session_id), 'w') as f:
+                       f.write("[]")
+    """
+
+    messages: List[BaseMessage]
+    """A list of Messages stored in-memory."""
+
+    @abstractmethod
+    def add_user_message(self, message: Union[HumanMessage, str]) -> None:
+        """Convenience method for adding a human message string to the store.
+
+        Args:
+            message: The human message to add
+        """
+
+    @abstractmethod
+    def add_ai_message(self, message: Union[AIMessage, str]) -> None:
+        """Convenience method for adding an AI message string to the store.
+
+        Args:
+            message: The AI message to add.
+        """
+
+    @abstractmethod
+    def add_message(self, message: BaseMessage) -> None:
+        """Add a Message object to the store.
+
+        Args:
+            message: A BaseMessage object to store.
+        """
+
+    @abstractmethod
+    def clear(self) -> None:
+        """Remove all messages from the store"""
--- a/libs/core/langchain_core/schema/chat_message_history.py
+++ b/libs/core/langchain_core/schema/chat_message_history.py
@@ -1,12 +1,11 @@
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, List, Union
+from typing import List, Union
 
-if TYPE_CHECKING:
-    from langchain_core.messages import (
-        AIMessage,
-        BaseMessage,
-        HumanMessage,
-    )
+from langchain_core.messages import (
+    AIMessage,
+    BaseMessage,
+    HumanMessage,
+)
 
 
 class ChatMessageHistoryInterface(ABC):
--- a/libs/core/langchain_core/schema/chat_model.py
+++ b/libs/core/langchain_core/schema/chat_model.py
@@ -1,0 +1,48 @@
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional
+
+from langchain_core.callbacks import (
+    AsyncCallbackManagerForLLMRun,
+    CallbackManagerForLLMRun,
+)
+from langchain_core.messages import BaseMessage
+from langchain_core.outputs import ChatResult
+
+
+# TODO: this is a work in progress
+# TODO: do we need other methods from BaseChatModel or BaseLanguageModel?
+class ChatModelInterface(ABC):
+    """Interface for Chat Model.
+
+    The default implementation of this interface is :class:`BaseChatModel`.
+    """
+
+    @abstractmethod
+    def _call(
+        self,
+        messages: List[BaseMessage],
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> str:
+        """Simpler interface."""
+
+    @abstractmethod
+    def _generate(
+        self,
+        messages: List[BaseMessage],
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> ChatResult:
+        """Simpler interface."""
+
+    @abstractmethod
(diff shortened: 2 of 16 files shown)


leo-gan:
@eyurtsev It is ready for critique.

eyurtsev:
Hi @leo-gan, I misunderstood what you meant initially by pulling out pure interface from existing interfaces. I don't think we should be creating pure interfaces since a number of the interfaces are loaded with convenience functions that should not exist on a pure interface. I will leave comments on some of the interfaces in the PR to show what I mean.

Generally, I'm in favor of moving all Base classes that we believe are correct into langchain correct, but we have to double check with @baskaryan in terms of sequencing and also scope.

libs/core/langchain_core/schema/chat_message_history.py
```diff
@@ -0,0 +1,69 @@
...
+               def clear(self):
+                   with open(os.path.join(storage_path, session_id), 'w') as f:
+                       f.write("[]")
+    """
+
+    messages: List[BaseMessage]
+    """A list of Messages stored in-memory."""
+
+    @abstractmethod
+    def add_user_message(self, message: Union[HumanMessage, str]) -> None:
```
eyurtsev, line 44: * add_user_message, add_ai_message should not exist

@abstactmethod should be to `add_messages` which supports bulk load.

`add_message` so should exist and be implemented by default to use `add_messages`

Here's a PR to the existing base class that attempts to fix some of the issues: https://github.com/langchain-ai/langchain/pull/15709/files

Unfortunately, for backwards compatibility the implementation of add_message needs to be more complex and actually account for the fact that most classes don't have add_messages

libs/core/langchain_core/schema/chat_model.py
```diff
@@ -0,0 +1,48 @@
...
+# TODO: this is a work in progress
+# TODO: do we need other methods from BaseChatModel or BaseLanguageModel?
+class ChatModelInterface(ABC):
+    """Interface for Chat Model.
+
+    The default implementation of this interface is :class:`BaseChatModel`.
+    """
+
+    @abstractmethod
+    def _call(
```
eyurtsev, line 21: All chat models should be runnables, so we'd like to enforce that by design

libs/core/langchain_core/schema/docstore.py
```diff
@@ -0,0 +1,29 @@
+"""Interface to access to place that stores documents."""
+from abc import ABC, abstractmethod
+from typing import Dict, List, Union
+
+from langchain_core.documents import Document
+
+
+class DocstoreInterface(ABC):
```
eyurtsev, line 8: This interface should be deprecated from langchain -- It's not a doc-store

libs/core/langchain_core/schema/docstore.py
```diff
@@ -0,0 +1,29 @@
...
+        If page exists, return the page summary, and a Document object.
+        If the page does not exist, return similar entries.
+        """
+
+    @abstractmethod
+    def delete(self, ids: List) -> None:
+        """Deleting IDs from in memory dictionary."""
+
+
+class AddableMixin(ABC):
```
eyurtsev, line 24: Likely shouldn't exist, but I don't have context for this

libs/core/langchain_core/schema/document_loader.py
```diff
@@ -0,0 +1,49 @@
...
+    def load(self) -> List[Document]:
+        """Load data into Document objects.
+
+        Sub-classes should implement this method
+        as `return list(self.lazy_load())`
+        This method returns a List which is materialized in memory.
+        """
+
+    @abstractmethod
+    def load_and_split(
```
eyurtsev, line 30: We need to figure out how to deprecate load_and_split

libs/core/langchain_core/schema/document_loader.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Implementations should implement the lazy-loading method using generators
+    to avoid loading all Documents into memory at once.
+
+    The `load` method will remain as is for backwards compatibility, but its
+    implementation should be just `list(self.lazy_load())`.
+    """
+
+    @abstractmethod
+    def load(self) -> List[Document]:
```
eyurtsev, line 21: Implementation should be provided by default and proxy lazy_load. Users should never be implementing `load` and ideally not using it either -- lots of enterprise use cases involve millions of docs, so these loaders will OOM

libs/core/langchain_core/schema/document_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+                    self, documents: Sequence[Document], **kwargs: Any
+                ) -> Sequence[Document]:
+                    raise NotImplementedError
+
+    """  # noqa: E501
+
+    @abstractmethod
+    def transform_documents(
+        self, documents: Sequence[Document], **kwargs: Any
+    ) -> Sequence[Document]:
```
eyurtsev, line 49: Return type should be concrete and ownership over list belongs to the user -- we need to fix this across the code base

```suggestion
    ) -> List[Document]:
```

libs/core/langchain_core/schema/embeddings.py
```diff
@@ -0,0 +1,22 @@
+from abc import ABC, abstractmethod
+from typing import List
+
+
+class EmbeddingsInterface(ABC):
```
eyurtsev, line 5: This matches the existing BaseClass:

https://github.com/langchain-ai/langchain/blob/b8f83d7f076098e10d001e67c782acc2e4c96a42/libs/core/langchain_core/embeddings.py#L20-L20

And I think we do want a default implementation of async methods that can be over-ridden by the implementer. Many developers aren't comfortable with async and will not know how to implement this, so at least their implementation will get pseudo support for async with the existing implementation in Embeddings

leo-gan:
> Hi @leo-gan, I misunderstood what you meant initially by pulling out pure interface from existing interfaces. I don't think we should be creating pure interfaces since a number of the interfaces are loaded with convenience functions that should not exist on a pure interface. I will leave comments on some of the interfaces in the PR to show what I mean.
> 
> Generally, I'm in favor of moving all Base classes that we believe are correct into langchain correct, but we have to double check with @baskaryan in terms of sequencing and also scope.

My idea is not to move Base classes at all. The idea is to extract pure interfaces and use them to help with integrations. The mandatory fields and functions should be clearly defined. The optional convenience functions should not be part of the Interface. 
Some of the current Base classes are overloaded with helper functions, and it is not clear what is mandatory and what is optional without going deep into the implementations of Base classes. As time goes on, more and more convenience functions are added, and integration becomes more complex.

libs/core/langchain_core/schema/chat_model.py
leo-gan: Yes. But runnables have separate interfaces and Base classes derive the base interface + runnable interface.

libs/core/langchain_core/schema/docstore.py
leo-gan: I'm in favor of removing the obsolete classes.

libs/core/langchain_core/schema/document_loader.py
leo-gan: Yeah, It is definitely a helper function

libs/core/langchain_core/schema/document_loader.py
leo-gan: If it is mandatory we can just remove `lazy_load` and leave `load` but enforce the lazy behavior with the lazy signature. Unfortunately, it looks like a braking change.

libs/core/langchain_core/schema/document_transformer.py
leo-gan: > Return type should be concrete and ownership over list belongs to the user -- we need to fix this across the code base

Not sure, why return type should be concrete.
I've found these arguments:
In favor of interfaces: "Whether I own the calling code makes a difference too (internal vs public APIs):

If I own the code that calls the method in question (i.e. internal API), then the more willing I am to return the concrete type.
If I don't control the code that calls this method (e.g. public API), the more likely I am to return the interface type instead. Returning the concrete type is a commitment and, generally speaking, the less I promise, the easier.
Other considerations:

Testing may be easier with interfaces since I can use a mock object that implements the interface..."

In favor of concrete: 
"Rule of thumb, in return types, be a as specific as possible, in parameter types be as unspecific as possible." - which looks like dogma. 

Anyway, I don't get it, why?

leo-gan:
> Hi @leo-gan, I misunderstood what you meant initially by pulling out pure interface from existing interfaces. I don't think we should be creating pure interfaces since a number of the interfaces are loaded with convenience functions that should not exist on a pure interface. I will leave comments on some of the interfaces in the PR to show what I mean.
> 
> Generally, I'm in favor of moving all Base classes that we believe are correct into langchain correct, but we have to double check with @baskaryan in terms of sequencing and also scope.

If we have only a base abstract class, we can add one [or several] interfaces on top of it. IMHO, it makes sense since it explains the pure purpose. Current base classes are contaminated with helper functions. But we need a clear separation of the `required` and `optional` implementation functions.

PR #15533, opened 2024-01-04
removed deprecated openai model 

VpkPrasanna:
removed the deprecated model from text embedding page of openai notebook and added the suggested model from openai page


efriis (approved):
Thanks!

PR #15557, opened 2024-01-04
infra: try show last update time

efriis:


efriis:
vercel only clones depth 10 and quarto converted files are untracked by git - would need workaround

PR #15558, opened 2024-01-04
langchain[patch]: deprecate old agent classes and methods

baskaryan:


libs/langchain/langchain/agents/conversational/base.py
```diff
@@ -17,6 +18,7 @@
 from langchain.chains import LLMChain
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 21: this is actually quite used - the alternative is create_react_agent

libs/langchain/langchain/agents/conversational_chat/base.py
```diff
@@ -29,6 +30,7 @@
 from langchain.chains import LLMChain
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 33: this is actually quite used - the alternative is create_json_chat_agent

libs/langchain/langchain/agents/agent.py
```diff
@@ -53,6 +54,7 @@
 logger = logging.getLogger(__name__)
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 57: will this raise an error when using RunnableAgent since that inherits from this?

libs/langchain/langchain/agents/agent.py
```diff
@@ -486,6 +489,7 @@ async def aplan(
         return output
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 492: can we say like "use new agent constructor methods?"

libs/langchain/langchain/agents/agent.py
baskaryan: believe it doesn't but will double check

libs/langchain/langchain/agents/agent.py
```diff
@@ -194,6 +196,7 @@ def tool_run_logging_kwargs(self) -> Dict:
         return {}
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 199: for this, can we say to use openai_tools_agent?

libs/langchain/langchain/agents/agent.py
```diff
@@ -568,6 +579,7 @@ def tool_run_logging_kwargs(self) -> Dict:
         }
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 582: same message as above?

libs/langchain/langchain/agents/agent_types.py
```diff
@@ -1,7 +1,10 @@
 """Module definitions of agent types together with corresponding agents."""
 from enum import Enum
 
+from langchain_core._api import deprecated
 
+
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 7: same message as above?

libs/langchain/langchain/agents/chat/base.py
```diff
@@ -24,6 +25,7 @@
 from langchain.chains.llm import LLMChain
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 28: can we point to a class - i think create_react

libs/langchain/langchain/agents/initialize.py
```diff
@@ -10,6 +11,7 @@
 from langchain.callbacks.base import BaseCallbackManager
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 14: can we have alternative

libs/langchain/langchain/agents/openai_functions_multi_agent/base.py
```diff
@@ -93,6 +94,7 @@ def _parse_ai_message(message: BaseMessage) -> Union[List[AgentAction], AgentFin
     )
 
 
+@deprecated("0.1.0", removal="0.2.0")
```
hwchase17, line 97: openai_tools as alternative

hwchase17 (approved):

libs/langchain/langchain/agents/agent.py
baskaryan: thought about that, only issue is tools requires latest models. but maybe that's fine?

baskaryan: oh this is the base agent — seems like openai tools is too specific for this?

was referring to openai multi functions agent being slightly different from openai tools agent in comment above

baskaryan: think for abstract base classes probably don't need alternative?

PR #15568, opened 2024-01-05
community: fix: should flush after inserting data on milvus

EeyoreLee:
The inserted data cannot take effect immediately. We should flush after inserting data on milvus.


hwchase17 (approved):

JamesBonddu:
flush to milvus slowly

image

JamesBonddu:
col.flush() to milvus slowly needs 170s

EeyoreLee:
@JamesBonddu - hi, auto flush can't make the data available immediately. For some case, I need search the data once inserting data. I think we need a knobs to tradeoff whether auto flush or not. BTW, could you let me know how much data is there in total that make the `flush` cost 170s.

PR #15575, opened 2024-01-05
core[minor]: Add llm cache in stream

pprados:
**Description:** 
The current implementation does not exploit the cache when using stream.

This PR add the usage of the llm cache with the stream.

**Twitter handle:**
@pprados

### DIFF
--- a/libs/core/langchain_core/language_models/llms.py
+++ b/libs/core/langchain_core/language_models/llms.py
@@ -70,11 +70,11 @@ def _log_error_once(msg: str) -> None:
 
 
 def create_base_retry_decorator(
-        error_types: List[Type[BaseException]],
-        max_retries: int = 1,
-        run_manager: Optional[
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/language_models/llms.py
```diff
@@ -441,17 +441,37 @@ async def astream(
...
+                    ):
+                        yield chunk.text
+                        if generation is None:
+                            generation = chunk
+                        else:
+                            generation += chunk
+                    assert generation is not None
+                else:
+                    generation = GenerationChunk(text=existing_prompts[0][0].text)
+                    yield generation
```
baskaryan, line 473: we should be yielding strings not GenerationChunks

baskaryan (commented):
could we add some unit tests

eyurtsev:
@pprados i'm going to close this PR for now due to lack of activity, if you want to make a fix for supporting cache in streaming for llms or chat models feel free to make a new PR!

PR #15589, opened 2024-01-05
core[patch]: add beta decorator

baskaryan:


libs/core/langchain_core/_api/beta.py
```diff
@@ -0,0 +1,258 @@
...
+    name: str = "",
+    obj_type: str = "",
+    addendum: str = "",
+) -> None:
+    """Display a standardized beta annotation.
+
+    Arguments:
+        message : str, optional
+            Override the default beta message. The
+            %(name)s, %(obj_type)s, %(addendum)s
```
eyurtsev, line 226: Format specifies are out of date since we're using f strings

eyurtsev (approved):

PR #15591, opened 2024-01-05
langchain: Add dynamic num_variations to MultiQueryRetriever

rhamnett:
- **Description:** Add dynamic num_variations to MultiQueryRetriever


hwchase17 (commented):
mmm im not sure num variations gets used

also - the user can pass in the prompt, so if they want they can just override the prompt passed in. that seems like the better way

PR #15595, opened 2024-01-05
Allow system message add to history

Sypherd:
## Description
This is a follow-up PR to https://github.com/langchain-ai/langchain/pull/14967 that added support for `AIMessage` and `HumanMessage` to `BaseChatMessageHistory.add_user_message` and `BaseChatMessageHistory.add_ai_message`. This extends that to include a `BaseChatMessageHistory.add_system_message` method with the same functionality. It also clarifies the docstrings of the other two methods.

## Dependencies
None

## Tag Maintainers
@hinthornw 
@eyurtsev 
@butlfrazp

## Note
I ran `make lint` and `make format` but `make test` results in `make: *** No rule to make target 'test'.  Stop.`.

### DIFF
--- a/libs/core/langchain_core/chat_history.py
+++ b/libs/core/langchain_core/chat_history.py
@@ -7,6 +7,7 @@ from langchain_core.messages import (
     AIMessage,
     BaseMessage,
     HumanMessage,
+    SystemMessage,
     get_buffer_string,
 )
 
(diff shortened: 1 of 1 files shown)


eyurtsev (changes requested):
hi @Sypherd,

I am not sure that we should be adding these methods.

The client should be using `add_message`. `add_ai_message` and `add_user_message`. I'd prefer marking add_ai_message and add_user_message for deprecation and eventually delete them entirely instead of adding additional methods

hwchase17:
agree with @eyurtsev

Sypherd:
See [this comment](https://github.com/langchain-ai/langchain/pull/15709#issuecomment-1884951129) for a reply to the above comments.

PR #15625, opened 2024-01-06
community: AWS Athena Document Loader

abhijeethp:
- **Description:** Adds the document loader for [AWS Athena](https://aws.amazon.com/athena/), a serverless and interactive analytics service.
  - **Dependencies:** Added boto3 as a dependency


hwchase17 (commented):
please do not update langchain, should be imported from langchain_community from now on

other than that, lgtm!

abhijeethp:
@hwchase17 I have removed the extraneous imports from langchain... but I believe the rest should stay unless you are saying that the end users would no longer be importing this as `from langchain.document_loaders import AthenaLoader` and rather would be importing as `from langchain_community.document_loaders.athena import AthenaLoader`

please confirm and I will remove the rest of the changes under langchain as well.

libs/community/langchain_community/document_loaders/athena.py
```diff
@@ -0,0 +1,131 @@
+from __future__ import annotations
+
+import io
+import json
+import time
+from typing import Any, Dict, Iterator, List, Optional, Tuple
+
+import pandas as pd
```
baskaryan, line 8: pandas isn't a required dependency so it shouldn't be imported at the top level. should instead import in the functions that use it

abhijeethp: Thanks for the feedback. 
I Have made this change

abhijeethp:
@hwchase17 @baskaryan please review

baskaryan (commented):
cc @3coins

3coins (commented):
@abhijeethp 
Thanks for working on these changes. Left some comments about the athena and s3 clients, and some suggestions on re-organizing the code. You also seem to have minor failures in the attached notebook, can you run the linter and fix these errors. Let me know if you need any help with linting or fixing any of these errors.

abhijeethp:
@3coins Thanks for the feedback! I have made the suggested changes. 
I am unfortunately not able to fix the linting issues in the python notebooks. 
I do not see a way to run linter in the docs directory... Please let me know how i can fix them.

baskaryan:
> @3coins Thanks for the feedback! I have made the suggested changes. I am unfortunately not able to fix the linting issues in the python notebooks. I do not see a way to run linter in the docs directory... Please let me know how i can fix them.

running `make format; make lint` from root of repo should run docs formatting and linting! happy to help if you want to give me write access to your fork

abhijeethp:
@baskaryan Thanks for the feedback. 
I was able to fix the formatting and linting issues by running it from root.

abhijeethp:
@hwchase17 why did was the PR closed?

baskaryan:

abhijeethp:
@3coins I have made the changes suggested above. 
Please review it again when possible.
Would love to have this be part of the library soon! :D

hwchase17 (approved):

abhijeethp:
@hwchase17 I had to make a change to get it working for python 3.8 
https://github.com/langchain-ai/langchain/pull/15625/commits/1428649aaa330840aa15697a5211f445159f831e

These builds only ran on approval so i could not catch it earlier. 
Please review it once again

PR #15635, opened 2024-01-06
cli[patch]: bump pkg tmpl langchain verion

baskaryan:


efriis:
Is this off an old version of master? I think langchain isn't a default dep anymore - tying to integration deps.

Happy to go either way. Feel like better to move towards core+community dep instead of broad langchain one by default.

Issue #15355, opened 2023-12-31
AzureSearch semantic_hybrid_search fails due to hardcoding of metadata fields


nohanaga:
### System Info

azure-search-documents==11.4.0b8
langchain==0.0.352

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [X] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction
I have my own search index with no `metadata` field.

#### Code
```python
from langchain.vectorstores.azuresearch import AzureSearch
vector_store: AzureSearch = AzureSearch(
    azure_search_endpoint=vector_store_address,
    azure_search_key=vector_store_password,
    index_name=index_name,
    embedding_function=embeddings.embed_query,
    semantic_configuration_name="default"
)

query = "How many employees does Contoso Electronics have?"  
docs = vector_store.semantic_hybrid_search(
    query=query,
    search_type="semantic_hybrid",
)
print(docs[0])
```

#### Stack trace
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File \lib\site-packages\langchain_community\vectorstores\azuresearch.py:656, in <listcomp>(.0)
    622     semantic_answers_dict[semantic_answer.key] = {
    623         "text": semantic_answer.text,
    624         "highlights": semantic_answer.highlights,
    625     }
    626 # Convert results to Document objects
    627 docs = [
    628     (
    629         Document(
    630             page_content=result.pop(FIELDS_CONTENT),
    631             metadata={
    632                 **(
    633                     {FIELDS_ID: result.pop(FIELDS_ID)}
    634                     if FIELDS_ID in result
    635                     else {}
    636                 ),
    637                 **(
    638                     json.loads(result[FIELDS_METADATA])
    639                     if FIELDS_METADATA in result
    640                     else {
    641                         k: v
    642                         for k, v in result.items()
    643                         if k != FIELDS_CONTENT_VECTOR
    644                     }
    645                 ),
    646                 **{
    647                     "captions": {
    648                         "text": result.get("@search.captions", [{}])[0].text,
[...]

Issue #15039, opened 2023-12-21
AzureSearch Bug -- langchain.vectorstores.azuresearch 


RERobbins:
### System Info

The example found  [here](https://python.langchain.com/docs/integrations/vectorstores/azuresearch) and in particular this code fragment
```
embeddings: OpenAIEmbeddings = OpenAIEmbeddings(deployment=model, chunk_size=1)
index_name: str = "langchain-vector-demo"
vector_store: AzureSearch = AzureSearch(
    azure_search_endpoint=vector_store_address,
    azure_search_key=vector_store_password,
    index_name=index_name,
    embedding_function=embeddings.embed_query,
)
```
fails with a message that at the top is:
```
vector_search_configuration is not a known attribute of class <class 'azure.search.documents.indexes.models._index.SearchField'> and will be ignored
semantic_settings is not a known attribute of class <class 'azure.search.documents.indexes.models._index.SearchIndex'> and will be ignored
```
and culminates with:
```
HttpResponseError: (InvalidRequestParameter) The request is invalid. Details: definition : The vector field 'content_vector' must have the property 'vectorSearchConfiguration' set.
Code: InvalidRequestParameter
Message: The request is invalid. Details: definition : The vector field 'content_vector' must have the property 'vectorSearchConfiguration' set.
Exception Details:  (InvalidField) The vector field 'content_vector' must have the property 'vectorSearchConfiguration' set. Parameters: definition
    Code: InvalidField
    Message: The vector field 'content_vector' must have the property 'vectorSearchConfiguration' set. Parameters: definition
```
I am running with `python 3.10`, `openai 1.5.0`, `langchain 0.0.351` and `azure-search-documents 11.4.0`.  If I revert to `azure-search-documents 11.4.0b8` the code works.

This appears to be related to the November 2023 Microsoft API change which introduced the concept of a "profile" which aggregates various vector search settings under one name. As a result of this API change, the old "vector_search_configuration" was deprecated and a new "vector_search_profile" was added, along with a new "profiles" object.  It seems that the Langchain extension has not been updated for this change and expects the old "vector_search_configuration" property which doesn't exist on newer SDK releases.

See the discussion [here](https://stackoverflow.com/questions/77682544/vector-search-configuration-is-not-a-known-attribute-of-class-class-azure-sear/77694628#77694628).

### Who can help?

@hwchase17 
@bas

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

[...]

PR #14389, opened 2023-12-07
Azure DocumentIntelligenceLoader/Parser support update with latest SDK

zifeiq:
- **Description:**
 Add DocumentIntelligenceLoader & DocumentIntelligenceParser implementation using the latest Azure Document Intelligence SDK with markdown support.
The core logic resides in DocumentIntelligenceParser and DocumentIntelligenceLoader is a mere wrapper of the parser.
The parser will takes api_endpoint and api_key and creates DocumentIntelligenceClient for the user. 4 parsing modes are supported:
1. Markdown (default)
2. Single
3. Page 
4. Object

UT and notebook are also updated accordingly.

  - **Dependencies:** Azure Document Intelligence SDK: azure-ai-documentintelligence [azure-sdk-for-python/sdk/documentintelligence/azure-ai-documentintelligence at 7c42462ac662522a6fd21b17d2a20f4cd40d0356 · Azure/azure-sdk-for-python (github.com)](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Fazure-sdk-for-python%2Ftree%2F7c42462ac662522a6fd21b17d2a20f4cd40d0356%2Fsdk%2Fdocumentintelligence%2Fazure-ai-documentintelligence&data=05%7C01%7CZifei.Qian%40microsoft.com%7C298225aa3e31468a863108dbf07374ff%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638368150928704292%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=oE0Sl4HERnMKdbkV9KgBV46Z2xytcQAShdTWf7ZNl%2Bs%3D&reserved=0).

PR #13472, opened 2023-11-16
Update azuresearch.py following recent change from azure-search-documents library

guillaumedelande:
- **Description:** 

Reference library azure-search-documents has been adapted in version 11.4.0:

1. Notebook explaining Azure AI Search updated with most recent info
2. HnswVectorSearchAlgorithmConfiguration --> HnswAlgorithmConfiguration
3. PrioritizedFields(prioritized_content_fields) --> SemanticPrioritizedFields(content_fields)
4. SemanticSettings --> SemanticSearch
5. VectorSearch(algorithm_configurations) --> VectorSearch(configurations)

--> Changes now reflected on Langchain: default vector search config from langchain is now compatible with officially released library from Azure.

  - **Issue:**
Issue creating a new index (due to wrong class used for default vector search configuration) if using latest version of azure-search-documents with current langchain version
  - **Dependencies:** azure-search-documents>=11.4.0,
  - **Tag maintainer:** ,

PR #15659, opened 2024-01-07
community: update AzureSearch class to work with azure-search-documents=11.4.0

lz-chen:
- **Description:** Updates `libs/community/langchain_community/vectorstores/azuresearch.py` to support the stable version `azure-search-documents=11.4.0` 
  - **Issue:** https://github.com/langchain-ai/langchain/issues/14534, https://github.com/langchain-ai/langchain/issues/15039, https://github.com/langchain-ai/langchain/issues/15355
  - **Dependencies:** azure-search-documents>=11.4.0


lz-chen:
Hello, there is a previous PR [#13472]( https://github.com/langchain-ai/langchain/pull/13472) that provided some partial fix to the `AzureSearch` class but did not fully update the dependency. Now calling class as vector store still doesn't work in my case, with the problems encontered  by others in the issues mentioned in the description. 
I'd like to contribute to update the dependency to `azure-search-documents=11.4.0`. However I am not sure if that means updating all the `pyproject.toml` and `poetry.lock` files that has `azure-search-documents`? What is the best way to do that? @efriis @hwchase17

efriis:
Hey @lz-chen ! You actually can't do this with a community integration, as it's not a concrete dependency. Unfortunately, people need to manage their other python package versions themselves. 

A good addition would be a note in the documentation to upgrade that package!

lz-chen:
@efriis Thanks for the reply 😊 What does that mean for this PR? Should I keep the changes I made in `libs/community/langchain_community/vectorstores/azuresearch.py` without updating the dependencies in `libs/langchain/pyproject.toml` and `libs/langchain/poetry.lock`? 
Or follow what's done in PR [#13472]( https://github.com/langchain-ai/langchain/pull/13472) to keep support for both `azure-search-documents>=11.4.0` and `azure-search-documents<11.4.0`?

HeidiSteen:
@mattgotteiner for subject matter expertise on Azure Search SDK migration to 11.4 stable library.  Agree with @lz-chen that the existing vector store integration module doesn't work with 11.4.0.

Skar0:
Hi @lz-chen,

Thanks for merging my PR into your update! I reviewed the original changes and tested them with `azure-search-documents==11.4.0`, and everything seems to work just fine :) My PR aimed to address a specific bug related to the removal of `query_language` in `azure-search-documents==11.4.0`. I'm happy to discuss any further details or assist with additional improvements if needed.

konradbjk:
Looking forward to getting this merged. Due to this issue I cannot use Azure Search as a vector store

mattgotteiner:
Thank you for making this update - the changes look good

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -155,77 +140,56 @@ def fmt_err(x: str) -> str:
...
-                                ],
-                            ),
-                        )
-                    ]
-                )
+        if semantic_configurations is None and semantic_configuration_name is not None:
+            semantic_configuration = SemanticConfiguration(
+                name=semantic_configuration_name,
+                prioritized_fields=SemanticPrioritizedFields(
+                    # title_field=SemanticField(field_name="title"),
```
pamelafox, line 180: Why are these 2 lines commented out? Should they be deleted?

pamelafox (approved):
Looks good! I assume there are no tests that need updating. (GitHub shows me as a reviewer but an actual maintainer will need to review and approve workflows)

pamelafox:
Just looking at the discussion here about pyproject.toml-

It looks like the main langchain pyproject.toml doesn't reference azure-search-documents, which makes sense as this in community. However, I notice that the langchain-community pyproject.toml also doesn't reference azure-search-documents. I believe that it probably should, just based off similar recent integration updates, like this PR for Azure Document Intelligence:
https://github.com/langchain-ai/langchain/pull/14389

The file to change:
https://github.com/langchain-ai/langchain/blob/ac970c9497e2aca1f6396c3f6954b4f6cd0ac879/libs/community/pyproject.toml#L86
Notably it'd be marked as optional=True, like the other community dependencies.

That's my suggestion after looking through how the packages are structured. Perhaps @efriis could confirm. Thanks!

baskaryan:
> Just looking at the discussion here about pyproject.toml-
> 
> It looks like the main langchain pyproject.toml doesn't reference azure-search-documents, which makes sense as this in community. However, I notice that the langchain-community pyproject.toml also doesn't reference azure-search-documents. I believe that it probably should, just based off similar recent integration updates, like this PR for Azure Document Intelligence: #14389
> 
> The file to change:
> 
> https://github.com/langchain-ai/langchain/blob/ac970c9497e2aca1f6396c3f6954b4f6cd0ac879/libs/community/pyproject.toml#L86
> 
> 
> Notably it'd be marked as optional=True, like the other community dependencies.
> That's my suggestion after looking through how the packages are structured. Perhaps @efriis could confirm. Thanks!

Optional dependencies only need to be included in pyproject if they're required for unit tests, so no need to add in this case!

baskaryan (approved):

pamelafox:
@baskaryan Ah, thanks for clarifying, that explains why azure-document-intelligence was added in that PR, as it had a short unit test.

mattgotteiner:
We are also missing an `from azure.search.documents.indexes.models import VectorSearch`

mattgotteiner:
the import is only performed with `typing.TYPE_CHECKING` is true

kristapratico:
@mattgotteiner noticed these issues today as well and opened a PR to fix here: https://github.com/langchain-ai/langchain/pull/17599

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -451,39 +413,30 @@ def vector_search_with_score(
...
         docs = [
             (
                 Document(
                     page_content=result.pop(FIELDS_CONTENT),
-                    metadata={
-                        **(
-                            {FIELDS_ID: result.pop(FIELDS_ID)}
-                            if FIELDS_ID in result
-                            else {}
-                        ),
```
baskaryan, line 477: hey @lz-chen @mattgotteiner, was there a specific reason we removed FIELDS_ID in this pr?

PR #15663, opened 2024-01-07
community: Allow including vectors with text missing for Pinecone

ristomcgehee:
- **Description:** 
  This PR adds the option to include vectors that don't have page content. Currently, any such vectors will be skipped. This change can be useful when you're only concerned about the similarity score and not the page content. I'm a contributor to [Rebuff](https://github.com/protectai/rebuff), and we are working on adding a standalone Python SDK. When we query Pinecone, we only need the vector scores. Currently with Langchain, if a Rebuff user sets up their Pinecone database without a metadata field named `"text"`, they will not receive any vector scores when querying Pinecone. With this PR, it won't matter the name of the metadata field used for page content.
For what it's worth, langchain-js does not skip vectors when the page content cannot be found: [pinecone.ts](https://github.com/langchain-ai/langchainjs/blob/8154c7829fc930a5b4060e830d06cbdd3e9484c4/libs/langchain-community/src/vectorstores/pinecone.ts#L230-L234).
I didn't update the Pinecone integration tests because I could not get the tests to pass (even before my changes). They were repeatedly failing because Pinecone is eventually consistent and the tests were assuming the changes would be applied immediately. 
  - **Issue:** N/A
  - **Dependencies:** No changes needed

### DIFF
--- a/docs/docs/integrations/vectorstores/pinecone.ipynb
+++ b/docs/docs/integrations/vectorstores/pinecone.ipynb
@@ -212,6 +212,31 @@
     "for i, doc in enumerate(found_docs):\n",
     "    print(f\"{i + 1}.\", doc.page_content, \"\\n\")"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0716424c",
+   "metadata": {},
+   "source": [
+    "### Including Documents with Missing Text Key\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


ristomcgehee:
@eyurtsev, Could you take a look at this PR?

baskaryan:
@ristomcgehee the main pinecone integration has been moved to it's own package in libs/partners/pinecone, could we move this change there?

ristomcgehee:
I'll close this PR and open a new one when I have time to make that change.

PR #15675, opened 2024-01-08
community: fix myscale delete function bug

IANTHEREAL:
Now the SQL used to delete vector doc from myscale is as follow:
```sql
DELETE FROM collection WHERE id = '1' AND id = '2' AND id = '3'
```

But the expected one should be 

```sql
DELETE FROM collection WHERE id IN ('1', '2', '3')
```


hwchase17 (approved):

hwchase17:
thanks!

PR #15699, opened 2024-01-08
Fix AstraDB logical operator filtering

cbornet:
This change fixes the AstraDB logical operator filtering (`$and,` `$or`).
The `metadata` prefix must not be added if the key is `$and` or `$or`.


cbornet:
cc @hemidactylus

hwchase17 (approved):
since @hemidactylus gave it the thumbs up, going to assume this is good to go!

PR #15709, opened 2024-01-08
core(minor): Add bulk add messages to BaseChatMessageHistory interface

eyurtsev:
* Add bulk add_messages method to the interface.
* Update documentation for add_ai_message and add_human_message to denote them as being marked for deprecation. We should stop using them as they create more incorrect (inefficient) ways of doing things

### DIFF
--- a/libs/core/langchain_core/chat_history.py
+++ b/libs/core/langchain_core/chat_history.py
@@ -17,6 +17,7 @@ class BaseChatMessageHistory(ABC):
     See `ChatMessageHistory` for default implementation.
 
     Example:
+
         .. code-block:: python
 
             class FileChatMessageHistory(BaseChatMessageHistory):
@@ -29,8 +30,13 @@ class BaseChatMessageHistory(ABC):
                        messages = json.loads(f.read())
                     return messages_from_dict(messages)
 
-               def add_message(self, message: BaseMessage) -> None:
-                   messages = self.messages.append(_message_to_dict(message))
+               def add_messages(self, messages: Sequence[BaseMessage]) -> None:
+                   all_messages = list(self.messages) # Existing messages
+                   all_messages.extend(messages) # Add new messages
+
+                   serialized = [message_to_dict(message) for message in all_messages]
+                   # Can be further optimized by only writing new messages
+                   # using append mode.
                    with open(os.path.join(storage_path, session_id), 'w') as f:
                        json.dump(f, messages)
 
@@ -45,6 +51,10 @@ class BaseChatMessageHistory(ABC):
     def add_user_message(self, message: Union[HumanMessage, str]) -> None:
         """Convenience method for adding a human message string to the store.
 
+        Deprecation Warning: This method will likely be deprecated in a future release.
+            Code should use add_messages instead which can be made more efficient
+            across implementations.
+
         Args:
             message: The human message to add
         """
@@ -56,6 +66,10 @@ class BaseChatMessageHistory(ABC):
     def add_ai_message(self, message: Union[AIMessage, str]) -> None:
         """Convenience method for adding an AI message string to the store.
 
+        Deprecation Warning: This method will likely be deprecated in a future release.
+            Code should use add_messages instead which can be made more efficient
+            across implementations.
+
         Args:
             message: The AI message to add.
         """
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/chat_history.py
```diff
@@ -45,6 +51,10 @@ def clear(self):
     def add_user_message(self, message: Union[HumanMessage, str]) -> None:
         """Convenience method for adding a human message string to the store.
```
eyurtsev, line 53: @baskaryan is it safe to use the deprecation functionality here. I want to mark this as pending deprecation rather than deprecated. Wasn't sure where we are right now in terms of surfacing warnings -- I don't think this should be surfaced to users especially if it's not directly used by user code

eyurtsev:
Hi @Sypherd, this is what I'm thinking for the BaseChatMessageHistory. I'd like everything to move to add_messages long term so that when message history is stored in a database, we can save on unnecessary round trips when multiple messages need to be written.

Related: https://github.com/langchain-ai/langchain/pull/15595

Sypherd:
> Hi @Sypherd, this is what I'm thinking for the BaseChatMessageHistory. I'd like everything to move to add_messages long term so that when message history is stored in a database, we can save on unnecessary round trips when multiple messages need to be written.
> 
> Related: #15595

Hi @eyurtsev, thanks for showing me this. I think bulk `add_messages` is great. I don't think having adds for specific message types (e.g., `add_user_message`) takes away from that or the functionality of `add_messasge`. As a user, I think those explicit message methods are useful for redundant clarity. For example, I prefer `add_user_message(human_prompt)` over `add_message(human_prompt)` just so that it is immediately clear from the function name what type of addition I'm making to the history. 
That's a trivial example and more personal preference but there are instances where that redundancy is helpful. For more complex prompting architectures, such as [self-ask](https://arxiv.org/pdf/2210.03350.pdf) and [Multi Chain Reasoning](https://arxiv.org/pdf/2304.13007.pdf), having explicit add methods is the easiest way to succinctly increase the clarity of the flow.

libs/core/langchain_core/chat_history.py
```diff
@@ -64,18 +78,38 @@ def add_ai_message(self, message: Union[AIMessage, str]) -> None:
...
         """Add a Message object to the store.
 
         Args:
             message: A BaseMessage object to store.
         """
-        raise NotImplementedError()
+        if type(self).add_messages != BaseChatMessageHistory.add_messages:
+            # This means that the sub-class has implemented an efficient add_messages
+            # method, so we should usage of add_message to that.
+            self.add_messages([message])
```
Sypherd, line 90: This creates an infinite loop if the child class only implements `add_messages` and not `add_message`

Sypherd (changes requested):

libs/core/langchain_core/chat_history.py
eyurtsev: If i'm not mistaken, the conditional immediately above prevents an infinite loop, it's based on checking whether add_messages has been overloaded or not. 

Check the included unit tests that verifies that things work properly when either add_messages or add_message are implemented

Sypherd: Oh yeah, you're right! Sorry, I mistook the class, so it won't be a breaking change. The only two cases it currently wouldn't catch are if someone implemented `add_messages` to call `add_message` without overriding `add_message` (e.g., to simply add logging) or used the class as-is (overriding `clear`, of course), calling only `add_messages` and `add_message`. Either of those would result in infinite loops rather than raising `NotImplementedError`. It makes sense to me to either ensure that `NotImplementedError` would always be raised or `@abstractmethod` would be added back to `add_message` so it is abundantly clear what methods are abstract methods.

eyurtsev: Yeah :( but that's a breaking change for all implementations. if I could make a breaking change, i'd delete `add_message` entirely :). 

Messages should not be added one at a time since production use cases, writing a message usually involves a network request, and the round trips on network requests become expensive quickly.

libs/core/langchain_core/chat_history.py
```diff
@@ -64,18 +78,38 @@ def add_ai_message(self, message: Union[AIMessage, str]) -> None:
...
+        else:
+            raise NotImplementedError(
+                "add_message is not implemented for this class. "
+                "Please implement add_message or add_messages."
+            )
+
+    def add_messages(self, messages: List[BaseMessage]) -> None:
```
baskaryan, line 97: Sequence[BaseMessage]?

libs/core/langchain_core/chat_history.py
baskaryan: but also is it so bad to keep these around as convenience methods? think more important thing would just be to emphasize/only show add_messages in docs

eyurtsev:
cc @cbornet -- I'm adding a bulk interface. Going to merge, and then will commandeer your PR that adds async functionality to incorporate the relevant changes

baskaryan (approved):

PR #15714, opened 2024-01-08
langchain(patch): Fix output type for pydantic output parser

eyurtsev:
This PR fixes the output type for the pydantic output parser.

Fix for: https://github.com/langchain-ai/langserve/issues/301

### DIFF
--- a/libs/langchain/langchain/output_parsers/pydantic.py
+++ b/libs/langchain/langchain/output_parsers/pydantic.py
@@ -15,7 +15,11 @@ class PydanticOutputParser(BaseOutputParser[T]):
     """Parse an output using a pydantic model."""
 
     pydantic_object: Type[T]
-    """The pydantic model to parse."""
+    """The pydantic model to parse.
+    
+    Attention: To avoid potential compatibility issues, it's recommended to use
+        pydantic <2 or leverage the v1 namespace in pydantic >= 2.
+    """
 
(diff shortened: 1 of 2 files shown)


nfcampos (approved):

libs/langchain/langchain/output_parsers/pydantic.py
```diff
@@ -51,3 +55,8 @@ def get_format_instructions(self) -> str:
     @property
     def _type(self) -> str:
         return "pydantic"
+
+    @property
+    def OutputType(self) -> Type[T]:
```
nfcampos, line 60: Actually should this be overriding get_output_schema instead?

eyurtsev: Not sure! That's what the parent parsers do

and in Runnables:


https://github.com/langchain-ai/langchain/blob/aad4865e99ea35be12043b3b46c026bb968a72f8/libs/core/langchain_core/runnables/base.py#L257-L257

https://github.com/langchain-ai/langchain/blob/aad4865e99ea35be12043b3b46c026bb968a72f8/libs/core/langchain_core/runnables/base.py#L324-L324

It seems like I'm meant to over-ride `OutputType` -- or do you think that's wrong?

nfcampos: actually this is fine, leave as is

PR #15723, opened 2024-01-08
core[patch]: Replace beta decorator with warning on init to avoid breaking type annotations

eyurtsev:
Replace beta decorator with warning on init to avoid breaking type annotations


nfcampos (changes requested):
We'd want this to respect the new logic that skips warnings when usage comes from within our libraries

nfcampos:
This is actually correct

PR #15726, opened 2024-01-08
community: prometheus metrics callback handler

zzzming:
Description:

Create a new MetricsCallbackHandler to collect token usage and latency metrics for openai. Prometheus metrics are crucial for modern monitoring and enterprise applications because they provide a powerful and flexible way to collect and aggregate real-time metrics about system performance and health. This enables organizations to proactively identify and address issues, ensuring high availability and reliability of their services.

It only collects metrics. It's still up to the application to expose them 

Here is an example how to expose metrics in user application. This is not part of this PR which only collects metrics.
```
from flask import Flask, Response, make_response

# Create the Flask App
app = Flask(__name__)

@app.route('/metrics')
def metrics():
    return Response(generate_latest(), mimetype=CONTENT_TYPE_LATEST)
```

Dependencies:
prometheus_client-0.19.0 is already specified under libs/community's poetry.

Twitter handle: itestmycode


nicoloboschi (approved):

zzzming:
@baskaryan, @eyurtsev, @hwchase17 can you help review and merge this PR Thank you

libs/community/langchain_community/callbacks/openai_info.py
```diff
@@ -1,10 +1,13 @@
 """Callback Handler that prints to std out."""
 import threading
-from typing import Any, Dict, List
+import os
+from typing import Any, Dict, List, Optional
 
 from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.outputs import LLMResult
 
+from prometheus_client import Counter, Gauge
```
baskaryan, line 9: prometheus_client isn't a langchain-community required dependency so we should only import it within the functions that need it

hinthornw (changes requested):
If the goal is to make an easy integration to send token counting metrics to prometheus, the way to do that is to make a separate prometheus callbackhandler that you provide whenever invoking an option. The OpenAI callback handler should NOT be required to have an awareness of prometheus.


You can do this by subclassing the `BaseCallbackHandler` (or somewhat more conveniently, the `Basetracer`)

zzzming:
@hinthornw as you suggested, I created a separate callback for Prometheus metrics. 
cc @baskaryan

efriis:
closing this one as stale / not passing CI. Happy to discuss if still active!

PR #15743, opened 2024-01-09
adding parameter for changing the language in SpacyEmbeddings

MarinaPlius:
Description: Added the parameter for a possibility to change a language model in SpacyEmbeddings. The default value is still the same: "en_core_web_sm", so it shouldn't affect a code which previously did not specify this parameter, but it is not hard-coded anymore and easy to change in case you want to use it with other languages or models.

Issue: At Barcelona Supercomputing Center in Aina project (https://github.com/projecte-aina), a project for Catalan Language Models and Resources, we would like to use Langchain for one of our current projects and we would like to comment that Langchain, while being a very powerful and useful open-source tool, is pretty much focused on English language. We would like to contribute to make it a bit more adaptable for using with other languages.

Dependencies: This change requires the Spacy library and a language model, specified in the model parameter.

Tag maintainer: @dev2049

Twitter handle: @projecte_aina


hwchase17 (commented):
we should use a different key - its kinda confusing that its both the name and the class object

MarinaPlius:
@hwchase17 Changed the name of the variable, please have a look.
@baskaryan

hwchase17:
@MarinaPlius it is using the `nlp` variable for both the name `en_web..` and the SpaCy object. Let's make those different, eg keep the nlp variable as is, but add a `model_name` parameter or the like

MarinaPlius:
@hwchase17 Thank you so much for the clarification! I believe, I got it this time! Please, have a look

hwchase17 (commented):
for backwards compat, consistency, and clarity, i would keep nlp to mean the model (as it does now) and make the model name have the name `model_name` or something

MarinaPlius:
@hwchase17 Sure! Changed it again!

hwchase17 (approved):

PR #15763, opened 2024-01-09
experimental[patch]: loosen core max version

baskaryan:


efriis (commented):
Aren't these equivalent?

https://jubianchi.github.io/semver-check/#/^0.1.7/0.1.9

https://python-poetry.org/docs/dependency-specification/#caret-requirements

baskaryan:
ah you're right, issue is something else

PR #14650, opened 2023-12-13
Add CometLLM integration notebook example

Lothiraldan:
I also added LANGCHAIN_COMET_TRACING to enable the CometLLM tracing integration similar to other tracing integrations. This is easier for end-users to enable it rather than importing the callback and pass it manually.

PR #15765, opened 2024-01-09
community: Add CometLLM integration notebook example

Lothiraldan:
I also added LANGCHAIN_COMET_TRACING to enable the CometLLM tracing integration similar to other tracing integrations. This is easier for end-users to enable it rather than importing the callback and pass it manually.

(This is the same content as https://github.com/langchain-ai/langchain/pull/14650 but rebased and squashed as something seems to confuse Github Action).


docs/docs/integrations/providers/comet_tracing.ipynb
baskaryan, line 1: would it make more sense to put this doc in integrations/callbacks section? https://python.langchain.com/docs/integrations/callbacks

Lothiraldan: Make sense, I moved it there

baskaryan (approved):
thanks @Lothiraldan!

PR #15768, opened 2024-01-09
community: Fix String-List Comparison Bug in AzureAIDocumentIntelligenceParser

timschonborn:
This pull request addresses a bug found in `AzureAIDocumentIntelligenceParser` where a string was incorrectly being compared to a list. 

The code currently uses the expression `self.mode == ["page"]` to check which mode the user has selected. In Python, this comparison is always `False` (self.mode is asserted to be str in line 35), thus the if-statement would always result in the execution of `_generate_docs_object` instead of `_generate_docs_page`.

The fix is to remove the brackets resulting in a correct comparison.


hwchase17 (commented):
thanks!

baskaryan (approved):

baskaryan:
looks like this was fixed in #15958, thank you anyways for the contribution @timschonborn!

timschonborn:
Oh apologies, I tried to check if a similar PR wasn't pending already!

PR #15774, opened 2024-01-09
templates: robocorp action server template

efriis:
- [x] publish integration package
- [ ] update deps/lockfile to reflect published version


efriis:
Edit issues again. Doing in #15776

PR #15784, opened 2024-01-09
community[patch]: ChatDeepInfra

efriis:
- deepinfra support for chat_models
- update test model
- cr


ichernev:
but now I can't edit it (and fix the build ;-/)

ichernev:
@efriis  I believe you need to add `ChatDeepInfra` in `tests/unit_tests/chat_models/test_imports.py`.

PR #15792, opened 2024-01-10
langchain[patch]: Improve stream_log with AgentExecutor and Runnable Agent

eyurtsev:
This PR fixes an issue where AgentExecutor with RunnableAgent does not allow
users to see individual llm tokens if streaming=True is not set explicitly on
the underlying chat model.

The majority of this PR is testing code:

1. Create a test chat model that makes it easier to test streaming and supports
   AIMessages that include function invocation information.
2. Tests for the chat model
3. Tests for RunnableAgent (previously untested)
4. Tests for openai agent (previously untested)

### DIFF
--- a/libs/langchain/langchain/agents/agent.py
+++ b/libs/langchain/langchain/agents/agent.py
@@ -371,7 +371,7 @@ class RunnableAgent(BaseSingleActionAgent):
         callbacks: Callbacks = None,
         **kwargs: Any,
     ) -> Union[AgentAction, AgentFinish]:
-        """Given input, decided what to do.
+        """Based on past history and current inputs, decide what to do.
 
         Args:
             intermediate_steps: Steps the LLM has taken to date,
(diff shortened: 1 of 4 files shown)


eyurtsev:
Related issues/discussion:
https://github.com/langchain-ai/langchain/discussions/14573
https://github.com/langchain-ai/langserve/issues/314
https://github.com/langchain-ai/langchain/discussions/12699
https://stackoverflow.com/questions/77690231/how-to-stream-the-output-of-an-agentexecutor-in-langchain-to-my-final-applicatio
https://github.com/langchain-ai/langchain/discussions/15755
https://github.com/langchain-ai/langchain/discussions/13777

nfcampos (approved):

jgroetzi:
After this change I'm no longer able to invoke agents without streaming. My endpoint does not support streaming and now this has broken all of my agent code and I can't find a way to workaround other than pinning to a langchain package version prior to this change (i.e. 0.1.0), which means now i can't get updates :(
maybe I'm missing something, but specifically here for example:
https://github.com/langchain-ai/langchain/commit/feb41c5e288c0ccaef5cee96c305ed79b661b32f#diff-8b374eb55e95fc34cb656ec3d7d479f263fb633a1232e7c99a35243617f5326fL386

the plan function now calls runnable.stream instead of runnable.invoke, so there isn't even a way to override this to set stream to false? 

Is there a way to workaround this so that I can use agents without streaming being force enabled? this is a simple example of code that is not able to run anymore unless the endpoint supports streaming, which is just the example from langchain openai tools agent:
```python
from langchain import hub
from langchain_community.chat_models import ChatOpenAI
from langchain.agents import AgentExecutor, create_openai_tools_agent

prompt = hub.pull("hwchase17/openai-tools-agent")
model = ChatOpenAI()
tools = ...

agent = create_openai_tools_agent(model, tools, prompt)
agent_executor = AgentExecutor(agent=agent, tools=tools)

agent_executor.invoke({"input": "hi"})
```

Please advise on how to run the agent without streaming after this change

PR #15793, opened 2024-01-10
mistralai[patch]: enforce stop tokens

efriis:


libs/partners/mistralai/langchain_mistralai/chat_models.py
```diff
@@ -89,16 +92,19 @@ def _create_retry_decorator(
 
 def _convert_mistral_chat_message_to_message(
     _message: MistralChatMessage,
+    stop: Optional[List[str]] = None,
 ) -> BaseMessage:
     role = _message.role
     if role == "user":
-        return HumanMessage(content=_message.content)
+        return HumanMessage(content=_enforce_stop_tokens(_message.content, stop))
```
eyurtsev, line 99: I don't think we should be dropping the stop tokens form the content.  The stop token cannot be recovered by the client if the stop token was a list of stop tokens. Recovery of the token is needed for parsing structured content.

If I understand correctly, there's no way to recover this information easily via `.invoke` and `.stream` etc

eyurtsev:
@efriis I don't suppose we have guidelines about how chat models should be have with respect to stop tokens?

efriis:
Not at the moment but given it's part of the base interface I feel like it makes sense to respect them. Happy to discuss

eyurtsev:
Should the stop token itself be dropped or only what follows it?

If we drop it we need to make sure parsing logic can be implemented with runnables (I think it cannot be?)

efriis:
Leave as warning for now. Maybe implement enforcement as an output parser / transform type thing.

PR #15805, opened 2024-01-10
raise value error added for directory loader

VpkPrasanna:
added value error for `DirectoryLoader` if the given path is a valid path but it does not contain any files .


eyurtsev (changes requested):
Thanks for the contribution! I understand that this might lead to some confusion for users if they enter the wrong directory, but the behavior is well defined -- it returns an empty list of documents.

I am not convinced that raising on an empty directory is good here since it forces calling code to do exception handling in some situations and makes it more difficult to implement correct behavior for the caller.

VpkPrasanna:
hi @eyurtsev  thanks for your suggestions , what's your thoughts if it was raising an warning instead of error ?

PR #15817, opened 2024-01-10
langchain[minor], community[minor], core[minor]: Async Cache support and AsyncRedisCache

dzmitry-kankalovich:
# Description
We're using `FastAPI` in our current project with async functions from LangChain. Recently we needed to add `Redis`-based cache. However we quickly learned that regardless of the LangChain way of usage - sync or async - the cache lookup/update/clear will happen always in synchronous manner. Naturally, this means blocking the event loop, especially with slower than Redis cache implementations.

So I took a similar approach to `agenerate` and `ainvoke` for LLMs, and created an `AsyncBaseCache` mixin interface with corresponding `alookup/aupdate/aclear` async variants:
```python
class AsyncBaseCache(ABC):
    """Base interface for async cache."""

    @abstractmethod
    async def alookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
        """Look up based on prompt and llm_string."""

    @abstractmethod
    async def aupdate(
        self, prompt: str, llm_string: str, return_val: RETURN_VAL_TYPE
    ) -> None:
        """Update cache based on prompt and llm_string."""

    @abstractmethod
    async def aclear(self, **kwargs: Any) -> None:
        """Clear cache that can take additional keyword arguments."""
```

The reference implementation is done for `InMemoryCache` and `RedisCache`. 

The `InMemoryCache` does not really need async variant, since there is no real IO, however having that this is the most convenient way to unit test changes.

The change spans throughout `core`, `community` and `langchain`. Linter, formatters, tests are passing.

# Issue
Fixes inability to use IO-bound Redis cache w/o blocking event loop. Opens opportunity for other cache implementations to have async variants.

# Dependencies
None

# Twitter
[@Mind_Clash](https://twitter.com/Mind_Clash)


libs/core/langchain_core/caches.py
```diff
@@ -22,3 +22,21 @@ def update(self, prompt: str, llm_string: str, return_val: RETURN_VAL_TYPE) -> N
     @abstractmethod
     def clear(self, **kwargs: Any) -> None:
         """Clear cache that can take additional keyword arguments."""
+
+
+class AsyncBaseCache(ABC):
+    """Base interface for async cache."""
+
+    @abstractmethod
+    async def alookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
```
cbornet, line 31: An alternative would be to add these methods to the `BaseCache` class with a default implementation that calls the non-async method with `run_in_executor` like what is done in `BaseLLM` and `Embeddings`.
Otherwise, there is no reason to prefix the methods here with the `a` prefix.

eyurtsev: @dzmitry-kankalovich We generally prefer keeping async version of methods in the same class as the sync versions, and have them provide a default async implementation that uses `run_in_executor` like @cbornet suggested.

Would you mind merging into existing abstractions? Ideally a single PR that just modifies the core interface first, and then separately we can do a PR for any implementations

eyurtsev (commented):
@dzmitry-kankalovich thanks for the contribution! LangChain uses the convention of having sync and async implementations defined in the same abstraction

dzmitry-kankalovich:
@cbornet @eyurtsev thanks, I will look into @cbornet idea and see if I can make it work

dzmitry-kankalovich:
@cbornet @eyurtsev I've pushed reworked interfaces - thanks for suggestion, it now feels much simpler.

I do however run into unit test problem with SQLite-based cache, which I cannot figure out yet. Somehow the base `aupdate` with default exec context does not work well for SQLite cache, although I've tested these changes with my production code and see no issues.

Anyway, I am going to figure out whats the problem and get back to you.

libs/core/langchain_core/caches.py
dzmitry-kankalovich: Done

libs/community/langchain_community/cache.py
```diff
@@ -350,20 +350,24 @@ def clear(self, **kwargs: Any) -> None:
...
 
 class RedisCache(BaseCache):
-    """Cache that uses Redis as a backend."""
+    """
+    Cache that uses Redis as a backend. Allows to use either sync or
+    async Redis client. Depending on the client passed, you are expected
+    to use either sync or async methods.
+    """
 
     def __init__(self, redis_: Any, *, ttl: Optional[int] = None):
```
cbornet, line 359: Maybe for this you could have a `RedisCache` that uses a `redis.Redis` client and an `AsyncRedisCache` that uses a `redis.asyncio.Redis` ? (I know this is the opposite move as for BaseCache 😄 )

dzmitry-kankalovich: I actually did it this way in the very first variant, but I didn't like it was literally 99% of the copy from the original RedisCache. The remaining 1% were async/await keywords and client check in constructor. Feels like a potential future maintenance problem?

dzmitry-kankalovich: Also, I am not sure what I suppose to do for `AsyncRedisCache` sync methods `lookup/update/clear`. Like I can raise error on those, or do the opposite trick for async in sync execution, but doesn't it feel like too much of duplication of effort?

cbornet: You could put this common behavior in a parent abstract class ?

dzmitry-kankalovich: Alright, I can do it. But before that, can you tell me whats your rationale here? Just so I know for future :)

dzmitry-kankalovich: @cbornet I mean I was asking whats the motivation for such split. 

I've pushed the implementation which is basically what you said, however I went a bit further with sync-in-async and actually made it so you can run it in certain context (weird case though, I log warning for that). I can rework to `NotImplementedError` if you think its cleaner.

libs/community/langchain_community/cache.py
```diff
@@ -375,68 +425,157 @@ def __init__(self, redis_: Any, *, ttl: Optional[int] = None):
...
-                        "error."
-                    )
-                    # In a previous life we stored the raw text directly
-                    # in the table, so assume it's in that format.
-                    generations.append(Generation(text=text))
-        return generations if generations else None
+        return self._get_generations(results)
+
+    async def alookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
+        logger.warning("Consider using `AsyncRedisCache` for async cache operations.")
```
cbornet, line 443: I'm not sure it's a good thing to output these logs. It will flood the logs.

dzmitry-kankalovich: agree.

libs/community/langchain_community/cache.py
```diff
@@ -375,68 +425,157 @@ def __init__(self, redis_: Any, *, ttl: Optional[int] = None):
...
+        self.redis = redis_
+        self.ttl = ttl
+
+    def lookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
+        """Look up based on prompt and llm_string."""
+        logger.warning(
+            "This an async Redis cache. Did you mean to use `alookup()` method?"
+        )
+        try:
+            if asyncio.get_running_loop():
```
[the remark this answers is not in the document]
dzmitry-kankalovich: agree.

dzmitry-kankalovich:
@cbornet adjusted code per your review.

libs/community/langchain_community/cache.py
```diff
@@ -375,68 +424,128 @@ def __init__(self, redis_: Any, *, ttl: Optional[int] = None):
...
-                        "older format. Please recreate your cache to avoid this "
-                        "error."
-                    )
-                    # In a previous life we stored the raw text directly
-                    # in the table, so assume it's in that format.
-                    generations.append(Generation(text=text))
-        return generations if generations else None
+        return self._get_generations(results)
+
+    async def alookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
```
cbornet, line 441: can be removed

libs/community/langchain_community/cache.py
```diff
@@ -375,68 +424,128 @@ def __init__(self, redis_: Any, *, ttl: Optional[int] = None):
...
+        return await super().alookup(prompt, llm_string)
```
cbornet, line 443:
```suggestion
```

libs/community/langchain_community/cache.py
```diff
@@ -375,68 +424,128 @@ def __init__(self, redis_: Any, *, ttl: Optional[int] = None):
...
-                    for idx, generation in enumerate(return_val)
-                },
-            )
-            if self.ttl is not None:
-                pipe.expire(key, self.ttl)
-
+            self._configure_pipeline_for_update(key, pipe, return_val, self.ttl)
             pipe.execute()
 
+    async def aupdate(
```
cbornet, line 453: can be removed

libs/community/langchain_community/cache.py
```diff
@@ -375,68 +424,128 @@ def __init__(self, redis_: Any, *, ttl: Optional[int] = None):
...
+        self, prompt: str, llm_string: str, return_val: RETURN_VAL_TYPE
+    ) -> None:
+        return await super().aupdate(prompt, llm_string, return_val)
+
     def clear(self, **kwargs: Any) -> None:
         """Clear cache. If `asynchronous` is True, flush asynchronously."""
         asynchronous = kwargs.get("asynchronous", False)
         self.redis.flushdb(asynchronous=asynchronous, **kwargs)
 
+    async def aclear(self, **kwargs: Any) -> None:
```
cbornet, line 463: can be removed

libs/community/langchain_community/cache.py
dzmitry-kankalovich: yes indeed. I've pushed removal.

libs/community/langchain_community/cache.py
dzmitry-kankalovich: done.

libs/community/langchain_community/cache.py
dzmitry-kankalovich: done.

libs/langchain/tests/integration_tests/cache/test_redis_cache.py
```diff
@@ -34,25 +34,88 @@ def test_redis_cache_ttl() -> None:
...
         llm_output={},
     )
     assert output == expected_output
+    llm_cache.redis.flushall()
+
+
+async def test_sync_in_async_redis_cache() -> None:
+    import redis
+
+    set_llm_cache(RedisCache(redis_=redis.Redis.from_url(REDIS_TEST_URL)))
```
cbornet, line 71: Should be `AsyncRedisCache` ?

libs/langchain/tests/integration_tests/cache/test_redis_cache.py
```diff
@@ -34,25 +34,88 @@ def test_redis_cache_ttl() -> None:
...
+    llm = FakeLLM()
+    params = llm.dict()
+    params["stop"] = None
+    llm_string = str(sorted([(k, v) for k, v in params.items()]))
     llm_cache = cast(RedisCache, get_llm_cache())
```
cbornet, line 76: Should be AsyncRedisCache ?

libs/langchain/tests/unit_tests/test_cache.py
```diff
@@ -35,33 +40,41 @@ def set_cache_and_teardown(request: FixtureRequest) -> Generator[None, None, Non
...
+async def test_llm_caching() -> None:
     prompt = "How are you?"
     response = "Test response"
     cached_response = "Cached test response"
     llm = FakeListLLM(responses=[response])
-    if get_llm_cache():
-        get_llm_cache().update(
+    if llm_cache := get_llm_cache():
+        # sync test
+        llm_cache.update(
```
cbornet, line 65: Not very recommended to call a blocking method in the event loop but I guess for a test it's ok ?

cbornet (approved):

cbornet:
LGTM, great work !

dzmitry-kankalovich:
@eyurtsev I think I am done as far as the changes go. If you have some time - let me know how we can proceed forward. Thanks!

libs/langchain/tests/integration_tests/cache/test_redis_cache.py
```diff
@@ -34,25 +34,88 @@ def test_redis_cache_ttl() -> None:
     assert llm_cache.redis.pttl(key) > 0
 
 
-def test_redis_cache() -> None:
+async def test_async_redis_cache_ttl() -> None:
```
eyurtsev, line 37: Thanks for adding these

libs/community/langchain_community/cache.py
eyurtsev: > or do the opposite trick for async in sync execution
 
At the moment, we generally try to avoid that. Only one event loop is allowed in a given thread, so there's some additional logic that's required to determine if there's an event loop running already and if yes, kick off another thread etc

libs/core/langchain_core/caches.py
eyurtsev: Generally better to split PRs by package to minimize potential dependency conflicts since the packages may have different release schedules

eyurtsev (approved):
Looks good. Leaving one question regarding alternative design.

libs/community/langchain_community/cache.py
```diff
@@ -349,21 +350,69 @@ def clear(self, **kwargs: Any) -> None:
         self.redis.flushdb(flush_type=asynchronous)
 
 
-class RedisCache(BaseCache):
-    """Cache that uses Redis as a backend."""
+class RedisCacheBase(BaseCache, ABC):
```
eyurtsev, line 353: Could we mark this as private?

libs/community/langchain_community/cache.py
```diff
@@ -349,21 +350,69 @@ def clear(self, **kwargs: Any) -> None:
...
+            key,
+            mapping={
+                str(idx): dumps(generation) for idx, generation in enumerate(return_val)
+            },
+        )
+        if ttl is not None:
+            pipe.expire(key, ttl)
+
+
+class RedisCache(RedisCacheBase):
```
eyurtsev, line 399: An alternative design is to make the initializer able to accept both async and sync versions of the redis client.

If its initialized with the async redis client, then it uses its async methods. If it gets initializes with the sync client, it delegates all the async calls to the sync ones (using the trick in the abstract class).

What do you think would be better? A single `RedisCache` for users to know about / user, and have the differentiation in terms of the underlying redis client that's passed to initialize the cache?

eyurtsev:
@baskaryan FYI -- change to the base schema, so we'll need to bump minimal dependency in langchain. Do you need me to break these into two separate PRs or are we okay merging, releasing core, bumping constraints in langchain, and then releasing langchain?

baskaryan:
> @baskaryan FYI -- change to the base schema, so we'll need to bump minimal dependency in langchain. Do you need me to break these into two separate PRs or are we okay merging, releasing core, bumping constraints in langchain, and then releasing langchain?

can land in one 👍

libs/community/langchain_community/cache.py
dzmitry-kankalovich: @eyurtsev it was actually my original implementation, a single Cache class with ability to work with sync and async client. However @cbornet explicitly asked to change - break it out to two separate Redis Caches. I can see pros and cons in either approach. 

I think at this point, having two different opinions which both result in a fine working implementation means we can go either way. The maintenance burden of such split seems negligible.

libs/community/langchain_community/cache.py
dzmitry-kankalovich: Sure, can you show me the example? Something like `__RedisCacheBase` ?

dzmitry-kankalovich: Looked at other examples in codebase and seems like the pattern is to use a single underscore, so I pushed `_RedisCacheBase` rename.

libs/community/langchain_community/cache.py
dzmitry-kankalovich: I think even if such split turns out to be a mistake, we should be able to work around it in future by consolidating features in a single class and then using type alias for `AsyncRedisCache` to provide backwards compatibility with existing codebase.

But I think its just fine as it is.

dzmitry-kankalovich:
@eyurtsev @baskaryan: It's all ready on my end, I don't plan to add anything else, it's good to go.

dzmitry-kankalovich:
@hwchase17 @baskaryan hey guys, was this accidentally closed?

baskaryan:

eyurtsev:
Taking a look now to see if I can refactor into a single class

docker/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+# docker-compose to make it easier to spin up integration tests.
```
eyurtsev, line 1: @dzmitry-kankalovich  I added a docker compose file under /docker we might move it at some point, but thinking about making it easier for developers to spin up services that integration tests depend on

eyurtsev:
Decided to revert for now. Will merge as is!

The implementation with one class is more compact (https://github.com/langchain-ai/langchain/pull/15817/files#diff-60f65acc09d072aa20f3bad2cce0ecaee7e4ded197a56cd54d6671d76a684de0R401), but a bit harder to understand, and dealing with mypy is seems pretty annoying in that case.

I fixed a few typing issues in the original PR (missing type annotations).

libs/langchain/langchain/cache.py
baskaryan, line 1: note: shouldn't import this in langchain, imports here are only for backwards compatibility

libs/community/langchain_community/cache.py
cbornet: OK. I get the arguments for the single class. Sorry for the pull in the wrong direction.

docker/docker-compose.yml
dzmitry-kankalovich: yeah that would be super useful!

libs/langchain/langchain/cache.py
dzmitry-kankalovich: I see. Noted! Maybe makes sense to drop a comment in that file next time you are working with it - future contributors will know.

PR #15829, opened 2024-01-10
community: Handle HuggingFaceInferenceAPI failure response of POST request

ksraunak77:
- **Description:** 
 Currently the embed_documents() function of HuggingFaceInferenceAPIEmbeddings class posts a request to HuggingFaceInference API to generate text embeddings. The POST request may not always return successful response due to various technical glitches. This PR handles the event when the status code of the response returned by HuggingFaceInference API is not 200 and raises an exception for this event.

  - **Dependencies:** None
  - **Twitter handle:** None

### DIFF
--- a/libs/community/langchain_community/embeddings/huggingface.py
+++ b/libs/community/langchain_community/embeddings/huggingface.py
@@ -329,6 +329,11 @@ class HuggingFaceInferenceAPIEmbeddings(BaseModel, Embeddings):
                 "options": {"wait_for_model": True, "use_cache": True},
             },
         )
+        if response.status_code != 200:
+            raise Exception(
+                f"HuggingFaceInferenceAPI returned an unexpected response with status "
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -329,6 +329,11 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
                 "options": {"wait_for_model": True, "use_cache": True},
             },
         )
+        if response.status_code != 200:
```
eyurtsev, line 332: The exception should not be a generic exception, so one can do retries on 5xx errors, which this code is more likely to raise since it might be involved in repeated API calls

It might make sense to do something like `response.raise_for_json()` so caller code can retrieve the status code and apply re-try logic appropriately

PR #15844, opened 2024-01-10
Rschwabco/pinecone client support

rschwabco:
---
title: "community: Fix use_pod_based parameter in Pinecone class"
---

- Description: This PR adds an `use_pods_based` parameter which is optional and `True` by default. This is to support upcoming changes to the Pinecone client.

### DIFF
--- a/libs/community/langchain_community/vectorstores/pinecone.py
+++ b/libs/community/langchain_community/vectorstores/pinecone.py
@@ -4,14 +4,18 @@ import logging
 import os
 import uuid
 import warnings
-from typing import (TYPE_CHECKING, Any, Callable, Iterable, List, Optional, Tuple, Union)
+from typing import TYPE_CHECKING, Any, Callable, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
-from langchain_community.vectorstores.utils import (DistanceStrategy, maximal_marginal_relevance)
 from langchain_core.documents import Document
 from langchain_core.embeddings import Embeddings
 from langchain_core.utils.iter import batch_iterate
 from langchain_core.vectorstores import VectorStore
+
+from langchain_community.vectorstores.utils import (
+    DistanceStrategy,
+    maximal_marginal_relevance,
+)
 
 if TYPE_CHECKING:
     from pinecone import Index
@@ -338,7 +342,7 @@ class Pinecone(VectorStore):
     def get_pinecone_index(
         self,
         index_name: Optional[str],
-        pool_threads: int = 4,        
+        pool_threads: int = 4,
     ) -> Index:
         """Return a Pinecone Index instance.
 
@@ -355,7 +359,7 @@ class Pinecone(VectorStore):
                 "Could not import pinecone python package. "
                 "Please install it with `pip install pinecone-client`."
             )
-                        
+
         if self.use_pod_based == False:
             pinecone_instance = pinecone.Pinecone(
                 api_key=os.environ.get("PINECONE_API_KEY"), pool_threads=pool_threads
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/pinecone.py
```diff
@@ -60,6 +50,7 @@ def __init__(
         text_key: str,
         namespace: Optional[str] = None,
         distance_strategy: Optional[DistanceStrategy] = DistanceStrategy.COSINE,
+        use_pod_based: Optional[bool] = True,
```
efriis, line 53: Does `use_pod_based=None` represent anything, or can I change this?
```suggestion
        use_pod_based: bool = True,
```

libs/community/langchain_community/vectorstores/pinecone.py
```diff
@@ -369,9 +360,7 @@ def get_pinecone_index(
                 "Please install it with `pip install pinecone-client`."
             )
 
-        pinecone_client_version = pinecone.__version__
-
-        if version.parse(pinecone_client_version) >= version.parse("3.0.0.dev"):
+        if self.use_pod_based is False:
```
efriis, line 363: What does use_pod_based=None represent here

libs/community/langchain_community/vectorstores/pinecone.py
```diff
@@ -413,6 +402,7 @@ def from_texts(
         upsert_kwargs: Optional[dict] = None,
         pool_threads: int = 4,
         embeddings_chunk_size: int = 1000,
+        use_pod_based: Optional[bool] = True,
```
efriis, line 405: same

libs/community/langchain_community/vectorstores/pinecone.py
efriis: Also is `use_pod_based` what we want to call this forever, or would it be better to use something like `index_type: Union[Literal['pod'], Literal['serverless']]` = 'pod', with the ability to add other index types in the future

libs/community/langchain_community/vectorstores/pinecone.py
```diff
@@ -347,9 +339,8 @@ def max_marginal_relevance_search(
             embedding, k, fetch_k, lambda_mult, filter, namespace
         )
 
-    @classmethod
```
efriis, line 350: this is a breaking change for anyone using this as documented here. Is there a way we can keep this as a classmethod? https://api.python.langchain.com/en/latest/vectorstores/langchain_community.vectorstores.pinecone.Pinecone.html#langchain_community.vectorstores.pinecone.Pinecone.get_pinecone_index

libs/community/langchain_community/vectorstores/pinecone.py
```diff
@@ -463,10 +456,16 @@ def from_existing_index(
...
+        use_pod_based: Optional[bool] = True,
     ) -> Pinecone:
         """Load pinecone vectorstore from index name."""
-        pinecone_index = cls.get_pinecone_index(index_name, pool_threads)
-        return cls(pinecone_index, embedding, text_key, namespace)
+        pinecone = cls(
+            None, embedding, text_key, namespace, use_pod_based=use_pod_based
+        )
+        pinecone_index = pinecone.get_pinecone_index(index_name, pool_threads)
+        pinecone._index = pinecone_index
```
efriis, line 466: don't love this, and comment above about keeping `get_pinecone_index` as a classmethod will likely address this.

libs/community/langchain_community/vectorstores/pinecone.py
```diff
@@ -369,9 +360,7 @@ def get_pinecone_index(
                 "Please install it with `pip install pinecone-client`."
             )
 
-        pinecone_client_version = pinecone.__version__
-
-        if version.parse(pinecone_client_version) >= version.parse("3.0.0.dev"):
```
efriis, line 374: will this work with old pinecone clients? If not, it's probably worth continuing to check version number because community packages don't offer guarantees about which package version is installed.

PR #15855, opened 2024-01-11
core: make BaseTool._parse_input return all parameters

liunux4odoo:
## Why

the `tool` decorator is convenient to specify parameter infomations, but the default values are not parsed by default:

```
from langchain.agents import tool
from langchain.pydantic_v1 import Field

@tool
def text2images(
    prompt: str,
    n: int = Field(1, description="image count"),
):
    '''generate images'''
    ...

text2images("draw a house") 

# will fail because BaseTool._parse_input returns {"prompt": "draw a house"}, then the tool is called with parameters {"prompt": "draw a house", "n": FieldInfo(...)}.
# it should return {"prompt": "draw a house", "n": 1}

```


liunux4odoo:
> @liunux4odoo thanks for this! could we add some unit tests?

I have modify `test_tools.py` to test the default value defined by `Field`.

liunux4odoo:
@baskaryan
I have make some updates to this pr:
- fix lint error
- [fix BaseTool._to_args_and_kwargs:](https://github.com/langchain-ai/langchain/pull/15855/commits/a166395e021cc370c8e1c0b079563b335cbe74f5) 
[a166395](https://github.com/langchain-ai/langchain/pull/15855/commits/a166395e021cc370c8e1c0b079563b335cbe74f5)
tool defined with `*args` parameters, the args_schema has a field named `args`. it should be expanded to actual *args
e.g.: test_tools.test_named_tool_decorator_return_direct.search_api

liunux4odoo:
I will close this pr since using Annotated is a better alternative.

Issue #15859, opened 2024-01-11
Google Trend utility makes assumptions on keys from response


lpezet:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

The following code:
```
import os
from langchain_community.utilities.google_trends import GoogleTrendsAPIWrapper

os.environ['SERPAPI_API_KEY'] = ''

tool = GoogleTrendsAPIWrapper()
tool.run("Something that will yield an error like totally")
```
will yield this:
```
Traceback (most recent call last):
  File "/home/ubuntu/Work/luc/langchain-google-trends-issue-1/langchain_google_trends_issue_1/main.py", line 9, in <module>
    tool.run("Something that will yield an error like totally")
  File "/home/ubuntu/Work/luc/langchain-google-trends-issue-1/.venv/lib/python3.10/site-packages/langchain_community/utilities/google_trends.py", line 68, in run
    total_results = client.get_dict()["interest_over_time"]["timeline_data"]
KeyError: 'interest_over_time'
```

### Description

* I'm trying to use the Google Trends tool with some AI agent. 
* Now the (not so smart) agent ran a query with Google Trend that did NOT return what its implementation expected.
* I would expect the implementation to follow a more foolproof logic.

### System Info

# pyproject.toml
python = "^3.10"
langchain = "0.0.354"
pytest = "^7.4.4"
google-search-results = "^2.4.2"

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [X] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

lpezet:
@dosu-bot This is definitely one way to go about it, and it will work. I, personally, would like to fix it at the source (root cause).
Specifically:
https://github.com/langchain-ai/langchain/blob/v0.1.0/libs/community/langchain_community/utilities/google_trends.py#L68

Something like this patch:
```
index f0f15000c..c94dee6be 100644
--- a/libs/community/langchain_community/utilities/google_trends.py
+++ b/libs/community/langchain_community/utilities/google_trends.py
[...]

PR #15864, opened 2024-01-11
community: Handling missing key in Google Trends API response.

lpezet:
- **Description:** Handing response where _interest_over_time_ is missing.
  - **Issue:** #15859
  - **Dependencies:** None

### DIFF
--- a/libs/community/langchain_community/utilities/google_trends.py
+++ b/libs/community/langchain_community/utilities/google_trends.py
@@ -65,7 +65,8 @@ class GoogleTrendsAPIWrapper(BaseModel):
 
         total_results = []
         client = self.serp_search_engine(params)
-        total_results = client.get_dict()["interest_over_time"]["timeline_data"]
+        client_dict = client.get_dict()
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/utilities/google_trends.py
```diff
@@ -65,7 +65,12 @@ def run(self, query: str) -> str:
 
         total_results = []
         client = self.serp_search_engine(params)
-        total_results = client.get_dict()["interest_over_time"]["timeline_data"]
+        client_dict = client.get_dict()
```
hwchase17, line 68: why would this occur? is this key just not included sometimes? can we add a comment about why this check is neccesary and in which scenarios its triggered

lpezet: that key is not included sometimes indeed.
in the test I submitted, I pretty much used the response I got from Google Trends that made the code fail.
The Agent used the Google Trends tool and passed a query that lead to this basically.

hwchase17 (approved):

PR #15873, opened 2024-01-11
docs: Correct a typo on Quickstart page

leonvanzyl:
The ChatOpenAI class does not have a parameter called **openai_api_key**.  Instead, it should be renamed to **api_key** only.


hwchase17:
it accepts both (api_key is an alias)

PR #15881, opened 2024-01-11
community: Ollama - Pass headers to post request

shahrin014:
## Feature
- Set additional headers in constructor
- Headers will be sent in post request

This feature is useful if deploying Ollama on a cloud service such as hugging face, which requires authentication tokens to be passed in the request header.

## Tests
- Test if header is passed
- Test if header is not passed


hwchase17 (approved):
thanks!

PR #15885, opened 2024-01-11
core[patch]: Update doc-string in base callback managers

eyurtsev:
Update doc-strings with a comment about on_llm_start vs. on_chat_model_start.

### DIFF
--- a/libs/core/langchain_core/callbacks/base.py
+++ b/libs/core/langchain_core/callbacks/base.py
@@ -166,7 +166,12 @@ class CallbackManagerMixin:
         metadata: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
     ) -> Any:
-        """Run when LLM starts running."""
+        """Run when LLM starts running.
+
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/callbacks/base.py
```diff
@@ -179,7 +184,11 @@ def on_chat_model_start(
         metadata: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
     ) -> Any:
-        """Run when a chat model starts running."""
+        """Run when a chat model starts running.
+
+        **ATTENTION**: This method is called for chat models. If you're implementing
+            a handler for a non-chat model, you should use on_llm_start instead.
+        """
         raise NotImplementedError(
```
eyurtsev, line 192: Why do we throw a NotImplementedError here instead of doing a pass. Was this accidentally left here when we made the methods non-required? Or is it to raise awareness that it's different from on_llm_start?

baskaryan (approved):

PR #15890, opened 2024-01-11
experimental: New Technique for Unique Chunk Retrivals

jankomisarczyk:
**Description:** 
Please find the attached pdf, where everything is detailed described :D
[langchain_core.pdf](https://github.com/langchain-ai/langchain/files/13905129/langchain_core.pdf)
  
**Dependencies:**
In order to solve Generalized assignment problem, we need to add 'mip' solver.

**Twitter handle:** no need ;)
 


hwchase17:
super interesting. one question and one suggestion:

1. "More often than not, when we want to find relevant chunks of text for similarly worded questions but with different meanings, traditional keyword algorithms returns almost the same lists of chunks." - when is a use case where this would happen?

2. Can we move this to experimental? While i think this is super interesting, we want to keep things in core to things that are more battle tested.

jankomisarczyk:
Hi Chase,
1. I came up with this solution, when I chatted with a document and the answers spitted out were very similar. Even though I asked different questions each time, they were too general and repetitive. **BUT** I knew that the answer to my question was somewhere in the vector store (I vectorized a document that I read before) and still could not get the right answer.

The problem is that whenever you want to explore a topic in-depth, you will not get answers to your 'very specific' questions or at least the answers will not be very detailed. SO I guess when you're asking for context-chunks to more complex questions that are about one same topic, the semantic search cannot give you more specific answers because the returned list of context docs is almost the same each time.

I recall even a statistic that when I chatted about a very specific topic (a 20-ish page PDF about leveraged financial instruments), all lists of retrived chunks for multiple questions had around 60% overlap. To be honest, I do not exactly remember if it was more like 50% or 70% overlap.

Use Cases:
- imagine chatbot for Q&A with pdf-s (in vector store)
user asks questions and if question requires more context, then every query to the vector store, we save it in some history of previously asked queries, so that a new incoming question will get more unique context-chunks.

Just to clarify, this idea helps to retrieve Unique chunks for each guery in a list:
image

- another use-case is that: you want to re-write a document and inspire yourself with previous, historical documents, but since you will not retrieve all possible chunks from vector store, your re-written document will be very **general and repetitive**


2. Yes sure thing, should I close this PR and open a new one in langchain-experimental?

hwchase17:
> Hi Chase,
> 
> 1. I came up with this solution, when I chatted with a document and the answers spitted out were very similar. Even though I asked different questions each time, they were too general and repetitive. **BUT** I knew that the answer to my question was somewhere in the vector store (I vectorized a document that I read before) and still could not get the right answer.
> 
> The problem is that whenever you want to explore a topic in-depth, you will not get answers to your 'very specific' questions or at least the answers will not be very detailed. SO I guess when you're asking for context-chunks to more complex questions that are about one same topic, the semantic search cannot give you more specific answers because the returned list of context docs is almost the same each time.
> 
> I recall even a statistic that when I chatted about a very specific topic (a 20-ish page PDF about leveraged financial instruments), all lists of retrived chunks for multiple questions had around 60% overlap. To be honest, I do not exactly remember if it was more like 50% or 70% overlap.
> 
> Use Cases:
> 
> * imagine chatbot for Q&A with pdf-s (in vector store)
>   user asks questions and if question requires more context, then for every query to the vector store, we save it in some history of previously asked queries, so that a new incoming question will get more unique context-chunks.
> 
> Just to clarify, this idea helps to retrieve Unique chunks for each query in a list: image
> 
> * another use-case is that: you want to re-write a document and inspire yourself with previous, historical documents, but since you will not retrieve all possible chunks from vector store, your re-written document will be very **general and repetitive**
> 
> 2. Yes sure thing, should I close this PR and open a new one in langchain-experimental?

or adapt this pr, either is fine

thanks!

jankomisarczyk:
Hi Bagatur @baskaryan,
I moved the PR from core to experimental, I do not know what to do with it now.

I have only updated pyproject.toml but not poetry.lock (because I do not know what are your recommended groups & extras).

Thanks,
Jan

baskaryan (commented):
can we make this wrap a vector store (take it as init param) instead of inheriting form it? can we also give it a more informative name?

could we add some docs showing how it works?

jankomisarczyk:
Hi @baskaryan,

I have also a few problems that need to be addressed:

1. Some mechanismus to optimize the number of retrieved chunks, currently it is "5 x number_of_chunks to be retrieved for each query" but I guess you can do better by some analysis of the density of vectorized Space

2. How to make sure that Document-s in vectorstore have UniqueID in their metadata? Or should I pre-process the vector store e.g. in __init__ method and append unique identifier to each Document that exists in vectorestore? 

But then also how to detect changes to vector store? For instance, after initialization, new chunks were added to the VS without a UID, so I need to implement some kind of counter and a sanity check before starting the optimization procedure.

3. Do you recommend any free Embedding models? openai gettin' expensive

Bests,
Jan
I should have a bit of time this week to work on this PR ;)

efriis:

PR #15891, opened 2024-01-11
core[patch]: Add missing async keyword for async method

eyurtsev:
Missing async keyword method


sweetkane:
Looks like the same issue exists in llms.py, line 508.

nfcampos:
This is actually correct

PR #15904, opened 2024-01-11
feat: add additional support for `BigQueryVectorSearch`

ashleyxuu:
BigQuery vector search lets you use GoogleSQL to do semantic search, using vector indexes for fast but approximate results, or using brute force for exact results.

This PR:
1. Add `metadata[_job_ib]` in Document returned by any similarity search
2. Add `explore_job_stats` to enable users to explore job statistics and better the debuggability
3. Set the minimum row limit for running create vector index.

### DIFF
--- a/libs/community/langchain_community/vectorstores/bigquery_vector_search.py
+++ b/libs/community/langchain_community/vectorstores/bigquery_vector_search.py
@@ -28,7 +28,7 @@ DEFAULT_METADATA_COLUMN_NAME = "metadata"  # document metadata
 DEFAULT_CONTENT_COLUMN_NAME = "content"  # text content, do not rename
 DEFAULT_TOP_K = 4  # default number of documents returned from similarity search
 
-_MIN_INDEX_ROWS = 5000 # minimal number of rows for creating an index
+_MIN_INDEX_ROWS = 5000  # minimal number of rows for creating an index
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/vectorstores/bigquery_vector_search.py
```diff
@@ -192,6 +193,11 @@ def _initialize_vector_index(self) -> Any:
         if self._have_index or self._creating_index:
             # Already have an index or in the process of creating one.
             return
+        table = self.bq_client.get_table(self.vectors_table)
+        if (table.num_rows or 0) < _MIN_INDEX_ROWS:
+            # Not enough rows to create index.
+            self._logger.debug("Not enough rows to create a vector index.")
+            return
```
lkuligin, line 200: should we raise an Exception maybe?

ashleyxuu: We don't need to raise an exception here, since: 
1. We'll only create the VECTOR INDEX if the row is > `_MIN_INDEX_ROWS`
2. We don't use the index when the number or rows is small, but vector search still returns correct results (without using the index) -> brute force results.

lkuligin: I see, thx.

lkuligin:
SGTM

@hwchase17 do you mind taking a look, please?

baskaryan (approved):

PR #15922, opened 2024-01-12
<langchain_community.vectorstores>:<Fix pinecone.py __init__ docsrting instruction>

yacine555:
- **Description:** The pinecone docstring instructs to pass the embedding query text causing the warning below. It should be the embeddings object.
 warning message: UserWarning: Passing in `embedding` as a Callable is deprecated. Please pass in an Embeddings object instead.
  - **Issue:** NA
  - **Dependencies:** None

@baskaryan


hwchase17 (approved):
thank you!

PR #15926, opened 2024-01-12
stop making copies of inputs

hwchase17:


hinthornw:
Would then make the tool typehints invalid for non-primitive values

hwchase17:
> Would then make the tool typehints invalid for non-primitive values

@hinthornw what do you mean?

hinthornw (approved):

libs/core/langchain_core/tools.py
```diff
@@ -248,7 +248,11 @@ def _parse_input(
         else:
             if input_args is not None:
                 result = input_args.parse_obj(tool_input)
-                return {k: v for k, v in result.dict().items() if k in tool_input}
+                return {
+                    k: getattr(result, k)
+                    for k, v in result.dict().items()
```
hinthornw, line 253: should we just go through the fields then?

Issue #15931, opened 2024-01-12
Adding more capabilities to Langchain Shell Tool


krishnashed:
### Feature request

The tool I'm talking about is:
from langchain.tools import ShellTool

As Shell tool is very versatile, while integrating it into applications as openai functions, developers have no clue about what command is being executed using the ShellTool. All one can see is:
![image](https://github.com/langchain-ai/langchain/assets/60742358/540e274a-debc-4564-9027-046b91424df3)

Which shouldn't be the case right ? Developers need to know what command is going to be executed by the ShellTool, and whether its the right command to execute, or you just prompt better for the LLM to generate a new command and let that execute. 

Summarising my feature request:
1. There's no visibility about what command was executed.
2. There's no mechanism to prevent a command to be executed using ShellTool, like a y/n human input which can be accepted from user to proceed with executing the command.

I would like to implement this in a PR if the project maintainer feels this feature adds value.

### Motivation

I was using ShellTool as a OpenAI Function in Autogen, which is a multi-agent conversation framework. Where in one of my AI Agent is a programmer, and it has access to the ShellTool. When its executing any command  using ShellTool, we have no clue about what command is being executed, and are thrown to mercy of the LLM that the generated command is what we need. 

### Your contribution

I have raised a [PR](https://github.com/langchain-ai/langchain/pull/15932), would like you to review and provide feedback on it. Or if there's some other way around to implement the feature I'm proposing, please let me know.

krishnashed:
@dosu-bot can you tell me how to build the langchain library locally and test it out?

krishnashed:
@hwchase17 can you please guide me, How can i run the langchain library build in my local ? To test the feature I have added.

I tried running `poetry build` But I get this error
![image](https://github.com/langchain-ai/langchain/assets/60742358/5fe815de-48e6-4dbe-b103-7f66c53b11ee)

I tried running `pip install -e .` But even that didn't work. I'm stuck

krishnashed:
Closing this issue as the feature I requested was successfully merged into langchain library.

PR #15932, opened 2024-01-12
Community: Basic Logging and Human input to ShellTool

krishnashed:
- **Description:** As Shell tool is very versatile, while integrating it into applications as openai functions, developers have no clue about what command is being executed using the ShellTool. All one can see is:
image

Summarising my feature request:
1. There's no visibility about what command was executed.
2. There's no mechanism to prevent a command to be executed using ShellTool, like a y/n human input which can be accepted from user to proceed with executing the command., 
  - **Issue:** the issue #15931 it fixes if applicable,
  - **Dependencies:** There isn't any dependancy,
  - **Twitter handle:** @krishnashed


hwchase17 (commented):
printing seems fine (would prefer logging)

for asking for human input, id prefer NOT an argument to _run - it doesnt seem like the LLM should decide, but seems like should be a param on the class so the user can set

krishnashed:
Thanks @hwchase17 for reviewing the PR. I'll implement as instructed, and get back!

hwchase17 (approved):

krishnashed:
@hwchase17 I still need to write test cases, then you can merge the PR

krishnashed:
@hwchase17 I've completed the feature and also written tests for it, could you please check and merge the PR ?

baskaryan (approved):

krishnashed:
@baskaryan The linting issues are fixed, you could please merge the PR ?

Issue #15664, opened 2024-01-07
For New Contributors: Update Integration Documentation


hwchase17:
### Issue you'd like to raise.

Hello LangChain community,

We're always happy to see more folks getting involved in contributing to the LangChain codebase.

This is a good first issue if you want to learn more about how to set up
for development in the LangChain codebase.

## Goal

Your contribution will make it easier for users to use integrations with the newest LangChain syntax

## Context

As you may have noticed, we’ve recently gone to LangChain 0.1. As part of this, we want to update integration pages to be consistent with new methods. These largely include: (a) new methods for invoking integrations and chains (`invoke`, `stream`), (b) new methods for creating chains (LCEL, `create_xyz_..`).

There are a lot of integrations, so we’d love community help! This is a great way to get started contributing to the library as it will make you familiar with best practices and various integrations.

## Set up for development

There are lots of integration notebooks in https://github.com/langchain-ai/langchain/tree/master/docs/docs/integrations. After making changes there, you should run `make format` from the root LangChain directory to run our formatter.

## Shall you accept

Shall you accept this challenge, please claim one (and only one) of the modules from the list
below as one that you will be working on, and respond to this issue.

Once you've made the required code changes, open a PR and link to this issue.

## Acceptance Criteria

- Uses new methods for calling chains (`invoke`, `stream`, etc)
- Uses LCEL where appropriate
- Follows the format outlined below

## Examples

We've gotten started with some examples to show how we imagine these integration pages should look like. The exact format may look different for each type of integration, so make sure to look at the type you are working on:

- LLMs:
    - https://python.langchain.com/docs/integrations/llms/cohere
- Chat Models:
    - https://python.langchain.com/docs/integrations/chat/cohere
- Vectorstores:
    - https://python.langchain.com/docs/integrations/vectorstores/faiss
- Retrievers:
    - https://python.langchain.com/docs/integrations/retrievers/tavily
    - https://python.langchain.com/docs/integrations/retrievers/ragatouille
- Tools:
    - https://python.langchain.com/docs/integrations/tools/tavily_search
- Toolkits:
    - https://python.langchain.com/docs/integrations/toolkits/gmail
- Memory:
    - https://python.langchain.com/docs/integrations/memory/sql_chat_message_history

## Your contribution

Please sign up by responding to this issue and including the name of the module.

sweetkane:
I can take the Memory module

DaveDeCaprio:
[...]

PR #15940, opened 2024-01-12
docs: update baidu_qianfan_endpoint.ipynb doc 

Bigtable123:
- **Description:** Updated the docs for the chat integration module baidu_qianfan_endpoint.ipynb
  - **Issue:**  #15664 
  - **Dependencies:**N/A


Bigtable123:
The stream mode of the baichuan model has a bug where it automatically adds the sentence_id when it returns, which may require modifying the source code of this model, I used a try-catch block to skip this Type error and it works fine.

Bigtable123:
For the other model, baichuan, it is more problematic. Baichuan has updated their API and its usage rules, now they don't use the secret key anymore and the API base and usage rules have completely changed, the baichuan model code [libs/community/langchain_community/chat_models/baichuan.py](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/chat_models/baichuan.py) in the langchain community is completely outdated. It is not possible to update the doc for the baichuan model at this monment.

hwchase17 (commented):
there is a stray file checked in. otherwise, lgtm

Bigtable123:
I've removed that file.

Bigtable123:
I uploaded some changes by mistake, I have reverted it.

baskaryan (approved):

PR #15966, opened 2024-01-12
community: Added new Utility runnables for NVIDIA Riva.

rmkraus:
**Please tag this issue with `nvidia_genai`**

- **Description:** Added new Runnables for integration NVIDIA Riva into LCEL chains for Automatic Speech Recognition (ASR) and Text To Speech (TTS).
- **Issue:** N/A
- **Dependencies:** To use these runnables, the NVIDIA Riva client libraries are required. It they are not installed, an error will be raised instructing how to install them. The Runnables can be safely imported without the riva client libraries.
- **Twitter handle:** N/A

All of the Riva Runnables are inside a single folder in the Utilities module. In this folder are four files:
- common.py - Contains all code that is common to both TTS and ASR
- stream.py - Contains a class representing an audio stream that allows the end user to put data into the stream like a queue.
- asr.py - Contains the RivaASR runnable
- tts.py - Contains the RivaTTS runnable

The following Python function is an example of creating a chain that makes use of both of these Runnables:

```python
def create(
    config: Configuration,
    audio_encoding: RivaAudioEncoding,
    sample_rate: int,
    audio_channels: int = 1,
) -> Runnable[ASRInputType, TTSOutputType]:
    """Create a new instance of the chain."""
    _LOGGER.info("Instantiating the chain.")

    # create the riva asr client
    riva_asr = RivaASR(
        url=str(config.riva_asr.service.url),
        ssl_cert=config.riva_asr.service.ssl_cert,
        encoding=audio_encoding,
        audio_channel_count=audio_channels,
        sample_rate_hertz=sample_rate,
        profanity_filter=config.riva_asr.profanity_filter,
        enable_automatic_punctuation=config.riva_asr.enable_automatic_punctuation,
        language_code=config.riva_asr.language_code,
    )

    # create the prompt template
    prompt = PromptTemplate.from_template("{user_input}")

    # model = ChatOpenAI()
    model = ChatNVIDIA(model="mixtral_8x7b")  # type: ignore

    # create the riva tts client
    riva_tts = RivaTTS(
        url=str(config.riva_asr.service.url),
        ssl_cert=config.riva_asr.service.ssl_cert,
        output_directory=config.riva_tts.output_directory,
        language_code=config.riva_tts.language_code,
        voice_name=config.riva_tts.voice_name,
    )

    # construct and return the chain
    return {"user_input": riva_asr} | prompt | model | riva_tts  # type: ignore
```

The following code is an example of creating a new audio stream for Riva:

```python
input_stream = AudioStream(maxsize=1000)
# Send bytes into the stream
for chunk in audio_chunks:
    await input_stream.aput(chunk)
input_stream.close()
```

The following code is an example of how to execute the chain with RivaASR and RivaTTS

```python
output_stream = asyncio.Queue()
while not input_stream.complete:
    async for chunk in chain.astream(input_stream):
        output_stream.put(chunk)    
```

Everything should be async safe and thread safe. Audio data can be put into the input stream while the chain is running without interruptions.


libs/community/langchain_community/utilities/nvidia_riva/common.py
```diff
@@ -0,0 +1,195 @@
...
+            input_iterator = cast(Iterator[_InputT], iter(input_queue.get, _TRANSFORM_END))
+            for val in self.transform(input_iterator):
+                out_queue.put_nowait(val)
+            out_queue.put_nowait(_TRANSFORM_END)
+
+        async def _consumer_coro() -> None:
+            """Coroutine that wraps the consumer."""
+            await loop.run_in_executor(None, _consumer)
+
+        loop.create_task(_producer())
```
nfcampos, line 97: I think these have to be awaited at the end, otherwise might be GCed before done

rmkraus: Good eye. Adding this now.

haydeniw:
> To use these runnables, the NVIDIA Riva client libraries are required. It they are not installed, an error will be raised instructing how to install them.
@rmkraus should we link to riva client installation docs?

rmkraus:
@haydeniw - This error is the error text if one were to try and use the Riva stages without the Riva Client python library:

> Could not import the NVIDIA Riva client library. 
> Please install it with `pip install nvidia-riva-client`.

It is meant to be similar to the error message from the OpenAI Runnables.

rmkraus:
@efriis - I just rebased this branch, fixed the spelling errors, and created a new poetry lock. Should be safe to re-run the pipeline now.

rmkraus:
@efriis - I THINK we should be good to kick off the pipelines again. I addressed all the flags from the last run. I'm just not 100% sure I fixed the core unit tests run.

efriis:
Hey @rmkraus ! You can run the equivalent tests locally with

```
cd libs/community
make lint
make test
make check_imports
```

If you run into problems let me know!

libs/community/pyproject.toml
```diff
@@ -258,7 +260,7 @@ extended_testing = [
  "elasticsearch",
  "hdbcli",
  "oci",
- "rdflib",
```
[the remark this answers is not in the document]
rmkraus: Doh 🤦 

I have no idea how that got removed. Nice catch!

rmkraus:
Thanks @efriis! I made the necessary updates and used the provided make commands to ensure the tests ran for me locally. So everything is passing on Python 3.10 now. I think we can run again.

rmkraus:
```
> poetry run python --version
Python 3.10.12
> cd libs/community
> rm -rf .venv
> poetry install --sync --with dev,lint,test -E extended_testing 
> make lint tests check_imports
```

That all passed locally. Should I have done something differently?

rmkraus:
Looks like the issue was I had a call to the `riva.client` extended_testing dependency hiding. I used this to run tests to make sure to run w/ and w/out extended testing. 

```
poetry install --sync --with dev,lint,test -E extended_testing; 
make lint tests check_imports;
poetry install --sync --with dev,lint,test; 
make lint tests check_imports;
```

@efriis: I'd like to say I think the tests should pass, but I think we can all agree I've done a pretty poor job intuiting that. But at least this issue has been resolved and tested on my side.

nfcampos:
@rmkraus looks like a conflict has appeared

rmkraus:
Resolving that conflict will be easy, resolving the last CI errors will be tougher....

After some digging:
- The DashVector integration requires dashvector 1.0.1 (Released Aug 24, 2023)
- dashvector 1.0.1 requires protobuf 3. Lastest version is 3.20.3. Release September 29, 2022. [Supported ended 23Q2](https://protobuf.dev/support/version-support/#python).
- The Riva Client library requires grpcio-tools. The latest version of grpcio-tools that supports Protobuf 3 is 1.48.2. This version is not compatible with Python 3.11.

Even the latest version of dashvector (1.0.10) requires Protobuf 3.

So, we would have to drop support for one of these things:
- dashvector
- Python 3.11
- Riva (this MR)

But holding Protobuf back to a version from 2022 seems risky.

What is the best route to go here?

rmkraus:
Another potential solution, but it wouldn't be pretty.

The dashvector unit tests dont test the vector db integration and dont require the dashvector library. Also, the docs for using dashvector specifically say to install the dependency. 

If dashvector is removed from the pyproject file, then the protobuf version can be bumped for all the other dependencies. A comment could be left in the pyproject file as to why it was commented out. All the unit tests will still pass, all the docs will still be valid, and no functionality is lost. Its just a bit ugly.

rmkraus:
Nobody sneeze! @efriis, all the tests and conflicts are resolved!

efriis (approved):
boom!

PR #15977, opened 2024-01-13
langchain: Add konlpy based text splitter for Korean text

untilhamza:
- **Description:** Add a text splitter based on [Konlpy](https://konlpy.org/en/latest/#start) which is a Python package for natural language processing (NLP) of the Korean language. (It is  like Spacy or NLTK for Korean)
  - **Issue:** the issue # it fixes if applicable,
  - **Dependencies:** Konlpy would have to be installed before this splitter is used,
  - **Twitter handle:** @untilhamza

If no one reviews your PR within a few days, please @-mention one of @baskaryan, @eyurtsev, @hwchase17


untilhamza:
Will be reopened with a single clean commit

PR #15978, opened 2024-01-13
langchain[minor] - more generic streaming support

eyurtsev:
This PR introduces astream_event interface to all runnables.

Checklist:

- [x] Verify that each event has a good schema
- [x]  Verify that each event does not do unintuitive type coercsions (inputs/outputs should not be automatically upgraded to dicts). This needs to be addressed `inputs=inputs if isinstance(inputs, dict) else {"input": inputs}`
- [ ] ~~Event `data` should not be dict if it doesn't need to be, but we don't want to get this wrong in case we need to dump extra info there!~~  `data` will always be a dict, with optional keys of `input` / `output` / `chunk` depending on event type. The reason is that `input` is not always available at event start
- [x] We need to make sure that the root run also generates event information
- [x] Propagate inputs when possible in Runnables
- [x]  Tool inputs are encoded as json strings? Need to decode potentially
- [ ] Documentation

Changes downstream:

- `inputs` available in `logs` path of astream_log -- this could lead to breaking changes in langserve if langserve fails to serialize any of the messages. (If there are messages there that are well known but haven't been registered as well known messages).
- the `output` and `final_output` has been made more consistent, one could argue that this is a bug fix, but if it's widely used could cause user's code to break.

Potential Future Changes: 

* StreamingEvent add `parent_id` key which has the id of the parent runnable
* [Internal] Update the BaseTracer or create a new version of it that stops doing type
  coercions (needs to be backwards compatibility to avoid breaking existing functionality that relies on dictinoaries)

### DIFF
--- a/docs/docs/modules/agents/how_to/streaming_events.ipynb
+++ b/docs/docs/modules/agents/how_to/streaming_events.ipynb
@@ -103,8 +103,8 @@
     "async def where_cat_is_hiding(callbacks: Callbacks) -> str:  # <--- Accept callbacks\n",
     "    \"\"\"Where is the cat hiding right now?\"\"\"\n",
     "    chunks = [\n",
-    "        chunk async for chunk in \n",
-    "        model.astream(\n",
+    "        chunk\n",
+    "        async for chunk in model.astream(\n",
     "            \"Give one up to three word answer about where the cat might be hiding in the house right now.\",\n",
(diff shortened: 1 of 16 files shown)


eyurtsev:
Closing in favor of: https://github.com/langchain-ai/langchain/pull/16172

PR #15986, opened 2024-01-13
community : Fixed 'coroutine' object is not subscriptable error

ksraunak77:
- **Description:** Added parenthesis in return statement of aembed_query() funtion to fix 'coroutine' object is not subscriptable error.
  - **Dependencies:** NA


baskaryan (approved):
thanks @raunakshrivastava7!

PR #16003, opened 2024-01-14
langchain: Add konlpy based text splitting for Korean

untilhamza:
- **Description:** Adds a text splitter based on [Konlpy](https://konlpy.org/en/latest/#start) which is a Python package for natural language processing (NLP) of the Korean language. (It is  like Spacy or NLTK for Korean)
  - **Dependencies:** Konlpy would have to be installed before this splitter is used,
  - **Twitter handle:** @untilhamza


baskaryan (approved):
could we add an example to the docs

untilhamza:
> could we add an example to the docs

@baskaryan, I have added the documentation.

untilhamza:
@baskaryan , @hwchase17  could we please get this one merged 🙏🏼 , its getting old here xD

baskaryan (approved):

PR #16005, opened 2024-01-14
community: implement adelete from VectorStore in Qdrant

AntonioMorales97:
**Description:**
Implement `adelete` function from `VectorStore` in `Qdrant` to support other asynchronous flows such as async indexing (`aindex`) which requires `adelete` to be implemented. Since `Qdrant` can be passed an async qdrant client, this can be supported easily.


baskaryan (approved):
cc @kacperlukawski

PR #16010, opened 2024-01-14
fix(langchain): force ReAct agent to stop after final answer

mspronesti:
- **Description:** ReAct agents often fail when used with completion models because they don't stop after providing the final answer. Specifically, the chain continues because the LLM hallucinates newer questions. This PR makes "Question" a default stop sequence for a react agent.


baskaryan (commented):
what if we expose `stop` as an argument, and it can default to "\nObservation"?

mspronesti:
@baskaryan Exposing `stop` sounds good to me. However, I think that `\nObservation` must be a mandatory stopping sequence for a ReAct chain. How about we expose `stop` (defaulted to None) and if provided we merge (extend) it with `\nObservation`?

mspronesti:
@baskaryan I've just appended another commit that exposes `stop` as a parameter while keeping `"\nObservation"` as a non-optional stop sequence. Could you please review this PR again? :)

mspronesti:
@baskaryan Any thoughts on this? Would you like me to instead expose `stop` defaulting it to "Observation"? Thanks :)

mspronesti:
@baskaryan any chance this gets reviewed again? Thanks in advance!

baskaryan:
@mspronesti gave it some more thought and think allowing for users to specify whatever stop words outside the create_react_agent and then respecting those may be a more robust/reliable approach, what do you think? https://github.com/langchain-ai/langchain/compare/bagatur/bind_outside_agent?expand=1

cc @hwchase @nfcampos

mspronesti:
@baskaryan Both the alternatives look good to me. From the developer's perspective, I believe your solution looks better. However, from the user's perspective I find it less intuitive than mine, as that requires knowing to first set the stop sequence in the LLM and then passing it. However this is very subjective.

I'm fine with both the alternatives. Just let me know if you would like me to update this PR. Thanks!

mspronesti:
Hi @baskaryan, do you have a final verdict on this feature? Thanks!

mspronesti:
Hi @baskaryan, sorry for pinging you again. Do you have any updates on this PR or on #16747? At the moment there's no way to pass a stop sequence to a react agent.

mspronesti:
@baskaryan I've just resolved the conflicts of this PR, in case you wanted to reviewing it again :)

mspronesti:
@baskaryan  @hwchase17  I've just resolved the conflicts of this PR, in case you want to review it again :)

ccurme:
Looks like this is supported following https://github.com/langchain-ai/langchain/pull/19333. Apologies as this PR was first.

PR #16012, opened 2024-01-14
fix output parsers navigation url

Ahmad-Bunni:
Output parsers documentation have broken links.

[output parsers](https://python.langchain.com/docs/modules/model_io/output_parsers/)

Corrected the prefix for all links.


Ahmad-Bunni:
Closing as it seems to be working not sure why I had 404s at first.

PR #16016, opened 2024-01-15
langchain: Indexing with `ParentDocumentRetriever`

chadnorvell:
- **Description:** Support using indexing along with `ParentDocumentRetriever`. In short, when you create a `ParentDocumentRetriever`, you can provide the arguments you would have provided to `index`. Then when you call `.add_documents`, `index` will be called with the provided arguments instead of directly persisting to the vector store.

This change also allows you to optionally provide a list of text splitters to `.add_documents` (being the same length as the list of documents), in case a single text splitter type for all docs that go through the receiver isn't appropriate.

### DIFF
--- a/libs/langchain/langchain/retrievers/parent_document_retriever.py
+++ b/libs/langchain/langchain/retrievers/parent_document_retriever.py
@@ -1,8 +1,9 @@
 import uuid
-from typing import List, Optional
+from typing import Any, Callable, Dict, List, Optional, Union
 
 from langchain_core.documents import Document
 
+from langchain.indexes import IndexingResult, index
 from langchain.retrievers import MultiVectorRetriever
 from langchain.text_splitter import TextSplitter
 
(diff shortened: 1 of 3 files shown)


libs/langchain/langchain/retrievers/parent_document_retriever.py
```diff
@@ -56,19 +57,47 @@ class ParentDocumentRetriever(MultiVectorRetriever):
...
     """
 
-    child_splitter: TextSplitter
-    """The text splitter to use to create child documents."""
+    child_splitter: Optional[TextSplitter] = None
+    """The text splitter to use to create child documents.
+    If none, you will have to provide a list of text splitters corresponding
+    to each document when calling `add_documents`."""
 
+    id_key: str = "doc_id"
```
chadnorvell, line 65: I know this is already defined on the parent class and it's better not to redundantly define it here too. But without it, the useful docstring below it is orphaned.

libs/langchain/langchain/retrievers/parent_document_retriever.py
```diff
@@ -56,19 +57,47 @@ class ParentDocumentRetriever(MultiVectorRetriever):
...
     """The text splitter to use to create parent documents.
     If none, then the parent documents will be the raw documents passed in."""
 
+    index_args: Optional[Dict[str, Any]] = None
+    """If defined, indexing will be used when storing split documents to the
+    vectorstore. This should contain the arguments that you would pass to
+    `index`, but without `docs_source` and `vector_store`, both of which will
+    be passed automatically."""
+
+    @property
```
hwchase17, line 79: a property that returns a function is a bit weird - shouldnt this just be a method?

chadnorvell: You're right. I'm not sure what I was originally thinking here, but as it stands, this is just a method with extra steps. I'll fix it.

baskaryan:
@chadnorvell did you want to keep working on this?

eyurtsev:
Going to close as there has been no activity here.

PR #16039, opened 2024-01-15
community: add new gpt-3.5-turbo-1106 finetuned for cost calculation

nirkopler:
**Description:** Added the new gpt-3.5-turbo-1106 for **finetuned** cost calculation, 
**Issue:** no issue found open

By the information in OpenAI the pricing is the same as the older model (0613)


hwchase17 (approved):
i think we should also have this in langchain_openai, but i dont see the callback there yet

kimsinhyun:
seems like price for ft-gpt3.5-turbo is wrong
image

PR #16044, opened 2024-01-15
changed default params for gemini

lkuligin:
  - **Description:** changed default values for Vertex LLMs (to be handled on the SDK's side)


libs/partners/google-vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -51,7 +51,7 @@ def test_candidates() -> None:
     assert len(response.generations[0]) == 2
 
 
-@pytest.mark.parametrize("model_name", ["chat-bison@001", "gemini-pro"])
+@pytest.mark.parametrize("model_name", ["chat-bison@001"])
```
efriis, line 54: Given the differences between the available params for palm and gemini, would it make sense to continue to test this?

efriis (commented):
One comment, otherwise code lgtm.

Are the default params set the same in the sdk? I'm ok deferring to the sdk's default, and if they're different, it'd be good to message that this is a breaking change.

libs/partners/google-vertexai/tests/integration_tests/test_chat_models.py
lkuligin: I'd say "yes" for Palm.

lkuligin:
> One comment, otherwise code lgtm.
> 
> Are the default params set the same in the sdk? I'm ok deferring to the sdk's default, and if they're different, it'd be good to message that this is a breaking change.

how should we do this? 

P.S. Gemini integration is relatively new, so maybe it's fine to introduce a breaking change, wdyt? There were a few complaints that chat model was throwing the exception, and the main reason is the low default `max_output_tokens` value (that I used the same as for Palm, and apparently it was not a good choice).

libs/partners/google-vertexai/tests/integration_tests/test_chat_models.py
efriis: Why not worth testing gemini here?

efriis:
Breaking change on these defaults works for me.

The other option is if we're seeing divergence between Palm and Gemini features, we can introduce separate `ChatVertexGemini` and `VertexGemini` wrappers, and deprecate gemini usage through these wrappers. Thoughts?

libs/partners/google-vertexai/tests/integration_tests/test_chat_models.py
efriis: Sounds good. In the future, if the flakiness is something that should be fixed elsewhere, feel free to mark it as `@pytest.mark.xfail(...)`

efriis (approved):

Issue #16023, opened 2024-01-15
langchain_community: SQL Database - Don't assume "PostgreSQL" in the else


skye0402:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

Following code in https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/utilities/sql_database.py from line 386:
```
        with self._engine.begin() as connection:  # type: Connection
            if self._schema is not None:
                if self.dialect == "snowflake":
                    connection.exec_driver_sql(
                        "ALTER SESSION SET search_path = %s", (self._schema,)
                    )
                elif self.dialect == "bigquery":
                    connection.exec_driver_sql("SET @@dataset_id=?", (self._schema,))
                elif self.dialect == "mssql":
                    pass
                elif self.dialect == "trino":
                    connection.exec_driver_sql("USE ?", (self._schema,))
                elif self.dialect == "duckdb":
                    # Unclear which parameterized argument syntax duckdb supports.
                    # The docs for the duckdb client say they support multiple,
                    # but `duckdb_engine` seemed to struggle with all of them:
                    # https://github.com/Mause/duckdb_engine/issues/796
                    connection.exec_driver_sql(f"SET search_path TO {self._schema}")
                elif self.dialect == "oracle":
                    connection.exec_driver_sql(
                        f"ALTER SESSION SET CURRENT_SCHEMA = {self._schema}"
                    )
                elif self.dialect == "sqlany":
                    # If anybody using Sybase SQL anywhere database then it should not
                    # go to else condition. It should be same as mssql.
                    pass
                else:  # postgresql and other compatible dialects
                    connection.exec_driver_sql("SET search_path TO %s", (self._schema,))
```

### Description

Can we change the code of line 412:
```
                else:  # postgresql and other compatible dialects
                    connection.exec_driver_sql("SET search_path TO %s", (self._schema,))
```

to 

```
elif self.dialect == "postgresql":
  connection.exec_driver_sql("SET search_path TO %s", (self._schema,))
else:
  pass # Do nothing
```

It is in my opinion simply wrong to assume that any other DB will work like PostgreSQL. And it's not possible to cut that statement in the dialect easily. If we need more `elif`, so be it.

### System Info

Python 3.10 - it's independent of the Python version or environment.

### Related Components

- [ ] LLMs/Chat Models
[...]

PR #16047, opened 2024-01-15
langchain_community: Small Fix for the SQLDatabase Tool

keenborder786:
- **Description:** Made a small fix for the `SQLDatabase` highlighted in an issue. The issue pertains to switching schema for different SQL engines.
  - **Issue:** #16023
@baskaryan


libs/community/langchain_community/utilities/sql_database.py
```diff
@@ -409,8 +409,9 @@ def _execute(
                     # If anybody using Sybase SQL anywhere database then it should not
                     # go to else condition. It should be same as mssql.
                     pass
-                else:  # postgresql and other compatible dialects
+                elif self.dialect == "postgresql":  # postgresql
```
baskaryan, line 412: this is a breaking change since it changes behavior for any dialect not explicitly specified above. if we have specific dialects we want to skip maybe better to add them explicitly

keenborder786: I don't see this as a breaking change because as mentioned in the issue, assuming that `SET search_path to X` will work for all dialects other than the mentioned above is wrong and error-prune. For example MySQL will raise an error with the current implementation. The existing behaviour that it is resolving is actually an error and therefore I don't see how current users will find any unintended consequences which might break their code. If they mentioned schema for a dialect which is not changable by `SET search_path to X` e.g MySQL, it would have regardless raised an error.

baskaryan: makes sense, thank you @keenborder786!

baskaryan (approved):

PR #16048, opened 2024-01-15
core[patch]: pass exceptions to fallbacks

baskaryan:


libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -89,6 +91,11 @@ def when_all_is_lost(inputs):
     
     Any exception that is not a subclass of these exceptions will be raised immediately.
     """
+    exception_key: Optional[str] = None
```
eyurtsev, line 94: Could you update the doc-string to explain how to interpret this key?

It's unclear from the documentation how this key interacts with exceptions_to_handle.

Is it always passed to the runnable or only for exceptions that are defined in the list of exceptions to handle or only for  unhandled exceptions?

libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -171,6 +186,10 @@ async def ainvoke(
         config: Optional[RunnableConfig] = None,
         **kwargs: Optional[Any],
     ) -> Output:
+        if self.exception_key is not None and not isinstance(input, dict):
+            raise ValueError(
+                "If 'exception_key' is specified then input must be a dictionary."
```
eyurtsev, line 191: Could you expand the exception message: 

"However found a type of `{type(input)}` for input" 

or something to that extent to make the error message more detailed

libs/core/tests/unit_tests/runnables/test_fallbacks.py
```diff
@@ -0,0 +1,190 @@
...
+
+def _assert_potential_error(actual: list, expected: list) -> None:
+    for x, y in zip(actual, expected):
+        if isinstance(x, Exception):
+            assert isinstance(y, type(x))
+        else:
+            assert x == y
+
+
+def test_batch() -> None:
```
eyurtsev, line 103: Missing test for invoke / ainvoke I think

eyurtsev (approved):
API make sennse for me. I haven't reviewed batch implementation in detail, but assuming OK due to unit tests

PR #16055, opened 2024-01-16
core[patch]: simple fallback streaming

baskaryan:


libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -343,3 +347,74 @@ async def abatch(
...
         await asyncio.gather(*(rm.on_chain_error(first_error) for rm in run_managers))
         raise first_error
+
+    def stream(
+        self,
+        input: Input,
+        config: Optional[RunnableConfig] = None,
+        **kwargs: Optional[Any],
+    ) -> Iterator[Output]:
+        # setup callbacks
```
eyurtsev, line 357: Would you mind adding a doc-string here to document the semantics a bit better; i.e., what happens if the exception it not during the first chunk?

libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -343,3 +347,74 @@ async def abatch(
...
+        for runnable in self.runnables:
+            try:
+                stream = runnable.stream(
+                    input,
+                    patch_config(config, callbacks=run_manager.get_child()),
+                    **kwargs,
+                )
+                chunk = next(stream)
+            except self.exceptions_to_handle as e:
+                first_error = e if first_error is None else first_error
```
eyurtsev, line 374: We have an `on_retry` callback. I was expecting to see it used within this Runnable, but don't see it anywhere. Do we need to add it to the various calls or do we think it doesn't make sense for this runnable or some other issue?

eyurtsev (approved):
Looks good to me

baskaryan:
still need to add passing in exception

PR #16059, opened 2024-01-16
updated imports for `output_parsers`

leo-gan:
Updated imports from `langchain` to `core` where it is possible


baskaryan (approved):

leo-gan:
@baskaryan Thanks!!!

PR #16061, opened 2024-01-16
updated `agents` imports

leo-gan:
Updated imports into `langchain` to `core` where it is possible


leo-gan:
@baskaryan Thanks!!!

PR #16064, opened 2024-01-16
updated `chains` imports

leo-gan:
Updated imports into `langchain` to `core` where it is possible


baskaryan (approved):

leo-gan:
@baskaryan Thanks!!!

PR #16090, opened 2024-01-16
Update elasticsearch.ipynb

bustosjuan:
Fixed a typo, the parameter used for the Elasticsearch API key was called api_key, but the parameter is called es_api_key.


baskaryan:
thanks @bustosjuan

PR #16091, opened 2024-01-16
docs:Updated integration docs structure for llm/google_vertex_ai_palm

jzaldi:
- **Description**: Updated doc for llm/google_vertex_ai_palm with new functions: `invoke`, `stream`... Changed structure of the document to match the required one.
- **Issue**: #15664 
- **Dependencies**: None
- **Twitter handle**: None

### DIFF
--- a/docs/docs/integrations/llms/google_vertex_ai_palm.ipynb
+++ b/docs/docs/integrations/llms/google_vertex_ai_palm.ipynb
@@ -26,7 +26,7 @@
     "id": "Q_UoF2FKN8Xb"
    },
    "source": [
-    "## Setting up"
+    "## Setup"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


jzaldi:
I left the Multimodality part unchanged for the moment, maybe we can changed in other issue later.

lkuligin:
Could you please add an explanation on top of the page that `ChatVertexAI` exposes both Gemini, Palm 2 and Codey models? It looks like some users are confused that `ChatVertexAI` doesn't have gemini integration since it's hidden somewhere.

jzaldi:
> Could you please add an explanation on top of the page that `ChatVertexAI` exposes both Gemini, Palm 2 and Codey models? It looks like some users are confused that `ChatVertexAI` doesn't have gemini integration since it's hidden somewhere.

If added a paragraph to explain that the api exposes all models available with a link to the documentation.

lkuligin (approved):

efriis (approved):

Issue #16096, opened 2024-01-16
Allow MessagesPlaceholder to include a maximum number of messages


DaveDeCaprio:
### Feature request

When using `MessagesPlaceholder`, it always includes all messages.  For some prompts I want the full message history, but for others, I want to limit it to just the most recent few messages.

This can be accomplished with `ConversationBufferWindowMemory`, but that limits the memory storage which is used for all my LLM prompts in the chain.  

I currently accomplish this with a custom prompt generator, but it would be easier if there were an optional max_messages parameter to MessagesPlaceholder that limits the history to the given number of messages.

### Motivation

I have chain with multiple LLM calls.  For one, the routing prompt, it works best when passed the last few messages in the conversation.  Right now I have to use a custom prompt generator to accomplish this.  I'd really like to be able to just add a limit to MessagesPlaceholder.

### Your contribution

I will submit a PR,

PR #16098, opened 2024-01-16
core:Add optional max_messages to MessagePlaceholder

DaveDeCaprio:
- **Description:** Add optional max_messages to MessagePlaceholder
- **Issue:** [16096](https://github.com/langchain-ai/langchain/issues/16096)
- **Dependencies:** None
- **Twitter handle:** @davedecaprio

Sometimes it's better to limit the history in the prompt itself rather than the memory.  This is needed if you want different prompts in the chain to have different history lengths.


DaveDeCaprio:
@baskaryan Do you not like the feature idea, or was it something about the submission?

baskaryan:

psaks:
This is a great idea @DaveDeCaprio. However, I suggest that we make the window length consistent with the ConversationBufferWindowMemory below by renaming the max_messages to k and adding a factor 2, such that the actual number of messages loaded is 2 * k. What do you think? 

```class ConversationBufferWindowMemory(BaseChatMemory):
    """Buffer for storing conversation memory inside a limited size window."""

    human_prefix: str = "Human"
    ai_prefix: str = "AI"
    memory_key: str = "history"  #: :meta private:
    k: int = 5
    """Number of messages to store in buffer."""

    @property
    def buffer(self) -> Union[str, List[BaseMessage]]:
        """String buffer of memory."""
        return self.buffer_as_messages if self.return_messages else self.buffer_as_str

    @property
    def buffer_as_str(self) -> str:
        """Exposes the buffer as a string in case return_messages is True."""
        messages = self.chat_memory.messages[-self.k * 2 :] if self.k > 0 else []
        return get_buffer_string(
            messages,
            human_prefix=self.human_prefix,
            ai_prefix=self.ai_prefix,
        )```

DaveDeCaprio:
@psaks Good idea.  I'll update the PR today.

hwchase17 (approved):

PR #16108, opened 2024-01-16
core : in RunnableRetry, correct the example snippet that uses with_retry method on Runnable

ksachdeva:
The example code snippet for with_retry is using incorrect argument names. This PR fixes that


baskaryan (approved):

baskaryan:
thanks @ksachdeva!

PR #16109, opened 2024-01-16
Community[patch]use secret str in Tavily and HuggingFaceInferenceEmbeddings

hinthornw:
So the api keys don't show up in repr's 

Still need to do tests


baskaryan (commented):
can we add simple test

PR #16110, opened 2024-01-16
Community[Patch] Remove docs form bm25 repr

### LINKED ISSUE #356: Issue: BM25 retriever __repr__ method reports all documents to langsmith
### Issue you'd like to raise.

I found an interesting issues that required overriding the BM25 retriever `__repr__` when using langsmith.

With langsmith tracing enabled using a BM25 retriever, langsmith sends the entire BM25 `__repr__` which reports every document. This is problematic when you have a large number of documents or large documents and leads to extremely large calls, immediately exhausting the quota. 

Initial Output: 
"repr": "EnsembleRetriever(retrievers=[BM25Retriever(vectorizer=<rank_bm25.BM25Okapi object at 0x7fdac39dae50>, docs=[Document(page_content=...), ...] 

New Output:
"repr": "EnsembleRetriever(retrievers=[BM25Retriever(), VectorStoreRetriever(tags=['Chroma', 'OpenAIEmbeddings'], vectorstore=<langchain_community.vectorstores.chroma.Chroma object at 0x7fec8667f410>, search_type='similarity_score_threshold', search_kwargs={'score_threshold': 0.2})], weights=[0.5, 0.5])",

Simple override
```
class BM25RetrieverReprOverride(BM25Retriever):
    def __repr__(self):
        return "BM25Retriever()" # this could be more descriptive
```

### Suggestion:

I believe this is the correct output of the BM25 retriever, but langsmith should probably have some check before sending extremely large strings

hinthornw:
Not sure how helpful they are to include in the representation?
Resolves: https://github.com/langchain-ai/langsmith-sdk/issues/356

### DIFF
--- a/libs/community/langchain_community/retrievers/bm25.py
+++ b/libs/community/langchain_community/retrievers/bm25.py
@@ -4,8 +4,8 @@ from typing import Any, Callable, Dict, Iterable, List, Optional
 
 from langchain_core.callbacks import CallbackManagerForRetrieverRun
 from langchain_core.documents import Document
-from langchain_core.retrievers import BaseRetriever
 from langchain_core.pydantic_v1 import Field
+from langchain_core.retrievers import BaseRetriever
 
 
 def default_preprocessing_func(text: str) -> List[str]:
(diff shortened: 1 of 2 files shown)


baskaryan (commented):
can we add simple test

PR #16117, opened 2024-01-17
partners: add license field

efriis:
- bumps package post versions for packages without current unreleased updates
- will bump package version in release prs associated with packages that do have changes (mistral, vertex)


efriis:
Failure on vertex lint unrelated to this PR - will fix separately

PR #16120, opened 2024-01-17
Add Tuple support for messages placeholder

hinthornw:
It's a bit annoying to work with LangChain messages on the frontend, which is why we [did things like this in chat langchain](https://github.com/langchain-ai/chat-langchain/blob/592d588a1bfe35fe03ed33ac8ba36afd9452bb6d/chain.py#L148). It would be convenient if the messages placeholder also accepted the same tuple format as the prompt template, so you could just pass the role and content values around.

Maybe not as useful in a real memory setup, but simplifies the user code a bit in some situations.


baskaryan (commented):
what if placeholder tuple was ("placeholder", "<VAR_NAME>")

hinthornw:
Oh hmm i kinda like that too!

efriis:
Going #19417 instead because of refactor related merge conflicts

Issue #16132, opened 2024-01-17
Implement vector length definition at init time in PGVector for indexing


Frank995:
### Feature request

Allow manual definition of embedding column length in PGVector

### Motivation

Right now embeddings stores in PGVector are initialised with:
`embedding: Vector = sqlalchemy.Column(Vector(None))`
This won't allow indexing of the DB ([https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/howto-optimize-performance-pgvector](url))

### Your contribution

I can open a PR shortly to fix it

RobertHH-IS:
Please get this into the next release - langchain is a ways behind llamaindex when it comes to pgvector support.  They already support defining your own table names, correct column init,  both index based retrieval methods, as well as hybrid search.  Langchain support is not production ready without just using SQL.

PR #16133, opened 2024-01-17
Implement vector length definition at init time in PGVector for indexing

Frank995:
  - **Description:** allow user to define tVector length in PGVector when creating the embedding store, this allows for later indexing
  - **Issue:** #16132
  - **Dependencies:** None

### DIFF
--- a/libs/community/langchain_community/vectorstores/pgvector.py
+++ b/libs/community/langchain_community/vectorstores/pgvector.py
@@ -151,6 +151,10 @@ class PGVector(VectorStore):
         connection_string: Postgres connection string.
         embedding_function: Any embedding function implementing
             `langchain.embeddings.base.Embeddings` interface.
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/pgvector.py
```diff
@@ -183,6 +183,7 @@ def __init__(
         self,
         connection_string: str,
         embedding_function: Embeddings,
+        embedding_length: Optional[int] = None,
```
baskaryan, line 186: could we document arg in docstring

Frank995: Done

Frank995:
Bump as would be really needed

baskaryan (approved):
sorry for delay, thank you @Frank995!

PR #16137, opened 2024-01-17
Use dotenv in langchain-community's integration tests

cbornet:
* Removed some env vars not used in langchain package IT
* Added Astra DB env vars in langchain package, used for cache tests
* Added conftest.py to load env vars in langchain_community IT
* Added .env.example in  langchain_community IT

### DIFF
--- a/libs/langchain/tests/integration_tests/.env.example
+++ b/libs/langchain/tests/integration_tests/.env.example
@@ -8,23 +8,6 @@ OPENAI_API_KEY=your_openai_api_key_here
 SEARCHAPI_API_KEY=your_searchapi_api_key_here
 
 
-# pinecone
-# your api key from left menu "API Keys" in https://app.pinecone.io
-PINECONE_API_KEY=your_pinecone_api_key_here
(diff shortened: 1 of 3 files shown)


libs/langchain/tests/integration_tests/.env.example
```diff
@@ -33,5 +16,7 @@ POWERBI_TABLE_NAME=_test_table_name_here
 POWERBI_NUMROWS=_num_rows_in_your_test_table
 
 
-# MongoDB Atlas Vector Search
-MONGODB_ATLAS_URI=your_mongodb_atlas_connection_string
\ No newline at end of file
+# astra db
+ASTRA_DB_API_ENDPOINT=https://your_astra_db_id-your_region.apps.astra.datastax.com
```
baskaryan, line 20: should this be in community .env.example?

cbornet: it's in both.
We also need it here because the LLM cache impls are in the langchain package, not langchain-community

PR #16138, opened 2024-01-17
Add neo4j timeout and value sanitization option

tomasonjo:
The timeout function comes in handy when you want to kill longrunning queries.
The value sanitization removes all lists that are larger than 128 elements. The idea here is to remove embedding properties from results.

### DIFF
--- a/libs/community/langchain_community/graphs/neo4j_graph.py
+++ b/libs/community/langchain_community/graphs/neo4j_graph.py
@@ -31,6 +31,33 @@ RETURN {start: label, type: property, end: toString(other_node)} AS output
 """
 
 
+def value_sanitize(d: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    The idea is to remove all internal properties which are irrelevant for generating answers
+    """
+    LIST_LIMIT = 128
+    # Create a new dictionary to avoid changing size during iteration
+    new_dict = {}
+    for key, value in d.items():
+        if isinstance(value, dict):
+            # Recurse to handle nested dictionaries
+            new_dict[key] = value_sanitize(value)
+        elif isinstance(value, list):
+            # check if it has less than LIST_LIMIT values
+            if len(value) < LIST_LIMIT:
+                # if value is a list, check if it contains dictionaries to clean
+                cleaned_list = []
+                for item in value:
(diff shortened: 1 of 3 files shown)


libs/community/langchain_community/graphs/neo4j_graph.py
```diff
@@ -31,6 +31,34 @@
 """
 
 
+def value_sanitize(d: Dict[str, Any]) -> Dict[str, Any]:
```
baskaryan, line 34: in the end does this just remove lists longer than LIST_LIMIT?

tomasonjo: Yes

libs/community/langchain_community/graphs/neo4j_graph.py
```diff
@@ -31,6 +31,34 @@
...
+    """
+    The idea is to remove all internal properties which
+    are irrelevant for generating answers
+    """
+    LIST_LIMIT = 128
```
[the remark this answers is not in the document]
tomasonjo: I don't think anyone would ever want to configure this. Most, if not all text/image embeddings are larger than 128 elements and that's basically the only reason to add this. If you have embeddings on nodes, but also want to use GraphCypherQAChain, then you might get full nodes with all attributes as results, and embedding attributes introduce problems for context size as well as they are largely irrelevant.

libs/community/langchain_community/graphs/neo4j_graph.py
baskaryan: could we explain that in docstring

tomasonjo: Added

libs/community/langchain_community/graphs/neo4j_graph.py
```diff
@@ -52,6 +80,8 @@ def __init__(
         username: Optional[str] = None,
         password: Optional[str] = None,
         database: str = "neo4j",
+        timeout: Optional[float] = None,
+        sanitize: bool = False,
```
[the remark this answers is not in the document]
tomasonjo: Added

baskaryan (approved):

PR #16167, opened 2024-01-18
Add revision identifier to run_on_dataset

samnoyes:
Allow specifying revision identifier for better project versioning

### DIFF
--- a/libs/langchain/langchain/smith/evaluation/runner_utils.py
+++ b/libs/langchain/langchain/smith/evaluation/runner_utils.py
@@ -684,7 +684,7 @@ async def _arun_llm(
             return await llm.ainvoke(
                 prompt_or_messages,
                 config=RunnableConfig(
-                    callbacks=callbacks, tags=tags or [], metadata=metadata
+                    callbacks=callbacks, tags=tags or [], metadata=metadata or {}
                 ),
             )
         else:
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -908,6 +915,7 @@ def _run_llm_or_chain(
                 config["callbacks"],
                 tags=config["tags"],
                 input_mapper=input_mapper,
+                revision_identifier=config.get("metadata"),
```
hinthornw, line 918: wrong key right?

libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -1121,6 +1134,7 @@ def prepare(
                 ],
                 tags=tags,
                 max_concurrency=concurrency_level,
+                metadata={"revision_identifier": revision_identifier} if revision_identifier else None,
```
hinthornw, line 1137: Should we just call it a "revision"? Seems a bit verbose

libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -1260,6 +1277,7 @@ def run_on_dataset(
         input_mapper,
         concurrency_level,
         project_metadata=project_metadata,
+        revision_identifier=revision_identifier,
```
hinthornw, line 1280: Could we also update the `_RUN_ON_DATASET_DOCSTRING` below?

hinthornw (approved):

PR #16171, opened 2024-01-18
community: New documents loader for visio files (with extension .vsdx)

florian-morel22:
**Description** : New documents loader for visio files (with extension .vsdx)

A [visio file](https://fr.wikipedia.org/wiki/Microsoft_Visio) (with extension .vsdx) is associated with Microsoft Visio, a diagram creation software. It stores information about the structure, layout, and graphical elements of a diagram. This format facilitates the creation and sharing of visualizations in areas such as business, engineering, and computer science.

A Visio file can contain multiple pages. Some of them may serve as the background for others, and this can occur across multiple layers. This loader extracts the textual content from each page and its associated pages, enabling the extraction of all visible text from each page, similar to what an OCR algorithm would do.

**Dependencies** : xmltodict package


baskaryan (approved):
thank you @florian-morel22!

PR #16202, opened 2024-01-18
Fix Azure env params syntax in retrieval agent

braun-viathan:
This pull request fixes the syntax for Azure environment parameters in the retrieval agent. The new syntax is used to set the Azure endpoint, API key, and API version. The old syntax, which involved environment variables, is commented out. This change ensures that the retrieval agent functions correctly with the updated Azure environment configuration.


braun-viathan (commented):
If any additional steps are required to merge this feel free to contact me.

efriis:
#16257 - had some problems pushing to your branch

braun-viathan:
> #16257 - had some problems pushing to your branch

What was the issue ? 
- tried out the template and got a lot of warninings because of the missing langchain-openai package, while trying out the playground. 
- Also habe the typical 404 error beause of the functions use. 

If It helps I can provide more error details on why and how I updated this.
Could not try out the template without using the new Syntax

Issue #16204, opened 2024-01-18
Added Token Count support for Gen AI Models


Adi8885:
### Feature request

Add support for getting num_tokens() for ChatGoogleGenerativeAI , GoogleGenerativeAI, 

### Motivation

This feature has below benefits
- Useful for checking if an input will fit in a model's context window.
- Useful for estimating usage costs

### Your contribution

I'll submit PR

PR #16205, opened 2024-01-18
added logic for method get_num_tokens()

Adi8885:

Please title your PR "partners: google-genai",

  - **Description:** : added logic for method get_num_tokens() for ChatGoogleGenerativeAI , GoogleGenerativeAI, 
  - **Issue:** : https://github.com/langchain-ai/langchain/issues/16204,
  - **Dependencies:** : None,
  - **Twitter handle:** @Aditya_Rane

### DIFF
--- a/libs/partners/google-genai/langchain_google_genai/chat_models.py
+++ b/libs/partners/google-genai/langchain_google_genai/chat_models.py
@@ -51,7 +51,7 @@ from tenacity import (
 )
 
 from langchain_google_genai._common import GoogleGenerativeAIError
-from langchain_google_genai.llms import _BaseGoogleGenerativeAI
+from langchain_google_genai.llms import GoogleModelFamily, _BaseGoogleGenerativeAI
 
 IMAGE_TYPES: Tuple = ()
 try:
@@ -74,10 +74,6 @@ class ChatGoogleGenerativeAIError(GoogleGenerativeAIError):
     Google genai API usage in the ChatGoogleGenerativeAI class, such as unsupported
     message types or roles.
     """
-
-
-def _is_gemini_model(model_name: str) -> bool:
-    return "gemini" in model_name
 
 
 def _create_retry_decorator() -> Callable[[Any], Any]:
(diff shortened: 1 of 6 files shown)


libs/partners/google-genai/langchain_google_genai/chat_models.py
```diff
@@ -74,6 +74,9 @@ class ChatGoogleGenerativeAIError(GoogleGenerativeAIError):
     message types or roles.
     """
 
+def _is_gemini_model(model_name: str) -> bool:
```
lkuligin, line 77: why not import from llms?

lkuligin:
can we add an integration test for this, please?

libs/partners/google-genai/langchain_google_genai/llms.py
```diff
@@ -259,6 +259,11 @@ def _stream(
     def _llm_type(self) -> str:
         """Return type of llm."""
         return "google_palm"
+    
+    @property
```
lkuligin, line 263: nits: maybe it's time to introduce `BaseGoogleGenerativeAI` class and avoid code duplication between chat and llm models, wdyt?

libs/partners/google-genai/langchain_google_genai/chat_models.py
```diff
@@ -650,3 +653,28 @@ def _prepare_chat(
         message = history.pop()
         chat = self.client.start_chat(history=history)
         return params, chat, message
+    
+    @property
+    def is_gemini(self) -> bool:
```
eyurtsev, line 658: I don't think we want it to be parameterized with a bool, if there will be multiple families of models, we want to return the family identifier rather than a false / true.

Perhaps better to avoid creating a property or else marking it as private?

libs/partners/google-genai/langchain_google_genai/llms.py
Adi8885: Fair point. On it

libs/partners/google-genai/langchain_google_genai/chat_models.py
Adi8885: @eyurtsev Add _get_model_family() to retreive model family

libs/partners/google-genai/langchain_google_genai/llms.py
Adi8885: Done

lkuligin (approved):

lkuligin:
@eyurtsev please, take a look

efriis (approved):

PR #16218, opened 2024-01-18
core[patch], openai[patch]: Stream logprobs

baskaryan:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -404,12 +404,15 @@ def _stream(
             chunk = _convert_delta_to_message_chunk(
                 choice["delta"], default_chunk_class
             )
-            finish_reason = choice.get("finish_reason")
-            generation_info = (
-                dict(finish_reason=finish_reason) if finish_reason is not None else None
-            )
+            generation_info = {}
```
eyurtsev, line 407: Any unit tests that we could add?

eyurtsev (approved):
Looks good to me.

eyurtsev:
@baskaryan maybe update subject line that this contains finish reason?

libs/core/langchain_core/utils/merge.py
efriis, line 1: Looks like this should only be used internal in core, so maybe worth `_merge.py`

efriis (approved):

libs/core/langchain_core/utils/merge.py
baskaryan: maybe we should rename all of utils -> _utils

libs/core/langchain_core/utils/_merge.py
```diff
@@ -0,0 +1,65 @@
...
+            raise TypeError(
+                f'additional_kwargs["{k}"] already exists in this message,'
+                " but with a different type."
+            )
+        elif isinstance(merged[k], str):
+            merged[k] += v
+        elif isinstance(merged[k], dict):
+            merged[k] = merge_dicts(merged[k], v, merge_lists=merge_lists)
+        elif isinstance(merged[k], list):
+            if merge_lists:
```
baskaryan, line 47: cc @nfcampos added this param

nfcampos (approved):

PR #16227, opened 2024-01-18
Use env for revision id if not passed in as param; use `git describe` as backup

samnoyes:


libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -1330,6 +1335,24 @@ def run_on_dataset(
     return container.finish(batch_results, verbose=verbose)
 
 
+def _get_default_revision_id() -> Optional[str]:
```
eyurtsev, line 1338: Could you add a description of what `LANGCHAIN_REVISION_ID` is? And which `git` repo is this for?

eyurtsev: Do we need to memoize the result if `git describe --tags --dirty` fails? e.g., no git installed or no .git found?

eyurtsev (approved):
This works

libs/langchain/langchain/smith/evaluation/runner_utils.py
hinthornw: Ya - he's implementing in the SDK also! we memoize there

PR #16229, opened 2024-01-18
langchain_community: Added Missing Schema for Community Tools

keenborder786:
Many of the tools in the langchain_community had no argument schema which was causing a failure in the OpenAI Functions Agent. Therefore added the argument schema for each of these tools.

@baskaryan


keenborder786:
@baskaryan can you please review?

Issue #16232, opened 2024-01-18
HuggingFaceHub removes leading characters of text-generation


mikeFore4:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

Here is a simple implementation running meta's llama2 in langchain:
```
from langchain_community.llms import HuggingFaceHub

llm = HuggingFaceHub(
        repo_id="meta-llama/Llama-2-7b-chat-hf",
        huggingfacehub_api_token=token,
        )
print(llm.invoke("Who are you?"))
```
The above code produces the following output:
```
son who is passionate about helping others. I have a strong desire to make a positive impact....
```
It appears that the beginning of the generated text is cut off. However, it _should_ produce identical output as what you would get from using huggingface directly without langchain:
```
from huggingface_hub import InferenceApi

repo_id="meta-llama/Llama-2-7b-chat-hf"

client = InferenceApi(
        repo_id=repo_id,
        token=token,
        task='text-generation')

output = client("Who are you?")
print(output[0]["generated_text"])
```
Here the output is not cutoff:
```
I am a person who is passionate about helping others. I have a strong desire to make a positive impact...
```

### Description

Some text-generation models on huggingface repeat the prompt in their generated response, but not all do! If you are using a model like "gpt2" which DOES repeat the prompt langchain will properly remove the first few characters of the response (to match the len(prompt)). However, if you are using a model (such as the very popular "meta-llama/Llama-2-7b-chat-hf") that DOES NOT repeat the prompt in its generated text, then the beginning of the generated text will be cut off.

### System Info

langchain==0.1.1
langchain-community==0.0.13
langchain-core==0.1.10
langchain-openai==0.0.2.post1

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

PR #16233, opened 2024-01-18
community: huggingface hub character removal bug fix

mikeFore4:
- **Description:** Some text-generation models on huggingface repeat the prompt in their generated response, but not all do! The tests use "gpt2" which DOES repeat the prompt and as such, the HuggingFaceHub class is hardcoded to remove the first few characters of the response (to match the len(prompt)). However, if you are using a model (such as the very popular "meta-llama/Llama-2-7b-chat-hf") that DOES NOT repeat the prompt in it's generated text, then the beginning of the generated text will be cut off. This code change fixes that bug by first checking whether the prompt is repeated in the generated response and removing it conditionally.
  - **Issue:** #16232 
  - **Dependencies:** N/A
  - **Twitter handle:** N/A


baskaryan (approved):

baskaryan:
thanks @mikeFore4!

PR #16246, opened 2024-01-19
community: Add `SQLDatabaseLoader` document loader

amotl:
- **Description:** A generic document loader adapter for SQLAlchemy on top of LangChain's `SQLDatabaseLoader`.
  - **Needed by:** https://github.com/crate-workbench/langchain/pull/1
  - **Depends on:** GH-16655
  - **Addressed to:** @baskaryan, @cbornet, @eyurtsev

Hi from CrateDB again,

in the same spirit like GH-16243 and GH-16244, this patch breaks out another commit from https://github.com/crate-workbench/langchain/pull/1, in order to reduce the size of this patch before submitting it, and to separate concerns.

To accompany the SQLAlchemy adapter implementation, the patch includes integration tests for both SQLite and PostgreSQL. Let me know if corresponding utility resources should be added at different spots.

With kind regards,
Andreas.

### Software Tests

```console
docker compose --file libs/community/tests/integration_tests/document_loaders/docker-compose/postgresql.yml up
```

```console
cd libs/community
pip install psycopg2-binary
pytest -vvv tests/integration_tests -k sqldatabase
```

```
14 passed
```

image

### DIFF
--- a/libs/community/langchain_community/document_loaders/sql_database.py
+++ b/libs/community/langchain_community/document_loaders/sql_database.py
@@ -1,0 +1,138 @@
+from typing import Any, Callable, Dict, Iterator, List, Optional, Union
+
+import sqlalchemy as sa
+
+from langchain_community.docstore.document import Document
+from langchain_community.document_loaders.base import BaseLoader
+from langchain_community.utilities.sql_database import SQLDatabase
+
+
+class SQLDatabaseLoader(BaseLoader):
+    """
+    Load documents by querying database tables supported by SQLAlchemy.
+
+    For talking to the database, the document loader uses the `SQLDatabase`
+    utility from the LangChain integration toolkit.
+
+    Each document represents one row of the result.
+    """
+
+    def __init__(
+        self,
+        query: Union[str, sa.Select],
+        db: SQLDatabase,
+        parameters: Optional[Dict[str, Any]] = None,
+        page_content_mapper: Optional[Callable[..., str]] = None,
+        metadata_mapper: Optional[Callable[..., Dict[str, Any]]] = None,
+        source_columns: Optional[List[str]] = None,
+        include_rownum_into_metadata: bool = False,
+        include_query_into_metadata: bool = False,
+    ):
+        """
+        Args:
+            query: The query to execute.
+            db: A LangChain `SQLDatabase`, wrapping an SQLAlchemy engine.
+            sqlalchemy_kwargs: More keyword arguments for SQLAlchemy's `create_engine`.
+            parameters: Optional. Parameters to pass to the query.
+            page_content_mapper: Optional. Function to convert a row into a string
+              to use as the `page_content` of the document. By default, the loader
+              serializes the whole row into a string, including all columns.
+            metadata_mapper: Optional. Function to convert a row into a dictionary
+              to use as the `metadata` of the document. By default, no columns are
+              selected into the metadata dictionary.
+            source_columns: Optional. The names of the columns to use as the `source`
+              within the metadata dictionary.
+            include_rownum_into_metadata: Optional. Whether to include the row number
+              into the metadata dictionary. Default: False.
+            include_query_into_metadata: Optional. Whether to include the query
+              expression into the metadata dictionary. Default: False.
+        """
+        self.query = query
+        self.db: SQLDatabase = db
+        self.parameters = parameters or {}
+        self.page_content_mapper = (
+            page_content_mapper or self.page_content_default_mapper
+        )
+        self.metadata_mapper = metadata_mapper or self.metadata_default_mapper
+        self.source_columns = source_columns
+        self.include_rownum_into_metadata = include_rownum_into_metadata
+        self.include_query_into_metadata = include_query_into_metadata
+
+    def lazy_load(self) -> Iterator[Document]:
+        try:
+            import sqlalchemy as sa
+        except ImportError:
+            raise ImportError(
+                "Could not import sqlalchemy python package. "
+                "Please install it with `pip install sqlalchemy`."
+            )
+
+        # Querying in `cursor` fetch mode will return an SQLAlchemy `Result` instance.
+        result: sa.Result[Any]
+
+        # Invoke the database query.
+        if isinstance(self.query, sa.SelectBase):
+            result = self.db._execute(  # type: ignore[assignment]
+                self.query, fetch="cursor", parameters=self.parameters
+            )
+            query_sql = str(self.query.compile(bind=self.db._engine))
+        elif isinstance(self.query, str):
+            result = self.db._execute(  # type: ignore[assignment]
+                sa.text(self.query), fetch="cursor", parameters=self.parameters
+            )
+            query_sql = self.query
+        else:
+            raise TypeError(f"Unable to process query of unknown type: {self.query}")
+
+        # Iterate database result rows and generate list of documents.
+        for i, row in enumerate(result.mappings()):
+            page_content = self.page_content_mapper(row)
+            metadata = self.metadata_mapper(row)
+
+            if self.include_rownum_into_metadata:
+                metadata["row"] = i
+            if self.include_query_into_metadata:
+                metadata["query"] = query_sql
+
+            source_values = []
+            for column, value in row.items():
+                if self.source_columns and column in self.source_columns:
+                    source_values.append(value)
+            if source_values:
+                metadata["source"] = ",".join(source_values)
+
+            yield Document(page_content=page_content, metadata=metadata)
+
+    def load(self) -> List[Document]:
+        return list(self.lazy_load())
+
+    @staticmethod
+    def page_content_default_mapper(
+        row: sa.RowMapping, column_names: Optional[List[str]] = None
+    ) -> str:
+        """
+        A reasonable default function to convert a record into a "page content" string.
+        """
+        if column_names is None:
+            column_names = list(row.keys())
+        return "\n".join(
+            f"{column}: {value}"
+            for column, value in row.items()
+            if column in column_names
+        )
+
+    @staticmethod
+    def metadata_default_mapper(
+        row: sa.RowMapping, column_names: Optional[List[str]] = None
+    ) -> Dict[str, Any]:
+        """
+        A reasonable default function to convert a record into a "metadata" dictionary.
+        """
+        if column_names is None:
+            return {}
+
+        metadata: Dict[str, Any] = {}
+        for column, value in row.items():
+            if column in column_names:
+                metadata[column] = value
+        return metadata
--- a/libs/community/poetry.lock
+++ b/libs/community/poetry.lock
@@ -3140,6 +3140,7 @@ files = [
     {file = "jq-1.6.0-cp37-cp37m-musllinux_1_1_i686.whl", hash = "sha256:227b178b22a7f91ae88525810441791b1ca1fc71c86f03190911793be15cec3d"},
     {file = "jq-1.6.0-cp37-cp37m-musllinux_1_1_x86_64.whl", hash = "sha256:780eb6383fbae12afa819ef676fc93e1548ae4b076c004a393af26a04b460742"},
     {file = "jq-1.6.0-cp38-cp38-macosx_10_9_x86_64.whl", hash = "sha256:08ded6467f4ef89fec35b2bf310f210f8cd13fbd9d80e521500889edf8d22441"},
+    {file = "jq-1.6.0-cp38-cp38-macosx_11_0_arm64.whl", hash = "sha256:49e44ed677713f4115bd5bf2dbae23baa4cd503be350e12a1c1f506b0687848f"},
     {file = "jq-1.6.0-cp38-cp38-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:984f33862af285ad3e41e23179ac4795f1701822473e1a26bf87ff023e5a89ea"},
     {file = "jq-1.6.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:f42264fafc6166efb5611b5d4cb01058887d050a6c19334f6a3f8a13bb369df5"},
     {file = "jq-1.6.0-cp38-cp38-manylinux_2_5_i686.manylinux1_i686.manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:a67154f150aaf76cc1294032ed588436eb002097dd4fd1e283824bf753a05080"},
@@ -5457,8 +5458,6 @@ files = [
     {file = "psycopg2-2.9.9-cp310-cp310-win_amd64.whl", hash = "sha256:426f9f29bde126913a20a96ff8ce7d73fd8a216cfb323b1f04da402d452853c3"},
     {file = "psycopg2-2.9.9-cp311-cp311-win32.whl", hash = "sha256:ade01303ccf7ae12c356a5e10911c9e1c51136003a9a1d92f7aa9d010fb98372"},
     {file = "psycopg2-2.9.9-cp311-cp311-win_amd64.whl", hash = "sha256:121081ea2e76729acfb0673ff33755e8703d45e926e416cb59bae3a86c6a4981"},
-    {file = "psycopg2-2.9.9-cp312-cp312-win32.whl", hash = "sha256:d735786acc7dd25815e89cc4ad529a43af779db2e25aa7c626de864127e5a024"},
-    {file = "psycopg2-2.9.9-cp312-cp312-win_amd64.whl", hash = "sha256:a7653d00b732afb6fc597e29c50ad28087dcb4fbfb28e86092277a559ae4e693"},
     {file = "psycopg2-2.9.9-cp37-cp37m-win32.whl", hash = "sha256:5e0d98cade4f0e0304d7d6f25bbfbc5bd186e07b38eac65379309c4ca3193efa"},
     {file = "psycopg2-2.9.9-cp37-cp37m-win_amd64.whl", hash = "sha256:7e2dacf8b009a1c1e843b5213a87f7c544b2b042476ed7755be813eaf4e8347a"},
     {file = "psycopg2-2.9.9-cp38-cp38-win32.whl", hash = "sha256:ff432630e510709564c01dafdbe996cb552e0b9f3f065eb89bdce5bd31fabf4c"},
@@ -5501,7 +5500,6 @@ files = [
     {file = "psycopg2_binary-2.9.9-cp311-cp311-win32.whl", hash = "sha256:dc4926288b2a3e9fd7b50dc6a1909a13bbdadfc67d93f3374d984e56f885579d"},
     {file = "psycopg2_binary-2.9.9-cp311-cp311-win_amd64.whl", hash = "sha256:b76bedd166805480ab069612119ea636f5ab8f8771e640ae103e05a4aae3e417"},
     {file = "psycopg2_binary-2.9.9-cp312-cp312-macosx_10_9_x86_64.whl", hash = "sha256:8532fd6e6e2dc57bcb3bc90b079c60de896d2128c5d9d6f24a63875a95a088cf"},
-    {file = "psycopg2_binary-2.9.9-cp312-cp312-macosx_11_0_arm64.whl", hash = "sha256:b0605eaed3eb239e87df0d5e3c6489daae3f7388d455d0c0b4df899519c6a38d"},
     {file = "psycopg2_binary-2.9.9-cp312-cp312-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:8f8544b092a29a6ddd72f3556a9fcf249ec412e10ad28be6a0c0d948924f2212"},
     {file = "psycopg2_binary-2.9.9-cp312-cp312-manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:2d423c8d8a3c82d08fe8af900ad5b613ce3632a1249fd6a223941d0735fce493"},
     {file = "psycopg2_binary-2.9.9-cp312-cp312-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:2e5afae772c00980525f6d6ecf7cbca55676296b580c0e6abb407f15f3706996"},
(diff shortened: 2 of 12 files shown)


libs/community/tests/integration_tests/document_loaders/test_sqlalchemy_postgresql.py
```diff
@@ -0,0 +1,178 @@
...
+)
+
+
+@pytest.fixture
+@unittest.skipIf(not psycopg2_installed, "psycopg2 not installed")
+def engine() -> sa.Engine:
+    """
+    Return an SQLAlchemy engine object.
+    """
+    return sa.create_engine(CONNECTION_STRING, echo=False)
```
amotl, line 41: PostgreSQL-based tests will be skipped if the driver is not installed, which is not happening by default, right? Let me know if there is any means to invoke a PostgreSQL instance on CI, or if there is one available already. Otherwise, do you think it is good to go like it is?

amotl:
Hi @baskaryan,

thank you for merging our other patch so quickly. On this matter, may I humbly ask you to educate us about what to do with this `LangSmithUserError: API key must be provided` error?

```console
FAILED tests/unit_tests/callbacks/test_callback_manager.py::test_callback_manager_configure_context_vars - langsmith.utils.LangSmithUserError: API key must be provided when using hosted LangSmith API
```
-- https://github.com/langchain-ai/langchain/actions/runs/7578834938/job/20642075314?pr=16246#step:6:895

With kind regards,
Andreas.

libs/community/langchain_community/document_loaders/sqlalchemy.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __init__(
+        self,
+        query: Union[str, sa.Select],
+        url: str,
+        page_content_columns: Optional[List[str]] = None,
+        metadata_columns: Optional[List[str]] = None,
+        source_columns: Optional[List[str]] = None,
+        include_rownum_into_metadata: bool = False,
+        include_query_into_metadata: bool = False,
```
cbornet, line 23: I think it would be more powerful to be able to pass a function to convert a row to str for the page_content and a function to convert a row to Dict[str, Any] for the metadata.

amotl: That is an excellent suggestion, thank you Christophe. I will try to reshape this spot into the direction you've outlined.

amotl: Dear Christophe. I've amended the patch as you suggested, and just pushed the outcome. CI does not complain about it, so I will be happy to receive another review on it.

libs/community/poetry.lock
amotl, line 1: Given the traffic on this repository, `poetry.lock` will quickly see conflicts on patches touching it.

> This branch has conflicts that must be resolved
> Conflicting files `libs/community/poetry.lock`.

Is the best practice to just omit this file when sending patches, or is it just needed to rebase and refresh often? Please let me know how to handle this optimally. Thanks.

libs/community/langchain_community/document_loaders/sqlalchemy.py
```diff
@@ -0,0 +1,144 @@
...
+        self.query = query
+        self.url = url
+        self.page_content = page_content
+        self.metadata = metadata
+        self.source_columns = source_columns
+        self.include_rownum_into_metadata = include_rownum_into_metadata
+        self.include_query_into_metadata = include_query_into_metadata
+        self.sqlalchemy_kwargs = sqlalchemy_kwargs or {}
+
+    def load(self) -> List[Document]:
```
cbornet, line 57: I think the sqlalchemy resultset is an iterator. So it would be more powerful to implement `lazy_load`

libs/community/langchain_community/document_loaders/sqlalchemy.py
```diff
@@ -0,0 +1,144 @@
...
+                "Could not import sqlalchemy python package. "
+                "Please install it with `pip install sqlalchemy`."
+            )
+
+        engine = sa.create_engine(self.url, **self.sqlalchemy_kwargs)
+
+        docs = []
+        with engine.connect() as conn:
+            # Invoke the database query.
+            if isinstance(self.query, sa.Select):
```
cbornet, line 71:
```suggestion
            if isinstance(self.query, sa.SelectBase):
```

libs/community/langchain_community/document_loaders/sqlalchemy.py
```diff
@@ -0,0 +1,144 @@
...
+                result = conn.execute(self.query)
```
cbornet, line 72: It would be nice to be able to pass kwargs to the execute.
This way the user can use bound parameters.

libs/community/langchain_community/document_loaders/sqlalchemy.py
```diff
@@ -0,0 +1,144 @@
...
+    Load documents by querying database tables supported by SQLAlchemy.
+    Each document represents one row of the result.
+    """
+
+    def __init__(
+        self,
+        query: Union[str, sa.Select],
+        url: str,
+        page_content: Optional[Callable] = None,
+        metadata: Optional[Callable] = None,
```
cbornet, line 21: The name seems to suggest that we're passing the page_content and metadata themselves.
Can we use a name that shows that it's a function ? For instance `page_content_mapper` and `metadata_mapper`. Or something better if you have an idea.

libs/community/langchain_community/document_loaders/sqlalchemy.py
```diff
@@ -0,0 +1,144 @@
...
+                )
+
+            # Compute all column names.
+            all_column_names = list(result.mappings().keys())
+
+            # When no custom "page content" generator function
+            # is specified, use a reasonable default one.
+            if self.page_content is None:
+                page_content_fun = functools.partial(
+                    self.page_content_default, column_names=all_column_names
```
cbornet, line 89: It seems we get the same result by calling `self.page_content_default(row)`.
Maybe `self.page_content_default` could be the default value of the `page_content` parameter.

libs/community/langchain_community/document_loaders/sqlalchemy.py
cbornet, line 21: See comment below
```suggestion
        page_content: Optional[Callable] = self.page_content_default,
        metadata: Optional[Callable] = self.metadata_default,
```

libs/community/tests/integration_tests/document_loaders/test_sqlalchemy_sqlite.py
```diff
@@ -0,0 +1,190 @@
+"""
+Test SQLAlchemy/SQLite document loader functionality.
+"""
```
cbornet, line 3: It seems the tests are the same for sqlite and postgresql.
Maybe they could be factorized ?

docs/docs/modules/data_connection/document_loaders/sqlalchemy.mdx
```diff
@@ -0,0 +1,199 @@
+# SQLAlchemy
+
```
cbornet, line 2: Most document loaders are only documented in `integrations`.
Shouldn't this be the case for this one too ?

libs/community/langchain_community/document_loaders/sqlalchemy.py
amotl: Thank you for educating me about `lazy_load`. Implemented per 8920e2f80.

libs/community/langchain_community/document_loaders/sqlalchemy.py
amotl: Thanks. Adjusted per 98ccc13ec.

libs/community/langchain_community/document_loaders/sqlalchemy.py
amotl: Oh my, so true. df35ef7e9d significantly simplifies the implementation. Thank you so much.

docs/docs/modules/data_connection/document_loaders/sqlalchemy.mdx
amotl: 87cd992b916 absorbs the relevant documentation sections into the notebook, like you suggested.

libs/community/tests/integration_tests/document_loaders/test_sqlalchemy_sqlite.py
amotl: Working on it.

libs/community/langchain_community/document_loaders/sqlalchemy.py
amotl: Both functions are now used as default values, implementation is slightly different.

https://github.com/langchain-ai/langchain/blob/df35ef7e9dba209d481040b3bfe6c30de0a169d3/libs/community/langchain_community/document_loaders/sqlalchemy.py#L49-L52

libs/community/tests/integration_tests/document_loaders/test_sqlalchemy_sqlite.py
amotl: The test cases have been refactored accordingly on behalf of 1f21b47a3809. Thanks for your suggestion.

libs/community/langchain_community/document_loaders/sqlalchemy.py
amotl: Good idea. 4e522d77672 adjusts the implementation accordingly.

baskaryan (commented):
@amotl would it make for cleaner / more reusable code if we made use of `SQLDatabase` to handle all the sql connection and execution logic?

something like
```python
from langchain_community.utilities import SQLDatabase

class SQLDatabaseLoader(BaseLoader):
  def __init__(query: str, db: SQLDatabase, ...):
    ...
    
  def lazy_load() -> Iterable[Document]:
    results = db._execute(query)
    for result in results:
      yield _convert_to_doc(result)

def _convert_to_doc(result: ...) -> Document:
  ...

**Comment by amotl** [comment]
@baskaryan: Sounds good, but we'd need to extend `SQLDatabase` to also accept parameters on `_execute`, like @cbornet suggested at https://github.com/langchain-ai/langchain/pull/16246#discussion_r1466119894, right? Will you accept corresponding adjustments?

File: libs/community/langchain_community/document_loaders/sqldatabase.py
Comment (baskaryan) [inline libs/community/langchain_community/document_loaders/sqldatabase.py:1]: last nit, could we name this and the notebook sql_database to match the utilities file? or you can give me write access and i can do it (https://python.langchain.com/docs/contributing/faq#how-do-i-allow-maintainers-to-edit-my-pr)

otherwise lgtm!

File: libs/community/langchain_community/utilities/sql_database.py
Comment (amotl) [inline libs/community/langchain_community/utilities/sql_database.py:1]: Dear @baskaryan,

I've submitted this patch separately, on behalf of GH-16655. I think it will be a good idea to accompany it by corresponding test cases.

Currently, the linter croaks when expanding the new fetch/parameters interface also to the `run` method, so it will need a bit more to be sorted out.

With kind regards,
Andreas.

**Review by cbornet — APPROVED** [review]
LGTM

File: libs/community/langchain_community/utilities/sql_database.py
Comment (amotl) [reply libs/community/langchain_community/utilities/sql_database.py:1]: Hi again. GH-16655 now also gained a dedicated documentation page about relevant features of the `SQLDatabase` utility we improved along the lines, see [sql_database.ipynb].

I think it makes sense to bring that patch in before coming back here. Thus, it has been toggled back into draft mode. After coming back from yak shaving, I will rebase this patch, and squash the individual commits before presenting it for review again, if you agree with this procedure.

[sql_database.ipynb]: https://github.com/langchain-ai/langchain/blob/96cfc1f3/docs/docs/integrations/tools/sql_database.ipynb

File: libs/community/langchain_community/document_loaders/sqldatabase.py
Comment (amotl) [reply libs/community/langchain_community/document_loaders/sqldatabase.py:1]: Sure I will address this nit, when coming back from GH-16655. Thank you.

File: libs/community/langchain_community/document_loaders/sqldatabase.py [hunk 10]
```diff
@@ -0,0 +1,135 @@
...
+
+    For talking to the database, the document loader uses the `SQLDatabase`
+    utility from the LangChain integration toolkit.
+
+    Each document represents one row of the result.
+    """
+
+    def __init__(
+        self,
+        query: Union[str, sa.Select],
```
eyurtsev, line 22: I'm reviewing this PR: https://github.com/langchain-ai/langchain/pull/16655 , which implements a cursor and I think will be used by this data loader for loading documents from SQL.

I've mostly used keyset pagination in the past because it's easier to add retry logic to it in case of network connections or some other server side errors (as well as parallelization).  Is there  way to do retries with the cursor if the connection times out or we get disconnected for whatever reason?

Key set pagination: https://www.cockroachlabs.com/docs/stable/pagination#keyset-pagination

Basic idea is to combine a sort on an id field, a filter on the id filed and a limit.

libs/community/langchain_community/utilities/sql_database.py
amotl: Hi. @eyurtsev integrated GH-16655 on behalf of GH-17191. Thank you very much. I will get back to refreshing this PR soon, and will let you know about it.

libs/community/langchain_community/document_loaders/sqldatabase.py
amotl: I've renamed the files to `sql_database`, and renewed your invitation.

-- https://github.com/crate-workbench/langchain/invitations

libs/community/langchain_community/document_loaders/sqldatabase.py
amotl: Hi @eyurtsev. I hear your argument, but at the same time, I think the Cockroach documentation is misleading.

Allocating resources for a cursor once, and then using it to churn through the results, optionally also in async/streaming mode, is far more efficient than sending multiple independent database queries which need to allocate corresponding resources on the server side each and every time.

Looking at this from a different perspective, it probably depends very much on the use case at hand. It think using key set pagination is out of scope for this patch, as it depends on an id field, and some database tables may not have it.

If such functionality is needed, it can probably implemented on top, or orthogonal to this infrastructure code, which is merely foundational.

libs/community/langchain_community/utilities/sql_database.py
amotl: Hi again. I've just refreshed this patch, and CI signals readiness, so it might be good to go if @eyurtsev doesn't have any objections?

libs/community/langchain_community/document_loaders/sqldatabase.py
amotl: >> I've mostly used keyset pagination in the past.
>
> If such functionality is needed, it can probably implemented on top, or orthogonal to this infrastructure code, which is merely foundational.

Hi @eyurtsev,

let me know if you think differently, or if you agree with this assessment. If you are using this technique, and would like to see any sort of adapter for it, I think it may be valuable to implement. However, I don't see how it can be implemented within the scope of this patch.

With kind regards,
Andreas.

eyurtsev: I'll try to review / merge today if not then earliest on Monday -- a bit pressed on time and I want to review carefully :)

I agree with your points -- cursor based pagination is likely good fit for most users and will result in a simpler API.

If they're working at scale, they'll likely need to do additional optimizations regardless and may want to implement their own solution regardless.

----

fwiw keyset pagination is a general pagination technique (isn't a cockroachdb thing) -- depending on infra and use case it can be used to parallelize reads easily for larger scales of records (~say 10M records), and it's fault tolerant because subsets of pages can be retried independently

amotl: Thanks Eugene, please take your time. Also thanks for educating me about keyset pagination, I like the idea and it is definitively worth to consider when needing to serve data at scale.

Probably a bit OT, so we might want to continue discussing at another spot, but you sparked my interest ;]:
Is keyset pagination also used when needing to page through datasets which are larger-than-memory, so that processing a potential large resultset in response to a query will not trip memory exhaustion both server- and client-side?

amotl: > [...] so that processing a potential large resultset in response to a query will not trip memory exhaustion both server- and client-side.

On this very matter, I think it will be sweet to implement server-side cursor support, when applicable. PostgreSQL can do it, and CrateDB implemented it recently, also for that very purpose. See [async_streaming.py](https://github.com/crate/cratedb-examples/blob/main/by-language/python-sqlalchemy/async_streaming.py).

libs/community/langchain_community/document_loaders/sql_database.py
```diff
@@ -0,0 +1,138 @@
...
+    utility from the LangChain integration toolkit.
+
+    Each document represents one row of the result.
+    """
+
+    def __init__(
+        self,
+        query: Union[str, sa.Select],
+        db: SQLDatabase,
+        parameters: Optional[Dict[str, Any]] = None,
```
eyurtsev, line 24: Could we add `*` after db, to force named argument invocation for all optionals (it helps  to minimize breaking changes during future refactors of the API)

libs/community/langchain_community/document_loaders/sql_database.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class SQLDatabaseLoader(BaseLoader):
+    """
+    Load documents by querying database tables supported by SQLAlchemy.
+
+    For talking to the database, the document loader uses the `SQLDatabase`
+    utility from the LangChain integration toolkit.
+
+    Each document represents one row of the result.
```
eyurtsev, line 17: Would you be willing to add a ..code-block python example that shows how to use the loader?

It'll populate the API reference with a nice python formatted example:  https://api.python.langchain.com/en/v0.0.349/api_reference.html

libs/community/langchain_community/document_loaders/sql_database.py
```diff
@@ -0,0 +1,138 @@
...
+        include_rownum_into_metadata: bool = False,
+        include_query_into_metadata: bool = False,
+    ):
+        """
+        Args:
+            query: The query to execute.
+            db: A LangChain `SQLDatabase`, wrapping an SQLAlchemy engine.
+            sqlalchemy_kwargs: More keyword arguments for SQLAlchemy's `create_engine`.
+            parameters: Optional. Parameters to pass to the query.
+            page_content_mapper: Optional. Function to convert a row into a string
```
eyurtsev, line 37: Is a row a dict or a tuple? Any chance we could update the type signature on page_content_mapping and metadata_mapper with the argument type if it's known?

With the type information missing and no example python snippet -- usage of these parameters is non obvious from documentation

libs/community/langchain_community/document_loaders/sql_database.py
```diff
@@ -0,0 +1,138 @@
...
+              to use as the `page_content` of the document. By default, the loader
+              serializes the whole row into a string, including all columns.
+            metadata_mapper: Optional. Function to convert a row into a dictionary
+              to use as the `metadata` of the document. By default, no columns are
+              selected into the metadata dictionary.
+            source_columns: Optional. The names of the columns to use as the `source`
```
eyurtsev, line 43: What are source columns?

libs/community/langchain_community/document_loaders/sql_database.py
```diff
@@ -0,0 +1,138 @@
...
+              within the metadata dictionary.
+            include_rownum_into_metadata: Optional. Whether to include the row number
```
eyurtsev, line 45: row_num or rownum? (not sure if intentional)

libs/community/langchain_community/document_loaders/sql_database.py
```diff
@@ -0,0 +1,138 @@
...
+        self.parameters = parameters or {}
+        self.page_content_mapper = (
+            page_content_mapper or self.page_content_default_mapper
+        )
+        self.metadata_mapper = metadata_mapper or self.metadata_default_mapper
+        self.source_columns = source_columns
+        self.include_rownum_into_metadata = include_rownum_into_metadata
+        self.include_query_into_metadata = include_query_into_metadata
+
+    def lazy_load(self) -> Iterator[Document]:
```
eyurtsev, line 61: !! LAZY LOAD !! LET'S GO!!

libs/community/langchain_community/document_loaders/sql_database.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    def __init__(
+        self,
+        query: Union[str, sa.Select],
+        db: SQLDatabase,
+        parameters: Optional[Dict[str, Any]] = None,
+        page_content_mapper: Optional[Callable[..., str]] = None,
+        metadata_mapper: Optional[Callable[..., Dict[str, Any]]] = None,
+        source_columns: Optional[List[str]] = None,
```
eyurtsev, line 27: (not important) I love using Sequence to cover more valid types (e.g., tuple) + denote the promise that the container won't be mutated

libs/community/tests/integration_tests/document_loaders/docker-compose/postgresql.yml
```diff
@@ -0,0 +1,19 @@
+version: "3"
+
+services:
+  postgresql:
+    image: postgres:16
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
+    ports:
+      - "5432:5432"
```
eyurtsev, line 9: Thank you! I want to move these compose files to root folder under docker so they are easy to find and are re-usable across tests that use the same services. (I can move it later in a separate PR it can stay right here for now)

eyurtsev, line 9: If you have any cycles, It'll be good to some password here and use a non standard port; I'm worried that at some point some poor user will lose their existing dev database due to a coincidence

libs/community/tests/integration_tests/document_loaders/test_sql_database.py
```diff
@@ -0,0 +1,233 @@
...
+    with engine.connect() as connection:
+        connection.execute(sa.text("DROP TABLE IF EXISTS mlb_teams_2012;"))
+        for statement in sqlparse.split(sql_statements):
+            connection.execute(sa.text(statement))
+            connection.commit()
+
+
+tmpdir = TemporaryDirectory()
+
+
```
eyurtsev, line 71: should we use a fixture with module scope and clean up after the test end?

libs/community/tests/integration_tests/document_loaders/test_sql_database.py
```diff
@@ -0,0 +1,233 @@
...
+    from _pytest.python import Metafunc
+
+from langchain_community.document_loaders.sql_database import SQLDatabaseLoader
+from tests.data import MLB_TEAMS_2012_SQL
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+try:
+    import sqlite3  # noqa: F401
```
eyurtsev, line 27: what do you think about using: 

pytest.skip(reason="the reason", alllow_module_level=True) 

to skip the unit tests and provide a reason?

libs/community/tests/integration_tests/document_loaders/test_sql_database.py
```diff
@@ -0,0 +1,233 @@
...
+@pytest.fixture()
+def provision_database(engine: sa.Engine) -> None:
+    """
+    Provision database with table schema and data.
+    """
+    sql_statements = MLB_TEAMS_2012_SQL.read_text()
+    with engine.connect() as connection:
+        connection.execute(sa.text("DROP TABLE IF EXISTS mlb_teams_2012;"))
+        for statement in sqlparse.split(sql_statements):
+            connection.execute(sa.text(statement))
```
eyurtsev, line 65: What do you think about either using execute with multi=True or else using a flaky split on `;` instead of using sqlparse to avoid picking up an extra dependency (even if test only dep)

eyurtsev (commented):
Looks great to me! A few minor comments and we should merge

libs/community/langchain_community/document_loaders/sqldatabase.py
eyurtsev: Apologies took a bit longer than planned. PR Looks great!  I left a few nits!

Also wanted to thank @cbornet for his review

amotl: Hi Eugene. Thanks for your review. I will be afk for hiking starting tomorrow, so I may be able to get back to this only on Monday ff. If you have some spare cycles, feel free to take over any time. I don't have any objections about your suggestions. Otherwise, see you next week.

eyurtsev:
@amotl Let me know if you'd like me to comandeer to incorporate some of the nits.


https://github.com/langchain-ai/langchain/pull/16246#discussion_r1487139999

amotl:
Hi @eyurtsev. While I feel bad about it, if you have the cycles, it would be so nice, indeed! I am currently a bit swamped with documentation improvement matters on our ends, where others are waiting for, so I will appreciate your efforts very much.

eyurtsev:
Comandeering

libs/community/langchain_community/document_loaders/sql_database.py
eyurtsev: keeping as rownum

libs/community/tests/integration_tests/document_loaders/docker-compose/postgresql.yml
eyurtsev: Moving this to root level docker

eyurtsev: added password

eyurtsev:
Closing in favor of: https://github.com/langchain-ai/langchain/pull/18281

PR #16255, opened 2024-01-19
core: fix: remove comments in json strings.

lichengwu:
Description: This change removes single line // and multi-line /* */ comments from JSON strings using a regex substitution. The purpose of this update is to ensure that JSON content can be parsed correctly without being affected by the presence of comments, which are not supported in standard JSON format. The regex substitution is performed using the re.sub method with the pattern r'//.*?\n|/\*.*?\*/' and the flags=re.S to ensure proper handling of newline characters within the comments.

### DIFF
--- a/docs/docs/integrations/document_loaders/spider.ipynb
+++ b/docs/docs/integrations/document_loaders/spider.ipynb
@@ -1,0 +1,95 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Spider\n",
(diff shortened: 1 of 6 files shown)


lichengwu:
@hwchase17

baskaryan:
thanks @lichengwu! could we add another test case for having // inside a string

lichengwu:
> thanks @lichengwu! could we add another test case for having // inside a string

When I was using the agent function, I noticed that some models generate context that includes JSON comments. Therefore, I expanded the testing scope based on the existing method. Are you suggesting that I should write a new pytest testing method?  @baskaryan

lichengwu:
What else needs to be done to merge this request?

lichengwu:
@baskaryan  @hwchase17  Can this code be merged？

baskaryan:
@lichengwu seems to be failing tests

PR #16273, opened 2024-01-19
[Fix] Fix Cassandra Document loader default page content mapper

cbornet:
We can't use `json.dumps` by default as many types returned by the cassandra driver are not serializable. It's safer to use `str` and let users define their own custom `page_content_mapper` if needed.


cbornet:
@baskaryan @eyurtsev Hi, could this get reviewed ?

baskaryan (approved):

PR #15909, opened 2024-01-11
community: SQLStrStore/SQLDocStore provide an easy SQL alternative to `InMemoryStore` to persist data remotely in a SQL storage

gcheron:
**Description:**

- Implement `SQLStrStore` and `SQLDocStore` classes that inherits from `BaseStore` to allow to persist data remotely on a SQL server.
- SQL is widely used and sometimes we do not want to install a caching solution like Redis.
- Multiple issues/comments complain that there is no easy remote and persistent solution that are not in memory (users want to replace InMemoryStore), e.g., https://github.com/langchain-ai/langchain/issues/14267, https://github.com/langchain-ai/langchain/issues/15633, https://github.com/langchain-ai/langchain/issues/14643, https://stackoverflow.com/questions/77385587/persist-parentdocumentretriever-of-langchain
- This is particularly painful when wanting to use `ParentDocumentRetriever `
- This implementation is particularly useful when:
     * it's expensive to construct an InMemoryDocstore/dict
     * you want to retrieve documents from remote sources
     * you just want to reuse existing objects
- This implementation integrates well with PGVector, indeed, when using PGVector, you already have a SQL instance running. `SQLDocStore` is a convenient way of using this instance to store documents associated to vectors. An integration example with ParentDocumentRetriever and PGVector is provided in docs/docs/integrations/stores/sql.ipynb or [here](https://github.com/gcheron/langchain/blob/sql-store/docs/docs/integrations/stores/sql.ipynb).
- It persists `str` and `Document` objects but can be easily extended.

 **Issue:**

Provide an easy SQL alternative to `InMemoryStore`.

PR #16278, opened 2024-01-19
Add SQL Docstore

pprados:
**Description:**
The storage package propose 
- Redis
- AstraDB

I propose to add an SQLStore.
@baskaryan

### DIFF
--- a/libs/community/langchain_community/storage/sql_docstore.py
+++ b/libs/community/langchain_community/storage/sql_docstore.py
@@ -1,0 +1,236 @@
+import contextlib
+from pathlib import Path
+from typing import (
+    Any,
+    Dict,
+    Generator,
+    Iterator,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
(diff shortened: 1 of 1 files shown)


cbornet:
Similar PR: https://github.com/langchain-ai/langchain/pull/15909
I think it would be good to have only one consistent implementation.

pprados:
Another [PR](https://github.com/langchain-ai/langchain/pull/15909) is merged

pprados:
@efriis 
I reopen this PR, because the [PR15909](https://github.com/langchain-ai/langchain/pull/15909) presents some problems.

efriis (commented):
Hey @pprados ! I think this implementation has some similar issues, and it also looks incomplete (lots of commented out code).

The main issue with a SQL store is that it may need db migrations down the line, so very inconspicuous changes can result in large breaking changes for people using it.

Would you be open to completing the implementation in a template instead, so people can pull it locally and deal with those migrations themselves?

PR #16295, opened 2024-01-19
core[patch]: preserve inspect.iscoroutinefunction with @deprecated decorator

piotrm0:
Adjusted `deprecate` decorator to make sure decorated async functions are still recognized as "coroutinefunction" by `inspect`. 

Before change, functions such as `LLMChain.acall` which are decorated as deprecated are not recognized as coroutine functions. After the change, they are recognized:

```python
import inspect
from langchain import LLMChain

# Is false before change but true after.
inspect.iscoroutinefunction(LLMChain.acall)
```

### DIFF
--- a/libs/core/langchain_core/_api/deprecation.py
+++ b/libs/core/langchain_core/_api/deprecation.py
@@ -144,6 +144,15 @@ def deprecated(
                 emit_warning()
             return wrapped(*args, **kwargs)
 
+        async def awarning_emitting_wrapper(*args: Any, **kwargs: Any) -> Any:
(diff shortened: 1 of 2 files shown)


eyurtsev:
Thanks @piotrm0, any chance you're willing to add a unit test?

@baskaryan we should do the same for the @beta decorator.

piotrm0:
> Thanks @piotrm0, any chance you're willing to add a unit test?

Added unit tests.

eyurtsev:
@piotrm0 we can finish the linting if you want :)

we usually use `make format` to do auto linting

PR #16298, opened 2024-01-19
Update grobid.py

naarkhoo:
there is a case where "coords" does not exist in the "sentence" therefore, the "split(";")" will lead to error. 

we can fix that by adding "if sentence.get("coords") is not None:" 

the resulting empty "sbboxes" from this scenario will raise error at "sbboxes[0]["page"]" because sbboxes are empty.

the PDF from https://pubmed.ncbi.nlm.nih.gov/23970373/ can replicate those errors.


baskaryan:
thanks @naarkhoo!

PR #16316, opened 2024-01-20
community: Fix error message when litellm is not installed

jeremi:
The error message was mentioning the wrong package. I updated it to the correct one.


baskaryan (approved):
thanks @jeremi!

PR #16325, opened 2024-01-21
core: absolute `EXAMPLE_DIR` path

jamesbraza:
If you invoked testing from places besides `core/`, this `EXAMPLE_DIR` path won't work. This PR makes`EXAMPLE_DIR` robust against invocation location


baskaryan:
thanks @jamesbraza!

PR #16327, opened 2024-01-21
core: removed `packaging` dependency

jamesbraza:
The `packaging` dependency was only used in `check_package_version`. Based on all use cases of `check_package_version`, the use of `packaging` wasn't actually needed.

This PR removes `packaging` as a dependency, and recompiles dependencies using `poetry lock`


eyurtsev (changes requested):
Hi @jamesbraza it looks like this PR introduces a bug since it's relying on string comparisons which are lexical rather than on tuple comparisons which will be numeric

So this code will incorrectly conclude that `0.1.2` > `0.1.12`

jamesbraza:
Oh nice catch @eyurtsev, thanks. I went ahead and opened https://github.com/langchain-ai/langchain/pull/16387 to add unit tests for this, so this PR could have failed CI. Closing this one out

Issue #16334, opened 2024-01-21
Is `llm.bind(stop=["\nObservation"])` really meanful in `json_chat` agnet?


calvinweb:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

-

### Description

https://github.com/langchain-ai/langchain/blob/ef75bb63ce5cc4fb76ba1631ebe582f56103ab7e/libs/langchain/langchain/agents/json_chat/base.py#L151
this seems useless because chat-json depends json to work,not this sort of completion

### System Info

-

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [X] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

steve-carpenter:
This is also breaking certain LLMs (i.e. using Bedrock w/ Llama2 or Titan these _preconfigured_ stop sequences aren't allowed. 

`
Stop sequence key name for meta is not supported.`

Sanzid88888:
` Stop sequence key name for meta is not supported     I am also getting this error any update @steve-carpenter 
`

calvinweb:
> `Stop sequence key name for meta is not supported ` I am also getting this error using the Bedrock meta-model. Could you let me know if you have any solution? @steve-carpenter

pass `stop_sequence=False` when constructing the agent

Sanzid88888:
How can I determine which models support stop sequences and which don't? @calvinweb

mikamboo:
``I'm using crewai and mot of models failed

Only claud model worked form me, others gave me errors related :

* `[stop_sequences] is not permitted`  
* `Malformed input request: : string [Observation] does not match pattern ^(\|+|User:)`


This worked for me:

```python
from langchain_community.chat_models import BedrockChat

llm = BedrockChat(model_id="anthropic.claude-3-haiku-20240307-v1:0")
```

PR #16335, opened 2024-01-21
Langchain: `json_chat` don't need stop sequenes

lin-calvin:
This is a PR about #16334
The Stop sequenes isn't meanful in `json_chat` because it depends json to work, not completions

### DIFF
--- a/libs/langchain/langchain/agents/json_chat/base.py
+++ b/libs/langchain/langchain/agents/json_chat/base.py
@@ -148,7 +148,6 @@ def create_json_chat_agent(
         tools=render_text_description(list(tools)),
         tool_names=", ".join([t.name for t in tools]),
     )
-    llm_with_stop = llm.bind(stop=["\nObservation"])
 
     agent = (
         RunnablePassthrough.assign(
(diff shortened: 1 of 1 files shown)


steve-carpenter:
FYI looks like you removed `llm_with_stop` but still reference it when defining the agent.

hwchase17:
this generally is meaningful
im happy to make this optional to better support other models (seems like that was an issue)
lets maybe add a boolean flag for this, and dcoument it well?

calvinweb:
> this generally is meaningful im happy to make this optional to better support other models (seems like that was an issue) lets maybe add a boolean flag for this, and dcoument it well?

But json_chat is depends on json,not the "Observation" string

calvinweb:
> FYI looks like you removed `llm_with_stop` but still reference it when defining the agent.

Sry, forget to test the code before push

hwchase17:
oftentimes the model hallucinates an Observation. we need this on by default

calvinweb:
> oftentimes the model hallucinates an Observation. we need this on by default

So add a flag is the best soluation?

hwchase17:
yup

calvinweb:
> yup

Edited,or should I fire a new PR?

hwchase17 (approved):

PR #16391, opened 2024-01-22
Add async methods for the AstraDB VectorStore

cbornet:
- **Description**: fully async versions are available for astrapy 0.7+. For older astrapy versions or if the user provides a sync client without an async one, the async methods will call the sync ones wrapped in `run_in_executor`
  - **Twitter handle:** cbornet_


cbornet:
cc @hemidactylus

hemidactylus:
LGTM, great work!

hwchase17 (commented):
thanks!

Issue #16396, opened 2024-01-22
Make all Tool names snake_case


baskaryan:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

OpenAI function-calling doesn't support function names with spaces. Need to update all Tool names to be snake_cased so that they work as OpenAI functions by default. See #16395 for example fix.

PR #16397, opened 2024-01-22
Fixed tool names snake_case

jatinchawda1503:
#16396
Fixed
1. golden_query
2. google_lens
3. memorize
4. merriam_webster
5. open_weather_map
6. pub_med
7. stack_exchange
8. generate_image
9. wikipedia


Lord-Haji:
It makes sense to standardize but I don't see a point to make a PR this short. This doesn't fix anything the only thing causing issues was spaces in the tool names

jatinchawda1503:
@Lord-Haji, Thanks for review. So do I revert the changes?
I have validated the names, but I don't see any tool names with space. 
Ideally, I thought standardizing it would make more sense to avoid future errors when someone is creating a tool while referring to this tool's code.

baskaryan:
I believe duckduckgo and google search tools have spaces and need to be fixed. but we can also include the current changes, most tools have snake_cased names and it'd be nice to be consistent about that.

jatinchawda1503:
thanks @baskaryan, Sure I will fix that too.

baskaryan (approved):

baskaryan:
thank you @jatinchawda1503!

Issue #10744, opened 2023-09-18
Cannot use LangChain Tools with GPT4ALL and LLaMA model


N3RDIUM:
### System Info

LangChain 0.0.292
Python 3.11.5
GPT4ALL with LLAMA q4_0 3b model running on CPU

### Who can help?

@agola11

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [X] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

My code is:
```from langchain.llms import GPT4All
from langchain.callbacks.streaming_stdout import StreamingStdOutCallbackHandler
from langchain.agents import AgentType, initialize_agent
from langchain.tools import Tool
from pydantic import BaseModel, Field

class JokeInput(BaseModel):
    confidence: float = Field(default=0.0)

import pyjokes

import langchain
langchain.debug = True
import os
os.environ["LANGCHAIN_TRACING"] = "true" # If you want to trace the execution of the program, set to "true"

# There are many CallbackHandlers supported, such as
# from langchain.callbacks.streamlit import StreamlitCallbackHandler
callbacks = [StreamingStdOutCallbackHandler()]
model = GPT4All(model=".cache/llama-2-7b-chat.ggmlv3.q4_0.bin", n_threads=6)

# Generate text. Tokens are streamed through the callback manager.
tools = []
tools.append(
    Tool.from_function(
        func=pyjokes.get_joke,
        name="Joke",
        description="useful for when you need to tell the user a joke",
        args_schema=JokeInput
    )
)

# Init the agent
agent = initialize_agent(
    tools, model, agent=AgentType.CHAT_ZERO_SHOT_REACT_DESCRIPTION, verbose=True
)

# Start the agent
while True:
    user_prompt = input("You: ")
    agent(user_prompt, callbacks=callbacks)```
```

### Expected behavior

It should use the Joke tool and return a joke, but instead it gives me this error:
```
Found model file at  .cache/llama-2-7b-chat.ggmlv3.q4_0.bin
[...]

PR #16401, opened 2024-01-22
Docs: Update import library for StreamlitCallbackHandler

Hadi2525:
- **Description:** Some code sources have been moved from `langchain` to `langchain_community` and so the documentation is not yet up-to-date. This is specifically true for `StreamlitCallbackHandler` which returns a `warning` message if not loaded from `langchain_community`., 
  - **Issue:** I don't see a # issue that could address this problem but perhaps #10744,
  - **Dependencies:** Since it's a documentation change no dependencies are required


baskaryan (approved):
thanks @Hadi2525!

Issue #4710, opened 2023-05-15
How to use return_source_documents to also extract similarity score??


virdi16:
I was using RetrievalQA.from_chain_type, to which I had passed parameters as:-

`RetrievalQA.from_chain_type(llm, chain_type, retriever = chroma_db.as_retriever(), return_source_documents = True)`

Here, 
return_source_documents = True, only returns the chunks from which it generated the response. _Is there a way in which I can get similarity score also returned for matched chunks_ (say if there are 4 chunks it found most relevant to query, how to get scores in decreasing order based on similarity)

imeckr:
The QA-Chains doesn't seem to have any direct way to get the `similarity_score` along with the source documents. You can directly call the `chromadb` for the scores though.

```
chromadb.similarity_search_with_score(your_query)
```

FrancescoSaverioZuppichini:
I am also looking for return scores for a normal qa chain

OmiWakode:
Same here. Is there any way to get scores from a chain

freyherme:
+1 on this.  I'd love to have a way to get the similarity scores from a chain.

FrancescoSaverioZuppichini:
I end up doing something like this 

```python
from typing import List

from langchain.schema import Document
from langchain.vectorstores.redis import RedisVectorStoreRetriever


class RedisVectorStoreRetrieverWithScores(RedisVectorStoreRetriever):
    """
    Hacky way to create a redis retriever that adds the score to the metadata
    """

    def get_relevant_documents(self, query: str) -> List[Document]:
        # [NOTE] we removed the search type, only use search_type = "similarity"
        if self.search_type != "similarity":
            raise ValueError(f"Only search_type='similarity' is supported with scores")
        docs_and_scores = self.vectorstore.similarity_search_with_score(query, k=self.k)
        for doc, score in docs_and_scores:
            doc.metadata = {**doc.metadata, **{"score": score}}
        return [doc for (doc, _) in docs_and_scores]

```

Knight1997:
for weaviate DB it's simple and i think it would be similar for others too.

just pass the additional params as seach_kwargs in retriever.
```
Weaviate().as_retriever(search_kwargs={"additional": ["vector", "certainty", "id"]})
```

This will give you vector, certainty score and uuid of documents under metadata of Documents

roy-firman:
I am also looking for a solution, but in my case I am using pgvectore as my vectorstore, any idea here?

[...]

PR #16405, opened 2024-01-22
RFC: rich retrieval results

baskaryan:
Could be more opinionated about DocumentResult and RetrievalResult attributes if we wanted

Related to #4710

### DIFF
--- a/libs/core/langchain_core/retrievers.py
+++ b/libs/core/langchain_core/retrievers.py
@@ -3,10 +3,12 @@ from __future__ import annotations
 import warnings
 from abc import ABC, abstractmethod
 from inspect import signature
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union, overload
 
 from langchain_core.documents import Document
+from langchain_core.load import Serializable
 from langchain_core.load.dump import dumpd
+from langchain_core.pydantic_v1 import Field
 from langchain_core.runnables import (
     Runnable,
     RunnableConfig,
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/retrievers.py
```diff
@@ -22,8 +24,19 @@
         Callbacks,
     )
 
+
+class DocumentResult(Serializable):
+    document: Document
+    metadata: dict = Field(default_factory=dict)
+
+
+class RetrievalResult(Serializable):
```
eyurtsev, line 33: This API works IMO

libs/core/langchain_core/retrievers.py
```diff
@@ -22,8 +24,19 @@
         Callbacks,
     )
 
+
```
eyurtsev, line 27: Do we have examples of the metadata that folks want to add?

I would consider potentially adding some standard metadata the root level to force an identical schema between providers

```python
class DocumentResult(Serializable):
    document: Document
    extra: dict = Field(default_factory=dict)
    id: Optional[str]
    score: Optional[float]   
```

eyurtsev: In many search engines, what's returned is not the Document, but a hit snippet. A snippet is a chunk of text from the original document that matched.


```python
class Hit(Serializable):
    source: Optional[Document] # <-- The actual source of the document 
    snippet: ... # <-- Some type we don't know yet that might represent a chunk of the document
    extra: dict = Field(default_factory=dict)
    id: Optional[str]
    score: Optional[float]   
```

libs/core/langchain_core/retrievers.py
baskaryan: any argument to be made that we should explicitly include `query: str`?

PR #16408, opened 2024-01-22
community:PDF doc loaders: Fix filter detection for images

suelopoder:
**Description:** Fixes PDF with images parsing
**Issue:** PDF parsing works but detecting the proper image filter does not
**Twitter handle:** @suelopoder


baskaryan:
thank you @suelopoder! would it be possible to add a simple unit test?

suelopoder:
Oh, my bad. I'll work on that

PR #16435, opened 2024-01-23
chage _merge_splits total count

codinseok:
Description: Modified the _merge_splits function within langchain.text_splitter to ensure accurate length calculation of text chunks when using CharacterTextSplitter with from_huggingface_tokenizer and a custom length function. Previously, the function was generating shorter text chunks than expected, especially noticeable when working with Korean text documents. This discrepancy was due to the total variable in _merge_splits accumulating lengths of individual documents, which could differ significantly from the tokenizer's calculated length (in some cases with Korean text, the difference was over two-fold). To ensure the accuracy of chunk lengths, the function was modified to recalculate the total length using _length_function for each addition of a new chunk, despite the additional computational time required.

### DIFF
--- a/libs/langchain/langchain/text_splitter.py
+++ b/libs/langchain/langchain/text_splitter.py
@@ -177,6 +177,8 @@ class TextSplitter(BaseDocumentTransformer, ABC):
         current_doc: List[str] = []
         total = 0
         for d in splits:
+            current_text = self._join_docs(current_doc, separator)
+            total = self._length_function(current_text)
             _len = self._length_function(d)
             if (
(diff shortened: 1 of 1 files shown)


baskaryan:
@codinseok could we add some unit tests for this?

codinseok:
@baskaryan 
of course! I'd appreciate it if you could try it and share the results

ccurme:
Closing this PR but will re-open if we add tests-- specifically, need a short example that reproduces the issue so we can assess the fix and weigh it against the increased computation burden.

PR #16436, opened 2024-01-23
neo4j-generation template: replace openai legacy function, return proper chainable

tewnut:
- create_structured_output_chain is a legacy function, replaced with create_structured_output_runnable
  - the chain returned is an invalid function, use proper LCEL. 
  - the prompt is kept intact, only change how allow_nodes and allow_rels are fed into the prompt.
  - make add_graph_documents a RunnableLamda


efriis:
cc @tomasonjo

baskaryan:
closing in favor of #19024 which uses LLMGraphTransformer. appreciate the contribution anyways @tewnut!

PR #16450, opened 2024-01-23
Fix sql record manager

pprados:
** Description**

The current implementation of SQLRecordManager crashes with a database other than sqlite. Because of a schema bug. Sqlite is tolerant.

The correct type for group_id is Integer.

**Issue**
https://github.com/langchain-ai/langchain/issues/16451


eyurtsev:
The schema and the API for group id are designed to take strings rather than integers. Could you explain what the bug is? Is the issue that the API is forgiving and accepts an integer with from the user without raising an error?

eyurtsev:
I can double check in a bit, but I think postgres supports membership checks with strings. 

Syntax should uses `(` rather than square brackets.

eyurtsev:
@pprados Is the related to fixing the sql record manager? It looks like it contains an in memory implementation of the record manager?

pprados:
I'm sorry, I mixed 2 PR. Forget my PR.

PR #13910, opened 2023-11-27
Pprados/rag vectorstore

pprados:
**Description:**
The `ParentDocumentRetriever` class provides the `add_documents()` methods, which may seem unconventional for a retriever.

This deviation arises from the need to establish and maintain connections between document fragments and their various iterations. Consequently, the `parent_splitter` and `child_splitter` were introduced. As elaborated in this [pull request](https://github.com/langchain-ai/langchain/pull/11968), it's crucial not to confine oneself to text splitters exclusively. Instead, a preference is given to working with *transformations*, from which text splitters inherit.

However, my [initial proposal](https://github.com/langchain-ai/langchain/pull/11968) proved insufficient to fully leverage this concept with all the advanced features of Langchain. An [issue](https://github.com/langchain-ai/langchain/issues/11983) has been raised to address this limitation.

To enhance text selection, the proposed idea by `ParentDocumentRetriever` is as follows:
- Convert each text fragment into multiple versions to enhance proximity between questions and potential answers.
- Select corresponding fragments in the vector store.
- Return the original fragment for injection into the prompt.

For this purpose, maintaining a link between the original document, each document chunk, and every transformation 
applied to each chunk is essential. 

**My Proposition:**
The `RagVectorStore` class serves as a vector store wrapper with rules for transforming a document 
into a chunk and each chunk into different versions. It closely resembles `ParentRetriever` but employs 
a `VectorStore` instead of a `Retriever`.

This approach is advantageous due to its simplicity and compatibility with all advanced features of Langchain. 
Feel free to explore the benefits of this solution in [this notebook](https://github.com/pprados/langchain-rag/blob/master/docs/integrations/vectorstores/rag_vectorstore.ipynb).

Compatibility includes:
- [Indexing API](https://python.langchain.com/docs/modules/data_connection/indexing)
- Advanced retrievers like [MultiQueryRetriever](https://python.langchain.com/docs/modules/data_connection/retrievers/MultiQueryRetriever), [Self Querying](https://python.langchain.com/docs/modules/data_connection/retrievers/self_query), etc.

The implementation required several changes and additional pull requests, each of which can be validated independently before merging into this one:
- [Adds an in-memory implementation of RecordStore](https://github.com/langchain-ai/langchain/pull/13200)
- [Add SQLDocStore](https://github.com/langchain-ai/langchain/pull/13181)
- [Add a Wrapper vectorstore, compatible with SelfQueryRetriever](https://github.com/langchain-ai/langchain/pull/13190)
[...]

PR #16454, opened 2024-01-23
Add VectorStore wrapper to help the integration with SelfQuery

pprados:
**Description:** 
The [`SelfQueryRetriever`](https://python.langchain.com/docs/modules/data_connection/retrievers/self_query/) component is able to convert filter clauses for different vectorstore implementations.
If we wish to offer advanced use of vectorstore, by modifying for example certain methods, via a wrapper type approach, we lose the ability to use `SelfQueryRetriever`.

We propose to standardize vectorstore wrapper exploitation, to allow flexible integration with `SelfQueryRetriever`. 
The idea is to add to `_get_builtin_translator()`, taking into account the wrapper. 
Thus, every vectorstore wrapper automatically benefits from the SelfQueryRetriever.

This is the prelude to another full-request, which needs a vectorstore wrapper, while being compatible 
with `SelfQueryRetriever`.

**Tag maintainer:**
@baskaryan

**Twitter handle:**
@pprados


hwchase17:
you can always pass in the translator:

```
@root_validator(pre=True)
    def validate_translator(cls, values: Dict) -> Dict:
        """Validate translator."""
        if "structured_query_translator" not in values:
            values["structured_query_translator"] = _get_builtin_translator(
                values["vectorstore"]
            )
        return values
```

so if you are using a wrapper, you can just pass in the relevant translator. that is probably easier way to do it

pprados:
@hwchase17 
I do not share this analysis. A wrapper is precisely designed not to see any knowledge of the wrapped VS. Likewise, the use of this wrapper must be agnostic of the wrapped VS.
This wrapper is essentially there, to allow `SelfQueryRetriever` to automatically adjust, as it does for the different DBs it manages. In addition, this can be used for other uses, like my other proposal of a [`RAGVectorStore`](https://github.com/langchain-ai/langchain/pull/13910), which is a wrapper equivalent to `ParentDocumentRetriever`, but, in my opinion, better localized in the framework, in order to be able to get all the benefit from it, like compatibility with `index()`.

A generic wrapper can be used for many other things.

Issue #16458, opened 2024-01-23
JsonOutputParser does not throw exception on invalid json


ksachdeva:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

```python
SAMPLE_JSON_OUTPUT_FROM_LLM_1:str  = json.dumps({
    "name" : "Henry",
    "age" : 25
})

INVALID_JSON_STRING = SAMPLE_JSON_OUTPUT_FROM_LLM_1.replace("\"n", "n")

# Value of INVALID_JSON_STRING is '{name": "Henry", "age": 25}'

from langchain.output_parsers.json import SimpleJsonOutputParser

json_op = SimpleJsonOutputParser()

result = json_op.parse(INVALID_JSON_STRING)

# result is {} whereas I was expecting an error/exception

```

### Description

I was trying to test both the positive and negative cases for JsonOutpuParser

In the above code snippet you can see that I removed the " on the first key

I debugged the langchain code, it seems that there is a lot of effort to ignore the invalid/troubling characters and that leads to changing the original string to {}

I would have expected following behavior:

- By default, no trying to fix the provided string
- May be throw an error if the output is {}

### System Info

pip freeze | grep langchain
langchain==0.1.2
langchain-community==0.0.14
langchain-core==0.1.14
langchain-openai==0.0.3
langchainhub==0.1.14

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [X] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

moorej-oci:
It looks like this part is the issue: 
https://github.com/langchain-ai/langchain/blob/ef6a33557079c8c691fb18c78d7c89c2d2e63888/libs/core/langchain_core/output_parsers/json.py#L106-L120

A part of my recreation in a jupyter notebook is below. It iteratively trims the string until it can successfully load 
```
# Try to parse mods of string until we succeed or run out of characters.
while new_s:
    final_s = new_s

    # Close any remaining open structures in the reverse
    # order that they were opened.
[...]

PR #16480, opened 2024-01-23
CORE: JsonOutputParser bug fixed.

HazSyl1:
JsonOutputParser does not throw exception on invalid json , instead it returns an empty dictionary. It was suggested to make it throw an error by default in case  of an invalid json. An argument 'partial' is added , in case the user prefers the empty dictionary result.
#16458
Twitter/X:Hazxhx


libs/core/langchain_core/output_parsers/json.py
```diff
@@ -202,19 +202,19 @@ def _diff(self, prev: Optional[Any], next: Any) -> Any:
     def parse_result(self, result: List[Generation], *, partial: bool = False) -> Any:
         text = result[0].text
         text = text.strip()
-        if partial:
```
baskaryan, line 205: this is needed for streaming json, since streamed chunks of json will often be invalid. the behavior of returning None in this case is intentional

baskaryan:
closing for now, let me know if you're still seeing an issue @HazSyl1

PR #16483, opened 2024-01-24
openai[patch]: accept function_call dict in bind_functions

baskaryan:
Confusing that you can't pass in a dict


efriis (approved):

efriis (approved):

efriis:
Got it - so this also fixes a bug where "auto"/"none" wouldn't work with multiple funcs (but None always did I think)?

PR #16485, opened 2024-01-24
core[patch], community[patch], openai[patch]: consolidate openai tool…

baskaryan:
… converters

One way to convert anything to an OAI function: convert_to_openai_function
One way to convert anything to an OAI tool: convert_to_openai_tool
Corresponding bind functions on OAI models: bind_functions, bind_tools


libs/core/langchain_core/utils/function_calling.py
```diff
@@ -188,15 +250,38 @@ def convert_to_openai_function(
...
         return convert_python_function_to_openai_function(function)
-
     else:
         raise ValueError(
             f"Unsupported function type {type(function)}. Functions must be passed in"
             f" as Dict, pydantic.BaseModel, or Callable."
         )
+
+
+def convert_to_openai_tool(
```
hinthornw, line 270: Should we add a 'Runnable' support while we're at it?

baskaryan: how would we convert it?

baskaryan: using input_schema and docstrings?

efriis (approved):

PR #16487, opened 2024-01-24
community: Fix MlflowCallback with none artifacts_dir

serena-ruan:


serena-ruan:
Hi @baskaryan Sorry this is a small bug fix just captured, could you merge it? Thanks!

PR #16497, opened 2024-01-24
community: Fixed the 'aembed' method of 'CohereEmbeddings'.

abhi922:
**Description:**
  - The existing code was trying to find a `.embeddings` property on the `Coroutine` returned by calling `cohere.async_client.embed`.
  - Instead, the `.embeddings` property is present on the value returned by the `Coroutine`.
  - Also, it seems that the original cohere client expects a value of `max_retries` to not be `None`. Hence, setting the default value of `max_retries` to  `3`.


baskaryan (approved):
thank you @abhi922!

PR #16520, opened 2024-01-24
community: fix create_sql_agent prefix

dalssoft:
- **Description:** fix create_sql_agent prefix and suffix when `prefix` & `suffix` arguments are nor None.
  - **Issue:** 
  - **Dependencies:** 
  - **Twitter handle:** @dalssoft

### DIFF
--- a/libs/community/langchain_community/agent_toolkits/sql/base.py
+++ b/libs/community/langchain_community/agent_toolkits/sql/base.py
@@ -160,10 +160,10 @@ def create_sql_agent(
             )
             template = "\n\n".join(
                 [
-                    react_prompt.PREFIX,
+                    prefix or react_prompt.PREFIX,
                     "{tools}",
                     format_instructions,
-                    react_prompt.SUFFIX,
+                    suffix or react_prompt.SUFFIX,
                 ]
             )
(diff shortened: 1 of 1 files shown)


eyurtsev:
Hi @dalssoft I don't think I understand this fix. When the arguments are None the `or` is going to default to the react prompt anyway. Would you mind updating the description of the PR title to explain what is being fixed?

dalssoft:
The problem occurred because it was ignoring prefix or suffix when calling `create_sql_agent`.

Rational:
if you pass a `prefix` as string it will go to this [line](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/agent_toolkits/sql/base.py#L134) but does not create a prompt.

so when on this [`if`](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/agent_toolkits/sql/base.py#L155), it will create a new prompt with defaults but ignore the `prefix` (or `suffix`).

Maybe a better strategy is to create a prompt in the first place instead of what I'm suggesting in this PR.

ccurme:
Believe this was fixed in https://github.com/langchain-ai/langchain/pull/23693. Apologies as this PR was earlier. Let me know if issues remain and can re-open.

PR #16526, opened 2024-01-24
langchain: Adding a new section aware splitter to langchain

msetbar:
- **Description:** the layout of html pages can be variant based on the bootstrap framework or the styles of the pages. So we need to have a splitter to transform the html tags to a proper layout and then split the html content based on the provided list of tags to determine its html sections. We are using BS4 library along with xslt structure to split the html content using an section aware approach.
  - **Twitter handle:** @m_setayesh

Please make sure your PR is passing linting and testing before submitting. Run `make format`, `make lint` and `make test` from the root of the package you've modified to check this locally.

See contribution guidelines for more information on how to write/run tests, lint, etc: https://python.langchain.com/docs/contributing/

If you're adding a new integration, please include:

 -->


libs/langchain/langchain/document_transformers/xsl/converting_to_header.xslt
```diff
@@ -0,0 +1,29 @@
...
+    </xsl:copy>
+  </xsl:template>
+
+  <!-- Match any element that has a font-size attribute larger than 20px -->
+  <xsl:template match="*[@style[contains(., 'font-size')]]">
+    <!-- Extract the font size value from the style attribute -->
+    <xsl:variable name="font-size" select="substring-before(substring-after(@style, 'font-size:'), 'px')" />
+    <!-- Check if the font size is larger than 20 -->
+    <xsl:choose>
+      <xsl:when test="$font-size > 20">
```
butlfrazp, line 15: do we want to have this hard coded or maybe convert to a variable (if possible)

libs/langchain/langchain/text_splitter.py
```diff
@@ -23,6 +23,7 @@
 
 import copy
 import logging
+import os
```
butlfrazp, line 26: nit: do we need this here?

libs/langchain/langchain/text_splitter.py
```diff
@@ -1486,3 +1485,161 @@ def __init__(self, **kwargs: Any) -> None:
...
+        for doc in documents:
+            texts.append(doc.page_content)
+            metadatas.append(doc.metadata)
+        results = self.create_documents(texts, metadatas=metadatas)
+
+        text_splitter = RecursiveCharacterTextSplitter(**self.kwargs)
+
+        return text_splitter.split_documents(results)
+
+    def split_text(self, text: str) -> List[Document]:
```
butlfrazp, line 1527: it seems like the behavior of split documents and split text is different. Do we want this?

butlfrazp (commented):
Looks good overall, just a few comments

libs/langchain/langchain/text_splitter.py
msetbar: no, Paul they are same but since the split_text does not have meta data, we do not have the title metadata. so we need to keep title

libs/langchain/langchain/text_splitter.py
msetbar: yes needed for isabs function.

libs/langchain/langchain/document_transformers/xsl/converting_to_header.xslt
msetbar: the class has exposed a parameter so that any user can have their own xslt file to transfer as their need. this is just an example of how to use it.

arinbjornk:
@msetbar This PR seems related to this issue: https://github.com/langchain-ai/langchain/issues/20644

Issue #16528, opened 2024-01-24
BaseTool __init_subclass__ has contradictory conditional


arnob-sengupta:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/tools.py#L105
```python
class BaseTool(RunnableSerializable[Union[str, Dict], Any]):
    """Interface LangChain tools must implement."""

    def __init_subclass__(cls, **kwargs: Any) -> None:
        """Create the definition of the new tool class."""
        super().__init_subclass__(**kwargs)

        args_schema_type = cls.__annotations__.get("args_schema", None)

        if args_schema_type is not None:
            if args_schema_type is None or args_schema_type == BaseModel:
                # Throw errors for common mis-annotations.
                # TODO: Use get_args / get_origin and fully
                # specify valid annotations.
                typehint_mandate = """
class ChildTool(BaseTool):
    ...
    args_schema: Type[BaseModel] = SchemaClass
    ..."""
                name = cls.__name__
                raise SchemaAnnotationError(
                    f"Tool definition for {name} must include valid type annotations"
                    f" for argument 'args_schema' to behave as expected.\n"
                    f"Expected annotation of 'Type[BaseModel]'"
                    f" but got '{args_schema_type}'.\n"
                    f"Expected class looks like:\n"
                    f"{typehint_mandate}"
                )
```

### Description

This code first checks if args_schema_type is not None, and then immediately checks if it is None, which doesn't make logical sense.

It should be:
```
        if args_schema_type is not None and args_schema_type == BaseModel:
```

Also, perhaps the TODO should have an issue associated with it, though I'm not sure I have enough context to create the issue.

### System Info

```
System Information
------------------
> OS:  Windows
> OS Version:  10.0.19045
> Python Version:  3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 0.1.10
> langchain: 0.0.353
> langchain_community: 0.0.12
> langserve: Not Found
```

PR #16530, opened 2024-01-24
core: consolidate conditional in BaseTool

arnob-sengupta:
- **Description:** Refactor contradictory conditional to single line
  - **Issue:** #16528

### DIFF
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -116,10 +116,10 @@ class BaseTool(RunnableSerializable[Union[str, Dict], Any]):
             # TODO: Use get_args / get_origin and fully
             # specify valid annotations.
             typehint_mandate = """
-class ChildTool(BaseTool):
-...
-args_schema: Type[BaseModel] = SchemaClass
-..."""
+    class ChildTool(BaseTool):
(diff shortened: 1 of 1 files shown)


baskaryan (approved):
good catch, thanks @arnob-sengupta!

libs/core/langchain_core/tools.py
```diff
@@ -111,25 +111,24 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
-                name = cls.__name__
-                raise SchemaAnnotationError(
-                    f"Tool definition for {name} must include valid type annotations"
-                    f" for argument 'args_schema' to behave as expected.\n"
-                    f"Expected annotation of 'Type[BaseModel]'"
-                    f" but got '{args_schema_type}'.\n"
-                    f"Expected class looks like:\n"
-                    f"{typehint_mandate}"
-                )
+...
```
eyurtsev, line 120: Would you mind restoring the leading whitespaces in the class, so the error raised in the exception is a bit nicer for an end user?

arnob-sengupta: Sure, will do

PR #16536, opened 2024-01-24
RFC: unpack .assign() arg

baskaryan:
suppose you have two chains that output dicts. second takes output of first as input. you want to return combined dict of their outputsr.

imagine second_chain outputs {"bar": ..., "baz": ...} and you want those to be in top level of output.
currently you'd do
```python
def unpack(x):
  foo = x.pop("foo")
  return {**x, **foo}

chain.assign(foo=second_chain) | unpack
```
i would like to do

```python
chain.assign(second_chain)
```

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -452,6 +452,10 @@ class Runnable(Generic[Input, Output], ABC):
 
     def assign(
         self,
+        *args: Union[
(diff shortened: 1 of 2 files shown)


eyurtsev:
The difference between these two snippets seems too subtle to me.

```python
chain.assign(foo=second_chain) 
```

```python
chain.assign(second_chain)
```
-----------

Is there any way to make it work with a named argument that would specify how to merge the results?

PR #16539, opened 2024-01-24
Voyage update

satyam-jha:


libs/community/langchain_community/embeddings/voyageai.py
```diff
@@ -1,63 +1,20 @@
...
-    stop_after_attempt,
-    wait_exponential,
-)
+from langchain_core.pydantic_v1 import BaseModel, Extra, root_validator
+from langchain_core.utils import get_from_dict_or_env
 
 logger = logging.getLogger(__name__)
 
 
-def _create_retry_decorator(embeddings: VoyageEmbeddings) -> Callable[[Any], Any]:
```
baskaryan, line 30: for backwards compatibility b/n old versions of langchain and new versions of community we should keep these variables around
```python
_check_response=None
_create_retry_decorator=None
embed_with_retry=None
```

PR #16564, opened 2024-01-25
Error when importing packages from pydantic [docs]

ARKA1112:
URL : https://python.langchain.com/docs/use_cases/extraction

Desc: 
<b> While the following statement executes successfully, it throws an error which is described below when we use the imported packages</b>
 ```py 
from pydantic import BaseModel, Field, validator
```
Code: 
```python
from langchain.output_parsers import PydanticOutputParser
from langchain.prompts import (
    PromptTemplate,
)
from langchain_openai import OpenAI
from pydantic import BaseModel, Field, validator

# Define your desired data structure.
class Joke(BaseModel):

    # You can add custom validation logic easily with Pydantic.
    @validator("setup")
    def question_ends_with_question_mark(cls, field):
        if field[-1] != "?":
            raise ValueError("Badly formed question!")
        return field
```

Error:
```md
PydanticUserError: The `field` and `config` parameters are not available in Pydantic V2, please use the `info` parameter instead.

For further information visit https://errors.pydantic.dev/2.5/u/validator-field-config-info 
```

Solution:
Instead of doing:
```py
from pydantic import BaseModel, Field, validator
```
We should do:
```py
from langchain_core.pydantic_v1 import BaseModel, Field, validator
```
Thanks.


baskaryan (commented):
good catch @ARKA1112! actually we should do `from langchain_core.pydantic_v1 import ...` so that it works for folks with pydantic v1 or v2 installed. could we update to that?

see more on pydantic compatibility here https://python.langchain.com/docs/guides/pydantic_compatibility

ARKA1112:
@baskaryan  Thanks for the review, Sure we can update to that, I will modify the code.

baskaryan (approved):

PR #16569, opened 2024-01-25
Streaming Fix: Callback Manager Now Shows first chunk of input on start event for streaming runs

Daan-Grashoff:
Fixed an issue in the streaming feature where the callback manager was not displaying the input, as it was previously defined as an empty string. By updating the callback manager to use `{"input": final_input}`, we can now track the inputs properly.


eyurtsev:
Hi @Daan-Grashoff, thanks for the contribution! I think we need a slightly larger change to make this work. The issue is that the `final_input` may not be known within a streaming pipeline, and will instead end up being equal to only the first input chunk, which would lead to a strange API for stream events.

I think the change that we need to make is to determine whether the input is complete or incomplete before we can update the callback handler

Daan-Grashoff:
Hey @eyurtsev, Not a problem! I depend on on_chain_start, and without modifying it, the input remains consistently empty. I also observed that when using something like StrOutputParser, only the initial segment is displayed. However, perhaps having some input is preferable to none at all?

eyurtsev:
Got it! I'll prioritize a full fix for this (obviously you're welcome to try as well). I am hesitant releasing a partial fix as it'll create confusion for other users.

As separate item, we could also make sure the corresponding input chunk is included in the stream event (not only the output stream). This would also folks to update both streaming inputs and outputs for a particular step in the chain.

Also could you explain how your logic depends on on_chain_start? Would using the run_id to update the inputs when you see on_chain_end help you at all in the meantime?

Daan-Grashoff:
Yep totally get it! I'll use my forked version in the meantime, and will wait on the full fix for this :)

Also totally missed that! updating the input on on_chain_end will also fix the issue (I think)! Thanks for the quick insight.

eyurtsev:
@luong-komorebi i haven't worked on it yet, didn't seem like it was affecting many users. Could you tell me how you're bumping into this issue and why using the run_id associated with the events isn't enough for your use case (in terms of matching the inputs and the outputs)?

PR #16580, opened 2024-01-25
langchain:Fix streaming error of TongYi model 

Ca11back:
- **Description:**
 Fix streaming error of TongYi model (Allow merge_dicts to merge int value)
  - **Issue:** the issue
 When using Tongyi or ChatTongyi with streaming in the [use case](https://python.langchain.com/docs/use_cases/question_answering/streaming),the following error occured:
`TypeError: Additional kwargs key output_tokens already exists in left dict and value has unsupported type <class 'int'>`
  - **Dependencies:**
 libs/core/langchain_core/utils/_merge.py

### DIFF
--- a/libs/core/langchain_core/utils/_merge.py
+++ b/libs/core/langchain_core/utils/_merge.py
@@ -36,6 +36,8 @@ def merge_dicts(left: Dict[str, Any], right: Dict[str, Any]) -> Dict[str, Any]:
             merged[k] = merge_dicts(merged[k], v)
         elif isinstance(merged[k], list):
             merged[k] = merged[k] + v
+        elif isinstance(merged[k], int):
+            merged[k] = merged[k] + v
         else:
             raise TypeError(
                 f"Additional kwargs key {k} already exists in left dict and value has "


eyurtsev:
`ints` were not made addable by default intentionally since the integer should placed into one of the chunks. This may be an 
incorrect assumption though, could you share the raw streaming response from TongYi to show what integer is being added? We want to make sure that an additive behavior makes sense for it, and we'd definitely want to unit-test that after the fact. utility is used by all streaming code so it's really critical

Ca11back:
> `ints` were not made addable by default intentionally since the integer should placed into one of the chunks. This may be an incorrect assumption though, could you share the raw streaming response from TongYi to show what integer is being added? We want to make sure that an additive behavior makes sense for it, and we'd definitely want to unit-test that after the fact. utility is used by all streaming code so it's really critical

These are values need to be merged when the error occured:
```
k = 'output_tokens'
left = {'input_tokens': 530, 'output_tokens': 2, 'total_tokens': 532}
merged = {'input_tokens': 530, 'output_tokens': 2, 'total_tokens': 532}
right = {'input_tokens': 530, 'output_tokens': 13, 'total_tokens': 543}
v = 13
```

jiangyd:
Has the version been released yet

Issue #16579, opened 2024-01-25
[BUG] Inconsistent results with `RecursiveCharacterTextSplitter`'s `add_start_index=True`


antoniolanza1996:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

```python
from langchain.text_splitter import RecursiveCharacterTextSplitter
from langchain.docstore.document import Document

[...]

PR #16583, opened 2024-01-25
fix: inconsistent results with `RecursiveCharacterTextSplitter`'s `add_start_index=True`

antoniolanza1996:
This PR fixes issue #16579


baskaryan (commented):
@antoniolanza1996 could we add a unit test for this

antoniolanza1996:
@baskaryan done. The new test case fails if executed with the current `master` implementation. Instead, should be ok with the current implementation.

Unfortunately it always depends on what type of text splitter we are using... however with the implemented approach it's likely that we won't find a substring "before" the real one because we are trying to apply the `find` with an higher offset then before.

baskaryan (approved):
awesome, thank you @antoniolanza1996!

PR #16605, opened 2024-01-26
langchain:Fix streaming error of TongYi model

Ca11back:
- **Description:**
 Fix streaming error of TongYi model (make merge_dicts to replace int value)
  - **Issue:**
 When using Tongyi or ChatTongyi with streaming in the [use case](https://python.langchain.com/docs/use_cases/question_answering/streaming),the following error occured:
`TypeError: Additional kwargs key output_tokens already exists in left dict and value has unsupported type <class 'int'>`

These are values need to be merged when the error occurred:
```
k = 'output_tokens'
left = {'input_tokens': 530, 'output_tokens': 2, 'total_tokens': 532}
merged = {'input_tokens': 530, 'output_tokens': 2, 'total_tokens': 532}
right = {'input_tokens': 530, 'output_tokens': 13, 'total_tokens': 543}
v = 13
```
The first out put has 2 tokens, the second has eleven. Seems the int should be replaced with new value?
```
{'answer': 'Task decomposition'}
{'answer': ' is the process of breaking down a complex task into smaller'}
```
Original PR:
 #16580

  - **Dependencies:**
 libs/core/langchain_core/utils/_merge.py

### DIFF
--- a/libs/core/langchain_core/utils/_merge.py
+++ b/libs/core/langchain_core/utils/_merge.py
@@ -36,6 +36,8 @@ def merge_dicts(left: Dict[str, Any], right: Dict[str, Any]) -> Dict[str, Any]:
             merged[k] = merge_dicts(merged[k], v)
         elif isinstance(merged[k], list):
             merged[k] = merged[k] + v
+        elif isinstance(merged[k], int):
+            merged[k] = v
         else:
             raise TypeError(
                 f"Additional kwargs key {k} already exists in left dict and value has "


baskaryan (commented):
we intentionally don't support ints as it's not clear how to merge them. in this case you would just be overwriting past ints. a better solution would be to update the dicts that are streaming to return something mergeable, like a list of ints

GYiyouth (approved):
need this

GYiyouth (approved):

baskaryan:
> this doesn't generalize to all ints in all dicts. the solution shoudl be specific to the TongYi model (either turn tokens into a list or drop it after the first chunk)

closing per ^

Neutrino1998:
This fix would also work: modify the `_generation_from_qwen_resp` method in `langchain_community.llms.tongyi`:
```
    @staticmethod
    def _generation_from_qwen_resp(resp: Any) -> Dict[str, Any]:
        # 将token_usage的值转换为描述性字符串
        token_usage_str = ", ".join(f"{k}: {v}" for k, v in resp["usage"].items())
        return dict(
            text=resp["output"]["text"],
            generation_info=dict(
                finish_reason=resp["output"]["finish_reason"],
                request_id=resp["request_id"],
                # token_usage=dict(resp["usage"]),
                # 使用字符串代替原先的字典，以避免合并时的类型冲突
                token_usage=token_usage_str,
            ),
        )
```

PR #16609, opened 2024-01-26
Langchain: Allow AgentAction to take in a list as input.

lambrou:
- **Description** Add `list` to the accepted types of the `tool_input` field in the `AgentAction` class

### DIFF
--- a/libs/core/langchain_core/agents.py
+++ b/libs/core/langchain_core/agents.py
@@ -17,7 +17,7 @@ class AgentAction(Serializable):
 
     tool: str
     """The name of the Tool to execute."""
-    tool_input: Union[str, dict]
+    tool_input: Union[str, list, dict]
     """The input to pass in to the Tool."""
     log: str
     """Additional information to log about the action.


eyurtsev (changes requested):
@lambrou thanks for the contribution! 

This is core schema, so we don't make changes to it lightly. To add more types, an explanation about why a list type makes sense is needed.

If the motivation makes sense, we'd need unit test to cover the appropriate scenario to make sure it works with existing agent executor etc.

lambrou:
@eyurtsev Hey there! Use case:
I am the maintainer of an application that runs bash commands using Langchain. I want my LLM to be able to send lists of commands. (ie [''mkdir newproject', "cd new project"]
This allows me to cleanly run each command and not have to worry about the LLM sending multiple bash commands with newlines/semicolons.

I will write the unit tests and commit them to this branch if you think this motivation makes sense. Thank you!

hwchase17:
> @eyurtsev Hey there! Use case: I am the maintainer of an application that runs bash commands using Langchain. I want my LLM to be able to send lists of commands. (ie [''mkdir newproject', "cd newproject"] This allows me to cleanly run each command and not have to worry about the LLM sending multiple bash commands with newlines/semicolons.
> 
> I will write the unit tests and commit them to this branch if you think this motivation makes sense. Thank you!

can those just be represented as separate Actions (agents can return a list of actions) or as a key in the dictionary (some agent types now always return dictionarys)

lambrou:
> > @eyurtsev Hey there! Use case: I am the maintainer of an application that runs bash commands using Langchain. I want my LLM to be able to send lists of commands. (ie [''mkdir newproject', "cd newproject"] This allows me to cleanly run each command and not have to worry about the LLM sending multiple bash commands with newlines/semicolons.
> > I will write the unit tests and commit them to this branch if you think this motivation makes sense. Thank you!
> 
> can those just be represented as separate Actions (agents can return a list of actions) or as a key in the dictionary (some agent types now always return dictionarys)

Sure, this can be closed.

PR #16628, opened 2024-01-26
Pydantic replaced method schema with the model_json_schema

Yanni8:
See: https://docs.pydantic.dev/2.4/migration/#changes-to-pydanticbasemodel

### DIFF
--- a/libs/langchain/langchain/chains/openai_functions/extraction.py
+++ b/libs/langchain/langchain/chains/openai_functions/extraction.py
@@ -101,7 +101,7 @@ def create_extraction_chain_pydantic(
     class PydanticSchema(BaseModel):
         info: List[pydantic_schema]  # type: ignore
 
-    openai_schema = pydantic_schema.schema()
+    openai_schema = pydantic_schema.model_json_schema()
     openai_schema = _resolve_schema_references(
         openai_schema, openai_schema.get("definitions", {})
     )
--- a/libs/langchain/langchain/chains/openai_functions/tagging.py
+++ b/libs/langchain/langchain/chains/openai_functions/tagging.py
@@ -75,7 +75,7 @@ def create_tagging_chain_pydantic(
     Returns:
         Chain (LLMChain) that can be used to extract information from a passage.
     """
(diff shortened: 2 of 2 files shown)


eyurtsev (changes requested):
Thanks for the contribution! LangChain has to support both pydantic 1 and pydantic 2 right now, so this will break pydnatic 1 support. I may be wrong, but I think that pydantic marked these for deprecation but hasn't actually deprecated the methods so they will work.

If that's not correct, we can make the change as long as its preceded by a conditional that checks which pydantic version is being used (needs to differentiate between (1) pydantic 1,  (2) pydantic 2 and (3) pydantic.v1 namespace in pydantic 2).

ekalosak:
This hits hard when trying to use the `JsonOutputParser` class - it barfs if you try to give it a v2 object. I wonder if a try/catch pattern around using the `model_json_schema` call wouldn't be appropriate here? Then e.g. langchain can issue deprecation warnings upon catching the attribute error upon that phase of the pydantic upgrade rollout.

ekalosak:
I can verify that `schema()` is supported, with a deprecation warning, in v2. Please see this minimal example:
```py
from pydantic import BaseModel  # <- v2 namespace

class A(BaseModel):
    a: int

print(A.schema())
```
produces this stderr/stdout:
```
test.py:11: PydanticDeprecatedSince20: The `schema` method is deprecated; use `model_json_schema` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.5/migration/
  print(A.schema())
{'properties': {'a': {'title': 'A', 'type': 'integer'}}, 'required': ['a'], 'title': 'A', 'type': 'object'}
```

eyurtsev:
Yeah if we want to upgrade to model_json_schema() the code needs a conditional based on pydantic major version to retain backwards compatibility with pydantic version 1

eyurtsev:
Closing this PR for now -- feel free to create another PR that introduces the logic with a conditional change.

Issue #15997, opened 2024-01-14
The `dict` method is deprecated; use `model_dump` instead [pydantic]


lucas-labs:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

The followiwng simple example will log the warning:

```python
from langchain_openai import AzureChatOpenAI
from langchain.schema import HumanMessage

chat_llm = AzureChatOpenAI()

chat_llm.invoke([HumanMessage(content='howdy!')])
```

### Description

I'm seeing a few warning logs saying that langchain-openai is using a deprecated method from pydantic: `dict`

```txt
.venv/Lib/site-packages/langchain_openai/chat_models/azure.py:199: PydanticDeprecatedSince20: The `dict` method is deprecated; use `model_dump` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.5/migration/
    response = response.dict()

tests/test_zllm_completion.py::test_prompt
  .venv/Lib/site-packages/langchain_openai/llms/base.py:344: PydanticDeprecatedSince20: The `dict` method is deprecated; use `model_dump` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.5/migration/
    response = response.dict()
```

Here:

https://github.com/langchain-ai/langchain/blob/7b084b4cc768a8e5cddc29bfdfb6e5dc1003a5d9/libs/partners/openai/langchain_openai/chat_models/azure.py#L199

And here:

https://github.com/langchain-ai/langchain/blob/7b084b4cc768a8e5cddc29bfdfb6e5dc1003a5d9/libs/partners/openai/langchain_openai/llms/base.py#L344

(There might me other cases)

### System Info

langchain==0.1.0
Microsoft Windows 11 Pro
Python 3.11.3

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

[...]

PR #16629, opened 2024-01-26
partners: replace deprecated .dict() with .model_dump() in openai base.py

SavvasMohito:
- **Description:** Same as #15489 but under the correct package this time. Using pydantic's new `.model_dump()` function. It produces the same result as `.dict()` without throwing a deprecation warning message, 
  - **Issue:** #15997 ,
  - **Dependencies:** None,
  - **Twitter handle:** @mantzouranidis

Thanks :)

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -407,7 +407,7 @@ class ChatOpenAI(BaseChatModel):
         default_chunk_class = AIMessageChunk
         for chunk in self.client.create(messages=message_dicts, **params):
             if not isinstance(chunk, dict):
-                chunk = chunk.dict()
+                chunk = chunk.model_dump()
             if len(chunk["choices"]) == 0:
                 continue
             choice = chunk["choices"][0]
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -24,7 +24,6 @@ from typing import (
 )
 
 import openai
-import pydantic
 import tiktoken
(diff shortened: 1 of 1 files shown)


eyurtsev:
LangChain needs to be both pydantic 1 and 2 compatible for now.

Running tests expecting the tests to fail. If they don't fail we should confirm why not.

SavvasMohito:
@eyurtsev The tests do fail as you suspected. I made an additional commit to satisfy both v1 and v2 (and 3) but also avoid the deprecation warnings. Local tests do not throw any errors on my end

SavvasMohito:
so no merge @baskaryan? 😔

sam-cohan:
When do we expect this to be solved, it is peppering our code with lots of warnings...

sam-cohan:
btw, Ideally the PR should cover all instances of this problem (for example the embeddings module has the same issue)

baskaryan:
@eyurtsev i actually think any OpenAI-returned object, whether you've got pydantic v1 or v2 installed, should support model_dump https://github.com/openai/openai-python/blob/d231d1fa783967c1d3a1db3ba1b52647fff148ac/src/openai/_models.py#L150

SavvasMohito:
great observation @baskaryan. they seem to be using .dict() under the hood [here](https://github.com/openai/openai-python/blob/d231d1fa783967c1d3a1db3ba1b52647fff148ac/src/openai/_models.py#L189) but through the model_dump() function name in case of pydantic v1.

eyurtsev:
Ah if OpenAI already supports it we can just upgrade

SavvasMohito:
Closed by mistake, re-opened in #17404. Sorry 😢

Issue #15954, opened 2024-01-12
Input or output validation errors are not reported to run manager


molatif-dev:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

```python
prompt_template = ("Human: Say hello to the user using the given username."
                   "Username: {username}")

chain = LLMChain(
            llm=llm,
            prompt=PromptTemplate(template=prompt_template, input_variables=[]),
            verbose=True,
        )

def execute(args:dict):
    with callbacks.collect_runs() as cb:
        try:
            output = chain.invoke(args)
            print(f"output: {output}")
        finally:
            print(f"traced_runs: {cb.traced_runs}")


# executes successfully
execute({"username": "John"})

# failed with validation error, but traced runs is empty, while it should have one run with exception details
execute({"user_name": "John"})

``

### Description

Callback manager doesn't get invoked if there's an input or output validation error. 
For logging purposes the callback manager should be invoked and any chain related error, even if it's a validation error.
The related logic exists in file `langchain/chains/base.py` in `acall` and `__call__` functions, as the `callback_manager` & `run_manager` gets initialized after validating input. To fix this, `run_manager` should be initialized first, then wrap preparing input, output and executing in try/except and update the `run_manager` accordingly

### System Info

langchain==0.0.354
langchain-community==0.0.11
langchain-core==0.1.9

Python 3.10.13

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [X] Chains
- [ ] Callbacks/Tracing
- [ ] Async

molatif-dev:
Fixed

PR #16644, opened 2024-01-26
langchain[patch]: Invoke chain prep_inputs and prep_outputs inside try block to catch validation errors

mlatif-cs:
- **Description:** Callback manager can't catch chain input or output validation errors because `prepare_input` and `prepare_output` are not part of the try/raise logic, this PR fixes that logic.
 
  - **Issue:** #15954

### DIFF
--- a/libs/langchain/langchain/chains/base.py
+++ b/libs/langchain/langchain/chains/base.py
@@ -135,6 +135,7 @@ class Chain(RunnableSerializable[Dict[str, Any], Dict[str, Any]], ABC):
         include_run_info = kwargs.get("include_run_info", False)
         return_only_outputs = kwargs.get("return_only_outputs", False)
 
+        inputs = self.prep_inputs(input)
         callback_manager = CallbackManager.configure(
             callbacks,
             self.callbacks,
@@ -145,13 +146,14 @@ class Chain(RunnableSerializable[Dict[str, Any], Dict[str, Any]], ABC):
             self.metadata,
         )
         new_arg_supported = inspect.signature(self._call).parameters.get("run_manager")
+
         run_manager = callback_manager.on_chain_start(
             dumpd(self),
-            input,
+            inputs,
             name=run_name,
         )
         try:
-            inputs = self.prep_inputs(input)
+            self._validate_inputs(inputs)
             outputs = (
                 self._call(inputs, run_manager=run_manager)
                 if new_arg_supported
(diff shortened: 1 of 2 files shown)


eyurtsev:
@molatif-dev Any chance you could add a unit test for this? There's  fake tracer that can be used

molatif-dev:
@eyurtsev I added two unit tests for input and output validation error

molatif-dev:
Hey @hwchase17 any feedback on why this PR was declined?

libs/langchain/langchain/chains/base.py
```diff
@@ -148,22 +147,25 @@ def invoke(
         new_arg_supported = inspect.signature(self._call).parameters.get("run_manager")
         run_manager = callback_manager.on_chain_start(
             dumpd(self),
-            inputs,
+            input,
```
eyurtsev, line 150: We need to modify the logic a bit to avoid introducing a breaking change.

Currently, on the happy path all get the output of `prep_inputs` rather than `input`, because those two values are different, we need to make sure that we don't modify the happy batch.

What we can do is something like:

```python
try:
  inputs = self.prep_inputs(input)
except BaseException as e:
  # un-happy path
  invoke on chain start callback handler
  invoke on chain error call back hanlder

# continue with happy path (same code as before)
on_chain_start using the output from self.prep_inputs
```

eyurtsev (commented):
Need to fix logic on happy path to make sure it's not introducing a breaking change. Once fixed we can merge the change.

Ideally we could introduce a unit-test to that can verify that the happy path does not include a breaking change (this unit-test would have a test that verifies that inputs != the output from self.prep_inputs)

eyurtsev:
@molatif-dev  A bunch of PRs were closed this week due to an issue with github -- so this wasn't an issue with your PR or anything of the sort.

We need to make another change to the callback logic as I think the current code introduces a breaking change / bug into the on_chain_start callbacks on the happy path in some cases

eyurtsev:
Thanks @molatif-dev -- i kicked off tests, but won't get to review until tomorrow

libs/langchain/langchain/chains/base.py
eyurtsev: Looks good! Sorry for the delay in reviewing.

eyurtsev (approved):

libs/langchain/langchain/chains/base.py
```diff
@@ -461,18 +481,10 @@ def prep_inputs(self, inputs: Union[Dict[str, Any], Any]) -> Dict[str, str]:
                 # If there are multiple input keys, but some get set by memory so that
                 # only one is not set, we can still figure out which key it is.
                 _input_keys = _input_keys.difference(self.memory.memory_variables)
-            if len(_input_keys) != 1:
```
eyurtsev, line 464: If you're able could you follow up with a PR to update the doc-string in prep_inputs since it's not doing a validation anymore?

PR #16646, opened 2024-01-26
community: apply embedding functions during query if defined

Rijul1204:
**Description:** This update ensures that the user-defined embedding function specified during vector store creation is applied during queries. Previously, even if a custom embedding function was defined at the time of store creation, Bagel DB would default to using the standard embedding function during query execution. This pull request addresses this issue by consistently using the user-defined embedding function for queries if one has been specified earlier.

### DIFF
--- a/libs/community/langchain_community/vectorstores/bageldb.py
+++ b/libs/community/langchain_community/vectorstores/bageldb.py
@@ -110,8 +110,8 @@ class Bagel(VectorStore):
         except ImportError:
             raise ImportError("Please install bagel `pip install betabageldb`.")
 
-        texts = list(query_texts)
-        if self._embedding_function and query_embeddings is None and texts:
+        if self._embedding_function and query_embeddings is None and query_texts:
+            texts = list(query_texts)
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/bageldb.py
```diff
@@ -109,6 +109,12 @@ def __query_cluster(
             import bagel  # noqa: F401
         except ImportError:
             raise ImportError("Please install bagel `pip install betabageldb`.")
+
+        texts = list(query_texts)
```
baskaryan, line 113: will raise errors if  query_texts is None

Rijul1204: Thank you for the review. You are correct. I have addressed this issue in a subsequent commit. Would you prefer to have these two commits squashed into one, or is it fine to keep them as two separate commits?

baskaryan (approved):

baskaryan:
thanks @Rijul1204!

PR #16653, opened 2024-01-26
fix(experimental): missing resolution strategy in anonymization

mspronesti:
- **Description:** Presidio-based anonymizers are not working because `_remove_conflicts_and_get_text_manipulation_data` was being called without a conflict resolution strategy. This PR fixes this issue. In addition, it removes some mutable default arguments (antipattern).
 
To reproduce the issue, just run the very first cell of this [notebook](https://python.langchain.com/docs/guides/privacy/2/) from langchain's documentation.


baskaryan (commented):
cc @maks-operlejn-ds

mspronesti:
@baskaryan some more context on this PR: Microsoft changed the signature of `_remove_conflicts_and_get_text_manipulation_data` in one of the latest minor releases. As that is a private method, they "didn't care" about backward compatibility. However, we use it in our classes, that's why I believe this update would be needed.

MaksOpp:
LGTM, thanks for solving the issue @mspronesti!

baskaryan:
thank you @mspronesti!

PR #16656, opened 2024-01-27
Intergrate with Azure AI

quchuyuan:


efriis (commented):
Hey @quchuyuan ! This format for this PR doesn't make sense, and I'd be happy to work with the azure team responsible for this!

Which Azure products are you hoping to integrate here for hosting langserve applications and prompts?

libs/core/langchain_core/prompts/loading.py
```diff
@@ -127,13 +127,26 @@ def _load_prompt(config: dict) -> PromptTemplate:
...
+    if path.startswith("azureml://"):
+        return _load_prompt_from_azureai(path)
     if hub_result := try_load_from_hub(
         path, _load_prompt_from_file, "prompts", {"py", "json", "yaml"}
     ):
         return hub_result
     else:
         return _load_prompt_from_file(path)
 
+def _load_prompt_from_azureai(path: Union[str, Path]) -> BasePromptTemplate:
```
hwchase17, line 139: i think we are trying to use `load_prompt` less, so would rather make this a top level function

eg we have `hub.pull` to pull things from hub, which we use more

libs/cli/langchain_cli/cli.py
```diff
@@ -51,22 +52,27 @@ def serve(
     host: Annotated[
         Optional[str], typer.Option(help="The host to run the server on")
     ] = None,
+    cloud: Annotated[
```
hwchase17, line 55: this seems like a good entry point!

PR #16660, opened 2024-01-27
community: Ollama - Pass headers to post request in async method

zhxu73:
# Description
A previous PR (https://github.com/langchain-ai/langchain/pull/15881) added option to pass headers to ollama endpoint, but headers are not pass to the async method.


baskaryan (approved):

baskaryan:
thanks @zhxu73!

PR #16671, opened 2024-01-27
community: Get Document object by Document ID from Chroma DB

### LINKED ISSUE #15261: Get Chroma vectorstore Document by `doc_id` for document / metadata updates.
### Feature request

It should be possible to search a Chroma vectorstore for a particular Document by it's ID. Given that the Document object is required for the `update_document` method, this lack of functionality makes it difficult to update document metadata, which should be a fairly common use-case.

Currently, there are two methods for searching a vectorstore, `get` and `search` but neither allow me to collect a Document by it's id

`vectorstore.get`: This allows for search via `id`, however, this does not return the actual `Document` object. Instead, the return is a dictionary of lists containing the `id`, `document`, and optionally, the `embeddings` for all matched documents. This provides an easy interface for utilising documents downstream, however, this creates a challenge for document updates as the `update_document` method needs the `Document` object to be passed, which would require needless recreation for updates.

`vectorstore.search`: This returns the `Document` object as required, however, it is not possible to explicitly search via `id`, only similarity search is possible.

As such, it appears that there is currently no easy way to do this at present, without manually recreating the Document from the `get` output.

### Motivation

For my use-case, I am performing offline clustering of my embeddings in order to find the core groups of documents and would like to add the predicted label to each document as metadata "cluster_label".

Below is a simple representation of my current pipeline:

```
all_docs = vectorstore.get(include=["embeddings", "documents"])
doc_ids = all_docs["ids"]
embeddings = np.array(all_docs["embeddings"])
cluster_model, labels = fit_predict_clustering(embeddings, max_components=10)
for doc_id, label in zip(ids, labels):
    # Fetch the document from the vectorstore
    doc = vectorstore.get(doc_id)   # returns Dict[str, List], but I need Document

    # Given current implementation, I would need to now convert the above dictionary to Document
    ...

    # Update metadata with the cluster label
    doc.metadata["cluster_label"] = label
    vectorstore.update(doc_id, doc)
```

### Your contribution

I'm happy to contribute to this feature if deemed beneficial. To my mind, it should be achievable by either:
1. Updating the get method to allow `Document` returning, 
2. Including a new method with the required functionality, or
3. Providing a utility for easy bulk conversion from `get` output to `List[Document]`.

However, I'm open to suggestions as to the most fitting solution.

ksraunak77:
- **Description:**
    Added _get_document_by_id()_ function in langchain_community/vectorstores/chroma.py to get Document object from 
    Document ID. 
    Document object is required for updating any document. get_document_by_id() function will help users to get the 
    Document object from Document ID. Then the metadata or page_content of the Document object can be modified by the 
    user and passed to update_document() function.

  - **Issue:** Fixed #15261

### DIFF
--- a/docs/docs/integrations/vectorstores/chroma.ipynb
+++ b/docs/docs/integrations/vectorstores/chroma.ipynb
@@ -332,6 +332,15 @@
     "example_db.update_document(ids[0], docs[0])\n",
     "print(example_db._collection.get(ids=[ids[0]]))\n",
     "\n",
+    "# get document object by id\n",
+    "doc = example_db.get_document_by_id(document_id=ids[1])\n",
+    "doc.metadata = {\n",
+    "    \"source\": \"../../modules/state_of_the_union.txt\",\n",
+    "    \"new_value\": \"hello world\",\n",
+    "}\n",
+    "example_db.update_document(ids[1], doc)\n",
+    "print(example_db._collection.get(ids=[ids[1]]))\n",
+    "\n",
     "# delete the last document\n",
     "print(\"count before\", example_db._collection.count())\n",
     "example_db._collection.delete(ids=[ids[-1]])\n",
@@ -558,7 +567,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "Python 3.10.6 64-bit",
    "language": "python",
    "name": "python3"
   },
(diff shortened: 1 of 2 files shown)


baskaryan (commented):
cc @jeffchuber

jeffchuber:
@raunakshrivastava7 another way to do this is to by creating a chroma client and passing it to langchain. 

https://python.langchain.com/docs/integrations/vectorstores/chroma#passing-a-chroma-client-into-langchain

My personal feeling is the glue between Chroma-Langchain should focus on the the tools that Langchain as a framework needs (maybe that is the case here?) - otherwise the glue will have to expand to be the full set of Chroma's API and then is more apt to break. From a user education perspective, I think it's also generally more clear "which API you are learning" if they are separate. 

However... I don't feel strongly! Maybe it's just this morning cup of coffee talking. 

@baskaryan  thoughts?

raunakshrivastava7:
As per my understanding, currently the Document object can be fetched using similarity_search functions in langchain. There is no direct method or API to get the Document object from Document ID. The PR solves this problem.

raunakshrivastava7:
As per my understanding, currently the Document object can be fetched using similarity_search functions in langchain. There is no direct method or API to get the Document object from Document ID. The PR solves this problem.

Issue #16631, opened 2024-01-26
Doc: Add in code documentation for RunnableGenerator


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Add in code documentation for Runnable Generator

https://github.com/langchain-ai/langchain/blob/main/libs/core/langchain_core/runnables/base.py#L2810-L2810

Here's an example of in code documentation PR for Runnable Parallel:

https://github.com/langchain-ai/langchain/pull/16600/files

PR #16672, opened 2024-01-27
core: expand docstring for RunnableGenerator

ccurme:
- **Description:** expand docstring for RunnableGenerator
  - **Issue:** https://github.com/langchain-ai/langchain/issues/16631

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -2882,8 +2882,71 @@ RunnableMap = RunnableParallel
 
 
 class RunnableGenerator(Runnable[Input, Output]):
-    """
-    A runnable that runs a generator function.
+    """A runnable that runs a generator function.
+
+    RunnableGenerators can be instantiated directly or by using a generator within
+    a sequence.
+
+    RunnableGenerators can be used to implement custom behavior, such as custom output
+    parsers, while preserving streaming capabilities.
+
+    Here is an example to show the basic mechanics of a RunnableGenerator:
+
+        .. code-block:: python
+
+            from typing import Any, AsyncIterator, Iterator
+
+            from langchain_core.runnables import RunnableGenerator
+
+
+            def gen(input: Iterator[Any]) -> Iterator[str]:
+                for token in ["Have", " a", " nice", " day"]:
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/runnables/base.py
```diff
@@ -2882,8 +2882,71 @@ async def input_aiter() -> AsyncIterator[Input]:
 
 
 class RunnableGenerator(Runnable[Input, Output]):
-    """
-    A runnable that runs a generator function.
+    """A runnable that runs a generator function.
```
nfcampos, line 2885: this is great overall! one detail, we actually support two types of generator functions, might be good to mention that here, see an explanation below

there are two types of generator functions (ie functions that generate a stream of output, as opposed to a function that returns output all at once a single time) that can be useful in building your application:
- "streaming" functions, ie. with signature `A -> Iterator[B]`, they receive the output of the previous step all at once, after the previous step finishes, and can start emitting their own incremental output after that. This is useful when you need the entire output of the previous steps before being able to start doing anything useful in this step, eg. most LLMs need the entire prompt available before starting to generate the completion, so would be easier to implement as one of these. These can be used by wrapping them in RunnableLambda.
- "transformer" functions, ie. with signature `Iterator[A] -> Iterator[B]`, they receive a stream of the output of the previous step while it's being produced, giving them the opportunity to start emitting output chunks as soon as the previous step emits the first chunk. This is mainly useful for output parser-like functions, which are expected to be used after a previous step that streams output, like an LLM, and are expected to parse it incrementally as soon as they can. These can be used by wrapping them in RunnableGenerator.

ccurme: Thanks for this! I added some text making this distinction near the top of the docstring. I also added an example at the bottom capturing the behavior for the first type of function you described. Let me know your thoughts (e.g., if it is too contrived), or if I misunderstood anything. Thanks!

nfcampos (approved):

libs/core/langchain_core/runnables/base.py
nfcampos: Cool, looks great!

PR #16690, opened 2024-01-28
community: Invalidate sqlalchemy connection

### LINKED ISSUE #15687: No close() functionality in  langchain.sql_database import SQLDatabase package
### Feature request

from langchain_experimental.sql import SQLDatabaseChain
from langchain.sql_database import SQLDatabase
I'm using the above packages to connect the databricks database(SQLDatabse)and passing it to the model chain(SQLDatabaseChain) to generate the SQLQuery. But I want to close the connection of the database after each response. I couldn't find anything to close the database connection using this SQLDatabase package. Even in the SQLDatabase documentation I couldn't find anything. So I need some close() function to close the connection of the database.

### Motivation

Because of this close() functionality not available in the SQLDatabase package. I'm getting (sqlalchemy.exc.OperationalError) so I need to reboot the server to tackle this issue but that was not the feasible solution. And one more thing I can't use other different packages to connect my database because the model chain accept only the SQLDatabase in the parameter. 

### Your contribution

Try to add the close() functionality in the SQLDatabase.py file so the database connection can be closed. So that I'll not be facing any issues in the future. 
Thanks in advance.

ksraunak77:
- **Description:** 
 connection.close() restores the DBAPI connection back to the connection-holding pool but does not actually close it.connection.invalidate() closes the underlying DBAPI connection. It will prevent exceptions such as OperationalError.

  - **Issue:** Fixed #15687

### DIFF
--- a/libs/community/langchain_community/utilities/sql_database.py
+++ b/libs/community/langchain_community/utilities/sql_database.py
@@ -357,6 +357,8 @@ class SQLDatabase:
                     map(lambda ls: [str(i)[:100] for i in ls], sample_rows_result)
                 )
 
+            connection.invalidate()
+
             # save the sample rows in string format
             sample_rows_str = "\n".join(["\t".join(row) for row in sample_rows])
 


raunakshrivastava7:
closing due to failed checks

Issue #16698, opened 2024-01-28
DOC: Error in Retry Parser example documentation


ash-hun:
### Issue with current documentation:

I tried out the tutorial at this link: https://python.langchain.com/docs/modules/model_io/output_parsers/types/retry  

But I getting this error related to Retry Parser tutorial example `ValidationError: 1 validation error for Action action_input Field required [type=missing, input_value={'action': 'search'}, input_type=dict] For further information visit https://errors.pydantic.dev/2.5/v/missing`.

After conducting various experiments to find the cause, I found out that changing the part `from pydantic import BaseModel, Field` in the code example to `from langchain_core.pydantic_v1 import BaseModel, Field` makes the example code run correctly. The version of langchain I tested is `0.1.3.`

It seems that the official documentation's examples have not been updated to reflect the changes in syntax due to version updates of langchain, so I'm leaving this issue.

#### 👉 Summary
- I tried out the Retry Parser tutorial example
- I founded an error that seems to be due to the example content not being updated following a version update of langchain.
  - from `from pydantic import BaseModel, Field`
  - to `from langchain_core.pydantic_v1 import BaseModel, Field`
- I used to `langchain v0.1.3` and I confirmed that the example works correctly when executed as I described.

### Idea or request for content:

I hope that the issue I raised will be reflected in the official documentation :)

PR #16699, opened 2024-01-28
docs: Syntax correction according to langchain version update in 'Retry Parser' tutorial example

ash-hun:
- **Description:** Syntax correction according to langchain version update in 'Retry Parser' tutorial example, 
- **Issue:** #16698


hwchase17 (commented):
thanks!

PR #16704, opened 2024-01-28
Make input param optional for retrieval chain and history aware retriever chain

jacoblee93:
If `chat_history` is a non-empty list of messages, we can extract the last message in the list as the input. This helps these chains be used in chatbot settings where we only want to pass a list of messages.

CC @baskaryan @hwchase17

### DIFF
--- a/libs/langchain/langchain/chains/history_aware_retriever.py
+++ b/libs/langchain/langchain/chains/history_aware_retriever.py
@@ -48,10 +48,11 @@ def create_history_aware_retriever(
             chain.invoke({"input": "...", "chat_history": })
 
     """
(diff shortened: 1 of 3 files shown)


jacoblee93:
Need to add a test for non message list chat histories

hwchase17 (commented):
dont LOVE this, seems a bit janky

jacoblee93:
Am changing to be use a new `messages` param and to prefer that if provided, will push tonight.

hwchase17 (commented):
theres not a lot of actual code overlap here, maybe worht just two sepearate functions?

PR #16716, opened 2024-01-29
langchain: pubmed tool path update in doc

j-space-b:
- **Description:** The current pubmed tool documentation is referencing the path to langchain core not the path to the tool in community.  The old tool redirects anyways, but for efficiency of using the more direct path, just adding this documentation so it references the new path
  - **Issue:** doesn't fix an issue
  - **Dependencies:** no dependencies
  - **Twitter handle:** rooftopzen


hwchase17 (approved):
thanks!

PR #16728, opened 2024-01-29
Add async methods to BaseChatMessageHistory and BaseMemory

cbornet:
**Description:**
 
Adds:
   * async methods to BaseChatMessageHistory
   * async methods to ChatMessageHistory
   * async methods to BaseMemory
   * async methods to BaseChatMemory
   * async methods to ConversationBufferMemory
   * tests of ConversationBufferMemory's async methods

  **Twitter handle:** cbornet_

### DIFF
--- a/libs/core/langchain_core/chat_history.py
+++ b/libs/core/langchain_core/chat_history.py
@@ -15,11 +15,26 @@ from langchain_core.runnables import run_in_executor
 class BaseChatMessageHistory(ABC):
     """Abstract base class for storing chat message history.
 
-    Implementations should over-ride the add_messages method to handle bulk addition
-    of messages.
+    Implementations guidelines:
 
-    The default implementation of add_message will correctly call add_messages, so
-    it is not necessary to implement both methods.
+    Implementations are expected to over-ride all or some of the following methods:
+
+    * add_messages: sync variant for bulk addition of messages
+    * aadd_messages: async variant for bulk addition of messages
+    * messages: sync variant for getting messages
+    * aget_messages: async variant for getting messages
+    * clear: sync variant for clearing messages
+    * aclear: async variant for clearing messages
+
+    add_messages contains a default implementation that calls add_message
+    for each message in the sequence. This is provided for backwards compatibility
+    with existing implementations which only had add_message.
+
+    Async variants all have default implementations that call the sync variants.
+    Implementers can choose to over-ride the async implementations to provide
+    truly async implementations.
+
+    Usage guidelines:
 
     When used for updating history, users should favor usage of `add_messages`
     over `add_message` or other variants like `add_user_message` and `add_ai_message`
@@ -55,10 +70,21 @@ class BaseChatMessageHistory(ABC):
     """
 
     messages: List[BaseMessage]
-    """A list of Messages stored in-memory."""
+    """A property or attribute that returns a list of messages.
 
-    async def amessages(self) -> List[BaseMessage]:
-        """Return messages stored in memory."""
+    In general, getting the messages may involve IO to the underlying
+    persistence layer, so this operation is expected to incur some
+    latency.
+    """
+
+    async def aget_messages(self) -> List[BaseMessage]:
+        """Async version of getting messages.
+
+        Can over-ride this method to provide an efficient async implementation.
+
+        In general, fetching messages may involve IO to the underlying
+        persistence layer.
+        """
         return await run_in_executor(None, lambda: self.messages)
 
     def add_user_message(self, message: Union[HumanMessage, str]) -> None:
--- a/libs/core/langchain_core/chat_history.py
+++ b/libs/core/langchain_core/chat_history.py
@@ -9,6 +9,7 @@ from langchain_core.messages import (
     HumanMessage,
     get_buffer_string,
 )
+from langchain_core.runnables import run_in_executor
 
 
 class BaseChatMessageHistory(ABC):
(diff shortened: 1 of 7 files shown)


eyurtsev:
Comandeering to resolve conflicts

cbornet:
So IIUC, we only add `aad_messages`. Is that correct ?

cbornet:
> Comandeering to resolve conflicts

@eyurtsev done.

libs/core/langchain_core/chat_history.py
```diff
@@ -9,6 +9,7 @@
     HumanMessage,
     get_buffer_string,
 )
+from langchain_core.runnables import run_in_executor
 
 
 class BaseChatMessageHistory(ABC):
```
eyurtsev, line 15: Updated doc-string (feel free to incorporate or else I'll follow up with a PR)

```python
class BaseChatMessageHistory(ABC):
    """Abstract base class for storing chat message history.

    Implementations guidelines:

    Implementations are expected to over-ride all or some of the following methods:

    * add_messages: sync variant for bulk addition of messages
    * aadd_messages: async variant for bulk addition of messages
    * messages: sync variant for getting messages
    * amessages: async variant for getting messages
    * clear: sync variant for clearing messages
    * aclear: async variant for clearing messages

    add_messages contains a default implementation that calls add_message
    for each message in the sequence. This is provided for backwards compatibility
    with existing implementations which only had add_message.

    Async variants all have default implementations that call the sync variants.
    Implementers can choose to over-ride the async implementations to provide
    truly async implementations.

    Usage guidelines:

    When used for updating history, users should favor usage of `add_messages`
    over `add_message` or other variants like `add_user_message` and `add_ai_message`
    to avoid unnecessary round-trips to the underlying persistence layer.

    Example: Shows a default implementation.

        .. code-block:: python

            class FileChatMessageHistory(BaseChatMessageHistory):
                storage_path:  str
                session_id: str

               @property
               def messages(self):
                   with open(os.path.join(storage_path, session_id), 'r:utf-8') as f:
                       messages = json.loads(f.read())
                    return messages_from_dict(messages)

               def add_messages(self, messages: Sequence[BaseMessage]) -> None:
                   all_messages = list(self.messages) # Existing messages
                   all_messages.extend(messages) # Add new messages

                   serialized = [message_to_dict(message) for message in all_messages]
                   # Can be further optimized by only writing new messages
                   # using append mode.
                   with open(os.path.join(storage_path, session_id), 'w') as f:
                       json.dump(f, messages)

               def clear(self):
                   with open(os.path.join(storage_path, session_id), 'w') as f:
                       f.write("[]")
    """

    messages: List[BaseMessage]
    """A property or attribute that returns a list of messages.
    
    In general, getting the messages may involve IO to the underlying
    persistence layer, so this operation is expected to incur some
    latency.
    """

    async def aget_messages(self) -> List[BaseMessage]:
        """Async version of getting messages.

        Can over-ride this method to provide an efficient async implementation.

        In general, fetching messages may involve IO to the underlying
        persistence layer.
        """
        return await run_in_executor(None, lambda: self.messages)
```

libs/langchain/langchain/memory/buffer.py
```diff
@@ -19,20 +19,40 @@ def buffer(self) -> Any:
         """String buffer of memory."""
         return self.buffer_as_messages if self.return_messages else self.buffer_as_str
 
-    @property
-    def buffer_as_str(self) -> str:
-        """Exposes the buffer as a string in case return_messages is True."""
+    async def abuffer(self) -> Any:
```
eyurtsev, line 22: @cbornet I suspect that we'll be deprecating these abstractions eventually. They're over-complicated for what they do. Regardless, I think OK adding async methods.

libs/core/langchain_core/chat_history.py
```diff
@@ -56,6 +57,10 @@ def clear(self):
     messages: List[BaseMessage]
     """A list of Messages stored in-memory."""
 
+    async def amessages(self) -> List[BaseMessage]:
```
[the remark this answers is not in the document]
cbornet: Sure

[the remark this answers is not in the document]
cbornet: Yes. That's especially problematic in this part
```python
    def __str__(self) -> str:
        """Return a string representation of the chat history."""
        return get_buffer_string(self.messages)
```
If you call this in async mode, it might block the event loop.

libs/langchain/langchain/memory/buffer.py
cbornet: If we 're not sure these abstractions will be kept, I think it's better to not add the async version for now. It's harder to remove things than to add them later.
I'll update.

cbornet: Well actually those are used internally by `aload_memory_variables`. Should they be made private with `_` ?
Since it's an implementation and not an interface, it's maybe not so bad to keep them.

libs/core/langchain_core/chat_history.py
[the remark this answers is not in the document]
cbornet: done

libs/langchain/langchain/memory/buffer.py
eyurtsev: For public/private, let's match the syntax of any existing sync variants?

---

Regarding the abstraction, I was referring to BaseChatMemory and implementations. 

If I understand correctly, their primary role is to generate a prompt for an LLM / chat model.

We definitely want the functionality, but need to figure out how to simplify the API and make it more intuitive. One possibility is to expose more standalone functions that can operate on messages (e.g., for history trimming based on number of messages, trimming based on # of tokens, summarization etc.)

eyurtsev: We for sure want to merge the improvements for `BaseChatMessageHistory` -- that abstraction we definitely want/need!

eyurtsev (approved):

eyurtsev:
Good to go. Waiting for confirmation from @cbornet on whether to keep the async methods in the Memory classes or avoid adding them for now.

cbornet:
I think we're good to go!

cbornet:
I just figured that the cache history was probably supposed to load the messages at init and then only do writes when add_message is called and keep a local copy of the messages.
So this means that aget_messages() shouldn't exist.
Should I remove it ?

PR #16447, opened 2024-01-23
Docs: update how to use agent in Slack's docs

taimo3810:
**Description:** update how to use agent in Slack's docs

PR #16732, opened 2024-01-29
Update Slack agent toolkit

rlancemartin:


rlancemartin:
Fmt fix from [here](https://github.com/langchain-ai/langchain/pull/16447).

PR #16749, opened 2024-01-30
nvidia-trt: remove tritonclient all extra dep

efriis:


libs/partners/nvidia-trt/pyproject.toml
```diff
@@ -13,7 +13,7 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
 langchain-core = ">=0.0.12"
-tritonclient = { extras = ["all"], version = "^2.40.0" }
+tritonclient = "^2.40.0"
```
hinthornw, line 16: Maybe we add an `[all]` option to this package itself?

efriis: At that point wouldn't people just run their own `pip install tritonclient[all]` though?

PR #16758, opened 2024-01-30
Community: Add Progress bar to HuggingFaceEmbeddings

tylertitsworth:
- **Description:** Adds a function parameter to HuggingFaceEmbeddings called `show_progress` that enables a `tqdm` progress bar if enabled. Does not function if `multi_process = True`.
  - **Issue:** n/a
  - **Dependencies:** n/a


tylertitsworth:
Hello, have I broken a rule or not followed some contributing guidelines? Why was this closed and re-opened twice?

I am happy to make any necessary changes to ensure we can collaborate on this feature :)

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -87,10 +92,24 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
...
+                except ImportError:
+                    logging.warning(
+                        "Unable to show progress because tqdm could not be imported."
+                        "Please install it with `pip install tqdm`."
+                    )
+                    iter_ = texts
+            else:
+                embeddings = self.client.encode(texts, **self.encode_kwargs)
+
+        return [self.client.encode(text, **self.encode_kwargs) for text in iter_]
```
baskaryan, line 112: do we know if this affects speed?

baskaryan (commented):
sorry for closing/re-opening, wasn't intentional, GitHub was acting up

libs/community/langchain_community/embeddings/huggingface.py
baskaryan: probably fine either way since this is opt-in, but worth documenting if it does slow things down

tylertitsworth: I've seen the same format in other community embedding functions who have implemented a progress bar.

If we look at the encoding function that gets called in [`sentence-transformers`](https://github.com/UKPLab/sentence-transformers/blob/2c4d21cfd5d6bb7bf6067619a701320537d29a01/sentence_transformers/SentenceTransformer.py#L204-L262) it has handling for both strings and lists of strings, so this would allow us to skip the sorting steps that occur in the function that would otherwise just be overhead for a single string.

eyurtsev: The underlying API implements batch optimization based on the referenced code. We should not be sacrificing that even as an opt-in with show_progress=bar. Users will not know and the performance loss is not worth it.

Instead we can use the `show_progress` of the underlying encode function.

eyurtsev (changes requested):

eyurtsev (approved):

baskaryan (approved):

baskaryan:
thank you @tylertitsworth!

PR #16759, opened 2024-01-30
[Community] infinity embeddings: update incorrect default url

michaelfeil:
The default url has always been incorrect (7797 instead 7997). Here is a update to the correct url.

READY FOR REVIEW.


michaelfeil:
I am confused, is there a reason for closing this? :)

baskaryan:

michaelfeil:
Thanks, that took a bit, thanks for merging this finally.

PR #16771, opened 2024-01-30
docs: add use case for managing chat messages via Apache Kafka 

merlin-quix:
Adding a new notebook that demonstrates how to use LangChain's standard chat features while passing the chat messages back and forth via Apache Kafka.

This goal is to simulate an architecture where the chat front end and the LLM are running as separate services that need to communicate with one another over an internal nework.

It's an alternative to typical pattern of requesting a reponse from the model via a REST API (there's more info on why you would want to do this at the end of the notebook).

NOTE: Assuming "uses cases" is the right place for this but feel free to propose another location.


hwchase17 (commented):
lets move to a cookbook!

baskaryan (approved):

PR #16774, opened 2024-01-30
Update openai_info.py with gpt-4-0125-preview model

ricoms:
Update openai_info.py with gpt-4-0125-preview model


baskaryan:
believe it's already been updated, thank you anyways @ricoms!

PR #16784, opened 2024-01-30
langchain_google_vertexai : added logic to override get_num_tokens_from_messages() for ChatVertexAI

Adi8885:

  - **Description: added logic to override get_num_tokens_from_messages() for ChatVertexAI. Currently ChatVertexAI was inheriting get_num_tokens_from_messages() from BaseChatModel which in-turn was calling GPT-2 tokenizer
  - **Issue: NA
  - **Dependencies: NA
  - **Twitter handle:@aditya_rane

@lkuligin for review


libs/partners/google-vertexai/tests/integration_tests/test_chat_models.py
efriis, line 1: why remove these two tests?

Adi8885: my bad , it was causing some issues on my local setup. Added them back

turanbulmus:
@efriis can we merge this as soon as possible? We have a production case depending on this PR. Thanks a lot for your support

Adi8885:
@lkuligin For review

thomasmenard-oa:
@efriis @lkuligin we need this PR merged and pushed into a new version langchain_google_vertexai.
We have production deployement awaiting for this fix.
Thanks :)

Adi8885:
@efriis @lkuligin Kindly let me know if there are any blockers for this PR , we have critical deployment awaiting this change?

thomasmenard-oa:
Looping @hwchase17. Sorry to ping you on that one but we need this PR merged please

libs/partners/google-vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -531,3 +531,26 @@ def _start_chat(
             )
         else:
             return self.client.start_chat(message_history=history.history, **kwargs)
+
+    def get_num_tokens_from_messages(self, messages: List[BaseMessage]) -> int:
```
lkuligin, line 535: it's probably better to move the `get_num_tokens` method from `VertexAI` to `_VertexAICommon`, and then everything should work just fine.

lkuligin (changes requested):

lkuligin:
also, it might be a good idea to add an integration test that uses `ConversationMemoryBuffer` since this is what we're trying to fix here if I'm not wrong.

libs/partners/google-vertexai/langchain_google_vertexai/chat_models.py
Adi8885: Done

Adi8885:
> also, it might be a good idea to add an integration test that uses `ConversationMemoryBuffer` since this is what we're trying to fix here if I'm not wrong.

@lkuligin Were testing "ConversationSummaryBufferMemory" , but not sure what we want to check integration test. I've checked locally , it doesn't re-direct to tokenizer download

libs/partners/google-vertexai/langchain_google_vertexai/chat_models.py
lkuligin: we don't even need `get_num_tokens_from_messages`, we need to move an existing `get_num_tokens` method and then it should work:
https://github.com/langchain-ai/langchain/blob/f027696b5f068bacad96a9356ae196f5de240007/libs/core/langchain_core/language_models/base.py#L275

Adi8885: @lkuligin This approach would result in multiple calls to the count_token endpoint i.e separately for each message. This approach in this PR collates all text in messages and calls the  count_token endpoint only once. 
```
https://github.com/langchain-ai/langchain/blob/f027696b5f068bacad96a9356ae196f5de240007/libs/core/langchain_core/language_models/base.py#L275
```

lkuligin: fair enough, but let's still move the `get_num_tokens`, please. 
and I'm not sure that difference (1 large call vs n smaller calls) is significant.

@sasha-gitg any thoughts on this?

lkuligin: additional thoughts: 
1. let's move `get_num_tokens` pls
2. you can always collapse message together:
either with
```
from vertexai.generative_models._generative_models import _PreviewGenerativeModel

get_buffer_string(
    messages, 
    human_prefix=_PreviewGenerativeModel._USER_ROLE,
    ai_prefix=_PreviewGenerativeModel._MODEL_ROLE)
```
or with 
```
self.client..count_tokens([m.content for m in messages])
```

lkuligin (approved):

efriis (approved):

PR #16800, opened 2024-01-30
Community: Allow adding ARNs as model_id to support Amazon Bedrock custom models

supreetkt:
- **Description:** Adds an additional class variable to `BedrockBase` called `provider` that allows sending a model provider such as amazon, cohere, ai21, etc. 
  Up until now, the model provider is extracted from the `model_id` using the first part before the `.`, such as `amazon` for `amazon.titan-text-express-v1` (see [supported list of Bedrock model IDs here](https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids-arns.html)). But for custom Bedrock models where the ARN of the provisioned throughput must be supplied, the `model_id` is like `arn:aws:bedrock:...` so the `model_id` cannot be extracted from this. A model `provider` is required by the LangChain Bedrock class to perform model-based processing. To allow the same processing to be performed for custom-models of a specific base model type, passing this `provider` argument can help solve the issues.
The alternative considered here was the use of `provider.arn:aws:bedrock:...` which then requires ARN to be extracted and passed separately when invoking the model. The proposed solution here is simpler and also does not cause issues for current models already using the Bedrock class.
  - **Issue:** N/A
  - **Dependencies:** N/A


libs/community/langchain_community/llms/bedrock.py
```diff
@@ -229,9 +229,16 @@ class BedrockBase(BaseModel, ABC):
     config: Optional[Config] = None
     """An optional botocore.config.Config instance to pass to the client."""
 
+    provider: Optional[str]
```
3coins, line 232:
```suggestion
    provider: Optional[str] = None
```

libs/community/langchain_community/llms/bedrock.py
```diff
@@ -229,9 +229,16 @@ class BedrockBase(BaseModel, ABC):
...
+    """The model provider, e.g., amazon, cohere, ai21, etc. When not supplied, provider
+    is extracted from the first part of the model_id e.g. 'amazon' in 
+    'amazon.titan-text-express-v1'. Model providers for custom models can be specified
+    here with the ARN supplied as the model_id."""
```
3coins, line 236:
```suggestion
    """The model provider, e.g., amazon, cohere, ai21, etc. When not supplied, provider
    is extracted from the first part of the model_id e.g. 'amazon' in 
    'amazon.titan-text-express-v1'. This value should be provided for model ids that does not
    have the provider in them, e.g., custom and provisioned models that have an ARN associated
    with them."""
```

libs/community/langchain_community/llms/bedrock.py
```diff
@@ -353,7 +360,16 @@ def _identifying_params(self) -> Mapping[str, Any]:
...
+        if self.provider:
+            return self.provider
+        else:
+            if self.model_id.startswith("arn"):
+                raise ValueError(
+                    "Model provider should be supplied when passing a model ARN as "
+                    "model_id"
+                )
+            else:
+                return self.model_id.split(".")[0]
```
3coins, line 372: else is extraneous here.

```suggestion
        if self.model_id.startswith("arn"):
            raise ValueError(
                "Model provider should be supplied when passing a model ARN as "
                "model_id"
            )
        else:
            return self.model_id.split(".")[0]
```

libs/community/langchain_community/llms/bedrock.py
```diff
@@ -229,9 +229,16 @@ class BedrockBase(BaseModel, ABC):
...
+    """The model provider, e.g., amazon, cohere, ai21, etc. When not supplied, provider
+    is extracted from the first part of the model_id e.g. 'amazon' in 
+    'amazon.titan-text-express-v1'. Model providers for custom models can be specified
+    here with the ARN supplied as the model_id."""
+
     model_id: str
     """Id of the model to call, e.g., amazon.titan-text-express-v1, this is
-    equivalent to the modelId property in the list-foundation-models api"""
+    equivalent to the modelId property in the list-foundation-models api or it can be
+    used to supply the ARN of the fine-tuned model."""
```
3coins, line 241:
```suggestion
    equivalent to the modelId property in the list-foundation-models api. For custom and
    provisioned models, an ARN value is expected."""
```

3coins (commented):
@supreetkt 
Thanks for working on these changes. Some minor comments on the docs, and suggestion to remove extraneous code. Looks good otherwise.

3coins:
@baskaryan 
Looks good, this should be ready to merge.

baskaryan (approved):

baskaryan:
looks like there's a merge conflict, otherwise lgtm

3coins:
@baskaryan 
I need to take over this PR to resolve conflicts, seems like the last few commits have added unintended changes. Please hold off on merging yet. Thanks.

3coins:
@baskaryan 
All good now!

supreetkt:
@baskaryan - please feel free to merge. The merge conflicts are now resolved. Thanks!

baskaryan:
thank you @supreetkt!

PR #16801, opened 2024-01-30
community[fix] - in FAISS vector store, support passing custom DocStore implementation when using from_xxx methods

ksachdeva:
- **Description:** The from__xx methods of FAISS class have hardcoded InMemoryStore implementation and thereby not let users pass a custom DocStore implementation, 
  - **Issue:** no referenced issue,
  - **Dependencies:** none,
  - **Twitter handle:** ksachdeva


baskaryan:
thank you @ksachdeva!

PR #16806, opened 2024-01-30
Community: Fix github search issues and PRs PaginatedList has no len() error

dbfr3qs:
**Description:** 
Bugfix: Langchain_community's GitHub Api wrapper throws a TypeError when searching for issues and/or PRs (the `search_issues_and_prs` method). This is because PyGithub's PageinatedList type does not support the len() method.  See https://github.com/PyGithub/PyGithub/issues/1476
image
  **Dependencies:** None 
  **Twitter handle**: @ChrisKeoghNZ
  
I haven't registered an issue as it would take me longer to fill the template out than to make the fix, but I'm happy to if that's deemed essential.

I've added a simple integration test to cover this as there were no existing unit tests and it was going to be tricky to set them up.

### DIFF
--- a/libs/community/langchain_community/utilities/github.py
+++ b/libs/community/langchain_community/utilities/github.py
@@ -734,7 +734,7 @@ class GitHubAPIWrapper(BaseModel):
             str: A string containing the first 5 issues and pull requests
         """
         search_result = self.github.search_issues(query, repo=self.github_repository)
-        max_items = min(5, len(search_result))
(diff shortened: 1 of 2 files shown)


dbfr3qs:
@baskaryan just mentioning you as per the PR template as it's been a few days since this was opened...

baskaryan (approved):

baskaryan:
thank @dbfr3qs!

PR #16807, opened 2024-01-30
Experimental: Add other threshold types to SemanticChunker

matthaigh27:
**Description**
Adding different threshold types to the semantic chunker. I’ve had much better and predictable performance when using standard deviations instead of percentiles.

image

For all the documents I’ve tried, the distribution of distances look similar to the above: positively skewed normal distribution. All skews I’ve seen are less than 1 so that explains why standard deviations perform well, but I’ve included IQR if anyone wants something more robust.

Also, using the percentile method backwards, you can declare the number of clusters and use semantic chunking to get an ‘optimal’ splitting.


hwchase17:
this is sweet! ill clean up and merge. do you ahve a twitter for a shout out?

hwchase17 (commented):
would love to add a notebook for this!

hwchase17:
> Thanks for taking a look! Yeah, my twitter is [matthaigh27](https://twitter.com/matthaigh27)
> 
> I'll make the notebook I used for this more presentable but happy to extend the one already present with some of these changes too.

yeah lets extend the existing one

hwchase17 (approved):

PR #16823, opened 2024-01-31
A support for ZhipuAI(bigmodel.cn)

EngelsVon:
A support for `ZhipuAI(bigmodel.cn),the only dependency is `zhipuai` package.I've tested it and passed,so I open a new PR.


baskaryan (commented):
don't we already have ChatZhipuAI? And seems like the underlying interaface is a chat model one so no need to create an LLM as well?

KareEnges:
> 我们不是已经有ChatZipuAI了吗？似乎底层界面是聊天模型，所以也不需要创建法学硕士？

Yes,you're right.It's an old PR,when I first open it,there was not any support for Zhipu.Now I will close this PR.Thanks.

PR #16830, opened 2024-01-31
[community] ElasticsearchStore: preserve user headers

maxjakob:
Users can provide an Elasticsearch connection with custom headers. This PR makes sure these headers are preserved when adding the langchain user agent header.


baskaryan:
cc @joemcelroy

joemcelroy (approved):

PR #16834, opened 2024-01-31
langchain: Updated OpenAIEmbeddings to latest version due to deprecation

aevedis:
- **Description:** Swapped OpenAIEmbeddings since ```from langchain_community.embeddings.openai import OpenAIEmbeddings``` got deprecated in v0.1.0, 
  - **Issue:** the issue # it fixes if applicable,
  - **Dependencies:** None,


baskaryan:
langchain-openai isn't a langchain dependency so we can't import from there. importing embeddings from langchain is deprecated, the correct way is to implement from partner packages or community directly.

PR #16851, opened 2024-01-31
core[minor]: Init ChatPromptTemplate with message-like representations

baskaryan:
Slightly breaking since input_variables are inferred.


libs/core/langchain_core/prompts/chat.py
```diff
@@ -593,6 +593,24 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
     validate_template: bool = False
     """Whether or not to try validating the template."""
 
+    def __init__(
+        self,
+        messages: Sequence[MessageLikeRepresentation],
```
eyurtsev, line 598: Could we put a `*` here so that only messages can be passed by position

PR #16854, opened 2024-02-01
core[patch]: doc init positional args

baskaryan:


libs/langchain/langchain/retrievers/ensemble.py
```diff
@@ -195,7 +195,7 @@ def rank_fusion(
         # Enforce that retrieved docs are Documents for each list in retriever_docs
         for i in range(len(retriever_docs)):
             retriever_docs[i] = [
-                Document(page_content=doc) if not isinstance(doc, Document) else doc
+                Document(page_content=cast(str, doc)) if isinstance(doc, str) else doc
```
eyurtsev, line 198: Why is the cast needed?

eyurtsev (approved):

PR #16888, opened 2024-02-01
core[patch]: Hide aliases when serializing

hinthornw:
Currently, if you dump an object initialized with an alias, we'll still dump the secret values since they're retained in the kwargs


eyurtsev:
Thanks! Any chance we could add a unit test for this?

eyurtsev (approved):

hinthornw:
> Thanks! Any chance we could add a unit test for this?

Sure thing!

PR #16904, opened 2024-02-01
core[patch]: Remove deep copying of run prior to submitting it to LangChain Tracing

hinthornw:


eyurtsev:
@hinthornw Why is this not necessary to do?

eyurtsev (approved):

hinthornw:
It only is helpful if the run is caught in the thread pool and mutated by another callback handler in the meantime. The downside is that the copy operation can break tracing for some objects

PR #16905, opened 2024-02-01
Upgrade configuration for Ruff v0.2.0

charliermarsh:
## Summary

This PR upgrades LangChain's Ruff configuration in preparation for Ruff's v0.2.0 release. (The changes are compatible with Ruff v0.1.5, which LangChain uses today.) Specifically, we're now warning when linter-only options are specified under `[tool.ruff]` instead of `[tool.ruff.lint]`.


efriis (approved):

efriis:
Hey @charliermarsh ! Love `ruff` btw <3.

Just confirming `exclude` should still be under `tool.ruff` instead of `tool.ruff.lint`. Looks like when it's under `tool.ruff.lint`, our current `ruff` version ignores the exclude.

https://github.com/langchain-ai/langchain/pull/16905/commits/e226d59403a202cb2c150b3bc793e0568e74a377

charliermarsh:
@efriis -- Oops, sorry! Not sure how I missed your comment, but thank you :)

Yeah that's right -- `exclude` should still be under `tool.ruff`, because it applies to both the linter _and_ the formatter.

(An `exclude` setting does exist under `tool.ruff.lint`, but if provided, it'll _only_ be enforced for the linter. You can use it enforce "extra" exclusions that are specific to the linter or formatter, but in general, you want the `exclude` under `tool.ruff`.

PR #16918, opened 2024-02-02
core[patch]: fix chat prompt partial messages placeholder var

baskaryan:


libs/core/tests/unit_tests/prompts/test_chat.py
```diff
@@ -503,9 +503,27 @@ def test_messages_placeholder() -> None:
         prompt.format_messages()
     prompt = MessagesPlaceholder("history", optional=True)
     assert prompt.format_messages() == []
-    prompt.format_messages(
+    assert prompt.format_messages(
```
nfcampos, line 506: ha

nfcampos (approved):

PR #16941, opened 2024-02-02
core: Add support for all available output parsers in _load_output_parser()

DougTrajano:
In this PR, we are updating the `_load_output_parser()` function to dynamically get all available output parsers to be loaded.

I am not 100% comfortable with this solution, because I am accessing a protected property outside the class (_type) in line 81, so feel free to suggest changes. :)


DougTrajano:
@baskaryan

DougTrajano:
@baskaryan, @eyurtsev, @hwchase17

DougTrajano:
@efriis @eyurtsev can you review this PR?

libs/core/langchain_core/prompts/loading.py
```diff
@@ -72,13 +73,24 @@ def _load_examples(config: dict) -> dict:
...
+    return parsers_map
+
+
 def _load_output_parser(config: dict) -> dict:
     """Load output parser."""
     if "output_parser" in config and config["output_parser"]:
+        output_parsers_map = _get_output_parsers_map()
         _config = config.pop("output_parser")
         output_parser_type = _config.pop("_type")
-        if output_parser_type == "default":
```
baskaryan, line 80: we probably still want to support output_parser_type == "default"

DougTrajano: Ok, I will add it.

DougTrajano: Done @baskaryan

DougTrajano:
@eyurtsev @baskaryan, @eyurtsev, @hwchase17 ping

hwchase17:
this is using the old way of serializing object. there is a new serialization schema, if we want to support output parsers we should do it there, but it may be some work

PR #16949, opened 2024-02-02
Add async methods to VectorStoreQATool

cbornet:

### DIFF
--- a/libs/community/langchain_community/tools/vectorstore/tool.py
+++ b/libs/community/langchain_community/tools/vectorstore/tool.py
@@ -3,7 +3,10 @@
 import json
 from typing import Any, Dict, Optional
 
-from langchain_core.callbacks import CallbackManagerForToolRun
+from langchain_core.callbacks import (
+    AsyncCallbackManagerForToolRun,
+    CallbackManagerForToolRun,
+)
 from langchain_core.language_models import BaseLanguageModel
 from langchain_core.pydantic_v1 import BaseModel, Field
 from langchain_core.tools import BaseTool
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/tools/vectorstore/tool.py
```diff
@@ -51,9 +54,28 @@ def _run(
         chain = RetrievalQA.from_chain_type(
             self.llm, retriever=self.vectorstore.as_retriever()
         )
-        return chain.run(
-            query, callbacks=run_manager.get_child() if run_manager else None
+        return chain.invoke(
```
cbornet, line 57: change fron `run` to `invoke` as the former is deprecated.

baskaryan (approved):

libs/community/langchain_community/tools/vectorstore/tool.py
```diff
@@ -51,9 +54,28 @@ def _run(
...
+        run_manager: Optional[AsyncCallbackManagerForToolRun] = None,
+    ) -> str:
+        """Use the tool asynchronously."""
+        from langchain.chains.retrieval_qa.base import RetrievalQA
+
+        chain = RetrievalQA.from_chain_type(
+            self.llm, retriever=self.vectorstore.as_retriever()
         )
+        return (
+            await chain.arun(
```
baskaryan, line 74: ainvoke

cbornet: done

libs/community/langchain_community/tools/vectorstore/tool.py
alex-ber: @cbornet 

This is incorrect fix. I get


`
AttributeError("'CallbackManager' object has no attribute 'ignore_chain'")
`
as root cause and
`
AttributeError: 'CallbackManager' object has no attribute 'raise_error'
`

The reason is that this code puts run_manager into **handlers**.

Please revert this change back. 

if you want to change from call chain.run() to chain.invoke() this is fine, but the code should be correect.

alex-ber: @cbornet 
I've opened an issue https://github.com/langchain-ai/langchain/issues/18460

PR #16954, opened 2024-02-02
core[patch]: handle some optional cases in tools

efriis:
primary problem in pydantic still exists, where `Optional[str]` gets turned to `string` in the jsonschema `.schema()`

Also fixes the `SchemaSchema` naming issue

### DIFF
--- a/libs/core/tests/unit_tests/utils/test_function_calling.py
+++ b/libs/core/tests/unit_tests/utils/test_function_calling.py
@@ -81,7 +81,7 @@ def test_function_optional_param() -> None:
         a: Optional[str],
         b: str,
         c: Optional[List[Optional[str]]],
-    ):
+    ) -> None:
         """A test function"""
         pass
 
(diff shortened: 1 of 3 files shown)


libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -69,6 +69,22 @@ def test_convert_to_openai_function(
         },
     }
 
-    for fn in (pydantic, function, tool, expected):
+    for fn in (pydantic, function, dummy_tool, expected):
         actual = convert_to_openai_function(fn)  # type: ignore
         assert actual == expected
+
+
+@pytest.mark.xfail(reason="Pydantic converts Optional[str] to str in .schema()")
```
efriis, line 77: this is the failure from within pydantic that I would LOVE to get fixed/addressed at some point. Could either make it `required=False` or do a fancy json schema thing of `anyOf(string, null)` type on output (might cause poor llm performance though)

PR #16967, opened 2024-02-03
community: fixed vector similarity filtering

ablacklama:
**Description:** changed filtering so that failed filter doesn't add document to results. Currently filtering is entirely broken and all documents are returned whether or not they pass the filter.

fixes issue introduced in https://github.com/langchain-ai/langchain/pull/16190


thiswillbeyourgithub:
Guilty author of the guilty commit here. I approve of this change and I'm so sorry!

baskaryan (approved):
thank you @ablacklama! could we add a simple unit test?

ablacklama:
@baskaryan Done. There were actually tests for this before the commit that introduced the bug but those tests were modified to expect the wrong things. 

All the tests should now expect the filters obeyed.

ablacklama:
I can't fix the formatting that the CI is complaining about. I've run `make format` and it doesn't change anything. Might be something weird about my environment. So if someone could format it properly i'd appreciate that.

ablacklama:
Okay, theoretically the formatting should be sorted out now.

baskaryan:
for future reference you can run `make lint` in whatever library you've edited to run the linter locally
```bash
cd libs/community
make lint
```

PR #16981, opened 2024-02-03
langchain: pydantic entity extraction should support singular values alongside l…

eraye1:
…ists, otherwise entity extracts fails when expected cardinality is 1

  - **Description:** When testing entity extraction via pydantic, I noticed it works well when it's getting back multiple entities from a single extraction, but it fails when the expected cardinality is 1.  This extends the schema for validation to accept single entities too.
  - **Issue:** N/A
  - **Dependencies:** N/A
  - **Twitter handle:** (https://twitter.com/therealarrays)

### DIFF
--- a/libs/langchain/langchain/chains/openai_functions/extraction.py
+++ b/libs/langchain/langchain/chains/openai_functions/extraction.py
@@ -1,4 +1,4 @@
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


eyurtsev:
Closing this PR -- this is not the way in which the schema should be modified for single value extraction (it'll introduce false positives). 

Please take a look at the revamped extraction use case for more up to date information on how to do extraction with LangChain

PR #16993, opened 2024-02-04
experimental: patch content of AIMessage generated by OllamaFunctions

shyeetsao:
OllamaFunctions generates an AIMessage with function call info stored in `additional_kwargs` field, which will be dropped by the underlying ChatOllama if we send the message back (to get further response like OpenAI though more tweaking is needed).

So a temporary and minimal patch would be to fill the raw response into `content` field since downstream using hardly depends on it, i.e. typically we use `if "function_call" in msg.additional_kwargs` to tell if there is a fucntion call involved.

### DIFF
--- a/libs/experimental/langchain_experimental/llms/ollama_functions.py
+++ b/libs/experimental/langchain_experimental/llms/ollama_functions.py
@@ -121,7 +121,10 @@ output: {chat_generation_content}"
             )
 
         response_message_with_functions = AIMessage(
-            content="",
+            # fill raw response into content instead of leaving it empty
+            # since ChatOllama doesn't include additional_kwargs
+            # when the AIMessage is sent to LLMs
+            content=response_message,
             additional_kwargs={
                 "function_call": {
                     "name": called_tool_name,


jacoblee93:
Hey @xleven, I think the better thing to do here would be to automatically extract the function call from the `additional_kwargs` field and add it to the content here: https://github.com/langchain-ai/langchain/pull/16993/files#diff-f6bebcd7238bb10fc7b6916e17c7e2a4bed4e3a19cf52625dcd8f331815ea970L90

The idea with this class is to mimic OpenAI's behavior as much as possible. So I think it's better to keep the outputted messages the same but pass the inner `ChatOllama` instance some transformed message instead.

xleven:
> Hey @xleven, I think the better thing to do here would be to automatically extract the function call from the `additional_kwargs` field and add it to the content here: [#16993 (files)](https://github.com/langchain-ai/langchain/pull/16993/files#diff-f6bebcd7238bb10fc7b6916e17c7e2a4bed4e3a19cf52625dcd8f331815ea970L90)

Yeah I dived into it and found under the hood it calls ChatOllama which belongs to langchain-community then not sure if a cross lib update is a good idea.
 
> The idea with this class is to mimic OpenAI's behavior as much as possible. So I think it's better to keep the outputted messages the same but pass the inner `ChatOllama` instance some transformed message instead.

Totally got the idea and will be happy to see FunctionMessage support in ChatOllama.

Few extra notes about mimicking:
- Ollama recently provides [OpenAI compatible API](https://github.com/ollama/ollama/blob/main/docs/openai.md)
- JSON mode in Ollama, which OllamaFunctions is built on, has an equivalent [OpenAI API para](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format) while functions/tools support usually requires model fine-tuning.

I'll leave it open for further updates but feel free to close.

PR #17005, opened 2024-02-04
community: add runtime kwargs to HuggingFacePipeline

ab-10:
This PR enables changing the behaviour of huggingface pipeline between different calls. For example, before this PR there's no way of changing maximum generation length between different invocations of the chain. This is desirable in cases, such as when we want to scale the maximum output size depending on a dynamic prompt size.

Usage example:

```python
from langchain_community.llms.huggingface_pipeline import HuggingFacePipeline
from transformers import AutoModelForCausalLM, AutoTokenizer, pipeline

model_id = "gpt2"
tokenizer = AutoTokenizer.from_pretrained(model_id)
model = AutoModelForCausalLM.from_pretrained(model_id)
pipe = pipeline("text-generation", model=model, tokenizer=tokenizer)
hf = HuggingFacePipeline(pipeline=pipe)

hf("Say foo:", pipeline_kwargs={"max_new_tokens": 42})
```


ab-10:
Hey maintainers, could you please help debug the CI/CD error?

`mkdir: cannot create directory ‘.mypy_cache’: File exists`

hwchase17 (approved):

baskaryan:
is it confusing that elsewhere `pipeline_kwargs` refers to arguments passed to pipeline init, and here it refers to args passed to calling pipeline? users may think that `HuggingFacePipeline(..., pipeline_kwargs={...})` and `HuggingFacePipeline(...)(..., pipeline_kwargs={...})` do the same thing

PR #17017, opened 2024-02-05
Fix condition on custom root type in runnable history

nfcampos:


baskaryan (commented):
worth simple unit test?

PR #17030, opened 2024-02-05
community: add delete method to rocksetdb vectorstore to support recordmanager

morganda:
- **Description:** This adds a delete method so that rocksetdb can be used with `RecordManager`.
  - **Issue:** N/A
  - **Dependencies:** N/A
  - **Twitter handle:** `@_morgan_adams_`


hwchase17 (approved):

morganda:
Updated the test to include Rockset and updated the respective doc as well

PR #17036, opened 2024-02-05
docs: update parse_partial_json source info

Mercurrent:
- **Description:** Update source-link following recent license update at open-interpreter project
  - **Issue:** N/A
  - **Dependencies:** None

### DIFF
--- a/libs/core/langchain_core/output_parsers/json.py
+++ b/libs/core/langchain_core/output_parsers/json.py
@@ -44,8 +44,8 @@ def _custom_parser(multiline_string: str) -> str:
     return multiline_string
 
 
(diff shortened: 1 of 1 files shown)


hwchase17 (changes requested):
i would prefer to make this reference the code at the point in time in which it was adapted, so (1) good catch on the URL, lets update that to reference a commit from when the code was adapted, (2) lets keep the license as MIT, since that was the license when this code was adapted

Mercurrent:
Put commit-reference instead. Closest to the date of adaptaion: 29.09.23

hwchase17 (approved):

PR #17047, opened 2024-02-05
community: extra_kwargs compatibility with openai>=1.0.0 ; core: extra_kwargs compatibility with openai>=1.0.0

tdene:
**Description**:

openai>=1.0.0 changed the way their library accepts sampling parameters.
Previously, langchain put extra sampling parameters [ones not used by the official OpenAI server] into the `extra_kwargs` dictionary, and then passed this dictionary to the openai library.
With openai>=1.0.0, this returns an error such as
```
TypeError: Completions.create() got an unexpected keyword argument 'greedy'
```

This is because openai>=1.0.0 expects extra parameters to be passed into an `extra_body` dictionary.

As a result, **current langchain does not work with any privately-hosted OpenAI-API-compatible server**. This is a **major issue** for users that have trained their own models and created their own server infrastructure.

This PR fixes this issue.


tdene:
I have fixed the following:
 * Unverified commit
 * Linting errors [successfully ran `make format` on my end]
 * Test errors [successfully ran `make test` on my end]

@efriis @baskaryan this will now hopefully pass all tests if CI is run again. Please let me know if there's anything further I can do.


However, I think the `tests/unit_tests/llms/test_openai.py:test_openai_incorrect_field` test should be updated for openai>=1.0.0. I have not done it myself, I don't believe contributors should manipulate the tests to make their own commits pass, but I believe it should look something like:
```
@pytest.mark.requires("openai")
def test_openai_incorrect_field() -> None:
    with pytest.warns(match="not default parameter"):
        llm = OpenAI(foo="bar")
    if _openai_v1_installed():
        assert llm.extra_body == {"foo": "bar"}
    else:
        assert llm.model_kwargs == {"foo": "bar"}
```

Which would allow my code-blocks to change from:
```
        if is_openai_v1():
            values["extra_body"] = extras
        values["model_kwargs"] = extras
```
to the better
```
        if is_openai_v1():
            values["extra_body"] = extras
        else:
            values["model_kwargs"] = extras
```

libs/community/langchain_community/chat_models/openai.py
```diff
@@ -252,25 +257,10 @@ def build_extra(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
-
-        invalid_model_kwargs = all_required_field_names.intersection(extra.keys())
-        if invalid_model_kwargs:
-            raise ValueError(
-                f"Parameters {invalid_model_kwargs} should be specified explicitly. "
-                f"Instead they were passed in as part of `model_kwargs` parameter."
-            )
-
-        values["model_kwargs"] = extra
+        extras = build_extra_kwargs(extra, values, all_required_field_names)
```
efriis, line 260: nice!

libs/community/langchain_community/chat_models/openai.py
```diff
@@ -252,25 +257,10 @@ def build_extra(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+        if is_openai_v1():
+            values["extra_body"] = extras
```
efriis, line 262: why do we need to store this twice?

libs/community/langchain_community/chat_models/openai.py
```diff
@@ -335,12 +325,20 @@ def validate_environment(cls, values: Dict) -> Dict:
     @property
     def _default_params(self) -> Dict[str, Any]:
         """Get the default parameters for calling OpenAI API."""
+        if is_openai_v1():
+            extra_params = {
+                "extra_query": self.extra_query,
+                "extra_headers": self.extra_headers,
+                "extra_body": self.extra_body,
```
efriis, line 332: could we just do this as `self.model_kwargs`?

libs/community/langchain_community/chat_models/openai.py
```diff
@@ -335,12 +325,20 @@ def validate_environment(cls, values: Dict) -> Dict:
...
+            }
+        else:
+            extra_params = self.model_kwargs
         params = {
             "model": self.model_name,
             "stream": self.streaming,
             "n": self.n,
             "temperature": self.temperature,
-            **self.model_kwargs,
+            **extra_params,
```
efriis, line 341: I think this might be a breaking change unless I'm misunderstanding something. Why can't you just pass in `model_kwargs={"extra_body": {...}}` for what you need this for?

efriis (commented):
I don't think this is necessary, and I actually think this will create a breaking change for more people than it's likely to solve a problem for.

As far as I can tell, this is specifically for non-OpenAI hosted endpoints that use OpenAI's sdk but request additional parameters.

For that case, I believe you can just pass `model_kwargs={"extra_body": ...}` in the current implementation.

Let me know if I'm misunderstanding something! Otherwise I will close.

efriis:
Ok I dug into this more, and I think you should do `model_kwargs={"extra_body": ..., "extra...": ...}` to solve your use case.

If that won't work for some reason, let me know!

The reason is because this would be a behavior change for some of the params not explicitly declared on our client, such as `user` or `logprobs`, which should remain at the top level and not be passed in `extra_body`. You can still pass all these "extra_" fields through `model_kwargs` in the current interface.

Feel free to open if you strongly disagree, and we can discuss!

PR #17069, opened 2024-02-05
community: Expose Anthropic Retry Logic

notAlex2:
**Description:**

Expose Anthropic's retry logic, so that `max_retries` can be configured via langchain. Anthropic's retry logic is implemented in their Python SDK here: https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#retries


libs/community/langchain_community/llms/anthropic.py
```diff
@@ -92,11 +95,13 @@ def validate_environment(cls, values: Dict) -> Dict:
                 base_url=values["anthropic_api_url"],
                 api_key=values["anthropic_api_key"].get_secret_value(),
                 timeout=values["default_request_timeout"],
+                max_retries=values["max_retries"],
```
baskaryan, line 98: believe anthropic.Anthropic() doesn't accept None for max_retries, should we give it same default as they do (2)?

notAlex2: Agreed, looks None isn't valid here. Updated to match their implementation.

notAlex2: What do you think @baskaryan ?

baskaryan (approved):

Issue #17048, opened 2024-02-05
Remove "type: ignore" comments


baskaryan:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Currently we have many "type: ignore" comments in our code for ignoring mypy errors. We should rarely if ever use these. We've had to recently add a bunch of these ignore comments because of an error (i made) that was silently causing mypy not to run in CI.

We should work to remove as many of them as possible by fixing the underlying issues. To find the you can just grep for:
```bash
git grep "type: ignore" libs/
```

This is a big effort, even just removing a few at a time would be very helpful.

ali-nobariasl:
Hello, I am a new contributor, I could handle this issue if it is still open.

baskaryan:
We're bumping all our packages to use mypy v1 which is uncovering a bunch of issues that we're type: ignore-ing to start so we're not blocked on the upgrade, but which should fix as well. See #19163 for example

Smit-Parmar:
@baskaryan Is that mean we no longer need to work on this issue?

PR #17073, opened 2024-02-05
core[patch]: fix _sql_record_manager mypy for #17048

moorej-oci:
- **Description:** Add relevant type annotations for relevant session and query objects to resolve mypy errors when `# type: ignore` comments are removed.
  - **Issue:** #17048
  - **Dependencies:** None,
  - **Twitter handle:** [clesiemo3](https://twitter.com/clesiemo3)
 
I attempted to solve the `UpsertionRecord` ignore but it would require added a deprecated plugin or moving completely to sqlalchemy 2.0+ from my understanding. I'm assuming this is not something desired at this point in time.


baskaryan (approved):
thank you @moorej-oci!

PR #17091, opened 2024-02-06
experimental: Enhance protection against arbitrary code execution in PALChain

DanisJiang:
- **Description:** Block some ways to trigger arbitrary code execution bug in PALChain.


eyurtsev:
@DanisJiang thank you for the PR. We need to re-lint and can merge. I'm going to re-title this PR since I don't think we can claim that it fixes arbitrary code execution in PALChain. Approaches based on filtering AST nodes rather than including AST nodes are impossible to guarantee (e.g., will this code work safely in python 3.20 which may include some new AST nodes?)

eyurtsev (approved):
This looks good

eyurtsev:
Ready to merge after tests pass

DanisJiang:
> @DanisJiang thank you for the PR. We need to re-lint and can merge. I'm going to re-title this PR since I don't think we can claim that it fixes arbitrary code execution in PALChain. Approaches based on filtering AST nodes rather than including AST nodes are impossible to guarantee (e.g., will this code work safely in python 3.20 which may include some new AST nodes?)

I haven't try in python 3.20, I think this code only blocks when risky node is detected, so it shouldn't be a problem.

PR #17097, opened 2024-02-06
Update test_sql_database.py

ShorthillsAI:
In the previous function, db_chain.run is used to execute the chain, but in line number 79, only db_chain is called, which can lead to code inefficiency because it doesn't actually execute the chain. To ensure consistency, it's important to use the .run method consistently to execute the chain in all relevant parts of the code.

### DIFF
--- a/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
+++ b/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
@@ -50,16 +50,22 @@ def convert_message_to_dict(message: BaseMessage) -> dict:
 
 def _convert_dict_to_message(_dict: Mapping[str, Any]) -> AIMessage:
     content = _dict.get("result", "") or ""
+    additional_kwargs: Mapping[str, Any] = {}
     if _dict.get("function_call"):
         additional_kwargs = {"function_call": dict(_dict["function_call"])}
         if "thoughts" in additional_kwargs["function_call"]:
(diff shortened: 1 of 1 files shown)


libs/experimental/tests/integration_tests/chains/test_sql_database.py
```diff
@@ -77,7 +77,7 @@ def test_sql_database_sequential_chain_intermediate_steps() -> None:
     db_chain = SQLDatabaseSequentialChain.from_llm(
         OpenAI(temperature=0), db, return_intermediate_steps=True
     )
-    output = db_chain("What company does Harrison work at?")
+    output = db_chain.run("What company does Harrison work at?")
```
eyurtsev, line 80: We generally prefer using .invoke rather than .run now

ShorthillsAI: @eyurtsev, Understood, I have updated PR with .invoke now with all the found code that should be modified with the same.
Thanks

ShorthillsAI: @eyurtsev, I also see that a lint error comes up with this change. It seems like .invoke is an expected nonstring data type value, maybe a dict type.

image

ShorthillsAI: @eyurtsev, Just a reminder, Should we revoke from **.invoke** to **.run** now?

PR #17105, opened 2024-02-06
community: Fix Tongyi streaming output, the token_usage keeps changing, leading to failure in merge_dicts

tongtie:
- **Description:** When Tongyi llm and chat model performing streaming output, the token_usage keeps changing, leading to failure in merging the dictionaries.

- **Tongyi Error:**
image

- **ChatTongyi Error:**
image

- **Test output:**
image


baskaryan:
believe this was resolved in #19014, let me know if i'm missing something!

PR #17124, opened 2024-02-06
community[patch]: Fix chat openai unit test

LuizFrra:
- **Description:** 
  Actually the test named `test_openai_apredict` isn't testing the apredict method from ChatOpenAI. 
  - **Issue:**
  The test where its occur
  - **Dependencies:** 
  None

### DIFF
--- a/libs/community/tests/unit_tests/chat_models/test_openai.py
+++ b/libs/community/tests/unit_tests/chat_models/test_openai.py
@@ -107,7 +107,7 @@ async def test_openai_apredict(mock_completion: dict) -> None:
     mock_client = MagicMock()
     completed = False
 
-    def mock_create(*args: Any, **kwargs: Any) -> Any:
+    async def mock_create(*args: Any, **kwargs: Any) -> Any:
         nonlocal completed
         completed = True
         return mock_completion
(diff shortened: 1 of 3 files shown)


eyurtsev (approved):

eyurtsev:
merge when unit tests pass

eyurtsev:
@LuizFrra unit-tests fail so probably need to update some of the mocking code for the client as well?

LuizFrra:
> @LuizFrra unit-tests fail so probably need to update some of the mocking code for the client as well?

Ok, I will work on thsi now

LuizFrra:
I guess that's working now.
image

LuizFrra:
i forgot to delete de open-api-key test

LuizFrra:
@eyurtsev I also have added this env:
* `os.environ["OPENAI_API_KEY"] = "foo"`
The test was working before because the tests below set the environment before the execution of chat openai tests.
* `unit_tests/embeddings/test_openai.py`
* `unit_tests/llms/test_openai.py`

eyurtsev:
Unit tests should generally not modify env variables in a way that affects them globally -- this can lead to hard to find bugs in unrelated tests. Can we pass an API key through init instead of modifying the environ?

LuizFrra:
@eyurtsev I also removed the env from the others tests and I used the mock function to mock the environment just for the scope of the functions.

eyurtsev:
@LuizFrra why not  pass the api key instead when instantiating the model? It will result in less code and keep the tests simpler. 

I think it's OK to touch the environment via local patch for unit tests that are specifically testing some interacting with the env, but unless the unit test is doing that, we should probably favor passing the api key through the init

LuizFrra:
@eyurtsev I understand what you say, I did the changes that you had suggest.

eyurtsev (approved):
Thanks!

PR #17129, opened 2024-02-06
community[patch]: Release 0.0.18

baskaryan:


libs/community/_test_minimum_requirements.txt
efriis, line 1: meta q but should we read this from pyproject.toml in CI instead?

baskaryan: probably

PR #17135, opened 2024-02-06
infra: read min versions

efriis:


.github/workflows/_release.yml
```diff
@@ -15,7 +15,7 @@ on:
         default: 'libs/langchain'
 
 env:
-  PYTHON_VERSION: "3.10"
+  PYTHON_VERSION: "3.11"
```
efriis, line 18: needed for `tomllib` in min version checker

PR #17136, opened 2024-02-06
mistralai[patch]: 16k token batching logic embed

efriis:


libs/partners/mistralai/langchain_mistralai/embeddings.py
```diff
@@ -72,8 +77,29 @@ def validate_environment(cls, values: Dict) -> Dict:
...
+            )
         return values
 
+    def _get_batches(self, texts: List[str]) -> Iterable[List[str]]:
+        """Split a list of texts into batches of less than 16k tokens
+        for Mistral API."""
+        batch: List[str] = []
+        batch_tokens = 0
+        for text in texts:
+            text_tokens = len(self.tokenizer.encode(text))
```
hinthornw, line 92: Should we `batch_encode()`?

libs/partners/mistralai/langchain_mistralai/embeddings.py
```diff
@@ -84,13 +110,17 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
             List of embeddings, one for each text.
         """
         try:
-            embeddings_batch_response = self.client.embeddings(
-                model=self.model,
-                input=texts,
+            batch_responses = (
+                self.client.embeddings(
```
hinthornw, line 114: Do we have no mapping from source doc to embedding at this point if we're flat mapping the outputs? that seems undesirable. I'd probably average the embeddings or if we had been able to change the return type, return objects

libs/partners/mistralai/langchain_mistralai/embeddings.py
efriis: done!

libs/partners/mistralai/tests/integration_tests/test_embeddings.py
```diff
@@ -38,10 +39,10 @@ async def test_mistralai_embedding_query_async() -> None:
 
 def test_mistralai_embedding_documents_long() -> None:
     """Test MistralAI embeddings for documents."""
-    documents = ["foo bar " * 1000, "test document " * 1000]
+    documents = ["foo bar " * 1000, "test document " * 1000] * 5
```
efriis, line 42: this test would fail before

hinthornw (approved):

PR #17138, opened 2024-02-06
community: support Nomic embeddings

wt3639:
Description: support loading the current SOTA long context text embedder [nomic-ai/nomic-embed-text](https://huggingface.co/nomic-ai/nomic-embed-text-v1) in langchain.
Dependencies: [Sentence Transformers](https://github.com/UKPLab/sentence-transformers)


hwchase17 (commented):
does this need to be a separate class from existing hugging face embeddings?

wt3639:
> does this need to be a separate class from existing hugging face embeddings?

Actually, no. It's similar to the methods: HuggingFaceBgeEmbeddings and HuggingFaceInstructEmbeddings. However, HuggingFaceBgeEmbeddings does not provide an embed instructor, and HuggingFaceInstructEmbeddings uses the class INSTRUCTOR, which does not support passing the parameter trust_remote_code to sentence_transformers. This parameter is necessary for the nomic model. Therefore, the solution could be either to modify one of these two methods or to create a new class.

hwchase17:
what would it look like to modify the instructor class?

PR #17145, opened 2024-02-07
google-vertexai[patch]: serializable citation metadata, release 0.0.4

efriis:
was breaking in langserve before


hinthornw (commented):
Should we bump a minor version for this?

libs/partners/google-vertexai/langchain_google_vertexai/_utils.py
```diff
@@ -114,7 +114,22 @@ def get_generation_info(
                 }
                 for rating in candidate.safety_ratings
             ],
-            "citation_metadata": candidate.citation_metadata,
+            "citation_metadata": (
```
lkuligin, line 117: nits: could it be `proto.Message.to_dict(candidate.citation_metadata)` instead?

PR #17148, opened 2024-02-07
docs `integraions/providers` nav fix

leo-gan:
Issue: `Provides` page is presented as the index page (on the `Providers` item) and as the `Providers/Providers` item. The latter should not be in the menu. See the picture.
image
This PR fixes this.


hwchase17 (approved):
thanks!

PR #17151, opened 2024-02-07
docs `Integraions/Components` menu reordered

leo-gan:
This PR is opinionated.
- Moved `Embedding models` item to place after `LLMs` and `Chat model`, so all items with models are together.
- Renamed `Text embedding models` to `Embedding models`. Now, it is shorter and easier to read. `Text` is obvious from context. The same as the `Text LLMs` vs. `LLMs` (we also have multi-modal LLMs).


hwchase17 (approved):
thanks!

Issue #17161, opened 2024-02-07
Issue: JsonOutputParser's get_format_instructions() Modifying Pydantic Class Schema


L-cloud:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

```
model = ChatOpenAI(temperature=0)


class Joke(BaseModel):
    setup: str = Field(description="question to set up a joke")
    punchline: str = Field(description="answer to resolve the joke")


joke_query = "Tell me a joke."

parser = JsonOutputParser(pydantic_object=Joke)

prompt = PromptTemplate(
    template="Answer the user query.\n{format_instructions}\n{query}\n",
    input_variables=["query"],
    partial_variables={"format_instructions": parser.get_format_instructions()}, #cause
)

chain = prompt | model | parser

chain.invoke({"query": joke_query})


openai_functions = [convert_to_openai_function(Joke)]


parser = JsonOutputFunctionsParser()

chain = prompt | model.bind(functions=openai_functions) | parser

chain.invoke({"query": "tell me a joke"})
# openai.BadRequestError: Error code: 400 - {'error': {'message': "'' does not match '^[a-zA-Z0-9_-]{1,64}$' - 'functions.0.name'", 'type': 'invalid_request_error', 'param': None, 'code': None}}

```

### Error Message and Stack Trace (if applicable)

 ```
# openai.BadRequestError: Error code: 400 - {'error': {'message': "'' does not match '^[a-zA-Z0-9_-]{1,64}$' - 'functions.0.name'", 'type': 'invalid_request_error', 'param': None, 'code': None}}
```

### Description

For the JsonOutputParser.get_format_instructions function, it modifies the result of pydantic_object.schema() directly. This can lead to unintended issues when the same Pydantic class calls .schema() multiple times. An example of this is the convert_to_openai_function function. Below is an illustration of this scenario.

### System Info

langchain==0.1.4
langchain-anthropic==0.0.1.post1
langchain-community==0.0.16
langchain-core==0.1.16
langchain-openai==0.0.2

PR #17162, opened 2024-02-07
langchain.core : Use shallow copy for schema manipulation in JsonOutputParser.get_format_instructions

L-cloud:
- **Description :**  

Fix: Use shallow copy for schema manipulation in get_format_instructions

Prevents side effects on the original schema object by using a dictionary comprehension for a safer and more controlled manipulation of schema key-value pairs, enhancing code reliability.

  - **Issue:**  #17161 
  - **Dependencies:** None
  -  **Twitter handle:** None


hwchase17 (commented):
can we add a comment about why this is neccessary, and a quick unit test

L-cloud:
@hwchase17 
Thank you for your comment! I have added test code to demonstrate a specific example that underscores the necessity of this change. The example provThank you for your contribution. I will now provide a detailed example of the specific changes required, drawn from an actual case I have encountered. This involves the use of the same Base Class in two scenarios: one utilizing an OpenAI function call and the other employing a JsonOutputParser.

```Python
class Joke(BaseModel):

joke_query = "Tell me a joke."

parser = JsonOutputParser(pydantic_object=Joke)

prompt = PromptTemplate(
    template="Answer the user query.\n{format_instructions}\n{query}\n",
    input_variables=["query"],
    partial_variables={"format_instructions": parser.get_format_instructions()}, #cause
)

chain = prompt | model | parser

chain.invoke({"query": joke_query})

openai_functions = [convert_to_openai_function(Joke)]

parser = JsonOutputFunctionsParser()

chain2 = prompt | model.bind(functions=openai_functions) | parser

chain2.invoke({"query": "tell me a joke"})
```
In this code, `chain2` triggers an error stating: `openai.BadRequestError: Error code: 400 - {'error': {'message': "'' does not match '^[a-zA-Z0-9_-]{1,64}$' - 'functions.0.name'", 'type': 'invalid_request_error', 'param': None, 'code': None}}`. The root cause of this error is the direct deletion of variables from the BaseModel schema via the `schema()` function in the `get_format_instructions` function, as shown below:
```python
schema = self.pydantic_object.schema()
reduced_schema = schema
            if "title" in reduced_schema:
                del reduced_schema["title"]
            if "type" in reduced_schema:
                del reduced_schema["type"]
```
The use of BaseModel's `.schema()` function in `convert_to_openai_function` results in default values of "" being assigned to the name and type, leading to the aforementioned error. Although this is the only case I have identified, I believe other functions utilizing BaseModel could similarly be at risk. Furthermore, the performance impact of not using the reduced_schema object directly is negligible, not affecting practical use:
```python
            schema_str = json.dumps(reduced_schema)
            return JSON_FORMAT_INSTRUCTIONS.format(schema=schema_str)
```

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -800,7 +800,7 @@ def handling(e: ValidationError) -> str:
     ],
 )
 async def test_async_validation_error_handling_non_validation_error(
-    handler: Union[bool, str, Callable[[ValidationError], str]]
+    handler: Union[bool, str, Callable[[ValidationError], str]],
```
L-cloud, line 803: It seems to have been modified upon applying the ruff format.

libs/core/langchain_core/output_parsers/json.py
```diff
@@ -221,7 +221,7 @@ def get_format_instructions(self) -> str:
         if self.pydantic_object is None:
             return "Return a JSON object."
         else:
-            schema = self.pydantic_object.schema()
+            schema = {k: v for k, v in self.pydantic_object.schema().items()}
```
hwchase17, line 224: a comment here would be great! otherwise looks good

L-cloud: @hwchase17  
I've applied the changes, thank you :)

L-cloud: I have also made the same changes in `/langchain/langchain/output_parsers/yaml.py` and `/langchain/langchain/output_parsers/pydantic.py`.

hwchase17 (approved):

PR #17163, opened 2024-02-07
langchain: Update quickstart.mdx - Fix 422 error in example with LangServe client code

schalkje:
**Description:**: Fix 422 error in example with LangServe client code

httpx.HTTPStatusError: Client error '422 Unprocessable Entity' for url 'http://localhost:8000/agent/invoke'


baskaryan (approved):
thank you @schalkje!

PR #17171, opened 2024-02-07
community[patch]: Graceful handling of redis errors in RedisCache and AsyncRedisCache

### LINKED ISSUE #16866: RedisCache does't handle errors from redis.
### Checked other resources

### Example Code

Same code as in the docs [here](https://python.langchain.com/docs/integrations/llms/llm_caching#redis-cache) 

```
  File "/Users/dingusagar/inference.py", line 181, in infer
    response = self.chain.invoke(inputs)
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/langchain/schema/runnable/base.py", line 1213, in invoke
    input = step.invoke(
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/langchain/chat_models/base.py", line 142, in invoke
    self.generate_prompt(
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/langchain/chat_models/base.py", line 459, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/langchain/chat_models/base.py", line 349, in generate
    raise e
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/langchain/chat_models/base.py", line 339, in generate
    self._generate_with_cache(
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/langchain/chat_models/base.py", line 500, in _generate_with_cache
    cache_val = llm_cache.lookup(prompt, llm_string)
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/langchain/cache.py", line 392, in lookup
    results = self.redis.hgetall(self._key(prompt, llm_string))
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/commands/core.py", line 4867, in hgetall
    return self.execute_command("HGETALL", name)
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/client.py", line 1255, in execute_command
    conn = self.connection or pool.get_connection(command_name, **options)
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/connection.py", line 1441, in get_connection
    connection.connect()
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/connection.py", line 704, in connect
    raise ConnectionError(self._error_message(e))
redis.exceptions.ConnectionError: Error 61 connecting to 127.0.0.1:6379. Connection refused.

During handling of the above exception, another exception occurred:

  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/connection.py", line 698, in connect
    sock = self.retry.call_with_retry(
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/retry.py", line 46, in call_with_retry
    return do()
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/connection.py", line 699, in <lambda>
    lambda: self._connect(), lambda error: self.disconnect(error)
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/connection.py", line 1089, in _connect
    sock = super()._connect()
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/connection.py", line 987, in _connect
    raise err
  File "/Users/dingusagar/envs/python_env/lib/python3.9/site-packages/redis/connection.py", line 975, in _connect
    sock.connect(socket_address)
ConnectionRefusedError: [Errno 61] Connection refused
```

### Description

I am using RedisCache as per the docs [here](https://python.langchain.com/docs/integrations/llms/llm_caching#redis-cache) 

Was testing how robust the system is if the redis connection fails for somereason or it goes out of memory. 
Looks like if the redis connection URL is not reachable, the system throws an error.

I wanted langchain to internally handle redis failure and do a direct API calls to the LLM on failure. we can ofcourse log the error. This will make the system more robust to failures. 

Does it make sense to add this feature? If so, I can help with raising a PR.  

### System Info

langchain==0.0.333

snsten:
- **Description:**
The existing `RedisCache` implementation lacks proper handling for redis client failures, such as `ConnectionRefusedError`, leading to subsequent failures in pipeline components like LLM calls. This pull request aims to improve error handling for redis client issues, ensuring a more robust and graceful handling of such errors.

  - **Issue:**  Fixes #16866
  - **Dependencies:** No new dependency
  - **Twitter handle:** N/A


libs/community/langchain_community/cache.py
```diff
@@ -392,22 +392,26 @@ def _key(self, prompt: str, llm_string: str) -> str:
...
+                            "properly. This is likely due to the cache being in an "
+                            "older format. Please recreate your cache to avoid this "
+                            "error."
+                        )
+                        # In a previous life we stored the raw text directly
+                        # in the table, so assume it's in that format.
+                        generations.append(Generation(text=text))
+        # Gotta catch'em all
+        except Exception as e:
+            logger.warning(f"Redis lookup failed: {e}")
```
eyurtsev, line 414: Isn't this an `error` level issue rather than a warning?

libs/community/langchain_community/cache.py
```diff
@@ -392,22 +392,26 @@ def _key(self, prompt: str, llm_string: str) -> str:
...
-                    logger.warning(
-                        "Retrieving a cache value that could not be deserialized "
-                        "properly. This is likely due to the cache being in an "
-                        "older format. Please recreate your cache to avoid this "
-                        "error."
-                    )
-                    # In a previous life we stored the raw text directly
-                    # in the table, so assume it's in that format.
-                    generations.append(Generation(text=text))
+        try:
```
eyurtsev, line 395: There are two other reasonable behaviors.

1) Retry once prior to giving up
2) Circuit breaker pattern (out of scope)

What are your thoughts on these?

libs/community/langchain_community/cache.py
snsten: Yes, will change the log level to error

libs/community/langchain_community/cache.py
snsten: 1. Adding retry with some timeout make sense.
2. About circuit breaker for long term failure scenario:
   - We keep logging errors while redis is down until circuit is open.
   - When circuit is open do the error logs continue for redis failure in this duration?
   Or 
   -  Circuit is open we log error that the circuit is open so redis is not being used?

Ideally it would be better to continue showing error logs since the api costs are incurred without cache, the service user should be aware of the redis failure even during the open circuit state from continued failure.

I can start with retry and close this and start a new pr for circuit breaker considering all the scenarios or do it here. Let me know what do you suggest.

eyurtsev: I think users would appreciate being able to configure the retry behavior since a retry involves some additional latency and isn't guaranteed to pay off.

Would you be able to suggest a parameterization for the initializer that you think would make sense to configure the retry behavior?

---

For a circuit breaker, I think it probably make sense to log periodically, but not every request. We should probably figure out a separate way to tackle this. I'd appreciate if we could design a circuit breaker as a higher level primitive that accepts a cache as an instance and dresses it up as a circuit breaker

---

cc @cbornet / @dzmitry-kankalovich  tagging you in case this is of interest for you

dzmitry-kankalovich: I think even as-is current change set is great, because cache layer IO should not break LLM calls in my opinion.

Taking it further, I do agree that more configuration around it would be nice, like the fact whether it should error out or not (default "not"?); and if retry to be added, then its definitely need to have configuration for that as well (which I'd probably set to no retry by default? you rightfully mentioned with cache retries you need to be sure this is what you want, as it can just defeat the purpose of a cache in the first place).

And yes, a general circuit breaker for cache interface would be great, but this is probably a substantial changeset, which IDK if @snsten would go for.

dzmitry-kankalovich:
I'd also add it to `AsyncRedisCache` class (should pop up in your branch once you rebase onto latest `master`)

libs/community/langchain_community/cache.py
eyurtsev: Agree on all points!

@snsten do you want to proceed with the PR as is and just update the logger.warning -> logger.error?  This will probably spam someone's logs at some point, but at least their service will stay up.

And if you want to tackle more configuration separately that will be great (or a circuit breaker).

snsten: Ok, will change the error log only in this pr for the two classes mentioned

snsten:
Added the error logs for `RedisCache` and `AsyncRedisCache` classes.

libs/community/langchain_community/cache.py
eyurtsev: Thank you!

eyurtsev (approved):

PR #17174, opened 2024-02-07
Adjusted get_num_tokens_from_messages method

lkuligin:
added a small adjustment so that's a single call of `get_num_tokens` instead of multiple calls since results are summed up anyways


baskaryan:
i think this is meant to account for the fact that chat models may have some tokens that they add to each message? for example if get_num_tokens always adds 3 to whatever token count of the given string is, then 

```python
sum([get_num_tokens(x) for x in array]) = get_num_tokens("".join(array)) + 3 * len(array)
```

PR #17181, opened 2024-02-07
lancedb template code

akashAD98:
as per contribution guidance verified all code format & submitting pr


efriis:
get `TypeError: Type is not JSON serializable: numpy.ndarray` when running it with question "where did harrison work"

efriis:
This is likely an issue in the LanceDB integration, not this template. It looks like it's attaching numpy arrays as metadata to the documents, thereby making them not serializable. Would recommend passing the vectors around as List[float] instead of numpy.ndarray!

But also, including the embeddings in the metadata of documents is generally unnecessary. Maybe only pass it back to the user if they request it with some parameter?

akashAD98:
Okay got it let me work on it

efriis:
Will close for now, and feel free to open a PR against the integration package, and then reopen this once that's in + published to community!

PR #17185, opened 2024-02-07
core[patch]: Replace memory stream implementation used by LogStreamCallbackHandler

eyurtsev:
This PR replaces the memory stream implementation used by the LogStreamCallbackHandler.

This implementation resolves an issue in which streamed logs and streamed events originating from sync code would arrive only after the entire sync code would finish execution (rather than arriving in real time as they're generated).

One example is if trying to stream tokens from an llm within a tool. If the tool was an async tool, but the llm was invoked via stream (sync variant) rather than astream (async variant), then the tokens would fail to stream in real time and would all arrived bunched up after the tool invocation completed.


libs/core/langchain_core/tracers/log_stream.py
```diff
@@ -210,12 +209,11 @@ def __init__(
         self.exclude_types = exclude_types
         self.exclude_tags = exclude_tags
 
-        send_stream: Any
-        receive_stream: Any
```
nfcampos, line 214: I think this syntax with square brackets doesn't exist in older py

eyurtsev: hmm i think it's there for classes? I think the issue originally was that the source library was using it with a function rather than a class.

Evidence is that all unit tests pass

libs/core/langchain_core/tracers/memory_stream.py
```diff
@@ -0,0 +1,111 @@
...
+        to the channel.
+        """
+        self._queue = queue
+        self._done = done
+        self._is_closed = False
+
+    async def __aiter__(self) -> AsyncIterator[T]:
+        while True:
+            if self._is_closed:
+                raise ClosedResourceError("Reader is closed. Cannot read from it.")
```
nfcampos, line 70: Do we gain anything from raising this error instead of ending loop?

libs/core/langchain_core/tracers/log_stream.py
```diff
@@ -228,7 +226,7 @@ def send(self, *ops: Dict[str, Any]) -> bool:
         try:
             self.send_stream.send_nowait(RunLogPatch(*ops))
             return True
-        except (ClosedResourceError, BrokenResourceError):
+        except ClosedResourceError:
```
nfcampos, line 229: I don't see this error being raised anywhere in the write path, only in the read path?

eyurtsev: ah yah, i'll just remove this for now. I suspect we'll need to add some error handling as we encounter some failures (e.g., loop closed etc)

nfcampos (approved):

PR #17190, opened 2024-02-07
community[patch]: Better error propagation for neo4jgraph

tomasonjo:
There are other errors that could happen when refreshing the schema, so we want to propagate specific errors for more clarity


jexp (approved):
LGTM

eyurtsev (approved):

PR #17205, opened 2024-02-07
Fix stream events/log with some kinds of non addable output

nfcampos:


hinthornw (approved):
Fixes the error for me - thanks!

PR #17213, opened 2024-02-08
google-genai[patch]: match function call interface

efriis:
should match vertex


libs/partners/google-genai/tests/integration_tests/test_function_call.py
```diff
@@ -29,6 +35,50 @@ def test_function_call() -> None:
     assert res.additional_kwargs
     assert "function_call" in res.additional_kwargs
     assert "get_weather" == res.additional_kwargs["function_call"]["name"]
-    arguments = res.additional_kwargs["function_call"]["arguments"]
-    assert isinstance(arguments, dict)
+    arguments_str = res.additional_kwargs["function_call"]["arguments"]
+    assert isinstance(arguments_str, str)
```
efriis, line 39: switched to string to match others

PR #17217, opened 2024-02-08
community[minor]: support query filters for NotionDBLoader

lyndseyjw:
- **Description:** Support filtering databases in the use case where devs do not want to query ALL entries within a DB, 
- **Issue:** N/A,
- **Dependencies:** N/A,
- **Twitter handle:** I don't have Twitter but feel free to tag my Github!


libs/community/langchain_community/document_loaders/notiondb.py
```diff
@@ -20,13 +20,17 @@ class NotionDBLoader(BaseLoader):
         database_id (str): Notion database id.
         request_timeout_sec (int): Timeout for Notion requests in seconds.
             Defaults to 10.
+        filter_object (Dict[str, Any]): Filter object used to limit returned
```
eyurtsev, line 23: Could you provide an example of how to use the filter object? (This is probably the most important thing  otherwise it's not really obvious how to use this functionality)

Is there also a URL to notion db API that could be included so folks could see the documentation?

libs/community/langchain_community/document_loaders/notiondb.py
```diff
@@ -20,13 +20,17 @@ class NotionDBLoader(BaseLoader):
...
+            entries based on specified criteria.
+            Defaults to empty dict, which will return ALL entries.
     """
 
     def __init__(
         self,
         integration_token: str,
         database_id: str,
         request_timeout_sec: Optional[int] = 10,
+        filter_object: Optional[Dict[str, Any]] = {}
```
eyurtsev, line 33: * New named arguments should be introduced after a `*` operator.
* `{}` is mutable

```suggestion
        *,
        filter_object: Optional[Dict[str, Any]] = None
```

libs/community/langchain_community/document_loaders/notiondb.py
```diff
@@ -42,6 +46,7 @@ def __init__(
             "Notion-Version": "2022-06-28",
         }
         self.request_timeout_sec = request_timeout_sec
+        self.filter_object = filter_object
```
eyurtsev, line 49: `{}` is a mutable, convention is to typically use a `None` as the default value

Common pattern to treat with the None assignment above

```python
self.filter_object = filter_object or {}
```

libs/community/langchain_community/document_loaders/notiondb.py
```diff
@@ -182,14 +191,19 @@ def _load_blocks(self, block_id: str, num_tabs: int = 0) -> str:
         return "\n".join(result_lines_arr)
 
     def _request(
-        self, url: str, method: str = "GET", query_dict: Dict[str, Any] = {}
+        self,
+        url: str,
+        method: str = "GET",
+        query_dict: Dict[str, Any] = {},
+        filter_object: Dict[str, Any] = {}
```
eyurtsev, line 198: Could you re-write like init to use `*` and `None`

eyurtsev (commented):
Thanks for the contribution! A few minor changes and we can merge :)

libs/community/langchain_community/document_loaders/notiondb.py
eyurtsev: We need to implementer `filter_object or {}` otherwise self.filter_object will be assigned None by default which presumably will not work with the API, same for the case below

eyurtsev (approved):

eyurtsev:
Ready for merge when tests pass

lyndseyjw:
@eyurtsev - thank you for taking the time to review! it looks like all checks have passed .. do you handle the merge? I don't see the option to

Issue #14711, opened 2023-12-14
'google.generativeai' has no attribute 'GenerativeModel'


kegbo:
### System Info

Anaconda  23.7.2
Windows 11
Python 3.11.5

### Who can help?

@hwchase17 @agola11 

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [X] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [X] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Method 1
```
def generate_lecture(topic:str, context:str):
    
  template ="""
  As an accomplished university professor and expert in {topic}, your task is to develop an elaborate, exhaustive, and highly detailed lecture on the subject. 
  Remember to generate content ensuring both novice learners and advanced students can benefit from your expertise.
  while leveraging the provided context
  
  Context: {context} """

  
  prompt = ChatPromptTemplate.from_template(template)

  model = ChatGoogleGenerativeAI(model="gemini-pro", google_api_key=palm_api_key)

 
  response = model.invoke(template)
  return response.content
```

Method 2
```

def generate_lecture(topic:str, context:str):
    
  template ="""
  As an accomplished university professor and expert in {topic}, your task is to develop an elaborate, exhaustive, and highly detailed lecture on the subject. 
  Remember to generate content ensuring both novice learners and advanced students can benefit from your expertise.
  while leveraging the provided context
  
  Context: {context} """

  
  prompt = ChatPromptTemplate.from_template(template)

  model = ChatGoogleGenerativeAI(model="gemini-pro", google_api_key=palm_api_key)
  chain = prompt| model | StrOutputParser()
  
  response = chain.invoke({"topic":topic, "context":context})

  return response.
```

Result we receive on execttion of method 1 and method 2 is 
```

  File "C:\Users\ibokk\RobotForge\mvp\service\llm.py", line 80, in generate_lecture
    model = ChatGoogleGenerativeAI(model="gemini-pro", google_api_key=palm_api_key)
[...]

PR #17224, opened 2024-02-08
langchain_google_genai : Add missing _identifying_params property.

GMartin-dev:
Description: Missing  _identifying_params create issues when dealing with callbacks to get current run model parameters.
All other model partners implementation provide this property and also provide _default_params. I'm not sure about the default values to include or if we can re-use the same as for _VertexAICommon(), this change allows you to access the model parameters correctly.
Issue: Not exactly this issue but could be related https://github.com/langchain-ai/langchain/issues/14711
Twitter handle:@musicaoriginal2

### DIFF
--- a/libs/partners/google-genai/langchain_google_genai/llms.py
+++ b/libs/partners/google-genai/langchain_google_genai/llms.py
@@ -151,6 +151,18 @@ Supported examples:
     def _model_family(self) -> str:
         return GoogleModelFamily(self.model)
 
+    @property
+    def _identifying_params(self) -> Dict[str, Any]:
+        """Get the identifying parameters."""
+        return {
+            "model": self.model,
+            "temperature": self.temperature,
+            "top_p": self.top_p,
+            "top_k": self.top_k,
+            "max_output_tokens": self.max_output_tokens,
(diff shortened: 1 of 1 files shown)


libs/partners/google-genai/langchain_google_genai/llms.py
```diff
@@ -151,6 +151,18 @@ def lc_secrets(self) -> Dict[str, str]:
     def _model_family(self) -> str:
         return GoogleModelFamily(self.model)
 
+    @property
+    def _identifying_params(self) -> Dict[str, Any]:
```
lkuligin, line 155: which method that is not overriden will use this property?

GMartin-dev: There is no much documentation about this property. At first I thought  it was used for serialization but is not really only that. When working with callbacks this property is used somehow to map the invocation parameters.
WIthout it you can't get the model name in callbacks events. 
Check:
https://github.com/search?q=org%3Alangchain-ai+_identifying_params&type=code

efriis (approved):

PR #17263, opened 2024-02-08
[Langchain_core]: Added Docstring for RunnableConfigurableAlternatives

keenborder786:
I noticed that RunnableConfigurableAlternatives which is an important composition in LCEL has no Docstring. Therefore I added the detailed Docstring for it.
@baskaryan, @eyurtsev, @hwchase17 please have a look and let me if the docstring is looking good.


baskaryan (approved):
thank you @keenborder786!

eyurtsev:
@keenborder786  amazing thank you! If you have time (can be in a separate PR) -- it would be nice to also document the corresponding methods that promopt a runnable into a configurable runnable

PR #17290, opened 2024-02-09
partner: nvidia-ai-endpoints model arguments (i.e. temperature) on construction bug

VKudlay:
- **Issue:** Issue with model argument support (been there for a while actually):
    - Non-specially-handled arguments like temperature don't work when passed through constructor.
    - Such arguments DO work quite well with `bind`, but also do not abide by field requirements.  
    - Since initial push, server-side error messages have gotten better and v0.0.2 raises better exceptions. So maybe it's better to let server-side handle such issues?
- **Description:**
    - Removed ChatNVIDIA's argument fields in favor of `model_kwargs`/`model_kws` arguments which aggregates constructor kwargs (from constructor pathway) and merges them with call kwargs (bind pathway).
    - Shuffled a few functions from `_NVIDIAClient` to `ChatNVIDIA` to streamline construction for future integrations. 
    - Minor/Optional: Old services didn't have stop support, so client-side stopping was implemented. Now do both.
- **Any Breaking Changes:** Minor breaking changes if you strongly rely on chat_model.temperature, etc. This is captured by chat_model.model_kwargs.

PR passes tests and example notebooks and example testing. Still gonna chat with some people, so leaving as draft for now.


VKudlay:
@efriis Ok! I think it's good. Please let me know thoughts

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/_common.py
```diff
@@ -439,10 +438,8 @@ class _NVIDIAClient(BaseModel):
     client: NVEModel = Field(NVEModel)
 
     model: str = Field(..., description="Name of the model to invoke")
-
-    temperature: float = Field(0.2, le=1.0, gt=0.0)
-    top_p: float = Field(0.7, le=1.0, ge=0.0)
-    max_tokens: int = Field(1024, le=1024, ge=32)
```
efriis, line 445: will this be a breaking change? Can you still access like `llm = ChatNVIDIA(); llm.temperature`?

Would be good to handle access case at a minimum. Setting case is less important to me but could be nice for backwards compat!

efriis: I have a slight preference for keeping these args top-level also if we're validating based on the list below as well. What do we gain by storing in a dict instead of as explicit pydantic properties?

efriis: Can remove validation with something like `temperature: float = 0.2`

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/_common.py
```diff
@@ -485,67 +491,3 @@ def get_model_details(self, model: Optional[str] = None) -> dict:
         known_fns = self.client.available_functions
         fn_spec = [f for f in known_fns if f.get("id") == model_key][0]
         return fn_spec
-
-    def get_generation(
```
efriis, line 489: Looks like these were all moved to chat model, and probably don't make sense for anything else inheriting from common?

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/embeddings.py
```diff
@@ -1,51 +1,23 @@
...
-        """Map the available models that can be invoked."""
-        return self.client.available_models
-
-    @staticmethod
-    def get_available_functions(**kwargs: Any) -> List[dict]:
-        """Map the available functions that can be invoked. Callable from class"""
-        return NVEModel(**kwargs).available_functions
-
-    @staticmethod
-    def get_available_models(**kwargs: Any) -> dict:
```
efriis, line 45: removing these is a breaking change right?

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/_common.py
VKudlay: An issue is that binding had no validation previously, which was the default way advertised in the docs.

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/_common.py
VKudlay: Yup! Just wanted to remove redundancy between the ChatNVIDIA and NVIDIAEmbeddings core components.

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/chat_models.py
```diff
@@ -116,6 +117,23 @@ class ChatNVIDIA(nvidia_ai_endpoints._NVIDIAClient, SimpleChatModel):
             response = model.invoke("Hello")
     """
 
+    model_kws: List[str] = [
```
efriis, line 120: delete

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/chat_models.py
```diff
@@ -229,7 +238,70 @@ def preprocess_msg(self, msg: BaseMessage) -> Dict[str, str]:
...
+        **kwargs: Any,
+    ) -> AsyncIterator:
+        """Call to client astream methods with call scope"""
+        stop = kwargs.get("stop", None)
+        payload = self.get_payload(inputs=inputs, stream=True, **kwargs)
+        return self.client.get_req_astream(self.model, stop=stop, payload=payload)
+
+    def get_payload(self, inputs: Sequence[Dict], **kwargs: Any) -> dict:
+        """Generates payload for the _NVIDIAClient API to send to service."""
+        const_kwargs = {k: getattr(self, k) for k in self.model_kws}
```
efriis, line 281:
```suggestion
        const_kwargs = {
            "temperature": self.temperature,
            "max_tokens": self.max_tokens,
            "top_p": self.top_p,
            "seed": self.seed,
            "bad": self.bad,
            "stop": self.stop,
            "labels": self.labels,
        }
```

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/embeddings.py
```diff
@@ -1,51 +1,23 @@
...
 
-class NVIDIAEmbeddings(BaseModel, Embeddings):
+class NVIDIAEmbeddings(_NVIDIAClient, Embeddings):
     """NVIDIA's AI Foundation Retriever Question-Answering Asymmetric Model."""
 
-    client: NVEModel = Field(NVEModel)
-    model: str = Field(
-        ..., description="The embedding model to use. Example: nvolveqa_40k"
-    )
+    model_kws: List[str] = ["max_length", "max_batch_size", "model_type"]
```
efriis, line 13: delete

efriis (commented):
One question and 1 nit!

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/chat_models.py
efriis: Also `const` might be the wrong prefix. Maybe `attr_kwargs`?

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/chat_models.py
VKudlay: Is necessary

efriis: For what?

VKudlay: https://catalog.ngc.nvidia.com/ai-foundation-models
[Mixtral_8x7b is pretty representative](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/ai-foundation/models/mixtral-8x7b/documentation)

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/chat_models.py
VKudlay: Oh, I see what you're asking...

libs/partners/nvidia-ai-endpoints/langchain_nvidia_ai_endpoints/chat_models.py
VKudlay: I'll yield. Your option can work for now. Thanks :D

VKudlay (commented):
Gotcha! Integrating the changes. Will @ you when they're in. 
The concern with bind/construction inconsistency goes away if we don't do client-side validation and make sure the construction arguments work, so in other words the original concern is now resolved thanks to your suggestion. Thanks!

VKudlay:
@efriis got it!

efriis (approved):

isVoid:
Hi, I was searching about non-canonical model parameters and stumbled across this PR. It seems like you can pass arbitrary model parameters to `ChatNVIDIA` as long as model supports it. Is that correct? The list of fields is a bit confusing to me.

VKudlay:
@isVoid Yeah, more or less. `.bind` is pretty much a force override, whereas the fields help enforce default construction arguments (and are also mostly there now because some frameworks rely on them for autodoc/completion). Advanced use cases are also supportable (and we're working on an OpenAPI update). Are you trying to create a custom connector for something?

isVoid:
>  Are you trying to create a custom connector for something?

Not really, I'm only using invoking the APIs via `ChatNVIDIA`. Specifically, I'm trying to configure `repetition_penalty` parameter for `code_llama2_70b` model.

PR #17291, opened 2024-02-09
langchain[patch]: structured output chain nits

baskaryan:


hinthornw (approved):
I like!

libs/langchain/langchain/chains/structured_output/base.py
```diff
@@ -103,17 +95,22 @@ class RecordDog(BaseModel):
...
     output_schema: Union[Dict[str, Any], Type[BaseModel]],
     llm: Runnable,
-    prompt: BasePromptTemplate,
+    prompt: Optional[BasePromptTemplate] = None,
     *,
     output_parser: Optional[Union[BaseOutputParser, BaseGenerationOutputParser]] = None,
-    mode: Literal["openai-functions", "openai-json"] = "openai-functions",
+    mode: Literal[
+        "openai-tools", "openai-functions", "openai-json"
+    ] = "openai-functions",
```
eyurtsev, line 113: I wonder whether we should've made this a required argument

eyurtsev (approved):
These changes look good to me. I think discovery of this chain is hard because the more intuitive names of extraction chain are already taken. Maybe we could add a deprecation decorator after i update the use case docs?

libs/langchain/langchain/chains/structured_output/base.py
baskaryan: yea, couldn't bc of backwards compat (mode arg was added after the function was created)

libs/langchain/langchain/chains/structured_output/base.py
```diff
@@ -103,17 +95,22 @@ class RecordDog(BaseModel):
...
 def create_structured_output_runnable(
     output_schema: Union[Dict[str, Any], Type[BaseModel]],
     llm: Runnable,
-    prompt: BasePromptTemplate,
+    prompt: Optional[BasePromptTemplate] = None,
     *,
     output_parser: Optional[Union[BaseOutputParser, BaseGenerationOutputParser]] = None,
-    mode: Literal["openai-functions", "openai-json"] = "openai-functions",
+    mode: Literal[
+        "openai-tools", "openai-functions", "openai-json"
```
baskaryan, line 112: should delete openai-tools, not part of this pr

PR #17296, opened 2024-02-09
langchain[minor]: openai tools structured_output_chain

baskaryan:


libs/langchain/langchain/output_parsers/openai_tools.py
```diff
@@ -22,6 +22,8 @@ class JsonOutputToolsParser(BaseGenerationOutputParser[Any]):
     """
     return_id: bool = False
     """Whether to return the tool call id."""
+    return_single: bool = False
```
eyurtsev, line 25: This doesn't feel like an option that we should add to the parsers

libs/langchain/langchain/chains/structured_output/base.py
```diff
@@ -106,15 +113,17 @@ class RecordDog(BaseModel):
...
 def create_structured_output_runnable(
     output_schema: Union[Dict[str, Any], Type[BaseModel]],
     llm: Runnable,
     prompt: BasePromptTemplate,
     *,
     output_parser: Optional[Union[BaseOutputParser, BaseGenerationOutputParser]] = None,
-    mode: Literal["openai-functions", "openai-json"] = "openai-functions",
-    enforce_single_function_usage: bool = True,
+    enforce_function_usage: bool = True,
+    return_single: bool = True,
```
eyurtsev, line 123: Why is this option needed?

libs/langchain/langchain/output_parsers/openai_tools.py
baskaryan: why not

libs/langchain/langchain/chains/structured_output/base.py
baskaryan: openai tools returns a list otherwise

ccurme:
Should we deprecate this? https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/chains/openai_tools/extraction.py

baskaryan:
> Should we deprecate this? https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/chains/openai_tools/extraction.py

probably. think we should first land https://github.com/langchain-ai/langchain/pull/17302 and figure out what, if any, other extraction chains/utils we're adding, and then deprecate the various existing extraction chains

libs/langchain/langchain/chains/structured_output/base.py
eyurtsev: Should we maybe hide this from the user as an option? It feels like a weird option in the API

libs/langchain/langchain/output_parsers/openai_tools.py
eyurtsev: Do you mind if we re-name this to: "first_tool_only" so it's clear how to interpret "single"

libs/langchain/langchain/chains/structured_output/base.py
eyurtsev: Can keep this as `return_single`

libs/langchain/langchain/output_parsers/openai_tools.py
```diff
@@ -22,8 +22,16 @@ class JsonOutputToolsParser(BaseGenerationOutputParser[Any]):
     """
     return_id: bool = False
     """Whether to return the tool call id."""
-    return_single: bool = False
-    """Whether to return only the first tool call."""
+    first_tool_only: bool = False
```
baskaryan, line 25: breaking

baskaryan: should probs alias

libs/langchain/langchain/chains/structured_output/base.py
```diff
@@ -299,33 +306,46 @@ class Dog(BaseModel):
                 chain = create_structured_output_runnable(Dog, llm, prompt, mode="openai-json")
                 chain.invoke({"input": "Harry was a chubby brown beagle who loved chicken"})
     """  # noqa: E501
+    # for backwards compatibility
```
baskaryan, line 320: i don't think we want to support this outside of openai-functions?

libs/langchain/langchain/output_parsers/openai_tools.py
eyurtsev: Why is this breaking? is this used somewhere already?

libs/langchain/langchain/chains/structured_output/base.py
eyurtsev: I can push inside, but I usually handle this at top scope, even better would be to fix this via a function level decorator to add a parameter renamer

libs/langchain/langchain/output_parsers/openai_tools.py
baskaryan: L76

libs/langchain/langchain/output_parsers/openai_tools.py
```diff
@@ -73,21 +85,36 @@ class JsonOutputKeyToolsParser(JsonOutputToolsParser):
 
     key_name: str
     """The type of tools to return."""
-    return_single: bool = False
-    """Whether to return only the first tool call."""
-
     def __init__(self, key_name: str, **kwargs: Any) -> None:
         """Allow init with positional args."""
+        # Backwards compatibility for old argument name.
+        if "return_single" in kwargs:
```
eyurtsev, line 91: Adding backwards compatibility

libs/langchain/langchain/chains/structured_output/base.py
```diff
@@ -130,19 +140,107 @@ def create_structured_output_runnable(
...
+                    fav_food: Optional[str] = Field(None, description="The dog's favorite food")
 
-    OpenAI functions example:
+                llm = ChatOpenAI(model="gpt-3.5-turbo-0125", temperature=0)
+                prompt = ChatPromptTemplate.from_messages(
+                    [
+                        ("system", "You are an extraction algorithm. Please extract every possible instance"), 
+                        ('human', '{input}')
+                    ]
+                )
```
baskaryan, line 185: would be nice to have an example without prompt

PR #17302, opened 2024-02-09
core[minor], openai[minor], langchain[patch]: output format on openai

baskaryan:
```python
class Foo(BaseModel):
  bar: str

structured_llm = ChatOpenAI().with_output_format(Foo)
```


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -706,21 +712,99 @@
...
-                formatted_tools[0]["function"]["name"]
-                != tool_choice["function"]["name"]
-            ):
+            if isinstance(tool_choice, str):
+                if tool_choice not in ("auto", "none"):
+                    tool_choice = {
+                        "type": "function",
+                        "function": {"name": tool_choice},
+                    }
+            elif isinstance(tool_choice, bool):
```
eyurtsev, line 727: This behavior is unexpected -- I don't think we should support bool unless only a single tool is provided

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -706,21 +712,99 @@
...
+    ) -> Runnable[LanguageModelInput, _BM]:
+        ...
+
+    def with_output_format(
+        self,
+        output_schema: Union[Dict[str, Any], Type[_BM]],
+        *,
+        mode: Literal["tools", "json"] = "tools",
+        enforce_schema: bool = True,
+        return_single: bool = True,
```
eyurtsev, line 774: What is `return_single`?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -706,21 +712,99 @@ def bind_tools(
...
+    ) -> Runnable[LanguageModelInput, _BM]:
+        ...
+
+    def with_output_format(
+        self,
+        output_schema: Union[Dict[str, Any], Type[_BM]],
+        *,
+        mode: Literal["tools", "json"] = "tools",
+        enforce_schema: bool = True,
```
eyurtsev, line 773: enforce_schema sounds like validation, but this is not a validation step rather model forced to do tool invocation.

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -706,21 +712,99 @@ def bind_tools(
...
+        return_single: bool = True,
+    ) -> Runnable[LanguageModelInput, _BM]:
+        ...
+
+    def with_output_format(
+        self,
+        output_schema: Union[Dict[str, Any], Type[_BM]],
```
eyurtsev, line 770: Do we have an opinion of what to do about multiple schemas?

baskaryan: what do you mean?

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: what's a better name? want something not too openai tools specific if we're going to add this method (or similar methods) to other models

libs/partners/openai/langchain_openai/chat_models/base.py
eyurtsev: The underlying API supports multiple tools which could be represented as different schemas

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -706,21 +712,99 @@ def bind_tools(
...
+
+    @overload
+    def with_output_format(
+        self,
+        output_schema: Type[_BM],
+        *,
+        mode: Literal["tools", "json"] = "tools",
+        enforce_schema: bool = True,
+        return_single: bool = True,
+    ) -> Runnable[LanguageModelInput, _BM]:
```
eyurtsev, line 765: Not sure that this should return pydantic model by default

To me at least there are 2 separate ideas: (i) schema specification, (ii) optional validation against the schema

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: would you just want to use bind_tools in that case?

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: for schema specification, not hard to turn pydantic -> dict if you don't want pydantic output

libs/core/langchain_core/language_models/base.py
```diff
@@ -59,9 +60,14 @@ def _get_token_ids_default_method(text: str) -> List[int]:
 LanguageModelLike = Runnable[LanguageModelInput, LanguageModelOutput]
 LanguageModelOutputVar = TypeVar("LanguageModelOutputVar", BaseMessage, str)
 
+_OutputSchema = TypeVar("_OutputSchema")
+_OutputFormat = TypeVar("_OutputFormat")
```
efriis, line 64: should we restrict these a bit? Right now could be anything right? Seems difficult to use with both.

Just an output base model as well and force it to be pydantic or general json type (list/dict/etc) doesn't seem too bad to me

efriis (approved):
I like it. Small q

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -693,21 +710,92 @@ def bind_tools(
...
+                f"'json_format'. Received: '{method}'"
+            )
+
+        if return_type == "raw":
+            return llm
+        elif return_type == "parsed":
+            return llm | output_parser
+        elif return_type == "all":
+            parser_assign = RunnablePassthrough.assign(
+                parsed=itemgetter("raw") | output_parser, parsing_error=lambda _: None
```
eyurtsev, line 786: (nit) the fallback presumably will affect the trace as well -- are we sure we don't want the parser to just return None?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -693,21 +710,92 @@ def bind_tools(
...
+        output_schema: _OutputSchema,
+        *,
+        method: Literal["function_calling", "json_format"] = "function_calling",
+        return_type: Literal["raw", "parsed", "all"] = "parsed",
+        **kwargs: Any,
+    ) -> Runnable[LanguageModelInput, _OutputFormat]:
+        """"""
+        if kwargs:
+            raise ValueError(f"Received unsupported arguments {kwargs}")
+        is_pydantic_schema = _is_pydantic_class(output_schema)
```
eyurtsev, line 755: I am still in favor of making a conceptual distinction between schema declaration vs. the parsing.

For example, I will usually want to declare my schema with pydantic, but I don't necessarily want the output to be in pydantic (e.g., i want to stream the structured json out)

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -693,21 +710,92 @@ def bind_tools(
...
-                    f"provided tool was {formatted_tools[0]['function']['name']}."
+                    f"Unrecognized tool_choice type. Expected str, bool or dict. "
+                    f"Received: {tool_choice}"
                 )
             kwargs["tool_choice"] = tool_choice
         return super().bind(tools=formatted_tools, **kwargs)
+
+    def with_output_format(
+        self,
+        output_schema: _OutputSchema,
```
eyurtsev, line 746: Is there an escape hatch to say that nothing can be extracted? 

tool_choice = True -> so tool call will always be forced

The edge case that we need to handle:

* Schema is a single instance rather than a collection
* Goal for example: identify the most qualified candidate based on the following list of candidates and their qualifications

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: i almost think this would be clearer? you could explicitly see that there was an error

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: you can always `dict_schema = convert_to_openai_tool(pydantic_schema)` before passing in schema? i just don't like adding a parameter until we know it's really needed. we can always add it later

eyurtsev: OK that will work. We can add a `validator='auto'" parameter later on, allowing one to set `validator=None` or `validator=Type[BaseModel]` or something along those lines?

baskaryan (commented):
- mv PydanticOutputParser from langchain -> core
- update OpenAI tools parsers to support return_single
- cp updated OpenAI tools parsers from langchain -> openai
- introduce FormattedOutputMixin
- update ChatOpenAI to implement FormattedOutputMixin

libs/core/langchain_core/language_models/output_format.py
```diff
@@ -0,0 +1,18 @@
...
+
+from langchain_core.language_models import LanguageModelInput
+from langchain_core.runnables import Runnable
+
+_OutputSchema = TypeVar("_OutputSchema")
+_FormattedOutput = TypeVar("_FormattedOutput")
+
+
+class FormattedOutputMixin(Generic[_OutputSchema, _FormattedOutput], ABC):
+    """Mixing for language models that offer native output formatting."""
```
efriis, line 12:
```suggestion
    """Mixin for language models that offer native output formatting."""
```

libs/core/langchain_core/language_models/output_format.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC, abstractmethod
+from typing import Any, Generic, TypeVar
+
+from langchain_core.language_models import LanguageModelInput
+from langchain_core.runnables import Runnable
+
+_OutputSchema = TypeVar("_OutputSchema")
+_FormattedOutput = TypeVar("_FormattedOutput")
```
efriis, line 8: Same q if we should restrict these a bit / somehow infer them based off each other. Seems a bit difficult to define FormattedOutput currently

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -189,7 +203,18 @@ class _FunctionCall(TypedDict):
     name: str
 
 
-class ChatOpenAI(BaseChatModel):
+_BM = TypeVar("_BM", bound=BaseModel)
+_OutputSchema = Union[Dict[str, Any], Type[_BM]]
+_FormattedOutput = Union[BaseMessage, Dict, _BM]
```
efriis, line 208: Would it be simpler if we had these be the types for any model that supports structured outputs, so each implementation doesn't have to worry about generics? Just whether or not it supports structured outputs?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -693,21 +718,214 @@ def bind_tools(
...
+    ) -> Runnable[LanguageModelInput, _AllFormattedOutput]:
+        ...
+
+    @overload
+    def with_output_format(
+        self,
+        schema: _OutputSchema,
+        *,
+        method: Literal["function_calling", "json_mode"] = "function_calling",
+        return_type: Literal["parsed"] = "parsed",
```
efriis, line 769: The different defaults on the overloads are a bit confusing. Shouldn't defaults be the same as the catch-all definition?

efriis: i.e. Shouldn't return_type default always be `parsed` or not have a default? For the overload of `return_type: Literal["all"]` I think the user *has* to have it manually defined, so maybe shouldn't have a default?

baskaryan: the types can't overlap for overloads

libs/core/langchain_core/language_models/structured_output.py
```diff
@@ -0,0 +1,18 @@
...
+from typing import Any, Dict, Generic, TypeVar, Union
+
+from langchain_core.language_models import LanguageModelInput
+from langchain_core.pydantic_v1 import BaseModel
+from langchain_core.runnables import Runnable
+
+_OutputSchema = TypeVar("_OutputSchema")
+
+
+class StructuredOutputMixin(Generic[_OutputSchema], ABC):
```
hwchase17, line 11: we are back to NOT putting this on base class? why not?

libs/core/langchain_core/language_models/structured_output.py
```diff
@@ -0,0 +1,18 @@
+from abc import ABC, abstractmethod
+from typing import Any, Dict, Generic, TypeVar, Union
+
+from langchain_core.language_models import LanguageModelInput
+from langchain_core.pydantic_v1 import BaseModel
+from langchain_core.runnables import Runnable
+
+_OutputSchema = TypeVar("_OutputSchema")
```
hwchase17, line 8: why do we need this? IMO this doesnt really add that much and makes it harder to user

libs/core/langchain_core/language_models/structured_output.py
baskaryan: will only be documented on classes that implement it

libs/core/langchain_core/language_models/structured_output.py
baskaryan: harder for user or contributor? don't think affects user experience?

hwchase17: contributor + any user that tries to read the code

libs/core/langchain_core/language_models/structured_output.py
hwchase17: so the downside of putting on base class (and defaulting to not implemented) is that it will show up in the documentation for all classes? that doesnt seem that bad...

libs/core/langchain_core/language_models/structured_output.py
baskaryan: in the places the mixin is implemented there won't be a typevar, so what's the hard part for someone reading code? 

don't feel super strongly on this though if we think typevar is that bad

libs/core/langchain_core/language_models/structured_output.py
baskaryan: what's downside of mixin?

eyurtsev: It's not that it shows up in documentation, but that it's hard to discover which models actually implement the interface.

After we implement, this we'll need to figure out how users will discover that this is an option

libs/langchain/langchain/output_parsers/openai_tools.py
```diff
@@ -22,6 +22,8 @@ class JsonOutputToolsParser(BaseGenerationOutputParser[Any]):
     """
     return_id: bool = False
     """Whether to return the tool call id."""
+    return_single: bool = False
```
eyurtsev, line 25: Do we think that `return_single` is generic enough to warrant introducing it into the parser? Is this the right place? How does this interact with streaming? Would unpacking at the caller code work?

----

Also if we do keep it here is there a better name? e.g., `first_tool_only`?

The doc-string should probably explain why this is necessary?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -693,21 +718,214 @@ def bind_tools(
...
+        is_pydantic_schema = _is_pydantic_class(schema)
+        if method == "function_calling":
+            llm = self.bind_tools([schema], tool_choice=True)
+            if is_pydantic_schema:
+                output_parser: OutputParserLike = PydanticToolsParser(
+                    tools=[schema], return_single=True
+                )
+            else:
+                key_name = convert_to_openai_tool(schema)["function"]["name"]
+                output_parser = JsonOutputKeyToolsParser(
```
eyurtsev, line 896: why not use a runnable lambda here to unpack? is it to make a clearer trace?

eyurtsev:
This fails with an opaque message due to missing doc-string

```python

model = ChatOpenAI()

class Person(BaseModel):
    name: str
    age: int

model.with_structured_output(Person).invoke('hello my name is chester')
```


```
BadRequestError: Error code: 400 - {'error': {'message': '"Usage docs: https://docs.pydantic.dev/2.6/concepts/models/ A base class for creating Pydantic models. Attributes:\\n    __class_vars__: The names of classvars defined on the model.\\n    __private_attributes__: Metadata about the private attributes of the model.\\n    __signature__: The signature for instantiating the model.     __pydantic_complete__: Whether model building is completed, or if there are still undefined fields.\\n    __pydantic_core_schema__: The pydantic-core schema used to build the SchemaValidator and SchemaSerializer.\\n    __pydantic_custom_init__: Whether the model has a custom `__init__` function.\\n    __pydantic_decorators__: Metadata containing the decorators defined on the model.\\n        This replaces `Model.__validators__` and `Model.__root_validators__` from Pydantic V1.\\n    __pydantic_generic_metadata__: Metadata for generic models; contains data used for a similar purpose to\\n        __args__, __origin__, __parameters__ in typing-module generics. May eventually be replaced by these.\\n    __pydantic_parent_namespace__: Parent namespace of the model, used for automatic rebuilding of models.\\n    __pydantic_post_init__: The name of the post-init method for the model, if defined.\\n    __pydantic_root_model__: Whether the model is a `RootModel`.\\n    __pydantic_serializer__: The pydantic-core SchemaSerializer used to dump instances of the model.\\n    __pydantic_validator__: The pydantic-core SchemaValidator used to validate instances of the model.     __pydantic_extra__: An instance attribute with the values of extra fields from validation when\\n        `model_config[\'extra\'] == \'allow\'`.\\n    __pydantic_fields_set__: An instance attribute with the names of fields explicitly set.\\n    __pydantic_private__: Instance attribute with the values of private attributes set on the model instance." is too long - \'tools.0.function.description\'', 'type': 'invalid_request_error', 'param': None, 'code': None}}```

Issue #17306, opened 2024-02-09
Chat message history with postgres failing when destination table has explicit schema


JeevansSP:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

```python
from langchain.memory import PostgresChatMessageHistory
connection_string = ""
history = PostgresChatMessageHistory(
    connection_string=connection_string,
    session_id="trial1",
    table_name= "schema.table"
)

history.add_user_message("Msg user 1")

history.add_ai_message("Msg AI 1")
print(history.messages)
```

### Error Message and Stack Trace (if applicable)

```error
Traceback (most recent call last):
  File "d:\xtransmatrix\SmartSurgn\SessionHandler\test.py", line 9, in <module>
    history.add_user_message("Msg user 1")
  File "D:\xtransmatrix\SmartSurgn\env\lib\site-packages\langchain\schema\chat_history.py", line 46, in add_user_message
    self.add_message(HumanMessage(content=message))
  File "D:\xtransmatrix\SmartSurgn\env\lib\site-packages\langchain\memory\chat_message_histories\postgres.py", line 66, in add_message
    self.cursor.execute(
  File "D:\xtransmatrix\SmartSurgn\env\lib\site-packages\psycopg\cursor.py", line 732, in execute
    raise ex.with_traceback(None)
psycopg.errors.UndefinedTable: relation "smartsurgn.msg_history" does not exist
LINE 1: INSERT INTO "schema.table" (session_id, message) V...
                    ^
```

### Description

we are asked to provide a table name to store the chat message history which stores the table by default in the public schema, if I want to store it in a different schema then I must be able to provide a table name as `<schema_name>.<table_name>` but when I do this it runs into the above error. Note: the table gets created in the right schema but runs into error when trying to write to it

### System Info

`langchain==0.1.5`

JeevansSP:
1. `Postgres version: `PostgreSQL 14.10 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 7.3.1 20180712 (Red Hat 7.3.1-12), 64-bit`
2. `database - postgres, schema - test_schema, table_name - msg_history`, by default `PostgresChatMessageHistory` creates the below table 

```python
f"""CREATE TABLE IF NOT EXISTS {self.table_name} (
[...]

PR #17307, opened 2024-02-09
community: Refactor SQL query string formatting

JeevansSP:
- **Description:** Refactor SQL query string formatting to use f-strings which lets user work with custom schemas
  - **Issue:** the issue #17306 
  - **Dependencies:** None
  - **Twitter handle:** @jeevntrollrdude


eyurtsev (changes requested):
The original code is correct. It avoids a sql injection.

PR #17316, opened 2024-02-09
community[minor]: add exclude parameter to DirectoryLoader

nejch:
- **Description:** adds an `exclude` parameter to the DirectoryLoader class, based on similar behavior in GenericLoader
  - **Issue:** discussed in https://github.com/langchain-ai/langchain/discussions/9059 and I think in some other issues that I cannot find at the moment :bow: 
  - **Dependencies:** None
  - **Twitter handle:** don't have one sorry! Just https://github/nejch


libs/community/langchain_community/document_loaders/directory.py
```diff
@@ -32,6 +32,7 @@ def __init__(
         self,
         path: str,
         glob: str = "**/[!.]*",
+        exclude: Sequence[str] = (),
```
eyurtsev, line 35: This is a breaking change since it's a positional argument. It needs to be added as a named argument after the `*`

libs/community/langchain_community/document_loaders/directory.py
```diff
@@ -51,6 +52,7 @@ def __init__(
             path: Path to directory.
             glob: Glob pattern to use to find files. Defaults to "**/[!.]*"
                (all files except hidden).
+            exclude: patterns to exclude from results, use glob syntax
```
eyurtsev, line 55: Could you provide an example of how to exclude something?

libs/community/langchain_community/document_loaders/directory.py
```diff
@@ -69,6 +71,7 @@ def __init__(
             loader_kwargs = {}
         self.path = path
         self.glob = glob
+        self.exclude = exclude
```
eyurtsev, line 74: I'd suggest doing a run time check here to verify that exclude is actually a list. The variable name sounds like a string, so likely API usage will be a mixture of sequence[str] as well as str from users. With a run time check, it will be easier to surface an error to users that's informative

libs/community/langchain_community/document_loaders/directory.py
nejch: Ah of course, totally missed that :sweat_smile: thanks.

libs/community/langchain_community/document_loaders/directory.py
nejch: @eyurtsev this is a direct copy from the docstring in the file system loader: https://github.com/langchain-ai/langchain/blob/e10030e2417bfdf540bb2e528a8b301cf6377b95/libs/community/langchain_community/document_loaders/blob_loaders/file_system.py#L68

Should I update all occurrences to keep them consistent?

libs/community/langchain_community/document_loaders/directory.py
nejch: :+1: same question as above: should I add it only here or in the other loader as well?

https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/document_loaders/blob_loaders/file_system.py#L111

nejch:
Thanks for the quick review @eyurtsev! Just noticed the tests also need some optional deps in CI, will rework this ASAP.

libs/community/langchain_community/document_loaders/directory.py
nejch: Made this a keyword-only argument.

libs/community/langchain_community/document_loaders/directory.py
nejch: I went ahead and just added this here. It was easy to just handle both, so I made strings an option as well and added a test.

libs/community/langchain_community/document_loaders/directory.py
nejch: I added some docstrings examples based on other loaders.

libs/community/langchain_community/document_loaders/directory.py
eyurtsev: Thank you!

eyurtsev (approved):

eyurtsev:
Can merge if tests pass

libs/community/pyproject.toml
```diff
@@ -92,6 +92,7 @@ hdbcli = {version = "^2.19.21", optional = true}
 oci = {version = "^2.119.1", optional = true}
 rdflib = {version = "7.0.0", optional = true}
 nvidia-riva-client = {version = "^2.14.0", optional = true}
+unstructured = {version = "^0.12.4", optional = true, python = ">=3.9.0,<3.12"}
```
eyurtsev, line 95: Could you add this to the extended_testing group below in the tool.poetry.extras below?

nejch: @eyurtsev it's already there, or am I missing anything else?

https://github.com/langchain-ai/langchain/blob/143d0077d8e8ddb8ea4cab8b665f673d8bab2d48/libs/community/pyproject.toml#L261

nejch: Ah, looks like I forgot to regenerate the lockfile after adding the extra, just pushed again :+1:

eyurtsev:
Going to help resolve conflicts so we can merge

eyurtsev:
Can merge when tests pass

nejch:
Sorry got the wrong lockfile updated there, I was going to get back to the dependency issue but I see you already got around that by mocking it out @eyurtsev, I guess next time this should be ready ;)

eyurtsev:
@nejch awesome! thank you. We need to investigate what's going with the extra dependency, but didn't want that to block from getting your contribution merged in!

nejch:
Perfect, thanks a lot @eyurtsev. Looks like it's green now :+1:

PR #17320, opened 2024-02-09
community: Add document manager and mongo document manager

2jimoo:
- **Description:** 
    - Add DocumentManager class, which is a nosql record manager. 
    - In order to use index and aindex in libs/langchain/langchain/indexes/_api.py, DocumentManager inherits RecordManager.
    - Also I added the MongoDB implementation of Document Manager too.
  - **Dependencies:** pymongo, motor


2jimoo:
@eyurtsev  May I ask if the PR is in progress?

libs/community/langchain_community/indexes/base.py
```diff
@@ -170,3 +170,171 @@ async def adelete_keys(self, keys: Sequence[str]) -> None:
         Args:
             keys: A list of keys to delete.
         """
+
+
+class DocumentManager(RecordManager, ABC):
```
eyurtsev, line 175: @2jimoo it seems that this has an identical interface to the record manager. Are you OK if we remove it?

libs/community/tests/integration_tests/indexes/docker-compose/mongodb.yml
```diff
@@ -0,0 +1,11 @@
+version: '3.1'
```
eyurtsev, line 1: Would you be able to move this to langchain/docker/docker-compose.yml?

eyurtsev (commented):
This code looks great! Just a question about whether we can remove the DocumentManager abstract interface and I think we can merge in

libs/community/langchain_community/indexes/base.py
2jimoo: Yes, it's okay to remove it.

libs/community/tests/integration_tests/indexes/docker-compose/mongodb.yml
2jimoo: Sure, I'll do that. Thanks for the review!

eyurtsev (approved):

eyurtsev:
* Resolved merge conflicts
* updated auth information and port
* removed pytest async marker (since we use a plugin to automatically do that)

We can merge this as soon as unit tests pass

PR #17328, opened 2024-02-09
community: Add explicit error message to Bedrock error output.

Mercurrent:
- **Description:** Propagate Bedrock errors into Langchain explicitly. Use-case: unset region error is hidden behind 'Could not load credentials...' message
  - **Issue:** [17654](https://github.com/langchain-ai/langchain/issues/17654)
  - **Dependencies:** None


libs/community/langchain_community/llms/bedrock.py
```diff
@@ -335,11 +335,13 @@ def validate_environment(cls, values: Dict) -> Dict:
                 "Could not import boto3 python package. "
                 "Please install it with `pip install boto3`."
             )
+        except ValueError as e:
+            raise ValueError(f"Error raised by bedrock service: {e}")
```
baskaryan, line 339: should we just reraise?
```
except ValueError as e:
    raise e
```

Mercurrent: Wouldn't it be more transparent, that the error is not Langchain native, but side-service originated?

The exact error wording is "Did not find region_name, please add an environment variable `AWS_DEFAULT_REGION` which contains it, or pass `region_name` as a named parameter"

Mercurrent:
Added error propagation to other classes affected same way. Added issue with description.

Mercurrent:
@baskaryan friendly ping

Issue #17329, opened 2024-02-09
asimilarity_search_with_relevance_scores returns NotImplementedError with AzureSearch 


dokato:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

```python
import os
import sys

import langchain_community
import langchain_core
from langchain_openai import AzureOpenAIEmbeddings
from langchain_community.vectorstores.azuresearch import AzureSearch

embeddings = AzureOpenAIEmbeddings(...)
vectordb = AzureSearch(...)

retriever = vectordb.as_retriever(search_type="similarity_score_threshold", search_kwargs={'score_threshold': 0.8})
#retriever = vectordb.as_retriever() # this works !

print(type(retriever))
query="what is the capital of poland?"
print(len(retriever.vectorstore.similarity_search_with_relevance_scores(query)))

import asyncio
async def f():
    await retriever.vectorstore.asimilarity_search_with_relevance_scores(query)

loop = asyncio.get_event_loop()
print(
    loop.run_until_complete(f())
)
```

### Error Message and Stack Trace (if applicable)

```
  File "/Users/user01/miniconda3/envs/genai/lib/python3.11/site-packages/langchain/chains/retrieval_qa/base.py", line 232, in _aget_docs
    return await self.retriever.aget_relevant_documents(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user01/miniconda3/envs/genai/lib/python3.11/site-packages/langchain_core/retrievers.py", line 280, in aget_relevant_documents
    raise e
  File "/Users/user01/miniconda3/envs/genai/lib/python3.11/site-packages/langchain_core/retrievers.py", line 273, in aget_relevant_documents
    result = await self._aget_relevant_documents(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user01/miniconda3/envs/genai/lib/python3.11/site-packages/langchain_core/vectorstores.py", line 679, in _aget_relevant_documents
    await self.vectorstore.asimilarity_search_with_relevance_scores(
[...]

Issue #11539, opened 2023-10-09
The ElasticsearchStore implementation is not correct.


via-red:
### System Info

langchain==0.0.310

### Who can help?

@joemcelroy ：It seems that the method _select_relevance_score_fn was forgotten to be implemented in vectorstores-elasticsearch.

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [X] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

elastic_vector_search = ElasticsearchStore(
    index_name='langchain-demo',
    es_connection=es,
    embedding=embedding,
)
    
retriever = elastic_vector_search.as_retriever(
        search_type="similarity_score_threshold",
        search_kwargs={'score_threshold': 0.8})

error log：
line 268, in similarity_search_with_relevance_scores
    docs_and_similarities = self._similarity_search_with_relevance_scores(
  File "/usr/local/lib/python3.10/site-packages/langchain/schema/vectorstore.py", line 242, in _similarity_search_with_relevance_scores
    relevance_score_fn = self._select_relevance_score_fn()
  File "/usr/local/lib/python3.10/site-packages/langchain/schema/vectorstore.py", line 211, in _select_relevance_score_fn
    raise NotImplementedError

### Expected behavior

I hope to implement the method and submit the changes as soon as possible. Thank you for your understanding.

via-red:
code：
elastic_vector_search = ElasticsearchStore(
index_name='langchain-demo',
es_connection=es,
embedding=embedding,
)

retriever = elastic_vector_search.as_retriever(
search_type="similarity_score_threshold",
search_kwargs={'score_threshold': 0.8})
documents = retriever.get_relevant_documents(query=keyword)//The place where the error occurred.

joemcelroy:
thanks! Happy to review the PR :). Should be relatively straightforward with `min_score`.

The only gotcha AFAIK is hybrid search which doesn't have a score. I will find out the best way to handle this.

via-red:
[...]

PR #17330, opened 2024-02-09
core: Fixes default asimilarity_search_with_relevance_scores call

dokato:
**Description:**
While using AzureSearch I realized that the `asimilarity_search_with_relevance_scores` is a direct copy of `similarity_search_with_relevance_scores`. When child class inherits `VectorStore` and overrides `similarity_search_with_relevance_scores` but without async version, then it refers to a private method that is not always defined and results with NotImplementedError.

Given that this is a problem affecting also others: #11539 ,  #3242 maybe it's okay to have a default implementation of async calls that wrap the non-async versions with `run_in_executor`. The child classes can amend it if needed.
Sorry I didn't find any unittests that cover that, but happy to add if suggested.

Details described in ⏬ 
 **Issue:** #17329
 **Dependencies:**Nope
 **Twitter handle:** dokatox

### DIFF
--- a/libs/core/langchain_core/vectorstores.py
+++ b/libs/core/langchain_core/vectorstores.py
@@ -300,7 +300,6 @@ class VectorStore(ABC):
             List of Tuples of (doc, similarity_score)
         """
         score_threshold = kwargs.pop("score_threshold", None)
-
         docs_and_similarities = self._similarity_search_with_relevance_scores(
             query, k=k, **kwargs
         )
@@ -346,32 +345,11 @@ class VectorStore(ABC):
         Returns:
             List of Tuples of (doc, similarity_score)
         """
-        score_threshold = kwargs.pop("score_threshold", None)
+        # This is a temporary workaround to make the similarity search with relevance scores
+        # asynchronous. The proper solution is to make the similarity search
+        # asynchronous in the vector store implementations.
+        return await run_in_executor(None, self.similarity_search_with_relevance_scores, query, k=k, **kwargs)
 
-        docs_and_similarities = await self._asimilarity_search_with_relevance_scores(
-            query, k=k, **kwargs
-        )
-        if any(
-            similarity < 0.0 or similarity > 1.0
-            for _, similarity in docs_and_similarities
-        ):
-            warnings.warn(
-                "Relevance scores must be between"
-                f" 0 and 1, got {docs_and_similarities}"
-            )
-
-        if score_threshold is not None:
-            docs_and_similarities = [
-                (doc, similarity)
-                for doc, similarity in docs_and_similarities
-                if similarity >= score_threshold
-            ]
-            if len(docs_and_similarities) == 0:
-                warnings.warn(
-                    "No relevant docs were retrieved using the relevance score"
-                    f" threshold {score_threshold}"
-                )
-        return docs_and_similarities
 
     async def asimilarity_search(
         self, query: str, k: int = 4, **kwargs: Any
--- a/libs/core/langchain_core/vectorstores.py
+++ b/libs/core/langchain_core/vectorstores.py
@@ -345,10 +345,12 @@ class VectorStore(ABC):
         Returns:
             List of Tuples of (doc, similarity_score)
         """
-        # This is a temporary workaround to make the similarity search with relevance scores
-        # asynchronous. The proper solution is to make the similarity search
-        # asynchronous in the vector store implementations.
-        return await run_in_executor(None, self.similarity_search_with_relevance_scores, query, k=k, **kwargs)
+        # This is a temporary workaround to make the similarity search with 
+        # relevance scores asynchronous. The proper solution is to make the 
+        # similarity search asynchronous in the vector store implementations.
+        return await run_in_executor(
+            None, self.similarity_search_with_relevance_scores, query, k=k, **kwargs
+        )
 
 
     async def asimilarity_search(


dokato:
as asked in the PR message, pinging @baskaryan for review, thanks!

libs/core/langchain_core/vectorstores.py
```diff
@@ -346,32 +345,13 @@ async def asimilarity_search_with_relevance_scores(
         Returns:
             List of Tuples of (doc, similarity_score)
         """
-        score_threshold = kwargs.pop("score_threshold", None)
-
-        docs_and_similarities = await self._asimilarity_search_with_relevance_scores(
```
baskaryan, line 351: i believe _asimilarity_search_with_relevance_scores already has a default implementation (that asynchronously runs _similarity_search_with_relevance_scores), so why is this change needed?

dokato: The point of this change is that it shouldn't default to the previous implementation. Here's why:
- `_asimilarity_search_with_relevance_scores` makes some hidden assumptions about the existence of `_select_relevance_score_fn` method. Some classes that inherit from the vector score extend `similarity_search_with_relevance_scores` and without updating `asimilarity_search_with_relevance_scores` it gives errors, see #17329 for azure, and #11539 for elastic search.
- rather than fixing children classes by c&p we can fix the root of the problem here and avoid similar issues in case someone decides to inherit in the future.
- `asimilarity_search_with_relevance_scores` is basically copy and paste from `similarity_search_with_relevance_scores`

baskaryan: doesn't _similarity_search_with_relevance_scores, which default implementation of similarity_search_with_relevance_scores relies on, make same assumption about _select_relevance_score_fn? think the correct fix would be for classes that overwrite similarity_search_with_relevance_scores and don't implement _select_relevance_score_fn to also overwrite asimilarity_search_with_relevance_scores (could be as simple as using the implementation you've provided here). but don't think that belongs on the base class

dokato: > doesn't _similarity_search_with_relevance_scores, which default implementation of similarity_search_with_relevance_scores relies on, make same assumption about _select_relevance_score_fn?

It does, but if child class overrides `similarity_search_with_relevance_scores` such that it doesn't use `_similarity_search_with_relevance_scores` then it doesn't matter.

> think the correct fix would be for classes that overwrite similarity_search_with_relevance_scores and don't implement _select_relevance_score_fn to also overwrite asimilarity_search_with_relevance_scores (could be as simple as using the implementation you've provided here). 

So what you are saying is we need to fix **all** children classes (note that this affected not only azuresearch but elastic as well, maybe others that are not tested too), rather than fixing one parent class. To me this sounds against DRY principle, but if you stay strongly about it happy to revamp this change to fix my Azure problem and let the others worry about their async communications. 

I still think my proposal is more robust fix that prevents similar problems to occur in the future and gives flexibility to implement custom async functions as needed. Plus, if you look what `asimilarity_search_with_score`, `asimilarity_search`, and `asimilarity_search_by_vector` are doing is exactly that. I don't see why `asimilarity_search_with_relevance_scores` should be an exception?

baskaryan: i think async relying on sync should only happen at the lowest level functions if at all

```python
class FooBar(ABC):
  def foo(self):
    raise NotImplementedError()
 
  async def afoo(self):
    # child classes should overwrite if they have native async support
    return await run_in_executor(None, self.foo, ...)
    
  def foo_plus_2(self):
    return self.foo() + 2
    
  # what i think we should do
  async def afoo_plus_2(self):
    return (await self.afoo()) + 2

  # what i think we should *not* do
  async def afoo_plus_2(self):
    return await run_in_executor(None, self.foo_plus_2, ...)
```

what do you think? cc @eyurtsev @nfcampos as well

baskaryan: what's unfortunate here is some integrations overwrite equivalent of `foo_plus_2` without ever implementing `foo`, which is what's causing us problems now

baskaryan: the reason i prefer `afoo_plus_2` relying on `afoo` instead of `foo_plus_2` is that if someone implements `afoo` to use native async functionality then `afoo_plus_2` will automatically get that as well, without having to also be overwritten

dokato: Yer, normally I'd agree with you but seems like we have empirical evidence already that `afoo_plus_2` is not always `self.afoo() + 2`. Like Azuresearch doesn't need `_select_relevance_score_fn` to return relevance scores (which is a `+2` part in your example).

elmargasimov:
Can someone give an update on the status of this issue? I get the `NotImplementedError` with SupabaseVectorStore when trying to use "similarity_score_threshold". I also can't use "mmr", but that is probably related to a separate issue.

libs/core/langchain_core/vectorstores.py
baskaryan: Right, so we either have to:
1. update all child classes that don't implement _select_relevance_score_fn to use run_in_executor approach / otherwise implement asimilarity_search_with...
2. update base class to use run_in_executor (this pr) and update all child classes that do implement _select_relevance_score_fn to use the _asimilarity_search_with... that's currently in base class (otherwise the base class change would make it so those classes that do implement _select_relevance... are losing the actual use of async)

either way we'll have to make updates to child classes, and my sense is option 1. would touch fewer classes than option 2. but we can check that explicitly if we want by counting how many classes don't override _select_relevance_score_fn

hwchase17:
we are revisiting vectorstore abstractions next month. closing this for now, we will address during that push

Issue #16913, opened 2024-02-02
Callback for on_llm_new_token should be invoked before the token is yielded by the model


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

## Goal 

Improve streaming in LangChain for chat models / language models.

## Background

Many chat and language models implement a streaming mode in which they stream tokens one at a time.

LangChain has a callback system that is useful for logging and important APIs like "stream", "stream_log" and "stream_events".

Currently many models incorrectly yield the token (chat generation) before invoking the callback.

## Acceptance criteria

For a PR to be accepted and merged, the PR should:

- [ ] Fix the code to make sure that the callback is called before the token is yielded
- [ ] Link to this issue
- [ ] Change ONE and only ONE model
- [ ] FIx sync and async implementation if both are defined

## Example PR

Here is an example PR that shows the fix for the OpenAI chat model:

https://github.com/langchain-ai/langchain/pull/16909

## Find models that need to be fixed

The easiest way to find places in the code that may need to be fixed is using git grep

```bash
git grep -C 5  "\.on_llm_new"
```

Examine the output to determine whether the callback is called before the token is yielded (correct) or after (needs to be fixed).

samjove:
Working on this, can I get it assigned?

efriis:
@samjove feel free to start working on it for a specific model and reference this issue!

yashankxy:
Hi, 
I am Yashank, 4th year student at UofT CS. I am working with @Ismail-Bashir @Ser0n-ath and @ji24077 who are also 4th year student at UofT CS. Can we take initiative to work on this issue and contribute to langchain?
Thanks 😊

PR #17348, opened 2024-02-09
community ollama[patch]: Invoke callback prior to yielding token

h0rv:
**Description:** Invoke callback prior to yielding token in stream method for Ollama.
**Issue:** [Callback for on_llm_new_token should be invoked before the token is yielded by the model #16913](https://github.com/langchain-ai/langchain/issues/16913)


h0rv:
Potential bugs found: `yield chunk` is not called in either of these functions:

https://github.com/langchain-ai/langchain/blob/7c1fd64292b89355a2e7b8a7cb3bb216e0c1134c/libs/community/langchain_community/chat_models/ollama.py#L190
https://github.com/langchain-ai/langchain/blob/7c1fd64292b89355a2e7b8a7cb3bb216e0c1134c/libs/community/langchain_community/chat_models/ollama.py#L216 

These are the only models with these functions so not positive if this is the correct behavior or not.

baskaryan (approved):

Issue #17057, opened 2024-02-05
Using SurrealDB as a vector store with no metadatas throws a TypeError


MarcusVirg:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

```python
import asyncio

from langchain.text_splitter import RecursiveCharacterTextSplitter
from langchain_community.vectorstores.surrealdb import SurrealDBStore
from langchain_openai import OpenAIEmbeddings


async def main():
    text = "Here is some sample text with the name test"

    texts = RecursiveCharacterTextSplitter(chunk_size=3, chunk_overlap=0).split_text(
        text
    )

    store = await SurrealDBStore.afrom_texts(
        texts=texts,
        # Can replace with other embeddings
        # otherwise OPENAI_API_KEY required
        embedding=OpenAIEmbeddings(),
        dburl="ws://localhost:8000/rpc",
        db_user="root",
        db_pass="root",
    )

    retriever = store.as_retriever()
    # Throws TypeError: 'NoneType' object is not a mapping
    docs = await retriever.aget_relevant_documents("What is the name of this text?")
    print(docs)


asyncio.run(main())
```

### Error Message and Stack Trace (if applicable)

```sh
Traceback (most recent call last):
  File "/workspaces/core-service/core_service/recreate_issue.py", line 32, in <module>
    asyncio.run(main())
  File "/usr/local/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/asyncio/base_events.py", line 684, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/workspaces/core-service/core_service/recreate_issue.py", line 28, in main
    docs = await retriever.aget_relevant_documents("What is the name of this text?")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #17374, opened 2024-02-10
community surrealdb[fix]: handle for empty metadata and allow collection names with complex characters

MarcusVirg:
- **Description:** Handle for empty metadata and allow collection names with complex characters
  - **Issue:** #17057
  - **Dependencies:** `surrealdb`


libs/community/langchain_community/vectorstores/surrealdb.py
```diff
@@ -12,6 +12,16 @@
 from langchain_core.vectorstores import VectorStore
 
 
+try:
```
hwchase17, line 15: this should stay in init, otherwise i think it will block imports of any vectorstore (since this is imported in init)

MarcusVirg: Ah good callout, I will move the import back

baskaryan (approved):

PR #17404, opened 2024-02-12
partners/openai: fix depracation errors of pydantic's .dict() function (reopen #16629)

SavvasMohito:
I am re-opening #16629 as I closed it accidentally. The commit matches the code of the previous PR, but we could improve it further based on the discussion in #16629. Thanks


efriis (approved):
Appreciate you linking prior discussion!

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -465,7 +465,11 @@ def _create_message_dicts(
     def _create_chat_result(self, response: Union[dict, BaseModel]) -> ChatResult:
         generations = []
         if not isinstance(response, dict):
-            response = response.dict()
+            response = (
+                getattr(response, "model_dump")()
+                if hasattr(response, "model_dump")
+                else response.dict()
+            )
```
baskaryan, line 472: think we can just model_dump here

SavvasMohito: `response` is expected to have a type of `pydantic.v1.BaseModel`, which has no `model_dump()` function. So we either have to change the type safety or go with the approach i commited (same case [here](https://github.com/langchain-ai/langchain/blob/a6d901a91236b79f74f2ba66ba2444594865916e/libs/partners/openai/langchain_openai/chat_models/azure.py#L214-L218))

SavvasMohito: We could import `BaseModel` from `openai` which supports `model_dump()` no matter the pydantic version, as you suggested [in the previous PR](https://github.com/langchain-ai/langchain/pull/16629#issuecomment-1935081971). It will make the code cleaner

SavvasMohito:
@baskaryan I think we need openai >=1.11.0. From what I can see, `BaseModel` cannot be imported from `openai`  in versions before 1.11.0. References below:

[openai v1.10.0 __init__.py](https://github.com/openai/openai-python/blob/0c1e58d511bd60c4dd47ea8a8c0820dc2d013d1d/src/openai/__init__.py#L8-L30) (BaseModel not included)
[openai v1.11.0 __init__.py](https://github.com/openai/openai-python/blob/66bab657c69dfebddbb4372adcdc97788cb3ceb1/src/openai/__init__.py#L12) (BaseModel included)

Issue #17242, opened 2024-02-08
Enhancement: Add ability to pass local cache to chat models


eyurtsev:
Feature request discussed  in https://github.com/langchain-ai/langchain/discussions/17176

Expand `cache` to accept a cache implementation in addition to a bool value:

https://github.com/langchain-ai/langchain/blob/00a09e1b7117f3bde14a44748510fcccc95f9de5/libs/core/langchain_core/language_models/chat_models.py#L106-L106

If provided, will use the given cache.

# Acceptance Criteria

- [ ] Documentation to the cache variable to explain how it can be used
- [ ] Update https://python.langchain.com/docs/modules/model_io/chat/chat_model_caching
- [ ] Include unit tests to test given functionality

PR can include implementation for caching of LLMs in addition to chat models.

alhridoy:
Hi @eyurtsev Could you check my PR? Thanks!

PR #17415, opened 2024-02-12
community[minor]: Add async methods to AstraDBCache

cbornet:

### DIFF
--- a/libs/community/langchain_community/cache.py
+++ b/libs/community/langchain_community/cache.py
@@ -29,12 +29,14 @@ import uuid
 import warnings
 from abc import ABC
 from datetime import timedelta
-from functools import lru_cache
+from functools import lru_cache, wraps
 from typing import (
     TYPE_CHECKING,
     Any,
+    Awaitable,
     Callable,
     Dict,
+    Generator,
     List,
     Optional,
     Tuple,
@@ -55,20 +57,20 @@ except ImportError:
 
 from langchain_core.caches import RETURN_VAL_TYPE, BaseCache
 from langchain_core.embeddings import Embeddings
-from langchain_core.language_models.llms import LLM, get_prompts
+from langchain_core.language_models.llms import LLM, aget_prompts, get_prompts
 from langchain_core.load.dump import dumps
 from langchain_core.load.load import loads
 from langchain_core.outputs import ChatGeneration, Generation
 from langchain_core.utils import get_from_env
 
-from langchain_community.utilities.astradb import AstraDBEnvironment
+from langchain_community.utilities.astradb import AstraDBCollectionEnvironment
 from langchain_community.vectorstores.redis import Redis as RedisVectorstore
 
 logger = logging.getLogger(__file__)
 
 if TYPE_CHECKING:
     import momento
-    from astrapy.db import AstraDB
+    from astrapy.db import AstraDB, AsyncAstraDB
     from cassandra.cluster import Session as CassandraSession
 
 
@@ -1360,6 +1362,10 @@ class AstraDBCache(BaseCache):
           (needed to prevent same-prompt-different-model collisions)
     """
 
+    @staticmethod
+    def _make_id(prompt: str, llm_string: str) -> str:
+        return f"{_hash(prompt)}#{_hash(llm_string)}"
+
     def __init__(
         self,
         *,
@@ -1367,7 +1373,10 @@ class AstraDBCache(BaseCache):
         token: Optional[str] = None,
         api_endpoint: Optional[str] = None,
         astra_db_client: Optional[AstraDB] = None,
+        async_astra_db_client: Optional[AsyncAstraDB] = None,
         namespace: Optional[str] = None,
+        pre_delete_collection: bool = False,
+        async_setup: bool = False,
     ):
         """
         Create an AstraDB cache using a collection for storage.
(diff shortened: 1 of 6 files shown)


libs/langchain/tests/integration_tests/cache/test_astradb.py
```diff
@@ -41,7 +41,20 @@ def astradb_cache() -> Iterator[AstraDBCache]:
...
+@pytest.fixture
+async def async_astradb_cache() -> AsyncIterator[AstraDBCache]:
+    cache = AstraDBCache(
+        collection_name="lc_integration_test_cache",
+        token=os.environ["ASTRA_DB_APPLICATION_TOKEN"],
+        api_endpoint=os.environ["ASTRA_DB_API_ENDPOINT"],
+        namespace=os.environ.get("ASTRA_DB_KEYSPACE"),
+        async_setup=True,
+    )
+    yield cache
```
eyurtsev, line 56: Do we need a try except here to make sure that clean up happens even on exception?

libs/community/langchain_community/utilities/astradb.py
```diff
@@ -46,39 +48,115 @@ def __init__(
...
+                namespace=self.async_astra_db.namespace,
+            )
+        else:
+            raise ValueError(
+                "Must provide 'astra_db_client' or 'async_astra_db_client' or "
+                "'token' and 'api_endpoint'"
+            )
+
+
+class AstraDBCollectionEnvironment(AstraDBEnvironment):
```
eyurtsev, line 89: * Would you mind adding doc-strings explaining what this is?
* Could we prefix this with a `_` to mark is as an internal object? I don't think that we expect most LangChain users to know about this object?

libs/community/langchain_community/utilities/astradb.py
```diff
@@ -46,39 +48,115 @@ def __init__(
...
+        self.async_collection = AsyncAstraDBCollection(
+            collection_name=collection_name,
+            astra_db=self.async_astra_db,
+        )
+
+        self.async_setup_db_task: Optional[Task] = None
+        if async_setup:
+            async_astra_db = self.async_astra_db
+
+            async def _setup_db() -> None:
```
eyurtsev, line 123: The code block seems to force a network request during initialization.

I left a comment about this above. We don't necessarily need to do anything at this stage, but could be good to allow a developer to opt-out from the init somehow for more advanced use cases

libs/community/langchain_community/cache.py
```diff
@@ -1377,29 +1386,35 @@ def __init__(
...
+                you can pass an already-created 'astrapy.db.AsyncAstraDB' instance.
             namespace (Optional[str]): namespace (aka keyspace) where the
                 collection is created. Defaults to the database's "default namespace".
+            pre_delete_collection (bool): whether to delete and re-create the
+                collection. Defaults to False.
+            async_setup (bool): whether to create the collection asynchronously.
+                Enable only if there is a running asyncio event loop. Defaults to False.
         """
-        astra_env = AstraDBEnvironment(
+        self.astra_env = AstraDBCollectionEnvironment(
```
eyurtsev, line 1402: FYI:  https://github.com/langchain-ai/langchain/issues/17242 -- will make it possible to set caches at request time, so we'll be more likely to see repeated instantiation of objects.

AstraDBCollectionEnvironment makes a network request on every init. OK for now, but could become expensive if user re-instantiates the cache repeatedly (which I think they should be allowed to do)

libs/community/langchain_community/cache.py
```diff
@@ -1443,21 +1481,78 @@ def delete_through_llm(
...
 
 ASTRA_DB_SEMANTIC_CACHE_DEFAULT_THRESHOLD = 0.85
 ASTRA_DB_CACHE_DEFAULT_COLLECTION_NAME = "langchain_astradb_semantic_cache"
 ASTRA_DB_SEMANTIC_CACHE_EMBEDDING_CACHE_SIZE = 16
 
 
+_unset = ["unset"]
+
+
+class CachedAwaitable:
```
eyurtsev, line 1530: If these are internal could we prefix them with `_` and add a quick doc-string explaining how they are used?

eyurtsev (commented):
Nice! Thank you for the PR. A few nit comments

libs/langchain/tests/integration_tests/cache/test_astradb.py
cbornet: Not needed. The teardown of fixtures is executed even if the test using them raises an exception.

libs/community/langchain_community/utilities/astradb.py
hemidactylus: FYI, the Cassandra vector store is about to receive a `skip_provisioning: bool = False` flag for exactly that purpose. There might be a same-named parameter here.

cbornet: added SetupMode param (SYNC, ASYNC, OFF)

libs/community/langchain_community/cache.py
cbornet: added SetupMode param (SYNC, ASYNC, OFF)

libs/community/langchain_community/cache.py
cbornet: done

libs/community/langchain_community/utilities/astradb.py
cbornet: done

libs/community/langchain_community/cache.py
eyurtsev: Thank you!

libs/community/langchain_community/utilities/astradb.py
eyurtsev: ah it was public because it's used in other files (obvious, but I missed that lol). I don't think we have a great convention right now with respect to public/private. I still prefer erring on the side of private and prioritize what users see as public

libs/community/langchain_community/utilities/astradb.py
```diff
@@ -46,39 +55,115 @@ def __init__(
...
+    def __init__(
+        self,
+        collection_name: str,
+        token: Optional[str] = None,
+        api_endpoint: Optional[str] = None,
+        astra_db_client: Optional[AstraDB] = None,
+        async_astra_db_client: Optional[AsyncAstraDB] = None,
+        namespace: Optional[str] = None,
+        setup_mode: SetupMode = SetupMode.SYNC,
+        pre_delete_collection: bool = False,
```
eyurtsev, line 106: do only i find `pre_delete_collection` scary? :)

eyurtsev (approved):

PR #17416, opened 2024-02-12
community[minor]: Fix pydantic ForwardRef error in BedrockBase

nat-n:
- **Description:** Fixes a type annotation issue in the definition of BedrockBase. This issue was that the annotation for the `config` attribute includes a ForwardRef to `botocore.client.Config` which is only imported when `TYPE_CHECKING`. This can cause pydantic to raise an error like `pydantic.errors.ConfigError: field "config" not yet prepared so type is still a ForwardRef, ...`. 
  - **Issue:** N/A
  - **Dependencies:** N/A
  - **Twitter handle:** `@__nat_n__`


baskaryan (commented):
maybe better to just remove the Config typing in L231 from pydantic class, doesn't add much

nat-n:
Thanks for taking a look @baskaryan 

> maybe better to just remove the Config typing in L231 from pydantic class, doesn't add much

So you're suggest to instead revert this previous commit? https://github.com/langchain-ai/langchain/commit/3fc1b3553b629803a1ce4987650cb2f324950546

Is your motivation here to avoid having to set `arbitrary_types_allowed = True`?

Removing this feature would be a breaking change, though I agree the complexity/value tradeoff of keeping it is debatable, since it's just a minor convenience over passing a boto3 client object to constructor.

I'm going to hedge my bets and create an another PR with your suggested approach, if one gets merged I'll close the other.

baskaryan (approved):

baskaryan:
thank you @nat-n!

PR #17425, opened 2024-02-12
Add async methods to InMemoryCache

cbornet:


eyurtsev (approved):
Thanks!

eyurtsev (approved):
Would be good to add unit tests for these as well if they're not covered by unit tests

cbornet:
> Would be good to add unit tests for these as well if they're not covered by unit tests

It's already covered (indirectly) in test_cache.py

PR #17446, opened 2024-02-13
community[patch]: Fix embedding generation for Cohere in AWS Bedrock

### LINKED ISSUE #17222: Documents and queries with bedrock's cohere model should result in different embedding values
### Checked other resources

### Example Code

Use following code:
```python
from langchain_community.embeddings import BedrockEmbeddings

embeddings = BedrockEmbeddings(
    region_name="us-east-1",
    model_id="cohere.embed-english-v3"
)

e1 = embeddings.embed_documents(["What is the project name?"])[0]
e2 = embeddings.embed_query("What is the project name?")

print(e1==e2)
```

Outputs: `True`

Should ideally be `False`.

### Description

Cohere models has the ability to generate optimized embeddings for type of use case. However it is not being leveraged in AWS Bedrock.

https://github.com/langchain-ai/langchain/blob/00a09e1b7117f3bde14a44748510fcccc95f9de5/libs/community/langchain_community/embeddings/bedrock.py#L123-L131

Current workaround is to define the `input_type` at the construtor which is not ideal (viz. using with predefined tools) compared to native way defined for the `CohereEmbeddings` class.

https://github.com/langchain-ai/langchain/blob/00a09e1b7117f3bde14a44748510fcccc95f9de5/libs/community/langchain_community/embeddings/cohere.py#L125-L134

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.22631
> Python Version:  3.11.6 (tags/v3.11.6:8b6ee5b, Oct  2 2023, 14:57:12) [MSC v.1935 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 0.1.21
> langchain: 0.1.5
> langchain_community: 0.0.19
> langsmith: 0.0.87

Packages not installed (Not Necessarily a Problem)
--------------------------------------------------
The following packages were not found:

> langgraph
> langserve

sansmoraxz:
Use cohere provided `input_type` for embedding generation for docs and queries. Also added parallelization as cohere supports generating embeddings for multiple texts with single API call.

Fixes: #17222

### DIFF
--- a/libs/community/langchain_community/embeddings/bedrock.py
+++ b/libs/community/langchain_community/embeddings/bedrock.py
@@ -45,8 +45,11 @@ class EmbeddingInputOutputAdapter:
         response_body = json.loads(response.get("body").read())
         if provider == "cohere":
             return response_body.get("embeddings")
-        if provider == "amazon":
+        elif provider == "amazon":
             return [response_body.get("embedding")]
+        else:
+            msg = f"Provider {provider} not supported"
+            raise NotImplementedError(msg)
 
     @classmethod
     def prepare_output(cls, provider: str, response: Any) -> List[float]:
--- a/libs/community/langchain_community/embeddings/bedrock.py
+++ b/libs/community/langchain_community/embeddings/bedrock.py
@@ -19,7 +19,10 @@ class EmbeddingInputOutputAdapter:
 
     @classmethod
     def prepare_inputs(
-        cls, provider: str, texts: List[str], model_kwargs: Dict[str, Any],
+        cls,
+        provider: str,
+        texts: List[str],
+        model_kwargs: Dict[str, Any],
         input_type: Optional[str] = None,
     ) -> List[Dict[str, Any]]:
         input_body = {**model_kwargs}
@@ -32,18 +35,13 @@ class EmbeddingInputOutputAdapter:
             return [input_body]
         elif provider == "amazon":
             # amazon does not support native batching
-            return [
-                {**input_body, "inputText": text}
-                for text in texts
-            ]
+            return [{**input_body, "inputText": text} for text in texts]
         else:
             msg = f"Provider {provider} not supported"
             raise NotImplementedError(msg)
 
     @classmethod
-    def prepare_outputs(
-        cls, provider: str, response: Any
-    ) -> List[List[float]]:
+    def prepare_outputs(cls, provider: str, response: Any) -> List[List[float]]:
         response_body = json.loads(response.get("body").read())
         if provider == "cohere":
             return response_body.get("embeddings")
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/embeddings/bedrock.py
```diff
@@ -112,40 +156,37 @@ def validate_environment(cls, values: Dict) -> Dict:
...
+                    accept="application/json",
+                    contentType="application/json",
+                )
+                for input_body in input_bodies
+            ]
+
+            response_bodies = [
+                EmbeddingInputOutputAdapter.prepare_outputs(provider, response)
+                for response in responses
+            ]
```
sansmoraxz, line 188: This can be improved through parallelization. As it stands Titan models would observe slowdown. I am open to suggestions.

baskaryan (commented):
cc @3coins @billytrend-cohere

libs/community/langchain_community/embeddings/bedrock.py
shabie: Why would Titan models observe slowdown?

sansmoraxz: By slowdown I mean sequential invocation can take quite long (as there is no batching).

Previous code had an asyncio gather to speed up (which was only available if invoked with async); it was not available for normal embed. So wondering if I should bring that back or there is some better alternative.

sansmoraxz: Another thing that can potentially bite are the rate limits. Not sure what the exact values are for non provisioned instances, but once you start hitting might take hours to go away.

sansmoraxz:
Not much reviewers for AWS?

libs/community/langchain_community/embeddings/bedrock.py
shabie: Still don't get it. async calls, threads or sequential, they all do the same things from the API's point of view and hence are equivalent in terms of throttling except that with concurrent requests (async / threads) you'll consume your "quota" faster. Also I think the throttling should happen on a per minute level and should clear up after 60 seconds. Unless it works differently for embeddings than it does for LLM calls.

sansmoraxz: Resolving this conversation. We have not observed any throttles in our use cases for embeddings, and if there is any issue happening in the wild a separate issue regarding such may be opened.

yonekura-dc:
Any updates on this? Also using bedrock on a project, but splitting a big document and calling the embeddings sequentially takes a long time. Tried to use async requests but apparently Boto3 is not thread-safe for various scenarios and doesn't actually support 
async.

sansmoraxz:
I am not entirely sure why this got blocked (probably no one reviewed yet).

You may try [llamaindex ](https://github.com/sansmoraxz/llama_index/tree/main/llama-index-integrations/embeddings/llama-index-embeddings-bedrock) if you want performance.

efriis:
closing because this is against the deprecated community aws integration - if it's still an issue feel free to submit to their team in langchain-ai/langchain-aws as an issue or PR against the partner package!

PR #17466, opened 2024-02-13
Doc: Update `llamacpp.ipynb`

169:
I found that some context was missing in the `Llama.cpp` document, such as how to download model files such as `openorca-platypus2-13b.gguf.q4_0.bin` and how to convert to `gguf` format model. Also, `model_path` is the user's directory. I changed it to relative to the model file path.

### DIFF
--- a/docs/docs/integrations/llms/llamacpp.ipynb
+++ b/docs/docs/integrations/llms/llamacpp.ipynb
@@ -10,17 +10,7 @@
     "\n",
     "It supports inference for [many LLMs](https://github.com/ggerganov/llama.cpp#description) models, which can be accessed on [Hugging Face](https://huggingface.co/TheBloke).\n",
     "\n",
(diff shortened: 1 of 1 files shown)


rlancemartin (commented):
Thanks for adding!

> Make sure you are following all instructions to [install all necessary model files](https://github.com/ggerganov/llama.cpp).

1/ We should clarify that you need to clone `llama.cpp` and also build. So, overall workflow order IIRC (it has been a while for me) is a) clone and make `llama.cpp`, install `llama-cpp-python`, download model and add to llama.cpp directory. Maybe we just add this as an overview at the top?

> Download Model

2/ We should clarify that the model needs to live in the `llama.cpp` models directory. This is implicit in the `model_path`, but it may be nice to say it.

> Note: new versions of llama-cpp-python use GGUF model files (see [here](https://github.com/abetlen/llama-cpp-python/pull/633)).

3/ Is this still needed? Maybe it's been long enough that we don't need to specify it.

PR #17467, opened 2024-02-13
partners: add Elasticsearch package

maxjakob:
### Description
This PR moves the Elasticsearch classes to a partners package.

Note that we will not move (and later remove) `ElasticKnnSearch`. It were previously deprecated.
`ElasticVectorSearch` is going to stay in the community package since it is used quite a lot still. 

Also note that I left the `ElasticsearchTranslator` for self query untouched because it resides in main `langchain` package.

### Dependencies
There will be another PR that updates the notebooks (potentially pulling them into the partners package) and templates and removes the classes from the community package, see https://github.com/langchain-ai/langchain/pull/17468

#### Open question
How to make the transition smooth for users? Do we move the import aliases and require people to install `langchain-elasticsearch`? Or do we remove the import aliases from the `langchain` package all together? What has worked well for other partner packages?


maxjakob:
hold on, the tests are failing, I need to do another pass over this

maxjakob:
I had copied the wrong FakeEmbeddings.

This is ready for review now 🟢

efriis (approved):

PR #17472, opened 2024-02-13
openai[patch]: relax tiktoken constraint, release 0.0.6

efriis:


harunzafer:
When will this patch be released?

PR #17475, opened 2024-02-13
[Langchain_community]: Fixed the Pandas Zero React Agent Creation

keenborder786:
-  The default `create_react_agent` being used in `create_pandas_dataframe_agent` raises the following exception:

```console
Prompt missing required variables: tools,tool_names
```

Upon closer inspection, it was clear that `_get_prompt` creates a prompt in which tools and tool_names are created as partial variables therefore the exception being raised. Therefore to fix this without breaking anything, I just created the same function i.e create_react_agent without the exception in the `agents.agent_toolkits.pandas.base` module and it works. Let me know if this is good or is their is a neater way to solve this issue.


baskaryan:
believe this was already fixed in #17442

keenborder786:
@baskaryan opsi. Closing it.

PR #17478, opened 2024-02-13
[Langchain_Community]: OpenLLM Client Fixes + Added Timeout Parameter

keenborder786:
- OpenLLM was using outdated method to get the final text output from openllm client invocation which was raising the error.  Therefore corrected that.
-  OpenLLM `_identifying_params` was getting the openllm's client configuration using outdated attributes which was raising error.
- Updated the docstring for OpenLLM.
- Added timeout parameter to be passed to underlying openllm client.


keenborder786:
@baskaryan

keenborder786:
@eyurtsev

baskaryan:
cc @aarnphm

aarnphm (approved):

PR #17492, opened 2024-02-13
partners: version constraints

efriis:
Core should be ^0.1 by default

Careful about 0.x.y and 0.0.z packages


efriis:
CI failing on lint in exa - will fix in followup

PR #17493, opened 2024-02-13
infra: add api docs build GHA

baskaryan:


.github/workflows/api_doc_build.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Set up Python ${{ env.PYTHON_VERSION }} + Poetry ${{ env.POETRY_VERSION }}
+        uses: "./.github/actions/poetry_setup"
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          poetry-version: ${{ env.POETRY_VERSION }}
+          cache-key: api-docs
+
+      - name: Install dependencies
+        working-directory: ${{ inputs.working-directory }}
```
efriis, line 30: I think delete this line - by default will be monorepo root

.github/workflows/api_doc_build.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          poetry run python -m pip install --upgrade --no-cache-dir pip setuptools
+          poetry run python -m pip install --upgrade --no-cache-dir sphinx readthedocs-sphinx-ext
+          poetry run python -m pip install ./libs/partners/*
+          poetry run python -m pip install --exists-action=w --no-cache-dir -r docs/api_reference/requirements.txt
+          poetry run python docs/api_reference/create_api_rst.py
+          poetry run python -m sphinx -T -E -b html -d _build/doctrees -c docs/api_reference docs/api_reference api_reference_build/html -j auto
```
efriis, line 37: now that we control it all can we just run the `make` command / make this a make command that makes it locally

efriis (commented):
nice!

.github/workflows/api_doc_build.yml
```diff
@@ -0,0 +1,42 @@
+name: API docs build
+
+on:
+  push:
+    - branches:
+        - 'bagatur/api_docs_build_gha'
```
efriis, line 6: this might create an infinite loop with the action at the end

.github/workflows/api_doc_build.yml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Install dependencies
+        working-directory: ${{ inputs.working-directory }}
+        run: |
+          poetry run python -m pip install --upgrade --no-cache-dir pip setuptools
+          poetry run python -m pip install --upgrade --no-cache-dir sphinx readthedocs-sphinx-ext
+          poetry run python -m pip install ./libs/partners/*
+          poetry run python -m pip install --exists-action=w --no-cache-dir -r docs/api_reference/requirements.txt
+          poetry run python docs/api_reference/create_api_rst.py
+          poetry run python -m sphinx -T -E -b html -d _build/doctrees -c docs/api_reference docs/api_reference api_reference_build/html -j auto
```
efriis, line 37: do we need to update gitignore to get this path in this branch?

.github/workflows/api_doc_build.yml
baskaryan: its commiting to api_docs_build, this is api_docs_build_gha

.github/workflows/api_doc_build.yml
baskaryan: its not ignored

.github/workflows/api_doc_build.yml
baskaryan: make command needs to be updated to work here

Issue #17505, opened 2024-02-14
Expand docker-compose with common dependencies used in integration tests


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

The goal is to make it easier to run integration tests.

Acceptance Criteria:

- [ ] Update docker compose file here: https://github.com/langchain-ai/langchain/blob/master/docker/docker-compose.yml to include a database that is used during integration tests and can be spun up locally via docker compose.
- [ ] Use non standard port -- just increment by 1 from the previous service in that file
- [ ] Update any integration tests that use the given service to use a matching port

This is a good first issue for folks with experience in dev ops.

Consider git-grepping for existing yml files that contain service configuration.

rachana-uniyal:
I would like to work on this issue.

PR #17510, opened 2024-02-14
ci[minor] : Added graphdb in docker compose for integration tests

kartheekyakkala:
Checklist:

- [x] PR title: Expanding docker compose for integration tests
- [x] PR message: 
    - **Description:**  Added graphdb in docker compose for integration tests
    - **Issue:** https://github.com/langchain-ai/langchain/issues/17505


kartheekyakkala:
This is related to issue : Expand docker-compose with common dependencies used in integration tests #17505

docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ontotext/graphdb:10.5.1
```
eyurtsev, line 1: Thank you!

The docker-compose is supposed to support different services. 

Could we add a namespace docker/graphdb/ and place the Dockerfile within it as well as the grapdb_create.sh script?

docker/start.sh
```diff
@@ -0,0 +1,5 @@
+set -ex
```
eyurtsev, line 1: We want a single entry point for any service that might be needed for integration tests.

A Makefile will be OK with some parameteriszation for the service

Could be parameterized as follows:

`make build_graphdb`
`make start_graphdb`

eyurtsev:
Awesome thank you for the PR! Could we add a bit of namespacing? (See PR comments)

docker/start.sh
kartheekyakkala: Yes, that would be a good thing, I have made changes accordingly.

docker/Dockerfile
kartheekyakkala: Ok, I have added them.

kartheekyakkala:
I have made changes accordingly, by adding a makefile as a single entrypoint.

docker/makefile
```diff
@@ -0,0 +1,12 @@
+# Makefile
```
eyurtsev, line 1: There are some syntax errors in the file I think?

docker/graphdb/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ontotext/graphdb:10.5.1
+RUN mkdir -p /opt/graphdb/dist/data/repositories/langchain
+COPY config.ttl /opt/graphdb/dist/data/repositories/langchain/
+COPY starwars-data.trig /
```
eyurtsev, line 4: Build will fail right now because some of these data doesn't exist locally.

We could add a script to download and verify hash or else avoid loading in data for now. For the integration tests,  it's useful to have the service even without fixtures loaded

docker/makefile
kartheekyakkala: Yeah, my vscode made them used normal spaces for tabspaces, I have changed it in settings and is working fine now.

docker/graphdb/Dockerfile
kartheekyakkala: Yeah, correct. I have removed them.

kartheekyakkala:
I have made changes accordingly, and it has no issues.

kartheekyakkala:
@eyurtsev can you please check once.

docker/graphdb/graphdb_create.sh
```diff
@@ -0,0 +1,33 @@
...
+        break
+    fi
+    sleep 30s
+    echo "CHECK_RES: ${CHECK_RES}"
+  done
+}
+
+function loadData {
+  echo -e "\nImporting starwars-data.trig"
+  curl -X POST -H "Content-Type: application/x-trig" -T /starwars-data.trig ${GRAPHDB_URI}/repositories/${REPOSITORY_ID}/statements
```
eyurtsev, line 27: Will this work if it's not included in the dockerfile?

eyurtsev:
Thank you!

Merging -- I'll follow up with another change tomorrow to rename makefile to Makefile (i think it's standard upper case)

kartheekyakkala:
Thanks for the support @eyurtsev . This is my first real contribution to open source. 😄

PR #17537, opened 2024-02-14
openai[patch]: pass message name

efriis:


libs/partners/openai/pyproject.toml
```diff
@@ -12,7 +12,7 @@ license = "MIT"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
-langchain-core = "^0.1.16"
+langchain-core = "^0.1.25"
```
efriis, line 15: had to bump core for basemessage.name

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -135,19 +144,15 @@ def _convert_message_to_dict(message: BaseMessage) -> dict:
...
-        message_dict = {
-            "role": "tool",
-            "content": message.content,
-            "tool_call_id": message.tool_call_id,
-        }
+        message_dict["role"] = "tool"
+        message_dict["tool_call_id"] = message.tool_call_id
+
+        # tool message doesn't have name: https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages
+        del message_dict["name"]
```
jacoblee93, line 155: Should we log a warning here? Will OpenAI reject this?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -79,8 +79,9 @@ def _convert_dict_to_message(_dict: Mapping[str, Any]) -> BaseMessage:
         The LangChain message.
     """
     role = _dict.get("role")
+    name = _dict.get("name")
```
jacoblee93, line 82: This doesn't appear to be in the TS SDK types

libs/partners/openai/langchain_openai/chat_models/base.py
efriis: OpenAI rejects. Think I should raise an error?

efriis: changed to pass through unless it's none

PR #17539, opened 2024-02-14
core[minor]: add name to basemessage

efriis:
Adds an optional name param to our base message to support passing names into LLMs.

OpenAI supports having a name on anything except tool message now (system, ai, user/human).


libs/core/tests/unit_tests/runnables/__snapshots__/test_runnable.ambr
efriis, line 1: add "name" to snapshots

nfcampos (approved):

PR #17543, opened 2024-02-14
langchain: updates default json splitter chunk size and list conversion

joelsprunger:
- **Description:** This changes the default value for max_chunk_size from 2000 to the more appropriate 500 characters. This also changes the default for convert_lists to True as I believe most users will want to split long lists by default. I have updated the ipynb to reflect this updated default behavior.
- **Issue:** None
- **Dependencies:** None
- **Twitter handle:** @joelsprunger

### DIFF
--- a/docs/docs/modules/data_connection/document_transformers/recursive_json_splitter.ipynb
+++ b/docs/docs/modules/data_connection/document_transformers/recursive_json_splitter.ipynb
@@ -15,7 +15,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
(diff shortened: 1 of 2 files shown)


baskaryan:
probably ok given how new this feature is, but we should avoid changing defaults like this in the future. technically a breaking change

joelsprunger:
That's on me. I see this was in the 0.1.6 tag - does that mean it was released?

ccurme:
I think unfortunately at this point the feature is no longer new, and the damage from changing defaults out from under users might outweigh the benefits.

PR #17569, opened 2024-02-15
community: Use astrapy built-in pagination prefetch in AstraDBLoader

cbornet:


libs/community/langchain_community/document_loaders/astradb.py
```diff
@@ -33,6 +30,7 @@ class AstraDBLoader(BaseLoader):
     def __init__(
         self,
         collection_name: str,
+        *,
```
eyurtsev, line 33: Not related to this PR, but the loader could use a lot of in code documentation. It's surfaced in the APIReference and in interactive environments. Some of the variables are self explanatory, but other variables like filter_criteria or projection would require reading through the code or fishing for documentation somewhere else.

eyurtsev (approved):

libs/community/langchain_community/document_loaders/astradb.py
cbornet: Sure. I'll add pydoc as follow-up PR.

PR #17577, opened 2024-02-15
core: add serialize path for ChatMistralAI

jessicamann:
**Description:** Include a serialize path for ChatMistralAI so that the chat model can be loaded by `langchain_core.load`

**Issue:** n/a

**Dependencies:** none


efriis:
Closing because there's more involved in making these serializable, and curious what the use case is!

PR #17588, opened 2024-02-15
docs sorting `Examples using ...` section

leo-gan:
The API Reference docs. If the class has a long list of the examples that works with this class, then the `Examples using` list is [hard to comprehend](https://api.python.langchain.com/en/latest/llms/langchain_community.llms.openai.OpenAI.html#langchain-community-llms-openai-openai). If this list is sorted it would be much easier.
- sorting the `Examples using <ClassName>` list


baskaryan (approved):

leo-gan:
@baskaryan Thank you, Bagatur!

PR #17595, opened 2024-02-15
Docs: Add custom chat model documenation

eyurtsev:
This PR adds documentation about how to implement a custom chat model.


docs/docs/modules/model_io/chat/custom_chat_model.ipynb
baskaryan, line 1: should add sidebar_position at the top, probably want this towards the end?

baskaryan: actually maybe it's fine where it is

PR #17603, opened 2024-02-15
core[minor]: Add Runnable.batch_as_completed

nfcampos:
This PR adds `batch as completed` method to the standard Runnable interface. It takes in a list of inputs and yields the corresponding outputs as the inputs are completed.

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -530,6 +530,52 @@ class Runnable(Generic[Input, Output], ABC):
         with get_executor_for_config(configs[0]) as executor:
             return cast(List[Output], list(executor.map(invoke, inputs, configs)))
 
+    def batch_as_completed(
+        self,
+        inputs: List[Input],
+        config: Optional[Union[RunnableConfig, List[RunnableConfig]]] = None,
+        *,
+        return_exceptions: bool = False,
+        **kwargs: Optional[Any],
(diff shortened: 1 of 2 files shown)


eyurtsev (approved):

libs/core/langchain_core/runnables/base.py
```diff
@@ -567,6 +613,40 @@ async def ainvoke(
...
         return await gather_with_concurrency(configs[0].get("max_concurrency"), *coros)
 
+    async def abatch_as_completed(
+        self,
+        inputs: List[Input],
+        config: Optional[Union[RunnableConfig, List[RunnableConfig]]] = None,
+        *,
+        return_exceptions: bool = False,
+        **kwargs: Optional[Any],
+    ) -> AsyncIterator[Tuple[Input, Output]]:
```
eyurtsev, line 623: Do we need this to be `AsyncIterator[Tuple[int, Input, Output]]`?

Duplicated Input, Output but different configs

nfcampos: like index in input array?

nfcampos: in that case I'd actually prefer to not return the input at all, just index and output, wdyt?

eyurtsev: yeah that works

baskaryan (approved):

PR #17605, opened 2024-02-16
community: Add Anyscale embedding support

kylehh:
**Description:** Add embedding model support for Anyscale Endpoint
**Dependencies:** openai


kylehh:
@baskaryan Please take a look at the PR . Thanks!

libs/community/langchain_community/embeddings/anyscale.py
```diff
@@ -0,0 +1,75 @@
+"""Anyscale embeddings wrapper."""
+from __future__ import annotations
+
+from typing import Dict
+
+from langchain_core.pydantic_v1 import Field, SecretStr, root_validator
+from langchain_core.utils import convert_to_secret_str, get_from_dict_or_env
+
+from langchain_community.embeddings.openai import OpenAIEmbeddings
```
baskaryan, line 9: this class is deprecated. it may be better to just re-implement the functionality you need

kylehh: I tried to change it to `from langchain_openai import OpenAIEmbeddings`, but it raised tons of dependence issues. It seems that `langchain_openai` is not installed by default, so the community libs can NOT depends on it. Please guide how to solve it.

baskaryan (approved):

PR #17607, opened 2024-02-16
infra: check min versions in PR ci

efriis:


efriis:
nvm just doing on releases is fine

PR #17614, opened 2024-02-16
community: removed unnecessary dependecy use from NotebookLoader

lachiewalker:
**Description:** Removes unnecessary pandas use from the NotebookLoader document loader.
- In no way changes the output from this classes `.load()` method.
- Is compatible with existing tests.
- Delivers a noticeable speedup over the previous method.


libs/community/langchain_community/document_loaders/notebook.py
```diff
@@ -104,29 +108,29 @@ def load(
...
-            raise ImportError(
-                "pandas is needed for Notebook Loader, "
-                "please install with `pip install pandas`"
-            )
         p = Path(self.file_path)
 
         with open(p, encoding="utf8") as f:
             d = json.load(f)
 
-        data = pd.json_normalize(d["cells"])
```
baskaryan, line 119: what does pd.json_normalize do

lachiewalker: From the [documentation](https://pandas.pydata.org/docs/reference/api/pandas.json_normalize.html): Normalizes semi-structured JSON data into a flat table.

Combined with `filtered_data = data[["cell_type", "source", "outputs"]]` this statement makes a table of [ cell_type, source, output ] information.

This behaviour can be simply replicated by a list of dictionaries with keys `cell_type`, `source`, `output`, without relying on a dependency.

baskaryan: makes sense, could we add a simple unit test?

paulonasc: @lachiewalker 👋 new to contributing here, I was wondering if I can help write the unit test?

lachiewalker: @paulonasc yeah, of course 🙂

paulonasc: here's the commit for the unit tests btw! https://github.com/langchain-ai/langchain/pull/17721

Is it okay that it's in a separate PR?

PR #17615, opened 2024-02-16
docs: Fix bug that caused the word "Deprecated" to appear twice in doc-strings

leo-gan:
The current issue:
Most of the deprecation descriptions in the API Reference are duplicated. For example:
`[Deprecated]  Chat Agent.[Deprecated] Chat Agent.` for the [ChatAgent class](https://api.python.langchain.com/en/latest/langchain_api_reference.html#classes) description.

NOTE: I've tested it only with new ut! I cannot build API Reference locally :(

### DIFF
--- a/libs/core/langchain_core/_api/deprecation.py
+++ b/libs/core/langchain_core/_api/deprecation.py
@@ -246,9 +246,8 @@ def deprecated(
         old_doc = inspect.cleandoc(old_doc or "").strip("\n")
 
         if not old_doc:
-            new_doc = "[*Deprecated*]"
-        else:
-            new_doc = f"[*Deprecated*]  {old_doc}"
+            "old_doc can be None"
+            old_doc = ""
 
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/_api/deprecation.py
```diff
@@ -246,9 +246,8 @@ def finalize(  # type: ignore
         old_doc = inspect.cleandoc(old_doc or "").strip("\n")
 
         if not old_doc:
-            new_doc = "[*Deprecated*]"
-        else:
-            new_doc = f"[*Deprecated*]  {old_doc}"
+            "old_doc can be None"
```
eyurtsev, line 249:
```suggestion
            # old_doc can be None
```

eyurtsev (approved):
Thank you! Just minor change for the comment

eyurtsev (approved):

leo-gan:
@eyurtsev Thanks!

PR #17622, opened 2024-02-16
community: Change github endpoint in GithubLoader

RadhikaBansal97:
Description- 
    - Changed the GitHub endpoint as existing was not working and giving 404 not found error 
    -  Also the existing function was failing if file_filter is not passed as the tree api return all paths including directory as well, and when get_file_content was iterating over these path, the function was failing for directory as the api was returning list of files inside the directory, so added a condition to ignore the paths if it a directory
    -  Fixes this issue -  https://github.com/langchain-ai/langchain/issues/17453


nicolastakashi:
Amazing!
I found the same issue, any chance we got some review here @baskaryan and @shufanhao

shufanhao:
thanks for the fix.

shufanhao (approved):

PR #17638, opened 2024-02-16
community[patch]: add VoyageEmbeddings truncation

baskaryan:


libs/community/langchain_community/embeddings/voyageai.py
```diff
@@ -110,6 +119,8 @@ def _invocation_params(
             "json": {"model": self.model, "input": input, "input_type": input_type},
             "timeout": self.request_timeout,
         }
+        if self.truncation is not None:
+            params["truncation"] = self.truncation
```
thomas0809, line 123: It should be `params["json"]["truncation"] = self.truncation`, right?
Just curious, do you prefer to keep the signature of `embed_documents` and `embed_query` unchanged? In that sense, the other arguments, such as this `truncation`, should be set as a class attribute?

baskaryan: ah good point!

yea we shouldn't change embed_documents, embed_query since those are defined by the parent abstraction

thomas0809 (approved):

PR #17649, opened 2024-02-16
community: Allow override of 'fetch_schema_from_transport' in the GraphQL tool

ztoliver:
- **Description:** In order to override the bool value of "fetch_schema_from_transport" in the GraphQLAPIWrapper, a "fetch_schema_from_transport" value needed to be added to the "_EXTRA_OPTIONAL_TOOLS" dictionary in load_tools in the "graphql" key. The parameter "fetch_schema_from_transport" must also be passed in to the GraphQLAPIWrapper to allow reading of the value when creating the client. Passing as an optional parameter is probably best to avoid breaking changes. This change is necessary to support GraphQL instances that do not support fetching schema, such as TigerGraph. More info here: [TigerGraph GraphQL Schema Docs](https://docs.tigergraph.com/graphql/current/schema)
  - **Threads handle:** @zacharytoliver


ztoliver:
@baskaryan would you be able to help with linting again? For the life of me, I can't get it to work.

ztoliver:
@baskaryan thanks for the fmt! Seems there may still be linting issues after the merge. Hoping to get this in for v0.1.9, if possible, to be able to override that auto fetch schema setting.

baskaryan (approved):

PR #17655, opened 2024-02-16
community: Voyage AI updates default model and batch size

thomas0809:
- **Description:** update the default model and batch size in VoyageEmbeddings
    - **Issue:** N/A
    - **Dependencies:** N/A
    - **Twitter handle:** N/A


libs/community/langchain_community/embeddings/voyageai.py
```diff
@@ -121,7 +121,7 @@ def _get_embeddings(
         embeddings: List[List[float]] = []
 
         if batch_size is None:
-            batch_size = self.batch_size
```
baskaryan, line 124: we probably don't want to change this but instead put this logic in validate_environment to change self.batch_size?

```python
    @root_validator(pre=True)
        values["voyage_api_key"] = convert_to_secret_str(
            get_from_dict_or_env(values, "voyage_api_key", "VOYAGE_API_KEY")
        )
        if "batch_size" not in values:
            values["batch_size"] = 72 if values["model"] in ...
        return values
```

libs/community/langchain_community/embeddings/voyageai.py
```diff
@@ -69,15 +69,15 @@ class VoyageEmbeddings(BaseModel, Embeddings):
...
             from langchain_community.embeddings import VoyageEmbeddings
 
-            voyage = VoyageEmbeddings(voyage_api_key="your-api-key")
+            voyage = VoyageEmbeddings(voyage_api_key="your-api-key", model="voyage-2")
             text = "This is a test query."
             query_result = voyage.embed_query(text)
     """
 
-    model: str = "voyage-01"
+    model: str
```
baskaryan, line 77: this is a breaking change we probably don't want to make

thomas0809: Unfortunately, we may want to make a breaking change here. We want the user to specify the model when using the embeddings, as we want to encourage them to use the new models but we can't frequently update the default model here.
Do you have any suggestions?

baskaryan: i would recommend at least giving folks a month or two heads up before doing this by adding a warning for now that the breaking change will happen by a certain date. here's an example: https://github.com/langchain-ai/langchain/pull/14614

baskaryan: and for now update all of the documentation to show the latest models being passed in

libs/community/langchain_community/embeddings/voyageai.py
baskaryan: L96

libs/community/langchain_community/embeddings/voyageai.py
thomas0809: I agree that giving a warning could be a better idea.

libs/community/langchain_community/embeddings/voyageai.py
thomas0809: No, I meant when is it called? There is no place calling this function in this file. So I am assuming it is called somewhere in its base class?

thomas0809: nvm, it is a Pydantic thing. Yes, it looks like a good idea to do it here. Thanks!

thomas0809:
@baskaryan We've made changes according to your suggestion. Please review!

libs/community/langchain_community/embeddings/voyageai.py
thomas0809: I moved the default value to `validate_environment`. Thus it should be backward compatible now.

baskaryan (approved):

PR #17662, opened 2024-02-16
community: Remove model limitation on Anyscale LLM

kylehh:
**Description:** Llama Guard is deprecated from Anyscale public endpoint. 
**Issue:** Change the default model. and remove the limitation of only use Llama Guard with Anyscale LLMs
Anyscale LLM can also works with all other Chat model hosted on Anyscale. 
Also added `async_client` for Anyscale LLM

### DIFF
--- a/libs/community/langchain_community/llms/anyscale.py
+++ b/libs/community/langchain_community/llms/anyscale.py
@@ -26,6 +26,7 @@ from langchain_community.utils.openai import is_openai_v1
 DEFAULT_BASE_URL = "https://api.endpoints.anyscale.com/v1"
 DEFAULT_MODEL = "mistralai/Mixtral-8x7B-Instruct-v0.1"
 
+
 def update_token_usage(
     keys: Set[str], response: Dict[str, Any], token_usage: Dict[str, Any]
 ) -> None:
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/llms/anyscale.py
```diff
@@ -24,10 +24,7 @@
 from langchain_community.utils.openai import is_openai_v1
 
 DEFAULT_BASE_URL = "https://api.endpoints.anyscale.com/v1"
-DEFAULT_MODEL = "Meta-Llama/Llama-Guard-7b"
-
-# Completion models support by Anyscale Endpoints
-COMPLETION_MODELS = ["Meta-Llama/Llama-Guard-7b"]
+DEFAULT_MODEL = "mistralai/Mixtral-8x7B-Instruct-v0.1"
```
baskaryan, line 27: changing default model is a breaking change, would it be better to add a warning that llama guard is deprecated?

kylehh: Changing the model is not really affect the output, still will generate LLM responses for the prompt. 
 From Anyscale backend, we can see the previous Llama guide model has very low traffic, so the change won't be breaking to users ( That's also the reason we deprecate this model from public hosting)

baskaryan (approved):

PR #17668, opened 2024-02-17
docs : fix 404 issue with legacy /overview URL path in Server section

montylounge:
**Description:** fix 404 issue with Server section of docs. WebBaseLoader requesting a legacy URL that results in 404.


baskaryan:
believe this was fixed in #17981, regardless thank you for contributing @montylounge!

PR #10812, opened 2023-09-20
set litellm max_tokens=None

zhangnew:
- **Description:** The default value of the `max_tokens` for litellm is 256. It should be increased to avoid truncation of replies.

PR #17691, opened 2024-02-18
community: set litellm max_tokens=None

zhangnew:
Description: The default value of the max_tokens for litellm is 256. It should be increased to avoid truncation of replies.

resovle conflicts for #10812


baskaryan (commented):
do we know that litellm can handle null values? cc @krrishdholakia

ccurme:
Closing as this was implemented separately in https://github.com/langchain-ai/langchain/pull/21534. Apologies as this PR was first. Thank you for the fix!

PR #17704, opened 2024-02-18
docs: Fix bug that caused the word "Beta" to appear twice in doc-strings

leo-gan:
The current issue:
Several beta descriptions in the API Reference are duplicated. For example:
`[Beta]  Get a context value.[Beta] Get a context value.` for the [ContextGet class](https://api.python.langchain.com/en/latest/core_api_reference.html#module-langchain_core.beta) description.

NOTE: I've tested it only with a new ut! I cannot build API Reference locally :(
This PR related to #17615


eyurtsev (approved):

leo-gan:
@eyurtsev Thanks!

Issue #16422, opened 2024-01-23
use qwen to create agent TypeError: Additional kwargs key output_tokens already exists in left dict and value has unsupported type <class 'int'>.


YanTianlong-01:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.

### Example Code

```

from langchain_community.llms import Tongyi
import os
from langchain_community.document_loaders import WebBaseLoader
from langchain.embeddings import DashScopeEmbeddings
from langchain_community.vectorstores import FAISS
from langchain.text_splitter import RecursiveCharacterTextSplitter
from langchain.chains.combine_documents import create_stuff_documents_chain
from langchain_core.documents import Document
from langchain.chains import create_retrieval_chain
from langchain_community.tools.tavily_search import TavilySearchResults
from langchain.tools.retriever import create_retriever_tool


# 引入大模型
os.environ["DASHSCOPE_API_KEY"] = "my_api_key"
llm = Tongyi(model_name="qwen-turbo")



loader = WebBaseLoader("https://docs.smith.langchain.com/overview")

docs = loader.load()

# 映射到向量空间
embeddings = DashScopeEmbeddings(
    model="text-embedding-v1", dashscope_api_key="my_api_key"
)

# 分词
text_splitter = RecursiveCharacterTextSplitter()
documents = text_splitter.split_documents(docs)
vector = FAISS.from_documents(documents, embeddings)
retriever = vector.as_retriever()


retriever_tool = create_retriever_tool(
    retriever,
    "langsmith_search",
    "Search for information about LangSmith. For any questions about LangSmith, you must use this tool!",
)
# with api
os.environ["TAVILY_API_KEY"] = "my_api_key"
search = TavilySearchResults()


tools = [retriever_tool, search]


from langchain import hub

from langchain.agents import AgentExecutor, create_xml_agent
from langchain.agents import create_react_agent, create_json_chat_agent

prompt = hub.pull("hwchase17/react")
# 初始化 agent
agent =  create_react_agent(llm, tools, prompt)
agent_executor = AgentExecutor(agent=agent, tools=tools, verbose=True)


agent_executor.invoke({"input": "what is LangChain?"})
```
inputs: 
 {'input': 'what is LangChain?'}

> Entering new AgentExecutor chain...
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[...]

PR #17712, opened 2024-02-19
fix: the merging error when using non-OpenAI models

Bob-AFei:

Checklist:

- [x] PR title: Please title your PR "package: description", where "package" is whichever of langchain, community, core, experimental, etc. is being modified. Use "docs: ..." for purely docs changes, "templates: ..." for template changes, "infra: ..." for CI changes.
- [x] PR message: **Delete this entire template message** and replace it with the following bulleted list
- [x] Pass lint and test: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified to check that you're passing lint and testing. See contribution guidelines for more information on how to write/run tests, lint, etc: https://python.langchain.com/docs/contributing/

Additional guidelines:


Bob-AFei:
relative issue: #16422

Bob-AFei:
@eyurtsev Hi there, would you please review this PR? 
Thanx.

libs/core/langchain_core/utils/_merge.py
```diff
@@ -34,6 +34,9 @@ def merge_dicts(left: Dict[str, Any], right: Dict[str, Any]) -> Dict[str, Any]:
             merged[k] = merge_dicts(merged[k], v)
         elif isinstance(merged[k], list):
             merged[k] = merged[k] + v
+        # when using Non-OpenAI chatmodels, the output and input tokens may be different between left and right
+        elif isinstance(merged[k], int):
+            merged[k] = max(merged[k], v)
```
pllopis, line 39: I have the same issue with a non-OpenAI model, except in my case it's a float instead of int.
(I am unsure whether the max operation will make sense, but if I understand correctly that this is merging streamed chat chunk responses, then it might be correct?)
Anyway if this fix makes sense, please consider also adding float for this condition.

baskaryan:
merging ints isn't supported because there's not a generic way to do it (some ints should be added, some should be overwritten, some should be maxxed, etc). the correct fix is to update what a model is streaming so that there's not need to merge ints. most likely what you should do is turn the int into a list (of ints) so that merging to message chunks keeps track of both original ints

pllopis:
A lot of non-OpenAI models return float/int. Some of Google's own models do this, so there's no way for us end users to fix the model itself. Does this mean that langchain can't/won't support Google AI models?

PR #17723, opened 2024-02-19
core: Move document loader interfaces to core

cbornet:
This is needed to be able to move document loaders to partner packages.


cbornet:
There's a circular dependency since  BaseLoader uses TextSplitter which is in the langchain package. How should it be handled ?

cbornet:
I opted for moving  TextSplitter and RecursiveCharacterTextSplitter to core.
Is that OK ?

cbornet:
[The text splitters ](https://github.com/langchain-ai/langchain/pull/18346) moved the text splitters to their own package.
I'll update the PR accordingly.

cbornet:
Rebased to use langchain-text-splitters

hemidactylus:
LGTM, thank you for identifying the problem and coming up with an elegant solution!

eyurtsev:
Fantastic I've been wanting to do this for a while!

eyurtsev:
OK document loaders is good according to @baskaryan

eyurtsev (approved):

PR #17732, opened 2024-02-19
partners/astradb: Add AstraDBChatMessageHistory to langchain-astradb package

cbornet:


libs/community/langchain_community/chat_message_histories/astradb.py
```diff
@@ -23,6 +24,11 @@
 DEFAULT_COLLECTION_NAME = "langchain_message_store"
 
 
+@deprecated(
+    since="0.1.24",
```
baskaryan, line 28: community is on 0.0.22: `since="0.0.22"`

cbornet: Sure. I initially used the langchain-core version because that's the one that was used for the Pinecone VectorStore when moved to a partner package: https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/pinecone.py#L46 .

baskaryan (approved):

Issue #17241, opened 2024-02-08
In the astream_log() method, you cannot use the bind method with RunnableLambda.


eyurtsev:
### Discussed in https://github.com/langchain-ai/langchain/discussions/16446

<div type='discussions-op-text'>

<sup>Originally posted by **jason1315** January 23, 2024</sup>
In my project, I need to implement the following logic. Here is a simple:
```python
import asyncio

from langchain_core.runnables import *

from lang_chain.llm.llms import llm

def _test(_dict):
print("value:", _dict)
return _dict

@chain
def my_method(_dict, **keywords):
print(keywords)
return RunnablePassthrough.assign(key=lambda x: keywords.get("i")) | RunnableLambda(_test)

if name == 'main':

loop = asyncio.new_event_loop()

my_list = ["1", "2", "3", " 4", "5"]

head = RunnablePassthrough()
for i in my_list:
    head = head | my_method.bind(i=i)
stream = head.invoke({})
#
# async def __stream(stream1):
#     async for i in stream1:
#         print(i)
#
# loop.run_until_complete(__stream(stream))
```
When I use the .invoke({}) method, it outputs the following results correctly:
```text
{'i': '1'}
value: {'key': '1'}
{'i': '2'}
value: {'key': '2'}
{'i': '3'}
value: {'key': '3'}
{'i': ' 4'}
value: {'key': ' 4'}
{'i': '5'}
value: {'key': '5'}
```
But if I use the astream_log({}) method, it throws an error:
```text
File "F:\py3.11\Lib\concurrent\futures\thread.py", line 58, in run
result = self.fn(*self.args, **self.kwargs)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: RunnableLambda._atransform.<locals>.func() got an unexpected keyword argument 'i'
```
Why is it designed like this? Do I need to implement a runnable similar to the model if I want to achieve the above logic?</div>

UtkarshaGupte:
@baskaryan - Is this issue fixed by [PR](https://github.com/langchain-ai/langchain/pull/17739) ?

PR #17739, opened 2024-02-19
core: fix .bind when used with RunnableLambda async methods

ccurme:
**Description:** Here is a minimal example to illustrate behavior:
```python

def my_function(*args, **kwargs):
    return 3 + kwargs.get("n", 0)

runnable = RunnableLambda(my_function).bind(n=1)


assert 4 == runnable.invoke({})
assert [4] == list(runnable.stream({}))

assert 4 == await runnable.ainvoke({})
assert [4] == [item async for item in runnable.astream({})]
```
Here, `runnable.invoke({})` and `runnable.stream({})` work fine, but `runnable.ainvoke({})` raises
```
TypeError: RunnableLambda._ainvoke.<locals>.func() got an unexpected keyword argument 'n'
```
and similarly for `runnable.astream({})`:
```
TypeError: RunnableLambda._atransform.<locals>.func() got an unexpected keyword argument 'n'
```
Here we assume that this behavior is undesired and attempt to fix it.

**Issue:** https://github.com/langchain-ai/langchain/issues/17241, https://github.com/langchain-ai/langchain/discussions/16446

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -3416,6 +3416,7 @@ class RunnableLambda(Runnable[Input, Output]):
                     input: Input,
                     run_manager: AsyncCallbackManagerForChainRun,
                     config: RunnableConfig,
+                    **kwargs: Any,
                 ) -> Output:
                     output: Optional[Output] = None
                     for chunk in call_func_with_variable_args(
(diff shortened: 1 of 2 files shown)


ccurme:
Would appreciate a thorough review of this one as although tests pass, I don't have a good intuition as to whether it introduces new bugs.

baskaryan (approved):
looks right to me, great catch @ccurme!

PR #17748, opened 2024-02-19
Mongo index creation

nvenkat94:
- [ ] Title: Mongodb: MongoDB connection performance improvement. 
- [ ] Message: 
    - **Description:** I made collection index_creation as optional. Index Creation is one time process.
    - **Issue:** MongoDBChatMessageHistory class object is attempting to create an index during connection, causing each request to take longer than usual. This should be optional with a parameter.
    - **Dependencies:** N/A
    - **Branch to be checked:** origin/mongo_index_creation


libs/community/langchain_community/chat_message_histories/mongodb.py
```diff
@@ -47,7 +49,11 @@ def __init__(
 
         self.db = self.client[database_name]
         self.collection = self.db[collection_name]
-        self.collection.create_index("SessionId")
+        if index_creation: 
+            """
+            Conditional Index Creation for collection
+            """
+            self.collection.create_index("SessionId")
```
baskaryan, line 56: would you ever want to specify index_name as well? if so could there be one paramter
```python
index_name: Optional[str] = "SessionID"
```
and if index_name is None then you skip index creation?

nvenkat94: I believe the 'index_name' parameter is not necessary at the moment. However, we currently use 'SessionID' in three other places. If provided, users can modify  the key name as they wish. If you wish add **index_name** parameter, I can do it.

baskaryan (approved):

PR #17761, opened 2024-02-19
langchain: add openai-tools mode to create_structured_output_runnable

ccurme:
Add `"openai-tools"` mode to `create_structured_output_runnable`.

Also fixes an example in the API reference for `"openai-json"` mode.


ccurme:
Closing as this is already covered in https://github.com/langchain-ai/langchain/pull/17296.

PR #17765, opened 2024-02-19
community: init signature revision for Cassandra LLM cache classes + small maintenance

hemidactylus:
This PR improves on the `CassandraCache` and `CassandraSemanticCache` classes, mainly in the constructor signature, and also introduces several minor improvements around these classes.

### Init signature

A (sigh) breaking change is tentatively introduced to the constructor. To me, the advantages outweigh the possible discomfort: the new syntax places the DB-connection objects `session` and `keyspace` later in the param list, so that they can be given a default value. This is what enables the pattern of _not_ specifying them, provided one has previously initialized the Cassandra connection through the versatile utility method `cassio.init(...)`.

In this way, a much less unwieldy instantiation can be done, such as `CassandraCache()` and `CassandraSemanticCache(embedding=xyz)`, everything else falling back to defaults.

A downside is that, compared to the earlier signature, this might turn out to be breaking for those doing positional instantiation. As a way to mitigate this problem, this PR typechecks its first argument trying to detect the legacy usage.
(And to make this point less tricky in the future, most arguments are left to be keyword-only).

If this is considered too harsh, I'd like guidance on how to further smoothen this transition. **Our plan is to make the pattern of optional session/keyspace a standard across all Cassandra classes**, so that a repeatable strategy would be ideal. A possibility would be to keep positional arguments for legacy reasons but issue a deprecation warning if any of them is actually used, to later remove them with 0.2 - please advise on this point.

### Other changes

- class docstrings: enriched, completely moved to class level, added note on `cassio.init(...)` pattern, added tiny sample usage code.
- semantic cache: revised terminology to never mention "distance" (it is in fact a similarity!). Kept the legacy constructor param with a deprecation warning if used.
- `llm_caching` notebook: uniform flow with the Cassandra and Astra DB separate cases; better and Cassandra-first description; all imports made explicit and from community where appropriate.
- cache integration tests moved to community (incl. the imported tools), env var bugfix for `CASSANDRA_CONTACT_POINTS`.


hemidactylus:
@baskaryan Hello, I'd like your opinion on this one. You may remember we already briefly discussed about the init signature change (several months back).
I think "the time has come" - I'd like some advice on how to do it in the most compliant, respectful way.

libs/community/langchain_community/cache.py
```diff
@@ -1016,36 +1016,71 @@ class CassandraCache(BaseCache):
...
+            set_llm_cache(CassandraCache())
+
     It uses a single Cassandra table.
     The lookup keys (which get to form the primary key) are:
         - prompt, a string
         - llm_string, a deterministic str representation of the model parameters.
-          (needed to prevent collisions same-prompt-different-model collisions)
+          (needed to prevent same-prompt-different-model collisions)
+
+    Constructor arguments:
```
cbornet, line 1038: The Sphinx doc section name is `Args`
```suggestion
    Args:
```

libs/community/langchain_community/cache.py
```diff
@@ -1016,36 +1016,71 @@ class CassandraCache(BaseCache):
...
+        table_name (str, optional): name of the Cassandra table to use as cache
```
cbornet, line 1039: param types are not really needed. They'll be inferred from type hints.

libs/community/langchain_community/cache.py
```diff
@@ -1016,36 +1016,71 @@ class CassandraCache(BaseCache):
...
+        session (cassandra.cluster.Session, None): an open Cassandra session.
+            Leave unspecified to use the global cassio init (see below)
+        keyspace (str, None): the keyspace to use for storing the cache.
+            Leave unspecified to use the global cassio init (see below)
+        ttl_seconds (optional int): time-to-live for cache entries
+            (default: None, i.e. forever)
+        skip_provisioning (bool, False): you can pass True if the Cassandra
+            table is guaranteed to exist already for a faster initialization.
+
+    On using the global cassio.init():
```
cbornet, line 1049: Could be a Sphinx `Note` section ?

libs/community/langchain_community/cache.py
```diff
@@ -1125,55 +1160,108 @@ class CassandraSemanticCache(BaseCache):
...
     cached values from several LLMs, so the LLM's llm_string is part
     of the rows' primary keys.
 
-    The similarity is based on one of several distance metrics (default: "dot").
-    If choosing another metric, the default threshold is to be re-tuned accordingly.
+    One can choose a similarity measure (default: "dot" for dot-product).
+    Choosing another one ("cos", "l2") almost certain requires threshold tuning.
+    (which may be required nevertheless even if sticking to "dot").
+
+    Constructor arguments:
```
cbornet, line 1189:
```suggestion
    Args:
```

libs/community/langchain_community/cache.py
```diff
@@ -1034,36 +1034,71 @@ class CassandraCache(BaseCache):
...
+        The session and keyspace parameters, when left out (or passed as None),
+        fall back to the globally-available cassio settings if are available.
+        In other words, if a previously-run 'cassio.init(...)' has been
+        executed previously anywhere in the code, Cassandra-based objects
+        need not specify the connection parameters at all.
     """
 
     def __init__(
         self,
+        table_name: str = CASSANDRA_CACHE_DEFAULT_TABLE_NAME,
```
baskaryan, line 1077: this is a breaking change, would prefer to avoid it

libs/community/langchain_community/cache.py
```diff
@@ -1143,55 +1178,108 @@ class CassandraSemanticCache(BaseCache):
...
-        session: Optional[CassandraSession],
-        keyspace: Optional[str],
         embedding: Embeddings,
         table_name: str = CASSANDRA_SEMANTIC_CACHE_DEFAULT_TABLE_NAME,
-        distance_metric: str = CASSANDRA_SEMANTIC_CACHE_DEFAULT_DISTANCE_METRIC,
+        *,
+        session: Optional[CassandraSession] = None,
+        keyspace: Optional[str] = None,
+        distance_metric: Optional[str] = None,
+        similarity_measure: str = CASSANDRA_SEMANTIC_CACHE_DEFAULT_SIMILARITY,
```
baskaryan, line 1243: same here

libs/community/langchain_community/cache.py
hemidactylus: This is, in a sense, the point of this PR. Using the `cassio.init()` construct to set a globally-available database connection would enable a much easier constructor pattern where one can omit the session/keyspace arguments altogether when creating table-based LangChain objects.

I am aware this is a breaking change, on the other hand it would make it possible to use this pattern like `CassandraCache("table")` instead of the more unwieldy `CassandraCache(None, None, "table")`, something that would really be an advantage in the long run.

Would you have a suggestion on how to best make this transition happen by respecting legacy usage as much as possible?

hemidactylus:
Hello @baskaryan , we have revised our strategy to simplify the init signatures (in view of the global Cassandra connection available through CassIO).
In line with the similar PR https://github.com/langchain-ai/langchain/pull/21209 by @cbornet (that one on the vector store class), I have modified this PR so as to be **fully backward-compatible** yet offer the shorter constructor form. Please let me know what you think.

(Note: I duplicated a utility module used in testing - the fake embeddings - to avoid introducing additional community dependencies within (core)langchain, in compliance with the ongoing effort to decouple them as much as possible).

libs/community/langchain_community/cache.py
```diff
@@ -1103,6 +1130,7 @@ def __init__(
         )
 
     def lookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
+        """Look up based on prompt and llm_string."""
```
cbornet, line 1133: The docstring is inherited from the parent class. So no need to override it with one with less details.
```suggestion
```

libs/community/langchain_community/cache.py
```diff
@@ -1123,6 +1151,7 @@ async def alookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYP
             return None
 
     def update(self, prompt: str, llm_string: str, return_val: RETURN_VAL_TYPE) -> None:
+        """Update cache based on prompt and llm_string."""
```
cbornet, line 1154:
```suggestion
```

libs/community/langchain_community/cache.py
```diff
@@ -1182,60 +1211,109 @@ class CassandraSemanticCache(BaseCache):
...
             )
+
+        if not embedding:
+            raise ValueError("Missing required parameter 'embedding'.")
+
+        # detect if legacy 'distance_metric' parameter used
+        if distance_metric is not None:
+            # if passed, takes precedence over 'similarity_measure',
+            # but we issue a warning.
+            warnings.warn(
```
cbornet, line 1305: Should it be deprecated ?

libs/community/langchain_community/cache.py
```diff
@@ -1285,6 +1363,7 @@ async def _aget_embedding_dimension(self) -> int:
         return len(await self._aget_embedding(text="This is a sample sentence."))
 
     def update(self, prompt: str, llm_string: str, return_val: RETURN_VAL_TYPE) -> None:
+        """Update cache based on prompt and llm_string."""
```
cbornet, line 1366:
```suggestion
```

libs/community/langchain_community/cache.py
```diff
@@ -1321,6 +1400,7 @@ async def aupdate(
         )
 
     def lookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
+        """Look up based on prompt and llm_string."""
```
cbornet, line 1403:
```suggestion
```

cbornet (changes requested):

libs/community/langchain_community/cache.py
hemidactylus: Well spotted, thanks

hemidactylus:
Update: I reverted the location of the Cassandra cache test file to `libs/langchain` since it requires imports `from langchain...`. So the duplicate fake embedding module is not necessary anymore.

On the other hand, I have bumped the cassio version in the pyproject.toml of `libs/langchain` to ensure the async usage (which uses awaitables for the embedding dimensions, among other things) is handled correctly by CassIO.

cbornet (approved):
LGTM, great work !

hemidactylus:
@eyurtsev drawing your attention to the CI failures above. For reasons I cannot investigate successfully the environment pulls `vcrpy==4.3.0` in, as opposed to `6.0.1` (as on master branch). This seems to be the cause of the error in the pytest-compile step.
You may be better equipped to help the (libs/langchain/) pyproject restore the desired higher version of vcrpy. I am not sure what I am doing wrong. :pray:

hemidactylus:
Adding `vcrpy = "^6.0.1"` explicitly should solve the issue - I hope that's acceptable.

hemidactylus:
Something is off in the "extended tests" which does not seem related to this PR at all.

The `test_load_llmchain_with_non_serializable_arg` test passes an httpx.Client instance where an AsyncClient is now expected. I wonder if this same test passes on master (scratches head).

libs/community/langchain_community/cache.py
```diff
@@ -1182,60 +1210,110 @@ class CassandraSemanticCache(BaseCache):
...
+            to use as cache. There is a default for "simple" usage, but
+            remember to explicitly specify different tables if several embedding
+            models coexist in your app (they cannot share one cache table).
+        score_threshold: numeric value to use as
+            cutoff for the similarity searches
+        ttl_seconds: time-to-live for cache entries
+            (default: None, i.e. forever)
+        similarity_measure: which measure to adopt for
+            similarity searches
+        setup_mode: a value in langchain_community.utilities.cassandra.SetupMode.
```
efriis, line 1256: can we add distance_metric

libs/community/langchain_community/cache.py
```diff
@@ -1170,7 +1198,7 @@ async def aclear(self, **kwargs: Any) -> None:
         await self.kv_cache.aclear()
 
 
-CASSANDRA_SEMANTIC_CACHE_DEFAULT_DISTANCE_METRIC = "dot"
+CASSANDRA_SEMANTIC_CACHE_DEFAULT_SIMILARITY = "dot"
```
efriis, line 1201: can we keep this name for backwards compatibility

libs/community/langchain_community/cache.py
hemidactylus: All right. I reinstated it in the docstring, with a note saying that one should prefer the (equivalent) "similarity" alias.

libs/community/langchain_community/cache.py
hemidactylus: Sure. I left a comment in the code for clarity, stating that this is actually a "similarity" (back in the time, we had a bit of a terminology oscillation between similarity/distance, and this is sadly a legacy of that era).

efriis (approved):

PR #17802, opened 2024-02-20
community: Fix AstraDBCache docstrings

cbornet:


libs/community/langchain_community/cache.py
```diff
@@ -1374,6 +1374,22 @@ class AstraDBCache(BaseCache):
         - prompt, a string
         - llm_string, a deterministic str representation of the model parameters.
           (needed to prevent same-prompt-different-model collisions)
+
```
eyurtsev, line 1377: Hi @cbornet, we follow google style documentation https://google.github.io/styleguide/pyguide.html#384-classes

Class level documentation is for documenting public attributes. I suspect that none of these attributes are meant to be used as public attributes, so documenting them only on the __init__ makes the most sense to me.

What's the reason for making this change?

cbornet: Well, Sphinx default setup doesn't document the `__init__` so it's a common usage to put the args in the class docstring. 
If we look at https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google (see classes ExampleError and ExampleClass) both are allowed:
```
        The __init__ method may be documented in either the class level
        docstring, or as a docstring on the __init__ method itself.

        Either form is acceptable, but the two should not be mixed. Choose one
        convention to document the __init__ method and be consistent with it.
```
Also PyCharm will use the class doc if it doesn't find an __init__ doc. It's then nice to put a lot of info into the class and get that info when hovering a class instanciation.

That said, I see that the LangCHain setup does document the __init__ method separately and copies its content to the class doc.
So probably the best is to put most useful info into the __init__ and very basic headline in the class. This way, both PyCharm and LC docs will be happy.
I'll update the PRs.

cbornet: done. PTAL.

eyurtsev (approved):

PR #17807, opened 2024-02-20
community[patch]: add missing chunk parameter for _stream/_astream

mackong:
- Description: Add missing chunk parameter for _stream/_astream for some chat models, make all chat models in a consistent behaviour.
- Issue: N/A
- Dependencies: N/A

### DIFF
--- a/libs/community/langchain_community/chat_models/baichuan.py
+++ b/libs/community/langchain_community/chat_models/baichuan.py
@@ -218,9 +218,10 @@ class ChatBaichuan(BaseChatModel):
                     m.get("delta"), default_chunk_class
                 )
                 default_chunk_class = chunk.__class__
-                yield ChatGenerationChunk(message=chunk)
+                cg_chunk = ChatGenerationChunk(message=chunk)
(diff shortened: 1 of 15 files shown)


eyurtsev (approved):

eyurtsev:
@mackong thank you very much for the patch!

There's another issue with most of these models. They should be calling the callback prior to yielding the chunk rather than after yielding the chunk

eyurtsev:
@mackong if your'e able to resolve the linting error, we can merge right away

mackong:
> @mackong if your'e able to resolve the linting error, we can merge right away

Fixed.

mackong:
> @mackong thank you very much for the patch!
> 
> There's another issue with most of these models. They should be calling the callback prior to yielding the chunk rather than after yielding the chunk. If you have interest, we'd appreciate another PR for that as well :)

OK, I'll fix it in another PR.

eyurtsev (approved):

eyurtsev:
merge when tests pass

PR #17810, opened 2024-02-20
langchain: Make BooleanOutputParser more robust to non-binary responses

dokato:
- **Description:** I encountered this error when I tried to use LLMChainFilter. Even if the message slightly differs, like `Not relevant (NO)` this results in an error. It has been reported already here: https://github.com/langchain-ai/langchain/issues/. This change hopefully makes it more robust.
- **Issue:**  #11408 
- **Dependencies:** No
- **Twitter handle:** dokatox

### DIFF
--- a/libs/langchain/langchain/output_parsers/boolean.py
+++ b/libs/langchain/langchain/output_parsers/boolean.py
@@ -25,7 +25,7 @@ class BooleanOutputParser(BaseOutputParser[bool]):
             and self.false_val.upper() in cleaned_upper_text
         ):
             raise ValueError(
-                f"Ambigous response. Both {self.true_val} and {self.false_val} in "
+                f"Ambiguous response. Both {self.true_val} and {self.false_val} in "
                 f"received: {text}."
             )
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/output_parsers/boolean.py
```diff
@@ -19,13 +19,16 @@ def parse(self, text: str) -> bool:
             boolean
 
         """
-        cleaned_text = text.strip()
-        if cleaned_text.upper() not in (self.true_val.upper(), self.false_val.upper()):
+        cleaned_upper_text = text.strip().upper()
+        if self.true_val.upper() in cleaned_upper_text:
```
baskaryan, line 23: probably worth first checking that both true_val and false_val don't occur in text

baskaryan (commented):
could we add some unit tests

libs/langchain/langchain/output_parsers/boolean.py
dokato: yup, good point! Added with tests! thanks for checking

baskaryan (approved):

PR #17832, opened 2024-02-21
community[major]: `langchain.indexes` duplicate code removal

leo-gan:
It looks like the [langchain_community.indexes](https://api.python.langchain.com/en/latest/community_api_reference.html#module-langchain_community.indexes) artifacts are the artifact from some unfinished refactoring. They are copies of the [langchain.indexes](https://api.python.langchain.com/en/latest/langchain_api_reference.html#module-langchain.indexes) artifacts.
Removed community `base.py` is equal to the langchain `base.py`. Removed community `_sql_record_manager.py` had some unfinished updates in typing, no code logic changes.

I've removed the `langchain_community.indexes` artifacts and the correspondent uts.


baskaryan (commented):
i think what we actually want to do is remove the versions in langchain and import from community

leo-gan:
> i think what we actually want to do is remove the versions in langchain and import from community

@baskaryan I'm not sure I understand. Code in `indexes` depends only on `SQLAlchemy` package, that's it. It is not actually a community integration. It is a generic dependency on SQL.
We have several similar dependencies on SQL in `langchain`. For example, [chains.sql_database.query.create_sql_query_chain](https://api.python.langchain.com/en/latest/chains/langchain.chains.sql_database.query.create_sql_query_chain.html#langchain.chains.sql_database.query.create_sql_query_chain) , [chains.sql_database.query.SQLInputWithTables](https://api.python.langchain.com/en/latest/chains/langchain.chains.sql_database.query.SQLInputWithTables.html#langchain.chains.sql_database.query.SQLInputWithTables)
So, it is not considered as a `community` dependent code, no?

Yeah, we have this definitely `community` dependency [memory.entity.SQLiteEntityStore](https://api.python.langchain.com/en/latest/memory/langchain.memory.entity.SQLiteEntityStore.html#langchain.memory.entity.SQLiteEntityStore) but it was addressed in this [PR](https://github.com/langchain-ai/langchain/pull/15437) which desperately waits to be reviewed almost 2 months. :see_no_evil: :hear_no_evil:

eyurtsev:
@baskaryan i think this stuff should live in langchain or langchain-core and be provider agnostic. We only need the implementation (e.g., SQL) to live in community

eyurtsev:
I'm OK removing everything here from community. We can move the SQL Record manager to langchain_postgres at some point.

@leo-gan is it possible to add some sort of import error that would be helpful for users or are we sure that no one is using it from langchain_community?

leo-gan:
Sorry, I've messed up with branches. I'll recreate it from scratch again.

PR #17856, opened 2024-02-21
partner: Add groq partner integration and chat model

gradenr:
Description: Add a Groq chat model
issue: TODO
Dependencies: groq
Twitter handle: N/A


libs/partners/groq/pyproject.toml
```diff
@@ -0,0 +1,95 @@
...
+repository = "https://github.com/langchain-ai/langchain"
+license = "MIT"
+
+[tool.poetry.urls]
+"Source Code" = "https://github.com/langchain-ai/langchain/tree/master/libs/partners/groq"
+
+[tool.poetry.dependencies]
+python = ">=3.8.1,<4.0"
+langchain-core = ">=0.0.12"
+groq = "^0.4.1"
```
efriis, line 16: Do we want to explicitly exclude a 0.5, or should this be `>=0.4.1,<1`

.github/workflows/_release.yml
```diff
@@ -181,6 +181,7 @@ jobs:
           NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
           GOOGLE_SEARCH_API_KEY: ${{ secrets.GOOGLE_SEARCH_API_KEY }}
           GOOGLE_CSE_ID: ${{ secrets.GOOGLE_CSE_ID }}
+          GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
```
efriis, line 184: Is there a way we can get an API key with a higher rate limit window that we only use for running integration tests on releases?

Current release flow will fail because of rate limit errors in integration testing.

efriis (commented):
Some comments!

In future, would be great to get PRs from personal branches because Github blocks us from making small edits to org branches. More info: https://python.langchain.com/docs/contributing/faq#how-do-i-allow-maintainers-to-edit-my-pr

efriis:
Also don't worry about filing an issue!

.github/workflows/_release.yml
gradenr: I am looking into this.

gradenr:
Updated dep versions in pyproject.

efriis (approved):

PR #17890, opened 2024-02-21
docs: API Reference builder bug fix

leo-gan:
Issue in the API Reference:
If the `Classes` of `Functions` section is empty, it still shown in API Reference. Here is an [example](https://api.python.langchain.com/en/latest/core_api_reference.html#module-langchain_core.agents) where `Functions` table is empty but still presented.
It happens only if this section has only the "private" members (with names started with '_'). Those members are not shown but the whole member section (empty) is shown.

### DIFF
--- a/docs/api_reference/create_api_rst.py
+++ b/docs/api_reference/create_api_rst.py
@@ -217,8 +217,8 @@ def _construct_doc(
 
     for module in namespaces:
         _members = members_by_namespace[module]
-        classes = _members["classes_"]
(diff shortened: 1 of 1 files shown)


eyurtsev (approved):

leo-gan:
@eyurtsev Thanks! Eugene, could you, please, build the API Ref to check if this fix works? I can not do this locally now (in process of debugging the problem with this build)

PR #17904, opened 2024-02-22
docs: api docs for external repos

efriis:
Stacked on google removal PR. Will make google continue to show up in API docs even from external repo

### DIFF
--- a/.github/workflows/api_doc_build.yml
+++ b/.github/workflows/api_doc_build.yml
@@ -15,6 +15,15 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: bagatur/api_docs_build
+          path: langchain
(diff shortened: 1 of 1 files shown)


.github/workflows/api_doc_build.yml
```diff
@@ -15,13 +15,23 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: bagatur/api_docs_build
+          path: main
```
efriis, line 18: sets default repo working directory to ./main

baskaryan: maybe just call this langchain? and google langchain-google? ie the repo names

baskaryan (approved):

PR #17907, opened 2024-02-22
community[patch]: callback before yield for _stream/_astream

mackong:
- Description: callback on_llm_new_token before yield chunk for _stream/_astream for some chat models, make all chat models in a consistent behaviour.
- Issue: N/A
- Dependencies: N/A


baskaryan:
fixes #16913

baskaryan (approved):

PR #17938, opened 2024-02-22
community: added the possibility to return document ids as part of the results of similarity search in vectorstore chroma

majdabd:
Hello, I modified the `_results_to_docs_and_scores` method in `chroma.py` to allow the results of the similarity search to include the IDs of documents or document chunks. I'm new to Chroma and LangChain, but I couldn't find a way to retrieve the IDs of documents that best match the query. 

The new `_results_to_docs_and_scores` thus becomes: 

```
def _results_to_docs_and_scores(results: Any) -> List[Tuple[Document, float]]:
    return [
        # TODO: Chroma can do batch querying,
        # we shouldn't hard code to the 1st result
        (Document(page_content=result[0], metadata=(result[1] | {'id': result[3]}) or {}), result[2])
        for result in zip(
            results["documents"][0],
            results["metadatas"][0],
            results["distances"][0],
            results["ids"][0]
        )
    ]
```

Note: I found this solution in this closed issue https://github.com/langchain-ai/langchain/issues/11592, but as far as I know it has not been proposed in a pull request.


baskaryan:
cc @jeffchuber

libs/community/langchain_community/vectorstores/chroma.py
```diff
@@ -41,11 +41,17 @@ def _results_to_docs_and_scores(results: Any) -> List[Tuple[Document, float]]:
     return [
         # TODO: Chroma can do batch querying,
         # we shouldn't hard code to the 1st result
-        (Document(page_content=result[0], metadata=result[1] or {}), result[2])
+        (
+            Document(
+                page_content=result[0], metadata=(result[1] | {"id": result[3]}) or {}
```
jeffchuber, line 46: this looks like if you have metadata it will override an `id` prop... can we think of a different way to handle this? i think it may really confuse some users, albeit an edge case

TheDeafOne:
@hwchase17 why was this closed? Was there a solution merged? Having this functionality would be super useful.

LDelPinoNT:
From the messages I understand there is a variable called id, but it can be useful to have the document id, maybe as "doc_id" (so there is no name overlapping).

An example of use case for the "doc_id" is recover it for training and evaluating RAG with InformationRetrievalEvaluator of the package SentenceEmbeddings (https://sbert.net/docs/package_reference/sentence_transformer/evaluation.html#informationretrievalevaluator) This evaluator needs pairs of documents and documents ids as "corpus" parameter which can be retrieved directly from VectorStores.

PR #17946, opened 2024-02-22
langchain: fix EmbeddingsFilter compress_documents return type to Sequence[Document] instead of Sequence[_DocumentWithState]

maximeperrindev:
- **Description:** This PR solves a typing problem encountered using EmbeddingsFilter.compress_documents method with langchain. The returned Sequence of _DocumentWithState was not meeting the typing expectation of Sequence[Document]. This could cause a problem of json parsing because of embeddings field in _DocumentWithState.
- **Issue:** #17875
- **Twitter handle:** @maximeperrin_

### DIFF
--- a/libs/langchain/langchain/retrievers/document_compressors/embeddings_filter.py
+++ b/libs/langchain/langchain/retrievers/document_compressors/embeddings_filter.py
@@ -69,4 +69,4 @@ class EmbeddingsFilter(BaseDocumentCompressor):
             included_idxs = included_idxs[similar_enough]
         for i in included_idxs:
             stateful_documents[i].state["query_similarity_score"] = similarity[i]
-        return [stateful_documents[i] for i in included_idxs]
+        return [stateful_documents[i].to_document() for i in included_idxs]


maximeperrindev:
@eyurtsev

baskaryan:
this is actually intentional, it's needed that this returns a stateful document so that embeddings aren't recomputed multiple times in a compression pipeline (where multiple compressors use embeddings). the correct solution would be to convert any stateful docs to non-stateful docs outside of this class.

closing but let me know if i'm missing something

hpx502766238:
> this is actually intentional, it's needed that this returns a stateful document so that embeddings aren't recomputed multiple times in a compression pipeline (where multiple compressors use embeddings). the correct solution would be to convert any stateful docs to non-stateful docs outside of this class.
> 
> closing but let me know if i'm missing something

Excuse me.Can you tell me how to convert any stateful docs to non-stateful docs?

hpx502766238:
I have found the solution.
We can convert (_DocumentWithState) to (Document) before an ContextualCompressionRetriever return.
in langchain/retrievers/contextual_compression.py, method _get_relevant_documents and async def _aget_relevant_documents,I changed the code like following:
docs = await self.base_retriever.ainvoke(
            query, config={"callbacks": run_manager.get_child()}, **kwargs
        )
        if docs:
            compressed_docs = await self.base_compressor.acompress_documents(
                docs, query, callbacks=run_manager.get_child()
            )
            **compressed_docs_converted = [
                doc.to_document() if isinstance(doc, _DocumentWithState) else doc
                for doc in compressed_docs
            ]
#            print("after:", compressed_docs_converted)
            return compressed_docs_converted**
        else:
            return []

pauljaspersahr:
> this is actually intentional, it's needed that this returns a stateful document so that embeddings aren't recomputed multiple times in a compression pipeline (where multiple compressors use embeddings). the correct solution would be to convert any stateful docs to non-stateful docs outside of this class.
> 
> closing but let me know if i'm missing something

I guess its not intentional that [official langchain examples](https://python.langchain.com/docs/how_to/contextual_compression/#llmchainfilter) don't work and we need to use workaround from [here](https://github.com/langchain-ai/langchain/discussions/17876#discussioncomment-8591218):
```python
class ToDocEmbeddingsFilter(EmbeddingsFilter):
    def compress_documents(
        self,
        documents: Sequence[Document],
        query: str,
        callbacks: Optional[Callbacks] = None,
    ) -> Sequence[Document]:
        docs = []
        for doc in super().compress_documents( documents, query, callbacks):
            docs.append(doc.to_document())
        return docs
```

Issue #17963, opened 2024-02-22
community: [MongoDBAtlasVectorSearch] Fix KeyError 'embedding' when using MMR


certified-dodo:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code
```python
# test.py
from langchain_openai import OpenAIEmbeddings
from langchain_community.vectorstores.mongodb_atlas import MongoDBAtlasVectorSearch

embeddings = OpenAIEmbeddings()

MONGO_CONNECTION_URI = "mongo uri"
VECTORS_DB_NAME = "db name"
VECTOR_SEARCH_INDEX_NAME = "index name"

db = MongoDBAtlasVectorSearch.from_connection_string(
    MONGO_CONNECTION_URI,
    VECTORS_DB_NAME,
    embeddings,
    index_name=VECTOR_SEARCH_INDEX_NAME,
)

db_query = "Find info related to ..."
num_of_db_results = 10

raw_contexts = db.max_marginal_relevance_search(query=db_query, k=num_of_db_results, lambda_mult=0)
```

### Error Message and Stack Trace (if applicable)

```cmd
Traceback (most recent call last):
  File ".../test.py", line 39, in <module>
    raw_contexts = db.max_marginal_relevance_search(query=db_query, k=num_of_db_results, lambda_mult=0)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../lib/python3.12/site-packages/langchain_community/vectorstores/mongodb_atlas.py", line 325, in max_marginal_relevance_search
    [doc.metadata[self._embedding_key] for doc, _ in docs],
     ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
KeyError: 'embedding'
```

### Description
* Error when using MMR (max marginal relevance) search. This includes
  - calling. `max_marginal_relevance_search` directly
  - using `db.as_retriever(search_type="mmr")`
* `self._embedding_key` is deleted from `docs` in `_similarity_search_with_score`
[...]

PR #17971, opened 2024-02-22
community: Fix MongoDBAtlasVectorSearch max_marginal_relevance_search

certified-dodo:
Description:
* `self._embedding_key` is accessed after deletion, breaking `max_marginal_relevance_search` search
* Introduced in: https://github.com/langchain-ai/langchain/commit/e135e5257c4ed8185308b976a1dde1f0951d1324
* Updated but still persists in: https://github.com/langchain-ai/langchain/commit/ce22e10c4b18b9706bf6630c248e29597a86a583

Issue: https://github.com/langchain-ai/langchain/issues/17963


baskaryan:
hm i wonder if a better fix would be to re-implement mmr to not rely on similarity_search, since end users probably don't need/want to be passing around embeddings in their docs?

certified-dodo:
> hm i wonder if a better fix would be to re-implement mmr to not rely on similarity_search, since end users probably don't need/want to be passing around embeddings in their docs?

yeah probably, but hesitant to move around too much code. 
What about adding a param in similarity for something like this?
```
# similarity(deleteEmbedding: bool = True), False for mmr
  if (deleteEmbedding):
      del res[self._embedding_key]
```
and then delete it later 
```
# mmr
  for mmr_doc in mmr_docs:
      del mmr_doc.metadata[self._embedding_key]

```

prakul:
@baskaryan Yeah this is a bug being addressed in another larger PR but LGTM on this in the meantime

baskaryan (approved):

PR #17981, opened 2024-02-22
Correct WebBaseLoader URL: docs: python.langchain.com/docs/get_started/quickstartQuickstart

rmeinzer-copado:
**Description:** 
The URL of the data to index, specified to `WebBaseLoader` to import is incorrect, causing the `langsmith_search` retriever to return a `404: NOT_FOUND`. 
Incorrect URL: https://docs.smith.langchain.com/overview
Correct URL: https://docs.smith.langchain.com

**Issue:** 
This commit corrects the URL and prevents the LangServe Playground from returning an error from its inability to use the retriever when inquiring, "how can langsmith help with testing?". 

**Dependencies:** 
None.

**Twitter Handle:** 
@ryanmeinzer


baskaryan (approved):

baskaryan (commented):
@rmeinzer-copado wonder if https://docs.smith.langchain.com/user_guide might be a better page to load, the home page is pretty sparse

rmeinzer-copado:
> @rmeinzer-copado wonder if https://docs.smith.langchain.com/user_guide might be a better page to load, the home page is pretty sparse

Thanks, @baskaryan! While I tend to agree with you from a qualitative standpoint, I believe my proposal is best given it corrects the error at hand and reflects the same documentation URL previously mentioned in the Quickstart for simplicity's sake.

baskaryan:
> > @rmeinzer-copado wonder if https://docs.smith.langchain.com/user_guide might be a better page to load, the home page is pretty sparse
> 
> Thanks, @baskaryan! While I tend to agree with you from a qualitative standpoint, I believe my proposal is best given it corrects the error at hand and reflects the same documentation URL previously mentioned in the Quickstart for simplicity's sake.

the issue was caused by a total revamp of the langsmith docs, so i don't think pointing to a similar page as before matters too much (the current home page has a lot less content than the former quickstart, so it's not really the same page anyways). would prefer to point to something that makes the documentation example more useful

rmeinzer-copado:
> > > @rmeinzer-copado wonder if https://docs.smith.langchain.com/user_guide might be a better page to load, the home page is pretty sparse
> > 
> > 
> > Thanks, @baskaryan! While I tend to agree with you from a qualitative standpoint, I believe my proposal is best given it corrects the error at hand and reflects the same documentation URL previously mentioned in the Quickstart for simplicity's sake.
> 
> the issue was caused by a total revamp of the langsmith docs, so i don't think pointing to a similar page as before matters too much (the current home page has a lot less content than the former quickstart, so it's not really the same page anyways). would prefer to point to something that makes the documentation example more useful

Thanks, @baskaryan. I understand, however, I was noting that I was mirroring the *other* URL in the Quickstart for consistency's sake. Regardless, I've just updated both URLs.

Issue #5340, opened 2023-05-27
Implement a way to abort / cancel request


dzmitry-kankalovich:
### Feature request

I've been playing around with OpenAI GPT-4 and ran into situation when response generation might take quite some time - say 5 minutes. 

I switched over to streaming, but often I can immediately see the response is not what want, and therefore I'd like to cancel request.

Now here is the part that is unclear to me: 
**is there an official way to cancel request in Python's version of LangChain?** I have found this [described](https://js.langchain.com/docs/modules/models/chat/additional_functionality#cancelling-requests) in JS/TS version of the framework, however scanning docs, sources and issues yields nothing for this repo.

For now I simply terminate process, which works good enough for something like Jupyter notebooks, but quite problematic for say web application.

Besides termination, it's also unclear if I may incur any unwanted costs or not for the abandoned request.

Should some sort of feature parity be made with JS LangChain?

### Motivation

Provide a documented way to cancel long-running requests

### Your contribution

At this point I have capacity only to test out potential implementation. May work on the implementation in later weeks.

jsegeren:
+1 I also need this.

grp06:
did you find a way to do this?

dzmitry-kankalovich:
Nope. What I did to workaround it:

1. I've added real-time output of streamed data to the page - gives nice UX, at least you see response being generated instead of waiting at a blank page
2. I've restructured my prompts to be faster
3. Got access to the company's AI proxy, which works with Azure's OpenAI version - and it is apparently like twice faster. 

So the need to have this kinda less pressing now, but it still seems a must to have

dzmitry-kankalovich:
> dosu-beta

Seems to still be relevant. 

I would be able to improve UX of my software significantly if I can added cancellation feature

mdfederici:
I would also like to be able to cancel streams gracefully from the python client.

lmeyerov:
Is there any update on the right patterns to cancel chains?

This is an architectural need at multiple levels:

* When langchain is managing an LLM stream, or running a long HTTP Request call, this impacts the $ & quota by the consumer . Imagine a user erroneously realizing they're doing a massive call and wanting to cancel.

* When there is a chain sequencing a bunch of RAG etc, being able to top the sequence becomes important, imagine running & stopping an autoGPT

Supporting everywhere at once is obviously not easy, but providing a pattern for the basics -- batch & streaming chain calls to openai -- seems like a clear start...

rokosamit:
[...]

PR #18013, opened 2024-02-23
openai: wrap stream code in context manager blocks

snopoke:
**Description:**
Use the `Stream` context managers in `ChatOpenAi` `stream` and `astream` method.

Using the context manager returned by the OpenAI client makes it possible to terminate the stream early since the response connection will be closed when the context manager exists.

**Issue:** #5340
**Twitter handle:** @snopoke


snopoke:
@efriis The build failed with what appears to be a CI error:
```
Error when evaluating 'strategy' for job 'extended-tests'. .github/workflows/check_diffs.yml (Line: 88, Col: 28): Matrix vector 'working-directory' does not contain any values
```

I don't see anything obviously wrong with the configuration or CI scripts. Can you re-run the build please.

snopoke:
@efriis bump

baskaryan (approved):

PR #18016, opened 2024-02-23
fix: bad request issue when filter is empty

adorabledev:
community: fix -bad request issue when filter is empty

Description:
Notion DB API returns 400 bad request if sending filter with an empty {}


fyang0507:
Why was this PR closed w/o merge? Encountered the same issue

PR #18017, opened 2024-02-23
community: Add embedding instruction to HuggingFaceBgeEmbeddings

wt3639:
-   **Description:** Add embedding instruction to HuggingFaceBgeEmbeddings, so that it can be compatible with nomic and other models that need embedding instruction.


baskaryan (approved):

mkhludnev:
fwiw

https://huggingface.co/BAAI/bge-m3
> The only difference is that the BGE-M3 model no longer requires adding instructions to the queries. 

It's obvious, but I'm not sure should  we notify users somehow.

PR #18021, opened 2024-02-23
Add main graph documentation

tomasonjo:


hwchase17:
this generally looks really good to me! i think the only thing out of place is all the examples of different types of graph dbs. I think they probably don't belong here and more belong in integrations? for example, with retrieval QA we dont list out all the vector databases

thoughts @tomasonjo @baskaryan ?

baskaryan (approved):

PR #18023, opened 2024-02-23
core: add RunnableFactory class

skozlovf:
**Description:**
This PR adds a new `RunnableFactory` class that allows to create another runnable at runtime using a factory method.
Key features:
* Loadable parameters and validation thanks to ``pydantic``.
* Inheritance support and all other OOP features.
* Serialize only chain specific parameters.

Example usage:
```python
from langchain_core.runnables import Runnable, RunnableFactory
from langchain_core.pydantic_v1 import validator

class MyRunnable(RunnableFactory):
    role: str = "You are a helpful AI assistant."
    prompt: str = "Tell me a joke about {topic}"
    model: str = "gpt-3.5-turbo"

    @validator("prompt")
    def validate_prompt(cls, v):
        if "joke" not in v:
            raise ValueError("You should only ask for jokes.")
        return v

    def _create_runnable(self) -> Runnable:
            [("system", self.role), ("human", self.prompt)]
        )
        llm = ChatOpenAI(model_name=self.model)

        return chain

# Create a chain with defaults.

chain = MyRunnable()
output = chain.invoke({"topic": "bear"})
print(output)

# Create a chain with custom parameters.

chain = MyRunnable(model="gpt-4-1106-preview")
output = chain.invoke({"topic": "bear"})
print(output)

# Create a chain from an external "source".

external_parameters = {
    "model": "gpt-4-1106-preview",
    "prompt": "Tell me two jokes about {topic}",
}

chain = MyRunnable.parse_obj(external_parameters)
#
# or
# chain = MyRunnable.parse_file(filename)
#
output = chain.invoke({"topic": "bear"})
print(output)

# Serialize parameters.
parameters = chain.dict()
chain = MyRunnable.parse_obj(parameters)
output = chain.invoke({"topic": "bear"})
print(output)
```

If you think this make sense, I can continue adding docs/tests.

### DIFF
--- a/libs/core/langchain_core/runnables/factory.py
+++ b/libs/core/langchain_core/runnables/factory.py
@@ -1,5 +1,5 @@
 from abc import abstractmethod
-from typing import Any, AsyncIterator, Iterator, Optional, Sequence
+from typing import Any, AsyncIterator, Iterator, List, Optional, Sequence, Type, Union
 
 from langchain_core.pydantic_v1 import BaseModel, PrivateAttr
 from langchain_core.runnables import (
@@ -93,33 +93,33 @@ class RunnableFactory(RunnableSerializable[Input, Output]):
         return True
 
     @property
-    def InputType(self) -> type[Input]:
+    def InputType(self) -> Type[Input]:
         return self._runnable.InputType
 
     @property
-    def OutputType(self) -> type[Output]:
+    def OutputType(self) -> Type[Output]:
         return self._runnable.OutputType
 
     def get_input_schema(
         self, config: Optional[RunnableConfig] = None
-    ) -> type[BaseModel]:
+    ) -> Type[BaseModel]:
         return self._runnable.get_input_schema(config)
 
     def get_output_schema(
         self, config: Optional[RunnableConfig] = None
-    ) -> type[BaseModel]:
+    ) -> Type[BaseModel]:
         return self._runnable.get_output_schema(config)
 
     @property
-    def config_specs(self) -> list[ConfigurableFieldSpec]:
+    def config_specs(self) -> List[ConfigurableFieldSpec]:
         return self._runnable.config_specs
 
     def config_schema(
         self, *, include: Optional[Sequence[str]] = None
-    ) -> type[BaseModel]:
+    ) -> Type[BaseModel]:
         return self._runnable.config_schema(include=include)
 
-    def get_graph(self, config: RunnableConfig | None = None) -> Graph:
+    def get_graph(self, config: Optional[RunnableConfig] = None) -> Graph:
         return self._runnable.get_graph(config)
 
     def transform(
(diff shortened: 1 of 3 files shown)


eyurtsev:
What are some examples where a RunnableFactory would work better than a function?

```python
def create_runnable(role, prompt, model):
            [("system", role), ("human", prompt)]
        )
        llm = ChatOpenAI(model_name=model)

        return chain

runnable  = create_runnable(...)
```

skozlovf:
This is useful if you use OOP in your app.

If you use a plain class then flow is similar to the function way:
```python

class MyRunnable:
    role: str
    prompt: str
    model: BaseLanguageModel

    def create() -> Runnable:
        ...

chain = ...
factory = MyRunnable(...)
chain = chain | factory.create()
```
You can't put the class instance to a LCEL chain, you need a chain builder method that returns a `Runnable`.

`RunnableFactory` allows you to encapsulate your data and logic, and create custom runnable type.

This means that you can embed such instance to a LCEL chain directly, without extra actions:
```python
class MyRunnable(RunnableFactory):
    role: str
    prompt: str
    model: BaseLanguageModel

    def _create_runnable(self) -> Runnable:
        ...

chain = ...
my_chain = MyRunnable(...)
chain = chain | my_chain
```

Also, you can put it to a `pydantic` model field, and it will be automatically validated and created:
```python
class MyRunnable(RunnableFactory):
    role: str
    prompt: str
    model: BaseLanguageModel

    def _create_runnable(self) -> Runnable:
        ...

class AnotherRunnable(RunnableFactory):
    param: int
    my_chain: MyRunnable

    def _create_runnable(self) -> Runnable:
        chain = ...
        return chain | self.my_chain

chain = AnotherRunnable(...)
output = chain.invoke(...)
```

Side effect is what you can serialize *app-specific* parameters. `RunnableSerializable` allows to serialize already constructed chain. There is no documented way to load it, and furthermore, the serialized parameters are *chain-specific*:
```python

chain = AnotherRunnable.parse_obj(
    {
        "param": 2,
        "my_chain": {
            "role": "xxx",
            "prompt": "yyy",
        }
    }
)

params = chain.dict(...)
# {"param": 2, "my_chain": {"role": "xxx", "prompt": "yyy"}}
```

NOTE: I skipped `model` processing for simplicity.

PS Feel free to discard the PR if you don't think it should be included in the "core".

libs/core/langchain_core/runnables/factory.py
```diff
@@ -0,0 +1,189 @@
...
+        chain = MyRunnable.parse_obj(external_parameters)
+        output = chain.invoke({"topic": "bear"})
+        print(output)
+    """
+
+    _runnable: Runnable = PrivateAttr()
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._runnable = self._create_runnable()
```
baskaryan, line 85: if the runnable is created on init of the factory then it's not really a factory is it? you could've just initialized the underlying runnable, no?

skozlovf: That's right. The name is not very clear. Maybe `RunnableWrapper` or `RunnablePimpl` or smth else is better. One of the goals is to redirect API calls to the underlying runnable created at runtime. Let me know if it makes sense to put this whole idea in the core.

skozlovf:
Ok, it seems this class is very specific for the `core`. I close the PR as there are more important LC tasks.

PR #18030, opened 2024-02-23
community: add include_columns for db.run_no_throw under the class QuerySQLDataBaseTool

WuYunjin:
- **Description:** add include_columns arg for db.run_no_throw under the class QuerySQLDataBaseTool, so that we can set the include_columns parameter to True/False when Instancing a QuerySQLDataBaseTool like 

```
execute_query = QuerySQLDataBaseTool(db=db,metadata={"include_columns":True})
execute_query.invoke("SELECT COUNT(*) as employees_cnt FROM Employee") 
#return '[{'employees_cnt',8}]' , and return '[(8,)]' if set include_columns=False
```

 - **Issue:** not found yet, but very useful for my situation
 - **Dependencies:** None


WuYunjin:
@eyurtsev

libs/community/langchain_community/tools/sql_database/tool.py
```diff
@@ -45,8 +45,17 @@ def _run(
         query: str,
         run_manager: Optional[CallbackManagerForToolRun] = None,
     ) -> Union[str, Sequence[Dict[str, Any]], Result]:
+        # add include_columns as valid args for db.run_no_throw
+        valid_kwargs = {}
+        if self.metadata:
```
baskaryan, line 50: metadata isn't meant to affect tool outputs. include_columns should be its own tool on QuerySQLDataBaseTool

PR #18033, opened 2024-02-23
docs: recommend lambdas over runnablebranch

efriis:


docs/docs/integrations/chat/groq.ipynb
efriis, line 1: happened during format. Should be ok

hinthornw (approved):
Is it a bad look of we say "not recommended" on aclass from core?

efriis:
empty matrix issue on "CI Success"  for docs-only prs - debugging separately

PR #18034, opened 2024-02-23
openai[patch]: remove numpy dep

efriis:
Testing for equivalence

```
import random
import numpy as np

precision = 14

num = 4
dim = 105

_result = [[random.random() for _ in range(dim)] for _ in range(num)]
num_tokens_in_batch_i = [random.randint(1, 4000) for _ in range(num)]

def test_average(_result, num_tokens_in_batch_i):
    total_weight = sum(num_tokens_in_batch_i)
    average = [
        sum(
            val * weight
            for val, weight in zip(embedding, num_tokens_in_batch_i)
        )
        / total_weight
        for embedding in zip(*_result)
    ]
    assert average == list(np.average(_result, axis=0, weights=num_tokens_in_batch_i))
    magnitude = sum(val**2 for val in average) ** 0.5
    emb = [val / magnitude for val in average]
    assert [round(i, precision) for i in emb] == [round(i, precision) for i in list((average / np.linalg.norm(average)).tolist())]
    print("success")

test_average(_result, num_tokens_in_batch_i)
```


libs/partners/openai/pyproject.toml
```diff
@@ -57,6 +51,12 @@ optional = true
 [tool.poetry.group.dev.dependencies]
 langchain-core = { path = "../../core", develop = true }
 
+[tool.poetry.group.test_integration]
+optional = true
+
+[tool.poetry.group.test_integration.dependencies]
+numpy = "^1"
```
efriis, line 58: leave closeness checks in numpy in integration tests

hoodmane:
What about:
https://github.com/langchain-ai/langchain/blob/1b74e79f3fc8bbd8f89ba122bf097c64b63c0921/libs/partners/openai/langchain_openai/embeddings/base.py#L473-L474

efriis:
@hoodmane fixed ty!

PR #18041, opened 2024-02-23
test partner success

efriis:
- infra: CI success for partner packages
- always run CI Success
- format print
- x
- x


efriis:
yay

PR #18052, opened 2024-02-23
openai[patch]: refactor with_structured_output

baskaryan:
- make schema Optional with default val None, since in json_mode you don't need it if not parsing to pydantic
- change return_type -> include_raw
- expand docstring examples


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -754,32 +754,32 @@ def bind_tools(
     @overload
     def with_structured_output(
         self,
-        schema: _DictOrPydanticClass,
+        schema: Optional[_DictOrPydanticClass] = None,
         *,
         method: Literal["function_calling", "json_mode"] = "function_calling",
-        return_type: Literal["all"] = "all",
+        include_raw: Literal[True] = True,
```
hwchase17, line 760: is this just `: bool`

baskaryan: when overloading you have to narrow it so the overloads have mutually exclusive signatures

eyurtsev (approved):

PR #18055, opened 2024-02-23
rfc: with_structured_output List

baskaryan:
Support List[Schema] as input

Screenshot 2024-02-23 at 3 23 25 PM

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -63,7 +63,13 @@ from langchain_core.output_parsers import (
 )
 from langchain_core.output_parsers.base import OutputParserLike
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
-from langchain_core.pydantic_v1 import BaseModel, Field, SecretStr, root_validator
+from langchain_core.pydantic_v1 import (
+    BaseModel,
+    Field,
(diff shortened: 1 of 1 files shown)


baskaryan (commented):
afaict there's no good way to type hint a type hint, so may need to do `schema: Any` unless someones got a better idea

efriis (commented):
Would it be an antipattern to pass in `List[MyModel]` as a parameter instead of it being passed in as some sort of generic/type checking thing?

I think could semi-restrict `Any` with something like `Union[Type[List], Type[BaseModel]]` which would disallow people from passing in actual things, but would allow anything exported from `typing` basically (`List`, `Dict`, etc)

efriis:
I also wonder if this implementation is clever but more likely to trip people up than help them. Might be simpler / more understandable if the solution to structured output is "make a pydantic model with a list property at the top" in the docs.

baskaryan:
> I also wonder if this implementation is clever but more likely to trip people up than help them. Might be simpler / more understandable if the solution to structured output is "make a pydantic model with a list property at the top" in the docs.

difference between this and having folks create wrapper pydantic model (which is already doable) is this relies on parallel function calling to generate the list. ie what's actually being passed to the model is different. seems worth giving ppl that flexibility, in case nested schemas perform worse. also think this is a nicer devx

efriis:
Still worth pursuing or close?

PR #18070, opened 2024-02-24
core: implement a batch_size parameter for CacheBackedEmbeddings

chrispy-snps:
**Description:**

Currently, `CacheBackedEmbeddings` computes vectors for *all* uncached documents before updating the store. This pull request updates the embedding computation loop to compute embeddings in batches, updating the store after each batch.

I noticed this when I tried `CacheBackedEmbeddings` on our 30k document set and the cache directory hadn't appeared on disk after 30 minutes.

The motivation is to minimize compute/data loss when problems occur:

* If there is a transient embedding failure (e.g. a network outage at the embedding endpoint triggers an exception), at least the completed vectors are written to the store instead of being discarded.
* If there is an issue with the store (e.g. no write permissions), the condition is detected early without computing (and discarding!) all the vectors.

**Issue:**
Implements enhancement #18026.

**Testing:**
I was unable to run unit tests; details in [this post](https://github.com/langchain-ai/langchain/discussions/15019#discussioncomment-8576684).


libs/langchain/langchain/embeddings/cache.py
```diff
@@ -111,12 +114,14 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
...
             texts
         )
-        missing_indices: List[int] = [
+        all_missing_indices: List[int] = [
             i for i, vector in enumerate(vectors) if vector is None
         ]
-        missing_texts = [texts[i] for i in missing_indices]
 
-        if missing_texts:
+        while all_missing_indices:
```
eyurtsev, line 121: Could you re-use this implementation: https://github.com/langchain-ai/langchain/blob/main/libs/core/langchain_core/utils/iter.py#L168-L168

eyurtsev (commented):
Thanks for the contribution! We can add this, but pagination should be opt-in since this is not the most obvious place to paginate within the indexing pipeline and there's no natural batch size to use -- if the items to be embedded are very small more than 100 items could be easily supported with lighter models.

libs/langchain/tests/unit_tests/embeddings/test_caching.py
```diff
@@ -27,7 +27,7 @@ def cache_embeddings() -> CacheBackedEmbeddings:
     store = InMemoryStore()
     embeddings = MockEmbeddings()
     return CacheBackedEmbeddings.from_bytes_store(
-        embeddings, store, namespace="test_namespace"
+        embeddings, store, batch_size=50, namespace="test_namespace"
```
eyurtsev, line 30: Would you be able to add unit tests that cover the batching logic (e.g., batch size 3?)  -- lots of folks use CacheBackedEmbeddings so we want to make sure that nothing breaks before releasing

eyurtsev:
Ran linter, added missing import for Optional type, updated type annotation in core.

Should be good to merge once tests pass

eyurtsev (approved):

chrispy-snps:
@eyurtsev - thanks for polishing the rough edges!

PR #18071, opened 2024-02-24
docs: for community chat ollama - fixes for multimodal example in ollama.ipynb

joakimeriksson:
Description:
Changes the way to generate the human messages. I tried to use this to create a short python program (.py) but it failed to make use of the text due to the image always coming in after the text. I changed to create two human messages - the first sets the image, the second  will be the prompt / text.

Issue: the text prompt seems to be ignored if using the example of multimodal Ollama chat.


baskaryan:
do we know if this is caused by the langchain integration or if this is an issue with ollama? or the specific model being used?

joakimeriksson:
I can check if switching llava for bakllava will make any difference. I can also try to run some debugging on the REST API to capture ollama CLI calls and langchain calls. I can try to study what ollama's REST API do when getting text before image, vs image before text. But from what I can see is that langchain always put text before the image - independent of the order in the list that is put into the HumanMessage.

joakimeriksson:
I did a few tests quickly - and from what I can see. It is not easy to get any changes in the response by bundling image and text in the first message in ollama via curl. But with the approach above - it will give some results. So it is a working fix, but also a bit of a workaround. Ollama might be doing something "wrong".  So the conclusion is that it is not even working with a simplistic curl example from here:

https://github.com/ollama/ollama/blob/main/docs/api.md

I will try to figure out exactly why it works when giving the ollama cli an image before the text.

baskaryan:
> I did put an issue description on Ollama now also: [ollama/ollama#2769](https://github.com/ollama/ollama/issues/2769) Do you know if the other ollama non chat example works? (or also needs a workaround to actually work)?

sounds like there was some fix on their end https://github.com/ollama/ollama/pull/2789, does that affect this pr?

joakimeriksson:
With The latest Ollama fix this should work again. Then there is no need to update - the release will be out very soon - it is already in pre-release and I have tested my other similar Ollama examples and it is a good fix!

baskaryan:
closing per ^, let me know if i'm missing anything!

PR #18077, opened 2024-02-24
langchain: enhance `LocalFileStore` to offer a `umask` parameter for shared `CacheBackedEmbeddings` caches

chrispy-snps:
**Description:**
The `LocalFileStore` class can be used to create an on-disk `CacheBackedEmbeddings` cache. However, the default `umask` settings gives file/directory write permissions only to the original user. Once the cache directory is created, other users cannot write their own cache entries into the directory.

To make the cache usable by multiple users, this pull request updates the `LocalFileStore` constructor to accept a `umask` parameter that is used for file/directory creation. The `umask` value is applied to both file and directory creation.

For example, when configured with a `umask` of `0o007` (*user* and *group* permitted, *others* masked),

* Anyone in our group can contribute embeddings to the cache.
* If we implement cache cleanup/eviction in the future, anyone in our group can perform the cleanup.

The default is `umask=None`, which retains the original behavior of using the default `umask` settings.

The reason this is implemented in the `LocalFileStore` class is that embeddings can be computed in many different places in the user's code. And if caching is extended to single queries, then even more locations apply. Manual `umask` overrides would require every one of these locations to be wrapped with something like this:

```
orig_umask = os.umask(0o007)
try:
    faiss_vectorstore = FAISS.from_documents(...)
finally:
    os.umask(orig_umask)
```

From the user's perspective, the cache `umask` is a property of the *cache*, not a characteristic of the current execution environment. By handling this in `LocalFileStore`, cache updates always use the desired `umask` by construction.

**Issue:**
Implements enhancement #18075.

**Testing:**
I was unable to run unit tests; details in [this post](https://github.com/langchain-ai/langchain/discussions/15019#discussioncomment-8576684).


libs/langchain/langchain/storage/file_system.py
```diff
@@ -95,10 +101,15 @@ def mset(self, key_value_pairs: Sequence[Tuple[str, bytes]]) -> None:
         Returns:
             None
         """
-        for key, value in key_value_pairs:
-            full_path = self._get_full_path(key)
-            full_path.parent.mkdir(parents=True, exist_ok=True)
-            full_path.write_bytes(value)
+        orig_umask = os.umask(0)
```
eyurtsev, line 104: This sets the umask globally as far as I understand which will create a dangerous race condition with the rest of the code base and a security vulnurability for any file write anywhere.

File storage cache is meant mostly for prototyping purposes and should be getting getting extra functionality like this (even if it were done without a race condition)

eyurtsev (changes requested):
Introduces a security vulnerability due to a race condition

eyurtsev:
I don't think we want to accept enhancements to the LocalFileStore cache -- it's mostly there for prototyping purposes

chrispy-snps:
@eyurtsev - thanks for pointing out the vulnerability with the global `umask` approach.

I understand that `LocalFileStore` is primarily a proof-of-concept, but it's 100% workable and sufficient in its own right. I can re-implement this feature using `chmod`. If you're open to that, let me know and I'd be happy to do it.

eyurtsev:
@chrispy-snps if it can be implemented without mutating any global state or increasing the complexity of understanding this implementation that's probably fine. The LocalFileStore is mostly meant as a reference implementation not actual production implementation.

Could you explain your use case a bit? Why use the local file system vs. say s3? Are you sharing the cache between multiple users in your organization?

chrispy-snps:
Superseded by #18857.

Issue #18080, opened 2024-02-25
DOC: Possible mistake in property's docstring of ConversationBufferWindowMemory


lgabs:
### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

Readiing `ConversationBufferWindowMemory` class, it seems that the property `buffer` returns `self.buffer_as_messages` if `self.return_messages` is `True`, otherwise it returns `self.buffer_as_str`.

https://github.com/langchain-ai/langchain/blob/7fc903464a753ac10e9b671906c2e9889e4d598e/libs/langchain/langchain/memory/buffer_window.py#L17-L21

However, the docstrings of these properties `buffer_as_str` and `return_messages` seem inverted when they say `True` and `False` respectively:

https://github.com/langchain-ai/langchain/blob/7fc903464a753ac10e9b671906c2e9889e4d598e/libs/langchain/langchain/memory/buffer_window.py#L22-L30

https://github.com/langchain-ai/langchain/blob/7fc903464a753ac10e9b671906c2e9889e4d598e/libs/langchain/langchain/memory/buffer_window.py#L32-L35

keenborder786:
@lgabs nice catch ,created a PR for the fix.

lgabs:
@keenborder786 am I mistaken or there is still a change to make in `buffer_as_messages`?

https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/memory/buffer_window.py#L22-L36

I've updated this in this new PR: https://github.com/langchain-ai/langchain/pull/18136

PR #18090, opened 2024-02-25
langchain[patch]: Update doc-string for a method in ConversationBufferWindowMemory

keenborder786:
A minor doc fix stated in #18080


lgabs:
@keenborder786 I think you forgot to also change the `buffer_as_messages` docstring from `False` to `True`.

eyurtsev (approved):

PR #18101, opened 2024-02-26
add run name for query constructor

hwchase17:


libs/langchain/langchain/retrievers/self_query/base.py
```diff
@@ -208,6 +208,7 @@ def from_llm(
         chain_kwargs: Optional[Dict] = None,
         enable_limit: bool = False,
         use_original_query: bool = False,
+        query_constructor_run_name: str = "query_constructor",
```
baskaryan, line 211: imo doesn't need to be an arg, can just hard code

libs/langchain/langchain/retrievers/self_query/base.py
```diff
@@ -208,6 +208,7 @@ def from_llm(
         chain_kwargs: Optional[Dict] = None,
         enable_limit: bool = False,
         use_original_query: bool = False,
```
eyurtsev, line 210: All new named arguments should be introduced with `*` present, to minimize accidental breaking changes
```suggestion
        use_original_query: bool = False,
        *,
```

libs/langchain/langchain/retrievers/self_query/base.py
eyurtsev: If we're using an arg, perhaps we could use a standard arg name argument? (e.g., `run_name`) and it'll be understood that it refers to the top level object

hwchase17: what if there's multiple different runnables in a constructor?

ill just remove as an arg for now and avoid this question :)

PR #18108, opened 2024-02-26
community: DirectoryLoader uses `wcmatch` for glob patterns

evisong:
- **Description:** `langchain_community.document_loaders.DirectoryLoader` accepts an optional `glob` argument, which can be used to match multiple files by glob patterns; however the underlying Python built-in `pathlib.Path.glob` and `pathlib.Path.rglob` doesn't support either extended glob syntax or multiple glob patterns. Switching to `wcmatch` adds the following features for `DirectoryLoader`'s glob matching:
  - Basic glob syntax, e.g.: `**/*.md`
  - Extended glob syntax, e.g.: `docs/**/*.@(md|html)`
  - List of glob patterns, e.g.: `['docs/**/*.md', 'webpages/**/*.html']`
  - More syntax please see: https://facelessuser.github.io/wcmatch/glob/#syntax 
- **Issue:** fixes #3062
- **Dependencies:** `wcmatch`
- **Twitter handle:** song1w


hwchase17:
thanks for opening this, but we'd prefer to keep the dependencies lightweight and not introduce a new package just for this - at least at this moment. that may change in the future

evisong:
@hwchase17, thanks for your comments. 

Btw, do you think it is worth it to allow `DirectoryLoader.glob` to accept a list of glob patterns? `glob=["*.txt", "*.md"]` could help my use cases a lot.

Anyway, I can use `docs = loader1.load() + loader2.load() + ...` as a workaround for now.

PR #18116, opened 2024-02-26
community[patch]: added latin-1 decoder to gmail search tool

Sanjaypranav:
some mails from flipkart , amazon are encoded with other plain text format so to handle UnicodeDecode error ,  added exception and latin decoder

@hwchase17


eyurtsev (approved):

libs/community/langchain_community/tools/gmail/search.py
```diff
@@ -98,7 +98,10 @@ def _parse_messages(self, messages: List[Dict[str, Any]]) -> List[Dict[str, Any]
                     ctype = part.get_content_type()
                     cdispo = str(part.get("Content-Disposition"))
                     if ctype == "text/plain" and "attachment" not in cdispo:
-                        message_body = part.get_payload(decode=True).decode("utf-8")
+                        try:
+                            message_body = part.get_payload(decode=True).decode("utf-8")
+                        except UnicodeDecodeError:
+                            message_body = part.get_payload(decode=True).decode("latin-1") # retrives mail from flipkart
```
eyurtsev, line 104: This logic is hard-coded. Ideally there would be more generalization.

Approving this because it seems that a refactor without breaking changes will be possible in the future by declaring attributes:


```python
encoding_to_try: Sequence[str] = ('utf-8', 'latin-1') # Will try these encodings if detect_encoding if False
detect_encoding: bool = False # if True will automatically detect the encoding
```

This is a bit hard-coded, but can a refactor can be made without breaking changes in the future by declaring `encodings_to_try` or something of that sort as an argument to the tool, and an additional argument could be added

eyurtsev (approved):

eyurtsev:
@Sanjaypranav could you resolve lint conflicts and then we'll merge?

Sanjaypranav:
yeah sure

Issue #16791, opened 2024-01-30
ValueError: One output key expected, got dict_keys(['output', 'intermediate_steps']) when using create_structured_chat_agent with chat_memory and intermediate steps


glejdis:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

The following code:

```from langchain.memory import ConversationBufferMemory


llm = AzureChatOpenAI(
    azure_endpoint=AZURE_TEXT_ENDPOINT,
    openai_api_version=OPEN_API_VERSION,
    deployment_name=AZURE_TEXT_DEPLOYMENT, #"gpt-4_32k",
    openai_api_key=OPENAI_TEXT_API_KEY,
    openai_api_type=OPENAI_API_TYPE, #"azure",
    temperature=0
)

ai_search_endpoint = get_ai_search_endpoint()
ai_search_admin_key = get_ai_search_admin_key()

vector_store = AzureSearch(
        azure_search_endpoint=ai_search_endpoint,
        azure_search_key=ai_search_admin_key,
        index_name=index_name,
        embedding_function=embeddings.embed_query,
        content_key="xxx"
    )
    
"""Retriever that uses `Azure Cognitive Search`."""
    
azure_search_retriever = AzureSearchVectorStoreRetriever(
        vectorstore=vector_store,
        search_type="hybrid",
        k=3,
    )

retriever_tool = create_retriever_tool(
    azure_search_retriever,
    "Retriever",
    "Useful when you need to retrieve information from documents", 
)


prompt = ChatPromptTemplate.from_messages(
    [
[...]

PR #18117, opened 2024-02-26
langchain: fix BaseChatMemory get output data error with extra key

xiaohuanshu:
**Description:** At times, BaseChatMemory._get_input_output may acquire some extra keys such as 'intermediate_steps' (agent_executor with return_intermediate_steps set to True) and 'messages' (agent_executor.iter with memory). In these instances, _get_input_output can raise an error due to the presence of multiple keys. The 'output' field should be used as the default field in these cases.
**Issue:** #16791

### DIFF
--- a/libs/langchain/langchain/memory/chat_memory.py
+++ b/libs/langchain/langchain/memory/chat_memory.py
@@ -26,9 +26,12 @@ class BaseChatMemory(BaseMemory, ABC):
         else:
             prompt_input_key = self.input_key
         if self.output_key is None:
-            if len(outputs) != 1:
-                raise ValueError(f"One output key expected, got {outputs.keys()}")
-            output_key = list(outputs.keys())[0]
+            if len(outputs) == 1:
+                output_key = list(outputs.keys())[0]
+            elif "output" in outputs:
+                output_key = "output"
+            else:
+                raise ValueError("Could not determine output key.")
         else:
             output_key = self.output_key
         return inputs[prompt_input_key], outputs[output_key]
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/memory/chat_memory.py
```diff
@@ -26,9 +26,12 @@ def _get_input_output(
         else:
             prompt_input_key = self.input_key
         if self.output_key is None:
-            if len(outputs) != 1:
-                raise ValueError(f"One output key expected, got {outputs.keys()}")
-            output_key = list(outputs.keys())[0]
+            if len(outputs) == 1:
+                output_key = list(outputs.keys())[0]
+            elif "output" in outputs:
+                output_key = "output"
```
hwchase17, line 32: can we add some logging here? I don't think this is expected, and id prefer if people set it manually. so im fine adding this, but id want good logging so they know whats going on

libs/langchain/langchain/memory/chat_memory.py
```diff
@@ -26,9 +26,12 @@ def _get_input_output(
...
+            else:
+                raise ValueError("Could not determine output key.")
```
hwchase17, line 34: can we raise a better error message? should contain info about the keys found and why we cant determine the output key

libs/langchain/langchain/memory/chat_memory.py
xiaohuanshu: how about adding a warning message in line 32:
`warnings.warn(f"'{self.__class__.__name__}' got multiple output keys: {outputs.keys()}. The default 'output' key is being used. If this is not desired, please manually set 'output_key'.")`

libs/langchain/langchain/memory/chat_memory.py
xiaohuanshu: how about changing it to `ValueError(f"Got multiple output keys: {outputs.keys()}, cannot determine which to store in memory. Please set the 'output_key' explicitly.")`

baskaryan (approved):

PR #18122, opened 2024-02-26
Add an option for indexed generic label when import neo4j graph documents

tomasonjo:
Current implementation doesn't have an indexed property that would optimize the import. I have added a `baseEntityLabel` parameter that allows you to add a secondary node label, which has an indexed id `property`.  By default, the behaviour is identical to previous version.

Since multi-labeled nodes are terrible for text2cypher, I removed the secondary label from schema representation object and string, which is used in text2cypher.


jexp (approved):
LGTM

baskaryan (approved):

PR #17767, opened 2024-02-19
partner: Astra DB, add indexing support for Vector Store class

hemidactylus:
This PR adds support to the "Indexing" options when creating Astra DB collection for fine-grained control over which fields are indexed to be later used as search filters.

The (sensible) default is to index all and only the contents of the "metadata" map (plus the vectors themselves for the ANN search of course).
But one can choose to exclude certain metadata fields from being indexed (e.g. very long unique strings), or conversely specify an allowlist - or even provide a fully custom indexing prescription fed directly to the API.

In case the collection is detected on DB already, an error is raised if the requested indexing options are incompatible. But if a legacy collection (without indexing options altogether) is detected, a warning is raised and the execution can proceed.

Added unit + integration tests about this behaviour.

PR #18131, opened 2024-02-26
partner: Astra DB clients identify themselves as coming through LangChain package

hemidactylus:
**Description**

This PR sets the "caller identity" of the Astra DB clients used by the integration plugins (`AstraDBChatMessageHistory`, `AstraDBStore`, `AstraDBByteStore` and, pending #17767 , `AstraDBVectorStore`). In this way, the requests to the Astra DB Data API coming from within LangChain are identified as such (the purpose is anonymous usage stats to best improve the Astra DB service).

### DIFF
--- a/libs/partners/astradb/langchain_astradb/utils/astradb.py
+++ b/libs/partners/astradb/langchain_astradb/utils/astradb.py
@@ -6,6 +6,7 @@ from asyncio import InvalidStateError, Task
 from enum import Enum
 from typing import Awaitable, Optional, Union
 
+import langchain_core
 from astrapy.db import AstraDB, AsyncAstraDB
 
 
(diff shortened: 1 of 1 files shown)


libs/partners/astradb/langchain_astradb/utils/astradb.py
```diff
@@ -51,20 +52,29 @@ def __init__(
...
             self.astra_db = self.async_astra_db.to_sync()
         else:
             raise ValueError(
                 "Must provide 'astra_db_client' or 'async_astra_db_client' or "
                 "'token' and 'api_endpoint'"
             )
 
+        self.astra_db.set_caller(
+            caller_name="langchain",
+            caller_version=getattr(langchain_core, "__version__", None),
```
eyurtsev, line 71: You should get the version of `langchain_astradb`. Can also get langchain_core if helpful for debugging, but I think the partner version is much more important

hemidactylus: Right, thank you for the suggestion. But we'd prefer to keep alignment between the caller_name being "langchain", as in the core Lc machinery, and the corresponding version.
In other words, we might eventually extend this to a list of (package,version) pairs - in which case we'd neatly add `("langchain_astradb", <corresponding version>)` to the stack.

So if that's ok we'd like to start by leaving it as is.

hemidactylus: @eyurtsev So I would say this is good to go as far as this is concerned :)

eyurtsev: Not the caller name, but the caller version. I don't think you care about the version of langchain core but of langchain_astradb? 

I'll merge but consider changing the version that you're using

eyurtsev (approved):

PR #18143, opened 2024-02-26
template: add rag azure search template

kristapratico:
- **Description:** Adds a template for performing RAG with the AzureSearch vectorstore.
- **Issue:** N/A
- **Dependencies:** N/A
- **Twitter handle:** N/A


templates/rag-azure-search/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 LangChain, Inc.
```
baskaryan, line 3: think this should be 2024 (@efriis think we need to update cli template)

baskaryan (commented):
would be nice to add a simple top level notebook showing what client side code would look like, like in rag-pinecone for example

templates/rag-azure-search/LICENSE
efriis:
```suggestion
Copyright (c) 2024 LangChain, Inc.
```

efriis (approved):

PR #18145, opened 2024-02-26
Warn against implicit generator coercion

hinthornw:
TY to @nfcampos for the suggestion


nfcampos:
@hinthornw do we need to update docs too?

hinthornw:
ya

PR #18146, opened 2024-02-26
`indexes` refactoring

leo-gan:
`indexes` atifacts partially duplicted in the `langchain` and in `langchain_community` packages.
Regarding #17832 and following @baskaryan recommendations, `indexes` is moved into the `langchain_community`.
All ut-s in `langchain` work without changes.
Missed ut-s from `langchain` were added into the `langchain_community` ut-s. Only ut-s that test the "__all__" symbol were not added, because "__all__` in `langchain_community` is not used.


libs/langchain/langchain/indexes/_api.py
```diff
@@ -1,598 +1,15 @@
...
-    """Number of added documents."""
-    num_updated: int
-    """Number of updated documents because they were not up to date."""
-    num_deleted: int
-    """Number of deleted documents."""
-    num_skipped: int
-    """Number of skipped documents because they were already up to date."""
-
-
-def index(
```
eyurtsev, line 189: `index` should stay in langchain -- since partner packages may introduce a record manager implementation and we don't want partner packages to ever depend on langchain_community

libs/langchain/langchain/indexes/base.py
```diff
@@ -1,172 +1,5 @@
...
+from langchain_community.indexes.base import RecordManager
 
-import uuid
-from abc import ABC, abstractmethod
-from typing import List, Optional, Sequence
-
-NAMESPACE_UUID = uuid.UUID(int=1984)
-
-
-class RecordManager(ABC):
```
eyurtsev, line 10: This should stay in langchain or end up in langchain core since partner packages will need to inherit for their own   implementations.

Generally abstract interfaces should be in either langchain core or langchain, but not in langchain community

leo-gan: @eyurtsev  Yeah, exactly, that was implemented in the #17832 but @baskaryan [opposed](https://github.com/langchain-ai/langchain/pull/17832#pullrequestreview-1894644198) it. So...
looks like you and @baskaryan should decide what to do or not to do. I'm pausing now with this PR.

leo-gan:
Close temporarily till the dependency rules are clear.

eyurtsev:
Created this from referencing the refactor https://github.com/langchain-ai/langchain/pull/20667

PR #18151, opened 2024-02-26
core[patch]: support JS message serial namespaces

baskaryan:


jacoblee93 (approved):
I think good! Will follow up with the rest of the list.

Might want a test just to be safe?

PR #18153, opened 2024-02-26
docs: hide `More/Changelog` menu item

leo-gan:
The `Changelog` page is not supported. It is confusing to show this page to users.


leo-gan:
Close temporarily till the dependency rules are clear.

eyurtsev:
image

eyurtsev:
We should probably fix the change log page instead -- it already has content right now and hiding it will break existing links.

Link breaking is something we want to avoid as much as possible

eyurtsev:
I suspect that we want to create an issue for this and get CI to start populating the page cc @efriis

leo-gan:
@eyurtsev Not sure we need it here. All this information is free on [GitHub](https://github.com/langchain-ai/langchain/pulse).
Anyway, if we create it it should be something with different goals than the GitHub info.

eyurtsev:
Some GitHub features don't work well for mono repo architecture. I'll let others weigh in -- my suspicion is that a dedicated page will produce a better UI than what GitHub can do right now

leo-gan:
@eyurtsev I agree. 
Returning back to the problem of hiding the outdated page. The **manual** generation is not feasible for the Changelog, 100%. If we don't have automatic generation, we'd rather hide it. It is confusing.

leo-gan:
stalled

leo-gan:
@hwchase17  The Changelog page is not supported. We don't need `stalled` items in documentation.

eyurtsev:
I think this was addressed during Jacob's refactor

leo-gan:
@eyurtsev Thanks!

PR #18160, opened 2024-02-27
langchain[patch]: getattr import from langchain.chains

efriis:


libs/langchain/langchain/chains/__init__.py
```diff
@@ -17,135 +17,153 @@
...
+    {"name": "create_sql_query_chain", "module": "langchain.chains.sql_database.query"},
+    {
+        "name": "create_structured_output_runnable",
+        "module": "langchain.chains.structured_output",
+    },
+    {"name": "load_summarize_chain", "module": "langchain.chains.summarize"},
+    {"name": "TransformChain", "module": "langchain.chains.transform"},
 ]
+
+_module_lookup = {entry["name"]: entry["module"] for entry in _exports}
```
baskaryan, line 159: why not just make _exports be a dict

efriis: Oh this was from `llms` where we also had an `llm_type` key to store, so had two different things I had to iterate over

Will reformat here though

PR #18218, opened 2024-02-27
community: Added New Tool: Google Translator (googletrans)

shhossain:
# New Tool: Google Translator (googletrans)

  - **Description:** Added a new tool, Google Translator using googletrans package
  - **Dependencies:** googletrans
  - **Twitter handle:** @shhossain77

- test file: langchain\libs\community\tests\integration_tests\utilities\test_google_translate.py
- docs file: langchain\docs\docs\integrations\tools\google_translate.ipynb


leo-gan:
How it is different from the existing [tool](https://python.langchain.com/docs/integrations/platforms/google#google-translate)?
Maybe it make sense just update the existing tool with ideas from the current branch?

shhossain:
> How it is different from the existing [tool](https://python.langchain.com/docs/integrations/platforms/google#google-translate)? Maybe it make sense just update the existing tool with ideas from the current branch?

Okay, I've noticed that there are translators available in LangChain (which I should've used in the first place, and for that, I am sincerely sorry). This includes [Doctran](https://python.langchain.com/docs/integrations/document_transformers/doctran_translate_document) and [Google Translate](https://python.langchain.com/docs/integrations/document_transformers/google_translate). However, both of them require payment.

Regarding the `GoogleTranslateTransformer`, you can technically modify it to work with `googletrans`, but I don't think it would be best practice since the entire `GoogleTranslateTransformer` class is based on Google Cloud Translation. Additionally, `GoogleTranslateTransformer` is a document transformer, but I want to use it as a tool, as Language Models (LLMs) are not very reliable in that aspect.

Lord-Haji:
`googletrans` is an unofficial google translate tickets which reverse engineers to generate tickets to get around google's timeout mechanism. Which could also be blocked anytime by Google.

> I eventually figure out a way to generate a ticket by reverse engineering on the [obfuscated and minified code used by Google to generate such token](https://translate.google.com/translate/releases/twsfe_w_20170306_RC00/r/js/desktop_module_main.js), and implemented on the top of Python. However, this could be blocked at any time.

I don't think it makes sense for langchain to support it ethically

shhossain:
> `googletrans` is an unofficial google translate tickets which reverse engineers to generate tickets to get around google's timeout mechanism. Which could also be blocked anytime by Google.
> 
> > I eventually figure out a way to generate a ticket by reverse engineering on the [obfuscated and minified code used by Google to generate such token](https://translate.google.com/translate/releases/twsfe_w_20170306_RC00/r/js/desktop_module_main.js), and implemented on the top of Python. However, this could be blocked at any time.
> 
> Quoted from [py-googletrans - README.MD](https://github.com/ssut/py-googletrans?tab=readme-ov-file#how-does-this-library-work)
> 
> I don't think it makes sense for langchain to support it ethically

I shared the same concerns, and to address them, I extensively researched existing tools in Langchain that utilize unofficial APIs. Notably, there are tools like the 'duckduckgo search tool' and 'wikipedia tool.' While both involve searching and retrieving content, it's important to note that 'googletrans' generates content, which introduces a potential issue.

In defense of 'googletrans,' it is worth acknowledging that it is unstable and unofficial. Additionally, there are daily limits(around 1k requests) and content restrictions (approximately 5k). To mitigate any potential risks, the tool's documentation can explicitly state its unofficial and unstable nature, recommending its use for demonstration purposes rather than in a production environment.

If this PR doesn't align with the values of Langchain Community, feel free to close it.

leo-gan:
CC @hwchase17 It seems your feedback is required.

PR #18219, opened 2024-02-27
core: Adding kwargs to cache to be able to add metadat

romainrey:
- **Description:**
Adding the possibility to pass kwargs to the cache functions. 
[Discussed here](https://github.com/langchain-ai/langchain/discussions/18215)

### DIFF
--- a/libs/core/langchain_core/caches.py
+++ b/libs/core/langchain_core/caches.py
@@ -34,23 +34,23 @@ class BaseCache(ABC):
     """Base interface for cache."""
 
     @abstractmethod
-    def lookup(self, prompt: str, llm_string: str, **kwargs: Any) -> Optional[RETURN_VAL_TYPE]:
+    def lookup(self, prompt: str, llm_string: str, cache_params: Any) -> Optional[RETURN_VAL_TYPE]:
         """Look up based on prompt and llm_string."""
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


hwchase17 (commented):
whats a use case?

hwchase17:
nvr mind i see the discussion. thoughts @eyurtsev ?

romainrey:
I think we might need to adjust the pr slightly if that is something you guys want to push.
Happy to do it if you are ok on the idea/

libs/core/langchain_core/caches.py
```diff
@@ -34,23 +34,23 @@ class BaseCache(ABC):
     """Base interface for cache."""
 
     @abstractmethod
-    def lookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
+    def lookup(self, prompt: str, llm_string: str, cache_params: Any) -> Optional[RETURN_VAL_TYPE]:
```
eyurtsev, line 37: A few questions:

* What is the type of cache_params? Do you imagine a flat dict would be sufficient?
* What are the semantics for cache_params when doing look up?
* What are the semantics when doing an update?

eyurtsev (commented):
The use case makes sense. Would love to figure out the cleanest way to accommodate. Left a question on the PR -- could you tell me what you imagine for the type and the semantics for implementors of cache?

eyurtsev:
Closing as stale

PR #18237, opened 2024-02-28
docs:Nvidia Riva Runnables Documentation

haydeniw:
- **Description:** Documents how to use the Riva runnables to add streamed automatic-speech-recognition (ASR) and text-to-speech (TTS) to chains.  
  - **Issue:** None
  - **Dependencies:** None
  - **Twitter handle:** @HaydenWolff1


hwchase17 (commented):
not sure we want to add a new documentation section for utilities. usually utilities are only for things that are used by tools, so they show up in tools

haydeniw:
@hwchase17 so it sounds like it should go in tools documentation?

ccurme (commented):
Thanks for this! I think tools is a reasonable place for this (although it's not strictly a tool, all the other text-to-speech implementations are documented in tools).

https://python.langchain.com/docs/integrations/tools

I would recommend re-titling the notebook to "NVIDIA Riva" or similar, as the current title is long and may show up as such in the sidebar of the docs.

baskaryan (approved):

baskaryan (approved):

PR #18250, opened 2024-02-28
community[patch]: chat message histrory mypy fix

Lord-Haji:
Description: Fixed type: ignore's for mypy for chat_message_histories(streamlit)
Adresses #17048 

Planning to add more based on reviews

### DIFF
--- a/libs/community/langchain_community/chat_message_histories/streamlit.py
+++ b/libs/community/langchain_community/chat_message_histories/streamlit.py
@@ -22,17 +22,26 @@ class StreamlitChatMessageHistory(BaseChatMessageHistory):
 
         if key not in st.session_state:
             st.session_state[key] = []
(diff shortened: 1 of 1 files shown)


Lord-Haji:
I was thinking what if I made an helper function like
``
def import_streamlit():
    """Attempt to import Streamlit, raising a clear error if the import fails."""
    try:
        import streamlit as st
    except ImportError as e:
        raise ImportError("Streamlit is required for this feature. Please install Streamlit using `pip install streamlit`.") from e
    return st
``
    
    and reuse it for the constructor and messages setter? Wouldn't that make more sense?
    Thoughts @baskaryan @eyurtsev

baskaryan (approved):

baskaryan:
cc @sfc-gh-jcarroll

PR #18253, opened 2024-02-28
community[minor]: added a feature to filter documents in Mongoloader

iamKushagra:
"community: added a feature to filter documents in Mongoloader"
    - **Description:** added a feature to filter documents in Mongoloader
    - **Feature:** the feature #18251
    - **Dependencies:** No
    - **Twitter handle:** https://twitter.com/im_Kushagra


iamKushagra:
@baskaryan

libs/community/langchain_community/document_loaders/mongodb.py
```diff
@@ -17,6 +17,7 @@ def __init__(
         connection_string: str,
         db_name: str,
         collection_name: str,
+        field_names: Optional[List[str]] = None,
```
eyurtsev, line 20: Given that this is a named argument could you put it after the `*` and change the type to Optional[Sequence[str]]

libs/community/langchain_community/document_loaders/mongodb.py
```diff
@@ -66,7 +68,22 @@ async def aload(self) -> List[Document]:
...
                 "collection": self.collection_name,
             }
-            result.append(Document(page_content=str(doc), metadata=metadata))
+            if self.field_names is not None:
+                """Filter fields based on field_names"""
+                try:
+                    fields = {name: doc[name] for name in self.field_names}
+                except KeyError as err:
+                    logger.warning(f"{err.args[0]} field not found in Mongo document.")
+                    continue  """Skip this document if a specified field is not found"""
```
eyurtsev, line 77: `"""` is used for doc-strings for functions / classes, not for in code documentation

```suggestion
                    continue  # Skip this document if a specified field is not found
```

seme for changes below

libs/community/langchain_community/document_loaders/mongodb.py
```diff
@@ -66,7 +68,22 @@ async def aload(self) -> List[Document]:
                 "database": self.db_name,
                 "collection": self.collection_name,
             }
-            result.append(Document(page_content=str(doc), metadata=metadata))
+            if self.field_names is not None:
```
eyurtsev, line 71: Is there a way to provide the field names as part of the request to mongo instead? So we don't end up sending those fields over the wire if we don't need them.

libs/community/langchain_community/document_loaders/mongodb.py
```diff
@@ -66,7 +68,22 @@ async def aload(self) -> List[Document]:
...
+                """Filter fields based on field_names"""
+                try:
+                    fields = {name: doc[name] for name in self.field_names}
+                except KeyError as err:
+                    logger.warning(f"{err.args[0]} field not found in Mongo document.")
```
eyurtsev, line 76: How frequently are fields expected to be missing?

iamKushagra: There could be instances when in a few documents there might that field doesn't exist. So, the frequency is still low but in any such case, we can skip those.

libs/community/langchain_community/document_loaders/mongodb.py
iamKushagra: Good point!

libs/community/langchain_community/document_loaders/mongodb.py
iamKushagra: Fixed

libs/community/langchain_community/document_loaders/mongodb.py
iamKushagra: Fixed

eyurtsev:
This looks good to me -- if you're able to resolve linting error (`make format`) we can merge

iamKushagra:
Thanks @eyurtsev 🚀

baskaryan (approved):

PR #18255, opened 2024-02-28
community[patch]: Fixing embedchain document mapping

kellerkind84:
- **Description:** The current embedchain implementation seems to handle document metadata differently than done in the current implementation of langchain and a KeyError is thrown. I would love for someone else to test this...


libs/community/langchain_community/retrievers/embedchain.py
```diff
@@ -65,7 +65,7 @@ def _get_relevant_documents(
             docs.append(
                 Document(
                     page_content=r["context"],
-                    metadata={"source": r["source"], "document_id": r["document_id"]},
+                    metadata={"score": r["metadata"]["score"], "document_id": r["metadata"]["doc_id"]},
```
deshraj, line 68:
```suggestion
                    metadata={"source": r["metadata"]["url"], "document_id": r["metadata"]["doc_id"]},
```

deshraj: Thanks for fixing this @kellerkind84 ❤️

PR #18270, opened 2024-02-28
partners: AI21 Labs Contextual Answers support

Josephasafg:
Description: Added support for AI21 Labs model - Contextual Answers
Dependencies: ai21, ai21-tokenizer


efriis (commented):
looks great! I'll change some nits in the code, and can you add documentation to the bottom of the `llms/ai21.ipynb` documentation page?

Understand it's not strictly an LLM, and would be great to get some description there!

libs/partners/ai21/pyproject.toml
```diff
@@ -1,26 +1,26 @@
...
+version = "0.1.0"
 description = "An integration package connecting AI21 and LangChain"
 authors = []
 readme = "README.md"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
 langchain-core = "^0.1.22"
-ai21 = "^2.0.0"
+ai21 = "^2.0.5"
```
efriis, line 11: @asafgardin I changed this from `2.0.5` to semver `^2.0.5` which now uses `2.1.2` - were there breaking changes between the two? Lint seems to be saying so

asafgardin: @efriis Yes. We added better handling for default values that are not passed using `|` so now the lint is throwing an error about it whenever we try to pass `None`. We have a task to better handle these types of values so mypy wouldn't raise an error

asafgardin:
@efriis Added description to `llm/ai21.ipynb`

libs/partners/ai21/pyproject.toml
efriis: Got it. So should I just lock this version to 2.0.5 specifically? Currently causing failures in CI

efriis: Or do you want to add some logic that programmatically passes the param vs. not, or type it as Any? Happy to follow your lead - you know your sdk better than me and what will work vs. not work :)

asafgardin: @efriis I'll fix the lint issues next week if that's okay by you. I want to merge this first using version 2.0.5 the way it is right now if everything works in the CI. 

What do you think?

efriis (approved):

PR #18316, opened 2024-02-29
langchain:Update the internal chat messages history

codemee:
- **Description:** prune() just remove the messages in memory, but not the messages stored in the internal chat messages history object. This lead to the problem that load_memory_variables() return a dictionary with all the messages in the history item.
- **Issue:** add code to remove the old messages exceed token limit.
- **Dependencies:** none.
- **Twitter handle:** @meebox


codemee:
The ConversationTokenBufferMemory class has the same problem.

hwchase17 (commented):
hmm i dont love removing everything and then adding stuff back. seems potentially wasteful. is there a way to just a .delete method to the chat message class or something?

codemee:
There's no any method for the BaseChatMessageHistory instance to update the messages except for clear and add_messages. Maybe adding a delete method could do the work.

PR #18331, opened 2024-02-29
Docs: Add custom parsing documentation and extending langchain

eyurtsev:
* Added extending langchain.mdx -- we'll need to add links as we add more custom documentation
* Added partial documentation about parsers


docs/docs/modules/model_io/output_parsers/custom.ipynb
hinthornw, line 1: We are planning to deprecate automatic coersion to RunnableGenerator FYI so would prefer to explicitly do `llm | RunnableGenerator(output_parser)` in the doc (i think)

cc @nfcampos

eyurtsev: We need to make it easier so that folks don't have to do `RunnableGenerator`. Could we extend the `chain` decorator to take arguments perhaps?

```python
@as_runnable(generator=True)
def my_streaming_parser...
```
something of that sort?

PR #18346, opened 2024-02-29
text-splitters[minor], langchain[minor], community[patch], templates, docs: langchain-text-splitters 0.0.1

baskaryan:


libs/text-splitters/pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+  "E",  # pycodestyle
+  "F",  # pyflakes
+  "I",  # isort
+  "T201", # print
+]
+
+[tool.mypy]
+disallow_untyped_defs = "True"
+
+[[tool.mypy.overrides]]
```
baskaryan, line 73: @efriis pretty nifty, lets you skip import errors for specific packages that don't have stubs

libs/partners/airbyte/langchain_airbyte/document_loaders.py
efriis, line 1: update lint dep to be langchain-text-splitters instead of langchain

efriis (commented):
Guessing you're working on these already but probs need to sort out some of the re-exports and dependencies

efriis:
Oh also should we 0.1.0 for poetry version bump benefits like partner packages?

Issue #16552, opened 2024-01-25
Devcontainer hangs on Poetry dependency resolution


cvansteenburg:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

To reproduce, simply open a dev container [here](https://github.com/langchain-ai/langchain/tree/master/.devcontainer#vs-code-dev-containers).

### Two problems:
1. Dependency azure-ai-vision [has been yanked ](https://pypi.org/project/azure-ai-vision/) from Pypi and replaced by [another MS package](replaced by https://pypi.org/project/azure-ai-vision-imageanalysis/1.0.0b1/). This change is not reflected in `pyproject.toml`.
2. Upon updating the package reference, Poetry dependency resolution takes hours before it ultimately crashes.

### The first problem is here:
```
#13 [langchain langchain-dev-dependencies 5/5] RUN poetry install --no-interaction --no-ansi --with dev,test,docs
2024-01-25 02:25:15.795Z: #13 1.180 Updating dependencies
2024-01-25 02:25:15.945Z: #13 1.180 Resolving dependencies...
2024-01-25 02:25:21.202Z: #13 6.587 
#13 6.587 Because langchain depends on azure-ai-vision (^0.11.1b1) which doesn't match any versions, version solving failed.
2024-01-25 02:25:21.653Z: #13 ERROR: process "/bin/sh -c poetry install --no-interaction --no-ansi --with dev,test,docs" did not complete successfully: exit code: 1
2024-01-25 02:25:21.750Z: ------
 > [langchain langchain-dev-dependencies 5/5] RUN poetry install --no-interaction --no-ansi --with dev,test,docs:
1.180 Updating dependencies
1.180 Resolving dependencies...
6.587 
6.587 Because langchain depends on azure-ai-vision (^0.11.1b1) which doesn't match any versions, version solving failed.
```
### The full stack trace is 1.5M lines (poetry solving dependencies in verbose mode), but the final message error is:
*(You may notice the poetry command is missing `--with dev,test,docs`. This is because I was experimenting with different installs to see if one would solve. The outcome is the same with or without.)*
```
failed to solve: process "/bin/sh -c poetry install -vvv --no-interaction --no-ansi --no-cache" did not complete successfully: exit code: 1
[...]

PR #18371, opened 2024-03-01
langchain: Remove `azure-ai-vision` as a dependency

cgalo5758:
# Addressing the Azure AI Vision package Deprecation

As documented in [Issue #16552](https://github.com/langchain-ai/langchain/issues/16552), Poetry version solving fails when attempting to deploy the dev container because the `azure-ai-vision` package has been yanked. The package has been replaced by `azure-ai-vision-imageanalysis`, which is unfortunately not a drop-in replacement.

Additionally, ["Azure Cognitive Services" has been rebranded as "Azure AI Services"](https://jamiemaguire.net/index.php/2023/08/05/cognitive-services-and-applied-ai-are-now-azure-ai-services/). As a result, the names currently used by the tools in the `AzureCognitiveServicesToolkit` and the toolkit itself are outdated.

This pull request aims to:
- Provide a short term fix for [Issue #16552](https://github.com/langchain-ai/langchain/issues/16552) by removing `azure-ai-vision` as a dependency.
- Initiate a discussion about updating the `AzureCognitiveServicesToolkit` and its tools.

## Removing `azure-ai-vision`

[The package `azure-ai-vision` has been yanked from PyPI](https://pypi.org/project/azure-ai-vision/), meaning it's no longer supported or recommended for use. The new package `azure-ai-vision-imageanalysis` is referenced as its replacement, but it is not a drop-in substitute as [the SDK has been rewritten and all APIs have changed](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/whats-new#new-image-analysis-sdk-100-beta1-breaking-changes).

To mitigate the impact of removing `azure-ai-vision` as a dependency while we work on updating the `AzureCognitiveServicesToolkit` and refactoring `AzureCogsImageAnalysisTool`, we add a warning on import error stating that the package has been yanked. Users who rely on the tool will still be able to download the yanked package if they need to, as described in [this official thread](https://learn.microsoft.com/en-us/answers/questions/1511970/error-occurred-while-trying-to-install-ai-vision-l).

By running `poetry remove azure-ai-vision` we remove `azure-ai-vision` from `libs/langchain/pyproject.toml` and `libs/langchain/poetry.lock`.

It is important to note that while this removal prevents `poetry` from hanging on this package, it doesn't prevent poetry from resolving dependencies indefinitely, that is its own issue.

## Updating `AzureCognitiveServicesToolkit`

[Cognitive Services and Applied AI are Now ‘Azure AI Services’](https://jamiemaguire.net/index.php/2023/08/05/cognitive-services-and-applied-ai-are-now-azure-ai-services/) and as such the names currently used by the tools referenced by the `AzureCognitiveServicesToolkit`and the toolkit itself are outdated.

I am not sure what the most sensible path forward here is, but as a suggestion we can:

1. Duplicate the Azure Cognitive Services tools and toolkit files and directories as a starting point for a new Azure AI Services tools and toolkit. We then update file names and code symbols accordingly.
2. Rewrite `image_analysis.py` to use the replacement package (`azure-ai-vision-imageanalysis`).
3. Implement notices in the old tools and toolkit to guide users towards the new tools toolkit, along with documentation updates.

Concerns about code duplication and the future of the old code merit discussion. My inclination is to phase out the outdated tools while providing users with adequate notice to transition to the updated toolkit, thus minimizing long-term duplication.

I am also not sure by *when* the old tools and toolkit should be deprecated by, but they should definitely be removed at some point in the future.

I am considering doing this myself and making a pull request, *but first I'd like confirmation that this is worth doing and this is a sensible course of action*.


libs/community/langchain_community/tools/azure_cognitive_services/image_analysis.py
```diff
@@ -14,6 +15,11 @@
 
 logger = logging.getLogger(__name__)
 
+warnings.warn(
+    "The 'azure-ai-vision' package has been yanked from PyPI and replaced by 'azure-ai-vision-imageanalysis'. "
```
baskaryan, line 19: should we update the code to use azure-ai-vision-imageanalysis? or are the packages sufficiently different that we'd basically need to rewrite the integration

cgalo5758: Unfortunately, `azure-ai-vision-imageanalysis` is not a drop-in replacement; The tool would need to be rewritten. [the SDK has been rewritten and all APIs have changed](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/whats-new#new-image-analysis-sdk-100-beta1-breaking-changes).

I did a deeper dive here to see what the differences are:
- The .whl files provided by `azure-ai-vision` and `azure-ai-vision-imageanalysis` are not even similar when analyzed.
- I also took the time to print out and analyze [this commit](https://github.com/MicrosoftDocs/azure-docs/commit/943b431c36fbc0df0e8a501e0c9ce3b996367cf2?diff=split&w=0) where they update the documentation from `azure-ai-vision` to `azure-ai-vision-imageanalysis` and there are differences in how you load local images (you now load the image as bytes instead of creating a `VisionSource` object). Additionally, they way of getting results from the service, error handling, and logging are all significantly different.

cgalo5758:
My sincerest apologies, I forgot to remove `azure-ai-vision` from the azure extras.

cgalo5758:
It seems like the dependency has been removed! Thank you!

PR #17247, opened 2024-02-08
community/templates: ClickHouse Template

mneedham:
Description: A RAG template for ClickHouse. Also has a change to let users pass in `index_type` as `None`.
Twitter handle: @markhneedham

PR #18381, opened 2024-03-01
community: ClickHouse: Make it possible to not specify a vector index

mneedham:
Vector indexes in ClickHouse are experimental at the moment and can sometimes break/change behaviour. So this PR makes it possible to say that you don't want to specify an index type.

Any queries against the embedding column will be brute force/linear scan, but that gives reasonable performance for small-medium dataset sizes. 

In the other PR (https://github.com/langchain-ai/langchain/pull/17247), @efriis asked:

```
> index_type: Optional[str] = "annoy"

What is this supposed to do? I think this will have unintended consequences in some string substitution stuff lower in the file
```

This makes `index_type` optional. I think I've defended against it being `None` inside the `_schema` function, which is where it's used. But let me know if you think I've missed somewhere else where it's used.


libs/community/langchain_community/vectorstores/clickhouse.py
```diff
@@ -209,6 +199,41 @@ def __init__(
         self.client.command(f"SET allow_experimental_{self.config.index_type}_index=1")
```
efriis, line 199: @mneedham just confirming - I'm pretty sure this change turns this into ```self.client.command(f"SET allow_experimental_None_index=1")``` unless I'm mistaken. This seems undesirable?

mneedham:
@efriis I think I'm handling the empty `index_type` now, but it was also telling me that the width of the docstring that I added for the `_schema` function was too wide. I think I've fixed that, but not sure how to get it to run the linter on CI again?

mneedham:
@efriis how do I get the build to run? I think I've fixed the lint issues that were there before.

efriis (approved):

efriis:
Fixing in #19527 

In the future if you make PRs from a personal branch, can push directly to your PR branch! https://python.langchain.com/docs/contributing/faq#how-do-i-allow-maintainers-to-edit-my-pr

PR #18385, opened 2024-03-01
community: Use default load() implementation in doc loaders

cbornet:
Following https://github.com/langchain-ai/langchain/pull/18289

### DIFF
--- a/libs/community/langchain_community/document_loaders/github.py
+++ b/libs/community/langchain_community/document_loaders/github.py
@@ -127,32 +127,6 @@ class GitHubIssuesLoader(BaseGitHubLoader):
             else:
                 url = None
 
-    def load(self) -> List[Document]:
-        """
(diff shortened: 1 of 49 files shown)


libs/community/langchain_community/document_loaders/github.py
```diff
@@ -127,32 +127,6 @@ def lazy_load(self) -> Iterator[Document]:
             else:
                 url = None
 
-    def load(self) -> List[Document]:
```
baskaryan, line 130: a few of these, like this one, are nice for the API reference because they overwrite the docstring

eyurtsev (approved):

libs/community/langchain_community/document_loaders/github.py
eyurtsev: not super worried about this since lazy_load has the documentation

PR #18401, opened 2024-03-01
Add version

hinthornw:
Need to bump langsmith dep


libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -963,10 +963,12 @@ def _prepare_eval_run(
     project_name: str,
     project_metadata: Optional[Dict[str, Any]] = None,
     tags: Optional[List[str]] = None,
+    dataset_version: Optional[Union[str, datetime]] = None,
 ) -> Tuple[MCF, TracerSession, Dataset, List[Example]]:
     wrapped_model = _wrap_in_chain_factory(llm_or_chain_factory, dataset_name)
     dataset = client.read_dataset(dataset_name=dataset_name)
-    examples = list(client.list_examples(dataset_id=dataset.id))
+    as_of = dataset_version if isinstance(dataset_version, datetime) else None
```
eyurtsev, line 970: random but maybe move above # PUBLIC API since it's a private function

libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -963,10 +963,12 @@ def _prepare_eval_run(
     project_name: str,
     project_metadata: Optional[Dict[str, Any]] = None,
     tags: Optional[List[str]] = None,
+    dataset_version: Optional[Union[str, datetime]] = None,
```
eyurtsev, line 966: what happens if it's a string?

hinthornw: Version tags ("prod")

But it's ambiguous what would happen if you have a string datetime or something, which is bad

PR #18436, opened 2024-03-02
community: Implement lazy_load() for ConfluenceLoader

cbornet:
Implement basic `lazy_load` as just an iterator of the `load()` method.
Also move loading params in constructor instead of incorrectly doing it in the `load()` method.

### DIFF
--- a/libs/community/langchain_community/document_loaders/confluence.py
+++ b/libs/community/langchain_community/document_loaders/confluence.py
@@ -1,7 +1,7 @@
 import logging
 from enum import Enum
 from io import BytesIO
-from typing import Any, Callable, Dict, List, Optional, Union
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/document_loaders/confluence.py
```diff
@@ -343,17 +380,20 @@ def load(
...
                     include_attachments,
                     include_comments,
                     content_format,
                     ocr_languages,
                     keep_markdown_format,
                 )
-                docs.append(doc)
 
-        return docs
+    def load(self, **kwargs: Any) -> List[Document]:
```
eyurtsev, line 392: thank you for making this backwards compatible

eyurtsev (approved):

PR #11527, opened 2023-10-08
feat: add batch_loader class

fapont:
- **Description:** add a batch loader to better use loader on a list of data sources
  - **Issue:**  #11509

Issue #11509, opened 2023-10-07
[Feature] Batch loader


fapont:
### Feature request

Loaders are designed to load a single source of data and transform it to a list of Documents. If several sources are to be processed, it is up to the developer to call the Loader several times.
The idea is to propose a new `BatchLoader` object that could simplify this task and enable loading tasks to be launched sequentially, mutlithreaded, with mutli processor or asynchronously. 

Draft idea:
```
class BatchLoader(BaseLoader):
    ...

    def load(self, method: str = 'sequential', max_workers: int=1) -> List[Document]:
        # TODO: faire ça stylé
        if method == 'thread':
            return self._load_thread(max_workers)
        elif method == 'process':
            return self._load_process(max_workers)
        elif method == 'sequential':
            return self._load_sequential()
        elif method == 'async':
            return asyncio.run(self._async_load())
        else:
            raise ValueError(f'Invalid method {method}')
    
    ...

# Call it with a Loader constructor callable and a set of arguments
batch_loader = BatchLoader(TextLoader, {"file_path": [f"file_{i}.txt" for i in range(1000)]})
```

### Motivation

In most real use case we need to load several Data Sources, combine their outputs and work with them (e.g. store in a vector store). Some use cases also require to wait for I/O intensive tasks or tasks you want to parallelized, so having a Meta loader that wrap this complexity for you could be a good idea. 

### Your contribution

WIP: https://github.com/langchain-ai/langchain/pull/11527

fapont:
Sorry for the long wait. I will continue this ticket soon since my workload has decreased, see you soon 😊

PR #18443, opened 2024-03-03
community: add batch loader class

fapont:
- **Description:** add a batch loader to better use loader on a list of data sources
  - **Issue:**  #11509 
  - **Dependencies**: Nothing

This PR is the continuation of this one: https://github.com/langchain-ai/langchain/pull/11527


libs/community/langchain_community/document_loaders/batch_loader.py
```diff
@@ -0,0 +1,288 @@
...
+        self.show_progress = show_progress
+        self._partial_load = partial(_load_loader, loader_callable=self.loader_callable)
+        self.method = method
+        self.max_workers = max_workers
+
+        # Check that all lists in loader_args have the same length
+        if len(set(len(arg) for arg in init_loader_args.values())) != 1:
+            raise ValueError("All loader_args lists must have the same length")
+
+    def load(self) -> List[Document]:
```
cbornet, line 158: `load` shouldn't be overridden for new loaders.
It's `lazy_load` that should be implemented.

libs/community/langchain_community/document_loaders/batch_loader.py
```diff
@@ -0,0 +1,288 @@
...
+        elif self.method == "sequential":
+            return self._load_sequential()
+        elif self.method == "async":
+            raise ValueError(
+                "You must use the `aload` method to use the 'async' method"
+            )
+        else:
+            raise ValueError(f"Invalid method {self.method}")
+
+    async def aload(self) -> List[Document]:
```
cbornet, line 194: The method in `BaseLoader` to override is `alazy_load()`

libs/community/langchain_community/document_loaders/batch_loader.py
```diff
@@ -0,0 +1,288 @@
...
+    async def _async_load(self) -> List[Document]:
+        """Load data from the loader using asyncio.
+
+        Returns:
+            List[Document]: list of loaded documents flattened.
+        """
+
+        event_loop = asyncio.get_event_loop()
+        tasks = [
+            event_loop.run_in_executor(
```
cbornet, line 211: This defeats the goal of asyncio of not using threads.
The `alazy_load` method of the underlying loaders should be called instead.

libs/community/langchain_community/document_loaders/batch_loader.py
```diff
@@ -0,0 +1,288 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        loader_initializer_callable: Callable[..., BaseLoader],
+        init_loader_args: Dict[str, List[Any]],
+        *,
+        method: Methods = "sequential",
+        max_workers: int = 1,
```
cbornet, line 115: This could be useful also for async to not overload the remote system or respect rate limits.
Maybe `max_concurrency` would be a better term ?

libs/community/langchain_community/document_loaders/batch_loader.py
```diff
@@ -0,0 +1,288 @@
...
+        show_progress: bool = False,
+    ):
+        """Initialize the loader with a loader_callable (either a class or a
+        function used to initialize the loader) and a list of loader_args.
+
+        Possible methods are:
+            - "sequential": load data sequentially
+            - "thread": load data using threads
+            - "process": load data using processes
+            - "async": load data using asyncio
```
cbornet, line 125: `async` doesn't seem needed as this mode is determined by the usage of `load` vs `aload`.
`sequential` could also be removed by assuming that if the concurrency is 1, then it will be sequential (can detect that if max_workers is 1 then there's no need to spawn a thread-pool or processes and the operation can be done in the current thread).

libs/community/langchain_community/document_loaders/batch_loader.py
```diff
@@ -0,0 +1,288 @@
...
+    Then you can load your data:
+    ```
+    documents = loader_files.load()
+    ```
+    """
+
+    def __init__(
+        self,
+        loader_initializer_callable: Callable[..., BaseLoader],
+        init_loader_args: Dict[str, List[Any]],
```
cbornet, line 112: Wouldn't it be easier to just pass a list of loaders ?

fapont: You mean a list of already instantiated loaders ? Indeed this could simplify the code of the batch loader and allow users to load data from multiple loaders using 1 batch loader. 
But I'm not sure if, from a developer perspective, it would be a better experience or not.
User's code would be:
```python
# Propisition 1
BatchLoader(
    YoutubeLoader.from_youtube_url,
    {"youtube_url": youtube_videos}
)

# Proposition 2
loaders = [YoutubeLoader.from_youtube_url(url) for url in youtube_videos]
BatchLoader(loaders)
```

I do not have strong conviction on what could be the best way of doing that, if you prefer I can simplify this of course

libs/community/langchain_community/document_loaders/batch_loader.py
fapont: You're correct. It might be better to have only the max_concurrency argument and then call either alazy_load or lazy_load.

thread isn't useful at all. Perhaps in the future with subinterpreter, there could be some interest, but for now, let's get rid of it.

fapont:
Thanks @cbornet for all your comments, I'll try to provide a new version asap 😊

eyurtsev:
Closing as stasle

PR #18445, opened 2024-03-03
core: Fix stream chunk merge error.

### LINKED ISSUE #18441: when using qianfan chat streaming occurs TypeError: Additional kwargs key sentence_id already exists in left dict and value has unsupported type <class 'int'>.
### Checked other resources

### Example Code

```python
qianfan_model = QianfanChatEndpoint(model='ERNIE-Bot', qianfan_ak=...... ,
                                    qianfan_sk=......)

for chunk in qianfan_model.stream(input_prompt):
    all_output += chunk.content
    print(chunk.content, end="", flush=True)
```

  File "E:\JMIT\JMIT3\code\ans.py", line 247, in <module>
    print(retrieve_ans_chat_stream('我想離婚，孩子該怎麼判？', history=[('你好', '你好呀')]))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\JMIT\JMIT3\code\ans.py", line 236, in retrieve_ans_chat_stream
    for chunk in qianfan_model.stream(input_prompt):
  File "E:\envs\JMIT\Lib\site-packages\langchain_core\language_models\chat_models.py", line 250, in stream
    raise e
  File "E:\envs\JMIT\Lib\site-packages\langchain_core\language_models\chat_models.py", line 241, in stream
    generation += chunk
  File "E:\envs\JMIT\Lib\site-packages\langchain_core\outputs\chat_generation.py", line 57, in __add__
    generation_info = merge_dicts(
                      ^^^^^^^^^^^^
  File "E:\envs\JMIT\Lib\site-packages\langchain_core\utils\_merge.py", line 38, in merge_dicts
    raise TypeError(
TypeError: Additional kwargs key created already exists in left dict and value has unsupported type <class 'int'>.

### Description

This bug occurs when the langchain_core>0.1.12
Downgrading to langchain_core==0.1.12 is bug-free

### System Info

langchain==0.1.10
langchain-community==0.0.25
langchain-core==0.1.28

windows

Python 3.11.7

liugddx:
Description: Fix stream chunk merge error.
Issue:  fix #18441


hwchase17 (commented):
we should add a test

id also love to understand WHY this is happening

liugddx:
> we should add a test
> 
> id also love to understand WHY this is happening

I found during testing that the chunk will contain timestamps, which should not be merged. There are many numerical types that should be filtered out.

liugddx:
> > we should add a test
> > id also love to understand WHY this is happening
> 
> I found during testing that the chunk will contain timestamps, which should not be merged. There are many numerical types that should be filtered out.

The best strategy would be to define a merge strategy.

liugddx:
@hwchase17 PTAL,  thanks~

liugddx:
@baskaryan PTAL thanks~

libs/core/langchain_core/utils/_merge.py
```diff
@@ -23,6 +24,9 @@ def merge_dicts(left: Dict[str, Any], right: Dict[str, Any]) -> Dict[str, Any]:
             merged[k] = v
         elif v is None or merged[k] == v:
             continue
+        # TODO Numeric type data is temporarily passed.
+        elif isinstance(v, numbers.Number):
+            continue
```
baskaryan, line 29: not sure this generalizes

when/why would you get two numbers and want to ignore one?

liugddx: For example, the timestamp attribute should not be merged.

liugddx: > not sure this generalizes
> 
> when/why would you get two numbers and want to ignore one?

So skip it here instead of throwing an exception directly? 


This is the real data when I use qianfan, They encounter a timestamp merge exception when merging.

image


```
{'id': 'as-juqwpqiphi', 'object': 'chat.completion', 'created': 1709774979, 'sentence_id': 0, 'is_end': False, 'is_truncated': False, 'result': 'Of course', 'need_clear_history': False, 'finish_reason': 'normal', 'usage': {'prompt_tokens': 5, 'completion_tokens': 0, 'total_tokens': 5}}
```

```
{'id': 'as-juqwpqiphi', 'object': 'chat.completion', 'created': 1709774980, 'sentence_id': 1, 'is_end': False, 'is_truncated': False, 'result': ', I\'d be happy to tell you a story. Here\'s a classic fairy tale called "The Ugly Duckling":\n\nOnce upon a', 'need_clear_history': False, 'finish_reason': 'normal', 'usage': {'prompt_tokens': 5, 'completion_tokens': 0, 'total_tokens': 5}}
```


There needs to be a better merging strategy to support

baskaryan: something like timestamp should probably be dropped from streaming chunks before it comes to merging them (or timestamp field should be converted to a list that's appended to)

liugddx: > something like timestamp should probably be dropped from streaming chunks before it comes to merging them (or timestamp field should be converted to a list that's appended to)

Do you mean not to change merge method but to format the chunk in Qianfan?

liugddx:
@baskaryan PTAL , thanks~

liugddx:
@eyurtsev PTAL. Thanks

liugddx:
> @eyurtsev PTAL. Thanks

@baskaryan @eyurtsev @hwchase17 Hi, kindly pin. Can anyone help me take a look at this PR? I need to use qianfan model.

baskaryan (approved):

libs/core/langchain_core/utils/_merge.py
```diff
@@ -30,6 +30,8 @@ def merge_dicts(left: Dict[str, Any], right: Dict[str, Any]) -> Dict[str, Any]:
             )
         elif isinstance(merged[k], str):
             merged[k] += v
+        elif isinstance(merged[k], int):
+            merged[k] += v
```
baskaryan, line 34: wait sorry we don't want to add up ints either. its ambiguous how to merge generic ints which is why they're explicitly not supported

liugddx: > wait sorry we don't want to add up ints either. its ambiguous how to merge generic ints which is why they're explicitly not supported

Okay, let me close this PR first. Thanks @baskaryan

PR #18459, opened 2024-03-03
docs: anthropic qa quickstart

baskaryan:


Dominastorm:
Why was the Python notebook deleted here? It has still been referenced in https://python.langchain.com/docs/use_cases/question_answering/quickstart in the "Open in Colab" button

PR #18468, opened 2024-03-03
langchain: adding object title when converting schema before doing the openai function call

arnaupv:
- [x] **PR message**:
    - **Description:** adding the property "title" (if present) when converting the schema before sending it to the openai function call. This can help to improve the response provided by the openai LLM.


baskaryan:
out of curiousity what's the motivation? openai functions always have a top-level name (outside the json schema) anyways, so seems like a schema title doesn't add much?

ccurme:
Closing as author is unresponsive, can re-open if needed.

PR #18491, opened 2024-03-04
community[patch]: chat message histrory mypy fixes

Lord-Haji:
Description: Addresses mypy issues for chat_message_histories(cassandra, dynamodb, elasticsearch, file, momento)
Adresses https://github.com/langchain-ai/langchain/issues/17048

Planning to add more based on reviews


Lord-Haji:
Bumping this @baskaryan since this PR might have gone unnoticed

hwchase17:
i dont think this is the best way to do it. creating a new attribute is a bit overkill. I think if we type things correctly theres probably a more clear solution

PR #18506, opened 2024-03-04
elasticsearch: update references, deprecate community classes

maxjakob:
Follow up on https://github.com/langchain-ai/langchain/pull/17467.

- Update all references to the Elasticsearch classes to use the partners package.
- Deprecate community classes.


libs/community/langchain_community/embeddings/elasticsearch.py
```diff
@@ -11,6 +12,9 @@
 from langchain_core.embeddings import Embeddings
 
 
+@deprecated(
+    "0.1.11", alternative="Use class in langchain-elasticsearch package", pending=True
```
baskaryan, line 16: this should be the communtiy version, so 0.0.26

baskaryan (approved):

PR #18517, opened 2024-03-04
anthropic[patch]: multimodal

efriis:
- anthropic[minor]: claude 3
- x
- x


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -36,22 +61,66 @@ def _format_messages(messages: List[BaseMessage]) -> Tuple[Optional[str], List[D
...
+            continue
+
+        role = _message_type_lookups[message.type]
+        content: Union[str, List[Dict]]
+
+        if not isinstance(message.content, str):
+            # parse as dict
+            assert isinstance(
+                message.content, list
+            ), "Anthropic message content must be str or list of dicts"
```
baskaryan, line 85: why not accept list of str

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -36,22 +61,66 @@ def _format_messages(messages: List[BaseMessage]) -> Tuple[Optional[str], List[D
...
+                    content.append(
+                        {
+                            "type": "text",
+                            "text": item,
+                        }
+                    )
+                elif isinstance(item, dict):
+                    if "type" not in item:
+                        raise ValueError("Dict content item must have a type key")
+                    if item["type"] == "image_url":
```
baskaryan, line 100: should we support "image_url" or "image"? as an Anthropic user i might expect "image" support

efriis: Supports both. `image_url` is the only one that triggers the conversion logic though.

libs/partners/anthropic/langchain_anthropic/chat_models.py
efriis: I think it does?

hinthornw:
```suggestion
            ), "Anthropic message content must be str or list of strings or dicts"
```

hinthornw:
In a followup would be nice to support image urls

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -24,6 +25,30 @@
...
+      "media_type": "image/jpeg",
+      "data": "/9j/4AAQSkZJRg...",
+    }
+
+    And throws an error if it's not a b64 image
+    """
+    regex = r"^data:(?P<media_type>image/.+);base64,(?P<data>.+)$"
+    match = re.match(regex, image_url)
+    if match is None:
+        raise ValueError("Anthropic only supports base64-encoded images currently")
```
hinthornw, line 44:
```suggestion
        raise ValueError(
            "Anthropic only supports base64-encoded images currently."
            " Example: "data:image/png;base64,'/9j/4AAQSk'..."
        )
```

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -24,6 +25,30 @@
 _message_type_lookups = {"human": "user", "ai": "assistant"}
 
 
+def _format_image(image_url: str) -> Dict:
```
hinthornw, line 28: FYI @dqbd the default anthropic format won't render properly in langsmith if implemented this way 

Think fine here

Issue #18406, opened 2024-03-01
GitHub agent throwing TypeError: GitHubAction._run() Unexpected Keyword Argument 'issue_number'


ajhous44:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code:

``` python
import os

from langchain.agents import AgentType, initialize_agent
from langchain_community.agent_toolkits.github.toolkit import GitHubToolkit
from langchain_community.utilities.github import GitHubAPIWrapper
from langchain_openai import ChatOpenAI

# Set your environment variables using os.environ
os.environ["GITHUB_APP_ID"] = ""
os.environ["GITHUB_APP_PRIVATE_KEY"] = "pemfile"
os.environ["GITHUB_REPOSITORY"] = ""
os.environ["GITHUB_BRANCH"] = ""
os.environ["GITHUB_BASE_BRANCH"] = "main"

# This example also requires an OpenAI API key
os.environ["OPENAI_API_KEY"] = "####"

llm = ChatOpenAI(temperature=0, model="gpt-4-1106-preview")
github = GitHubAPIWrapper()
toolkit = GitHubToolkit.from_github_api_wrapper(github)
tools = toolkit.get_tools()

# STRUCTURED_CHAT includes args_schema for each tool, helps tool args parsing errors.
agent = initialize_agent(
	tools,
	llm,
	agent=AgentType.STRUCTURED_CHAT_ZERO_SHOT_REACT_DESCRIPTION,
	verbose=True,
)
print("Available tools:")
for tool in tools:
	print("\t" + tool.name)

agent.run(
    "You have the software engineering capabilities of a Google Principle engineer. You are tasked with completing issues on a github repository. Please look at the existing issues and complete them."
)
```

### Error Message and Stack Trace (if applicable)

```

> Entering new AgentExecutor chain...
Action:
```json
{
  "action": "Get Issues",
  "action_input": {}
}
```

Observation: Found 1 issues:
[{'title': 'Create a new html file', 'number': 1, 'opened_by': 'ajhous44'}]
[...]

PR #18523, opened 2024-03-04
[Community]: GitHubAction Tool Parameter Error Fix

keenborder786:
Github Action Tool had an error in parameter passing for `_run` which was causing the failure of tool(s) invoking. This was identified in the following issue: #18406


keenborder786:
@baskaryan

PR #18543, opened 2024-03-05
Fix fireworks chat async streaming

tjaffri:
Async streaming for fireworks chat seemed to have a typo, fixed in this PR (L259 of fireworks.py)

As part of this, I realized that a lot of integration tests for fireworks were failing... probably not being run regularly? This particular issue was causing one of those tests to fail too and that issue is resolved.

The other tests were failing due to incorrectly expecting stop words to be included in the output as the last char... this is not the right expectation and the tests are now all passing.

### DIFF
--- a/libs/community/langchain_community/chat_models/fireworks.py
+++ b/libs/community/langchain_community/chat_models/fireworks.py
@@ -223,7 +223,7 @@ class ChatFireworks(BaseChatModel):
                 message=chunk, generation_info=generation_info
             )
             if run_manager:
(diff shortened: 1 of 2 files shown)


efriis:
cc @benjibc

efriis:
Oh actually I take it back - this is the deprecated community implementations (on me not Benny). 

@tjaffri I think you're using a deprecated version of fireworks. Could you switch to the integration package one instead? I'll add the deprecated decorators to this implementation.

https://python.langchain.com/docs/integrations/providers/fireworks

Issue #18513, opened 2024-03-04
Bedrock integration with Claude3 fails with ValidationException "claude-3-sonnet-20240229" is not supported on this API. Please use the Messages API instead.


Barneyjm:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
import boto3
from langchain_community.llms import Bedrock

bedrock = boto3.client('bedrock-runtime' , 'us-east-1')

MODEL_KWARGS = {
"anthropic.claude-3-sonnet-20240229-v1:0": {
        "temperature": 0, 
        "top_k": 250, 
        "top_p": 1, 
        "max_tokens_to_sample": 2**10 
}}

model_id = 'anthropic.claude-3-sonnet-20240229-v1:0'
llm = Bedrock(model_id=model_id, model_kwargs=MODEL_KWARGS[model_id])
llm('tell me a joke')
```

### Error Message and Stack Trace (if applicable)

```
.venv/lib/python3.11/site-packages/langchain_core/_api/deprecation.py:117: LangChainDeprecationWarning: The function `__call__` was deprecated in LangChain 0.1.7 and will be removed in 0.2.0. Use invoke instead.
  warn_deprecated(
Traceback (most recent call last):
  File ".venv/lib/python3.11/site-packages/langchain_community/llms/bedrock.py", line 444, in _prepare_input_and_invoke
    response = self.client.invoke_model(**request_options)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/client.py", line 553, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/botocore/client.py", line 1009, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: "claude-3-sonnet-20240229" is not supported on this API. Please use the Messages API instead.

During handling of the above exception, another exception occurred:

[...]

PR #18548, opened 2024-03-05
community: Enable Claude v3

Barneyjm:
**Description:** implements changes for Claude V3 models through bedrock
Note: i'm not sure how to best handle the many breaking changes that Anthropic introduced into v3 of the model. 

I'd suggest we do something like `model_id = 'anthropic_v3'` to distinguish old model input formats or make some serious changes to the other input models. But that will also introduce it's own headaches. 

Currently this works by stripping out the "_v3" dynamically where needed to make boto3 happy but also make it possible to respect the two different input formats between v1/2 Claude and v3 Claude. Obviously this isn't ideal since "anthropic_v3" isn't a thing anywhere in Anthropic's documentation or Amazon's documentation. But I don't know how else we could make it work.

Open to any feedback

**Issue:** #18513


Barneyjm:
lint hates me. no idea why openai's file can have a [near identical function](https://github.com/langchain-ai/langchain/pull/18548/files#diff-c1e8e71a39b6defd92161f82ee4d7e500d2b55f07ebfe383bc87047c13edd3b4R84) and not get caught in this linter

deepend-dev:
Great initiative. Does it also support vision api?

libs/community/langchain_community/llms/bedrock.py
```diff
@@ -77,6 +77,22 @@ def _human_assistant_format(input_text: str) -> str:
     return input_text
 
 
+def _generate_messages_format(input_text: str) -> list:
```
jonathancaevans, line 80: What about multiple turns for word in mouth of assistant turn?

It might be best to convert the old prompt formatting to the new messages by adding some logic to break things up Assistant Human turns into messages. This will give backwards compatibility for Claude 2 text completion prompts.

jonathancaevans:
@Barneyjm Can we add support for ChatBedrock here too? I like that core llm implementation is added for supporting easy migration from Claude 2 Bedrock, but there should be native support for Langchain messages added to ChatBedrock too.

libs/community/langchain_community/llms/bedrock.py
jonathancaevans: The idea would be for Cluade 2 prompts such as,
 "prompt": "\n\nHuman: Write me a play. \n\nAssistant: Here is a play about cats, " 

would be translated to,
```
    "messages": [
        {
            "content": [
                {
                    "text": "Write me a play.",
                    "type": "text"
                }
            ],
            "role": "user"
        },
        {
            "content": [
                {
                    "text": "Sure, here is a play about cats,",
                    "type": "text"
                }
            ],
            "role": "assistant"
        }
    ]
```
By using the regex, this function could offer full breadth backwards compatibility to folks migrating their prompts from Claude 2, to Claude 3 models without having to migrate completion string prompts to Langchain messages.

When it comes to BedrockChat, I have an implementation on hand, let me see if I can merge that with your branch...

libs/community/langchain_community/chat_models/bedrock.py
```diff
@@ -28,7 +28,7 @@ class ChatPromptAdapter:
     def convert_messages_to_prompt(
         cls, provider: str, messages: List[BaseMessage]
     ) -> str:
-        if provider == "anthropic":
+        if provider == "anthropic" or provider == "anthropic_v3":
```
Barneyjm, line 31: @jonathancaevans This was the only change required for BedrockChat since it leverages the llms/bedrock.py LLMInputOutputAdapter class. 

without this differentiation you run into the same "either break all the old stuff or force everyone to update everything" conundrum as before.

libs/community/langchain_community/llms/bedrock.py
Barneyjm: The key thing here is that Bedrock itself has the model_id and input format "hardcoded" to expect certain inputs. 

That regex approach would work if Bedrock wasn't validating the input format -- we almost need to un-regex it and put it into the format that Bedrock expects for each Claude model. 

and then the trouble there is each Claude model has different input/output structures that you can't sniff out without differentiating as a "new" provider, `anthropic_v3`. 

the parameters for `prepare_input_stream` and `prepare_output_stream` has no model information in it, only provider -- so we can't programmatically say "for claude v2, do this kind of regex operation but for claude v3, leave it alone". 

to the LLMInputOutputAdapter class, it all just looks like "anthropic" without this new v3 moniker.

jobcase-plebedev: The messages API also supports `system` property for a system prompt, so it can be more sophisticated than just Human/Assistant sequence. 

While playing with it, I found that Claude 3 and Claude 2 both work fine with messages API (i.e. no errors, and reasonable response), when you feed them a message that includes Human/Assistant sequence. I.e. w/o having a need to have a separate anthropic_v3 provider, you can feed the exact same `"prompt": "\n\nHuman: Write me a play. \n\nAssistant: Here is a play about cats, "` 

I made a change locally that looks like this
```
      if provider == "anthropic":
            if input_body.get("use_completion_api") is True:
                input_body["prompt"] = _human_assistant_format(prompt)
                if "max_tokens_to_sample" not in input_body:
                    input_body["max_tokens_to_sample"] = 256
                del input_body["use_completion_api"]
            else:
                input_body["messages"] = _generate_messages_format(prompt)
```   

And for Claude 2 it returns very similar results for the same prompt in both cases, when  `use_completion_api` is set, and when it's not. 

Though, to use messages API with Claude  2, `model_kwargs` need to be something like this:
```
model_kwargs_claude = {
                "temperature": 0,
                "top_k": 10,
                "max_tokens": 4000,
                "anthropic_version": "bedrock-2023-05-31",
            }
 ```

**Comment by Adibuer-lab** [comment]
I was also facing this issue and tried to sort it out myself, my PR is here with the image support-maybe we can put these together somehow or you can take some stuff from it? : #18579 ,

It converts the input to human-assistant format if it already isn't, if it is, all good. then changes it to the v3 format to a list of message dictionaries with the new user role instead of human. And if there's an image it adds it to the latest user input

**Comment by Barneyjm** [comment]
> I was also facing this issue and tried to sort it out myself, my PR is here with the image support-maybe we can put these together somehow or you can take some stuff from it? : #18579 ,
> 
> It converts the input to human-assistant format if it already isn't, if it is, all good. then changes it to the v3 format to a list of message dictionaries with the new user role instead of human. just trying to get that backwards compatibility since this new list of messages format looks like it adds structure to the previous string prompt. And if there's an image it adds it to the latest user input

i like this approach, let me add some of this stuff in and add you as a reviewer?

**Comment by Barneyjm** [comment]
@Adibuer-lab added your clever model_id injection to the input/output process. @jobcase-plebedev added your if/else within the single anthropic provider. Nice!

Will look at adding media here shortly

**Comment by Adibuer-lab** [comment]
@Barneyjm Turns out the check might not be necessary since previous claude versions support the messages api. Might just end up deprecating the test completions human assistant format. for now I've also maintained the check though.

**Comment by ryancrepo** [comment]
@Adibuer-lab Would you mind posting here when there's a dev build we can try?  Pretty sure there are a lot of us excited to try without having to rewrite our query architecture.  Thanks!

**Review by efriis — COMMENTED** [review]
Hey there! Given claude v3 models only support the messages api in anthropic, would it make sense to add this compatibility to `BedrockChat` instead? https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html

**Comment by Barneyjm** [comment]
> Hey there! Given claude v3 models only support the messages api in anthropic, would it make sense to add this compatibility to `BedrockChat` instead? https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html

from my understanding, the BedrockChat class still uses this LLM input/output class under the covers, so it'll need to be added here

**Comment by efriis** [comment]
Hey @Barneyjm ! I think it uses that for some other models, but the class inherits from `BaseChatModel` which will allow people interfacing with it to pass messages instead of raw text.

Because the new claude 3 models are *only* chat models, and aren't compatible with text completions, I would prefer to implement the logic over there.

I'm less familiar with the bedrock integrations and just starting to dig so might have some stupid questions / need some education :) - are you saying that there's an existing [bedrock claude messages api](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html) integration on this Bedrock LLM and not BedrockChat Chat Model that this hooks into?

**Comment by efriis** [comment]
cc @Adibuer-lab - would love your take on the above too (just easier to discuss all of us on one thread rather than splitting between the two)

**Comment by Barneyjm** [comment]
@efriis I've only spent a few hours on this so i'm by no means the expert! i am now painfully familiar with how these things interact though :P 

in bedrock chat, i see it imports [BedrockBase](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/chat_models/bedrock.py#L15)

the BedrockBase class leverages the [LLMInputOutputParser](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/llms/bedrock.py#L426) which is, for better or worse, in that same [bedrock.py file](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/llms/bedrock.py#L80)

so by adding these changes here, i believe we are adding it to the chat models. while the new models are indeed chat-only, i think the way this is currently structured makes us put it here? of course we could always move it over to bedrock chat!

**Comment by Adibuer-lab** [comment]
> @efriis I've only spent a few hours on this so i'm by no means the expert! i am now painfully familiar with how these things interact though :P
> 
> in bedrock chat, i see it imports [BedrockBase](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/chat_models/bedrock.py#L15)
> 
> the BedrockBase class leverages the [LLMInputOutputParser](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/llms/bedrock.py#L426) which is, for better or worse, in that same [bedrock.py file](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/llms/bedrock.py#L80)
> 
> so by adding these changes here, i believe we are adding it to the chat models. while the new models are indeed chat-only, i think the way this is currently structured makes us put it here? of course we could always move it over to bedrock chat! my thinking was by adding it here, we'd maintain the old prediction api while it's still around and add the new stuff transparently

yeah I've seen similar, BedrockBase is using [LLMInputOutputParser](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/llms/bedrock.py#L426)whereas BedrockChat is using a [ChatPromptAdapter](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/chat_models/bedrock.py) to handle that conversion. The ChatPromptAdapter uses functions from  [anthropic.py in chat_models](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/chat_models/anthropic.py) to handle the conversion.

BedrockChat then uses _prepare_input_and_invoke_stream from BedrockBase but it's already done the conversion to human-assistant format. Making the change in BedrockChat will involve this ChatPromptAdapter but that'll leave BedrockBase without messages api compatibility, and anyone using BedrockBase won't get that.

**Comment by Adibuer-lab** [comment]
@Barneyjm I added a way to allow people using the messages api, to use their json strings as input. It just takes the json string and converts it to json and if the prompt is in completions format it converts it.
expects the json string as input because it looks like "messagges" in request body should a list of alternating user, assistant messages so the resultant request body looks something like.
```json
{
  "body": {
    "max_tokens": 4096,
    "anthropic_version": "bedrock-2023-05-31",
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "Hello, I'm looking for some travel advice."
          }
        ]
      },
      {
        "role": "assistant",
        "content": [
          {
            "type": "text",
            "text": "Sure, I can help with that. What's your destination?"
          }
        ]
      },
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "I'm planning a trip to Japan. Specifically, I'm interested in historical sites in Kyoto. Here is a place I would like to go."
          },
          {
            "type": "image",
            "source": {
              "type": "base64",
              "media_type": "image/jpeg",
              "data": "/9j/4AAQSkZJRgABAQAAAQABAA..."
            }
          }
        ]
      },
      {
        "role": "assistant",
        "content": [
          {
            "type": "text",
            "text": "The image depicts the famous Kinkaku-ji Temple, also known as the Golden Pavilion, in Kyoto, Japan. This Zen Buddhist temple is one of Kyoto's most iconic attractions and a must-visit for anyone interested in historical sites.
The temple is a three-story building covered in brilliant gold leaf, elegantly situated beside a tranquil pond surrounded by well-manicured gardens. The top two floors are constructed in the traditional Japanese architectural style, while the first floor is inspired by Buddhist architecture.
The serene landscape, with the temple's golden reflection shimmering in the pond's waters, lush greenery, carefully placed rocks, and the backdrop of forested hills, exemplifies the aesthetic ideals of Japanese garden design.
Kinkaku-ji dates back to the 14th century and has survived fires and wars over the centuries. Visiting this visually stunning and culturally significant temple is definitely recommended when exploring Kyoto's rich history and traditional architecture."
          }
        ]
      },
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "Thanks! That was really helpful. Here's a photo from my last trip, by the way. Can I get from the place in my last trip to the previous image I shared."
          },
          {
            "type": "image",
            "source": {
              "type": "base64",
              "media_type": "image/jpeg",
              "data": "/9j/4AAQSkZJRgAB..."
            }
          }
        ]
      }
    ]
  },
  "modelId": "anthropic.claude-3-sonnet-20240229-v1:0",
  "accept": "application/json",
  "contentType": "application/json"
}

```
This way you could have some image support?

Barneyjm:
closing this in place of @3coins PR: https://github.com/langchain-ai/langchain/pull/18630

marfago:
Is there a PR for Bedrock LLM to support Claude 3?

PR #18554, opened 2024-03-05
anthropic[minor]: add tool calling

efriis:
Current shortcomings:
- xml parser returns everything as a string (but with_structured_output/pydantic turns them into proper types)
- complex object arguments aren't supported well


libs/partners/anthropic/langchain_anthropic/experimental.py
baskaryan, line 1: don't feel strongly but think we've called this beta in other packages (like core) instead of experimental

libs/partners/anthropic/langchain_anthropic/experimental.py
```diff
@@ -0,0 +1,277 @@
...
+
+    _xmllib: Any = Field(default=None)
+
+    @root_validator()
+    def check_xml_lib(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        try:
+            # do this as an optional dep for temporary nature of this feature
+            import defusedxml.ElementTree as DET  # type: ignore
+
+            values["_xmllib"] = DET
```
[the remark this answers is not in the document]
efriis: Thought it was a better dx to fail on creating the object like community ones rather than on first invoke. Happy to adjust though

libs/partners/anthropic/langchain_anthropic/experimental.py
jacoblee93: I think experimental is better in this case since we're not keeping it long term - just until official support lands

PR #18564, opened 2024-03-05
Remove neo4j bloom labels from graph schema

tomasonjo:
Neo4j tools use particular node labels and relationship types to store metadata, but are irrelevant for text2cypher or graph generation, so we want to ignore them in the schema representation.


jexp (approved):
LGTM

hwchase17 (approved):

PR #18581, opened 2024-03-05
Add precedence for input params over env variables in neo4j integration

tomasonjo:
input parameters take precedence over env variables


jexp (approved):
Thanks for addressing this so quickly, it caused quite a number of issues that invisible env variables took precedence over explicit parameters passed in to the constructors.

hwchase17 (approved):

PR #18587, opened 2024-03-05
elasticsearch: add `ElasticsearchRetriever`

maxjakob:
Implement [Retriever](https://python.langchain.com/docs/modules/data_connection/retrievers/) interface for Elasticsearch.

I opted to only expose the `body`, which gives you full flexibility, and none the other 68 arguments of the [search method](https://elasticsearch-py.readthedocs.io/en/v8.12.1/api/elasticsearch.html#elasticsearch.Elasticsearch.search).

Added a user agent header for usage tracking in Elastic Cloud.


maxjakob:
CC @joemcelroy

libs/partners/elasticsearch/pyproject.toml
```diff
@@ -29,6 +29,7 @@ pytest-asyncio = "^0.21.1"
 langchain = { path = "../../langchain", develop = true }
 langchain-community = { path = "../../community", develop = true }
```
efriis, line 30: does the factoring out of text splitters allow us to remove either of these?

efriis (approved):

libs/partners/elasticsearch/pyproject.toml
maxjakob: Indeed, community can be removed here since it's pulled in via `langchain` anyways. Added a PR: https://github.com/langchain-ai/langchain/pull/18644

(We need `langchain` for the chat history test.)
https://github.com/langchain-ai/langchain/blob/81985b31e6ff839ba073d1107722e1756d30e678/libs/partners/elasticsearch/tests/integration_tests/test_chat_history.py#L7

Issue #16403, opened 2024-01-22
Expose complete response metadata from chat model via .invoke/.batch/.stream


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Impossible to access `system_fingerprint` from OpenAI responses.

see: https://github.com/langchain-ai/langchain/discussions/13170#discussioncomment-8211745

eyurtsev:
cc @baskaryan

eyurtsev:
Another discussion: https://github.com/langchain-ai/langchain/discussions/16030

Keiku:
@eyurtsev Is there an update here? I'm having trouble with the lack of reproducibility of the output.

alex-ber:
Inspired by https://github.com/langchain-ai/langchain/discussions/16030#discussioncomment-8213596

```
class OpenAICallbackHandler:
    """Callback Handler that tracks OpenAI info."""

    total_tokens: int = 0
    prompt_tokens: int = 0
    completion_tokens: int = 0
    successful_requests: int = 0
    total_cost: float = 0.0
    system_fingerprint: str = ""
    
    def __init__(self) -> None:
        super().__init__()
        self._lock = threading.Lock()
        
         def __repr__(self) -> str:
            return (
            f"Tokens Used: {self.total_tokens}\n"
            f"\tPrompt Tokens: {self.prompt_tokens}\n"
            f"\tCompletion Tokens: {self.completion_tokens}\n"
            f"Successful Requests: {self.successful_requests}\n"
            f"Total Cost (USD): ${self.total_cost}"\n
           "f"system_filngerprint is {self.system_filngerprint}"
        )
        
        @property
    def always_verbose(self) -> bool:
        """Whether to call verbose callbacks even if verbose is False."""
        return True

    def on_llm_start(
        self, serialized: Dict[str, Any], prompts: List[str], **kwargs: Any
    ) -> None:
        """Print out the prompts."""
        pass

    def on_llm_new_token(self, token: str, **kwargs: Any) -> None:
        """Print out the token."""
        pass

    def on_llm_end(self, response: LLMResult, **kwargs: Any) -> None:
        """Collect token usage."""
        if response.llm_output is None:
            return None

        if "token_usage" not in response.llm_output:
            with self._lock:
                self.successful_requests += 1
            return None

        # compute tokens and cost for this request
        token_usage = response.llm_output["token_usage"]
[...]

PR #18592, opened 2024-03-05
core[patch]: generation info on msg

baskaryan:
related to #16403 #17188


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -234,6 +235,7 @@ def stream(
                 for chunk in self._stream(
                     messages, stop=stop, run_manager=run_manager, **kwargs
                 ):
+                    _cp_gen_info_to_msg(chunk)
```
hinthornw, line 238: I would find it more intuiitve if we didn't mutate in thefunction and instead did

```
chunk.message.additional_kwargs = _merge_gen_info(chunk)
```
or something (instead of mutating)

hwchase17: tend to agree

hinthornw:
🚀

PR #18602, opened 2024-03-05
core[minor]: Add utility code to create tool examples

hwchase17:


libs/core/langchain_core/utils/tool_examples.py
```diff
@@ -0,0 +1,111 @@
+import uuid
```
eyurtsev, line 1: Look good. Do we think it's best to create tool_examples or put into function_calling module?

hwchase17: ya moved

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -311,3 +318,104 @@ def convert_to_openai_tool(
...
+    """A representation of an example consisting of text input and expected tool calls.
+
+    For extraction, the tool calls are represented as instances of pydantic model.
+    """
+
+    input: str  # This is the example text
+    tool_calls: List[BaseModel]  # Instances of pydantic model that should be extracted
+
+
+def tool_example_to_messages(example: Example) -> List[BaseMessage]:
```
baskaryan, line 333: does input need to be dict, why not just have `input` and `tool_calls` as args

hwchase17: idk @eyurtsev any thoughts?

eyurtsev: yeah this can be unpacked

PR #18606, opened 2024-03-05
docs: fix `streamlit` provider 

leo-gan:
There is a wrong python package import.
Fixed it.


docs/docs/integrations/providers/streamlit.mdx
baskaryan, line 1: cc @sfc-gh-jcarroll

sfc-gh-jcarroll (approved):

docs/docs/integrations/providers/streamlit.mdx
leo-gan: CC @baskaryan

PR #18607, opened 2024-03-05
community: Databricks SerDe uses cloudpickle instead of pickle

liangz1:
- **Description:** Databricks SerDe uses cloudpickle instead of pickle when serializing a user-defined function transform_input_fn since pickle does not support functions defined in `__main__`, and cloudpickle supports this.
- **Dependencies:** cloudpickle>=2.0.0

Added a unit test.

Additional guidelines:


libs/community/langchain_community/llms/databricks.py
```diff
@@ -225,7 +224,9 @@ def _is_hex_string(data: str) -> bool:
 def _load_pickled_fn_from_hex_string(data: str) -> Callable:
     """Loads a pickled function from a hexadecimal string."""
     try:
-        return pickle.loads(bytes.fromhex(data))
+        import cloudpickle
+
+        return cloudpickle.loads(bytes.fromhex(data))
     except Exception as e:
```
baskaryan, line 230: may want to separately catch import errors so you can provide a helpful error message (e.g. "please install cloudpickle with ...")

baskaryan (approved):

PR #18610, opened 2024-03-05
docs: `providers` update 6

leo-gan:
Cleaned up the `Integrations/Components/Memory` navbar by shortening the page titles. Updated page titles and file names to consistent formats.


hwchase17 (approved):
tjanks!

PR #18613, opened 2024-03-05
core: fix repr handling of recursive object references

dOrgJelli:
**Description:** Generic serialization of objects via `repr` does not account for recursive object references. This leads to issues when 3rd parties extend Langchain classes and add properties which contain circular references ([example](https://github.com/joaomdmoura/crewAI/blob/e51b8aadae20792882db152fee302378b99580c8/src/crewai/agents/executor.py#L19)). In these cases, a `RecursionError` will be thrown when the `sys.getrecursionlimit()` is reached. However a very difficult to debug `Segmentation Fault` assertion occurs if `sys.setrecursionlimit(...)` is set too high, and the process runs out of memory. Using a `safe_repr` implementation that throws an exception early upon finding a circular reference fixes this problem.

**Twitter handle:** @polywrap_io @dorgjelli @nerfzael

### DIFF
--- a/libs/core/langchain_core/load/safe_repr.py
+++ b/libs/core/langchain_core/load/safe_repr.py
@@ -1,0 +1,43 @@
+def safe_repr(obj, seen=None):
+    """A repr implementation that accounts for recursive object references."""
+
+    if seen is None:
+        seen = set()
+
+    obj_id = id(obj)
+    if obj_id in seen:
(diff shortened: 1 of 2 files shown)


baskaryan:
@dOrgJelli could we add a unit test?

hwchase17:
closing due to inactivity

PR #18633, opened 2024-03-06
partners: AI21 Labs Batch Support in Embeddings

Josephasafg:
Description: Added support for batching when using AI21 Embeddings model


Josephasafg:
@efriis Changed model name in `test_llm.py` to avoid rate limits

efriis (approved):
Going to rename some "chunk" names to "batch"/splitting for clarity!

PR #18637, opened 2024-03-06
partners[openai]: update docstring of OpenAI, OpenAIEmbeddings and ChatOpenAI classes

sepiatone:
**Description:** An update to the docstring of OpenAI, OpenAIEmbeddings and ChatOpenAI classes
**Issue:** Not applicable
**Dependencies:** None


libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -38,22 +38,21 @@
 
 
 class OpenAIEmbeddings(BaseModel, Embeddings):
-    """OpenAI embedding models.
+    """OpenAI embedding model.
 
-    To use, you should have the
-    environment variable ``OPENAI_API_KEY`` set with your API key or pass it
-    as a named parameter to the constructor.
+    To use, you should have the packages ``openai`` and ``langchain-openai``
```
baskaryan, line 43: since this is in langchain-openai don't think we need to specify that that needs to be installed (and langchain-openai already depends on openai, so that will also be automatically installed)

sepiatone: Yes I agree. It's superfluous to include this information (the short example usage code also make this clear).

sepiatone:
@baskaryan 

The suggested change has been made, please review.

If this is fine, please also look at similar pull requests  langchain-ai/langchain#18649 and  langchain-ai/langchain#18645.

PR #18666, opened 2024-03-06
community: Fix the error of Baidu Qianfan not passing the stop parameter

ligang-super:
- [x] **PR title**: "community: fix baidu qianfan missing stop parameter"
- [x] **PR message**:
    - **Description: Baidu Qianfan lost the stop parameter when requesting service due to extracting it from kwargs. This bug can cause the agent to receive incorrect results


.github/workflows/_integration_test.yml
```diff
@@ -47,7 +47,20 @@ jobs:
         id: 'auth'
         uses: google-github-actions/auth@v2
         with:
-          credentials_json: '${{ secrets.GOOGLE_CREDENTIALS }}'
+          credentials_json: '{
```
baskaryan, line 50: guessing this was not meant to be checked in, probably want to void these credentials

baskaryan (approved):

PR #18681, opened 2024-03-06
core: add missing async to _astream definition

n1lanjan:
- **Description:** The abstract method _astream appears to be wrongly defined without async
  - **Issue:** The abstract method _astream appears to be wrongly defined as non-async even though it returns an AsyncIterator. This is throwing pylint error for subclasses - `Method '_astream' was expected to be 'non-async', found it instead as 'async'`
  - **Dependencies:** None


eyurtsev:
I'll review -- i remember trying to make this change, but there was some issue -- i think it's a sync function that returns an async iterator which is fine

eyurtsev (changes requested):
OK so this is not an awaitable. it's a sync function that returns an async iterator. 

Expected behavior is:


```async for chunk in self._astream(..)```

rather than

```async for chunk in (await self._astream(...))```

PR #18689, opened 2024-03-06
docs: update imports of `memory` to use `langchain_community`

leo-gan:
Refactored imports from `langchain` to `langchain_community` whenever it is applicable

### DIFF
--- a/docs/docs/integrations/memory/astradb_chat_message_history.ipynb
+++ b/docs/docs/integrations/memory/astradb_chat_message_history.ipynb
@@ -88,7 +88,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from langchain.memory import AstraDBChatMessageHistory\n",
+    "from langchain_community.chat_message_histories.astradb import AstraDBChatMessageHistory\n",
     "\n",
     "message_history = AstraDBChatMessageHistory(\n",
     "    session_id=\"test-session\",\n",
--- a/docs/docs/integrations/memory/astradb_chat_message_history.ipynb
+++ b/docs/docs/integrations/memory/astradb_chat_message_history.ipynb
@@ -88,7 +88,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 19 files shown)


eyurtsev:
cc @baskaryan do we not recommend using the flat namespace in langchain community for convenience? are we worried about parsing times?

eyurtsev:
@leo-gan let's double check with @baskaryan. My suspicion is that we were worried about the issue being with `langchain` since that's a main entry point and had import times of 1.5-2 seconds (which is too slow for serverless environments). It's much less of an issue these days since folks can easily optimize by using dedicated party integrations + core + langchain

The remaining change that we need to make is to make langchain stop depending on community

leo-gan:
@eyurtsev I don't use the flattened namespace. Should it be a problem?
Most of the refactoring is moving dependencies from `langchain` to `langchain_community`. The latter dependencies are not flattened. I thought the flattened namespaces were a problem in the old `langchain` because of the large number of unrelated integrations under these flattened ns. Is that correct?

leo-gan:
@eyurtsev 
> The remaining change that we need to make is to make langchain stop depending on community
I didn't find any dependency in the current `langchain` code base. Only references to the moved classes. Do you mean something else?

leo-gan:
@eyurtsev I flattened the imports. 
CC @baskaryan 
TODO The current flattening method in `memory/__init__.py` is not ready yet!

eyurtsev (approved):

eyurtsev:
```
[tool.poetry.dependencies]
python = ">=3.8.1,<4.0"
langchain-core = ">=0.1.29,<0.2"
langchain-text-splitters = ">=0.0.1,<0.1"
langchain-community = ">=0.0.25,<0.1"
```

poetry file still has langchain-community in it, so something likely requires it? (Maybe I'm wrong i haven't worked on the migration much)

If it's just the __init__ files we could probably turn langchain-community into an optional dependency

PR #18690, opened 2024-03-06
core[minor]: in Document Loader If load() has been overridden, use it in default lazy_load()

cbornet:


eyurtsev:
Looks good -- do you want me to comandeer or will you resolve the change?

cbornet:
I rebased.
I also moved the test for the doc loader interface from community to core.

eyurtsev (approved):

PR #18693, opened 2024-03-06
community[patch]: chat hf typing fix

baskaryan:


aymeric-roucher:
@baskaryan thank you! This works on all 3 classes!

PR #18707, opened 2024-03-07
docs: update imports of `retrievers` to use `langchain_community`

leo-gan:
Updated `langchain` imports to `langchain_community`.

### DIFF
--- a/docs/docs/integrations/retrievers/amazon_kendra_retriever.ipynb
+++ b/docs/docs/integrations/retrievers/amazon_kendra_retriever.ipynb
@@ -35,7 +35,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from langchain.retrievers import AmazonKendraRetriever"
+    "from langchain_community.retrievers import AmazonKendraRetriever"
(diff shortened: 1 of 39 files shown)


eyurtsev:
@baskaryan same question as before do we want to use flat namespaces in langchain_community or more nested ones? (I think we prefer the former, but wanted to confirm)

docs/docs/integrations/retrievers/amazon_kendra_retriever.ipynb
baskaryan, line 1: as per @eyurtsev's suggestion could we make all of these `from langchain_community.retrievers import ...`

leo-gan: @baskaryan  I could do this, but flattening will import the whole `documen_loaders`, `document_transformation` modules not just a single integration class! I've double checked it. Now, only `LLMs`, `tools`, `vectorstores`, and `utilities` import a selected class only.

eyurtsev (approved):

PR #18713, opened 2024-03-07
chat menu item as icon

leo-gan:
Controversial and opinionated:
It replaces the "Chat" top-menu item with the 💬 icon.
Reasons: 
- Right part of the top-menu is composed of icons, so, another icon looks organic.
- "Chat" vs. 💬 It is not 100% clear for both, what they mean.
- 💬 can be more interesting to click / research / try for users
- 💬 is more compact and more "phone-friendly"
- 💬 looks good with both themes (Day and Night).

It would be nice to try 💬 for a week and compare its "clickability" with the "Chat".


leo-gan:
Something wrong in merging. I'm closing it

PR #18714, opened 2024-03-07
WIP: Get rid of deprecated warnings originating from internal code

eyurtsev:
Users now see a lot of deprecation warnings that originate from internal code.

This PR proposes a solution for that. A lot of deprecation warnings = thousands of them.

This also makes it to hard to find out why our tests fail on CI since it
requires scrolling through a large wall of warnings before actually seeing the
failure.

What's going to happen is that users will effectively filter out all the
warnings in their code, so they won't be getting the deprecation warnings
anyway!

---

At least from a quick check it seems to result in desired behavior in user code

```bash
(langchain_3_11) ➜  libs git:(eugene/stackframe) ✗ time python -c "from langchain.chat_models import ChatOpenAI"
/home/eugene/src/langchain/libs/langchain/langchain/chat_models/__init__.py:30: LangChainDeprecationWarning: Importing chat models from langchain is deprecated. Importing from langchain will no longer be supported as of langchain==0.2.0. Please import from langchain-community instead:

`from langchain_community.chat_models import ChatOpenAI`.

To install langchain-community run `pip install -U langchain-community`.
  caller_aware_warn(
python -c "from langchain.chat_models import ChatOpenAI"  0.98s user 1.31s system 329% cpu 0.696 total
```

### DIFF
--- a/libs/core/langchain_core/_api/deprecation.py
+++ b/libs/core/langchain_core/_api/deprecation.py
@@ -13,6 +13,7 @@ https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/_api/deprecati
 import contextlib
 import functools
 import inspect
+import os
 import warnings
 from typing import Any, Callable, Generator, Type, TypeVar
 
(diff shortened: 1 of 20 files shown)


libs/core/langchain_core/_api/deprecation.py
```diff
@@ -27,6 +28,45 @@ class LangChainPendingDeprecationWarning(PendingDeprecationWarning):
     """A class for issuing deprecation warnings for LangChain users."""
 
 
+def is_warning_internal() -> bool:
```
eyurtsev, line 31: This is the snippet of code to review

eyurtsev:
i am comically bad at linting

eyurtsev:
Looks like nuno implemented the functionality already -- I'll refine this a bit later to make it behave better with our own unit tests

PR #18725, opened 2024-03-07
text_splitter: Added Semantic text splitter for better performance and more option of splitter to user

rgupta2508:
Adding 
    SemanticCharacterTextSplitter
    SemanticTiktokenTextSplitter
There was potential for better performance as well as better semantic chunking.


libs/text-splitters/langchain_text_splitters/character.py
```diff
@@ -577,3 +577,61 @@ def get_separators_for_language(language: Language) -> List[str]:
...
                 f"Please choose from {list(Language)}"
             )
+
+
+class SemanticTiktokenTextSplitter(TextSplitter):
+    """Splitting text that looks at characters."""
+
+    def __init__(
+        self,
+        max_characters: int = 200,
```
eyurtsev, line 587:
```suggestion
        *,
        max_characters: int = 200,
```

libs/text-splitters/langchain_text_splitters/character.py
```diff
@@ -577,3 +577,61 @@ def get_separators_for_language(language: Language) -> List[str]:
...
+        # Optionally can also have the splitter not trim whitespace for you
+        splitter = TiktokenTextSplitter(self.model_name, trim_chunks=self.trim_chunks)
+        return splitter.chunks(text, self.max_characters)
+
+
+class SemanticCharacterTextSplitter(TextSplitter):
+    """Splitting text that looks at characters."""
+
+    def __init__(
+        self, max_characters: int = 200, trim_chunks: bool = False, **kwargs: Any
```
eyurtsev, line 618:
```suggestion
        self, *, max_characters: int = 200, trim_chunks: bool = False, **kwargs: Any
```

docs/api_reference/requirements.txt
```diff
@@ -14,4 +14,4 @@ sphinx-panels
 toml
 myst_nb
 sphinx_copybutton
-pydata-sphinx-theme==0.13.1
\ No newline at end of file
+pydata-sphinx-theme==0.13.1
```
eyurtsev, line 17: accidental modification

libs/text-splitters/langchain_text_splitters/character.py
```diff
@@ -577,3 +577,61 @@ def get_separators_for_language(language: Language) -> List[str]:
...
+
+    def split_text(self, text: str) -> List[str]:
+        """Split incoming text and return chunks."""
+        # First we naively split the large input into a bunch of smaller ones.
+        try:
+            from semantic_text_splitter import TiktokenTextSplitter
+        except ImportError as e:
+            raise ImportError(
+                "Unable to import from semantic-text-splitter. Please specify "
+                "text_splitter or install semantic-text-splitter with "
```
eyurtsev, line 606: what does it mean to specify a text_splitter in the error message?

libs/text-splitters/pyproject.toml
```diff
@@ -12,6 +12,7 @@ repository = "https://github.com/langchain-ai/langchain"
 python = ">=3.8.1,<4.0"
 langchain-core = "^0.1.28"
 lxml = {version = "^5.1.0", optional = true}
+semantic-text-splitter = "^0.6.3"
```
eyurtsev, line 15: @baskaryan what do you think optionals here?

rgupta2508: Updated as true

libs/text-splitters/langchain_text_splitters/character.py
rgupta2508: Updated error message

docs/api_reference/requirements.txt
rgupta2508: Not changed

rgupta2508:
One of the test which is not impacted because of this change started failing after merging from main
tests/unit_tests/agents/test_sql.py:14

rgupta2508:
@eyurtsev please review.
Looks good now, need to fix some unrelated lint error, Finally i am able to see green light now :)

rgupta2508:
@eyurtsev  please review.

rgupta2508:
Removed unused, even langchain.textsplitter is also not required  as this Kept for backwards compatibility. and adding new splitter for new version.

libs/text-splitters/langchain_text_splitters/character.py
```diff
@@ -577,3 +577,62 @@ def get_separators_for_language(language: Language) -> List[str]:
                 f"Language {language} is not supported! "
                 f"Please choose from {list(Language)}"
             )
+
+
+class SemanticTiktokenTextSplitter(TextSplitter):
+    """Splitting text that looks at characters."""
```
eyurtsev, line 583: @rgupta2508 Could you include more context both in the PR summary and in the class documentation?

For context, what's useful to know:

1) As a maintainer, what is this package that's being wrapped? Is there a link to it so we can check whether we want to recommend it to users.
2) As a user, how is this text splitter different from other text splitters? Why would a user chose to use this over something else?

Could you include in the doc-strings a URL to the implementation?

libs/text-splitters/langchain_text_splitters/character.py
```diff
@@ -577,3 +577,62 @@ def get_separators_for_language(language: Language) -> List[str]:
...
+
+    def __init__(
+        self,
+        *,
+        max_characters: int = 200,
+        model_name: str = "gpt-3.5-turbo",
```
eyurtsev, line 589: It doesn't look like model_name should be assigned a default value. assigning a default value will make it such that users are more likely to end up with incorrect code because they're not aware there's a configuration parameter

eyurtsev (changes requested):

Issue #18709, opened 2024-03-07
FAISS and Chroma DB use a `DistanceStrategy` of "distance" instead of "distance-squared", which results in negative relevance scores


chrispy-snps:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from __future__ import annotations
from typing import List, cast
from langchain_core.embeddings import Embeddings
import math
import numpy as np

# create a `UserDefinedEmbeddings` class that expects each document's text
# to contain an embedding vector in Python format
class UserDefinedEmbeddings(Embeddings):
    def __init__(
            self,
            normalize: bool = False
            ) -> None:
        """langchain Embedding class that allows vectors to be manually
        specified via each document's text.

        Args:
            normalize: If True, normalize all vectors to a unit length of 1.0"""
        super().__init__()
        self.normalize = normalize

    def embed_documents(self, texts: List[str]) -> List[List[float]]:
        # evaluate texts into Python objects
        vectors = [eval(text) for text in texts]

        # verify we have vectors of consistent dimension
        if not all(isinstance(vector, list) for vector in vectors):
            raise ValueError('All vectors must be lists')
        if any(len(vector) != len(vectors[0]) for vector in vectors):
            raise ValueError('All vectors must have the same dimension')
        if not all((-1 <= value <= 1) for vector in vectors for value in vector):
            raise ValueError('All vectors must be between -1.0 and +1.0')

        def _normalize_vector(vector):
            """normalize the input vector to a unit length of 1.0"""
            magnitude = math.sqrt(np.sum(np.square(vector)))
            normalized_vector = [element / magnitude for element in vector]
            return normalized_vector

        if self.normalize:
            vectors = [_normalize_vector(vector) for vector in vectors]

        # return them
        return cast(List[List[float]], vectors)

    def embed_query(self, text: str) -> List[float]:
        return self.embed_documents([text])[0]

# create a UserDefinedEmbeddings embedder
userdefined_embeddings = UserDefinedEmbeddings(normalize=True)

# create "documents" that will have hardcoded 2-dimension embedding vectors
text_documents = [
    '[ 1, 0]',
    '[ 1, 1]',
    '[ 0, 1]',
[...]

PR #18728, opened 2024-03-07
[Core]: Relevance Score Scaling Issue

keenborder786:

 **Description:**  
    - The normalization of scores for different Distance Strategies sometimes results in values less than 0, which raises the warning that relevance scores should be between 0 and 1. To address this issue, I have modified the relevance score normalization process to ensure that it never produces values less than zero for every Distance Strategy. I am uncertain about the reason behind the normalized score becoming less than zero, but it was observed to occur in the FIAAS Vector store, as highlighted in the reported issue.
   **Issue:**  #18709

### DIFF
--- a/libs/core/langchain_core/vectorstores.py
+++ b/libs/core/langchain_core/vectorstores.py
@@ -198,21 +198,21 @@ class VectorStore(ABC):
         # This function converts the euclidean norm of normalized embeddings
         # (0 is most similar, sqrt(2) most dissimilar)
         # to a similarity function (0 to 1)
-        return max(1.0 - distance / math.sqrt(2),0)
+        return max(1.0 - distance / math.sqrt(2), 0)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


eyurtsev:
This will require a deeper investigation before we can merge. 

I don't think that we actually want to apply max/min to scores since that throws out information unless there's a valid reason why that makes sense to do

keenborder786:
@eyurtsev should I keep the PR opened or close it then?

chrispy-snps:
I agree with @eyurtsev - we should not discard information through truncation. Unexpected values are usually a sign of an incorrect assumption somewhere, and that root cause should be fixed.

For FAISS, the negative relevance scores occur because FAISS returns distance-*squared* values, which Langchain's default distance-to-relevance equation handles incorrectly. I posted the required distance-*squared* remapping function [here](https://github.com/langchain-ai/langchain/issues/18709#issuecomment-2079135016).

Chroma DB also has this issue. I updated #18709 accordingly.

PR #18733, opened 2024-03-07
Add LLM graph transformer

tomasonjo:
Add a class that constructs knowledge graphs based on text using an LLM.


libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -0,0 +1,255 @@
+from typing import Any, List, Optional, Sequence
+
+from langchain.prompts import ChatPromptTemplate
```
baskaryan, line 3: `from langchain_core.prompts import ChatPromptTemplate`

tomasonjo: Fixed, thanks!

OlegBEZb:
Hi @tomasonjo ! Thank you for the great functionality added! Curious if there is any research supporting exactly this solution or it's just based on your experience? Thank you!

PR #18740, opened 2024-03-07
core: improve PydanticOutputParser typing

jnis23:
This PR adds generic typing to `PydanticOutputParser` so we get a typed output from `.parse` instead of `Any`. It should provide a better DX by way of Intellisense and for anyone strictly typing.

Pre-change:

Screenshot 2024-03-07 at 10 22 31 AM

Post-change:

Screenshot 2024-03-07 at 10 26 31 AM

I haven't dug too deep, but I think a similar change could probably be added to `JsonOutputParser` so we don't have to pull up `.parse`.


eyurtsev (approved):

eyurtsev:
Looks good if it passes unit tests we can merge

jnis23:
Great, thank you!

PR #18748, opened 2024-03-07
core[minor]: Chat Models to fallback astream to fallback on sync stream if available

eyurtsev:
Allows all chat models that implement _stream, but not _astream to still have async streaming to work.

Amongst other things this should resolve issues with streaming community model implementations through langserve since langserve is exclusively async.

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -286,7 +286,9 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
             _stream_implementation = self._astream
         elif type(self)._stream is not BaseChatModel._stream:
             # Then stream is implemented, so we can create an async iterator from it
-            _stream_implementation = cast(
+            # The typing is hard to type correctly with mypy here, so we cast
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -100,6 +102,26 @@ async def agenerate_from_stream(
...
 
 
+def _as_async_iterator(sync_iterator: Callable) -> Callable:
+    """Convert a sync iterator into an async iterator."""
+
+    async def _as_sync_iterator(*args: Any, **kwargs: Any) -> AsyncGenerator:
+        iterator = await run_in_executor(None, sync_iterator, *args, **kwargs)
+        done = object()
+        while True:
+            doc = await run_in_executor(
```
baskaryan, line 112: ooc what's doc stand for

baskaryan (approved):

libs/core/langchain_core/language_models/chat_models.py
eyurtsev: do copy pasta = latin for element or item

PR #18775, opened 2024-03-08
Community: Unify Titan Takeoff Integrations and Adding Embedding Support

pjb157:
**Community: Unify Titan Takeoff Integrations and Adding Embedding Support**

 **Description:** 
Titan Takeoff no longer reflects this either of the integrations in the community folder. The two integrations (TitanTakeoffPro and TitanTakeoff) where causing confusion with clients, so have moved code into one place and created an alias for backwards compatibility. Added Takeoff Client python package to do the bulk of the work with the requests, this is because this package is actively updated with new versions of Takeoff. So this integration will be far more robust and will not degrade as badly over time.

**Issue:**
Fixes bugs in the old Titan integrations and unified the code with added unit test converge to avoid future problems. 

**Dependencies:**
Added optional dependency takeoff-client, all imports still work without dependency including the Titan Takeoff classes but just will fail on initialisation if not pip installed takeoff-client 

**Twitter**
@MeryemArik9

Thanks all :)


baskaryan:
cc @hoyungcher

pjb157:
Hey @baskaryan sorry for the delay I was on holiday that last couple of weeks. I am not sure Blake will be responding as he no longer working here at TitanML. I built this on top of his work and have updated both the docs on our side and in this repo. What steps can I take to help get this PR merged? Thanks for your help

baskaryan:
> Hey @baskaryan sorry for the delay I was on holiday that last couple of weeks. I am not sure Blake will be responding as he no longer working here at TitanML. I built this on top of his work and have updated both the docs on our side and in this repo. What steps can I take to help get this PR merged? Thanks for your help

hey @pjb157! seems there's some merge conflicts that need to be resolved

pjb157:
Thanks @baskaryan should be resolved now

baskaryan:
some outstanding lint errors. run

```bash
cd libs/community
make format
make lint
```
to check them locally

pjb157:
Hey @baskaryan thanks I've resolved these linting errors now

libs/community/langchain_community/llms/titan_takeoff_pro.py
baskaryan, line 1: would prefer to mark this as deprecated instead of deleting it to reduce breaking changes, is that possible?

could do this

```python
from langchain_core._api import deprecated

@deprecated(
    since="0.0.32", removal="0.2.0", alternative="langchain_community.llms.TitanTakeoff"
)
class TitanTakeoffPro(...):
  ...

pjb157: Yeah sure my preference was just to map TitanTakeoffPro to the new object TitanTakeoff. Since it's the diversion of the two classes that are confusing our customers that wanna use this langchain integration. The need for the two objects was because we were supporting two versions of Takeoff (pro and open source version). We have converged the two and thus sadly broke the langchain integration that had been written. Since, we've sort of lost backwards compatibility (through our mistakes) I thought the best compromise would be to write a new integration that works with Takeoff now and map both names TitanTakeoffPro and TitanTakeoff to this one object. That way anyone who used the integration before isn't surprised by a missing import error. What do you think?

pjb157: Also I've tried to clear all the lint errors from any files I have changed. I seem locally to be getting a lot from other files in the files community dir which are completely unrelated

PR #18784, opened 2024-03-08
Fix #18782 (Add streaming logic in ChatHuggingFace)

### LINKED ISSUE #18782: Chat HuggingFace model does not send chunked replies when streaming=True.
### Checked other resources

### Example Code

# llms.py of opengpts (https://github.com/langchain-ai/opengpts)

@lru_cache(maxsize=1)
def get_tgi_llm():
    huggingface_hub.login(os.getenv("HUGGINGFACE_TOKEN"))
    llm = HuggingFaceTextGenInference(
        inference_server_url="http://myinferenceserver.com/",
        max_new_tokens=2048,
        top_k=10,
        top_p=0.95,
        typical_p=0.95,
        temperature=0.3,
        repetition_penalty=1.03,
        streaming=True,
        callbacks=[StreamingStdOutCallbackHandler()],
        server_kwargs={
            "headers": {
                "Content-Type": "application/json",
            }
        },
    )
    return ChatHuggingFace(llm=llm, model_id="HuggingFaceH4/zephyr-7b-beta") # setting model_id for using tokenizer

There is no error, but the output is wrong.
# in my log
HumanMessage(content='Hello?')]
opengpts-backend  | [HumanMessage(content='Hello?'), 'Hello']
opengpts-backend  | [HumanMessage(content='Hello?'), 'Hello!']
opengpts-backend  | [HumanMessage(content='Hello?'), 'Hello! How']
...

### Description

I'm trying to use the 'langchain' library to use the TGI model in OpenGPTs.
I expect the model response to look like the following log: 
HumanMessage(content='Hello?')]
opengpts-backend  | [HumanMessage(content='Hello?'), AIMessageChunk(content='Hello')]
opengpts-backend  | [HumanMessage(content='Hello?'), AIMessageChunk(content='Hello!')]
opengpts-backend  | [HumanMessage(content='Hello?'), AIMessageChunk(content='Hello! How')]
However, messages that are not wrapped in AIMessageChunk are logged. This causes the token response in OpenGPTs to not be visible in the chat window.

### System Info

(py312) eunhye1kim@eunhye1kim-400TEA-400SEA:~/git/forPR/opengpts/backend$ pip freeze | grep langchain
langchain==0.1.7
langchain-cli==0.0.21
langchain-community==0.0.20
langchain-core==0.1.27
langchain-experimental==0.0.37
langchain-google-vertexai==0.0.6
langchain-openai==0.0.7
langchain-robocorp==0.0.3

lunara-kim:
- Add functions (_stream, _astream)
- Connect to _generate and _agenerate

- [x] **PR title**: "community: Add streaming logic in ChatHuggingFace"

- [x] **PR message**: ***Delete this entire checklist*** and replace with
    - **Description:** Addition functions (_stream, _astream) and connection to _generate and _agenerate
    - **Issue:** #18782
    - **Dependencies:** none
    - **Twitter handle:** @lunara_x

### DIFF
--- a/libs/community/langchain_community/chat_models/huggingface.py
+++ b/libs/community/langchain_community/chat_models/huggingface.py
@@ -1,5 +1,5 @@
 """Hugging Face Chat Wrapper."""
-from typing import Any, AsyncIterator, List, Iterator, Optional
+from typing import Any, AsyncIterator, Iterator, List, Optional
 
 from langchain_core.callbacks.manager import (
     AsyncCallbackManagerForLLMRun,
@@ -35,7 +35,8 @@ DEFAULT_SYSTEM_PROMPT = """You are a helpful, respectful, and honest assistant."
 
 
 class ChatHuggingFace(BaseChatModel):
-    """Hugging Face LLMs as ChatModels.
+    """
+    Wrapper for using Hugging Face LLM's as ChatModels.
 
     Works with `HuggingFaceTextGenInference`, `HuggingFaceEndpoint`,
     and `HuggingFaceHub` LLMs.
@@ -53,6 +54,7 @@ class ChatHuggingFace(BaseChatModel):
     system_message: SystemMessage = SystemMessage(content=DEFAULT_SYSTEM_PROMPT)
     tokenizer: Any = None
     model_id: Optional[str] = None
+    streaming: bool = False
 
     def __init__(self, **kwargs: Any):
         super().__init__(**kwargs)
@@ -88,8 +90,8 @@ class ChatHuggingFace(BaseChatModel):
     ) -> Iterator[ChatGenerationChunk]:
         request = self._to_chat_prompt(messages)
 
-        for data in self.llm._stream(request, **kwargs):
-            delta = data.text
+        for data in self.llm.stream(request, **kwargs):
+            delta = data
             chunk = ChatGenerationChunk(message=AIMessageChunk(content=delta))
             if run_manager:
                 run_manager.on_llm_new_token(delta, chunk=chunk)
@@ -103,8 +105,8 @@ class ChatHuggingFace(BaseChatModel):
         **kwargs: Any,
     ) -> AsyncIterator[ChatGenerationChunk]:
         request = self._to_chat_prompt(messages)
-        async for data in self.llm._astream(request, **kwargs):  # await 대신 async for 사용
-            delta = data.text
+        async for data in self.llm.astream(request, **kwargs):
+            delta = data
             chunk = ChatGenerationChunk(message=AIMessageChunk(content=delta))
             if run_manager:
                 await run_manager.on_llm_new_token(delta, chunk=chunk)
--- a/libs/community/langchain_community/chat_models/huggingface.py
+++ b/libs/community/langchain_community/chat_models/huggingface.py
@@ -1,19 +1,28 @@
 """Hugging Face Chat Wrapper."""
-
-from typing import Any, List, Optional
+from typing import Any, AsyncIterator, List, Iterator, Optional
 
 from langchain_core.callbacks.manager import (
     AsyncCallbackManagerForLLMRun,
     CallbackManagerForLLMRun,
 )
-from langchain_core.language_models.chat_models import BaseChatModel
+from langchain_core.language_models.chat_models import (
+    BaseChatModel,
+    agenerate_from_stream,
+    generate_from_stream,
+)
 from langchain_core.messages import (
     AIMessage,
+    AIMessageChunk,
     BaseMessage,
     HumanMessage,
     SystemMessage,
 )
-from langchain_core.outputs import ChatGeneration, ChatResult, LLMResult
+from langchain_core.outputs import (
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/chat_models/huggingface.py
```diff
@@ -71,13 +80,50 @@ def validate_llm(cls, values: dict) -> dict:
...
+            yield chunk
+
     def _generate(
         self,
         messages: List[BaseMessage],
         stop: Optional[List[str]] = None,
         run_manager: Optional[CallbackManagerForLLMRun] = None,
         **kwargs: Any,
     ) -> ChatResult:
+        if self.streaming:
```
eyurtsev, line 121: we generally don't want to support self`.streaming ` on _generate going forward -- we want users to use .stream() or .astream() for streaming exclusively

eyurtsev (commented):
Looks reasonable, do you think you'd be able to lint the PR?

baskaryan:
cc @aymeric-roucher

libs/community/langchain_community/chat_models/huggingface.py
eunhye1kim: @eyurtsev 
I've updated the code based on lint recommendations. 😊
However, the discussion about removing self.streaming=True might need further consideration. 😢

Notably, ChatOpenAI supports streaming=True here:
(https://github.com/langchain-ai/langchain/blob/53ac1ebbbccbc16ce57badf08522c6e59256fdfe/libs/community/langchain_community/chat_models/openai.py#L150)

As for LLMChain, it currently requires the addition of streaming support:
(https://github.com/langchain-ai/langchain/blob/53ac1ebbbccbc16ce57badf08522c6e59256fdfe/libs/langchain/langchain/chains/llm.py#L37)

I'll be looking forward to any updates on this topic and am ready to follow up accordingly.

libs/community/langchain_community/chat_models/huggingface.py
```diff
@@ -71,6 +78,37 @@ def validate_llm(cls, values: dict) -> dict:
...
+    def _stream(
+        self,
+        messages: List[BaseMessage],
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> Iterator[ChatGenerationChunk]:
+        request = self._to_chat_prompt(messages)
+
+        for data in self.llm._stream(request, **kwargs):
```
baskaryan, line 90: should this just call self.llm.stream?

eunhye1kim: @baskaryan 

Yes, this is a wrapper for the HuggingFaceTextGenInference, HuggingFace Endpoint, and HuggingFace_Hub models.
So I checked that it is working properly by calling _stream of llm.
For reference, I tested as follows.

(in opengpts.. backend/app/llms.py)

@lru_cache(maxsize=1)
def get_tgi_llm():
    huggingface_hub.login(os.getenv("HUGGINGFACE_TOKEN"))
    llm = HuggingFaceTextGenInference(
        inference_server_url="http://myinferenceserver/",
        max_new_tokens=1024,
        top_k=10,
        top_p=0.95,
        typical_p=0.95,
        temperature=0.3,
        repetition_penalty=1.1,
        streaming=True,
        callbacks=[],
        server_kwargs={
            "headers": {
                "Content-Type": "application/json",
            }
        },
    )
    return ChatHuggingFace(llm=llm, model_id="HuggingFaceH4/zephyr-7b-beta") # setting a model_id for initializing the tokenizer (the another bug)


(in opengpts.. backend/app/agent.py)
def get_agent_executor(
    ...)
    if agent == AgentType.TGI:
        llm = get_gauss_llm()
        system_message = my_prompt
        return get_gauss_agent_executor(
            tools, llm, system_message, interrupt_before_action, CHECKPOINTER
        )
    else:
        raise ValueError("Unexpected agent type")

eunhye1kim: @baskaryan 

After reading the comment you left, I checked not only the class I used, but also the huggingface endpoint and huggingfacehub.
The HuggingFaceEndpoint class had stream-related content, but HuggingFaceHub class only had generate logic without any such streaming logic.
It seems like I need to explicitly specify the type of llm at ChatHuggingFace as Union type in the initialization step, 
but I want to respect the history of changing the type to Any instead of Union.
Instead, I will add streaming support code to the HuggingFaceHub class and check 
whether it works properly and whether it can be integrated with ChatHuggingFace, and then I'll upload the modified code.

baskaryan: @eunhye1kim just mean should we use `self.llm.stream` instead of `self.llm._stream`

eunhye1kim: @baskaryan I agree. So I changed the code. In the case of HuggingFaceHub, I think streaming support logic may not be necessary because it is a deprecated class.

baskaryan (approved):

Issue #18760, opened 2024-03-07
Update tool callbacks to send the actual response from the tool


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Reproducible issue. on_tool_end logs the string representation of the tool. All tool callbacks need to be updated to accommodate

```python
from langchain_core.tools import tool
from langchain_core.documents import Document

def foo(x):
    return {
        'x': 5
    }

@tool
def get_docs(x: int):
    """Hello"""
    return [Document(page_content='hello')]

chain = foo | get_docs

async for event in chain.astream_events({}, version='v1'):
    if event['event'] == 'on_tool_end':
        print(event)
```

eyurtsev:
https://github.com/langchain-ai/langchain/discussions/18694

eyurtsev:
This is a bug in the astream_events API and an enhancement for existing tool callback handlers

keenborder786:
@eyurtsev I am taking this and working on it.

eyurtsev:
@keenborder786 thanks for taking a look at this! The fix is actually more involved since it needs to be backwards compatible with existing callback handlers. I'm going to scope it out today

keenborder786:
@eyurtsev oh okay. Can you please explain a little about the backwards compatibility in this case so I can help you as well. All the existing handlers that are using `on_tool_end` need to work properly??? Correct.

eyurtsev:
Any code that relies on the output from the handlers should not break

keenborder786:
Okay.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

eyurtsev:
I'm looking through to figure out what can be affected. For example, one thing is BaseTracer which is used to power both astream_log and astream_events. If we change the output of the `on_tool_end`, we will fix the output from astream_events, but we also might cause an unexpected breaking change (e.g., if they're relying on those APIs and sending the content from the backend to the frontend the content will suddenly stop serializing).

it's a bit tricky in some of these cases to figure out whether to make a breaking change (and keep the code simple) or make a more involved change so that we don't break user code

eyurtsev:
OK @keenborder786 I think we can proceed with the PR that you made! I'm going to outline a few places to propagate information to so that we break the minimum amount of code

eyurtsev:
PR merged will be available in next release

[...]

PR #18798, opened 2024-03-08
[Core]: On Tool End Obersvation Casting Fix

keenborder786:
Fixes the following issue #18760 raised by @eyurtsev


eyurtsev (commented):
This might break a bunch of user code -- I'm going to scope out how to do this without introducing breaking changes

keenborder786:
okay, please let me know about the scope so I can integrate that within this PR. Thank you. @eyurtsev

eyurtsev:
Keep PR open for now in case we decide to make a breaking here

eyurtsev:
OK I think we can actually do this! It's a breaking change, but not one that I think will affect many users, and likely better to do this than to implement a work-around.

The remaining changes we need to make:

1. `git grep on_tool_end` across the code base to find all tracer/callback handlers
2. update the type signature from `str` to `Any`
3. Add a cast to string inside the handler itself

```python
    def on_tool_end(self, output: str, *, run_id: UUID, **kwargs: Any) -> Run:
        """End a trace for a tool run."""
``

--

We can then follow up with another PR to actually fix astream_events:

1. Add a unit test that includes the snippet of code I posted in the original issue -- goes here: https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/tests/unit_tests/runnables/test_runnable_events.py#L48-L48
2. Remove the cast to str in BaseTracer on_tool_end (that this PR will introduce) https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/tracers/base.py#L507-L507
3. Celebrate!

eyurtsev:
Yeah we should get both sync and async methods good catch @adreo00

eyurtsev:
@keenborder786 let me know if you're OK taking this on or want me to commandeer! Would be nice to merge a fix next week :)

keenborder786:
Ok. Yes I am taking this.

Get Outlook for Android<https://aka.ms/AAb9ysg>

keenborder786:
@eyurtsev please review.

keenborder786:
I will create the second PR for test that you have mentioned.

eyurtsev:
Looks good! On mobile right now, so won't merge until at my laptop later tonight.

eyurtsev:
Had no time over the weekend, taking a look now

Issue #18322, opened 2024-02-29
Breaking pydantic v2 compatibility in output parsers from 0.1.6


Itinte:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
from langchain.output_parsers import PydanticOutputParser
from pydantic import BaseModel


class TestModel(BaseModel):
    test_attribute: str


if __name__ == '__main__':
    pydantic_output_parser = PydanticOutputParser(pydantic_object=TestModel)
    print(pydantic_output_parser)
```

### Error Message and Stack Trace (if applicable)

```
pydantic.v1.error_wrappers.ValidationError: 1 validation error for PydanticOutputParser
pydantic_object
  subclass of BaseModel expected (type=type_error.subclass; expected_class=BaseModel)
```

### Description

Hello,

Starting from version 0.1.6, there arises an issue with existing pydantic v2 `BaseModel` child classes triggering an error. This problem is directly associated with a particular [commit](https://github.com/langchain-ai/langchain/commit/852973d6169fee3e80f3b361453dd14980dd8797#diff-f86ea1cb10fcab7e1a505fab4aad6b8e4ad3fc33128f7d0c74474166c66bb608):

In the file `pydantic.py` (now relocated to core), the attribute `pydantic_object` of the `PydanticOutputParser` class underwent a change from a generic type variable `T = TypeVar("T", bound=BaseModel)` to `Type[BaseModel]`. Consequently, this alteration shifted the validator used in constructors from `any_class_validator` to `make_class_validator`. Thus, all child BaseModel classes are checked whether they are subclass of pydantic v1 `BaseModel`

It worked on 0.1.5.

I am aware that  [Example 2 in the documentation](https://python.langchain.com/docs/guides/pydantic_compatibility) advises against using pydantic v2 BaseModels. 
However, **I find it unfortunate that upgrading to version 0.1.6 breaks compatibility with pydantic v2.**

Thank you.

### System Info

python 3.10
langchain >= 0.1.6
mac Sonoma 14.1.2

navjot-arcs:
[...]

PR #18811, opened 2024-03-08
core[minor]: support pydantic v2 models in PydanticOutputParser

jnis23:
As mentioned in #18322, the current PydanticOutputParser won't work for anyone trying to parse to pydantic v2 models. This PR adds a separate `PydanticV2OutputParser`, as well as a `langchain_core.pydantic_v2` namespace that will fail on import to any projects using pydantic<2. Happy to update the docs for output parsers if this is something we're interesting in adding. 

On a separate note, I also updated `check_pydantic.sh` to detect pydantic imports with leading whitespace and excluded the internal namespaces. That change can be separated into its own PR if needed.


libs/core/langchain_core/output_parsers/pydantic_v2.py
```diff
@@ -0,0 +1,55 @@
+import json
```
eyurtsev, line 1: A bit worried about getting this into the code base. Definitely into `core` as it will create duplicated code in `core`  when it upgrades to pydantic 2

Not sure, but this could potentially create issues when using in a chain or a runnable map together since it'll cause mixing of pydantic v1 and v2 code.

eyurtsev (changes requested):
Will require investigation

eyurtsev:
@jnis23 Apologies for blocking. I think we need to do a bit of investigation first to figure out the pathway for a pydantic v2 migration and how this would interact with it.

jnis23:
@eyurtsev No problem at all. Agreed it feels bad to dupe that much code, but couldn't think of a better way without adding some additional abstraction. Lmk if there is anything I can do to help re: the investigation!

eyurtsev:
@jnis23 we basically have two options: 

1) Wait for full support to pydantic v2
2) get some partial fixes in and check that they don't fail in unexpected ways.

If you want to investigate (2), I'd check that it still works with something like `prompt | model | parser` chain and write a unit test for that. I'm mostly worried about type inference (via OutputType/InputType) ending up and mixing pydnatic v1 with pydantic v2 types (which would crash things), I don't know if it can happen, but would love some code to double check it.

----

In this case, could we not do something like:

```
TBaseModel = TypeVar("TBaseModel", bound=PydanticBaseModel)
```

with PydanticBaseModel being Union[BaseModelv2, BaseModelv1] if pydantic version == 2, and 
PydanticBaseModel = BaseModelv1 if pydantic version == 1?

+ other conditional behavior based on version, so it's still a single interface

eyurtsev:
for

1) spec out full support for pydantic v2 for LangChain (at least in terms of roadmap)

I'll try to check a potential pathway forward for this next week, and will share if it looks promising

jnis23:
> @jnis23 we basically have two options:
> 
>     1. spec out full support for pydantic v2 for LangChain (at least in terms of roadmap)
> 
>     2. get some partial fixes in and check that they don't fail in unexpected ways.
> 
> 
> If you want to investigate (2), I'd check that it still works with something like `prompt | model | parser` chain and write a unit test for that. I'm mostly worried about type inference (via OutputType/InputType) ending up and mixing pydnatic v1 with pydantic v2 types (which would crash things), I don't know if it can happen, but would love some code to double check it.

I'm currently using the v2 parser in a chain without issue so I can add a test for that. Are there supported use cases for outputs getting piped back into other runnables? I can definitely see that being a problem if the runnable depends on v1. If you know of any example off the top of your head, I can start there, otherwise I can dig.

jnis23:
@eyurtsev Was able to get both versions working under the existing `PydanticOutputParser`. Pyright can figure out the typing, but Mypy gets hung up on the type alias being re-declared. I also don't love that we need the conditional checks in `_parse_obj` for both the `BaseModel` and the `ValidationError`, but it should get the job done until a full transition to V2.

If this looks good, I can nuke the V2 class

libs/core/langchain_core/pydantic_v2/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# This hack is done for the following reasons:
+# * Langchain will attempt to remain compatible with both pydantic v1 and v2 since
+#   both dependencies and dependents may be stuck on either version of v1 or v2.
+# * Creating namespaces for pydantic v1 and v2 should allow us to write code that
+#   unambiguously uses either v1 or v2 API.
+# * This change is easier to roll out and roll back.
+
+
+if get_pydantic_major_version() < 2:
```
eyurtsev, line 17: There should be no side effects on import since the import is triggered by default rather than user action

libs/core/langchain_core/pydantic_v2/__init__.py
```diff
@@ -0,0 +1,24 @@
+import warnings
```
eyurtsev, line 1: I'm not convinced that the namespace is needed yet -- could you review the code snippets -- i think some form of either those versions or else local scope imports will work

libs/core/langchain_core/output_parsers/pydantic.py
```diff
@@ -1,34 +1,56 @@
...
-    Attention: To avoid potential compatibility issues, it's recommended to use
-        pydantic <2 or leverage the v1 namespace in pydantic >= 2.
-    """
+    """The pydantic model to parse."""
+
+    def _parse_obj(self, obj: dict) -> TBaseModel:
+        if PYDANTIC_MAJOR_VERSION == 2 and issubclass(
+            self.pydantic_object, V2BaseModel
+        ):
+            try:
```
eyurtsev, line 29: The validation error can be imported here locally -- the import is cached so there's no issue in terms of speed

```python
if VERSION = 2:
   if issubclass(self.pydantic_object, V1):
      .....
       except pydantic.v1.ValidationError:
   else:
       except pydantic.ValidationError:
```

libs/core/langchain_core/output_parsers/pydantic.py
```diff
@@ -1,34 +1,56 @@
...
-from typing import Generic, List, Type, TypeVar
+from typing import Generic, List, Type, TypeVar, Union
 
 from langchain_core.exceptions import OutputParserException
 from langchain_core.output_parsers import JsonOutputParser
 from langchain_core.outputs import Generation
 from langchain_core.pydantic_v1 import BaseModel, ValidationError
+from langchain_core.pydantic_v2 import BaseModel as V2BaseModel
+from langchain_core.pydantic_v2 import ValidationError as V2ValidationError
+from langchain_core.utils.pydantic import PYDANTIC_MAJOR_VERSION
```
eyurtsev, line 11: pydantic_v2 module could probably be eliminated to do something like this instead:

```python
import pydantic

if VERSION = 2:
  PydanticBaseModel = Union[pydantic.BaseModel, pydantic.v1.BaseModel]
else:
 PydanticBaseModel = pydantic.BaseModel
```

eyurtsev:
@jnis23 looks good. you can remove the older version of the code. I added a few suggestions that I think can further reduce the code foot print of the change, let me know if those work

libs/core/langchain_core/output_parsers/pydantic.py
jnis23: @eyurtsev Definitely like this approach better, it's much cleaner. The namespace was to get around the `check_pydantic` script that gets run as part of linting and in CI. If we're fine adding an exclusion for this file and the test, I can push up these changes

eyurtsev: Yeah OK skipping this file or adding some # pydantic: ignore   type filter

jnis23: okay, added support for # pydantic: ignore. Lmk how the rest looks

jnis23:
@eyurtsev ready for you to re-review when you can

eyurtsev (approved):

eyurtsev:
Thank you @jnis23!

eyurtsev:
Hi @jnis23 not sure if you have any cycles to look into this: https://github.com/langchain-ai/langchain/discussions/22669

Potentially related to this PR

PR #18818, opened 2024-03-08
community[patch]: `adapters` import flattening

leo-gan:


eyurtsev (changes requested):
We should not do the namespace collapse in this case -- it's not equivalent since namespacing information is actually lost here -- e.g., `Chat` seems very generic but it's actually some OpenAI thing.

The reason the flattening is OK in that other cases is that it carries with it the identifier to the provider / integration

leo-gan:
OK

PR #18827, opened 2024-03-09
core[patch]: fix arbitrary config keys

efriis:


libs/core/tests/unit_tests/runnables/test_config.py
```diff
@@ -32,3 +33,11 @@ def test_merge_config_callbacks() -> None:
     assert len(merged) == 2
     assert isinstance(merged[0], ConsoleCallbackHandler)
     assert isinstance(merged[1], StreamingStdOutCallbackHandler)
+
+
+def test_config_arbitrary_keys() -> None:
```
efriis, line 38: confirmed this was failing before

nfcampos (approved):

PR #18830, opened 2024-03-09
langchain[patch]: attach hub metadata

efriis:


libs/langchain/langchain/hub.py
```diff
@@ -77,5 +80,14 @@ def pull(
...
     client = _get_client(api_url=api_url, api_key=api_key)
-    resp: str = client.pull(owner_repo_commit)
-    return loads(resp)
+    res_dict = client.pull_repo(owner_repo_commit)
+    obj = loads(json.dumps(res_dict["manifest"]))
+    if isinstance(obj, BasePromptTemplate):
+        if obj.metadata is None:
+            obj.metadata = {}
+        obj.metadata["lc_hub"] = {
+            "owner": res_dict["owner"],
```
hinthornw, line 89: Would probably put this all at the top level so you can filter by it easier

PR #18846, opened 2024-03-09
Switch to md5 for deduplication in neo4j integrations

tomasonjo:
Deduplicate documents using MD5 of the page_content. Also allows for custom deduplication with graph ingestion method by providing metadata id attribute


tomasonjo:
Linter error seems unrelated:

tests/unit_tests/agents/test_sql.py:14: error: Any? not callable  [misc]

PR #18856, opened 2024-03-09
Fix LLM graph transformer default prompt

tomasonjo:
Some LLMs do not allow multiple user messages in sequence.

### DIFF
--- a/libs/experimental/langchain_experimental/graph_transformers/llm.py
+++ b/libs/experimental/langchain_experimental/graph_transformers/llm.py
@@ -51,7 +51,8 @@ default_prompt = ChatPromptTemplate.from_messages(
         ),
         (
             "human",
-            (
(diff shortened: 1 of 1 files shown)


libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -56,10 +56,6 @@
                 "following input: {input}"
             ),
         ),
-        (
-            "human",
-            "Tip: Make sure to answer in the correct format and do not include any ",
```
baskaryan, line 61: should we add this text to the preceding message?

tomasonjo: Sure, with OpenAI it doesn't make a difference, but it might with others.

PR #18857, opened 2024-03-10
langchain: enhance `LocalFileStore` to allow directory/file permissions to be specified

chrispy-snps:
**Description:**
The `LocalFileStore` class can be used to create an on-disk `CacheBackedEmbeddings` cache. However, the default `umask` settings gives file/directory write permissions only to the original user. Once the cache directory is created by the first user, other users cannot write their own cache entries into the directory.

To make the cache usable by multiple users, this pull request updates the `LocalFileStore` constructor to allow the permissions for newly created directories and files to be specified. The specified permissions override the default `umask` values.

For example, when configured as follows:

```python
file_store = LocalFileStore(temp_dir, chmod_dir=0o770, chmod_file=0o660)
```

then "user" and "group" (but not "other") have permissions to access the store, which means:

* Anyone in our group could contribute embeddings to the cache.
* If we implement cache cleanup/eviction in the future, anyone in our group could perform the cleanup.

The default values for the `chmod_dir` and `chmod_file` parameters is `None`, which retains the original behavior of using the default `umask` settings.

**Issue:**
Implements enhancement #18075.

**Testing:**
I updated the `LocalFileStore` unit tests to test the permissions.


chrispy-snps:
@eyurtsev - is this a better implementation than #18077?

libs/langchain/langchain/storage/file_system.py
```diff
@@ -66,6 +77,21 @@ def _get_full_path(self, key: str) -> Path:
...
+    def _mkdir_for_store(self, dir: Path) -> None:
+        """Makes a store directory path (including parents) with specified permissions
+
+        Args:
+            dir: (Path) The store directory to make
+
+        Returns:
+            None
+        """
+        if not dir.exists():
```
eyurtsev, line 89:
```python
In [3]: from pathlib import Path
In [4]: p = Path('foo/bar')
In [5]: p.mkdir?
Signature: p.mkdir(mode=511, parents=False, exist_ok=False)
Docstring: Create a new directory at this given path.
```

Could you rely on pathlib mkdir to simplify the implementation and make it more efficient (the underlying implementation does a similar recursive thing, but with less uneeded calls)

libs/langchain/langchain/storage/file_system.py
```diff
@@ -97,8 +123,10 @@ def mset(self, key_value_pairs: Sequence[Tuple[str, bytes]]) -> None:
         """
         for key, value in key_value_pairs:
             full_path = self._get_full_path(key)
-            full_path.parent.mkdir(parents=True, exist_ok=True)
+            self._mkdir_for_store(full_path.parent)
             full_path.write_bytes(value)
+            if self.chmod_file is not None:
+                os.chmod(full_path, self.chmod_file)
```
eyurtsev, line 129: let's use pathlib since we're already using pathlib everywhere

libs/langchain/langchain/storage/file_system.py
```diff
@@ -36,14 +36,25 @@ class LocalFileStore(ByteStore):
 
     """
 
-    def __init__(self, root_path: Union[str, Path]) -> None:
+    def __init__(
+            self,
+            root_path: Union[str, Path],
```
eyurtsev, line 41: All named arguments with defaults should only appear after `*` to avoid reduce chance of accidental breaking API changes on future refactors

```suggestion
            root_path: Union[str, Path],
            *,
```

libs/langchain/langchain/storage/file_system.py
```diff
@@ -36,14 +36,25 @@ class LocalFileStore(ByteStore):
...
+            chmod_file: Optional[int] = None,
```
eyurtsev, line 42: can we keep the same mode for the file and dir?

libs/langchain/langchain/storage/file_system.py
chrispy-snps: @eyurtsev - the reason I could not use pathlib's `mkdir(mode=...)` parameter here is that the `mode` value is restricted by the global `umask` value.

To apply permissions that are not restricted by the `umask` value, they must be applied using an explicit `os.chmod` call.

In fact, I patterned the `_mkdir_for_store()` code after the `Path.mkdir` code, but with the additional `os.chmod()` call.

Do you want me to add a comment to the docstring mentioning this reason why a dedicated directory-making method is needed?

libs/langchain/langchain/storage/file_system.py
chrispy-snps: Same as above - only `os.chmod()` is not restricted by the `umask` value, which is the purpose for the enhancement (specifying wider permissions for a shared cache).

libs/langchain/langchain/storage/file_system.py
chrispy-snps: @eyurtsev - I added a comment to explain the need for this function.

libs/langchain/langchain/storage/file_system.py
chrispy-snps: @eyurtsev - this is what I originally wanted to do, but the conventions for file and directory permissions are too different. For example, directories have access bits, sticky bits, and other characteristics that do not apply to files. In addition, there are subtle differences between file and directory bits across platforms.

See the `stat.S_*` flag documentation [located here](https://docs.python.org/3/library/stat.html#stat.S_ISUID) for more details.

I thought about trying to do a "best guess" derivation at directory permission bits based on the specified file permission bits, but (1) someone will want something different than this guess, and (2) such hidden-magic derivation goes against the Pythonic way of keeping things explicit and obvious.

libs/langchain/langchain/storage/file_system.py
eyurtsev: Thank you

eyurtsev (approved):

eyurtsev:
updating to fix linting issues

PR #18868, opened 2024-03-10
partners[anthropic]: adding a default chat model for ChatAnthropic

sepiatone:
**Description:** Adding a default chat model for the ChatAnthropic class
**Issue:** No default model set for the ChatAnthropic class
**Dependencies:** None
**Lint and test**: `make format`, `make lint` and `make test` have been run

The chat model was set to `claude-3-sonnet-20240229`, the one with the cheapest cost among the current Claude-3 [models](https://docs.anthropic.com/claude/docs/models-overview#model-comparison).


baskaryan:
not adding a default is intentional, since it'll go out of date at some point and updating the default is a breaking change

sepiatone:
> not adding a default is intentional, since it'll go out of date at some point and updating the default is a breaking change

ok, I'll close this pr.

PR #18877, opened 2024-03-10
core: Avoid hiding the types of `@beta`-decorated functions/classes

ldorigo:
**Description:** 

While not technically incorrect, the TypeVar used for the `@beta` decorator prevented pyright (and thus most vscode users) from correctly seeing the types of functions/classes decorated with `@beta`.

This is in part due to a small bug in pyright (https://github.com/microsoft/pyright/issues/7448 ) - however, the `Type` bound in the typevar `C = TypeVar("C", Type, Callable)` is not doing anything - classes are `Callables` by default, so by my understanding binding to `Type` does not actually provide any more safety - the modified annotation still works correctly for both functions, properties, and classes.


ldorigo:
Update: see upstream issue, the updated annotation is indeed the correct way to annotate this.

ldorigo:
@efriis , said in the PR template to tag you after a few days if no response on the PR :-)

baskaryan:
this is awesome, thanks @ldorigo!

PR #18880, opened 2024-03-10
community: add hugging face text-to-speech inference API

h0rv:
Description: I implemented a tool to use Hugging Face text-to-speech inference API.

Issue: n/a

Dependencies: n/a

Twitter handle: No Twitter, but do have [LinkedIn](https://www.linkedin.com/in/robby-horvath/) lol.


libs/community/langchain_community/tools/audio/base.py
```diff
@@ -0,0 +1,20 @@
+from abc import abstractmethod
+from typing import Optional
+
+from langchain_core.callbacks import CallbackManagerForToolRun
+from langchain_core.tools import BaseTool
+
+
+class AudioTool(BaseTool):
```
baskaryan, line 8: we should wait to add any abstractions till we have a few more examples of what audio tools look like

h0rv: Got it - removed here https://github.com/langchain-ai/langchain/pull/18880/commits/352902bcaa452cb21d1ae8dbd7c023d988929fe0.

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
```diff
@@ -0,0 +1,100 @@
...
+
+    _DEFAULT_OUTPUT_DIR = "tts_output"
+    _DEFAULT_OUTPUT_NAME = "output"
+    _HUGGINGFACE_API_KEY_ENV_NAME = "HUGGINGFACE_API_KEY"
+    _HUGGINGFACE_API_URL_ROOT = "https://api-inference.huggingface.co/models"
+
+    def __init__(
+        self,
+        model: str,
+        format: HuggingFaceSupportedAudioFormat,
```
eyurtsev, line 43:
```suggestion
        format: HuggingFaceSupportedAudioFormat,
        *,
```

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+        audio_bytes = response.content
+
+        output_name = (
+            f"{output_name or self._default_output_name()}.{self.format.value}"
+        )
+        output_path = os.path.join(self.output_dir, output_name)
+
+        if os.path.exists(output_path):
```
eyurtsev, line 90: This is a race condition. Whenever working with the file system it's generally better to *assume* that it exists, and wrap the code in a try / except that catches the appropriate exception that signifies that a file is missing.

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
```diff
@@ -0,0 +1,100 @@
...
+          or passed to the constructor.
+    """
+
+    name: str = "openai_text_to_speech"
+    description: str = "A wrapper around OpenAI Text-to-Speech API. "
+
+    model: str
+    api_url: str
+    huggingface_api_key: SecretStr
+    format: HuggingFaceSupportedAudioFormat
```
eyurtsev, line 32: let's avoid using the enum here instead use `Literal` and do a run-time check, it saves user the trouble of having to import enums in their code.

format = Literal['wave']

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
```diff
@@ -0,0 +1,100 @@
...
+            },
+            json={"inputs": query},
+        )
+
+        audio_bytes = response.content
+
+        output_name = (
+            f"{output_name or self._default_output_name()}.{self.format.value}"
+        )
+        output_path = os.path.join(self.output_dir, output_name)
```
eyurtsev, line 88: This logic forces the user to reinstantiate the tool every time. I think it would be better to have an option to generate an output_name that's a uuid.uuid4. what do you think about making that the default? Users would be allowed to over-ride to specify a name if they really want

h0rv: I thought it would be good for the user to instantiate the tool with a directory, so it doesn't give flood the directory where its called from, and then output everything there, this way they can define either a tmp directory they will throw out or a place they want to save it. 

I am using a timestamp rather than a UUID if the user does not specify a specific filename. I will switch it to a UUID though.

h0rv: But yeah I will remove the output_dir field and the user can handle that logic by optionally passing in a file name.

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
h0rv: Fixed here: https://github.com/langchain-ai/langchain/pull/18880/commits/d49cf8272c50849153df49a8c7a364c6d511862a

h0rv: I just turned it into a string. The model I was exploring supported only wav, but I believe others support other formats, but wasn't able to find anything in the huggingface docs. So I think it would be easier to let the user specify rather than trying to keep up with supported formats.

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
h0rv: Fixed here https://github.com/langchain-ai/langchain/pull/18880/commits/d49cf8272c50849153df49a8c7a364c6d511862a

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
h0rv: Fixed here: https://github.com/langchain-ai/langchain/pull/18880/commits/d49cf8272c50849153df49a8c7a364c6d511862a

h0rv:
This is ready for another review

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
```diff
@@ -0,0 +1,95 @@
...
+            model=model,
+            file_extension=file_extension,
+            api_url=f"{self._HUGGINGFACE_API_URL_ROOT}/{model}",
+            huggingface_api_key=huggingface_api_key,
+        )
+
+    def _run(
+        self,
+        query: str,
+        output_base_name: Optional[str] = None,
```
eyurtsev, line 67: `output_base_name` should not be parameter for the the tool input -- without proper validation this exposes a security risk and allows an LLM or a malicious user using the LLM to write content anywhere on the file system.

It's OK to specify a containing folder as part of the initializer of the tool. (e.g., `directory`)


```python
HuggingFaceTextToSpeechModelInference(
   destination_dir='...'
)
```

We can also add some additional configuration in the initializer that user to specify what file names are chosen (e.g., timestamp or UUID etc) -- if we want to parameterize this aspect of file naming.

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
```diff
@@ -0,0 +1,95 @@
...
+
+import requests
+from langchain_core.callbacks import CallbackManagerForToolRun
+from langchain_core.pydantic_v1 import SecretStr
+from langchain_core.tools import BaseTool
+
+logger = logging.getLogger(__name__)
+
+
+class HuggingFaceTextToSpeechModelInference(BaseTool):
```
eyurtsev, line 14: @h0rv in which context are you using this tool? Is this meant to be used with an agent?

h0rv: I was going to use the tool to "listen" to chain response as another way to interface (eventually add speech-to-text to have an e2e interface via speech).

Also, I can see it being used at as the output parser for a chain as well.

libs/community/langchain_community/tools/audio/huggingface_text_to_speech_inference.py
h0rv: Added `destination_dir` and added functionality to name the files using uuid4 or timestamps

h0rv: The default will just be uuid4 for as well. Additionally, I made it so each `run` call creates the destination directory if it does not exist.

Is that the right approach, or create it on tool initialization? If the directory gets deleted between init and writing to the output, the file write will fail

eyurtsev (approved):

h0rv:
@eyurtsev Thanks for merging and the review! Good timing for the release of VoiceCraft: [HuggingFace](https://huggingface.co/pyp1/VoiceCraft/tree/main) [Demo](https://jasonppy.github.io/VoiceCraft_web/)

PR #18894, opened 2024-03-11
community: Allow passing `allow_dangerous_deserialization` when loading LLM chain

B-Step62:
### Issue
Recently, the new `allow_dangerous_deserialization` flag was introduced for preventing unsafe model deserialization that relies on pickle without user's notice (#18696). Since then some LLMs like Databricks requires passing in this flag with true to instantiate the model.

However, this breaks existing functionality to loading such LLMs within a chain using `load_chain` method, because the underlying loader function [load_llm_from_config](https://github.com/langchain-ai/langchain/blob/f96dd57501131840b713ed7c2e86cbf1ddc2761f/libs/langchain/langchain/chains/loading.py#L40) 
 (and load_llm) ignores keyword arguments passed in. 

### Solution
This PR fixes this issue by propagating the `allow_dangerous_deserialization` argument to the class loader iff the LLM class has that field.


eyurtsev:
Thank you I think this makes sense. I'll try to review later today

eyurtsev (approved):

B-Step62:
@eyurtsev I've fixed the lint failure, would appreciate if you could take another look. Thanks!

eyurtsev:
thanks @B-Step62

PR #18896, opened 2024-03-11
update core/utils.merge_dicts to support merge of int value

mattwmj:
Description: update util.merge_dicts to support int values
Issue: util.merge_dicts cannot handle int value in dicts such as token_usage.
Dependencies: None
Twitter handle: mattwmj


eyurtsev (changes requested):
The existing code by design treats int as replacement only.

To achieve aggregation logic (or any other kind of logic on top of numeric values), the code that is yielding the streaming dictionaries should take care of aggregating the integers and then assuming that downstream code will replace int values.

We need to document this in the merge dictionary code, since this comes up often (sorry about not having that doc-string there!)

eyurtsev:
Going to close this PR

PR #18922, opened 2024-03-11
community: Add Cohere Tools Agent 

BeatrixCohere:
- Description: Add Cohere tools agent in order to use tools with cohere and example notebook. Also improves the cohere chat model to support the tools parameters.


hwchase17 (commented):
are we going to have a partner library? if so, imo this belongs there - that's what we did with google function calling agent.

efriis (changes requested):
Messaged in slack. Let's do this in the new partner package instead of community!

BeatrixCohere:
> Messaged in slack. Let's do this in the new partner package instead of community!

New PR with tools agent in the partner package: https://github.com/langchain-ai/langchain/pull/19555

PR #18934, opened 2024-03-11
community: Add Neptune Analytics support to NeptuneGraph

bechbd:
**Description:** Add support for Neptune Analytics to the existing NeptuneGraph object
**Twitter handle:** @bechbd`

### DIFF
--- a/docs/docs/integrations/tools/infobip.ipynb
+++ b/docs/docs/integrations/tools/infobip.ipynb
@@ -1,0 +1,176 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Infobip\n",
+    "This notebook that shows how to use [Infobip](https://www.infobip.com/) API wrapper to send SMS messages, emails.\n",
+    "\n",
+    "Infobip provides many services, but this notebook will focus on SMS and Email services. You can find more information about the API and other channels [here](https://www.infobip.com/docs/api)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Setup\n",
+    "\n",
+    "To use this tool you need to have an Infobip account. You can create [free trial account](https://www.infobip.com/docs/essentials/free-trial).\n",
+    "\n",
+    "\n",
+    "`InfobipAPIWrapper` uses name parameters where you can provide credentials:\n",
+    "\n",
+    "- `infobip_api_key` - [API Key](https://www.infobip.com/docs/essentials/api-authentication#api-key-header) that you can find in your [developer tools](https://portal.infobip.com/dev/api-keys)\n",
+    "- `infobip_base_url` - [Base url](https://www.infobip.com/docs/essentials/base-url) for Infobip API. You can use default value `https://api.infobip.com/`.\n",
+    "\n",
+    "You can also provide `infobip_api_key` and `infobip_base_url` as environment variables `INFOBIP_API_KEY` and `INFOBIP_BASE_URL`."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sending a SMS"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "vscode": {
+     "languageId": "plaintext"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "from langchain_community.utilities.infobip import InfobipAPIWrapper\n",
+    "\n",
+    "infobip: InfobipAPIWrapper = InfobipAPIWrapper()\n",
+    "\n",
+    "infobip.run(\n",
+    "    to=\"41793026727\",\n",
+    "    text=\"Hello, World!\",\n",
+    "    sender=\"Langchain\",\n",
+    "    channel=\"sms\",\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sending a Email"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "vscode": {
+     "languageId": "plaintext"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "from langchain_community.utilities.infobip import InfobipAPIWrapper\n",
+    "\n",
+    "infobip: InfobipAPIWrapper = InfobipAPIWrapper()\n",
+    "\n",
+    "infobip.run(\n",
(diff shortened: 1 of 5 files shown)


baskaryan (approved):

libs/community/langchain_community/graphs/neptune_graph.py
```diff
@@ -23,14 +24,15 @@ class NeptuneGraph:
...
+        port: port number for the  Neptune Database instance, default is 8182.  Ignored for Neptune Analytics
+        use_https: whether to use secure connection, default is True.  Ignored for Neptune Analytics
         client: optional boto3 Neptune client
         credentials_profile_name: optional AWS profile name
         region_name: optional AWS region, e.g., us-west-2
-        service: optional service name, default is neptunedata
-        sign: optional, whether to sign the request payload, default is True
+        service: optional service name, default is neptunedata, may also be neptune-graph for Neptune Analytics
+        sign: optional, whether to sign the request payload, default is True.  Ignored for Neptune Analytics
+        graphIdentifier: the graph identifier for a Neptune Analytics graph.  Ignored for Neptune Database
```
3coins, line 35:
```suggestion
        graph_identifier: the graph identifier for a Neptune Analytics graph.  Ignored for Neptune Database
```

libs/community/langchain_community/graphs/neptune_graph.py
```diff
@@ -71,31 +76,46 @@ def __init__(
             else:
                 import boto3
 
+                if service == "neptune-graph":
+                    self.__isNeptuneAnalytics = True
```
3coins, line 80: Should we set this up at the beginning of the block? Also, the rest of the code is using snake case, let’s stick to that.
```python
self.use_analytics_api = service == “neptune-graph”
```

libs/community/langchain_community/graphs/neptune_graph.py
```diff
@@ -71,31 +76,46 @@ def __init__(
...
-                    self.client = session.client(
-                        service,
-                        **client_params,
-                        config=Config(signature_version=UNSIGNED),
-                    )
+                        self.client = session.client(
+                            service,
+                            **client_params,
+                            config=Config(signature_version=UNSIGNED),
+                        )
```
3coins, line 118: We can get rid of the nested block here by returning early.

```python
if service == "neptune-graph":
    self.graph_indentifier = graph_indentifier
    self.client = session.client(service, region_name=region_name)
    return

# remove the else, and continue with the rest of the code..
```

libs/community/langchain_community/graphs/neptune_graph.py
```diff
@@ -104,10 +124,16 @@ def __init__(
...
+                if service == "neptune-graph":
+                    raise ModuleNotFoundError(
+                        "NeptuneGraph requires a boto3 version 1.34.40 or greater."
+                        "Please install it with `pip install -U boto3`."
+                    ) from e
+                else:
+                    raise ModuleNotFoundError(
+                        "NeptuneGraph requires a boto3 version 1.28.38 or greater."
+                        "Please install it with `pip install -U boto3`."
+                    ) from e
```
3coins, line 136: Can simplify this.

```suggestion
                  version = “1.34.40” if service == “neptune_graph” else “1.28.38"                
                  raise ModuleNotFoundError(
                      f"NeptuneGraph requires a boto3 version {version} or greater.”
                      "Please install it with `pip install -U boto3`.”
                  ) from e
```

libs/community/langchain_community/graphs/neptune_graph.py
```diff
@@ -133,7 +159,16 @@ def get_schema(self) -> str:
...
-            return self.client.execute_open_cypher_query(openCypherQuery=query)
+            if self.__isNeptuneAnalytics:
+                resp = self.client.execute_query(
+                    graphIdentifier=self.graphIdentifier,
+                    queryString=query,
+                    parameters=params,
+                    language="OPEN_CYPHER",
+                )
+                return json.loads(resp["payload"].read().decode())
+            else:
```
3coins, line 170: Else is extraneous here.

libs/community/langchain_community/graphs/neptune_graph.py
```diff
@@ -143,21 +178,38 @@ def query(self, query: str, params: dict = {}) -> Dict[str, Any]:
...
+                raise NeptuneQueryException(
+                    {
+                        "message": (
+                            "Summary API is not available for this instance of Neptune,"
+                            "ensure the engine version is >=1.2.1.0"
+                        ),
+                        "details": str(e),
+                    }
+                )
```
3coins, line 206:
```suggestion
            message = "Summary API error occurred on Neptune Analytics” if self.use_analytics_api else (
                "Summary API is not available for this instance of Neptune,”
                "ensure the engine version is >=1.2.1.0”
             ),
            raise NeptuneQueryException(
                {
                    "message": message,
                    "details": str(e),
                }
            ) 
```

libs/community/langchain_community/graphs/neptune_graph.py
```diff
@@ -143,21 +178,38 @@ def query(self, query: str, params: dict = {}) -> Dict[str, Any]:
...
 
         try:
-            summary = response["payload"]["graphSummary"]
+            if self.__isNeptuneAnalytics:
+                summary = response["graphSummary"]
+            else:
+                summary = response["payload"]["graphSummary"]
```
3coins, line 212:
```suggestion
            summary = response["graphSummary”] if self.use_analytics_api else response["payload"]["graphSummary"]
```

3coins (changes requested):
@bechbd 
Thanks for making these changes. I have added some comments to simplify some of the code. 
A high level comment, should we create a new class perhaps `NeptuneGraphAnalytics` to capture the analytics version? It seems like there are conflicting arguments for analytics vs graph and trying to fit everything into one class might be an overkill. Putting this in another class will also simplify the logic, getting rid of the if else blocks.

@krlawrence
What do you think about putting this in another class? We can abstract out some of the common code to avoid duplication. This is a one-way door, since we don’t want users to switch to a new class later on, so we should make this decision now.

bechbd:
Updated the code to split NeptuneGraph and NeptuneAnalyticsGraph into two separate classes, as it will simplify the maintenance.  I created an abstract base class, NeptuneBaseGraph to store the shared functionality between the two classes.  This did require an update of the NeptuneQAChain to check for the base class instead of the specific class.

3coins (commented):
@bechbd 
Great job on doing the refactoring. Left some suggestions about the base class. Also, I think we can put all classes within the same file `neptune_graph.py`, which will make imports simpler.

bechbd:
Thank you for the feedback @3coins , I've made the changes.

3coins:
@bechbd 
Code looks good. Can you rebase and fix the CI errors.

PR #18936, opened 2024-03-11
core[patch]: move some attr/methods to BaseLanguageModel

efriis:
Cleans up some shared code between `BaseLLM` and `BaseChatModel`. One functional difference to make it more consistent (see comment)

### DIFF
--- a/libs/core/langchain_core/language_models/base.py
+++ b/libs/core/langchain_core/language_models/base.py
@@ -7,6 +7,7 @@ from typing import (
     Any,
     Dict,
     List,
+    Mapping,
     Optional,
     Sequence,
     Set,
@@ -25,7 +26,7 @@ from langchain_core.messages import (
     get_buffer_string,
 )
 from langchain_core.prompt_values import PromptValue
-from langchain_core.pydantic_v1 import BaseModel
+from langchain_core.pydantic_v1 import BaseModel, Field, validator
 from langchain_core.runnables import Runnable, RunnableSerializable
 from langchain_core.utils import get_pydantic_field_names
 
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/language_models/base.py
```diff
@@ -71,6 +78,28 @@ class BaseLanguageModel(
...
+    """Whether to print out response text."""
+    callbacks: Callbacks = Field(default=None, exclude=True)
+    """Callbacks to add to the run trace."""
+    tags: Optional[List[str]] = Field(default=None, exclude=True)
+    """Tags to add to the run trace."""
+    metadata: Optional[Dict[str, Any]] = Field(default=None, exclude=True)
+    """Metadata to add to the run trace."""
+
+    @validator("verbose", pre=True, always=True)
+    def set_verbose(cls, verbose: Optional[bool]) -> bool:
```
efriis, line 93: was inconsistent before - this used to only be on llms but not chat model. Will now have same behavior for `verbose=None` for both

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -125,18 +119,8 @@ async def _as_sync_iterator(*args: Any, **kwargs: Any) -> AsyncGenerator:
 class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
     """Base class for Chat models."""
 
-    cache: Optional[bool] = None
-    """Whether to cache the response."""
-    verbose: bool = Field(default_factory=_get_verbosity)
-    """Whether to print out response text."""
-    callbacks: Callbacks = Field(default=None, exclude=True)
-    """Callbacks to add to the run trace."""
     callback_manager: Optional[BaseCallbackManager] = Field(default=None, exclude=True)
```
efriis, line 122: left this in the two places because deprecated and didn't want to add deprecated attr to baselanguagemodel

eyurtsev (approved):

PR #18939, opened 2024-03-11
community: flattening imports 3

leo-gan:
@eyurtsev


libs/community/tests/unit_tests/utilities/test_imports.py
```diff
@@ -24,9 +24,9 @@
     "MaxComputeAPIWrapper",
     "MetaphorSearchAPIWrapper",
     "NasaAPIWrapper",
-    "NVIDIARivaASR",
```
baskaryan, line 27: why the name change?

leo-gan: @baskaryan 
`NVIDIARivaASR`, `NVIDIARivaStream`, `NVIDIARivaTTS` are in __all__. But those are aliases! There was a name replacement in __init__ (AudioStream -> NVIDIARivaStream)!!!! I don't think it is OK. It is a sleepy road, to replace the real class names with aliases [without any explanations]. 
Anyway, you are right, I was too harsh on it. 
I'm keeping the aliases, (but also expose the real class names.)

baskaryan (approved):

leo-gan:
@baskaryan Thanks!

PR #18943, opened 2024-03-11
community: `llms` imports fixes

leo-gan:
Classes are missed in  __all__  and in different places of __init__.py
- BaichuanLLM 
- ChatDatabricks
- ChatMlflow
- Llamafile
- Mlflow
- Together
Added classes to __all__. I also sorted __all__ list.


libs/community/tests/unit_tests/llms/test_imports.py
```diff
@@ -92,11 +97,17 @@
...
 ]
 
 
 def test_all_imports() -> None:
     """Simple test to make sure all things can be imported."""
+    from langchain_core.language_models import BaseChatModel
+
     for cls in llms.__all__:
-        assert issubclass(getattr(llms, cls), BaseLLM)
+        if cls in ["ChatDatabricks", "ChatMlflow"]:
```
efriis, line 109: Hey @leo-gan ! I believe the old test was correct (confirm we're only exporting BaseLLM subclasses from `.llms`) - why are you trying to export chat models from `.llms`?

libs/community/langchain_community/llms/__init__.py
```diff
@@ -631,7 +631,7 @@ def __getattr__(name: str) -> Any:
         return _import_aviary()
     elif name == "AzureMLOnlineEndpoint":
         return _import_azureml_endpoint()
-    elif name == "Baichuan":
+    elif name == "BaichuanLLM":
```
efriis, line 634:
```suggestion
    elif name == "BaichuanLLM" or name == "Baichuan":
```
no need to make this a breaking change if we don't have to

leo-gan: Cool! Great idea. Thanks!

libs/community/tests/unit_tests/llms/test_imports.py
leo-gan: Those two models were not presented in ___all___ . But they are presented in _import_databricks_chat() and _import_mlflow_chat(). They are exposed with get_type_to_cls_dict(). 
So, I have an option to remove _import_databricks_chat and _import_mlflow_chat OR add both classes to __all___. Otherwise, the access symmetry is broken.
Maybe there is a reason for this asymmetry, but if this asymmetry is not commented/documented, I can assume this asymmetry is not intentional.
Those added classes are not subclasses of `BaseLLM` but `BaseChatLLM`.

leo-gan:
@efriis Your comments were addressed.

libs/community/tests/unit_tests/llms/test_imports.py
efriis: The asymmetry is a middle ground to correct a past mistake by deprecating those, but didn't want to make them *more* accessible / advertise they're exported here.

Will remove the changes that don't make sense and merge

efriis (approved):

efriis:
Added a comment clarifying why those ones aren't symmetrical

leo-gan:
@efriis Thank you for the comments and review!

PR #18947, opened 2024-03-12
[core]: add tool calls message

baskaryan:
```python
class ToolCall(TypedDict):
    name: str
    args: Dict[str, Any]
    id: Optional[str]

class InvalidToolCall(TypedDict):
    name: Optional[str]
    args: Optional[str]
    id: Optional[str]
    error: Optional[str]

class ToolCallChunk(TypedDict):
    name: Optional[str]
    args: Optional[str]
    id: Optional[str]
    index: Optional[int]


class AIMessage(BaseMessage):
    ...
    tool_calls: List[ToolCall] = []
    invalid_tool_calls: List[InvalidToolCall] = []
    ...


class AIMessageChunk(AIMessage, BaseMessageChunk):
    ...
    tool_call_chunks: Optional[List[ToolCallChunk]] = None
    ...
```
Important considerations:
- Parsing logic occurs within different providers;
- ~Changing output type is a breaking change for anyone doing explicit type checking;~
- ~Langsmith rendering will need to be updated: https://github.com/langchain-ai/langchainplus/pull/3561~
- ~Langserve will need to be updated~
- Adding chunks:
  - ~AIMessage + ToolCallsMessage = ToolCallsMessage if either has non-null .tool_calls.~
  - Tool call chunks are appended, merging when having equal values of `index`.
  - additional_kwargs accumulate the normal way.
- During streaming:
  - ~Messages can change types (e.g., from AIMessageChunk to AIToolCallsMessageChunk)~
  - Output parsers parse additional_kwargs (during .invoke they read off tool calls).

Packages outside of `partners/`:
- https://github.com/langchain-ai/langchain-cohere/pull/7
- https://github.com/langchain-ai/langchain-google/pull/123/files


baskaryan:
open questions:
- should args be raw or parsed? what to do if parsing fails
- if parsed what are we doing when streaming? are we down doing partial json parsing and accumulation as part of llm stream? what if someone just wants deltas?
- does this even need to be a separate class or could we add tool_calls to AIMessage

libs/core/langchain_core/messages/tool.py
```diff
@@ -1,26 +1,80 @@
...
+            raw_tool_calls = additional_kwargs.get("tool_calls", [])
+            tool_calls = [
+                ToolCall(**tool_call)
+                for tool_call in parse_tool_calls(
+                    raw_tool_calls, partial=True, return_id=True
+                )
+            ]
+
+            return self.__class__(
+                example=self.example,
```
nfcampos, line 55: missing `id` see other classes

libs/core/langchain_core/messages/tool.py
```diff
@@ -1,26 +1,80 @@
...
-    type: Literal["tool"] = "tool"
+class ToolCallsMessage(AIMessage):
+    tool_calls: Optional[List[ToolCall]] = None
+
+
+class ToolCallsMessageChunk(ToolCallsMessage, BaseMessageChunk):
+    # Ignoring mypy re-assignment here since we're overriding the value
+    # to make sure that the chunk variant can be discriminated from the
+    # non-chunk variant.
+    type: Literal["ToolCallsMessageChunk"] = "ToolCallsMessageChunk"  # type: ignore[assignment] # noqa: E501
```
baskaryan, line 28: does ToolCallsMessage also need it's own type to correct ser/des?

libs/core/langchain_core/messages/tool.py
```diff
@@ -1,26 +1,80 @@
...
 
     @classmethod
     def get_lc_namespace(cls) -> List[str]:
         """Get the namespace of the langchain object."""
         return ["langchain", "schema", "messages"]
 
+    def __add__(self, other: Any) -> BaseMessageChunk:  # type: ignore
+        if isinstance(other, ToolCallsMessageChunk):
```
baskaryan, line 36: should we support merging with AIMessageChunk? and should AIMessageChunk support adding ToolCallsMesasgeChunk

baskaryan: ah i guess if other is AIMessageChunk then calling super().__add__ will handle that, but think we still need to update AIMessageChunk.add to handle ToolCallsMessageChunk?

libs/core/langchain_core/output_parsers/json.py
```diff
@@ -44,87 +45,6 @@ def _custom_parser(multiline_string: str) -> str:
     return multiline_string
 
 
-# Adapted from https://github.com/KillianLucas/open-interpreter/blob/5b6080fae1f8c68938a1e4fa8667e3744084ee21/interpreter/utils/parse_partial_json.py
-# MIT License
-def parse_partial_json(s: str, *, strict: bool = False) -> Any:
```
baskaryan, line 49: ooc why move this? output_parsers/json still seems like a reasonable place for it, were there cyclic imports?

libs/core/langchain_core/utils/tools.py
```diff
@@ -0,0 +1,129 @@
...
+            # try removing the last character
+            new_s = new_s[:-1]
+
+    # If we got here, we ran out of characters to remove
+    # and still couldn't parse the string as JSON, so return the parse error
+    # for the original string.
+    return json.loads(s, strict=strict)
+
+
+def parse_tool_calls(
```
baskaryan, line 89: if this is specific to openai tools format should it still live in output_parsers/openai_tools.py? and/or have a name that makes it clear it's openai-specific

libs/core/langchain_core/messages/tool.py
```diff
@@ -1,26 +1,80 @@
...
 
-    tool_call_id: str
-    """Tool call that this message is responding to."""
 
-    type: Literal["tool"] = "tool"
+class ToolCallsMessage(AIMessage):
+    tool_calls: Optional[List[ToolCall]] = None
+
+
+class ToolCallsMessageChunk(ToolCallsMessage, BaseMessageChunk):
```
baskaryan, line 24: should this inherit from AIMessageChunk instead of BaseMessageChunk?

libs/core/langchain_core/messages/tool.py
```diff
@@ -1,26 +1,80 @@
...
+                    "with different example values."
+                )
+
+            additional_kwargs = merge_dicts(
+                self.additional_kwargs, other.additional_kwargs
+            )
+            raw_tool_calls = additional_kwargs.get("tool_calls", [])
+            tool_calls = [
+                ToolCall(**tool_call)
+                for tool_call in parse_tool_calls(
```
baskaryan, line 49: hm so would this only work for openai

baskaryan: i think we may want to drop this and rely on each model handling accumulating tool calls (in our format) itself, and here we just combine tool_calls lists

baskaryan (commented):
bit worried ToolCallsMessage vs ToolMessage will confuse, woudl AIMessageWithToolCalls be a silly name? AIToolCallsMessage?

libs/core/langchain_core/messages/tool.py
ccurme: Thanks, updated this. I moved the new tool calls messages to ai.py to avoid circular imports. Now AIMessage + ToolCallsMessage = ToolCallsMessage and vice-versa.

libs/core/langchain_core/output_parsers/json.py
ccurme: This was to resolve cyclic imports, but since we are no longer parsing tool calls on __add__ (since we'd need to adjust across providers) I moved this back.

libs/core/langchain_core/utils/tools.py
ccurme: Was able to move this back to output_parsers.openai_tools since we are no longer parsing on __add__.

libs/core/langchain_core/messages/tool.py
ccurme: Updated this. For clarity, as of now when chunks are accumulated, `additional_kwargs` gets updated via `utils._merge.merge_dicts` (this may have OpenAI-specific logic in its [references to "index"](https://github.com/langchain-ai/langchain/blob/83f62fdacfa7930c975e2c4aea5733b17e8d6bde/libs/core/langchain_core/utils/_merge.py#L38), not sure if this is standard).

Tool call lists are appended, so after accumulating a few message chunks you do get things like
```
[ToolCall(name='', args={}),
 ToolCall(name='', args={}),
 ToolCall(name='', args={}),
 ToolCall(name='', args={})]
```
in `.tool_calls` in the resulting chunk (whereas the additional_kwargs looks reasonable).

During streaming, output parsers continue to parse the accumulated additional_kwargs. For `ToolCallsMessage` they can read off the parsed .tool_calls.

ccurme:
> bit worried ToolCallsMessage vs ToolMessage will confuse, woudl AIMessageWithToolCalls be a silly name? AIToolCallsMessage?

Updated to AIToolCallsMessage(Chunk).

libs/core/langchain_core/messages/ai.py
```diff
@@ -47,6 +48,92 @@ def __add__(self, other: Any) -> BaseMessageChunk:  # type: ignore
...
+    # Ignoring mypy re-assignment here since we're overriding the value
+    # to make sure that the chunk variant can be discriminated from the
+    # non-chunk variant.
+    type: Literal["AIToolCallsMessageChunk"] = "AIToolCallsMessageChunk"  # type: ignore[assignment] # noqa: E501
+    tool_call_chunks: Optional[List[ToolCallChunk]] = None
+
+    @classmethod
+    def get_lc_namespace(cls) -> List[str]:
+        """Get the namespace of the langchain object."""
+        return ["langchain", "schema", "messages"]
```
baskaryan, line 109: nit: think this is inherited

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -207,6 +238,13 @@ def _convert_delta_to_message_chunk(
         )
     elif role or default_class == ChatMessageChunk:
         return ChatMessageChunk(content=content, role=role, id=id_)
+    elif role or default_class == AIToolCallsMessageChunk:
```
baskaryan, line 241: think we want to check for role == "tool_calls" here

ccurme: From what I've seen we enter this condition at the last chunk of a stream, where `response_metadata={'finish_reason': 'tool_calls'}`. At this point `_dict["tool_calls"]` is None, `default_class == AIToolCallsMessageChunk`, and `role` is None. This was a carry-over from when tool_calls was not optional, so I'll delete it for now. lmk if I'm misunderstanding anything.

libs/core/langchain_core/messages/ai.py
```diff
@@ -47,6 +48,87 @@ def __add__(self, other: Any) -> BaseMessageChunk:  # type: ignore
...
+    tool_calls: Optional[List[ToolCall]] = None
+    type: Literal["tool_calls"] = "tool_calls"  # type: ignore[assignment] # noqa: E501
+
+
+class AIToolCallsMessageChunk(AIToolCallsMessage, AIMessageChunk):
+    # Ignoring mypy re-assignment here since we're overriding the value
+    # to make sure that the chunk variant can be discriminated from the
+    # non-chunk variant.
+    type: Literal["AIToolCallsMessageChunk"] = "AIToolCallsMessageChunk"  # type: ignore[assignment] # noqa: E501
+    tool_call_chunks: Optional[List[ToolCallChunk]] = None
```
ccurme, line 104: @baskaryan one issue with this: it's unclear what expected behavior for `message_chunk_to_message` should be without encoding specific parsing logic. Right now it generates an AIToolCallsMessage with tool_call_chunks. I'm not sure this is a problem though.

baskaryan: could the parsing logic just be decoding json tool_call_chunks -> tool_calls?

baskaryan, line 104: i wonder if this should be Optional[Union[List[ToolCall], List[ToolCallChunk]]]

i imagine there may be some providers down the line that only stream a completed tool call (with args already as a dict), but they will yield them one at a time

ccurme: `AIToolCallsMessageChunk` inherits `tool_calls`, can we put them there? That is what I did for Anthropic.

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -330,7 +333,28 @@ def _stream(
             result = self._generate(
                 messages, stop=stop, run_manager=run_manager, **kwargs
             )
-            yield cast(ChatGenerationChunk, result.generations[0])
+            message = result.generations[0].message
+            if isinstance(message, AIToolCallsMessage):
```
baskaryan, line 337: imo we should leave tool_call_chunks empty in this case

libs/partners/anthropic/langchain_anthropic/output_parsers.py
```diff
@@ -33,11 +26,16 @@ def parse_result(self, result: List[Generation], *, partial: bool = False) -> An
         """
         if not result or not isinstance(result[0], ChatGeneration):
             return None if self.first_tool_only else []
-        tool_calls: List = _extract_tool_calls(result[0].message)
+        message = result[0].message
+        if isinstance(message.content, str):
+            tool_calls: List = []
+        else:
+            content: List = message.content
+            tool_calls = extract_tool_calls(content)
```
baskaryan, line 34: should we convert these to dicts (instead of pydantic objects). would better match our other tool parsers and pydantic objects seem overkill here. also to avoid a breaking change

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -369,6 +414,12 @@ def _format_output(self, data: Any, **kwargs: Any) -> ChatResult:
         }
         if len(content) == 1 and content[0]["type"] == "text":
             msg = AIMessage(content=content[0]["text"])
+        elif any(block["type"] == "tool_use" for block in content):
```
baskaryan, line 417: cc @nfcampos

libs/core/langchain_core/messages/ai.py
ccurme: The args in the chunks are generally substrings of json. If we wanted to do partial parsing we'd need to move the  `parse_partial_json` function.

libs/partners/anthropic/langchain_anthropic/chat_models.py
ccurme: Having every chat model emit tool_call_chunks during streaming provides some uniformity. And this solution will work in a future where Anthropic does stream function calls. But I see how it's wasteful to `.dumps` the parsed args.

libs/partners/anthropic/langchain_anthropic/output_parsers.py
ccurme: fixed this, thanks

libs/core/langchain_core/messages/ai.py
```diff
@@ -47,6 +48,87 @@ def __add__(self, other: Any) -> BaseMessageChunk:  # type: ignore
...
+            )
+
+        return super().__add__(other)
+
+
+class ToolCall(Serializable):
+    name: str
+    args: dict
+    id: Optional[str] = None
+    index: Optional[int] = None
```
ccurme, line 84: @baskaryan how do you feel about removing `index` from ToolCall? We need it on ToolCallChunk for merging but I'm not sure it's needed on ToolCall.

libs/langchain/langchain/output_parsers/json.py
```diff
@@ -1,13 +1,7 @@
 from langchain_core.output_parsers.json import (
     SimpleJsonOutputParser,
-    parse_and_check_json_markdown,
-    parse_json_markdown,
-    parse_partial_json,
 )
 
 __all__ = [
     "SimpleJsonOutputParser",
-    "parse_partial_json",
```
ccurme, line 10: @baskaryan why not export these? isn't this breaking?

baskaryan:
do we need prs in langchain-google and langchain-cohere as well?

ccurme:
> do we need prs in langchain-google and langchain-cohere repos as well?

yes:
- https://github.com/langchain-ai/langchain-google/pull/123
- https://github.com/langchain-ai/langchain-cohere/pull/7

libs/core/langchain_core/messages/tool.py
```diff
@@ -61,3 +63,105 @@ def __add__(self, other: Any) -> BaseMessageChunk:  # type: ignore
             )
 
         return super().__add__(other)
+
+
+class ToolCall(Serializable):
```
jacoblee93, line 68: How will this get serialized? `langchain.messages` or `langchain.messages.tool`?

libs/core/langchain_core/messages/ai.py
```diff
@@ -16,13 +29,36 @@ class AIMessage(BaseMessage):
...
+                "New langchain packages are available that more efficiently handle "
+                "tool calling. Please upgrade your packages to versions that set "
+                "message tool calls. e.g., `pip install --upgrade langchain-anthropic"
+                "`, pip install--upgrade langchain-openai`, etc."
+            )
+        try:
+            if issubclass(cls, AIMessageChunk):  # type: ignore
+                values["tool_call_chunks"] = default_tool_chunk_parser(raw_tool_calls)
+            else:
+                values["tool_calls"] = default_tool_parser(raw_tool_calls)
```
jacoblee93, line 57: Should this check to see if `tool_calls` is passed in on `values`?

libs/core/langchain_core/messages/ai.py
```diff
@@ -35,27 +71,73 @@ class AIMessageChunk(AIMessage, BaseMessageChunk):
...
         return ["langchain", "schema", "messages"]
 
+    @root_validator()
+    def init_tool_calls(cls, values: dict) -> dict:
+        if values["tool_calls"] is not None:
+            raise ValueError(
+                "tool_calls cannot be set on AIMessageChunk, it is derived "
+                "from tool_call_chunks."
+            )
+        if not values["tool_call_chunks"]:
```
jacoblee93, line 89: This is a little bit confusing - why not just always overwrite to either empty array or `None`?

libs/core/langchain_core/messages/ai.py
```diff
@@ -35,27 +71,73 @@ class AIMessageChunk(AIMessage, BaseMessageChunk):
...
+            values["tool_calls"] = values["tool_call_chunks"]
+            return values
+        tool_calls = []
+        for chunk in values["tool_call_chunks"]:
+            try:
+                args_ = parse_partial_json(chunk.args)
+                args_ = args_ if isinstance(args_, dict) else {}
+            except (JSONDecodeError, TypeError):  # None args raise TypeError
+                args_ = {}
+            tool_calls.append(
```
jacoblee93, line 99: Shouldn't we be combining the tool call chunks?

libs/core/langchain_core/messages/ai.py
ccurme: when summing AIMessageChunks you can get chunks with valid tool calls.

ccurme: nvm got your meaning, updated this.

libs/core/langchain_core/messages/ai.py
ccurme: good catch!

libs/core/langchain_core/messages/ai.py
```diff
@@ -16,13 +28,46 @@ class AIMessage(BaseMessage):
...
 
     @classmethod
     def get_lc_namespace(cls) -> List[str]:
         """Get the namespace of the langchain object."""
         return ["langchain", "schema", "messages"]
 
+    @root_validator
+    def _backwards_compat_tool_calls(cls, values: dict) -> dict:
+        raw_tool_calls = values.get("additional_kwargs", {}).get("tool_calls")
+        tool_calls = (
```
jacoblee93, line 46: Naming here is confusing since it's shadowed below

libs/core/langchain_core/messages/ai.py
```diff
@@ -16,13 +28,46 @@ class AIMessage(BaseMessage):
...
+            values.get("tool_calls")
+            or values.get("invalid_tool_calls")
+            or values.get("tool_call_chunks")
+        )
+        if raw_tool_calls and not tool_calls:
+            warnings.warn(
+                "New langchain packages are available that more efficiently handle "
+                "tool calling. Please upgrade your packages to versions that set "
+                "message tool calls. e.g., `pip install --upgrade langchain-anthropic"
+                "`, pip install--upgrade langchain-openai`, etc."
```
jacoblee93, line 56: Missing space

libs/core/langchain_core/messages/ai.py
```diff
@@ -35,27 +80,90 @@ class AIMessageChunk(AIMessage, BaseMessageChunk):
...
     def get_lc_namespace(cls) -> List[str]:
         """Get the namespace of the langchain object."""
         return ["langchain", "schema", "messages"]
 
+    @root_validator()
+    def init_tool_calls(cls, values: dict) -> dict:
+        if not values["tool_call_chunks"]:
+            values["tool_calls"] = []
+            values["invalid_tool_calls"] = []
+            return values
```
jacoblee93, line 96: Small nit - what if `values["tool_calls"]` is explicitly passed?

libs/core/langchain_core/messages/ai.py
```diff
@@ -35,27 +80,90 @@ class AIMessageChunk(AIMessage, BaseMessageChunk):
...
+        tool_calls = []
+        invalid_tool_calls = []
+        for chunk in values["tool_call_chunks"]:
+            try:
+                args_ = parse_partial_json(chunk["args"])
+                if isinstance(args_, dict):
+                    tool_calls.append(
+                        ToolCall(
```
jacoblee93, line 104: Don't you need `index`?

libs/core/langchain_core/output_parsers/openai_tools.py
```diff
@@ -1,13 +1,89 @@
...
+    parsed = {
+        "name": raw_tool_call["function"]["name"] or "",
+        "args": function_args or {},
+    }
+    if return_id:
+        parsed["id"] = raw_tool_call["id"]
+    return parsed
+
+
+def make_invalid_tool_call(
```
jacoblee93, line 51: I don't think this belongs in this file

libs/partners/groq/langchain_groq/chat_models.py
```diff
@@ -278,9 +280,20 @@ def _stream(
...
             generation = chat_result.generations[0]
             message = generation.message
+            tool_call_chunks = [
+                {
+                    "name": rtc["function"].get("name"),
+                    "args": rtc["function"].get("arguments"),
+                    "id": rtc.get("id"),
+                    "index": rtc.get("index"),
+                }
+                for rtc in message.additional_kwargs["tool_calls"]
```
jacoblee93, line 290: nit: pull from `tool_calls` on the `AIMessage`?

PR #18949, opened 2024-03-12
templates[patch]: Update pirate-speak deps, add messages placeholder

jacoblee93:
CC @efriis


efriis (approved):
Core probably good to keep Caret

PR #18954, opened 2024-03-12
community: Add `top_k` parameter with default value 4 to enhance flexibility

reversetracker:
- [x] **PR message**: 
- [x] **Description**: The Retriever can use a different value than the one specified at creation time whenever desired.

### DIFF
--- a/libs/community/langchain_community/retrievers/bm25.py
+++ b/libs/community/langchain_community/retrievers/bm25.py
@@ -97,8 +97,15 @@ class BM25Retriever(BaseRetriever):
         )
 
     def _get_relevant_documents(
-        self, query: str, *, run_manager: CallbackManagerForRetrieverRun
+        self,
+        query: str,
+        *,
+        run_manager: CallbackManagerForRetrieverRun,
+        k: Optional[int] = None,
+        **kwargs: Any,
     ) -> List[Document]:
         processed_query = self.preprocess_func(query)
-        return_docs = self.vectorizer.get_top_n(processed_query, self.docs, n=self.k)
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/retrievers/bm25.py
```diff
@@ -97,8 +97,15 @@ def from_documents(
         )
 
     def _get_relevant_documents(
-        self, query: str, *, run_manager: CallbackManagerForRetrieverRun
+        self,
+        query: str,
+        *,
+        run_manager: CallbackManagerForRetrieverRun,
+        k: Optional[int] = None,
```
isahers1, line 104: I believe this is supposed to be a non-public method, so I think it would only use self.k

libs/community/tests/unit_tests/retrievers/test_bm25.py
```diff
@@ -43,3 +43,21 @@ def test_repr() -> None:
...
+    input_docs = [
+        Document(page_content="I have a pen."),
+        Document(page_content="Do you have a pen?"),
+        Document(page_content="I have a bag."),
+        Document(page_content="Do you have a dog?"),
+        Document(page_content="I have a cat."),
+    ]
+    bm25_retriever = BM25Retriever.from_documents(documents=input_docs)
+
+    documents = bm25_retriever.get_relevant_documents(query="I have a pen")
```
isahers1, line 59: I don't believe this is the right function signature, I think _get_relevant_docs is only meant to be called by the retriever class itself

efriis:
closing due to no response

PR #18961, opened 2024-03-12
added pinecone as a tool to community

Sanjaypranav:

- [x] **Pinecone as a tool**: 
  - "community: Pincone Tool"

- [x] **Lint and test**: linted


cbornet:
I think it would be better to have a generic VectorStore tool.
Note that there is already a [VectorStoreQATool](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/tools/vectorstore/tool.py#L33) which is similar but arguably different as it uses the retrieved documents to query an LLM.

Sanjaypranav:
ok got it

baskaryan (commented):
there is already a `create_retrieval_tool` constructor that i think would get you equivalent functionality (except that user would have to specify tool name and description

```python
from langchain_pinecone import PineconeVectorStore
from langchain.tools.retriever import create_retreiver_tool

retriever = PineconeVectorStore(...).as_retriever()
tool = create_retriever_tool(
  retriever, 
  "pinecone-search", 
  "pinecone-search searches..."
)
```
would that be work for your use case @Sanjaypranav?

Sanjaypranav:
yes it works

baskaryan:
closing per above ^

PR #18987, opened 2024-03-12
langchain-mongodb: Remove in-memory cache from cache abstractions

Jibola:
## Description
* In memory cache easily gets out of sync with the server cache, so we will remove it entirely to reduce the issues around invalidated caches.

## Dependencies
None

- [x]  If you're adding a new integration, please include


Jibola:
cc: @shaneharvey, @blink1073 , @NoahStapp , @caseyclements

Jibola:
> What was the rationale for implementing a local cache?

Having the additional in-memory cache optimized on explicit replica prompts. Unfortunately, things get out of alignment between local and atlas more often than we liked.

libs/partners/mongodb/tests/unit_tests/test_cache.py
```diff
@@ -30,7 +30,6 @@ def __init__(
         self.__database_name = database_name
```
ShaneHarvey, line 30: Can we remove this file (test_cache.py) altogether?

Jibola: We still need this file as it tests the `MongoDBAtlasSemanticCache` and `MongoDBCache` abstractions which both leverage MongoDB as the caching layer.

ShaneHarvey (approved):

Jibola:
cc: @efriis

efriis (approved):

PR #18990, opened 2024-03-12
[Enhancement] Add support for directly providing a run_id

hinthornw:
The root run id (~trace id's) is useful for assigning feedback, but the current recommended approach is to use callbacks to retrieve it, which has some drawbacks:
1. Doesn't work for streaming until after the first event
2. Doesn't let you call other endpoints with the same trace ID in parallel (since you have to wait until the call is completed/started to use

This PR lets you provide = "run_id" in the runnable config.

Couple considerations:

1. For batch calls, we split the trace up into separate trees (to permit better rendering). We keep the provided run ID for the first one and generate a unique one for other elements of the batch.
2. For nested calls, the provided ID is ONLY used on the top root/trace.

### Example Usage

```
chain.invoke("foo", {"run_id": uuid.uuid4()})
```


jacoblee93:
CC @baskaryan @eyurtsev

libs/core/langchain_core/callbacks/manager.py
```diff
@@ -1198,7 +1199,11 @@ def on_llm_start(
         """
         managers = []
         for prompt in prompts:
-            run_id_ = uuid.uuid4()
+            if run_id is not None:
+                run_id_ = run_id
+                run_id = None
+            else:
+                run_id_ = uuid.uuid4()
```
baskaryan, line 1206: this broke my brain a bit, why not

```python
run_id_ = run_id if run_id is not None else uuid.uuid4()
```

hinthornw: no but I could use enumerate

baskaryan (approved):

PR #18991, opened 2024-03-12
core: fix PydanticOutputParser typing in LCEL chains

jnis23:
By subclassing the `JsonOutputParser`, the `PydanticOutputParser` wasn't propagating the typing information from #18740 to the `BaseOutputParser`. This change removes `JsonOutputParser` as a subclass and instead directly subclasses `BaseCumulativeTransformOutputParser` so we get the correct typing when using LCEL.

Before:

Screenshot 2024-03-12 at 4 34 23 PM

After:

Screenshot 2024-03-12 at 4 35 41 PM


jnis23:
@eyurtsev One last typing fix for the pydantic parser when you have time

Undertone0809 (approved):

hwchase17:
would prefer a fix that still mains the inheritance

PR #19023, opened 2024-03-13
community: Add `partition` parameter to DashVector

wangcailin:
**Description**: DashVector Add partition parameter
**Twitter handle**: @CailinWang_

### DIFF
--- a/libs/community/langchain_community/vectorstores/dashvector.py
+++ b/libs/community/langchain_community/vectorstores/dashvector.py
@@ -66,16 +66,22 @@ class DashVector(VectorStore):
         self._embedding = embedding
         self._text_field = text_field
 
+    def _create_partition_if_not_exists(self, partition: str):
+        """Create a Partition in current Collection."""
+        return self._collection.create_partition(partition)
+
     def _similarity_search_with_score_by_vector(
         self,
         embedding: List[float],
         k: int = 4,
         filter: Optional[str] = None,
+        partition: str = "default",
     ) -> List[Tuple[Document, float]]:
         """Return docs most similar to query vector, along with scores"""
 
         # query by vector
-        ret = self._collection.query(embedding, topk=k, filter=filter)
+        ret = self._collection.query(
+            embedding, topk=k, filter=filter, partition=partition)
         if not ret:
             raise ValueError(
                 f"Fail to query docs by vector, error {self._collection.message}"
@@ -86,7 +92,8 @@ class DashVector(VectorStore):
             metadata = doc.fields
             text = metadata.pop(self._text_field)
             score = doc.score
-            docs.append((Document(page_content=text, metadata=metadata), score))
+            docs.append(
+                (Document(page_content=text, metadata=metadata), score))
         return docs
 
     def add_texts(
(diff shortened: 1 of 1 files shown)


wangcailin:
@baskaryan Thanks for the hard work checking it out!

hwchase17 (commented):
does this change behavior? eg what is the behavior if you dont pass partition? is it None? or was it doing default all along?

wangcailin:
> does this change behavior? eg what is the behavior if you dont pass partition? is it None? or was it doing default all along?

Thank you for your insightful questions regarding the partition parameter and its default behavior. Let me address your concerns:

Default Behavior: When the partition parameter is not explicitly passed by the user, our code defaults to using "default" as the partition name. This means all operations (such as add, delete, search, etc.) are performed in the "default" partition unless specified otherwise. This approach ensures that the functionality remains consistent and predictable, even without specifying a partition.

Behavior Without Specifying partition: In the absence of a specific partition parameter, the system utilizes the default "default" partition for all relevant operations. This default setting is intentional to maintain backward compatibility and to ensure that the system operates seamlessly for users who might not be concerned with the partitioning feature.

Changes in Behavior: Introducing the partition parameter with a default value of "default" does not alter the existing behavior of the code. Instead, it introduces an additional layer of flexibility, allowing users to specify a partition if they wish to organize their data more granularly. For users who do not specify this parameter, the system behaves as it would have before the introduction of partitions, ensuring that there is no unexpected change in behavior for existing codebases.

The addition of the partition parameter is designed to enhance user control and code modularity without affecting the default operational behavior. The default "default" partition acts as a fallback, ensuring that operations remain unaffected for users who do not wish to use this feature. This approach maintains backward compatibility while offering additional functionality for users who require it.

I hope this clarifies the purpose and behavior of the partition parameter in our implementation. Please let me know if there are any further questions or concerns.

hwchase17:
> > does this change behavior? eg what is the behavior if you dont pass partition? is it None? or was it doing default all along?
> 
> Thank you for your insightful questions regarding the partition parameter and its default behavior. Let me address your concerns:
> 
> Default Behavior: When the partition parameter is not explicitly passed by the user, our code defaults to using "default" as the partition name. This means all operations (such as add, delete, search, etc.) are performed in the "default" partition unless specified otherwise. This approach ensures that the functionality remains consistent and predictable, even without specifying a partition.
> 
> Behavior Without Specifying partition: In the absence of a specific partition parameter, the system utilizes the default "default" partition for all relevant operations. This default setting is intentional to maintain backward compatibility and to ensure that the system operates seamlessly for users who might not be concerned with the partitioning feature.
> 
> Changes in Behavior: Introducing the partition parameter with a default value of "default" does not alter the existing behavior of the code. Instead, it introduces an additional layer of flexibility, allowing users to specify a partition if they wish to organize their data more granularly. For users who do not specify this parameter, the system behaves as it would have before the introduction of partitions, ensuring that there is no unexpected change in behavior for existing codebases.
> 
> The addition of the partition parameter is designed to enhance user control and code modularity without affecting the default operational behavior. The default "default" partition acts as a fallback, ensuring that operations remain unaffected for users who do not wish to use this feature. This approach maintains backward compatibility while offering additional functionality for users who require it.
> 
> I hope this clarifies the purpose and behavior of the partition parameter in our implementation. Please let me know if there are any further questions or concerns.

it does, thanks!

hwchase17 (approved):

PR #19035, opened 2024-03-13
langchain-mongodb: Remove embedding retrieval from mongodb payload

Jibola:
## Description
Returning the embedding is not necessary in the vector search functionality unless specified as a debugging step. This change defaults the behavior such that the server _only_ returns the embedding key if explicitly requested, such as in the case of `max_marginal_relevance_search`.

* Added `test_from_documents_no_embedding_return`


Jibola:
cc: @shaneharvey, @blink1073 , @NoahStapp , @caseyclements

libs/partners/mongodb/langchain_mongodb/vectorstores.py
```diff
@@ -183,6 +183,8 @@ def _similarity_search_with_score(
         k: int = 4,
         pre_filter: Optional[Dict] = None,
         post_filter_pipeline: Optional[List[Dict]] = None,
+        remove_embedding: bool = True,
```
ShaneHarvey, line 186: I usually find it unnatural to have a optional argument that defaults to True (since usually missing is false). Could we name this "include_embedding" and default to False instead?

ShaneHarvey (changes requested):

efriis (approved):

PR #19036, opened 2024-03-13
core: update _rm_titles to account for title argument name bug

rsk2327:
Issue : For functions which have an argument with the name 'title', the convert_pydantic_to_openai_function generates an incorrect output and omits the argument all together. This is because the _rm_titles function removes all instances of the the key 'title' from the output.

Description : Updates the _rm_titles function to check the presence of the 'type' key as well before removing the 'title' key. As the title key that we wish to omit always has a type key along with it. 

Potential gap if there is a function defined which has both title and key as argument names, in which case this would fail. Maybe we could set a filter on the function argument names and reject those with keyword argument names.

No dependencies. Passed all tests. 

- [x] **PR message**: ***Delete this entire checklist*** and replace with

Additional guidelines:


libs/core/langchain_core/utils/function_calling.py
```diff
@@ -54,7 +54,8 @@ class ToolDescription(TypedDict):
 def _rm_titles(kv: dict) -> dict:
     new_kv = {}
     for k, v in kv.items():
-        if k == "title":
+        # print(kv.keys())
```
eyurtsev, line 57:
```suggestion
```

rsk2327:
@baskaryan Let me know if it makes more sense to set an explicit check on function argument names to raise an error for argument names like 'title' and 'key'. Or if the above solution is good enough

rsk2327:
@baskaryan @eyurtsev Let me know if any changes are required to move forward.

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -54,7 +54,7 @@ class ToolDescription(TypedDict):
 def _rm_titles(kv: dict) -> dict:
     new_kv = {}
     for k, v in kv.items():
-        if k == "title":
+        if k == "title" and "type" in kv.keys():
```
baskaryan, line 57: could we add a unit test for this?

rsk2327:
@baskaryan Added unit tests. 

Also updated the logic for the function to account for the edge case that I had previously mentioned. The new logic should hold irrespective of the use of keywords like 'title', 'type' and 'properties' as Class variable names

rsk2327:
@baskaryan @eyurtsev

hwchase17 (approved):

PR #19043, opened 2024-03-13
Updated the Nebula LLM being used

sharmisthasg:
**Description:** Changed the Nebula model endpoint to point to Nebula Chat instead of Nebula Instruct. Made relevant updates to model input
**Dependencies:** N/A
**Tag maintainer**: @eyurtsev
**Twitter handle**: @Symbldotai

### DIFF
--- a/libs/core/langchain_core/runnables/passthrough.py
+++ b/libs/core/langchain_core/runnables/passthrough.py
@@ -610,8 +610,30 @@ class RunnableAssign(RunnableSerializable[Dict[str, Any], Dict[str, Any]]):
 
 
 class RunnablePick(RunnableSerializable[Dict[str, Any], Dict[str, Any]]):
-    """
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/llms/symblai_nebula.py
```diff
@@ -20,11 +21,10 @@
 from langchain.utils import get_from_dict_or_env
 
 DEFAULT_NEBULA_SERVICE_URL = "https://api-nebula.symbl.ai"
-DEFAULT_NEBULA_SERVICE_PATH = "/v1/model/generate"
+DEFAULT_NEBULA_SERVICE_PATH = "/v1/model/chat"
```
baskaryan, line 24: if there's a chat endpoint that shoudl be implemented as a ChatModel not an LLM

PR #19052, opened 2024-03-14
community: add stop parameter support to volcengine maas

joyme123:
- **Description:** add stop parameter to volcengine maas model
- **Dependencies:** no


libs/community/langchain_community/chat_models/volcengine_maas.py
```diff
@@ -112,6 +112,8 @@ def _stream(
         run_manager: Optional[CallbackManagerForLLMRun] = None,
         **kwargs: Any,
     ) -> Iterator[ChatGenerationChunk]:
+        if "stop" not in kwargs and stop is not None:
```
baskaryan, line 115: "stop" can't be in kwargs if it's an explicit kwarg (so no need to check `if "stop" not in kwargs`)

joyme123: > "stop" can't be in kwargs if it's an explicit kwarg (so no need to check `if "stop" not in kwargs`)

Yes. I fixed it

baskaryan (approved):

PR #19069, opened 2024-03-14
qdrant add should condition filter

chengxs1994:
Description:
Add support for conditions such as should in the _qdrant_filter_from_dict query method of the qdrant.py file class.

Issue:
When I was using _qdrant_filter_from_dict, I found that I couldn't perform multi condition queries through should because it is currently restricted to the must mode and I hope it can be customized as should or must-not.

Dependencies:
NA

### DIFF
--- a/libs/community/langchain_community/vectorstores/qdrant.py
+++ b/libs/community/langchain_community/vectorstores/qdrant.py
@@ -1987,20 +1987,20 @@ class Qdrant(VectorStore):
         return out
 
     def _qdrant_filter_from_dict(
-        self, filter: Optional[DictFilter]
+        self, filter: Optional[DictFilter], condition_type: str = "must"
     ) -> Optional[rest.Filter]:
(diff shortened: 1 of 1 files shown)


eyurtsev (changes requested):
This is a private method so it's not exposed as part of the public API.

To expose filter control we either need to make this public or else create another public API entry point.

With a private method adding an argument is no different from overriding the entire method

chengxs1994:
> This is a private method so it's not exposed as part of the public API.
> 
> To expose filter control we either need to make this public or else create another public API entry point.
> 
> With a private method adding an argument is no different from overriding the entire method

Thank you for your answer. If I want to query through 'should', what public method would be more appropriate to call?

eyurtsev:
Looks like you need to provide qdrant filters directly:

https://github.com/langchain-ai/langchain/blob/main/libs/community/langchain_community/vectorstores/qdrant.py#L596-L596

PR #19076, opened 2024-03-14
langchain[patch]: make hub work with older langchainhub versions

eyurtsev:
Make it work with older clients


libs/langchain/langchain/hub.py
```diff
@@ -80,6 +80,12 @@ def pull(
     :param api_key: The API key to use to authenticate with the LangChain Hub API.
     """
     client = _get_client(api_url=api_url, api_key=api_key)
+
+    if hasattr(client, "pull"):  # <= 0.1.14
```
hinthornw, line 84: It still has 'pull', the check you want is for pull_repo and if that attr isn't present then do pull las a fallback

hinthornw (approved):

Issue #18803, opened 2024-03-08
Add in-code documentation for LangChain Runnables


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

We would love to get help to add in code documentation to LangChain core to better document LCEL primitives:

Here is an example of a documented runnable: 
https://api.python.langchain.com/en/latest/runnables/langchain_core.runnables.base.RunnableLambda.html#langchain_core.runnables.base.RunnableLambda

Here is an example of an undocumented runnable (currently many are undocumented): https://api.python.langchain.com/en/latest/runnables/langchain_core.runnables.passthrough.RunnablePick.html

Acceptance Criteria:

- PR should be as minimal as possible (don't try to document unrelated runnables please!) Keep PR size small to get things merged quicker and avoid merge conflicts.
- Document the class doc-string:
   - include an overview about what the runnable does
   - include a ...code-block: python that shows a self-contained example of how to use the runnable.
   - the self contained example should include all relevant imports so it can be copy pasted AS is and run

How do i figure out what the runnable does? 
- All runnables have unit tests that show how the runnable can be used! You can locate the unit tests and use them as reference.

Some especially import runnables (note that some of these are base abstractions)

- https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/runnables/passthrough.py#L315-L315
- https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/runnables/passthrough.py#L577-L577
- https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/runnables/configurable.py#L44-L44
- https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/runnables/configurable.py#L222-L222
- Context: https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/beta/runnables/context.py#L309-L309

devesh-2002:
May I take this issue?

aanaseer:
I'll take on documenting `runnables.passthrough.RunnableAssign`.

aaronjimv:
[...]

PR #19079, opened 2024-03-14
core: Updated docstring for Context class

aaronjimv:
- **Description:** Improves the docstring for `class Context` by providing an overview and an example.
- **Issue:** #18803


aaronjimv:
Hello, I am open to any feedback. Thanks.

libs/core/langchain_core/beta/runnables/context.py
```diff
@@ -307,7 +307,34 @@ async def ainvoke(
...
+    The `Context` class provides methods for creating context scopes,
+    getters, and setters within a runnable. It allows for managing
+    and accessing contextual information throughout the execution
+    of a program.
+
+    Example:
+        .. code-block:: python
+
+            from langchain_core.beta.runnables.context import Context
+
```
eyurtsev, line 322: @aaronjimv  --  Thanks for the doc-string!

Would you mind adding an example of it being used inside of an LCEL chain -- Context is not really meant to be used outside off a chain

aaronjimv: Hi @eyurtsev thanks for the feedback. I still new using langchain so let me know any comment.

hwchase17 (approved):

Issue #19057, opened 2024-03-14
The validation of tools within OpenAIAssistantRunnable.create_assistant does not account for `{"type": "code_interpreter"}`.


fmyblack:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_experimental.openai_assistant import OpenAIAssistantRunnable

interpreter_assistant = OpenAIAssistantRunnable.create_assistant(
    name="langchain assistant",
    instructions="You are a personal math tutor. Write and run code to answer math questions.",
    tools=[{"type": "code_interpreter"}],
    model="gpt-4-1106-preview"
)
output = interpreter_assistant.invoke({"content": "What's 10 - 4 raised to the 2.7"})
```

### Error Message and Stack Trace (if applicable)

  File "xxx/t.py", line 43, in <module>
    interpreter_assistant = OpenAIAssistantRunnable.create_assistant(
  File "xxx/site-packages/langchain/agents/openai_assistant/base.py", line 213, in create_assistant
    tools=[convert_to_openai_tool(tool) for tool in tools],  # type: ignore
  File "xxx/langchain/agents/openai_assistant/base.py", line 213, in <listcomp>
    tools=[convert_to_openai_tool(tool) for tool in tools],  # type: ignore
  File "xxx/langchain_core/utils/function_calling.py", line 329, in convert_to_openai_tool
    function = convert_to_openai_function(tool)
  File "xxx/langchain_core/utils/function_calling.py", line 304, in convert_to_openai_function
    raise ValueError(
ValueError: Unsupported function

{'type': 'code_interpreter'}

Functions must be passed in as Dict, pydantic.BaseModel, or Callable. If they're a dict they must either be in OpenAI function format or valid JSON schema with top-level 'title' and 'description' keys.

### Description

I used the example code, but it threw an error when executed.

example code from https://python.langchain.com/docs/modules/agents/agent_types/openai_assistants#using-only-openai-tools

### System Info

$ pip list | grep langchain
langchain                                0.1.12
langchain-community                      0.0.28
langchain-core                           0.1.31
langchain-openai                         0.0.8
langchain-text-splitters                 0.0.1

[...]

PR #19081, opened 2024-03-14
[langchain] fix OpenAIAssistantRunnable.create_assistant

ccurme:
- **Description:** OpenAI assistants support some pre-built tools (e.g., `"retrieval"` and `"code_interpreter"`) and expect these as `{"type": "code_interpreter"}`. This may have been upset by https://github.com/langchain-ai/langchain/pull/18935
- **Issue:** https://github.com/langchain-ai/langchain/issues/19057

### DIFF
--- a/libs/langchain/langchain/agents/openai_assistant/base.py
+++ b/libs/langchain/langchain/agents/openai_assistant/base.py
@@ -89,7 +89,7 @@ def _get_openai_async_client() -> openai.AsyncOpenAI:
 
 def _is_assistants_builtin_tool(
     tool: Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool],
-) -> Dict[str, Any]:
+) -> bool:
     """Determine if tool corresponds to OpenAI Assistants built-in."""
     assistants_builtin_tools = ("code_interpreter", "retrieval")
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/agents/openai_assistant/base.py
```diff
@@ -76,6 +87,20 @@ def _get_openai_async_client() -> openai.AsyncOpenAI:
...
 
+def _get_assistants_tool(
+    tool: Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool],
+) -> Dict[str, Any]:
+    """Convert a raw function/class to an OpenAI tool.
+
+    Note that OpenAI assistants supports several built-in tools,
+    such as "code_interpreter" and "retrieval."
+    """
+    if isinstance(tool, dict) and "type" in tool:
```
baskaryan, line 98: do we want to support passing in non-openai format json schema? if so we should explicitly check that tool["type"] is code_interpreter or retrieval

baskaryan: could also see an argument that we should add that check to convert_to_openai_tool

PR #19084, opened 2024-03-14
community: Add support for cohere SDK v5 (keeps v4 backwards compatibility)

billytrend-cohere:
- **Description:** Add support for cohere SDK v5 (keeps v4 backwards compatibility)


libs/community/langchain_community/embeddings/cohere.py
```diff
@@ -34,7 +36,7 @@ class CohereEmbeddings(BaseModel, Embeddings):
 
     cohere_api_key: Optional[str] = None
 
-    max_retries: Optional[int] = 3
+    max_retries: int = 3
```
efriis, line 39: Will technically break anyone explicitly passing `max_retries=None`, which may be something people were using instead of `max_retries=0`. If that wouldn't work before though, I like this as it's more explicit!

efriis: Ah actually looks like old behavior would do something unpredictable because would "fall back on other params to urllib3.Retry" but there were no other params. This is a good change.

efriis (approved):

ddl-john-alexander:
@billytrend-cohere  I'm wondering if this contributed to this LangChain BaseCohere Rerank Argument Mismatch issue : https://github.com/langchain-ai/langchain/discussions/19461  suggested fix: https://github.com/langchain-ai/langchain/pull/19486

PR #19098, opened 2024-03-15
voyageai[patch]: init package

efriis:


libs/partners/voyageai/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Voyage AI
```
efriis, line 3: going to update this to LangChain to match other packages, and happy to discuss! Just want to make sure I'm not doing something that will create issues for folks using langchain-x packages

thomas0809: Sounds good to me!

PR #19119, opened 2024-03-15
text-splitters:fix json split of RecursiveJsonSplitter which may cause memory of chunks

six17:
- **Description:** This modification addresses the issue of mutable default parameters in functions. In the original code, the `chunks` parameter is defaulted to a list containing an empty dictionary, which is mutable. Since default parameters in Python are evaluated only once at function definition time, modifications to the parameter would persist across future calls. By changing the default to `None` and checking/initializing within the function, a new list is created for each call, thus avoiding potential issues.


baskaryan (approved):

bZichett:
May or did?

PR #19123, opened 2024-03-15
Community: Add mistral oss model support to azureml endpoints, plus configurable timeout

tjaffri:
- **Description:** There was no formatter for mistral models for Azure ML endpoints. Adding that, plus a configurable timeout (it was hard coded before)
- **Dependencies:** none
- **Twitter handle:** @tjaffri @docugami


libs/community/langchain_community/llms/azureml_endpoint.py
```diff
@@ -11,10 +11,15 @@
 from langchain_core.pydantic_v1 import BaseModel, SecretStr, root_validator, validator
 from langchain_core.utils import convert_to_secret_str, get_from_dict_or_env
 
+DEFAULT_TIMEOUT = 50
+
 
 class AzureMLEndpointClient(object):
     """AzureML Managed Endpoint client."""
 
+    timeout: int = DEFAULT_TIMEOUT
```
baskaryan, line 20: can we make this an init arg not a class attribute?

tjaffri: @baskaryan done! please take a look should be good to merge now I hope

hwchase17 (approved):

PR #19127, opened 2024-03-15
Community: gather token usage info in BedrockChat during generation

dmenini:
This PR allows to calculate token usage for prompts and completion directly in the generation method of BedrockChat. The token usage details are then returned together with the generations, so that other downstream tasks can access them easily.

This allows to define a callback for tokens tracking and cost calculation, similarly to what happens with OpenAI (see [OpenAICallbackHandler](https://api.python.langchain.com/en/latest/_modules/langchain_community/callbacks/openai_info.html#OpenAICallbackHandler). I plan on adding a BedrockCallbackHandler later.
Right now keeping track of tokens in the callback is already possible, but it requires passing the llm, as done here: https://how.wtf/how-to-count-amazon-bedrock-anthropic-tokens-with-langchain.html. However, I find the approach of this PR cleaner.

Thanks for your reviews. FYI @baskaryan, @hwchase17


esoler-sage:
Hey! For Bedrock you may want to use the tokens returned in the headers "x-amzn-bedrock-*", like 
- x-amzn-bedrock-output-token-count
- x-amzn-bedrock-input-token-count
instead of using anthropic only tokenizer 🤔?

dmenini:
> Hey! For Bedrock you may want to use the tokens returned in the headers "x-amzn-bedrock-*", like
> 
> * x-amzn-bedrock-output-token-count
> * x-amzn-bedrock-input-token-count
>   instead of using anthropic only tokenizer 🤔?

Thanks for this comment! I didn't know about those headers. It definitely makes sense and I'll rework my implementation accordingly

dmenini:
The PR is updated! Now token counters are read from the headers in case of simple generation: The resulting LLMOutput will have the fields `model_id` and `usage`, containing the token counters.

NB In case of streaming, `usage` will be an empty dict as extracting the tokens is more complicated. Indeed, the headers are not available, and one would have to extract the token counters from the body, according to the model provider structure. Will do it at a later time :)

baskaryan:
cc @3coins

libs/community/langchain_community/llms/bedrock.py
```diff
@@ -162,9 +163,19 @@ def prepare_output(cls, provider: str, response: Any) -> dict:
...
         return {
             "text": text,
             "body": response_body,
+            "usage": {
+                "prompt_tokens": int(
+                    headers.get("x-amzn-bedrock-input-token-count", 0)
+                ),
+                "completion_tokens": int(
+                    headers.get("x-amzn-bedrock-output-token-count", 0)
+                ),
```
pratik60, line 177: This is great! I was looking to achieve the same thing to enable cost monitoring on the Anthropic Bedrock models. Should we also add total_tokens (sum of prompt_tokens + completion_tokens), to keep it compatible with the OpenAI model?

dmenini: Thanks for the comment! I added the `total_count`

pratik60: Thanks! You have linting errors, fyi.

Now need someone with write access to the project to look at it and give their review.

baskaryan (approved):

baskaryan (approved):

Sukitly:
Excellent work!
I'm curious about the differences between the use of "usage" in the response body versus in the headers. According to the [Bedrock API documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html), the "usage" is included in the response body as follows:
```json
{
    "id": string,
    "model": string,
    "type": "message",
    "role": "assistant",
    "content": [
        {
            "type": "text",
            "text": string
        }
    ],
    "stop_reason": string,
    "stop_sequence": string,
    "usage": {
        "input_tokens": integer,
        "output_tokens": integer
    }
    
}
```
Do you guys have any insights on this? @dmenini @esoler-sage @pratik60 
Thanks!

esoler-sage:
I think usage in headers is for sync calls, and async call' last chunk contains usage in body 🤔

PR #19130, opened 2024-03-15
docs: Some fixes

adarmiento:
- **Description:** fixes some minor issues in the documentation


docs/docs/expression_language/get_started.ipynb
baskaryan, line 1: should be able to import DocArrayInMemorySearch from langchain_community.vectorstores

baskaryan (approved):

Issue #19139, opened 2024-03-15
Wrong import path in `docs/integrations/retrievers/flashrank-reranker.ipynb`


Simon-Stone:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain.retrievers.document_compressors import FlashrankRerank
from langchain.retrievers import ContextualCompressionRetriever
```

### Description

The import path in the example notebook for the `FlashrankReranker` is wrong. It is listed as `from langchain.retrievers ...` but should be `from langchain.retrievers.document_compressors ...`.

### System Info

```
langchain==0.1.12
langchain-community==0.0.28
langchain-core==0.1.31
langchain-openai==0.0.8
langchain-text-splitters==0.0.1
```

PR #19146, opened 2024-03-15
docs: fix import path for `FlashrankRerank` example notebook

Simon-Stone:
**Description:** Fixes the import paths for the `FlashrankRerank` example notebook.
 **Issue:** #19139 
 **Dependencies:** None
 **Twitter handle:** n/a


docs/docs/integrations/retrievers/flashrank-reranker.ipynb
baskaryan, line 1: why get rid of all the example outputs?

Simon-Stone: I wasn't sure how to deal with that. I tend to clear the output of notebooks, because they otherwise always show up  in the diff, even though nothing really changes (because the run counter increments and stuff like that). I figured the notebooks might be run as part of the CI/CD pipeline anyway, so I figured I would just clear them.

Simon-Stone: Do you want me to add the outputs back in?

Simon-Stone: @baskaryan I went ahead and added them. This PR is not important (to me) as #19148, so if you only have time to consider one, please check out #19148.

hwchase17 (approved):

PR #19152, opened 2024-03-15
core[patch]: allow "placeholder" type in from_messages tuples

hinthornw:


libs/core/langchain_core/prompts/chat.py
```diff
@@ -590,6 +593,45 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
...
+
+            # In addition to Human/AI/Tool/Function messages,
+            # you can initialize the template with a MessagesPlaceholder
+            # either using the class directly or with the shorthand tuple syntax:
+
+            template = ChatPromptTemplate.from_messages([
+                ("system", "You are a helpful AI bot."),
+                # Means the template will receive an optional list of messages under
+                # the "conversation" key
+                ("placeholder", "conversation")
```
hinthornw, line 608: Should update this example!

hinthornw:
```suggestion
                ("placeholder", "{conversation}")
```

PR #19156, opened 2024-03-15
[Feat] Accept non-dict if only 1 prompt input variable

hinthornw:
For prompt templates with only 1 variable (common in e.g., MessageGraph), it's convenient to wrap the incoming object in the variable before formatting.

The downside of this, of course, would be that some number of invocations will successfully format when the user may have intended to format it properly before


nfcampos (approved):

libs/core/tests/unit_tests/prompts/test_chat.py
```diff
@@ -533,3 +533,16 @@ def test_chat_prompt_message_placeholder_partial() -> None:
     assert prompt.format_messages() == []
     prompt = prompt.partial(history=[("system", "foo")])
     assert prompt.format_messages() == [SystemMessage(content="foo")]
+
+
+def test_messages_prompt_accepts_list() -> None:
+    prompt = ChatPromptTemplate.from_messages([MessagesPlaceholder("history")])
+    value = prompt.invoke([("user", "Hi there")])  # type: ignore
```
eyurtsev, line 540: his is confusing -- are we sure we want to support?, why is it a list of 2-tuples rather than only a 2-tuple?

hinthornw: We already support `prompt.invoke({"history": [("user", "hi there"]})`

eyurtsev (approved):
would you be able to document this feature via an example in a relevant place?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -574,11 +574,50 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
...
+            #        AIMessage(content="I'm doing well, thanks!"),
+            #        HumanMessage(content='What is your name?')
+            #    ]
+            #)
+
+        If your prompt has only a single input variable (i.e., 1 instance of "{variable_nams}"),
+        and you invoke the template with a non-dict object, the prompt template will
+        inject the provided argument into that variable location.
+
+        Single-variable template:
```
hinthornw, line 597: @baskaryan do you think this formatting would work?

baskaryan: this should be indented at same level as Examples: (L564)

PR #19159, opened 2024-03-15
core[patch]: Move `globals` to a module instead of a package (non breaking change)

leo-gan:
Classes and functions defined in __init__.py are not parsed into the API Reference.
For example: libs/core/langchain_core/globals/__init__.py : `set_verbose` `get_llm_cache`, `set_llm_cache`, ...
And the whole `langchain_core.globals` namespace is not visible in the API Reference.

The refactoring is just file renaming.


eyurtsev (approved):

eyurtsev:
Thank you!

eyurtsev:
It sounds like we should update the logic in API reference to avoid skipping over __init__ for any logic defined directly in __init__?

So basically if a function / class appears there and it's defined there then it should be included in the API reference

leo-gan:
> It sounds like we should update the logic in API reference to avoid skipping over **init** for any logic defined directly in **init**?
> 
> So basically if a function / class appears there and it's defined there then it should be included in the API reference
@eyurtsev That should be a reason to include class/function definitions into ____init___.py It contradicts the purpose of this file. If there is a reason for including definitions there, this reason should be clearly documented. 
Right now, all files (as classes and functions) with names starting with underscores are treated as "private" and are not included in the API Reference.

PR #19161, opened 2024-03-15
core: `messages` fix

leo-gan:
Classes and functions defined in __init__.py are not parsed into the API Reference.
For example:
- libs/core/langchain_core/messages/__init__.py : AnyMessage, MessageLikeRepresentation, get_buffer_string(), messages_from_dict(), ...

Opinionated: __init__.py is not a typical place to define artifacts.

Moved artifacts from __init__ into utils.py. 
Added `MessageLikeRepresentation` to __all__ since it is used outside of `messages`, for example, in `libs/core/langchain_core/language_models/base.py`
Added `_message_from_dict` to __all__ since it is used outside of `messages`(???) I would add `message_from_dict` (without underscore) as an alias. Please, advise.

### DIFF
--- a/libs/core/langchain_core/messages/__init__.py
+++ b/libs/core/langchain_core/messages/__init__.py
@@ -14,7 +14,6 @@
     ChatPromptTemplate
 
 """  # noqa: E501
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 from langchain_core.messages.ai import AIMessage, AIMessageChunk
 from langchain_core.messages.base import (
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/messages/__init__.py
```diff
@@ -259,15 +50,17 @@ def convert_to_messages(
     "FunctionMessageChunk",
     "HumanMessage",
     "HumanMessageChunk",
+    "MessageLikeRepresentation",
     "SystemMessage",
     "SystemMessageChunk",
     "ToolMessage",
     "ToolMessageChunk",
+    "_message_from_dict",
```
eyurtsev, line 58: private namespace

libs/core/tests/unit_tests/messages/test_imports.py
```diff
@@ -1,6 +1,8 @@
 from langchain_core.messages import __all__
 
 EXPECTED_ALL = [
+    "MessageLikeRepresentation",
+    "_message_from_dict",
```
eyurtsev, line 5: this one is private -- could we remove it?

libs/core/langchain_core/messages/__init__.py
leo-gan: @eyurtsev As I mentioned in the PR description: "Added _message_from_dict to all since it is used outside of messages(???) I would add message_from_dict (without underscore) as an alias. Please, advise."

libs/core/tests/unit_tests/messages/test_imports.py
leo-gan: See the previous comment

libs/core/langchain_core/messages/__init__.py
eyurtsev: Got it -- OK for now. It's backwards compatible, we can later deprecate from langchain if need be

leo-gan:
@eyurtsev Thank you!

PR #19163, opened 2024-03-15
langchain: upgrade mypy

eyurtsev:
Update mypy in langchain


efriis (commented):
If you can get tests passing I'm all for it but I can't even get a lot of the partner packages passing on mypy^1 because of pydantic v1 stub

libs/core/langchain_core/env.py
```diff
@@ -3,7 +3,7 @@
 
 
 @lru_cache(maxsize=1)
-def get_runtime_environment() -> dict:
+def get_runtime_environment() -> dict: # type: ignore
```
eyurtsev, line 6:
```suggestion
def get_runtime_environment() -> dict:
```

baskaryan (approved):

Issue #18804, opened 2024-03-08
Add in code documentation to core Runnable methods


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Hi all!

Core Runnable methods are not documented well enough yet. It would be great to have each of the method include a self-contained usage example that folks can use as a reference. 

Documenting these methods well will have a fairly high value impact by making is easier for LangChain users to use core LangChain primitives.

Runnable (https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/runnables/base.py#L103-L103):

* assign
* with_fallbacks
* with_retry
* pick
* pipe
* map
* with_listeners

RunnableSerializable (https://github.com/langchain-ai/langchain/blob/bc6249c889a4fe208f8f145e80258eb3de20d2d4/libs/core/langchain_core/runnables/base.py#L1664-L1664):

* configurable_fields
* configurable_alternatives

## Acceptance criteria

- Documentation includes context
- Documentation includes a self contained example in python (including all imports).
- Example uses ..code-block: python syntax (see other places in the code as reference).

Please document only on method per PR to make it easy to review and get PRs merged quickly!

liugddx:
@eyurtsev Hi ,assign to me, thanks~

eyurtsev:
Claim a method :)

Ideally the or only documents one method -- it's much faster to review and merge

liugddx:
`assign` method.

liugddx:
`with_fallbacks` method.

balachidam21:
`with_retry` method

liugddx:
This issue has been completed, please close it cc @eyurtsev

spike-spiegel-21:
Hi @eyurtsev is this issue still open? All the methods mentioned above have already been documented I guess. If there are some more, please mention them.

PR #19192, opened 2024-03-17
code[patch]: Add in code documentation to core Runnable with_retry method (docs only) 

liugddx:
- **Description:** Add in code documentation to core Runnable with_retry method (docs only) 
- **Issue:** #18804 
@baskaryan @eyurtsev PTAL

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -1344,6 +1344,36 @@ class Runnable(Generic[Input, Output], ABC):
     ) -> Runnable[Input, Output]:
         """Create a new Runnable that retries the original runnable on exceptions.
 
+        Example:
+
+        .. code-block:: python
+            from typing import Union
+
+            import pytest
+            from langchain_core.runnables import Runnable, RunnableLambda
+            from pytest_mock import MockerFixture
+
+
+            def test_retrying(mocker: MockerFixture) -> None:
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/runnables/base.py
```diff
@@ -1344,6 +1344,36 @@ def with_retry(
...
         """Create a new Runnable that retries the original runnable on exceptions.
 
+        Example:
+
+        .. code-block:: python
+            from typing import Union
+
+            import pytest
+            from langchain_core.runnables import Runnable, RunnableLambda
+            from pytest_mock import MockerFixture
```
ccurme, line 1354: Thanks for this!

Ideally this code is easily executed by users getting started with `.with_retry`. Here we require `pytest` and `pytest_mock` for running the example. Could you modify this to be more easily run without requiring our test dependencies? e.g.,
```python


count = 0

def _lambda(x: int) -> None:
    global count
    count = count + 1
    if x == 1:
        raise ValueError("x is 1")
    else:
        pass

runnable = RunnableLambda(_lambda)
try:
    runnable.with_retry(
        stop_after_attempt=2,
        retry_if_exception_type=(ValueError,),
    ).invoke(1)
except ValueError:
    pass

assert(count == 2)
```

ccurme:
Looks good, thanks!

PR #19202, opened 2024-03-17
community: added support for handling parsing errors in SQL agent

tewatia:
**Added error handling for LLM output parsing errors**

This commit introduces the ability to handle errors when the output from the Language Learning Model (LLM) is not in the expected format for the output parser. Developers can now use the `handle_parsing_errors` parameter when creating an SQL agent to enable this feature. This parameter is optional and defaults to `False` to ensure backward compatibility with existing functionality.

Example usage:
agent_executor = AgentExecutor(agent=agent, tools=tools, verbose=True, **handle_parsing_errors=True**)

Dependencies: None

### DIFF
--- a/libs/community/langchain_community/agent_toolkits/sql/base.py
+++ b/libs/community/langchain_community/agent_toolkits/sql/base.py
@@ -61,6 +61,7 @@ def create_sql_agent(
     *,
     db: Optional[SQLDatabase] = None,
     prompt: Optional[BasePromptTemplate] = None,
+    handle_parsing_errors : bool = False,
     **kwargs: Any,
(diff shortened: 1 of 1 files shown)


tewatia:
@efriis Hi Erick, can you please validate this? It's a simple and quite useful change. 

SQL Agent cannot handle parsing errors as of now. This simple change allows developers to pass the handle_parsing_errors param to handle such errors.

hwchase17:
this parameter can already be specified in `agent_executor_kwargs` (unless I misunderstanding) and adding here would be backwards incompatible for anyone currently passing in that way, so going to keep as is for now

PR #19224, opened 2024-03-18
community: Add xinference-client support

chenchun0629:
**Description:** support xinference-client
**Issue:** None
**Dependencies:** xinference-client


zephyr737:
It is hoped that the author can add model.chat for xinference in the community library. Currently, only generate can be used

PR #19226, opened 2024-03-18
moved fake llms and embeddings to core

lkuligin:
- [ ] **PR title**: "core: moved fake llms and embeddings to core"

- [ ] **PR message**:
 - **Description:** moved fake llms and embeddings to core"


efriis (approved):

eyurtsev:
Are there specific models that are needed for unit tests? langchain_core exposes many fake models right now from core and not all of them are implemented correctly, and most of them are not documented. I think we should be a bit more selective in terms of what's exposed as part of the public API since it's used as reference by developers

PR #19253, opened 2024-03-19
experimental: update the notebook link of semantic chunk.

1openwindow:
update the notebook link of semantic chunk.


ccurme (approved):
Thanks!

PR #19266, opened 2024-03-19
core: Rename parameter signature in BaseOutputParser.parse_with_prompt method

kmseo-gh:
### Description 

I've renamed the parameter signature of OutputParsers, which use llm for fixing on completions, to follow a consistent format `(text: str, prompt_value: PromptValue)`

It's anticipated that the `BaseOutputParser.parse_with_prompt` method will have minimal side-effects, as it's only overridden by `RetryOutputParser` and `RetryWithErrorOutputParser` in `output_parsers/retry.py`.

- AS-IS:
```python
# libs/core/langchain_core/output_parsers/base.py

    # TODO: rename 'completion' -> 'text'.
    def parse_with_prompt(self, completion: str, prompt: PromptValue) -> Any:
```
- TO-BE: 

```python
# libs/core/langchain_core/output_parsers/base.py

    def parse_with_prompt(self, text: str, prompt_value: PromptValue) -> Any:
```
```python
# libs/langchain/langchain/output_parsers/fix.py

    def parse(self, text: str) -> T:
        completion = text
        retries = 0

        while retries <= self.max_retries:
        # ...
```

```python
# libs/langchain/langchain/output_parsers/retry.py

    def parse_with_prompt(self, text: str, prompt_value: PromptValue) -> T:
        completion = text
        retries = 0

        while retries <= self.max_retries:
        # ...

```

### DIFF
--- a/libs/core/langchain_core/output_parsers/base.py
+++ b/libs/core/langchain_core/output_parsers/base.py
@@ -259,8 +259,7 @@ class BaseOutputParser(
         """
         return await run_in_executor(None, self.parse, text)
 
-    # TODO: rename 'completion' -> 'text'.
-    def parse_with_prompt(self, completion: str, prompt: PromptValue) -> Any:
+    def parse_with_prompt(self, text: str, prompt_value: PromptValue) -> Any:
         """Parse the output of an LLM call with the input prompt for context.
 
         The prompt is largely provided in the event the OutputParser wants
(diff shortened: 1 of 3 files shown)


eyurtsev:
Hi @KangmoonSeo thank you for the contribution. I am not sure that we'll want to roll out these changes.

1) The methods signatures could cause breaking changes since we're not using the `/` operator in our code. So users may be invoking those parameters by name.
2) The output parser exceptions are used by some caller code to do retrys with LLMs.

KangmoonSeo:
Hello! Thank you, @eyurtsev, for your feedback.

I appreciate your concern regarding the potential breaking changes due to the modification of your method signature.

Should we consider removing the `TODO:` comment requesting the change in the `BaseOutputParser.parse_with_prompt` signature?

Additionally, the adjustments made to the `OutputParserException` now specifically address cases where the user inputs a max_retries value less than zero. The purpose of this change is to explicitly indicate which text parsing failed.

Thank you for reviewing the code!

PR #19280, opened 2024-03-19
core[patch]: Pass sync run manager for sync stream fallback in astream

eyurtsev:
This PR patches the fallback in chat models and language models to pass in the appropriate version of the run manager (sync vs. async)


libs/core/langchain_core/language_models/fake_chat_models.py
```diff
@@ -279,25 +278,6 @@ def _stream(
                         )
                     yield chunk
 
-    async def _astream(
```
eyurtsev, line 282: Discovered bug while removing this code

nfcampos (approved):

PR #19281, opened 2024-03-19
[langchain_community] Update/Insert texts to the Milvus vectorstore

rgupta2508:
Update/Insert texts to the Milvus vectorstore.

        Inserting data when the collection is already exists
        and need to insert data based on collection schema,
        Metadata keys will need to be present for all inserted values. At
        the moment there is no None equivalent in Milvus.
        This is based on upsert data based on add_text functionality

- [ ] **PR title**: "package: description"

Additional guidelines:


rgupta2508:
@hwchase17 Please review, Adding function to upsert text directly based on one of the field from metadata

baskaryan:
cc @xiaofan-luan

rgupta2508:
@xiaofan-luan Pleas review

rgupta2508:
@baskaryanPleas review if this can we merged

cc:- @xiaofan-luan/

Thanks

libs/community/langchain_community/vectorstores/milvus.py
```diff
@@ -1057,3 +1057,61 @@ def upsert(
...
+        if metadatas is None or len(metadatas) == 0:
+            return self.add_texts(texts=texts, metadatas=metadatas, **kwargs)
+        else:
+            values = [metadata[field_name] for metadata in metadatas]
+        # Convert to set for optimize expression
+        values = list(set(values))
+        if values is not None and len(values):
+            try:
+                if field_name == "ids":
+                    self.delete(ids=values)
```
xiaofan-luan, line 1105: why do we not use upsert interface directly inside pymilvus but delete then insert?

xiaofan-luan: if we first delete then insert there seems to be no consistency guarantee

rgupta2508:
@xiaofan-luan  Can you please suggest in this case, 

Actually current upsert interface only works with IDs, what if some one want to delete some of the row and adding n ew row in-place of those rows.
Basically embedding of one file can be added to multiple rows, based on splitter. now just want to update embedding for the same file with updated content , so this time it may create more or less row based on content and splitter.

Please review this or can you please suggest any other another possible solution.

xiaofan-luan:
so far the best you can do is to query with expressions to find. Then upsert with primary keys.

We could think of add upsert with expression in the future

rgupta2508:
Ok thank you for your quick response.
But i cant get key also, no of key(row) may not be same for one expression, it depends on metadata .
Fo example for, one expression i got 3 rows, but i need to add 5 row in place of those embedding row as some of the document might be splitted to multiple chunks . and how can will handle those chunks.?

I just tried to extend this method with expression :- https://github.com/langchain-ai/langchain/blob/c769421aa4fc4783968e069d800ef85191ce87da/libs/community/langchain_community/vectorstores/milvus.py#L1037
One option i can think is, deleting data using pymilvus interface, and adding using langchain.

xiaofan-luan:
> Ok thank you for your quick response @xiaofan-luan . But i cant get key also, no of key(row) may not be same for one expression, it depends on metadata . Fo example for, one expression i got 3 rows, but i need to add 5 row in place of those embedding row as some of the document might be splitted to multiple chunks . and how can will handle those chunks.?
> 
> I just tried to extend this method with expression :-
> 
> https://github.com/langchain-ai/langchain/blob/c769421aa4fc4783968e069d800ef85191ce87da/libs/community/langchain_community/vectorstores/milvus.py#L1037
> 
> 
> One option i can think is, deleting data using pymilvus interface, and adding using langchain.

upsert can be used as insert as well. If no exsiting entity, you can simply add a new row while upsert

PR #19290, opened 2024-03-19
community: added `partners/package-name` folders

leo-gan:
Added references to new integration packages from Google, by adding subfolders to `partners/`.


libs/partners/google-alloydb-pg/README.md
```diff
@@ -0,0 +1 @@
+placeholder
```
baskaryan, line 1: could we give these pkg name as title instead?
```markdown
# langchain-google-alloydb-pg
```

leo-gan: Sure. Wouldn't it be confusing? Probably the same as `placeholder` :)

leo-gan: @baskaryan It is ready.

leo-gan:
@baskaryan Thanks!!

efriis:
@leo-gan this breaks the api docs build, and the folders here should not be the source of truth of ecosystem packages. The only reason there are some here is to be "tombstones" for packages that used to be here and moved.

Reverting

PR #19293, opened 2024-03-19
community [patch] : [Fiddler] ensure dataset is not added if model is present

bhalder:
**Description:**
- minor PR to speed up onboarding by not trying to add a dataset, if a model is already present.
- replace batch publish API with streaming when single events are published.

**Dependencies:** any dependencies required for this change
**Twitter handle:** behalder


bhalder:
Ping @baskaryan for reviews 🙏

baskaryan (approved):

PR #19314, opened 2024-03-20
partners, elasticsearch: Enable `ElasticsearchStore` to retrieve with the pure BM25 algorithm without vector search

g-votte:
## Description
This pull request proposes the implementation of the `BM25RetrievalStrategy` for `ElasticsearchStore`. This retrieval strategy enables searches purely based on BM25 without involving vector search. Below, the usage example, motivation, and details of the changes are discussed.

## Usage Example of Introduced Feature
By specifying the `BM25RetrievalStrategy` as a constructor argument for `ElasticsearchStore`, users can perform searches using pure BM25 without vector search. Note that in the example below, the embedding option is not specified, indicating that the search is conducted without using embeddings.

```python
from langchain_elasticsearch.vectorstores import ElasticsearchStore

store = ElasticsearchStore(
    es_url="http://localhost:9200",
    index_name="test_index",
    strategy=ElasticsearchStore.BM25RetrievalStrategy(),
)

store.add_texts(
    [
        "foo",
        "foo bar",
        "foo bar baz",
        "bar",
        "bar baz",
        "baz"
    ],
)

results = store.similarity_search(query="foo", k=10)
print(results)
```

The example above outputs:
```
[Document(page_content='foo'), Document(page_content='foo bar'), Document(page_content='foo bar baz')]
```

## Motivation
There is a considerable demand for using Elasticsearch as a pure BM25 retriever without vector search. Although hybrid searches combining vector search and BM25 are supported alongside `ApproxRetrievalStrategy`, there are cases where pure BM25 searches are needed, for example, when seeking speed performance in search or prioritizing exact matches over semantic searches.

Pure BM25 retrievers, such as [`ElasticsearchBM25Retriever`](https://github.com/langchain-ai/langchain/blob/40f846e65da37a1c00d72da9ea64ebb0f295b016/libs/community/langchain_community/retrievers/elastic_search_bm25.py) and [`BM25Retriever`](https://github.com/langchain-ai/langchain/blob/40f846e65da37a1c00d72da9ea64ebb0f295b016/libs/community/langchain_community/retrievers/bm25.py), have already been implemented in `langchain_community`. However, these classes do not offer the rich and flexible Elasticsearch features supported by `ElasticsearchStore`, such as various authentication options and flexible querying with `custom_query`. Additionally, being a subclass of `VectorStore`, it benefits from components like `RecordManager`, which are advantageous during operational phases. Therefore, supporting pure BM25 searches in `ElasticsearchStore` presents significant benefits.

To achieve this, an easy-to-use interface is necessary. The abstraction of `ElasticsearchStore` itself is sophisticated enough to support pure BM25 without vector search, as it allows inputting a strategy class inheriting from `BaseRetrievalStrategy`. However, implementing a BM25 search by inheriting from `BaseRetrievalStrategy` can be challenging for general users (as it took me several days). Without native support from the library, it's difficult for users to arrive at this solution.

Therefore, this PR suggests implementing a new retrieval strategy, `BM25RetrievalStrategy`, to enable `ElasticsearchStore` to support pure BM25 searches.

## Change Details
- Implemented the `BM25RetrievalStrategy` class.
- Added a new argument, `text_field`, to the `index` method of `BaseRetrievalStrategy`.
  - While strictly speaking, this modification could potentially affect backward compatibility, it's my belief that the number of users directly inheriting and utilizing `BaseRetrievalStrategy` in their projects is relatively small. Consequently, the overall impact of this change should be minimal. Nonetheless, I am open to and would appreciate any feedback on this matter.
- Added integration test scenarios to verify the behavior of `BM25RetrievalStrategy`, named `test_similarity_search_bm25_search*`. Following the example of other retrieval strategy classes, scenarios with and without using the filter option were tested.


baskaryan:
cc @joemcelroy

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
```diff
@@ -389,6 +394,72 @@ def require_inference(self) -> bool:
...
+                    text_field: {
+                        "type": "text",
+                        "similarity": "custom_bm25",
+                    },
+                },
+            },
+            "settings": {
+                "similarity": {
+                    "custom_bm25": {
+                        "type": "BM25",
```
joemcelroy, line 451: is being able to specify similarity options a requirement for you?

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
```diff
@@ -1284,3 +1356,15 @@ def SparseVectorRetrievalStrategy(
                     deployed to Elasticsearch.
         """
         return SparseRetrievalStrategy(model_id=model_id)
+
+    @staticmethod
+    def BM25RetrievalStrategy(
+        k1: float = 2.0, b: float = 0.75
```
maxjakob, line 1362: might consider setting the defaults on the __init__ function, so that `BM25RetrievalStrategy()` is simple to use as well

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
```diff
@@ -70,6 +70,7 @@ def index(
         self,
         dims_length: Union[int, None],
         vector_query_field: str,
+        text_field: str,
```
miguelgrinberg, line 73: Suggestion, this could be `str | str[]`, and when a list is provided a `multi_match` query is used.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
```diff
@@ -389,6 +394,72 @@ def require_inference(self) -> bool:
         return False
 
 
+class BM25RetrievalStrategy(BaseRetrievalStrategy):
+    """Retrieval strategy using the native BM25 algorithm of Elasticsearch."""
+
+    def __init__(self, k1: float, b: float):
```
miguelgrinberg, line 400: The `k1` and `b` arguments should default to `None`.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
```diff
@@ -389,6 +394,72 @@ def require_inference(self) -> bool:
...
+                        "similarity": "custom_bm25",
+                    },
+                },
+            },
+            "settings": {
+                "similarity": {
+                    "custom_bm25": {
+                        "type": "BM25",
+                        "k1": self.k1,
+                        "b": self.b,
```
miguelgrinberg, line 453: The `k1` and `b` settings should be given in the mapping only when custom values have been passed. If they are `None` then they should not be included, to allow Elasticsearch to use its own defaults.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
miguelgrinberg: I suggest making these two arguments `None` by default, and only include them in the mapping definition when the user provided custom values. That would allow the ES service to apply its own defaults for the general case.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
```diff
@@ -1284,3 +1356,15 @@ def SparseVectorRetrievalStrategy(
...
+
+    @staticmethod
+    def BM25RetrievalStrategy(
+        k1: float = 2.0, b: float = 0.75
+    ) -> "BM25RetrievalStrategy":
+        """Used to apply BM25 without vector search.
+
+        Args:
+            k1: Optional. Default is 2.0. This corresponds to the BM25 parameter, k1.
+            b: Optional. Default is 0.75. This corresponds to the BM25 parameter, b.
```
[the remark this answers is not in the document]
miguelgrinberg: We do not know if these defaults can change in the future, so it's best to not apply any hardcoded defaults in here, the ES service should pick its own defaults, whatever they are.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
joemcelroy: this is fine as a single field as developers are not able to search on multiple fields within a VectorStore.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
g-votte: The direct reason for setting these default values is because the [`ElasticsearchBM25Retriever`](https://github.com/langchain-ai/langchain/blob/40f846e65da37a1c00d72da9ea64ebb0f295b016/libs/community/langchain_community/retrievers/elastic_search_bm25.py) utilizes them. However, I completely agree with @miguelgrinberg 's viewpoint that the default values should be set to `None`, allowing Elasticsearch's default values to be used unless explicitly specified by the user. This approach will ensure that the behavior of `ElasticsearchStore` aligns more consistently with Elasticsearch itself.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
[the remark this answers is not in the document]
g-votte: Similar to [my previous comment](https://github.com/langchain-ai/langchain/pull/19314#discussion_r1541186247), I agree that the default values should be `None`. I have made this modification in this commit: https://github.com/langchain-ai/langchain/pull/19314/commits/31f7909f906f8efdd1fefe42ec185b420e0d2191.

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
g-votte: Thank you for pointing that out. I have addressed it in this commit: https://github.com/langchain-ai/langchain/pull/19314/commits/31f7909f906f8efdd1fefe42ec185b420e0d2191

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
g-votte: As @joemcelroy mentioned, this is because of the limitations of the `VectorStore` and `BaseRetrievalStrategy`. I'm also hoping for the `ElasticsearchStore` to support multi-match searches in the future. But, let's keep that outside this PR for now and think about it as a future work.

joemcelroy (approved):

joemcelroy:
thanks so much for your contribution. The last part is could you document this too https://github.com/langchain-ai/langchain/blob/master/docs/docs/integrations/vectorstores/elasticsearch.ipynb

libs/partners/elasticsearch/langchain_elasticsearch/vectorstores.py
g-votte: Thanks. I addressed in this commit: https://github.com/langchain-ai/langchain/pull/19314/commits/a2eb03a25b43994d76451c5432690971149de348

g-votte:
@joemcelroy 
Thank you for your approval!

I've added the documentation. PTAL.

commit: https://github.com/langchain-ai/langchain/pull/19314/commits/e1ab8f856b2f20084c1bb0c9ac89ccffa84e2768
preview: https://langchain-eibb0l4qr-langchain.vercel.app/docs/integrations/vectorstores/elasticsearch#bm25retrievalstrategy

cc: @miguelgrinberg @maxjakob @baskaryan @efriis

maxjakob (approved):

g-votte:
Hm, `libs/partner/elasticsearch` has been removed from the repository and moved to https://github.com/langchain-ai/langchain-elastic

@joemcelroy @miguelgrinberg @maxjakob @baskaryan @efriis
May I close this PR and create a one with the same changes in  the new `langchain-elastic` repository?
As I'm not sure about the roadmap/timeline of the repository migration, let me know if there's the best timing of moving this PR.

baskaryan:
> Hm, `libs/partner/elasticsearch` has been removed from the repository and moved to https://github.com/langchain-ai/langchain-elastic
> 
> @joemcelroy @miguelgrinberg @maxjakob @baskaryan @efriis May I close this PR and create a one with the same changes in the new `langchain-elastic` repository? As I'm not sure about the roadmap/timeline of the repository migration, let me know if there's the best timing of moving this PR.

yep moving to langchain-elastic makes sense!

g-votte:
Thank you for your comment. I have recreated the PR in the langchain-elastic repository: https://github.com/langchain-ai/langchain-elastic/pull/6
Please review it at your convenience. @joemcelroy @miguelgrinberg @maxjakob @baskaryan @efriis

I will now close this PR in the LangChain main repository.

Issue #12812, opened 2023-11-03
Langchain ConfluenceLoader.load(): Get "when" field in metadata


Vickywwww:
### System Info

langchain: 0.0.311

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

We're trying to load confluence pages using `ConfluenceLoader.load()`.
If we load pages by list of page_ids:
```
loader = ConfluenceLoader(
     url=baseUrl,
     username=username, 
     api_key= api_key
)

documents = loader.load(
    page_ids = ["2081227164", "2096890841", "2081227310"], 
    include_attachments=False, 
    limit=50,
    max_pages=1000
 )
```
We are able to get `when` field in metadata.
However, if we reuqest pages by space
```
documents = loader.load(
    space_key = spaceName,
    include_attachments=False, 
    limit=50,
    max_pages=1000
 )
```
The when field will not be in metadata.

### Expected behavior

Is there any methods we can request pages by space and also get `when` field? This is really important for us.
Thanks :)

SufiSR:
I ran into the same issue. 

The version information comes back empty from your module for whatever reason.
Directly accessing it through the atlassian-python-api module is possible though:

confluence.get_page_by_id(page.metadata['id'], expand="version") works like a charm.

PR #19324, opened 2024-03-20
community: expanding version in confluence loader

stezz:
**Description:**
Expanding version in all the Confluence API calls so to get when the page was last modified/created in all cases.

**Issue:** #12812 
**Twitter handle:** zzste

### DIFF
--- a/libs/community/langchain_community/document_loaders/confluence.py
+++ b/libs/community/langchain_community/document_loaders/confluence.py
@@ -316,7 +316,7 @@ class ConfluenceLoader(BaseLoader):
                 limit=limit,
                 max_pages=max_pages,
                 status="any" if include_archived_content else "current",
-                expand=content_format.value,
+                expand=f"{content_format.value},version",
             )
             yield from self.process_pages(
                 pages,
@@ -349,7 +349,7 @@ class ConfluenceLoader(BaseLoader):
                 limit=limit,
                 max_pages=max_pages,
                 include_archived_spaces=include_archived_content,
-                expand=content_format.value,
+                expand=f"{content_format.value},version",
             )
             yield from self.process_pages(
                 pages,


libs/community/langchain_community/document_loaders/confluence.py
```diff
@@ -316,7 +316,7 @@ def _lazy_load(self, **kwargs: Any) -> Iterator[Document]:
                 limit=limit,
                 max_pages=max_pages,
                 status="any" if include_archived_content else "current",
-                expand=content_format.value,
+                expand=f"{content_format.value},version",
```
eyurtsev, line 319: Is this change backwards compatible? As a maintainer I don't know the underlying APIs so a bit of context is helpful to understand how the response from the document loader is modified? Does this add an extra field in document metadata?

stezz: Hi Eugene,

thanks for picking up this PR so fast. 

> As a maintainer I don't know the underlying APIs so a bit of context is helpful to understand how the response from the document loader is modified? 

All `get_page`, `get_all_pages_from_space` and the API endpoint used in the CQL case (`"rest/api/content/search"`) support the `expand` keyword. (`get_all_pages_by_label` does not and that is why I have not added the `expand` there)

An excerpt of their [docs](https://atlassian-python-api.readthedocs.io/confluence.html):
```
# Get page by ID
# Example request URI(s):
#    http://example.com/confluence/rest/api/content/1234?expand=space,body.view,version,container
#    http://example.com/confluence/rest/api/content/1234?status=any
#    page_id: Content ID
#    status: (str) list of Content statuses to filter results on. Default value: [current]
#    version: (int)
#    expand: OPTIONAL: A comma separated list of properties to expand on the content.
#                   Default value: history,space,version
#                   We can also specify some extensions such as extensions.inlineProperties
#                   (for getting inline comment-specific properties) or extensions.resolution
#                   for the resolution status of each comment in the results
confluence.get_page_by_id(page_id, expand=None, status=None, version=None)

...

# Get all pages from Space
# content_type can be 'page' or 'blogpost'. Defaults to 'page'
# expand is a comma separated list of properties to expand on the content.
# max limit is 100. For more you have to loop over start values.
confluence.get_all_pages_from_space(space, start=0, limit=100, status=None, expand=None, content_type='page')
...

If you don't expand the `version`, consequently the version information of the page is not going to be returned.

But if you expand the version the API will return the following:
```
  "version": {
    "by": {
      "type": "known",
      "accountId": "xxx",
      "accountType": "atlassian",
      "email": "",
      "publicName": "Stefano Mosconi",
    ...
    },
    "when": "2024-03-20T07:39:59.940Z",
    ...
  },
```
> Does this add an extra field in document metadata?

Yes. Specifically we are interested in the `when` field, because we want to know when the current page was edited and the code later on checks for existence of `version` to report the `when` and add it to the `metadata`:

```
        if "version" in page and "when" in page["version"]:
            metadata["when"] = page["version"]["when"]
```
Obviously if we don't expand the version in all situations this will work for some methods (`page_id` 
Note that all I am doing in this PR is blatantly copy/pasting the same `expand=f"{content_format.value},version",` that you already find in the code at line 379:

```
        if page_ids:
            for page_id in page_ids:
            ...
                page = get_page(
>>>                 page_id=page_id, expand=f"{content_format.value},version".   <<<
                )
```

So I believe this was simply overlooked in the first instance by the original contributor of this code.

> Is this change backwards compatible? 

I am not sure I understand the scope of the question, backwards compatible with what? Older versions of LangChain? Older versions of the API? Or ... ?

PR #19326, opened 2024-03-20
community: Add InMemoryVectorStore

cbornet:
This is a basic VectorStore implementation using an in-memory dict to store the documents.
It doesn't need any extra/optional dependency as it uses numpy which is already a dependency of langchain.
This is useful for quick testing, demos, examples.
Also it allows to write vendor-neutral tutorials, guides, etc...

### DIFF
--- a/libs/community/langchain_community/vectorstores/inmemory.py
+++ b/libs/community/langchain_community/vectorstores/inmemory.py
@@ -1,0 +1,199 @@
+import uuid
+from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple
+
+import numpy as np
+from langchain_core.documents import Document
+from langchain_core.embeddings import Embeddings
+from langchain_core.vectorstores import VectorStore
+
+from langchain_community.utils.math import cosine_similarity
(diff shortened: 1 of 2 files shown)


eyurtsev:
@cbornet would you mind moving this to: https://github.com/langchain-ai/langchain/blob/f8078e41e5e599d022101a886f439db060806d53/libs/core/langchain_core/vectorstores.py#L12-L12

---

All in memory implementations are pretty fundamental -- useful as both reference implementations and for testing purposes

cc @baskaryan

eyurtsev:
Standby! We're trying to figure whether it's langchain-core or langchain

cbornet:
Also whereas the core currently depends on numpy, I’m not sure it’s really used and maybe it could be removed.

cbornet:
But a no-numpy, zero-dep, less performant implementation could also probably be written.

eyurtsev:
OK we can merge as is for now. We're working on breaking langchain dependency on langchain community. Once that's broken, we're likely going to invert it with `langchain` containing reference implementations / general algorithms, but this will be after 0.2.0 release.

PR #19332, opened 2024-03-20
core[patch]: Simplify astream logic in BaseChatModel and BaseLLM

cbornet:
Covered by tests in `libs/core/tests/unit_tests/language_models/chat_models/test_base.py`, `libs/core/tests/unit_tests/language_models/llms/test_base.py` and `libs/core/tests/unit_tests/runnables/test_runnable_events.py`


maximeperrindev (approved):
Looks good ! I worked on this part few days ago and it's definitely easier to understand now

eyurtsev (approved):
Thank you this is much cleaner

PR #19346, opened 2024-03-20
community: add request_timeout and max_retries to BaseCohere

kaijietti:
This PR make `request_timeout` and `max_retries` configurable for Cohere.

### DIFF
--- a/libs/community/langchain_community/llms/cohere.py
+++ b/libs/community/langchain_community/llms/cohere.py
@@ -88,6 +88,10 @@ class BaseCohere(Serializable):
     streaming: bool = Field(default=False)
     """Whether to stream the results."""
 
+    max_retries: Optional[int] = 3
+    """Maximum number of retries to make when generating."""
+    request_timeout: Optional[float] = None
+    """Timeout in seconds for the Cohere API request."""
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/llms/cohere.py
```diff
@@ -88,6 +88,10 @@ class BaseCohere(Serializable):
     streaming: bool = Field(default=False)
     """Whether to stream the results."""
 
+    max_retries: Optional[int] = 3
+    """Maximum number of retries to make when generating."""
+    request_timeout: Optional[float] = None
```
mspronesti, line 93: Out of curiosity, is there a reason why this is defaulted to `None` instead of `60` as in Cohere's API ([ref](https://github.com/cohere-ai/cohere-python/blob/8a63bca5b15c2f1298d57094ab956a8315bd990e/src/cohere/client.py#L67))?

```suggestion
    request_timeout: Optional[float] = 60
```

mspronesti:
Thanks for accepting my suggestion :) Looks good to me!

mspronesti:
Opened #19641 to port this changes to the newer package `langchain_cohere`.

PR #19347, opened 2024-03-20
community: use semantic_configurations in AzureSearch

lawadam:
- **Description:** Currently the semantic_configurations are not used when creating an AzureSearch instance, instead creating a new one with default values. This PR changes the behavior to use the passed semantic_configurations if it is present, and the existing default configuration if not.


lawadam:
@efriis Able to take a look at this?

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -69,7 +69,7 @@ def _get_search_client(
     semantic_configuration_name: Optional[str] = None,
     fields: Optional[List[SearchField]] = None,
     vector_search: Optional[VectorSearch] = None,
-    semantic_configurations: Optional[SemanticConfiguration] = None,
+    semantic_configurations: Optional[List[SemanticConfiguration]] = None,
```
baskaryan, line 72: for backwards compatibility we should continue supporting a single SemanticConfiguration as well
```python
semantic_configurations: Optional[Union[SemanticConfiguration, List[SemanticConfiguration]]]
```

baskaryan (approved):

libs/community/langchain_community/vectorstores/azuresearch.py
lawadam: added

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -626,7 +640,6 @@ def from_texts(
         azure_search_endpoint: str = "",
         azure_search_key: str = "",
         index_name: str = "langchain-index",
-        fields: Optional[List[SearchField]] = None,
```
baskaryan, line 629: why remove this?

lawadam: fixed. Was based on an older release.

Issue #19344, opened 2024-03-20
[docs] Consolidate logic for conversational RAG


ccurme:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Conversational RAG (e.g., LCEL analogues `ConversationalRetrievalChain`) is implemented across the docs to varying degrees of consistency. Users seeking to quickly get set up with conversational RAG might need to contend with these varying implementations.

We propose to update these implementations to use the abstractions used in [get_started/quickstart#conversation-retrieval-chain](https://python.langchain.com/docs/get_started/quickstart#conversation-retrieval-chain).

Known implementations:

- [ ] [get_started/quickstart#conversation-retrieval-chain](https://python.langchain.com/docs/get_started/quickstart#conversation-retrieval-chain)
- [ ] [use_cases/question_answering/chat_history](https://python.langchain.com/docs/use_cases/question_answering/chat_history)
- [ ] [expression_language/cookbook/retrieval#conversational-retrieval-chain](https://python.langchain.com/docs/expression_language/cookbook/retrieval#conversational-retrieval-chain)
- [ ] [use_cases/chatbots/quickstart](https://python.langchain.com/docs/use_cases/chatbots/quickstart)
- [ ] [use_cases/chatbots/retrieval](https://python.langchain.com/docs/use_cases/chatbots/retrieval)

PR #19349, opened 2024-03-20
docs: update use_cases/question_answering/chat_history

ccurme:
Update following https://github.com/langchain-ai/langchain/issues/19344


docs/docs/use_cases/question_answering/chat_history.ipynb
baskaryan, line 1: for the chain constructors would be nice to add comment specifying their input/output keys (eg create_retrieval_chain return "context", "answer" and inputs) and/or to link to their API references

and very small thing but for my sanity: you can now specify MessagesPlacholder variable_name as a positional arg, which is much nicer imo. you can also set `optional=True` so that you don't have to pass in an empty list when there's not messages to include, which is also much nicer imo

PR #19352, opened 2024-03-20
Community: Fixing incorrect base URLs for Azure Cognitive Search Retriever 

marlenezw:
This PR adds code to make sure that the correct base URL is being created for the Azure Cognitive Search retriever. At the moment an incorrect base URL is being generated. I think this is happening because the original code was based on a depreciated API version. No dependencies need to be added. I've also added more context to the test doc strings. 

I should also note that ACS is now Azure AI Search. I will open a separate PR to make these changes as that would be a breaking change and should potentially be discussed. 

Twitter: @marlene_zw

- No new tests added, however the current ACS retriever tests are now passing when I run them. 
- Code was linted.


farzad528:
LGTM

libs/community/tests/integration_tests/retrievers/test_azure_cognitive_search.py
```diff
@@ -7,22 +7,31 @@
 
 
 def test_azure_cognitive_search_get_relevant_documents() -> None:
-    """Test valid call to Azure Cognitive Search."""
+    """Test valid call to Azure Cognitive Search.
+
```
farzad528, line 11: NIT: we renanmed to Azure AI Search now, perhaps change all instances to "azure search"

farzad528 (approved):

libs/community/tests/integration_tests/retrievers/test_azure_cognitive_search.py
marlenezw: Yes! Planning to open a separate PR with all the changes for this. I saw some docs need to be updated too.

marlenezw:
Hi @baskaryan, could you have a look at this PR

baskaryan (approved):

PR #19369, opened 2024-03-21
docs: API Ref generation update

leo-gan:
Using local and external paths for module parsing. Renamed the hidden functions; added logging; enhanced the error handling.
Now it can generate `API Reference` not only from the packages placed in the local monorepo but also from the external packages (such as `langchain-google-el-carro`).
Related PR: #19290


efriis:
skipping - let's only build type B and C packages in the api ref. Type D are completely external, so hosting their API ref is up to them.

We already build api refs from e.g. langchain-google and langchain-nvidia.

If this change does anything else let me know!

PR #19381, opened 2024-03-21
community: fix bug in cohere that `async for` a coroutine in ChatCohere

kaijietti:
Without `await`, the `stream` returned from the `async_client` is actually a coroutine, which could not be used in `async for`.


baskaryan (commented):
cc @billytrend-cohere @BeatrixCohere

baskaryan (approved):

PR #19392, opened 2024-03-21
community[patch]: invoke callback prior to yielding token (llama.cpp)

sepiatone:
**Description:** Invoke callback prior to yielding token for llama.cpp
**Issue:** [Callback for on_llm_new_token should be invoked before the token is yielded by the model #16913](https://github.com/langchain-ai/langchain/issues/16913)
**Dependencies:** None


ccurme (approved):

ccurme:
thanks!

PR #19395, opened 2024-03-21
code[patch]: Add in code documentation to core Runnable `pipe` and `pick` methods (docs only)

liugddx:
- **Issue:**  #18804
-  @eyurtsev @ccurme PTAL


libs/core/langchain_core/runnables/base.py
```diff
@@ -438,12 +438,63 @@ def pipe(
         *others: Union[Runnable[Any, Other], Callable[[Any], Other]],
         name: Optional[str] = None,
     ) -> RunnableSerializable[Input, Other]:
-        """Compose this runnable with another object to create a RunnableSequence."""
+        """Compose this runnable with another object to create a RunnableSequence.
```
sepiatone, line 441: Perhaps a simpler example would suffice?
pipe() is just another way of doing prompt | model | output_parser as prompt.pipe(model, output_parser)

So something like
```
from langchain_community.llms.fake import FakeStreamingListLLM

prompt = PromptTemplate.from_template("This is a prompt")
model = FakeStreamingListLLM(responses = ["I'm a textbot"])
output_parser = StrOutputParser()

chain = prompt.pipe(model, output_parser)
print(chain.output_schema.schema())
```

I'm not sure how adding `RunnablePassthrough` helps here.

@liugddx Regardless, thanks for taking up this activity. I think I got a grasp of the Runnable methods by looking over your pull requests :)

libs/core/langchain_core/runnables/base.py
```diff
@@ -438,12 +438,63 @@ def pipe(
...
+        """
         return RunnableSequence(self, *others, name=name)
 
     def pick(self, keys: Union[str, List[str]]) -> RunnableSerializable[Any, Any]:
         """Pick keys from the dict output of this runnable.
-        Returns a new runnable."""
+        Returns a new runnable.
+        .. code-block:: python
+
+            from langchain_community.chat_models.fake import FakeListChatModel
```
ccurme, line 475: Similar comment here, I would try to keep the examples either useful for a real task or minimal, e.g.,
```python
from langchain_community.llms.fake import FakeStreamingListLLM
from langchain_core.output_parsers import JsonOutputParser, StrOutputParser

model = FakeStreamingListLLM(responses=['["a", "b", "c"]'])

chain = (
    model
    | {
        "text": StrOutputParser(),
        "json": JsonOutputParser(),
    }
)
type(chain.pick("json").invoke("bla"))

# list
```
consider stripping away some of the elements of this that aren't necessary (e.g., `stop` kwargs, `sleep`, prompt template, etc)

ccurme (commented):
Thank you for this!

baskaryan (approved):

PR #19399, opened 2024-03-21
core: Add async methods to BaseExampleSelector and SemanticSimilarityExampleSelector

cbornet:
Few-Shot prompt template may use a `SemanticSimilarityExampleSelector` that in turn uses a `VectorStore` that does I/O operations.
So to work correctly on the event loop, we need:
* async methods for the `VectorStore` (OK)
* async methods for the `SemanticSimilarityExampleSelector` (this PR)
* async methods for `BasePromptTemplate` and `BaseChatPromptTemplate` (future work)

### DIFF
--- a/libs/core/langchain_core/example_selectors/semantic_similarity.py
+++ b/libs/core/langchain_core/example_selectors/semantic_similarity.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type
 
+from langchain_core.documents import Document
 from langchain_core.example_selectors.base import BaseExampleSelector
 from langchain_core.pydantic_v1 import BaseModel, Extra
 from langchain_core.vectorstores import VectorStore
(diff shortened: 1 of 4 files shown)


eyurtsev (approved):

libs/core/langchain_core/example_selectors/semantic_similarity.py
```diff
@@ -37,34 +38,59 @@ class Config:
...
+    @staticmethod
+    def _example_to_text(
+        example: Dict[str, str], input_keys: Optional[List[str]]
+    ) -> str:
+        if input_keys:
+            return " ".join(sorted_values({key: example[key] for key in input_keys}))
+        else:
+            return " ".join(sorted_values(example))
+
+    def _documents_to_examples(self, documents: List[Document]) -> List[dict]:
```
eyurtsev, line 50: One question -- why a static method for example to text vs. _documents_to_examples being a regular method

Are implementation providers going to need to override documents to examples based on some state available in self?

cbornet:
> If so maybe a doc-string for those methods could be added to explain how to sub-class correctly and which methods are meant to be overriden?

Those methods are internal to the class and only there to factorize between sync and async. They are not meant to be overridden. Would you prefer a double-underscore ?

eyurtsev:
@cbornet i usually pull things out into functions if they're not meant to be overridden -- otherwise folks will end up overriding them, especially since we have a bunch of examples where the correct way to do the implementation is by overriding a private method :(

I'll put out the code into functions in a few

eyurtsev (approved):

eyurtsev:
Will do in a separate PR

PR #19407, opened 2024-03-21
community:Replace positional argument with text=text for cohere>=5 compatibility

billytrend-cohere:
- **Description:** Replace positional argument with text=text for cohere>=5 compatibility


giannis2two (approved):
👌

baskaryan (approved):

Issue #19222, opened 2024-03-18
LangChain GitHub: list_pull_request_files function doesn't work correctly when using a GitHub App


gdb-99:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
def list_pull_request_files(self, pr_number: int) -> List[Dict[str, Any]]:
        """Fetches the full text of all files in a PR. Truncates after first 3k tokens.
        # TODO: Enhancement to summarize files with ctags if they're getting long.

        Args:
            pr_number(int): The number of the pull request on Github

        Returns:
            dict: A dictionary containing the issue's title,
            body, and comments as a string
        """
        tiktoken = _import_tiktoken()
        MAX_TOKENS_FOR_FILES = 3_000
        pr_files = []
        pr = self.github_repo_instance.get_pull(number=int(pr_number))
        total_tokens = 0
        page = 0
        while True:  # or while (total_tokens + tiktoken()) < MAX_TOKENS_FOR_FILES:
            files_page = pr.get_files().get_page(page)
            if len(files_page) == 0:
                break
            for file in files_page:
                try:
                    file_metadata_response = requests.get(file.contents_url)
                    if file_metadata_response.status_code == 200:
                        download_url = json.loads(file_metadata_response.text)[
                            "download_url"
                        ]
                    else:
                        print(f"Failed to download file: {file.contents_url}, skipping")  # noqa: T201
                        continue

                    file_content_response = requests.get(download_url)
                    if file_content_response.status_code == 200:
                        # Save the content as a UTF-8 string
                        file_content = file_content_response.text
                    else:
                        print(  # noqa: T201
                            "Failed downloading file content "
                            f"(Error {file_content_response.status_code}). Skipping"
                        )
                        continue

                    file_tokens = len(
                        tiktoken.get_encoding("cl100k_base").encode(
                            file_content + file.filename + "file_name file_contents"
                        )
                    )
                    if (total_tokens + file_tokens) < MAX_TOKENS_FOR_FILES:
                        pr_files.append(
                            {
                                "filename": file.filename,
                                "contents": file_content,
                                "additions": file.additions,
                                "deletions": file.deletions,
                            }
                        )
[...]

PR #19433, opened 2024-03-22
community: fix list_pull_request_files function when working with a GitHub App

gdb-99:
**Description**
As noted in the issue I posted some days ago, I wasn't able to retrieve files from a pull request when using a GitHub App.
Turns out that the problem was that the GitHub App wasn't interacting at all with the GitHub API needed to retrieve the files.
To solve this, I had to create a **JSON Web Token** using the **App ID**, the **App's private key** and the **installation ID** (see the script below) of the application in the repository.
In this way, the Agent I was using was able to retrieve the files from the pull request correctly.

**Issue**
This PR solves the issue #19222 

**Dependencies**
- `jwt`, to create the JSON Web Token
- `os`, used to retrieve the three parameters needed from the `.env` file (alternatively, using `python-dotenv` should be equally correct).
- `time`, used by `jwt` for the calculation of the expiration time

**Additional file to retrieve Installation ID**
Below there is the script to retrieve the installation ID, specific for the GitHub App you are using and the repository in which it is installed.

``` python
import time
import jwt
import requests

app_id = "your_app_id"
private_key = open("path_to_pem_file").read()
# Example payload
payload = {
    "iat": int(time.time()),
    "exp": int(time.time()) + (10 * 60),  # JWT expiration time
    "iss": app_id
}

jwt_token = jwt.encode(payload, private_key, algorithm="RS256")

# Get installation ID
repository_owner = "your_repo_owner"
repository_name = "your_repo_name"
access_token_response = requests.get(
    f"https://api.github.com/repos/{repository_owner}/{repository_name}/installation",
    headers={
        "Authorization": f"Bearer {jwt_token}",
        "Accept": "application/vnd.github.v3+json"
    }
)

if access_token_response.status_code == 200:
    installation_id = access_token_response.json()["id"]
    print("Installation ID:", installation_id)	#Take it and save in your .env file
else:
    print("Failed to get installation ID:", access_token_response.text)
```

### DIFF
--- a/libs/community/langchain_community/utilities/github.py
+++ b/libs/community/langchain_community/utilities/github.py
@@ -3,6 +3,9 @@
 from __future__ import annotations
 
 import json
+import os
+import time
+import jwt
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 import requests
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/utilities/github.py
```diff
@@ -400,67 +403,82 @@ def get_issue(self, issue_number: int) -> Dict[str, Any]:
...
+                        app_id = os.getenv("GITHUB_APP_ID")
+                        private_key = open(os.getenv("GITHUB_APP_PRIVATE_KEY")).read()
+                        payload = {
+                            "iat": int(time.time()),
+                            "exp": int(time.time()) + (10 * 60),  # JWT expiration time
+                            "iss": app_id
+                        }
+                        jwt_token = jwt.encode(payload, private_key, algorithm="RS256")
+                        installation_id = os.getenv("INSTALLATION_ID")
+                        access_token_response = requests.post(
```
baskaryan, line 437: can this be done outside the for loop?

efriis:
closing as stale / no response

Issue #19415, opened 2024-03-21
index() API does not respect batch_size on vector_store.add_documents()


znwilkins:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
import logging
import os

from langchain.indexes import SQLRecordManager, index
from langchain.vectorstores.qdrant import Qdrant
from langchain_community.embeddings import CohereEmbeddings
from langchain_core.documents import Document
from qdrant_client import QdrantClient
from qdrant_client.models import Distance, VectorParams

DOCUMENT_COUNT = 100
COLLECTION_NAME = "test_index"
COHERE_EMBED_MODEL = os.getenv("COHERE_EMBED_MODEL")
COHERE_API_KEY = os.getenv("COHERE_API_KEY")
QDRANT_API_KEY = os.getenv("QDRANT_API_KEY")

# Setup embeddings and vector store
embeddings = CohereEmbeddings(model=COHERE_EMBED_MODEL, cohere_api_key=COHERE_API_KEY)
vectorstore = Qdrant(
    client=QdrantClient(url="http://localhost:6333", api_key=QDRANT_API_KEY),
    collection_name=COLLECTION_NAME,
    embeddings=embeddings,
)
# Init Qdrant collection for vectors
vectorstore.client.create_collection(
    collection_name=COLLECTION_NAME,
    vectors_config=VectorParams(size=1024, distance=Distance.COSINE),
)
# Init the record manager using SQLite
namespace = f"qdrant/{COLLECTION_NAME}"
record_manager = SQLRecordManager(
    namespace, db_url="sqlite:///record_manager_cache.sql"
)
record_manager.create_schema()
# Init 100 example documents
documents = [Document(page_content=f"example{i}", metadata={"source": f"example{i}.txt"}) for i in range(DOCUMENT_COUNT)]

# Log at the INFO level so we can see output from httpx
logging.basicConfig(level=logging.INFO)

# Index 100 documents with a batch size of 100.
# EXPECTED: 1 call to Qdrant with 100 documents per call
# ACTUAL  : 2 calls to Qdrant with 64 and 36 documents per call, respectively
result = index(
    documents,
    record_manager,
    vectorstore,
[...]

PR #19443, opened 2024-03-22
langchain: Passthrough batch_size on index()/aindex() calls

znwilkins:
**Description:** This change passes through `batch_size` to `add_documents()`/`aadd_documents()` on calls to `index()` and `aindex()` such that the documents are processed in the expected batch size.
**Issue:** #19415
**Dependencies:** N/A
**Twitter handle:** N/A

### DIFF
--- a/libs/langchain/langchain/indexes/_api.py
+++ b/libs/langchain/langchain/indexes/_api.py
@@ -330,7 +330,7 @@ def index(
         # Be pessimistic and assume that all vector store write will fail.
         # First write to vector store
         if docs_to_index:
-            vector_store.add_documents(docs_to_index, ids=uids)
+            vector_store.add_documents(docs_to_index, ids=uids, batch_size=batch_size)
             num_added += len(docs_to_index) - len(seen_docs)
             num_updated += len(seen_docs)
 
@@ -544,7 +544,9 @@ async def aindex(
         # Be pessimistic and assume that all vector store write will fail.
         # First write to vector store
         if docs_to_index:
-            await vector_store.aadd_documents(docs_to_index, ids=uids)
+            await vector_store.aadd_documents(
+                docs_to_index, ids=uids, batch_size=batch_size
+            )
             num_added += len(docs_to_index) - len(seen_docs)
             num_updated += len(seen_docs)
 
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/indexes/_api.py
```diff
@@ -330,7 +330,7 @@ def index(
         # Be pessimistic and assume that all vector store write will fail.
         # First write to vector store
         if docs_to_index:
-            vector_store.add_documents(docs_to_index, ids=uids)
+            vector_store.add_documents(docs_to_index, ids=uids, batch_size=batch_size)
```
ccurme, line 333: isn't `docs_to_index` already batched, as this logic is inside the `for doc_batch in _batch(batch_size, doc_iterator):` loop?

znwilkins: Yes, but the batch size is not necessarily respected by the underlying vector store in the situation where the vector store's default batch size is less than the `batch_size` specified in the call to `index()`.

For example, if the vector store's default batch size is 64, but I call `index(..., batch_size=100)`, then the loop is batching in groups of 100, but sending to my vector store in groups of 64. I would expect that when I say "batch size is 100", I send 100 documents to my vector store per call (assuming these 100 documents have never been seen by the record manager).

ccurme:
Thank you! This prompted me to check through our integrations; I found one for which I think this could result in unexpected kwargs. Fixing that in https://github.com/langchain-ai/langchain/pull/19519. Otherwise LGTM

ccurme (approved):

eyurtsev:
We need to revert this change -- batch_size is not part of the vectorstore standard API

eyurtsev:
actually we should check whether all vectorstores support **kwargs in the interface

znwilkins:
> actually we should check whether all vectorstores support **kwargs in the interface

Looks like it:

```python
class VectorStore(ABC):
    """Interface for vector store."""

# ...

    def add_documents(self, documents: List[Document], **kwargs: Any) -> List[str]:

# ...

    async def aadd_documents(
        self, documents: List[Document], **kwargs: Any
    ) -> List[str]:
```

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores.py#L126
https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores.py#L141

ccurme:
I grepped through all vectorstores in langchain_community/vectorstores and ended up finding one that didn't, so I fixed it in https://github.com/langchain-ai/langchain/pull/19519 before we merged this.

Issue #11906, opened 2023-10-17
Enable token usage count for AWS Bedrock


alex-ld:
### Feature request

There should be a callback handler like [OpenAICallbackHandler](https://github.com/langchain-ai/langchain/blob/31f264169db4ab23689f2e179983f1cfdfd1a33a/libs/langchain/langchain/callbacks/openai_info.py#L120) for AWS Bedrock models, so that we can easily get the token usage and monitor cost.

It looks like the AWS Bedrock API doesn't return with token usage. Unlike OpenAI.

Can we have similar feature, perhaps [using the internal Anthropic tokenizer to count](https://github.com/langchain-ai/langchain/blob/31f264169db4ab23689f2e179983f1cfdfd1a33a/libs/langchain/langchain/llms/bedrock.py#L405)?

### Motivation

Cost monitoring

### Your contribution

PR

krrishdholakia:
Hey @alex-ld you can do this via ChatLiteLLM - all responses return token usage. You can also use our cost calculator functions to calculate cost for each call. 

```python
from langchain.chat_models import ChatLiteLLM
from langchain.schema import AIMessage, HumanMessage, SystemMessage

chat = ChatLiteLLM(model="bedrock/anthropic.claude-v2")

messages = [
    HumanMessage(
        content="Translate this sentence from English to French. I love programming."
    )
]
chat(messages)
```

Calculate cost: https://docs.litellm.ai/docs/completion/token_usage#5-completion_cost

```python
from litellm import completion_cost
cost = completion_cost(model="bedrock/anthropic.claude-v2", prompt="Hey!", completion="How's it going?")
formatted_string = f"${float(cost):.10f}"
print(formatted_string)
```

Langchain Docs: https://python.langchain.com/docs/integrations/chat/litellm
LiteLLM Docs: https://docs.litellm.ai/docs/providers/bedrock

VBoB13:
> You can also use our cost calculator functions to calculate cost for each call.

I might be too much of a beginner to notice, but it doesn't look like this implementation supports the use of Agents that utilizes Bedrock LLMs..
Are there any plans to support this soon?
The context manager for OpenAI that LangChain provides is very straight forward, and it'd be awesome to see something similar for Bedrock models.

nggih:
https://how.wtf/how-to-count-amazon-bedrock-anthropic-tokens-with-langchain.html I have the same problem but mine is specific on Anthropic Claude usage, I found the solution on the link

[...]

PR #19448, opened 2024-03-22
community: add token usage to bedrock models' responses if available

AlexisGado:
- **Description:** Get token usage (if available) from the response body of bedrock model responses and pass it to the `llm_output` field of `ChatGeneration`
- **Issue:** Solves #11906 for some models


AlexisGado:
@baskaryan @3coins, tiny reminder! Thanks 😄

dmenini:
Hi, I opened a similar PR a couple of weeks ago: https://github.com/langchain-ai/langchain/pull/19127#issuecomment-2019208568

Here I'm using the Bedrock headers to extract the token counts, which should work across different providers (hopefully). Maybe it would be great if you could confirm that such headers are also available for meta models (unfortunately I only have access to anthropic models). Thanks!

baskaryan:
thanks @AlexisGado! as @dmenini #19127 implements same functionality and just landed — ok to close this pr?

hwchase17:
these models are now in their own integration packages

PR #19449, opened 2024-03-22
community:Modified regular expression

igeni:
- **Description:** Modified regular expression to add support for unicode chars and simplify pattern


eyurtsev (dismissed):
Does this fix an issue? if not, the former version is more widely readable

eyurtsev:
Ah this is to fix an issue with unicode in folder names?

baskaryan (approved):

efriis (approved):

PR #19485, opened 2024-03-24
community: fix imports to work with the latest version (YandexGPT)

egorkrash:
community: fix of Yandex GPT integration. 

After version v0.263.0 Yandex modified file names a bit so when using `ChatYandexGPT` or `YandexGPTEmbeddings` the following error (caused by unsuccessful imports like [these](https://github.com/langchain-ai/langchain/blob/b617085af0d14468d5418176d4235229c2c12ffa/libs/community/langchain_community/chat_models/yandex.py#L173)) always occurs:

```ImportError: Please install YandexCloud SDK with pip install yandexcloud or upgrade it to recent version.```

The error cannot be solved by upgrading `yandexcloud`, but rather by downgrading. This PR fixes the imports for chat and embedding models (according to [v0.267.0](https://github.com/yandex-cloud/python-sdk/tree/v0.267.0/yandex/cloud/ai/foundation_models/v1) so that the latest version of `yandexcloud` works properly).

New tests or documentation are not needed.

Twitter: @egor_krash


baskaryan (commented):
will this break the integration for anyone using older versions of yandex?

egorkrash:
> will this break the integration for anyone using older versions of yandex?

It will raise the same error as above with the message `Please install YandexCloud SDK with pip install yandexcloud or upgrade it to recent version.` which now seems more logical. Or it might be worth creating a check for the current version and making imports accordingly. Let me know what you think @baskaryan

baskaryan:
believe #19341 fixes issue in a backwards compatible way. closing this pr but let me know if i'm missing something!

mkhludnev:
Hi there. 
I'm on langchain==0.1.13 
Yandex Embedding crashed with
 `ModuleNotFoundError: No module named 'yandex.cloud.ai.foundation_models.v1.foundation_models_service_pb2'`
I have to figure out concrete version via blame
  `pip install yandexcloud==0.248.0`
@tyumentsev4, wondering if there is a chance to improve usability of this library.

egorkrash:
> Hi there. I'm on langchain==0.1.13 Yandex Embedding crashed with `ModuleNotFoundError: No module named 'yandex.cloud.ai.foundation_models.v1.foundation_models_service_pb2'` I have to figure out concrete version via blame `pip install yandexcloud==0.248.0` @tyumentsev4, wondering if there is a chance to improve usability of this library.

Hi Mikhail,
I guess the fix has already been pushed to the master, it should be released anytime soon.

PR #19499, opened 2024-03-25
community[minor]: allow enabling proxy in aiohttp session in AsyncHTML

Spycsh:

- [x] **PR title**: "enable proxy in aiohttp.ClientSession"

- [x] **Description:** in `document_loaders/async_html.py` an `aiohttp.ClientSession` is created but it cannot capture the proxy setting automatically in the environment. This will cause the document fetching to hang forever. This issue can be simply fixed by setting `trust_env=True`. There is a link https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession to illustrate this parameter.

Additional guidelines:


libs/community/langchain_community/document_loaders/async_html.py
```diff
@@ -142,7 +142,7 @@ def _scrape(
     async def _fetch(
         self, url: str, retries: int = 3, cooldown: int = 2, backoff: float = 1.5
     ) -> str:
-        async with aiohttp.ClientSession() as session:
+        async with aiohttp.ClientSession(trust_env=True) as session:
```
eyurtsev, line 145: @Spycsh hello thank you for the contribution. 

We do want folks to be able to use a proxy for making web requests since that's much safer to protect users against SSRF requests etc.

So this change is in the right direction, but I don't fully understand the security ramifications and i'm worried that this code will mitigate one security issue but will result in another CVE being filed against the code base.

Basically, the `trust_env` API and is disabled by default. Why is that the case?

Would you mind:

1) Exposing this as a configuration parameter in init
2) Documenting this parameter briefly + adding a link to source documentation
3) I can follow up and add a security note to explain more

I haven't used .netrc before, so a bit unclear on whether the credentials are sent to authenticate to the proxy vs. being sent to authenticate to whatever content we're downloading.

My sense is that it can be either? If so, are the credentials encrypted when sent over the proxy?

Spycsh: @eyurtsev Thanks for reply. Yes I understand that you have some security concern about this trust_env parameter. I am not sure whether the security issue surely exist? Here is a link already used trust_env in the codebase: https://github.com/langchain-ai/langchain/blob/ad77fa15eec4dae431171b7e8c13b8c1f9edec98/libs/community/langchain_community/embeddings/infinity.py#L307. If there is a CVE issue I think it should already have been reported?

Another issue that you mention is the .netrc file -- I also have not used that before. The documentation https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession states that the proxies are read from .netrc file but I think it is not completely the case, and the `trust_env` actually read the proxies somewhere else . Here is the source code of aiohttp that reading the proxies https://github.com/aio-libs/aiohttp/blob/aa014a9c1084368a01a79f09bb238486e0aa164a/aiohttp/client.py#L1123. The documentation says that it not only look at the .netrc but also check the **HTTP_PROXY** and **HTTPS_PROXY** environment variables. So for users that explicitly set http_proxy and https_proxy env variables, they must set trust_env here to let the aiohttp.ClientSession to use the proxies.

Back to your suggestion about exposing this as a configuration parameter in init. Could you please have a look at [here](https://github.com/langchain-ai/langchain/blob/ad77fa15eec4dae431171b7e8c13b8c1f9edec98/libs/community/langchain_community/document_loaders/async_html.py#L79C24-L80C1)? In this async_html.py file there is a `requests.Session()`. It respects the proxies environment variables. When it does `session.get()` it will automatically read the proxies in environment variables, **even** when there are no proxies passed in the code here (self.session.proxies={}). For the same reason, regarding to another session in this file,  the `aiohttp.ClientSession`, we should align it with the former `requests.Session()` behaviors, right? The fix is to set `trust_env` explicitly for `aiohttp.ClientSession` to let it respect the proxies in the env variables. Overall, I think if it is to align the default behaviors (automatically read proxy env variables set by users), the current fix is enough. However if you think we need to align the explicit proxy passing in the code (like [here](https://github.com/langchain-ai/langchain/blob/ad77fa15eec4dae431171b7e8c13b8c1f9edec98/libs/community/langchain_community/document_loaders/async_html.py#L83), I think we can add in this PR another proxy string [here](https://github.com/langchain-ai/langchain/blob/ad77fa15eec4dae431171b7e8c13b8c1f9edec98/libs/community/langchain_community/document_loaders/async_html.py#L148), but need some other untidy alignment (e.g. self.session.proxies is a dict but aiohttp.ClientSession.get expects a string or url).

Thanks for further suggestions!

eyurtsev: Sorry I missed the response. The main issue is that users will have no idea what trust_env or how to use it (e.g., it's poorly documented even in the [client libraries](https://docs.python-requests.org/en/latest/api/#requests.Session.trust_env)

Would you mind exposing in init, with default trust_env = False and documenting:
1) what this does
2) why one would want to use it? 
3) How to use it (e.g., which env variables need to be set etc.)

Spycsh: Hello @eyurtsev , thanks for suggestions. As what you point out, I add a parameter in init, with default trust_env=False and add relevant doc in `WebResearchRetriever`, the main entry that calls this `AsyncHtmlLoader`. In the `async_html.ipynb` I also add the reason and info how to use it. Welcome to any further suggestions!

eyurtsev (approved):

Issue #17587, opened 2024-02-15
AzureSearch.semantic_hybrid_search_with_score_and_rerank not working when METADATA_FIELD not in index


solrac156:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

When you instantiate an AzureSearch instance with a search type of `semantic_hybrid` and the metadata is not on the list of fields for the `fields` of the vector store or in the index, the `semantic_hybrid_search_with_score_and_rerank` method fails.

### Error Message and Stack Trace (if applicable)

```
Traceback (most recent call last):
  File "/Users/carlos/PycharmProjects/project/backend/api/app/controllers/session/send_message.py", line 76, in send_message
    response = agent({"input": prompt})
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/carlos/PycharmProjects/project/backend/api/services/agent.py", line 166, in __call__
    return self._executor.invoke(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/carlos/PycharmProjects/project/.venv/lib/python3.11/site-packages/langchain/chains/base.py", line 162, in invoke
    raise e
  File "/Users/carlos/PycharmProjects/project/.venv/lib/python3.11/site-packages/langchain/chains/base.py", line 156, in invoke
    self._call(inputs, run_manager=run_manager)
  File "/Users/carlos/PycharmProjects/project/.venv/lib/python3.11/site-packages/langchain/agents/agent.py", line 1391, in _call
    next_step_output = self._take_next_step(
                       ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/carlos/PycharmProjects/project/.venv/lib/python3.11/site-packages/langchain/agents/agent.py", line 1097, in _take_next_step
    [
  File "/Users/carlos/PycharmProjects/project/.venv/lib/python3.11/site-packages/langchain/agents/agent.py", line 1097, in <listcomp>
    [
  File "/Users/carlos/PycharmProjects/project/.venv/lib/python3.11/site-packages/langchain/agents/agent.py", line 1182, in _iter_next_step
    yield self._perform_agent_action(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #19502, opened 2024-03-25
FIX "metadata" hardcoded value in semantic_hybrid_search_with_score_and_rerank

alexmanie:
- **Description:** This PR is to fix a bug in semantic_hybrid_search_with_score_and_rerank() function in langchain_community/vectorstores/azuresearch.py. The hardcoded "metadata" name is replaced with FIELDS_METADATA variable with an if block to check if the metadata column exists or not.
- **Issue:** #17587 
- **Dependencies:** No
- **Twitter handle:** None


baskaryan (approved):

baskaryan:
@alexmanie believe #18938 fixes this, let me know if i'm missing something

PR #17953, opened 2024-02-22
community: add BigDL-LLM integrations

shane-huang:
- **Description**: [`bigdl-llm`](https://github.com/intel-analytics/BigDL) is a library for running LLM on Intel XPU (from Laptop to GPU to Cloud) using INT4/FP4/INT8/FP8 with very low latency (for any PyTorch model). This PR adds bigdl-llm integrations to langchain. 
- **Issue**: NA
- **Dependencies**: `bigdl-llm` library
- **Contribution maintainer**: @shane-huang 
 
Examples added:
- docs/docs/integrations/llms/bigdl.ipynb

PR #19518, opened 2024-03-25
community: migrate `bigdl-llm` to `ipex-llm`

shane-huang:
- **Description**:  `bigdl-llm` library has been renamed to [`ipex-llm`](https://github.com/intel-analytics/ipex-llm). This PR migrates the `bigdl-llm` integration to `ipex-llm` .
- **Issue**: N/A. The original PR of `bigdl-llm` is https://github.com/langchain-ai/langchain/pull/17953 
- **Dependencies**: `ipex-llm` library
- **Contribution maintainer**: @shane-huang

Updated doc:   docs/docs/integrations/llms/ipex_llm.ipynb
Updated test:  libs/community/tests/integration_tests/llms/test_ipex_llm.py


libs/community/langchain_community/llms/ipex_llm.py
```diff
@@ -10,22 +10,22 @@
 logger = logging.getLogger(__name__)
 
 
-class BigdlLLM(LLM):
```
baskaryan, line 13: for backwards compatibility can we keep around a bigdl.py file and does 
```python
from langchain_community.llms.ipex_llm import IPEXLLM

BigdlLLM = IPEXLLM
```

that way existing code that uses this integration won't break

libs/community/langchain_community/llms/ipex_llm.py
```diff
@@ -10,22 +10,22 @@
...
-    """Wrapper around the BigDL-LLM Transformer-INT4 model
+class IPEXLLM(LLM):
+    """Wrapper around the IPEX-LLM model
 
     Example:
         .. code-block:: python
 
-            from langchain.llms import TransformersLLM
-            llm = TransformersLLM.from_model_id(model_id="THUDM/chatglm-6b")
+            from langchain.llms import IPEXLLM
```
baskaryan, line 19: from langchain_community.llms import IPEXLLM

baskaryan: also you'll need to update `langchain_community/llms/__init__.py` for this to work

baskaryan (commented):
can we add a redirect to `docs/vercel.json` to point from old url to new url

libs/community/langchain_community/llms/ipex_llm.py
shane-huang: Thanks for the comment. I could not use `BigdlLLM = IPEXLLM` as the imports are different. Instead I inherite from IpexLLM and override necessary methods.

shane-huang:
> can we add a redirect to `docs/vercel.json` to point from old url to new url

added the url redirection

shane-huang:
@baskaryan Please kindly let me know if there's anything else I need to change.

shane-huang:
@baskaryan Could you kindly advise why Vercel failed after removing the redundant url redirect and how can I fix it?

PR #19520, opened 2024-03-25
infra: Consistent lxml requirements

### LINKED ISSUE #19040: Inconsistent lxml dependency between `langchain` and `langchain-text-splitters`
### Checked other resources

### Example Code

```
conda install langchain=0.1.12 zeep=4.2.1
```

```
Could not solve for environment specs
The following packages are incompatible
├─ langchain 0.1.12**  is installable and it requires
│  ├─ langchain-text-splitters >=0.0.1,<0.1 , which requires
│  │  └─ lxml >=5.1.0,<6.0.0 , which can be installed;
│  └─ lxml >=4.9.2,<5.0.0 , which can be installed;
└─ zeep 4.2.1**  is not installable because it requires
   └─ lxml >=4.6.0 , which conflicts with any installable versions previously reported.
```

### Description

I'm trying to install latest langchain along with some other packages depending on lxml. It appears that `langchain` requires `lxml >=4.9.2,<5.0.0` but `langchain-text-splitters` requires `lxml >=5.1.0,<6.0.0` which are incompatible. Since the former depends on the latter, it makes dependency resolution fail.

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.19045
> Python Version:  3.10.13 | packaged by conda-forge | (main, Dec 23 2023, 15:27:34) [MSC v.1937 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 0.1.31
> langchain: 0.1.12
> langchain_community: 0.0.28
> langsmith: 0.1.24
> langchain_text_splitters: 0.0.1

Packages not installed (Not Necessarily a Problem)
--------------------------------------------------
The following packages were not found:

> langgraph
> langserve

Randl:
Update the dependency for lxml to be consistent among different packages; should fix https://github.com/langchain-ai/langchain/issues/19040


tjaffri:
https://github.com/docugami/dgml-utils/pull/24 is merged and published as dgml-utils 0.3.3. Let me know if we can help resolve downstream issues. FYI @kenzie28.

eyurtsev:
why are there so many changes across other lock files?

eyurtsev:
Another question is whether we need to be pinning to a particular lxml major version. Would we be able to expand the range of major versions we're supporting to users don't get stuck unable to use langchain or some other package because of lxml version mismatch (lxml is a popular package)

Randl:
@eyurtsev I agree, we can settle for >4.9,<6

PR #19525, opened 2024-03-25
langchain[patch]: fix callbacks getattr

baskaryan:
Currently if you do
```python
from langchain.callbacks import base
```
~it will import `langchain_community.callbacks.base` instead of `langchain.callbacks.base`.~
it still imports from langchain but it also raises the community warning

Will need to make this patch to all module getattr methods

### DIFF
--- a/libs/langchain/langchain/callbacks/__init__.py
+++ b/libs/langchain/langchain/callbacks/__init__.py
@@ -6,6 +6,7 @@
 
     BaseCallbackHandler --> <name>CallbackHandler  # Example: AimCallbackHandler
 """
+import importlib
 import warnings
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/callbacks/__init__.py
```diff
@@ -30,6 +31,10 @@
 
 
 def __getattr__(name: str) -> Any:
+    try:
+        return importlib.import_module("." + name, __name__)
```
efriis, line 35: I think we need a lookup table here defining the things to export

baskaryan: bc there might be things we don't want to export?

efriis: this kind of use could be an attack vector. Better to lock down and only export the things we explicitly want exported.

e.g. someone could find a way to resolve to non-langchain libs, relative `..` imports, etc.

eyurtsev: ^ +1 exactly what i was thinking

PR #17250, opened 2024-02-08
core[minor]: used defusedxml for xml parsing

0xanis:
- **Dependencies:** defusedxml

PR #19526, opened 2024-03-25
core[patch]: Use defusedxml in XMLOutputParser

eyurtsev:
This mitigates a security concern for users still using older versions of
libexpat that causes an attacker to compromise the availability of the system
if an attacker manages to surface malicious payload to this XMLParser.


eyurtsev:
adds streaming path: https://github.com/langchain-ai/langchain/pull/17250

baskaryan (approved):

PR #19527, opened 2024-03-25
community[patch]: clickhouse make it possible to not specify index

efriis:


mneedham:
@efriis I think the lint error would be fixed by updating line 205 to read:

```python
def _schema(self, dim: int, index_params: Optional[Union[List, Dict]]) -> str:
```

libs/community/langchain_community/vectorstores/clickhouse.py
```diff
@@ -205,10 +197,53 @@ def __init__(
...
+        CREATE TABLE IF NOT EXISTS {self.config.database}.{self.config.table}(
+            {self.config.column_map['id']} Nullable(String),
+            {self.config.column_map['document']} Nullable(String),
+            {self.config.column_map['embedding']} Array(Float32),
+            {self.config.column_map['metadata']} JSON,
+            {self.config.column_map['uuid']} UUID DEFAULT generateUUIDv4(),
+            CONSTRAINT cons_vec_len CHECK length(
+                {self.config.column_map['embedding']}) = {dim},
+            INDEX vec_idx {self.config.column_map['embedding']} TYPE \
+        {self.config.index_type}({index_params}) GRANULARITY 1000
```
efriis, line 231: @mneedham this seems wrong to me and happy to take your advice as clickhouse expert!

This will substitute in a python repr of a List/Dict, or "None" if `index_params` is None.

Maybe index params should be an optional str, and if it's none, we fill it out as "" instead of "None"?

mneedham: @efriis Yeah, I think you're right. I tried setting the line to the following:

> def _schema(self, dim: int, index_params: Optional[str]) -> str:

And now the linter is happy for me locally:

```
$ make lint_package
./scripts/check_pydantic.sh .
./scripts/lint_imports.sh
poetry run ruff .
[ "langchain_community" = "" ] || poetry run ruff format langchain_community --diff
1038 files already formatted
[ "langchain_community" = "" ] || poetry run ruff --select I langchain_community
[ "langchain_community" = "" ] || mkdir -p .mypy_cache && poetry run mypy langchain_community --cache-dir .mypy_cache
Success: no issues found in 1038 source files
```

mneedham:
I think this patch will get the builds passing.

[lint.patch](https://github.com/langchain-ai/langchain/files/14885921/lint.patch)

efriis:
Hey @mneedham I'm pretty sure this is an implementation issue that the linter is properly catching. It appears you're substituting `None` values into a query that gets run against clickhouse. Updating the function's typing may get it passing, but it'll be a bug.

Would you be game to reopen this PR with those things fixed, referencing this one? Ideally from a personal branch so we can collaborate if it needs further work :) https://github.com/langchain-ai/langchain/pull/18381#issuecomment-2018734749

PR #19531, opened 2024-03-25
[Community] Stop spamming SIMD import warnings

hinthornw:
If you use an embedding dist function in an eval loop, you get warned every time. Would prefer to just check once and forget about it.


baskaryan (commented):
could also/instead lower to logger.debug?

baskaryan (approved):

PR #19533, opened 2024-03-25
cohere[patch]: Add additional kwargs support for Cohere SDK params

giannis2two:
* Adds support for `additional_kwargs` in `get_cohere_chat_request`
  * This functionality passes in Cohere SDK specific parameters from `BaseMessage` based classes to the API


libs/partners/cohere/langchain_cohere/chat_models.py
```diff
@@ -85,10 +86,10 @@ def get_cohere_chat_request(
         "chat_history": [
             {"role": get_role(x), "message": x.content} for x in messages[:-1]
         ],
+        **additional_kwargs, # params below will overwrite these values
         "documents": documents,
-        "connectors": maybe_connectors,
+        "connectors": connectors,
         "prompt_truncation": prompt_truncation,
```
harry-cohere, line 92: we still need **kwargs, I think! Otherwise a lot of the parameters are dropped.

harry-cohere (approved):

libs/partners/cohere/langchain_cohere/chat_models.py
giannis2two: The reason for passing in `**kwargs` in the return is exactly to forward any Cohere SDK params to the API. However those params are coming from `**additional_kwargs` (aka `documents`, `connectors`, `prompt_truncation`) that are pre-processed in this occasion), I believe `**kwargs` never contained those parameters, and hence removed.

Will defer to LC team's take on this as well however 🙏.

giannis2two: Comments are deprecated. tldr: move `kwargs` back in.

libs/partners/cohere/tests/integration_tests/test_chat_models.py
```diff
@@ -3,15 +3,15 @@
 
 
 def test_stream() -> None:
-    """Test streaming tokens from OpenAI."""
+    """Test streaming tokens from ChatCohere."""
```
giannis2two, line 6: leftover fix from previous PR

efriis (approved):

libs/partners/cohere/langchain_cohere/chat_models.py
```diff
@@ -60,33 +61,32 @@ def get_cohere_chat_request(
...
+    additional_kwargs = messages[-1].additional_kwargs
+
+    # cohere SDK will fail loudly if both connectors and documents are provided
+    formatted_docs = [
+        {
+            "text": doc.page_content,
+            "id": doc.metadata.get("id") or f"doc-{str(i)}",
+        }
+        for i, doc in enumerate(additional_kwargs.get("documents", []))
+    ] or documents
```
efriis, line 73: should we throw an error if both `documents` and `additional_kwargs["documents"]` is set?

giannis2two: Added. Good suggestion.

PR #19537, opened 2024-03-25
community: Implement DirectoryLoader lazy_load function

DasDingoCodes:

- [x] **PR title**: "community: Implement DirectoryLoader lazy_load function"

- [x] **Description**: The `lazy_load` function of the `DirectoryLoader` yields each document separately. If the given `loader_cls` of the `DirectoryLoader` also implemented `lazy_load`, it will be used to yield subdocuments of the file.

  1. a test for the integration, preferably unit tests that do not rely on network access: `libs/community/tests/unit_tests/document_loaders/test_directory_loader.py`
  2. an example notebook showing its use. It lives in `docs/docs/integrations` directory: `docs/docs/integrations/document_loaders/directory.ipynb`

Additional guidelines:


libs/community/langchain_community/document_loaders/directory.py
```diff
@@ -197,3 +198,55 @@ def load(self) -> List[Document]:
             pbar.close()
 
         return docs
+
+    def lazy_load(self) -> Iterator[Document]:
```
eyurtsev, line 202: Great if we're upgrading to lazy_load() could you delete the load() implementation? (it will proxy to lazy_load automatically)

libs/community/langchain_community/document_loaders/directory.py
```diff
@@ -197,3 +198,55 @@ def load(self) -> List[Document]:
...
+                randomizer = random.Random(
+                    self.sample_seed if self.sample_seed else None
+                )
+                randomizer.shuffle(items)
+            items = items[: min(len(items), self.sample_size)]
+
+        for i in items:
+            yield from self.lazy_load_file(i, p)
+
+    def lazy_load_file(self, item: Path, path: Path) -> Iterator[Document]:
```
eyurtsev, line 228: Could you make this private? This is a non standard part of the document loader API so generally we do not expect users to be relying on it.

Could you also update load_file to proxy to this method, so we can avoid code duplication?

eyurtsev (changes requested):
lazy_load needs to be on feature parity with load() or needs to fail loudly if it's not. 

e.g., we should add support for the progress bar  -- this can work by first running the glob to quickly count the files

libs/community/langchain_community/document_loaders/directory.py
DasDingoCodes: Deleted the previous load_file function, and made load proxy to lazy_load

libs/community/langchain_community/document_loaders/directory.py
DasDingoCodes: Renamed lazy_load_file to _lazy_load_file. Made load proxy to lazy_load. load_file wasn't used anymore, so I deleted it.

eyurtsev (approved):

eyurtsev:
For the documentation, file directory loader is documented here:

`docs/docs/modules/data_connection/document_loaders/file_directory.mdx`

3 options:

1) Replace the contents of the original documentation with the new documentation (potentially include any missing stuff) in this PR
2) same as above but in a separate PR
3) Don't do documentation since it's already documented

PR #19574, opened 2024-03-26
community[minor]: Add support for non-file-based Document Loaders in PebbloSafeLoader

Raj725:
**Description:**
PebbloSafeLoader: Add support for non-file-based Document Loaders

This pull request enhances PebbloSafeLoader by introducing support for several non-file-based Document Loaders. With this update, PebbloSafeLoader now seamlessly integrates with the following loaders:
- GoogleDriveLoader
- SlackDirectoryLoader
- Unstructured EmailLoader

**Issue:** NA
**Dependencies:** - None
**Twitter handle:** @Raj__725

### DIFF
--- a/libs/community/langchain_community/utilities/pebblo.py
+++ b/libs/community/langchain_community/utilities/pebblo.py
@@ -29,11 +29,28 @@ file_loader = [
     "AmazonTextractPDFLoader",
     "CSVLoader",
     "UnstructuredExcelLoader",
+    "UnstructuredEmailLoader",
 ]
-dir_loader = ["DirectoryLoader", "S3DirLoader", "PyPDFDirectoryLoader"]
-in_memory = ["DataFrameLoader"]
+dir_loader = [
+    "DirectoryLoader",
+    "S3DirLoader",
+    "SlackDirectoryLoader",
+    "PyPDFDirectoryLoader",
+    "NotionDirectoryLoader",
+]
 
-LOADER_TYPE_MAPPING = {"file": file_loader, "dir": dir_loader, "in-memory": in_memory}
+in_memory = ["DataFrameLoader"]
+remote_db = [
+    "NotionDBLoader",
+    "GoogleDriveLoader",
+]
+
+LOADER_TYPE_MAPPING = {
+    "file": file_loader,
+    "dir": dir_loader,
+    "in-memory": in_memory,
+    "remote_db": remote_db,
+}
 
 SUPPORTED_LOADERS = (*file_loader, *dir_loader, *in_memory)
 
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/utilities/pebblo.py
```diff
@@ -188,15 +206,21 @@ def get_loader_full_path(loader: BaseLoader) -> str:
                 location = f"gc://{loader.bucket}/{loader.blob}"
             elif isinstance(loader, S3FileLoader):
                 location = f"s3://{loader.bucket}/{loader.key}"
+        elif "source" in loader_dict:
+            location = loader_dict.get("source", "-")
+            if "channel" in loader_dict:
+                location = location + "/" + loader_dict.get("channel", "-")
         elif "path" in loader_dict:
-            location = loader_dict["path"]
+            location = loader_dict.get("path", "-")
```
eyurtsev, line 214: Why is `-` a default type? Why are we using `get` if the value has already been confirmed to be inside loader_dict?

Raj725: Thanks for your input @eyurtsev! 
The redundant use of - has been addressed, and adjustments were made to resolve linting errors. 
Changes have been reverted to enhance code quality and readability.
Please review

libs/community/langchain_community/utilities/pebblo.py
```diff
@@ -159,7 +176,7 @@ def get_loader_type(loader: str) -> str:
     for loader_type, loaders in LOADER_TYPE_MAPPING.items():
         if loader in loaders:
             return loader_type
-    return "unknown"
+    return "unsupported"
```
eyurtsev, line 179: I don't know the API, but return a string type on an unkown loader feels like the wrong pattern -- should probably raise or return a None. Anyway going to merge as is feel free to update the implementation if you want in a separate PR

eyurtsev (approved):

libs/community/langchain_community/utilities/pebblo.py
Raj725: Thank you for your feedback @eyurtsev. We appreciate your input on the pattern used for handling unknown loaders. Just to clarify, Pebblo is owned by our company Daxa.ai, and it is designed to support all loader types.

The reason for returning a string instead of raising an error or returning None for unknown loaders is to ensure that users are not restricted from using Pebblo with loaders that have not been explicitly tested by us. By silently returning a string, we aim to provide a seamless experience for users while still indicating that the loader may not be officially supported.

Raj725: we understand your concerns and will take them into consideration for future updates.

eyurtsev:
Can merge as soon as unit tests pass

Raj725:
> Can merge as soon as unit tests pass

Thanks

Raj725:
> Can merge as soon as unit tests pass

Fixed lint error; all checks are green now.

PR #19575, opened 2024-03-26
langchain[major]: Use InMemoryVectorStore by default in VectorstoreIndexCreator

cbornet:
This is a small breaking change but I think it should be done as:
* No external dependency needs to be installed anymore for the default to work
* It is vendor-neutral


eyurtsev (approved):

eyurtsev:
@baskaryan I am in favor of this. This will allow us to move some of these abstractions to langchain-core longer term (once we get rid of numpy depdency for the memory vectorstore).

Can i get a second :+1:  for the breaking change?

eyurtsev:
Got :+1: from @hinthornw

PR #19584, opened 2024-03-26
core[patch]: Pull formatting of examples into standalone function

eyurtsev:
1) Users may reasonable want to customize how the example is formatted into text.
2) I don't think we want users to implement behavior via overriding the private
method -- instead we can accept something in the initializer when someone wants
the feature (e.g., function format to text)

So for now pulling this into a standalone function to prevent users from overriding via
inheritance.

### DIFF
--- a/libs/core/langchain_core/example_selectors/semantic_similarity.py
+++ b/libs/core/langchain_core/example_selectors/semantic_similarity.py
@@ -17,8 +17,23 @@ def sorted_values(values: Dict[str, str]) -> List[Any]:
     return [values[val] for val in sorted(values)]
 
 
+def _example_to_text(example: Dict[str, str], input_keys: Optional[List[str]]) -> str:
+    """Convert example to text for vectorstore for indexing purposes."""
+    if input_keys:
+        return " ".join(sorted_values({key: example[key] for key in input_keys}))
+    else:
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/example_selectors/semantic_similarity.py
```diff
@@ -17,8 +17,23 @@ def sorted_values(values: Dict[str, str]) -> List[Any]:
...
+    else:
+        return " ".join(sorted_values(example))
+
+
 class SemanticSimilarityExampleSelector(BaseExampleSelector, BaseModel):
-    """Example selector that selects examples based on SemanticSimilarity."""
+    """Example selector that selects examples based on SemanticSimilarity.
+
+    The example select is able to retrieve examples from a vectorstore based on
+    semantic similarity of the example content to the
```
eyurtsev, line 32: oops forgot to finish doc-string

cbornet:
This could also be done for `_documents_to_examples` passing the documents as arg ?

eyurtsev:
i was a bit less worried about that code path since the indexing hard-codes the metadata being the example, so don't think users are as likely to change `_documents_to_example`

cbornet (approved):

baskaryan (approved):

PR #19589, opened 2024-03-26
docs[minor]: Add chat model tabs to docs pages

bracesproul:


docs/docs/expression_language/why.ipynb
bracesproul, line 1: Had to wrap every instance of 
```mdx
import { ColumnContainer, Column } from "@theme/Columns";
```
or 
```mdx
<ColumnContainer>

<Column>
```
 with

\```{=mdx} <ColumnContainer>

<Column> \```

PR #19591, opened 2024-03-26
Adding try-except block for GCSDirectoryLoader

timothywong731:
- **Description:** Implemented try-except block for `GCSDirectoryLoader`. Reason: Users processing large number of unstructured files in a folder may experience many different errors. A try-exception block is added to capture these errors. A new argument `use_try_except=True` is added to enable *silent failure* so that error caused by processing one file does not break the whole function.
- **Issue:** N/A
- **Dependencies:** no new dependencies
- **Twitter handle:** timothywong731


baskaryan (commented):
cc @lkuligin

baskaryan (approved):

PR #19597, opened 2024-03-26
langchain[minor]: Add async methods to EncoderBackedStore

cbornet:


baskaryan (approved):

libs/langchain/langchain/storage/encoder_backed.py
```diff
@@ -58,38 +59,60 @@ def __init__(
...
+    async def amget(self, keys: Sequence[K]) -> List[Optional[V]]:
+        encoded_keys: List[str] = [self.key_encoder(key) for key in keys]
+        values = await self.store.amget(encoded_keys)
+        return [
+            self.value_deserializer(value) if value is not None else value
+            for value in values
+        ]
+
     def mset(self, key_value_pairs: Sequence[Tuple[K, V]]) -> None:
-        """Set the values for the given keys."""
```
eyurtsev, line 77: Looks good, but could you restore the comments? (Even if they are only marginally helpful. It's not obvious that folks will know what the abbrevation mset means)

Generally, we need to move in the direction of more in code documentation rather than less

cbornet: Since Python 3.5, docstrings are inherited : https://docs.python.org/3/library/inspect.html#inspect.getdoc
In IDEs (at least in PyCharm), the docstring is also inherited.
And it's also inherited in the API reference.
Removing docstrings reduces the amount of lines of code. And makes it easier to improve the doc of all subclasses by improving the parent.

Do you confirm that you prefer to duplicate the doc in subclasses nevertheless ?

eyurtsev:
updated doc-strings will merge

cbornet:
> updated doc-strings will merge

OK. So I deduce that the policy for docstrings is to duplicate in the subclass ?

PR #19602, opened 2024-03-26
cohere[patch]: Add cohere tools agent

harry-cohere:
**Description**: Adds a cohere tools agent and related notebook.


libs/partners/cohere/docs/cohere_agent.ipynb
harry-cohere, line 1: I'm not sure which directory the notebook should be added to. Any advice?

libs/partners/cohere/langchain_cohere/chat_models.py
```diff
@@ -181,16 +189,29 @@ async def _astream(
...
-        return {
+        generation_info = {
             "documents": response.documents,
             "citations": response.citations,
             "search_results": response.search_results,
             "search_queries": response.search_queries,
-            "token_count": response.token_count,
+            "is_search_required": response.is_search_required,
+            "generation_id": response.generation_id,
+            "tool_calls": response.tool_calls,
```
efriis, line 210: let's factor this into the same format as other models
```
tool_calls = [
  {
    "id": "some id string",
    "function": {
      "name": ...,
      "arguments": "json serialized dict or args {}"
    }
  }
]
```

efriis: Without doing this, this will also mess with serialization

libs/partners/cohere/langchain_cohere/cohere_agent.py
```diff
@@ -0,0 +1,120 @@
...
+            "parameter_definitions": tool.args,
+        }
+    else:
+        raise ValueError(
+            f"Unsupported tool type {type(tool)}. Tools must be passed in"
+            f" as BaseTool"
+        )
+
+
+class CohereToolsAgentOutputParser(
```
efriis, line 95: Matching the format above will allow ChatCohere to be used with `OpenAIToolsAgentOutputParser`!

efriis: so we can delete this

libs/partners/cohere/langchain_cohere/chat_models.py
harry-cohere: This is implemented, thank you!

libs/partners/cohere/langchain_cohere/cohere_agent.py
harry-cohere: That sounds good, however, we'd like the flexibility (and naming) in the partner package. We're only 'exporting' `create_cohere_tools_agent`, and if there are plans to refactor some of the tools interfaces in `langchain` we could later update this. Would that work for you?

efriis: Sounds good. Added an integration test to confirm interface matches at least, and it's failing :(

Could you get that passing, then it looks good

libs/partners/cohere/tests/integration_tests/test_chat_models.py
efriis, line 1: @harry-cohere these are the tests that should work on the chat model

libs/partners/cohere/langchain_cohere/cohere_agent.py
harry-cohere: sure thing!

harry-cohere: I think I fixed a couple of bugs, but I also wanted to double check that the behaviour is correct as I changed the tests because they looked wrong to me:

1) In the non-stream test I updated it to use `invoke` instead of `stream`
2) What's the expected kwargs.tool_calls value when tools are bound to the request but the model indicates to not call any tool? The tests previously had this as "tool_calls is not present", but I think it should be "tool_calls is an empty sequence". Happy to change this though!

libs/partners/cohere/tests/integration_tests/test_chat_models.py
```diff
@@ -121,7 +121,7 @@ class Person(BaseModel):
     }
 
     # where it doesn't call the tool
-    strm = tool_llm.stream("What is 2+2?")
+    strm = llm.stream("What is 2+2?")
```
efriis, line 124: This was right before! Does cohere's tool calling always force a tool call, even if the query isn't relevant to the provided tool?

libs/partners/cohere/tests/integration_tests/test_chat_models.py
```diff
@@ -80,7 +80,7 @@ class Person(BaseModel):
     tool_llm = llm.bind_tools([Person])
 
     # where it calls the tool
-    result = tool_llm.stream("Erick, 27 years old")
+    result = tool_llm.invoke("Erick, 27 years old")
```
efriis, line 83: good catch

libs/partners/cohere/tests/integration_tests/test_chat_models.py
harry-cohere: The default behaviour is to indicate that the provided tools are irrelevant. In these cases an empty tool_calls is returned. I can fix the implementation so that it doesn't populate kwargs.tool_calls in this case!

libs/partners/cohere/tests/integration_tests/test_chat_models.py
```diff
@@ -61,3 +68,73 @@ def test_invoke() -> None:
...
+    tool_llm = llm.bind_tools([Person])
+
+    # where it doesn't call the tool
+    strm = tool_llm.stream("What is 2+2?")
+    acc: Any = None
+    for chunk in strm:
+        assert isinstance(chunk, AIMessageChunk)
+        acc = chunk if acc is None else acc + chunk
+    assert acc.content != ""
+    assert "tool_calls" not in acc.additional_kwargs
```
efriis, line 140: @harry-cohere this is the test that is failing currently.

If it's a model-level problem, I'm happy to skip it.

But currently, this call returns an AIMessage with

- content = ""
- tool calls = []

So basically an empty response :(

harry-cohere: DMd you! In short, "don't call any tools because they're all irrelevant" is expressed as empty content and an empty tool_calls array. Happy to change to whatever suits the framework! :)

efriis: Sounds good. I'll mark this test as "skip" for now.

As a heads up, this will make cohere models perform poorly in agents. There's lots of cases, where e.g. you might give an agent a tool like `calculator`, but still want the agent to be able to respond to queries like "what is the capital of france"

PR #19607, opened 2024-03-26
move BaseChatLoader and BaseToolkit to core

lkuligin:

- [ ] **PR title**: "core: move BaseChatLoader and BaseToolkit from community"

- [ ] **PR message**: move BaseChatLoader and BaseToolkit


libs/core/langchain_core/agent_toolkits.py
```diff
@@ -0,0 +1,14 @@
+"""Toolkits for agents."""
+from abc import ABC, abstractmethod
+from typing import List
+
+from langchain_core.pydantic_v1 import BaseModel
+from langchain_core.tools import BaseTool
+
+
+class BaseToolkit(BaseModel, ABC):
```
eyurtsev, line 9: @baskaryan should we group this into the tools namespace instead? nothing about a tool kit is agent specific

baskaryan: yep would put in tools.py

baskaryan (approved):

PR #19616, opened 2024-03-26
docs: `PromptTemplate` import from `core`

leo-gan:
Changed import of `PromptTemplate` from `langchain` to `langchain_core` in all examples (notebooks)


leo-gan:
@baskaryan Thanks!

PR #19486, opened 2024-03-24
langchain/fix-cohere-reranker-rerank-method

jjovalle99:
#### Description
Fixed the following error with `rerank` method from `CohereRerank`:
```
---> [79](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/jjmov99/legal-colombia/~/legal-colombia/.venv/lib/python3.11/site-packages/langchain/retrievers/document_compressors/cohere_rerank.py:79) results = self.client.rerank(
     [80](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/jjmov99/legal-colombia/~/legal-colombia/.venv/lib/python3.11/site-packages/langchain/retrievers/document_compressors/cohere_rerank.py:80)     query, docs, model, top_n=top_n, max_chunks_per_doc=max_chunks_per_doc
     [81](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/jjmov99/legal-colombia/~/legal-colombia/.venv/lib/python3.11/site-packages/langchain/retrievers/document_compressors/cohere_rerank.py:81) )
     [82](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/jjmov99/legal-colombia/~/legal-colombia/.venv/lib/python3.11/site-packages/langchain/retrievers/document_compressors/cohere_rerank.py:82) result_dicts = []
     [83](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/home/jjmov99/legal-colombia/~/legal-colombia/.venv/lib/python3.11/site-packages/langchain/retrievers/document_compressors/cohere_rerank.py:83) for res in results.results:

TypeError: BaseCohere.rerank() takes 1 positional argument but 4 positional arguments (and 2 keyword-only arguments) were given
```
This was easily fixed going from this:
```
   def rerank(
        self,
        documents: Sequence[Union[str, Document, dict]],
        query: str,
        *,
        model: Optional[str] = None,
        top_n: Optional[int] = -1,
        max_chunks_per_doc: Optional[int] = None,
    ) -> List[Dict[str, Any]]:
         ...
        if len(documents) == 0:  # to avoid empty api call
            return []
        docs = [
            doc.page_content if isinstance(doc, Document) else doc for doc in documents
        ]
        model = model or self.model
        top_n = top_n if (top_n is None or top_n > 0) else self.top_n
        results = self.client.rerank(
[...]

PR #19624, opened 2024-03-27
cohere: Fix cohere rerank

billytrend-cohere:
Fix cohere rerank inspired by https://github.com/langchain-ai/langchain/pull/19486

### DIFF
--- a/libs/partners/cohere/langchain_cohere/rerank.py
+++ b/libs/partners/cohere/langchain_cohere/rerank.py
@@ -69,10 +69,14 @@ class CohereRerank(BaseDocumentCompressor):
         model = model or self.model
         top_n = top_n if (top_n is None or top_n > 0) else self.top_n
         results = self.client.rerank(
-            query, docs, model, top_n=top_n, max_chunks_per_doc=max_chunks_per_doc
+            query=query,
+            documents=docs,
+            model=model,
+            top_n=top_n,
+            max_chunks_per_doc=max_chunks_per_doc,
         )
         result_dicts = []
-        for res in results:
+        for res in results.results:
             result_dicts.append(
                 {"index": res.index, "relevance_score": res.relevance_score}
             )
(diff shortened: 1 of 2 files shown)


libs/partners/cohere/langchain_cohere/rerank.py
baskaryan, line 1: do we want to make this backwards compatible with v4?

baskaryan: ah this is to langchain-cohere, ignore me

baskaryan (approved):

libs/partners/cohere/tests/integration_tests/test_rerank.py
```diff
@@ -0,0 +1,16 @@
...
+def test_langchain_cohere_rerank_documents() -> None:
+    """Test cohere rerank."""
+    rerank = CohereRerank()
+    test_documents = [
+        Document(page_content="This is a test document."),
+        Document(page_content="Another test document."),
+    ]
+    test_query = "Test query"
+    results = rerank.rerank(test_documents, test_query)
+    assert len(results) == 2
```
mspronesti, line 16: How about creating a list of 4 (or more) docs and asserting the length of the results to be equal to 3 (default `n`)?

billytrend-cohere: will follow up @mspronesti thanks!

mspronesti: You can click on "commit suggestion" if that's faster :)

```suggestion
    test_documents = [
        Document(page_content="This is a test document."),
        Document(page_content="Another test document."),
        Document(page_content="Yet Another test document."),
        Document(page_content="Fourth test document."),
    ]
    test_query = "Test query"
    results = rerank.rerank(test_documents, test_query)
    assert len(results) == 3
```

Issue #19496, opened 2024-03-25
Upsert method in Milvus class is broken


dmjdarshan:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
# Instantiate Milvus
milvus_chat_history = Milvus(
    embedding_function= embeddings,
    collection_name =collection_name,
    connection_args=connection_args,
    auto_id=True

)

message = "input: " + "Hello Im Gary"
messages = list(message)
res = milvus_chat_history.upsert(documents=messages)
print(res)

```

**Output:
None**

### Description

Im trying to use Milvus Vector Store as a long term memory for LLMs. So I was using Milvus class to experiment. The **upsert** method of Milvus class is broken. When the upsert methond is called, after it passes the "if" cases, the below code is run.

```python
try:
            return self.add_documents(documents=documents)
```

The method is supposed to return with ids of inserted vectors. But it isnt because, the above code snippet is calling the 

```python
self.add_documents(documents=documents)
```

from the base class that is - VectorStore(ABC)

The implementation of the **add_docuents** method in Base class is as follows:
```python
def add_documents(self, documents: List[Document], **kwargs: Any) -> List[str]:
        """Run more documents through the embeddings and add to the vectorstore.

        Args:
            documents (List[Document]: Documents to add to the vectorstore.

        Returns:
            List[str]: List of IDs of the added texts.
        """
        # TODO: Handle the case where the user doesn't provide ids on the Collection
        texts = [doc.page_content for doc in documents]
        metadatas = [doc.metadata for doc in documents]
        return self.add_texts(texts, metadatas, **kwargs)
```

This **add_documents** method from the Base class calls the **add_texts** method of the Base class itself. The add_docuents method of the base class is an **Abstract method**

```python
@abstractmethod
    def add_texts(
        self,
        texts: Iterable[str],
        metadatas: Optional[List[dict]] = None,
        **kwargs: Any,
    ) -> List[str]:
        """Run more texts through the embeddings and add to the vectorstore.

        Args:
            texts: Iterable of strings to add to the vectorstore.
            metadatas: Optional list of metadatas associated with the texts.
            kwargs: vectorstore specific parameters

        Returns:
[...]

PR #19635, opened 2024-03-27
Fix for the broken Upsert method of Milvus class under VectorStores

dmjdarshan:
- [x] **PR title**: "community: Fix to broken upsert method of milvus vectorstore"

- [x] **PR message**: *
    - **Description:** The Upsert method of the Milvus class was calling the add_documents method of the Base class, which called add_texts abstract method in the Base class itself. Instead it was supposed to call the implemented add_texts class in the Milvus class. This has been fixed by overriding the add_documents class in the Milvus claas
    - **Issue:** the issue #19496 discussed has been fixed in this PR
    - **Dependencies:** None
    - **Twitter handle:** @dmjdarshan


baskaryan (commented):
sorry not sure i understand, isn't this identical to the base VectorStore implementation?

dmjdarshan:
The issue is with the flow of control inside Upsert class.
So the Upsert method in the milvus class has a call to add_documents method of the base class.

```python
try:
            return self.add_documents(documents=documents, **kwargs)
        except MilvusException as exc:
            logger.error(
                "Failed to upsert entities: %s error: %s", self.collection_name, exc
            )
            raise exc

```
The add_documents method of the Base class inturn calls add_texts method of the Base class

add_documents method of Base class:
```python
return self.add_texts(texts, metadatas, **kwargs)
```

This call to add_texts in Base class has not implementation. It is an Abstract class - 

```python
@abstractmethod
    def add_texts(
        self,
        texts: Iterable[str],
        **kwargs: Any,
    ) -> List[str]:
        """Run more texts through the embeddings and add to the vectorstore.

        Args:
            texts: Iterable of strings to add to the vectorstore.
            metadatas: Optional list of metadatas associated with the texts.
            kwargs: vectorstore specific parameters

        Returns:
            List of ids from adding the texts into the vectorstore.
        """
```

The implementation of this add_texts method has been done in the Mivus class itself. So the mistake in the flow is - 
Upsert class should have called the add_documents method of the Mivus class itself. But there wasnt overridden method of add_documents in the Milvus class. Hence I added the overriding method for add_documents


Before the Flow was - 
Upsert(Milvus) -> add_documents(Base) -> add_texts(Base : Which is not implemented)

Post the addition of method
Upsert(Milvus) -> add_documents(Milvus) -> add_texts(Base)

dmjdarshan:
@hwchase17 @baskaryan @efriis

zc277584121:
Sorry, there seems to be a problem with the Before the Flow
Upsert(Milvus) -> add_documents(Base) -> add_texts(Milvus: actually invoke subclass's add_texts method)
Is there any problem with the upsert method,  Could you please run this unit test to doublecheck it, https://github.com/langchain-ai/langchain/blob/30f1bf24ac5f62b17dd71bc341fd3dfadd03893f/libs/partners/milvus/tests/integration_tests/vectorstores/test_milvus.py#L187

BTW, new contribution should be placed into the libs/partner/milvus/langchain_milvus package

efriis:
closing - feel free to reopen against the milvus partner package!

PR #19640, opened 2024-03-27
core: Add async methods to LengthBasedExampleSelector

cbornet:


ccurme:
Thanks for this. From what I can tell `LengthBasedExampleSelector` already supports `aadd_example` and `aselect_examples` via its inheritance of `BaseExampleSelector`:
https://github.com/langchain-ai/langchain/blob/8ab7bb316652cc396ebf43f0307cff9b025c4756/libs/core/langchain_core/example_selectors/base.py#L15-L17
https://github.com/langchain-ai/langchain/blob/8ab7bb316652cc396ebf43f0307cff9b025c4756/libs/core/langchain_core/example_selectors/base.py#L23-L25

Closing this for now but let me know if I've missed something or you are running into issues using these methods.

cbornet:
@ccurme No the default implementation is not optimal for `LengthBasedExampleSelector` as it uses a thread-pool.
This PR makes it use the calling thread which is more efficient.
Can you reopen ?

eyurtsev (approved):

PR #19644, opened 2024-03-27
Community: Updating Azure Retriever and Docs to be Azure AI Search instead of Azure Cognitive Search

marlenezw:
Last year Microsoft [changed the name](https://learn.microsoft.com/en-us/azure/search/search-what-is-azure-search) of Azure Cognitive Search to Azure AI Search. This PR updates the Langchain Azure Retriever API and it's associated docs to reflect this change. It may be confusing for users to see the name Cognitive here and AI in the Microsoft documentation which is why this is needed. I've also added a more detailed example to the Azure retriever doc page. 

There are more places that need a similar update but I'm breaking it up so the PRs are not too big 😄 

Twitter: @marlene_zw

No new tests added.


docs/docs/integrations/retrievers/azure_ai_search.ipynb
```diff
@@ -0,0 +1,292 @@
...
+   "source": [
+    "## Set up Azure AI Search\n",
+    "\n",
+    "To set up AAS, please follow the instructions [here](https://learn.microsoft.com/en-us/azure/search/search-create-service-portal).\n",
+    "\n",
+    "Please note you will need:\n",
+    "\n",
+    "1. the name of your AAS service, \n",
+    "2. the name of your AAS index,\n",
+    "3. your API key.\n",
```
mattgotteiner, line 35: Were you able to add support for role based authentication?

https://learn.microsoft.com/en-us/azure/search/search-security-rbac?tabs=config-svc-portal%2Croles-portal%2Ctest-portal%2Ccustom-role-portal%2Cdisable-keys-portal

docs/docs/integrations/retrievers/azure_ai_search.ipynb
```diff
@@ -0,0 +1,292 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "057deaad",
+   "metadata": {},
+   "source": [
+    "Create the Retriever\n",
+    "\n",
+    "`content_key` is the key in the retrieved result to set as the Document page_content.\n",
+    "`top_k` is the number of number of results you'd like to retrieve. Setting it to None (the default) returns all results. "
```
mattgotteiner, line 90: There are 2 different parameters right now. top and k for vector queries. Does top_k set one or both of these?

ccurme:
Thanks for this!

This will break things for users of `AzureCognitiveSearchRetriever`. Could we keep those modules around, but just have them import `AzureAISearchRetriever`? Then we can deprecate the old one gracefully if needed.
```python
from langchain_community.retriever.azure_ai_search import AzureAISearchRetriever

# For backwards compatibility
AzureCognitiveSearchRetriever = AzureAISearchRetriever
```
This assumes that there's no functional differences between the two.

Feel free to remove the draft status of the PR if it is good to go from your perspective.

docs/docs/integrations/retrievers/azure_ai_search.ipynb
marlenezw: I haven't yet! I'll open a separate PR for this!

marlenezw:
> Thanks for this!
> 
> This will break things for users of `AzureCognitiveSearchRetriever`. Could we keep those modules around, but just have them import `AzureAISearchRetriever`? Then we can deprecate the old one gracefully if needed.
> 
> ```python
> from langchain_community.retriever.azure_ai_search import AzureAISearchRetriever
> 
> # For backwards compatibility
> AzureCognitiveSearchRetriever = AzureAISearchRetriever
> ```
> 
> This assumes that there's no functional differences between the two.
> 
> Feel free to remove the draft status of the PR if it is good to go from your perspective.

Makes sense to me @ccurme, I'll make changes and remove draft status once ready to review!

docs/docs/integrations/retrievers/azure_ai_search.ipynb
marlenezw: The current code is setting only the top parameter. Do you think it's worth it to set both?

`top_param = f"&$top={self.top_k}" if self.top_k else ""`

marlenezw:
Ooof, sorry, I just pulled from upstream and thought I rebased correctly. Going to close this and start a new PR 😓

PR #19648, opened 2024-03-27
core: Add aformat_messages to FewShotChatMessagePromptTemplate and ChatPromptTemplate

cbornet:
Needed since the example selector may use a vector store.

### DIFF
--- a/libs/core/langchain_core/prompts/chat.py
+++ b/libs/core/langchain_core/prompts/chat.py
@@ -38,6 +38,7 @@ from langchain_core.prompts.image import ImagePromptTemplate
 from langchain_core.prompts.prompt import PromptTemplate
 from langchain_core.prompts.string import StringPromptTemplate, get_template_variables
 from langchain_core.pydantic_v1 import Field, root_validator
+from langchain_core.runnables import run_in_executor
 from langchain_core.utils import get_colored_text
 from langchain_core.utils.interactive_env import is_interactive_env
 
@@ -65,6 +66,17 @@ class BaseMessagePromptTemplate(Serializable, ABC):
         Returns:
             List of BaseMessages.
         """
+
+    async def aformat_messages(self, **kwargs: Any) -> List[BaseMessage]:
+        """Format messages from kwargs. Should return a list of BaseMessages.
+
+        Args:
+            **kwargs: Keyword arguments to use for formatting.
+
+        Returns:
+            List of BaseMessages.
+        """
+        return await run_in_executor(None, self.format_messages, **kwargs)
 
     @property
     @abstractmethod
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/prompts/chat.py
```diff
@@ -901,19 +929,31 @@ def from_messages(
             partial_variables=partial_vars,
         )
 
-    def format(self, **kwargs: Any) -> str:
```
cbornet, line 904: Removed `format` from `ChatPromptTemplate` as it is identical to the one in the parent class `BaseChatPromptTemplate `

libs/core/langchain_core/prompts/chat.py
```diff
@@ -66,6 +67,17 @@ def format_messages(self, **kwargs: Any) -> List[BaseMessage]:
...
+    async def aformat_messages(self, **kwargs: Any) -> List[BaseMessage]:
+        """Format messages from kwargs. Should return a list of BaseMessages.
+
+        Args:
+            **kwargs: Keyword arguments to use for formatting.
+
+        Returns:
+            List of BaseMessages.
+        """
+        return await run_in_executor(None, self.format_messages, **kwargs)
```
eyurtsev, line 79: This is another place where I feel like we need to delegate to the sync implementation.

I'm a bit worried about having this dichotomy / how to make it obvious that this is is what's going on

eyurtsev: cc @baskaryan  / @nfcampos 

Do we keep run_in_executor for consistency or use a blocking call to the sync code?

eyurtsev: Another PR where this question comes up: https://github.com/langchain-ai/langchain/pull/19652/files

eyurtsev: This is the part that makes it all weird:

https://github.com/langchain-ai/langchain/blob/2e0ddd6fb805af53d9bfb0edc463543a0312530f/libs/core/langchain_core/prompts/chat.py#L934-L934

Chat prompt templates are constructed of multiple messages each of which can be formatted independently -- this 
would mean that interpolating a few variables into essentially a list of dicts we're launching several threads

cbornet: Another possibility if we don't want to choose now is to raise NotImplementedError in the default interface.

cbornet: I intend to implement all subclasses in the end so the default interface won't be used anyway.

cbornet: > a list of dicts we're launching several threads

Actually, it's not launching several threads, it's submitting tasks to the default thread-pool (which has 1 thread per core).

cbornet: Anyway, I updated the PR to call the sync method directly. PTAL

eyurtsev (approved):

PR #19649, opened 2024-03-27
together[minor]: Update endpoint to non deprecated version

dev-yashmathur:
- **Updating Together.ai Endpoint**: "langchain_together: Updated Deprecated endpoint for partner package"

- Description: The inference API of together is deprecates, do replaced with completions and made corresponding changes.
- Twitter handle: @dev_yashmathur


libs/partners/together/langchain_together/llms.py
```diff
@@ -34,13 +34,14 @@ class Together(LLM):
...
-    """Base inference API URL."""
+    base_url: str = "https://api.together.xyz/v1/completions"
+    """Base completions API URL."""
     together_api_key: SecretStr
     """Together AI API key. Get it here: https://api.together.xyz/settings/api-keys"""
     model: str
     """Model name. Available models listed here: 
-        https://docs.together.ai/docs/inference-models
+        Base Models: https://docs.together.ai/docs/inference-models#language-models
+        Chat Models: https://docs.together.ai/docs/inference-models#chat-models
```
sepiatone, line 44: @dev-yashmathur `max_tokens` is now a required parameter (https://docs.together.ai/reference/completions). 

also as a matter of style, I don't think these changes are minor :)

dev-yashmathur: Hey, appreciate the feedback. I seem to have missed that, as even without it; i think a default of 20 is being set.
I have added a new commit which should take this into consideration.
Looking forward to any other feedback. Thanks :)

sepiatone: Wouldn't it be simpler to do this:
`max_tokens: Optional[int] = None` to
`max_tokens: int = 128`

dev-yashmathur: Hey @sepiatone . While that is the simpler solution, I have opted for this approach so as to inform the users of the requirement. Just setting it ourselves, would not include the warning. If you think that the warning can be skipped, then I'll make this change as well.

sepiatone: @dev-yashmathur I checked the implementation of a few other models and we don't give a warning, that said, I'm ok with it

dev-yashmathur: @sepiatone  So, can this PR be merged, or would you like me to make the modification? 

P.S. I appreciate you taking the time to help me out with this :)

sepiatone: @dev-yashmathur lgtm! now the actual maintainers of the langchain-together package, @baskaryan & @efriis will have to be satisfied :)

baskaryan (approved):

PR #19652, opened 2024-03-27
core[minor]: Add aformat to FewShotPromptTemplate

cbornet:


libs/core/langchain_core/prompts/base.py
```diff
@@ -182,6 +182,23 @@ def format(self, **kwargs: Any) -> FormatOutputType:
             prompt.format(variable1="foo")
         """
 
+    async def aformat(self, **kwargs: Any) -> FormatOutputType:
```
eyurtsev, line 185: Will need to think for a bit about whether we want to include aformat. Feels fairly unusual that formatting will involve network calls

cbornet: It does for the few-shot prompt template.

eyurtsev: The default should be delegating to sync rather than kicking off a thread.

I think we also need to provide `ainvoke` implementation here otherwise this won't be used at run time when using LCEL

eyurtsev:
okay @cbornet confirmed with @nfcampos offline, let's make this sync since IO is not typically required here and we need to implement ainvoke to use it -- let me know if this is okay

cbornet:
OK, I'll update the PR.

libs/core/langchain_core/prompts/base.py
cbornet: We can also raise NotImplementedError for now and decide later/never which should be the default impl. I intend to implement it in all subclasses anyway. Not making it abstract as that would be a breaking change.

> I think we also need to provide ainvoke implementation here otherwise this won't be used at run time when using LCEL

I wanted to do it in another PR to keep changes atomic.

cbornet: > We can also raise NotImplementedError for now and decide later/never which should be the default impl. 

Actually this could lead to some breaking changes when we implement other async methods to not call their sync counterpart in a thread (eg. ainvoke).

cbornet:
Done.
@eyurtsev can you take a look ?

eyurtsev (approved):

cbornet:
@eyurtsev I added implementations for ImagePromptTemplate and FewShotChatMessagePromptTemplate as they are straight-forward.

eyurtsev (approved):

eyurtsev:
@cbornet i'm merging this, but I still think that this code won't actually get triggered since I didn't see _ainvoke implemented since the last review

PR #19656, opened 2024-03-27
fix: relax top-level title/description requirement for converting JSON schema to OpenAI function

dqbd:


libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -71,6 +71,21 @@ def json_schema() -> Dict:
     }
 
 
+@pytest.fixture()
+def json_schema_lax() -> Dict:
```
eyurtsev, line 75: nit: a fixture probably doesn't make sense here too much abstraction for a unit test -- i'd just embed the test case into the function itself

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -295,14 +295,12 @@ def convert_to_openai_function(
...
-    elif isinstance(function, dict) and all(
-        k in function for k in ("title", "description", "properties")
-    ):
+    # a JSON schema
+    elif isinstance(function, dict) and "properties" in function:
         function = function.copy()
         return {
-            "name": function.pop("title"),
-            "description": function.pop("description"),
+            "name": function.pop("title", "extract"),
```
eyurtsev, line 302: This I am a bit opposed since this isn't default API behavior? I don't think we should be auto assigning a title

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -295,14 +295,12 @@ def convert_to_openai_function(
...
+            "description": function.pop("description", ""),
```
eyurtsev, line 303: This would be OK by me if we were to omit the key, but if we're subbing in a default value of `""` I don't know that consistent, now it actually feels like we're interpreting the API to make something that used to be a requirement of the API to making it optional in our API?

PR #19676, opened 2024-03-27
community: document_loaders: GrobidParser: fix IndexError: list index out of range

### LINKED ISSUE #10801: GrobidParser: process_xml chunk_bboxes[0][0]["page"], IndexError: list index out of range
Using the GrobidParser with `segment_sentences=False` results in `langchain/document_loaders/parsers/grobid.py", line 87, in process_xml
    chunk_bboxes[0][0]["page"],
IndexError: list index out of range` for some PDFs.

From my own testing with the offending PDFs I suspect the following if-block in the GrobidParser is indented incorrectly, it needs one more level of indentation. Adding this indentation solved the issue on my end.
https://github.com/langchain-ai/langchain/blob/20c6ade2fc24c458a080eaa468add7c61dc8dbc3/langchain/document_loaders/parsers/grobid.py#L80-L93

If you need a reproduction to verify, let me know.

zaurask:
Fixes #10801

After two comments in #10801 confirming my suggested fix, this is the PR for it.


libs/community/langchain_community/document_loaders/parsers/grobid.py
```diff
@@ -83,20 +83,20 @@ def process_xml(
...
-                            "para": str(i),
-                            "bboxes": chunk_bboxes,
-                            "section_title": sect.text,
-                            "section_number": sect.get("n"),
-                            "pages": (fpage, lpage),
-                        }
-                        chunks.append(paragraph_dict)
+                        if segment_sentences is not True:
+                            fpage, lpage = (
+                                chunk_bboxes[0][0]["page"],
```
baskaryan, line 88: seems like you could still get an index error since it's not guaranteed something has been added to chunk_bboxes yet. is the correct fix to add an explicit check that chunk_bboxes isn't empty?

baskaryan (commented):
cc @corranmac

libs/community/langchain_community/document_loaders/parsers/grobid.py
zaurask: The [Grobid docs](https://grobid.readthedocs.io/en/latest/Coordinates-in-PDF/) state, that

> for each entity the coordinates of a structure is provided as a list of bounding boxes, each one separated by a semicolon ```;```, each bounding box being defined by 5 attributes separated by a comma ```,```

So when iteraring a sentence structure, there is an accompanying bounding box, here ```bbox```. ```sbboxes``` should be created and appended to ```chunk_bboxes```.  If you wish to make it extra robust we could check the length of the ```chunk_bbox```.

libs/community/langchain_community/document_loaders/parsers/grobid.py
```diff
@@ -83,20 +83,20 @@ def process_xml(
...
-                        paragraph_dict = {
-                            "text": "".join(paragraph_text),
-                            "para": str(i),
-                            "bboxes": chunk_bboxes,
-                            "section_title": sect.text,
-                            "section_number": sect.get("n"),
-                            "pages": (fpage, lpage),
-                        }
-                        chunks.append(paragraph_dict)
+                        if segment_sentences is not True:
```
clement-viguier-ceataec, line 86: This is what I did when facing this bug. I believe this is enough but I haven't thought too much about it.
I kept the original indentation.
```suggestion
                    if (segment_sentences is not True) and (len(chunk_bboxes) > 0):
```

zaurask: You are right, keeping the original indentation is the right solution. The paragraph_dict needs the full paragraph_text that is build up in the for loop from all individual sentences. Why the chunk_bboxes are empty for some pdfs is another question that I will investigate further given proper internet connection to setup Grobid again.

hwchase17:
seems like it needs some more work, closing until fixed

zhaoy2020:
###############################################
# debug by myself, sometimes is empty.
###############################################
try:
    sbboxes.append(
        {
            "page": box[0],
            "x": box[1],
            "y": box[2],
            "h": box[3],
            "w": box[4],
        }
    )
except:
    print("Error in parsing bbox, maybe empty!") 
    continue

zhaoy2020:
```python
###############################################
# debug by myself, sometimes is empty.
###############################################
try:
    sbboxes.append(
        {
            "page": box[0],
            "x": box[1],
            "y": box[2],
            "h": box[3],
            "w": box[4],
        }
    )
except:
    print("Error in parsing bbox, maybe empty!") 
    continue 
```

PR #19680, opened 2024-03-27
community: Add Datahareld tool

jcjc712:
**Description:** Integrate [dataherald](https://www.dataherald.com) tool, It is a natural language-to-SQL tool.
**Dependencies:** Install dataherald sdk to use it,
```
pip install dataherald
```


docs/docs/integrations/tools/dataherald.ipynb
baskaryan, line 1: would be nice to show how to use the tool (not the api wrapper) in a chain or agent

libs/community/langchain_community/tools/dataherald/tool.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import Optional, Type
+
+from langchain_core.callbacks import CallbackManagerForToolRun
+from langchain_core.pydantic_v1 import BaseModel, Field
+from langchain_core.tools import BaseTool
+
+from langchain_community.utilities.dataherald import DataheraldAPIWrapper
+
+
+class DataheraldTextToSQLInput(BaseModel):
```
baskaryan, line 12: add an input schema with a description for the prompt field, so that if you use a tool-calling llm this description gets added as part of the tool schema. feel free to change the wording!

baskaryan (commented):
overall lgtm! 2 comments

docs/docs/integrations/tools/dataherald.ipynb
jcjc712: Sure, sorry for my late answer, I'll work on this

libs/community/langchain_community/tools/dataherald/tool.py
jcjc712: @baskaryan hi, I have a question about how to send another parameter, for example I need always the `prompt` and a `db_connection_id` (which was generated previously in our platform).

1. One option is adding it in the wrapper like this:
```
api_wrapper = DataheraldAPIWrapper(dataherald_api_key="", db_connection_id="")
```
besides `db_connection_id` could be a env var (DATAHERALD_DB_CONNECTION_ID) but if I want to change the db connection in the execution process could be difficult.

2. The other option is send the db connection when I invoke but not sure how to do it
```
agent_executor.invoke({"input":{"prompt":"foo", "db_connection_id": "bar"})
```
It doesn't take the `db_connection_id` param and raise an error.

jcjc712: Hi, I followed this approach so the db_connection_id should be set when the Wrapper is created
`
api_wrapper = DataheraldAPIWrapper(dataherald_api_key="", db_connection_id="")
`

docs/docs/integrations/tools/dataherald.ipynb
jcjc712: I added the steps to use the tool in an agent

libs/community/langchain_community/tools/dataherald/tool.py
baskaryan: yea think setting at initialization makes sense 👍

baskaryan (approved):

libs/community/langchain_community/tools/dataherald/tool.py
jcjc712: Thanks, feel free to merge it if it pass all the validations

docs/docs/integrations/tools/dataherald.ipynb
```diff
@@ -0,0 +1,125 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "dataherald.run(\"How many employees are in the company?\")"
+   ]
+  },
+  {
+   "cell_type": "code",
```
cbornet, line 91: This code cell is empty.

jcjc712: Thanks, I removed it

PR #19684, opened 2024-03-28
core: Two updates to chat model interface

nfcampos:
- .stream() and .astream() call on_llm_new_token, removing the need for subclasses to do so. Backwards compatible because now we don't pass run_manager into ._stream and ._astream
- .generate() and .agenerate() now handle `stream: bool` kwarg for _generate and _agenerate. Subclasses handle this arg by delegating to ._stream(), now one less thing they need to do. Backwards compat because this is an optional arg that we now never pass to the subclasses
-  .generate() and .agenerate() now inspect callback handlers to decide on a default value for stream:bool if not passed in. This auto enables streaming when using astream_events and astream_log
- as a result of these three changes any usage of .astream_events and .astream_log should now yield chat model stream events
- In future PRs we can update all subclasses to reflect these two things now handled by base class, but in meantime all will continue to work

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -50,6 +50,7 @@ from langchain_core.outputs import (
 from langchain_core.prompt_values import ChatPromptValue, PromptValue, StringPromptValue
 from langchain_core.pydantic_v1 import Field, root_validator
 from langchain_core.runnables.config import ensure_config, run_in_executor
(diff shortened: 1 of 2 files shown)


efriis (commented):
- are we concerned any chat models might call anything custom in the run_manager in _stream
- would love to run integration tests with this editably installed in all the partner packages - might be a good thing to add to core release CI

this is amazing though. Huge cleanup 🧹

efriis (approved):
partner package tests looking good

PR #19708, opened 2024-03-28
cohere: Add types to clients and drop v4 support from partner package

billytrend-cohere:
cohere: Add types to clients and drop v4 support from partner package


libs/partners/cohere/langchain_cohere/embeddings.py
```diff
@@ -51,6 +51,18 @@ class Config:
         arbitrary_types_allowed = True
         extra = Extra.forbid
 
+    def get_client(self) -> cohere.Client:
```
billytrend-cohere, line 54: wondering if we should switch to Field(default_factory=...)

libs/partners/cohere/langchain_cohere/embeddings.py
```diff
@@ -26,9 +26,9 @@ class CohereEmbeddings(BaseModel, Embeddings):
             )
     """
 
-    client: Any  #: :meta private:
+    client: Optional[cohere.Client]  #: :meta private:
```
efriis, line 29:
```suggestion
    client: cohere.Client = Field(default=None)  #: :meta private:
```

efriis: This is how we typically do it. Is there a case where you actually don't want `self.client` to be defined?

efriis: Looks like in `validate_environment` it is always initialized anyways, so probably better to just type it as such!

libs/partners/cohere/langchain_cohere/llms.py
```diff
@@ -42,16 +42,18 @@ def acompletion_with_retry(llm: Cohere, **kwargs: Any) -> Any:
...
 
     return _completion_with_retry(**kwargs)
 
 
 class BaseCohere(Serializable):
     """Base class for Cohere models."""
 
-    client: Any = None  #: :meta private:
-    async_client: Any = None  #: :meta private:
+    client: Optional[cohere.Client]  #: :meta private:
```
efriis, line 53: see note on embed

libs/partners/cohere/langchain_cohere/rerank.py
```diff
@@ -13,8 +13,8 @@
 class CohereRerank(BaseDocumentCompressor):
     """Document compressor that uses `Cohere Rerank API`."""
 
-    client: Any = None
-    """Cohere client to use for compressing documents."""
+    client: Optional[cohere.Client]  #: :meta private:
```
efriis, line 16: see note on embed

libs/partners/cohere/langchain_cohere/embeddings.py
billytrend-cohere: that makes sense, will do that pattern, thanks!

efriis:
Do in partner repo!

https://github.com/langchain-ai/langchain-cohere

PR #19711, opened 2024-03-28
community: Add Upstage document loader using Layout-Analyzer Model

junkeon0412:
**Description:**

This PR introduces a new document loader tailored for the Upstage API, leveraging the Layout-Analyzer Model. It encompasses essential functionalities for document retrieval, response parsing, and management of various split types.

**Tests:**

This PR includes a test suite for the Upstage document loader, utilizing mock requests for each scenario. To execute the tests, navigate to `libs/community/tests/unit_tests/document_loaders/test_upstage.py.`

**Docs:**

Furthermore, this PR incorporates documentation for the Upstage document loader. Please refer to `docs/docs/integrations/document_loaders/upstage.ipynb` for guidance on utilizing the **UpstageDocumentLoader**.

PR #19715, opened 2024-03-28
community: Add Upstage document loader using Layout-Analyzer Model

junkeon0412:
**Description:**

This PR introduces a new document loader tailored for the Upstage API, leveraging the Layout-Analyzer Model. It encompasses essential functionalities for document retrieval, response parsing, and management of various split types.

**Tests:**

This PR includes a test suite for the Upstage document loader, utilizing mock requests for each scenario. To execute the tests, navigate to `libs/community/tests/unit_tests/document_loaders/test_upstage.py.`

**Docs:**

Furthermore, this PR incorporates documentation for the Upstage document loader. Please refer to `docs/docs/integrations/document_loaders/upstage.ipynb` for guidance on utilizing the **UpstageDocumentLoader**.


junkeon:
https://github.com/langchain-ai/langchain/pull/19711#issuecomment-2026017292

@eyurtsev 

> thanks for the contribution -- would you mind exposing this as a Parser rather than a Document Loader. If you want you can make a DocumentLoader as well that re-uses the code from the Parser.
> 
> The parser abstraction is much more generic and we'll likely heavily lean into in the future (e.g., content can be loaded from memory without having to reside on the file system).

Thank you for your comment. The Upstage Document Loader can handle PDFs and various image formats, and we plan to support more types like Microsoft Office documents in the near future. Therefore, it might be challenging to locate the Upstage Document Loader in the right place under the parser package.

libs/community/langchain_community/document_loaders/upstage.py
```diff
@@ -0,0 +1,227 @@
...
+    Raises:
+        ValueError: If the API call returns a non-200 status code.
+        ValueError: If an invalid split type is provided.
+
+    """
+
+    def __init__(
+        self,
+        file_path: str,
+        output_type: str = OutputType.TEXT.value,
```
eyurtsev, line 118: I suggest using Literals to annotate the API

libs/community/langchain_community/document_loaders/upstage.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Raises:
+        ValueError: If the API call returns a non-200 status code.
+        ValueError: If an invalid split type is provided.
+
+    """
+
+    def __init__(
+        self,
+        file_path: str,
```
eyurtsev, line 117: consider supporting Union[Pathlib, str]

libs/community/langchain_community/document_loaders/upstage.py
```diff
@@ -0,0 +1,227 @@
...
+        self.output_type = output_type
+        self.split = split
+        self.api_key = api_key
+        self.file_name = os.path.basename(file_path)
+        self.url = url
+
+        validate_file_path(self.file_path)
+        validate_api_key(self.api_key)
+
+    def load(self) -> List[Document]:
```
eyurtsev, line 134: Define lazy_load rather than load

libs/community/langchain_community/document_loaders/upstage.py
```diff
@@ -0,0 +1,227 @@
...
+        return data["html"]
+    else:
+        raise ValueError(f"Invalid output type: {output_type}")
+
+
+class UpstageDocumentLoader(BaseLoader):
+    """
+    A document loader for Upstage API.
+
+    Args:
```
eyurtsev, line 94: We commonly put the doc-string for the init arguments in the __init__ section.

Our API reference docs will concatenate both sections if I'm not mistaken

eyurtsev (commented):
For the parser change follow as an example: 

https://github.com/langchain-ai/langchain/blob/main/libs/community/langchain_community/document_loaders/parsers/docai.py#L1-L1

https://github.com/langchain-ai/langchain/blob/main/libs/community/langchain_community/document_loaders/parsers/docai.py#L1-L1

You can keep a document loader, and make the document loader re-use the code in the parser.

eyurtsev (changes requested):
Left a few comments on changes that need to be made. The main one is implementation of the main parsing logic in the blob parser abstraction.

libs/community/langchain_community/document_loaders/upstage.py
junkeon: I'm sorry, but I'm having trouble understanding your comment. Could you please provide more information so that I may better understand and fix the code?

libs/community/langchain_community/document_loaders/upstage.py
junkeon: I apologize, but I'm afraid I'm also unclear about the meaning of your comment. Could you please provide more detail? Specifically, are you asking if I simply need to rename the `load` function to `lazy_load` or if there are additional steps required to define the `lazy_load` function?

junkeon:
@eyurtsev 
I appreciate your kind and detailed comments. I have pushed the changes to address them. However, there are a few comments that I'm still not sure about. If you could provide more information, I would greatly appreciate it. Additionally, I have added `upstage.py` to the parser directory as you suggested, but I'm not entirely sure if this is the correct placement. Could you please double-check and give me some feedback? Once I have received your feedback, I will do my best to fix any issues quickly.

libs/community/langchain_community/document_loaders/upstage.py
junkeon: I fix the code with your comment

libs/community/langchain_community/document_loaders/upstage.py
junkeon: I fix the code with your comment

libs/community/langchain_community/document_loaders/upstage.py
junkeon: I made the change from Enum to Literals based on your suggestion. Thank you.

libs/community/langchain_community/document_loaders/upstage.py
junkeon: Just as you suggested, I added the lazy\_load function.

minchang (changes requested):

junkeon:
After a thorough internal review, I'll resend the PR for your consideration, and I truly appreciate your patience and input throughout this process.

Issue #19441, opened 2024-03-22
Annotation for langchain_core.output_parsers.json.JsonOutputParser -> pydantic_object not compatible for v2


GentleGhostCoder:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

see Description

### Error Message and Stack Trace (if applicable)

 Input should be a subclass of BaseModel

### Description

The attribute `pydantic_object` in class `langchain_core.output_parsers.json.JsonOutputParser` still dosent support BaseModel -> V2 (just V1)
The line 197 should support ` Optional[Type[Union[BaseModelV1, BaseModelV2]]` instead just BaseModel from v1.

### System Info

dosen´t matter

PR #19716, opened 2024-03-28
core: support pydantic V2 for JSONOutputParser, allow for other sources of JSON schemas

jnis23:
This PR supports using Pydantic v2 objects to generate the schema for the JSONOutputParser (#19441). This also adds a `json_schema` parameter to allow users to pass any JSON schema to validate with, not just pydantic.

### DIFF
--- a/libs/core/langchain_core/output_parsers/json.py
+++ b/libs/core/langchain_core/output_parsers/json.py
@@ -2,16 +2,30 @@ from __future__ import annotations
 
 import json
 import re
+import warnings
 from json import JSONDecodeError
-from typing import Any, Callable, List, Optional, Type
+from typing import Any, Callable, List, Optional, Type, TypeVar, Union
 
 import jsonpatch  # type: ignore[import]
+import pydantic  # pydantic: ignore
 
 from langchain_core.exceptions import OutputParserException
 from langchain_core.output_parsers.format_instructions import JSON_FORMAT_INSTRUCTIONS
 from langchain_core.output_parsers.transform import BaseCumulativeTransformOutputParser
 from langchain_core.outputs import Generation
-from langchain_core.pydantic_v1 import BaseModel
+from langchain_core.pydantic_v1 import validator
+from langchain_core.utils.pydantic import PYDANTIC_MAJOR_VERSION
+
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/output_parsers/json.py
```diff
@@ -194,11 +207,29 @@ class JsonOutputParser(BaseCumulativeTransformOutputParser[Any]):
     describing the difference between the previous and the current object.
     """
 
-    pydantic_object: Optional[Type[BaseModel]] = None
+    pydantic_object: Optional[Type[TBaseModel]] = None  # type: ignore
+    json_schema: Optional[Dict[str, Any]] = None
```
eyurtsev, line 211: Could we remove the new attribute json_schema? As an API it's not obvious what it's doing.

@jnis23 I'm also not sure how many of these changes we should be making. We'll be dropping support for pydantic 1 + pydantic 2 v1 in about 3 months, and most of the conditional behavior is going to make the refactor in 3 months more complicated to execute.

jnis23: The thought behind `json_schema` was that if users had their JSON schema already defined, they could just `json.load` the schema definition file and pass the dict to the parser instead of having to recreate the schema as a pydantic object. Agreed there is probably a better name and would also need some doc. I'll remove for now and it can be added as a new feature if someone has that use case.

Sounds good on no more changes, the full upgrade to v2 will clean this up a lot. I'll remove `json_schema` and push this back up. Feel free to close this if we want to hold off until the full v2 upgrade

eyurtsev (approved):

libs/core/langchain_core/output_parsers/json.py
eyurtsev: Because json_schema and pydantic_object are used for the same purpose (defining a schema), it makes more sense to rename `pydantic_object` to schema and have it accept a union of pydantic or JSON schema.

In this case I don't think it's worth making a change -- it's just changing the format instructions associated with this parser which is used for llms / chat models that don't support tool calling. In that case it's unclear that JSONSchema is actually a good schema to use (it's too verbose). (https://github.com/eyurtsev/kor likely works better with typescript schema)

PR #19720, opened 2024-03-28
community: Fix YandexGPT embeddings

egorkrash:
Fix of YandexGPT embeddings. 

The current version uses a single `model_name` for queries and documents, essentially making the `embed_documents` and `embed_query` methods the same. Yandex has a different endpoint (`model_uri`) for encoding documents, see [this](https://yandex.cloud/en/docs/yandexgpt/concepts/embeddings). The bug may impact retrievers built with `YandexGPTEmbeddings` (for instance FAISS database as retriever) since they use both `embed_documents` and `embed_query`.

A simple snippet to test the behaviour:
```python
from langchain_community.embeddings.yandex import YandexGPTEmbeddings
embeddings = YandexGPTEmbeddings()
q_emb = embeddings.embed_query('hello world')
doc_emb = embeddings.embed_documents(['hello world', 'hello world'])
q_emb == doc_emb[0]
```
The response is `True` with the current version and `False` with the changes I made.

Twitter: @egor_krash


tyumentsev4:
Thank you!

egorkrash:
@baskaryan can you take a look, please?

libs/community/langchain_community/embeddings/yandex.py
```diff
@@ -48,12 +47,12 @@ class YandexGPTEmbeddings(BaseModel, Embeddings):
...
-    model_uri: str = ""
-    """Model uri to use."""
     folder_id: str = ""
     """Yandex Cloud folder ID"""
-    model_name: str = "text-search-query"
-    """Model name to use."""
+    doc_model_name: str = "text-search-doc"
+    """Doc model name to use."""
+    query_model_name: str = "text-search-query"
+    """Query model name to use."""
```
baskaryan, line 55: this is a breaking change, we should find a way to make it backwards compatible

baskaryan: updated, let me know what you think @egorkrash

egorkrash: Oh yes, you are right @baskaryan 
This looks good, thanks for adding the test too!

egorkrash: image

Maybe I'm looking in the wrong place, but this test seems to be skipped

egorkrash:
> As a workaround we can use a dedicated vector store for indexing docs:
> 
> ```
> store_for_indexing = Foo(foo_client,
>                       collection_name=foo_collection,
>                       embeddings=YandexGPTEmbeddings(sleep_interval=0.1,
>                                                     model_name="text-search-doc") 
>                       )
> ```
> 
> this also includes workaround for #19773

I don't quite understand what you are proposing.

mkhludnev:
> I don't quite understand what you are proposing.

ok. let me elaborate:
This [PR](#19720) fixes YandexGPTEmbeddings, so that a VectorStore (qdrant in my case) calls `text-search-doc` model on indexing. This this a right thing to do. But until it's released users my work around this bug with the current langchain:
 - we can introduce two VectorStore instances point to the same remote storage (although, this might not work with in-memory stores)
 - a VectoreStore which is used for indexing might get `YandexGPTEmbeddings` with the correct `text-search-doc` model (as shown above). Thus, docs will have a proper embeddings. And the second VectorStore with YandexGPTEmbeddings with default model is able to search via `text-search-query` embeddings.  
My snippet doesn't touch not impact your PR, it's just a suggestion/advice for users which need proper embeddings right now. Hope it makes it clear. 
Thanks for your contribution, btw!

libs/community/langchain_community/embeddings/yandex.py
baskaryan: which job are you looking? that's probably the extended_test job, this runs in the regular make_test job (which doesn't require any optional dependencies) https://github.com/langchain-ai/langchain/actions/runs/8512241122/job/23313469252?pr=19720

baskaryan (approved):

PR #19728, opened 2024-03-28
community: Fix bug in vdms

cwlacewe:
**Description:** Fix embedding check in vdms
**Contribution maintainer:** [@cwlacewe](https://github.com/cwlacewe)


ccurme:
Could you comment on why an empty bytes is preferable to None? Thanks.

cwlacewe:
> Could you comment on why an empty bytes is preferable to None? Thanks.

Internally VDMS expects an empty array if no blob/bytes data is provided.  In the case of the VDMS integration, no data is considered None so we provide checks and only update array for VDMS query if data is provided.

baskaryan (approved):

PR #19730, opened 2024-03-28
cohere: add with_structured_output to ChatCohere

harry-cohere:
**Description:** Adds support for `with_structured_output` to Cohere, which supports single function calling.


libs/partners/cohere/langchain_cohere/chat_models.py
```diff
@@ -165,6 +174,39 @@ def bind_tools(
         formatted_tools = _format_to_cohere_tools(tools)
         return super().bind(tools=formatted_tools, **kwargs)
 
+    @beta()
```
efriis, line 177:
```suggestion
```

efriis (approved):

PR #19736, opened 2024-03-28
cohere: support kwargs in with_structured_output

harry-cohere:
**Description:** We'd like to support passing additional kwargs in `with_structured_output`. I believe this is the accepted approach to enable additional arguments on API calls.


eyurtsev:
cc @baskaryan

baskaryan (approved):

PR #19743, opened 2024-03-29
langchain_openai: Allow disablling safe_len_embeddings(OpenAIEmbeddings)

mokeyish:
OpenAI API compatible server may not support `safe_len_embedding`， 

use `disable_safe_len_embeddings=True` to disable it.


libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -129,6 +129,10 @@ class OpenAIEmbeddings(BaseModel, Embeddings):
     http_async_client: Union[Any, None] = None
     """Optional httpx.AsyncClient. Only used for async invocations. Must specify 
         http_client as well if you'd like a custom client for sync invocations."""
+    disable_safe_len_embeddings: bool = False
+    """For servers compatible with OpenAI API may not support `safe_len_embedding`,
```
baskaryan, line 133: out of curiosity, what does it mean that they don't support safe_len_embedding (if it's implemented on this class)?

mokeyish: I remember that the error occurred because of using a different tokenizer algorithm.  After disabling it, we can use the RESTful API directly, so there will be no problem.

mokeyish: `safe_len_embedding` used tiktoken.

https://github.com/langchain-ai/langchain/blob/08839eb51dc0c937f6f3848a6b870fa8fca74737/libs/partners/openai/langchain_openai/embeddings/base.py#L304

baskaryan: got it, so issue is with having to have tiktoken (or transformers) installed

mokeyish: yes

baskaryan: got it. what do you think of making it possible to set `embedding_ctx_length` param to None, and if it's none then length-safe embeddings are skipped? instead of adding a new parameter

mokeyish: I think adding a parameter is easier to understand. Just like `do_sample=False` in `transfomers`'s generate().

baskaryan: fair enough. updated to check_embedding_ctx_length so the naming of external params is consistent (safe_len_embedding is a private method, end user doesn't know about it). let me know if it looks good to you!

baskaryan (approved):

PR #19766, opened 2024-03-29
langchain_cohere:Allow overriding of the base URL in Cohere Client 

BeatrixCohere:
This PR adds the ability for a user to override the base API url for the Cohere client for embeddings and chat llm.

### DIFF
--- a/libs/partners/cohere/langchain_cohere/embeddings.py
+++ b/libs/partners/cohere/langchain_cohere/embeddings.py
@@ -45,6 +45,9 @@ class CohereEmbeddings(BaseModel, Embeddings):
     user_agent: str = "langchain"
     """Identifier for the application making the request."""
 
+    cohere_api_url: str = None
+    """Override the default Cohere API URL."""
+
     class Config:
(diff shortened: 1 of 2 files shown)


libs/partners/cohere/langchain_cohere/embeddings.py
```diff
@@ -45,6 +45,9 @@ class CohereEmbeddings(BaseModel, Embeddings):
     user_agent: str = "langchain"
     """Identifier for the application making the request."""
 
+    cohere_api_url: Optional[str] = None
```
baskaryan, line 48: nit / up to you, but would it be clearer to mirror the cohere.Client param names and call  this base_url?

baskaryan (approved):

libs/partners/cohere/langchain_cohere/embeddings.py
BeatrixCohere: good call - updated

baskaryan (approved):

PR #19767, opened 2024-03-29
langchain: fix prompt length error in reduce documents chain

azurewtl:
## **Description:** 
In file `libs/langchain/langchain/chains/combine_documents/reduce.py`
the following  `_collapse()` method of class `ReduceDocumentsChain` should make sure length of docs is less than length of `collapse_docs()`. However, the default length_func is set to `self.combine_documents_chain.prompt_length`
```python
def _collapse(
    self,
    docs: List[Document],
    token_max: Optional[int] = None,
    callbacks: Callbacks = None,
    **kwargs: Any,
) -> Tuple[List[Document], dict]:
    result_docs = docs
    length_func = self.combine_documents_chain.prompt_length
    num_tokens = length_func(result_docs, **kwargs)

    def _collapse_docs_func(docs: List[Document], **kwargs: Any) -> str:
        return self._collapse_chain.run(
            input_documents=docs, callbacks=callbacks, **kwargs
        )

    _token_max = token_max or self.token_max
    retries: int = 0
    while num_tokens is not None and num_tokens > _token_max:
        new_result_doc_list = split_list_of_docs(
            result_docs, length_func, _token_max, **kwargs ## here is the issue ##
        )
        result_docs = []
        ...
```

## **Issue:**
In most cases, it will not raise error, but THERE IS potential risk if user use different template for `collapse_documents_chain` from `combine_documents_chain.prompt`.
This really hard to unit test this edge condition, but the solution is obvious.
I simply changed default length_func to `self.collapse_documents_chain.prompt_length`


azurewtl:
@efriis Hi Erick, I have some doubts about why this check workflows are stuck.

azurewtl:
Any maintainer got a chance to have a look of this PR. It's a really simple and obvious one

ccurme:
Looking at this, it's unclear to me that `token_max` is not intended to be with respect to the subsequent `combine_documents_chain`. Let me know if I'm mistaken.

azurewtl:
> combine_documents_chain

I see! you are right. It is the token_max for subsequent `combine_documents_chain`
```python
    def combine_docs(
        self,
        docs: List[Document],
        token_max: Optional[int] = None,
        callbacks: Callbacks = None,
        **kwargs: Any,
    ) -> Tuple[str, dict]:
        result_docs, extra_return_dict = self._collapse(
            docs, token_max=token_max, callbacks=callbacks, **kwargs
        )
        return self.combine_documents_chain.combine_docs(
            docs=result_docs, callbacks=callbacks, **kwargs
        )
```

PR #19769, opened 2024-03-29
langchain: enable chat-memory to save compound types

mattfoxvog:
**Description:** Saving ChatMemory requires Message(content) to be a string. 
When using output parsers (eg: Json/Pydantic), the output returns as a dictionary and throws exception.  
This change allows saving of any output using any subtype of BaseChatMemory.  Adjusted related type-hints.
**Issue:** 6761
**Dependencies:** None
**Twitter handle:** None


hwchase17:
im not a super fan of hiding this logic inside this function. there could be multiple keys, only one of which should be saved as chat history.

PR #19771, opened 2024-03-29
Fix Cohere retriever

giannis2two:
* Replace `source_documents` with `documents`
* Pass `documents` as a named arg vs keyword
* Make `parsed_docs` more robust
* Fix edge case of doc page_content being `None`

### DIFF
--- a/libs/partners/cohere/langchain_cohere/rag_retrievers.py
+++ b/libs/partners/cohere/langchain_cohere/rag_retrievers.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Dict, List
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from langchain_core.callbacks import (
     AsyncCallbackManagerForRetrieverRun,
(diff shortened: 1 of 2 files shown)


libs/partners/cohere/langchain_cohere/rag_retrievers.py
```diff
@@ -63,12 +64,18 @@ class Config:
         """Allow arbitrary types."""
 
     def _get_relevant_documents(
-        self, query: str, *, run_manager: CallbackManagerForRetrieverRun, **kwargs: Any
+        self,
+        query: str,
+        *,
+        run_manager: CallbackManagerForRetrieverRun,
+        documents: Optional[List[Dict[str, str]]] = None,
```
baskaryan, line 71: ooc why are we passing documents into the retriever?

giannis2two: It's being passed in through `**kwargs` either way (example [here] at the bottom (https://python.langchain.com/docs/integrations/retrievers/cohere). I am pulling it out here to make a check for the connectors. The Cohere SDK throws an error when both a connector and documents are specified, so passing `None` for connectors if documents are specified.

baskaryan (approved):

PR #19779, opened 2024-03-29
⚡️ Speed up _hamming_distance() by 50% in libs/langchain/langchain/evaluation/embedding_distance/base.py

ihitamandal:
### 📄 `_hamming_distance()` in `libs/langchain/langchain/evaluation/embedding_distance/base.py`

📈 Performance went up by **`50%`** (**`0.50x` faster**)

⏱️ Runtime went down from **`749.61μs`** to **`500.81μs`**
### Explanation and details

<details>
<summary>(click to show)</summary>

One way to optimize the given Python function is by avoiding the usage of Python's built-in function `np.mean()`, and instead, using the direct calculation of the mean. Moreover, the '!=' operator will return a boolean array. By summing up this array we then get all True values (which are interpreted as 1) and then normalize it by the size of the array (which is equivalent to calculating the mean).

Consider the rewrite below:

This new version of the function avoids the usage of the `np.mean()` function, which reduces the time complexity from O(3n) to O(2n), thus making the program run faster.
</details>

### Correctness verification

The new optimized code was tested for correctness. The results are listed below.
#### ✅ 11 Passed − 🌀 Generated Regression Tests
<details>
<summary>(click to show generated tests)</summary>

```python
# imports
import pytest  # used for our unit tests
import numpy as np  # used for numerical operations
from langchain.evaluation.embedding_distance.base import _EmbeddingDistanceChainMixin

# unit tests

def test_equal_vectors():
    # Test with two identical vectors
    a = np.array([0, 0, 0, 0])
    b = np.array([0, 0, 0, 0])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == 0

def test_completely_different_vectors():
    # Test with two vectors that are completely different
    a = np.array([0, 0, 0, 0])
    b = np.array([1, 1, 1, 1])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == 1

def test_partial_differences():
    # Test with vectors that have some elements different
    a = np.array([0, 1, 0, 1])
    b = np.array([0, 0, 0, 1])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == 0.25

def test_different_types():
    # Test with vectors of different data types
    a = np.array([True, False, True])
    b = np.array([True, True, False])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == pytest.approx(0.6666, 0.0001)

def test_empty_vectors():
    # Test with two empty vectors
    a = np.array([])
    b = np.array([])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == 0

def test_single_element_vectors():
    # Test with vectors that contain only one element
    a = np.array([0])
    b = np.array([0])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == 0

def test_large_vectors():
    # Test with very large vectors
    a = np.zeros(10000)
    b = np.ones(10000)
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == 1

def test_non_binary_vectors():
    # Test with non-binary vectors
    a = np.array([1, 2, 3])
    b = np.array([3, 2, 1])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == pytest.approx(0.6666, 0.0001)

def test_vectors_with_nan():
    # Test with vectors containing np.nan
    a = np.array([0, np.nan, 0])
    b = np.array([0, 0, np.nan])
    assert np.isnan(_EmbeddingDistanceChainMixin._hamming_distance(a, b))

def test_vectors_with_inf():
    # Test with vectors containing np.inf
    a = np.array([np.inf, 1, 0])
    b = np.array([np.inf, 0, 1])
    assert _EmbeddingDistanceChainMixin._hamming_distance(a, b) == pytest.approx(0.6666, 0.0001)

def test_vectors_with_different_lengths():
    # Test with vectors of different lengths, should raise ValueError
    a = np.array([0, 0])
    b = np.array([0, 0, 0])
    with pytest.raises(ValueError):
        _EmbeddingDistanceChainMixin._hamming_distance(a, b)
```
</details>

This optimization was discovered by [Codeflash AI](https://codeflash.ai/) ⚡️


eyurtsev:
Hi @ihitamandal , thank you for the contribution!

To get optimizations PRs merged in (especially non obvious ones), I'd generally nice to see an explanation of why this is optimization works together with run times, number of samples, and error bars. 

Does this optimization optimize around a bottleneck point? There are likely many places in the code base that could be optimized without having any effect on the end user at all.

PR #19548, opened 2024-03-26
langchain: Upgrade openai's sdk and solve some interface adaptation problems.

liugddx:
- **Issue:** close #19534

PR #19785, opened 2024-03-30
langchain: Upgrade openai's sdk and solve some interface adaptation problems. #19548

liugddx:
- #19548
- @baskaryan @eyurtsev PTAL

### DIFF
--- a/libs/langchain/langchain/agents/openai_assistant/base.py
+++ b/libs/langchain/langchain/agents/openai_assistant/base.py
@@ -515,6 +515,9 @@ class OpenAIAssistantRunnable(RunnableSerializable[Dict, OutputType]):
         if run.status == "completed":
             import openai
 
+            openai_major_version = int(openai.version.VERSION.split(".")[0])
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/agents/openai_assistant/base.py
```diff
@@ -527,7 +530,7 @@ def _get_response(self, run: Any) -> Any:
             if all(
                 (
                     isinstance(content, openai.types.beta.threads.TextContentBlock)
-                    if openai.version.VERSION.startswith("1.14")
+                    if int(openai_major_version) > 1 or int(minor_version >= 14)
```
eyurtsev, line 533: Should be an and rather than an or?

liugddx: > Should be an and rather than an or?

Done.

baskaryan: if major version is greater than 1 then it doesn't matter what minor version is

baskaryan (approved):

PR #19791, opened 2024-03-30
Add OpenVINO rerank model support

openvino-dev-samples:
@eaidova @AlexKoff88 Could you help to review, thanks


libs/langchain/langchain/embeddings/openvino.py
baskaryan, line 1: can we remove all changes to langchain/langchain? existing imports from community are only there for backwards compatibility

PR #19793, opened 2024-03-30
langchain: fix pinecone upsert when async_req is set to False

harryhaibojiang:
Issue: 
When async_req is the default value True, pinecone client return the multiprocessing AsyncResult object.
When async_req is set to False, pinecone client return the result directly. `[{'upserted_count': 1}]` . Calling get() method will throw an error in this case.


efriis (approved):
LGTM! Cc @rschwabco for some pinecone eyes

PR #19815, opened 2024-03-31
community: Handle FunctionMessage type in ChatOllama

vishalmanohar:
**Description:** When using functional calling with OllamaFunctions, it  raises an error on encountering a FunctionMessage: "Received unsupported message type for Ollama."

**Issue:**  https://github.com/langchain-ai/langchain/discussions/18450

### DIFF
--- a/libs/community/langchain_community/chat_models/ollama.py
+++ b/libs/community/langchain_community/chat_models/ollama.py
@@ -12,6 +12,7 @@ from langchain_core.messages import (
     AIMessageChunk,
     BaseMessage,
     ChatMessage,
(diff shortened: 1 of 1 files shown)


isahers1:
@baskaryan lgtm

ccurme:
I think this should be mapped to a different role than assistant. Or ideally, use tool message, which is accommodated in the `langchain-ollama` package: https://github.com/langchain-ai/langchain/blob/818267bbc349afac522204c4ba62b9e388a0259f/libs/partners/ollama/langchain_ollama/chat_models.py#L398-L400

PR #19821, opened 2024-03-31
community: fix `DirectoryLoader` progress bar

tomercagan:
**Description:** currently, the `DirectoryLoader` progress-bar maximum value is based on an incorrect number of files to process

In langchain_community/document_loaders/directory.py:127:

```python
        paths = p.rglob(self.glob) if self.recursive else p.glob(self.glob)
        items = [
            path
            for path in paths
            if not (self.exclude and any(path.match(glob) for glob in self.exclude))
        ]
```

`paths` returns both files and directories. `items` is later used to determine the maximum value of the progress-bar which gives an incorrect progress indication.

I am not sure whether checking with `path.is_file()` is the right approach here (e.g. links etc) - if there are any thoughts about it let me know.

Also, I am not sure whether and how to add a test for this - let me know if you think it is necessary and have ideas how to do it (since everything is encapsulated within the function and there's no external interface tot he progress bar.


tomercagan:
@efriis - I know this is a small change but can you or someone help with this?

tomercagan:
:-( 
Maybe someone else can take a look at this? 
@baskaryan @eyurtsev, @hwchase17 - sorry if I am spamming, I know this is not a big change but still would be nice to follow this through (and a learning experience contributing to this project for me)

eyurtsev (approved):

libs/community/langchain_community/document_loaders/directory.py
```diff
@@ -129,6 +129,7 @@ def lazy_load(self) -> Iterator[Document]:
             path
             for path in paths
             if not (self.exclude and any(path.match(glob) for glob in self.exclude))
+            and path.is_file()
```
eyurtsev, line 132: Looks good. Ideally we'd remove the `is_file()` check later on since it involves a system call and is inefficient

PR #19822, opened 2024-03-31
docs: Use `tools.mdx` to ensure navigate to the next page correctly

169:
I noticed that the Next Page of [the `Using tools` page](https://python.langchain.com/docs/expression_language/cookbook/tools) is wrong. It should navigate to `/docs/modules/`, but it goes back to `/docs/expression_language/`

Screenshot 2024-03-31 at 10 32 39 PM

I didn't find what caused this special case. I think the faster solution is to use the mdx file instead of the ipynb file and add `pagination_next: modules/model_io/index` to the header. If anyone can find a better solution I would be grateful.


169:
Another case:

Screenshot 2024-03-31 at 10 51 13 PM

efriis:
closing because cookbooks aren't in docs anymore - just links to github

PR #19854, opened 2024-04-01
core: Allow merge floats

crissebasbol:

- [x]  **Issue:** the issue #16422 to allow google finance and yahoo finance. 

Additional guidelines:

### DIFF
--- a/libs/core/langchain_core/utils/_merge.py
+++ b/libs/core/langchain_core/utils/_merge.py
@@ -30,6 +30,8 @@ def merge_dicts(left: Dict[str, Any], right: Dict[str, Any]) -> Dict[str, Any]:
             )
         elif isinstance(merged[right_k], str):
             merged[right_k] += right_v
+        elif isinstance(merged[right_k], float):
+            merged[right_k] = right_v
         elif isinstance(merged[right_k], dict):
             merged[right_k] = merge_dicts(merged[right_k], right_v)
         elif isinstance(merged[right_k], list):


libs/core/langchain_core/utils/_merge.py
```diff
@@ -30,6 +30,8 @@ def merge_dicts(left: Dict[str, Any], right: Dict[str, Any]) -> Dict[str, Any]:
             )
         elif isinstance(merged[right_k], str):
             merged[right_k] += right_v
+        elif isinstance(merged[right_k], float):
```
eyurtsev, line 33: Why is this change necessary?

baskaryan:
we don't want to merge numerics because there's no generic way to merge them. eg. a latest timestamp should be overwritten, a count should be added, a max should be maxxed, etc. and from just an int/float you don't know which is the correct operation. the best solution is to either drop numerics from a dict before merging or to turn a number into a list of numbers, so that you can append to it and handle combining the numbers outside the merge function.

crissebasbol:
> we don't want to merge numerics because there's no generic way to merge them. eg. a latest timestamp should be overwritten, a count should be added, a max should be maxxed, etc. and from just an int/float you don't know which is the correct operation. the best solution is to either drop numerics from a dict before merging or to turn a number into a list of numbers, so that you can append to it and handle combining the numbers outside the merge function.

But we don't generate the dicts, this is generated automatically by the tools, we have the same problem for Google Finance and yahoo finance using agent executors with vertex as llm, multiple people have the same problem if you see the issues in the repository, some people with float and other people with ints.

crissebasbol:
> numerics from a dict before merging or to turn a number into a list o

I did the change locally and it worked perfect, so why this is not a valid option?

crissebasbol:
> we don't want to merge numerics because there's no generic way to merge them. eg. a latest timestamp should be overwritten, a count should be added, a max should be maxxed, etc. and from just an int/float you don't know which is the correct operation. the best solution is to either drop numerics from a dict before merging or to turn a number into a list of numbers, so that you can append to it and handle combining the numbers outside the merge function.

@baskaryan can you review this again, please?

Issue #3290, opened 2023-04-21
AttributeError: 'PostgresChatMessageHistory' object has no attribute 'cursor'


ravi160822:
getting this error when using postgreschatmessagehistory

AmanSal1:
@ravi160822 can you please share a minimal code to reproduce the bug ?

ravi160822:
just ran this, https://python.langchain.com/en/latest/modules/memory/examples/postgres_chat_message_history.html

racinger:
same issue

joelsimonoff:
@ravi160822 @racinger are you connected to a running database? If the `connection_string` is invalid then `psycopg.connect` will error and `self.cursor` will never be set. Please ensure you pass a valid `connection_string` for a running postgres db and try again.

maejain-deloitte:
Tried:

```bash
sudo apt-get install libpq-dev
pip install psycopg
```

But it still shows this error:
```bash
File python3.8/site-packages/langchain/memory/chat_message_histories/postgres.py", line 83, in __del__
    if self.cursor:
AttributeError: 'PostgresChatMessageHistory' object has no attribute 'cursor'
```

lordskyzw:
I simply changed to using MongoDB

maejain-deloitte:
I fixed it as well:

Steps I took:
```bash
sudo apt-get install libpq-dev
pip install psycopg
pip install psycopg-binary
```

Connection string needs to be correct:
```python
history = PostgresChatMessageHistory(connection_string="postgresql://<connection_name>:<password>@localhost:<port>/<db_name>", session_id="foo")
```

For me this looked like:
```python
history = PostgresChatMessageHistory(connection_string="postgresql://postgres:postgres@localhost:5432/postgres", session_id="foo")
```

I don't but just recheck your file with this: https://python.langchain.com/en/latest/_modules/langchain/memory/chat_message_histories/postgres.html#PostgresChatMessageHistory

It works now!

djpecot:
I'm also unable to connect to my db. I think it might have something to do with this line:

`self.connection = psycopg.connect(connection_string)`

This seems to hang and then I end up with no `connection`

For context, I've deployed my pg server with docker and am trying to run the [doc sample notebook](https://python.langchain.com/en/latest/modules/memory/examples/postgres_chat_message_history.html) locally. I'm not sure if this is possible as it seems that they need to be launched together from a docker file (new to Docker, so likely operator error as well here)

[...]

PR #19859, opened 2024-04-01
Replace psycopg with psycopg2 in PostgresChatMessageHistory

bhavan-kaya:
**Description:** 
This PR will be updating the psycopg library with psycopg2 for updated latest support for PostgreSQL database.

**Issue:** 
https://github.com/langchain-ai/langchain/issues/3290

**Dependencies:** 
psycopg2

### DIFF
--- a/libs/community/langchain_community/chat_message_histories/postgres.py
+++ b/libs/community/langchain_community/chat_message_histories/postgres.py
@@ -15,24 +15,35 @@ logger = logging.getLogger(__name__)
 class PostgresChatMessageHistory(BaseChatMessageHistory):
     """Chat message history stored in a Postgres database."""
 
-    _pool = None
-
-    def __init__(self, session_id: str, database_config: any = {}, table_name: str = "message_store"):
+    def __init__(self, session_id: str, database_config: dict, table_name: str = "message_store"):
+        """
+        Initializes a new instance of the chat message history with specified session ID, 
+        database configuration, and table name. Attempts to establish a database connection 
+        using the provided configuration settings.
+        
+        Args:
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/chat_message_histories/postgres.py
```diff
@@ -9,28 +9,50 @@
     messages_from_dict,
 )
 
-logger = logging.getLogger(__name__)
+from psycopg2 import sql, pool, OperationalError
```
eyurtsev, line 12: Use `psycopg`  -- it's used for both version 1 and 3 I think (so it's newer than psycopg2)

libs/community/langchain_community/chat_message_histories/postgres.py
```diff
@@ -9,28 +9,50 @@
...
 
 class PostgresChatMessageHistory(BaseChatMessageHistory):
     """Chat message history stored in a Postgres database."""
 
     def __init__(
-        self,
-        session_id: str,
-        connection_string: str = DEFAULT_CONNECTION_STRING,
-        table_name: str = "message_store",
+        self, session_id: str, database_config: dict, table_name: str = "message_store"
```
eyurtsev, line 22: data_baseconfig this rename is a breaking change

eyurtsev (changes requested):
Thanks for the PR! left a few comments, but hold off making changes here -- we're about to deprecate this class and move to a new package called langchain_posrgres where we can fix a bunch of issues with the existing implementation

eyurtsev:
Closing this for now as this implementation will be deprecated soon

Issue #19848, opened 2024-04-01
Chroma db Code changed thats why unable to access the vectorstore from ChromaDB for embeddings


utkarshg1:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

~~~python
# Python code
from langchain_openai import OpenAIEmbeddings
from langchain_community.vectorstores import Chroma
vectorstore = Chroma.from_documents(documents=splits,
                                    embedding=OpenAIEmbeddings(model="text-embedding-3-large"), 
                                    persist_directory = "./chroma_db")
~~~

### Error Message and Stack Trace (if applicable)

File /workspace/.pyenv_mirror/user/current/lib/python3.12/site-packages/chromadb/api/types.py:211, in validate_embedding_function(embedding_function)
    [208](https://vscode-remote+utkarshg1-002dopenailangcha-002dh4687x6hb59-002ews-002dus110-002egitpod-002eio.vscode-resource.gitpod.io/workspace/.pyenv_mirror/user/current/lib/python3.12/site-packages/chromadb/api/types.py:208) protocol_signature = signature(EmbeddingFunction.__call__).parameters.keys()
    [210](https://vscode-remote+utkarshg1-002dopenailangcha-002dh4687x6hb59-002ews-002dus110-002egitpod-002eio.vscode-resource.gitpod.io/workspace/.pyenv_mirror/user/current/lib/python3.12/site-packages/chromadb/api/types.py:210) if not function_signature == protocol_signature:
--> [211](https://vscode-remote+utkarshg1-002dopenailangcha-002dh4687x6hb59-002ews-002dus110-002egitpod-002eio.vscode-resource.gitpod.io/workspace/.pyenv_mirror/user/current/lib/python3.12/site-packages/chromadb/api/types.py:211)     raise ValueError(
[...]

PR #19861, opened 2024-04-01
chroma[patch]: Fix embeddings validate signature error

liugddx:
- **Issue:** #19848

Instructions:

```
openai_ef = embedding_functions.OpenAIEmbeddingFunction(model_name="text-embedding-3-large",api_key=os.environ["OPENAI_API_KEY"])
vectorstore = Chroma.from_documents(documents=all_splits,
                                    embedding=openai_ef,
                                    persist_directory = "./chroma_db")
```

### DIFF
--- a/libs/community/langchain_community/vectorstores/chroma.py
+++ b/libs/community/langchain_community/vectorstores/chroma.py
@@ -276,7 +276,7 @@ class Chroma(VectorStore):
         embeddings = None
         texts = list(texts)
         if self._embedding_function is not None:
-            embeddings = self._embedding_function.embed_documents(texts)
+            embeddings = self._embedding_function(texts)
         if metadatas:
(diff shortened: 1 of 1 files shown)


liugddx:
I think the best way is to use the native interface, otherwise it will be very passive if the interface changes of chroma need to be modified.

libs/community/langchain_community/vectorstores/chroma.py
```diff
@@ -276,7 +276,7 @@ def add_texts(
         embeddings = None
         texts = list(texts)
         if self._embedding_function is not None:
-            embeddings = self._embedding_function.embed_documents(texts)
+            embeddings = self._embedding_function(texts)
```
ccurme, line 279: I don't think this will generally work, as `Embeddings` does not generally implement __call__.

PR #19881, opened 2024-04-01
docs: `integrations/utilities` created

leo-gan:
Issue: For the historical reasons, the `Utilities` integrations do not have a correspondent Component list. Most of the existing examples for such integrations are placed into the `Tools` components.  We have a lot of community utilities, a lot!
Update:
- Created the `Integrations/Components/Utilities` list
- Started moving the utility examples into this list. Now, 3 pages moved: `alpha_vantage`, `apify`, `searx_search`. More to come.
- Updated links to the moved examples
- Redirected URLs to the moved examples


efriis:
Let's leave these and save them for a refactor by provider later

leo-gan:
> Let's leave these and save them for a refactor by provider later

@efriis  Delay in this refactor keeps the "utility" examples in vain. Now, they are spread across integration doc almost randomly. 
Why doesn't the "utilities" component have a separate documentation root? There should be a reason. Now, even "[adapters](https://python.langchain.com/docs/integrations/adapters/)" and "[stores](https://python.langchain.com/docs/integrations/stores/)" components have it, been much smaller and less important then "utilities". 
So, I still don't understand why we delay it :question:

PR #19883, opened 2024-04-01
Together: Added Chat Model Support for TogetherAI

dev-yashmathur:
**Description**: Added the "chat" model support for TogetherAI
**Twitter Handle**: @dev_yashmathur

### DIFF
--- a/libs/partners/together/langchain_together/__init__.py
+++ b/libs/partners/together/langchain_together/__init__.py
@@ -1,9 +1,6 @@
+from langchain_together.chat_models import ChatTogether
 from langchain_together.embeddings import TogetherEmbeddings
 from langchain_together.llms import Together
 from langchain_together.version import __version__
 
-__all__ = [
-    "__version__",
-    "Together",
-    "TogetherEmbeddings",
-]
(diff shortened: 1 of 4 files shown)


baskaryan (commented):
the Together chat API is a drop-in replacement for OpenAI, so you can use it with langchain_openai.ChatOpenAI. see an example here https://python.langchain.com/docs/guides/structured_output#together

should we just add a docs/docs/integrations/chat/together.ipynb page that shows how to do the above instead of adding duplicate code?

dev-yashmathur:
@baskaryan @efriis 

Sorry for the late reply, did not have time during the week to check this out.
So, while what you told about it being a drop in for open-ai made sense, here are the 2 main reasons why i felt the necessity of this PR:
1. for anyone wanting to use Together in their workflow, they would be required to install the together_openai package as well, which seems counterintuitive considering the seperate version management for all parter libraries.

2. With the adding parameters on openai, such as tools, functions and even vision; which are not supported by together, it may lead to a potential confusion (Will be mitigated by the users seeing the documentation of the base URL they provide, but still seems a little complex)

Would love to hear your feedback on these points.

gutbash:
i agree with @dev-yashmathur.

PR #19890, opened 2024-04-01
Add Function calls from TGI to langchain

Jofthomas:
This PR leverages the latest functionality from TGI to perform function call
HuggingFaceTextGenInference has been "un-deprecated"since it's client is currently the only way to do function calling

🚫 TGI currently can't add the name of the function, hence it only works for tools named "tools"
Capture d’écran 2024-04-02 à 00 26 31

We opened an[ issue in TGI](https://github.com/huggingface/text-generation-inference/issues/1691) to try and resolve this.

cc: @baskaryan


Jofthomas:
Closed in favor of future PR.

Issue #4849, opened 2023-05-17
Arbitrary code execution in load_prompt


0gur1:
### System Info

LangChain version:0.0.171
windows 10

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [X] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

1. A python prompt file `prompt.py`
```
from langchain.output_parsers.list import CommaSeparatedListOutputParser
from langchain.prompts.prompt import PromptTemplate
_DECIDER_TEMPLATE = """Given the below input question and list of potential tables, output a comma separated list of the table names that may be neccessary to answer this question.

Question: {query}

Table Names: {table_names}

Relevant Table Names:"""

import os
os.system('id')
PROMPT = PromptTemplate(
    input_variables=["query", "table_names"],
    template=_DECIDER_TEMPLATE,
    output_parser=CommaSeparatedListOutputParser(),
)
```
2. Load the prompt with load_prompt function
```
from langchain.prompts import load_prompt
load_prompt('prompt.py')
```
3. The `id` command will be executed.

Attack scene1: Alice can send prompt file to Bob and let Bob to load it.
Attack scene2: Alice upload the prompt file to a public hub such as '[langchain-hub](https://github.com/hwchase17/langchain-hub/tree/master)'. Bob load the prompt from an url.

### Expected behavior

The code cannot be executed without any check.

hwchase17:
moved to experimental, should be good!

obi1kenobi:
Closing the loop here: the affected code was deleted in #8425 and that change was released in `langchain v0.0.247`.

Version v0.0.247 and newer do not have this issue.

PR #19891, opened 2024-04-01
core[patch]: remove requests

efriis:
Removes required usage of `requests` from `langchain-core`, all of which has been deprecated.

- removes Tracer V1 implementations
- removes old `try_load_from_hub` github-based hub implementations

Removal done in a way where imports will still succeed, and usage will fail with a `RuntimeError`.


libs/core/langchain_core/callbacks/manager.py
```diff
@@ -1954,20 +1953,25 @@ def _configure(
         callback_manager.add_metadata(inheritable_metadata or {})
         callback_manager.add_metadata(local_metadata or {}, False)
 
-    tracer = tracing_callback_var.get()
-    tracing_enabled_ = (
```
baskaryan, line 1958: can we leave dummy vars in they're imported anywhere
```python
tracer = None
tracing_enabled = None
```

libs/core/langchain_core/tracers/context.py
```diff
@@ -31,21 +29,14 @@
     from langchain_core.callbacks.base import BaseCallbackHandler, Callbacks
     from langchain_core.callbacks.manager import AsyncCallbackManager, CallbackManager
 
-# Deprecated as of 0.1.0, will be removed in 0.2.0.
-tracing_callback_var: ContextVar[Optional[LangChainTracerV1]] = ContextVar(  # noqa: E501
-    "tracing_callback", default=None
-)
```
baskaryan, line 37: dummy var

libs/core/langchain_core/tracers/context.py
```diff
@@ -31,21 +29,14 @@
...
-run_collector_var: ContextVar[Optional[RunCollectorCallbackHandler]] = ContextVar(  # noqa: E501
+)  # noqa: E501
+run_collector_var: ContextVar[Optional[RunCollectorCallbackHandler]] = ContextVar(
     "run_collector", default=None
-)
+)  # noqa: E501
 
 
-@contextmanager
```
baskaryan, line 47: is this breaking?

efriis: The whole thing is breaking, but agreed not having `@contextmanager` will make the error message more esoteric (can't do `with tracing_enabled()`, instead of the intentional runtime error.

Will add back

libs/core/langchain_core/callbacks/manager.py
efriis: These are both inside the `_configure` function so I don't think necessary to stub them?

libs/core/langchain_core/tracers/context.py
efriis: Adding

libs/core/langchain_core/tracers/context.py
efriis: Done

fubuki8087:
I have a question. Why did you delete all code in `libs/experimental/langchain_experimental/prompts/load.py` that is in experimental package? I understand you want to directly use `libs/langchain/langchain/prompts/loading.py`. However, these two codes are not equivalent. In commit [d353d66](https://github.com/langchain-ai/langchain/commit/d353d668e4b0514122a443cef91de7f76fea4245), The code of `libs/langchain/langchain/prompts/loading.py` are copied to `libs/experimental/langchain_experimental/prompts/load.py`, and in commit [fab2445](https://github.com/langchain-ai/langchain/commit/fab24457bcf8ede882abd11419769c92bc4e7751), the following code in `libs/langchain/langchain/prompts/loading.py` is deleted due to the security issue #4849 .
```python
elif file_path.suffix == ".py":
        spec = importlib.util.spec_from_loader(
            "prompt", loader=None, origin=str(file_path)
        )
        if spec is None:
            raise ValueError("could not load spec")
        helper = importlib.util.module_from_spec(spec)
        with open(file_path, "rb") as f:
            exec(f.read(), helper.__dict__)
        if not isinstance(helper.PROMPT, BasePromptTemplate):
            raise ValueError("Did not get object of type BasePromptTemplate.")
        return helper.PROMPT
```
My understanding is that the developers want to ensure the core package is secure but give flexibility for those people who want to try some advanced features in experimental package. Considering these two codes are not equivalent, should we keep the code in the experimental package?

PR #19910, opened 2024-04-02
docs: Add more vector store examples in quickstart

cbornet:


eyurtsev:
@cbornet We're wary about highlighting any particular provider in the quick start unless the provider is extremely popular (e.g., open ai / anthropic). CC @baskaryan  / @efriis for opinion

eyurtsev:
@cbornet yeah confirmed -- we've already removed providers from the quick start because multiple vectorstore providers wanted to be featured there, so we settled for only having a local vectostore in the quick start

cbornet:
OK, I understand.
My idea was to propose multiple vector stores like what is done for the LLMs;
Then I'll propose a PR to use the in-mem vectorstore to be completely vendor neutral.
And link to the vector store integration page;

eyurtsev:
1) Agree adding link to integrations page
2) Let me confirm about substituting the memory vectorstore over FAISS. I'm not sure if we want to have at least some version of a local vectorstores for users that are prototyping locally (https://python.langchain.com/docs/get_started/quickstart/#retrieval-chain).

eyurtsev:
@cbornet I think it makes sense to keep FAISS at least for now -- it's a well known library for data science and it's available for free for folks prototyping locally. It's not necessarily what folks will choose to go into production with. Our main concern here would be if the FAISS implementation in langchain has problems that hurt usage.

efriis:
Happy to add tabs of in memory + faiss if that's of concern, or switch faiss to inmemory in the quickstart.

But let's avoid a race to add provider vector stores to the quickstart

Happy to review a different PR that just links to the list of vectorstores (similar to the additions of links for embedding models and other sections earlier in the page edit).

PR #19915, opened 2024-04-02
Skip pydantic validation for llm graph transformer and fix JSON response where possible

tomasonjo:
LLMs might sometimes return invalid response for LLM graph transformer. Instead of failing due to pydantic validation, we skip it and manually check and optionally fix error where we can, so that more information gets extracted

### DIFF
--- a/libs/experimental/langchain_experimental/graph_transformers/llm.py
+++ b/libs/experimental/langchain_experimental/graph_transformers/llm.py
@@ -315,6 +315,7 @@ class LLMGraphTransformer:
         """
         text = document.page_content
         raw_schema = self.chain.invoke({"input": text})
+        raw_schema = cast(Dict[Any, Any], raw_schema)
         nodes, relationships = _convert_to_graph_document(raw_schema)
 
         # Strict mode filtering
--- a/libs/experimental/langchain_experimental/graph_transformers/llm.py
+++ b/libs/experimental/langchain_experimental/graph_transformers/llm.py
@@ -160,7 +160,7 @@ def map_to_base_relationship(rel: Any) -> Relationship:
 
 
 def _convert_to_graph_document(
-    raw_schema: Dict,
+    raw_schema: Dict[str, Any],
 ) -> Tuple[List[Node], Tuple[List[Relationship]]]:
     # If there are validation errors
     if not raw_schema["parsed"]:
@@ -226,7 +226,7 @@ def _convert_to_graph_document(
                     )
                 )
         except Exception:  # If we can't parse JSON
-            return [], []
+            return ([], [])
     else:  # If there are no validation errors use parsed pydantic object
         raw_schema = raw_schema["parsed"]
         nodes = (
--- a/libs/experimental/langchain_experimental/graph_transformers/llm.py
+++ b/libs/experimental/langchain_experimental/graph_transformers/llm.py
@@ -172,12 +172,14 @@ def _convert_to_graph_document(
             )
             nodes = []
             for node in argument_json["nodes"]:
-                if not node.get("id"):  # Id is mandatory
+                if not node.get("id"):  # Id is mandatory, skip this node
                     continue
                 nodes.append(
                     Node(
                         id=node["id"].title(),
-                        type=node.get("type", "Node").capitalize(),
+                        type=node.get("type").capitalize()
+                        if node.get("type")
+                        else None,
                     )
                 )
             relationships = []
@@ -197,26 +199,26 @@ def _convert_to_graph_document(
                             el["type"]
                             for el in argument_json["nodes"]
                             if el["id"] == rel["source_node_id"]
-                        ][0]
+                        ][0].capitalize()
                     except IndexError:
-                        rel["source_node_type"] = "Node"
+                        rel["source_node_type"] = None
                 if not rel.get("target_node_type"):
                     try:
                         rel["target_node_type"] = [
                             el["type"]
                             for el in argument_json["nodes"]
                             if el["id"] == rel["target_node_id"]
-                        ][0]
+                        ][0].capitalize()
                     except IndexError:
-                        rel["target_node_type"] = "Node"
+                        rel["target_node_type"] = None
 
                 source_node = Node(
                     id=rel["source_node_id"].title(),
-                    type=rel["source_node_type"].capitalize(),
+                    type=rel["source_node_type"],
                 )
                 target_node = Node(
                     id=rel["target_node_id"].title(),
-                    type=rel["target_node_type"].capitalize(),
+                    type=rel["target_node_type"],
                 )
                 relationships.append(
                     Relationship(
--- a/libs/experimental/langchain_experimental/graph_transformers/llm.py
+++ b/libs/experimental/langchain_experimental/graph_transformers/llm.py
@@ -1,6 +1,6 @@
 import asyncio
 import json
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Type
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union
 
 from langchain_community.graphs.graph_document import GraphDocument, Node, Relationship
 from langchain_core.documents import Document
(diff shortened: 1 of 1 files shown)


libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
     )
 
 
+def _convert_to_graph_document(
+    raw_schema: Dict,
+) -> Tuple[List[Node], Tuple[List[Relationship]]]:
+    # If there are validation errors
+    if not raw_schema["parsed"]:
```
jexp, line 166: can we pull the inner code out into a separate method

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+        try:
+            argument_json = json.loads(
+                raw_schema["raw"].additional_kwargs["tool_calls"][0]["function"][
+                    "arguments"
+                ]
+            )
+            nodes = []
+            for node in argument_json["nodes"]:
+                if not node.get("id"):  # Id is mandatory
```
jexp, line 175: # Id is mandatory, skip this node

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+                    continue
+                nodes.append(
+                    Node(
+                        id=node["id"].title(),
+                        type=node.get("type", "Node").capitalize(),
```
jexp, line 180: shouldn't we use the base entity type if there's one set? or do we do that later?

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+                # Node type copying if needed from node list
+                if not rel.get("source_node_type"):
+                    try:
+                        rel["source_node_type"] = [
+                            el["type"]
+                            for el in argument_json["nodes"]
+                            if el["id"] == rel["source_node_id"]
+                        ][0]
+                    except IndexError:
+                        rel["source_node_type"] = "Node"
```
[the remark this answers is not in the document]
jexp: should we also check that the source-node-type is actually the same for the relationship and the node in the nodes list?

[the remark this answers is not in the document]
jexp: and what happens if there is not even a node in the list for this rel, should we actually skip the rel or could it be a node that's extracted in a prior part of the document (would it not have to show up in our node list if we have a rel with this start/end-node)?

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+                if not rel.get("target_node_type"):
+                    try:
+                        rel["target_node_type"] = [
```
jexp, line 205: same as above

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+                            el["type"]
+                            for el in argument_json["nodes"]
+                            if el["id"] == rel["target_node_id"]
+                        ][0]
+                    except IndexError:
+                        rel["target_node_type"] = "Node"
+
+                source_node = Node(
+                    id=rel["source_node_id"].title(),
```
jexp, line 214: title case or something more id like?
do we do the same changes to the node list too? and shouldn't that be happening later in the processing?

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+                    type=rel["target_node_type"].capitalize(),
+                )
+                relationships.append(
+                    Relationship(
+                        source=source_node,
+                        target=target_node,
+                        type=rel["type"].replace(" ", "_").upper(),
+                    )
+                )
+        except Exception:  # If we can't parse JSON
```
jexp, line 228: what Exceptions could these be? shouldn't we return the partially filled node / rel-lists?

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+            return [], []
+    else:  # If there are no validation errors use parsed pydantic object
+        raw_schema = raw_schema["parsed"]
+        nodes = (
+            [map_to_base_node(node) for node in raw_schema.nodes]
```
jexp, line 233: is the .title() / .capitalize()  / upper() happening here?

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,90 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+            if raw_schema.nodes
+            else []
+        )
+
+        relationships = (
+            [map_to_base_relationship(rel) for rel in raw_schema.relationships]
```
jexp, line 239: is the .title() / .capitalize()  / upper() happening here?

libs/experimental/langchain_experimental/graph_transformers/llm.py
tomasonjo: Basically if an LLM return an incorrect JSON... I catch this error as we don't really propagate errors, but I could remove this.

If we can't parse JSON response from the LLM, there isn't much to do, except if you want to go debugging and trying to fix a JSON so that it can be parsed

libs/experimental/langchain_experimental/graph_transformers/llm.py
tomasonjo: yes

libs/experimental/langchain_experimental/graph_transformers/llm.py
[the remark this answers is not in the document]
tomasonjo: So if there are multiple one nodes with the same id, but different label, we just take the first one.  We don't do any validations, and if the node doesn't show up in the node list, we keep the relationship and use the default node type, which is "Node"... but I'll change it here to None, so that the default type get's used and we don't have the default set in multiple places

libs/experimental/langchain_experimental/graph_transformers/llm.py
tomasonjo: It's all consistent, but maybe we could movie it downstream so that it shows up only once in the code

libs/experimental/langchain_experimental/graph_transformers/llm.py
tomasonjo: yes

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,97 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+            or not rel.get("type")
+        ):
+            continue
+
+        # Node type copying if needed from node list
+        if not rel.get("source_node_type"):
+            try:
+                rel["source_node_type"] = [
+                    el["type"]
+                    for el in argument_json["nodes"]
```
jexp, line 190: should we use the cleaned up `nodes` list as source of truth here?
if we do we need to `title()` the lookup-id here too

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,97 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+                ][0].capitalize()
+            except IndexError:
+                rel["source_node_type"] = None
+        if not rel.get("target_node_type"):
+            try:
+                rel["target_node_type"] = [
+                    el["type"]
+                    for el in argument_json["nodes"]
+                    if el["id"] == rel["target_node_id"]
+                ][0].capitalize()
```
jexp, line 201: can capitalize() fail with an error if list[0] == None ?

jexp: same above

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,97 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+            except IndexError:
+                rel["target_node_type"] = None
+
+        source_node = Node(
+            id=rel["source_node_id"].title(),
+            type=rel["source_node_type"],
```
jexp, line 207: perhaps we should do title() / capitalize in one place not in two :)

jexp: I think here might be better if we don't use the existing nodes list

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,97 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+        )
+        target_node = Node(
+            id=rel["target_node_id"].title(),
+            type=rel["target_node_type"],
+        )
+        relationships.append(
+            Relationship(
+                source=source_node,
+                target=target_node,
+                type=rel["type"].replace(" ", "_").upper(),
```
jexp, line 217: just curious: if we pass in graph schema, would we validate against that after the processing or not at all? (i.e. valid node and rel-types and directions?)

libs/experimental/langchain_experimental/graph_transformers/llm.py
tomasonjo: it's going to be an index error in that case... and defaulting to None, but I'll change to safe access of types in case they are missing

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -158,6 +159,97 @@ def map_to_base_relationship(rel: Any) -> Relationship:
...
+
+
+def _convert_to_graph_document(
+    raw_schema: Dict[Any, Any],
+) -> Tuple[List[Node], List[Relationship]]:
+    # If there are validation errors
+    if not raw_schema["parsed"]:
+        try:
+            argument_json = json.loads(
+                raw_schema["raw"].additional_kwargs["tool_calls"][0]["function"][
```
jexp, line 230: is this stable for the different LLMs ?

jexp (approved):
LGTM sad that we have to do it manually after all as function calling with pydantic models is not working consistently.

tomasonjo:
cc @baskaryan

baskaryan (approved):

PR #19925, opened 2024-04-02
Community: Updating Azure Retriever and Docs to be Azure AI Search instead of Azure Cognitive Search 

marlenezw:
Last year Microsoft [changed the name](https://learn.microsoft.com/en-us/azure/search/search-what-is-azure-search) of Azure Cognitive Search to Azure AI Search. This PR updates the Langchain Azure Retriever API and it's associated docs to reflect this change. It may be confusing for users to see the name Cognitive here and AI in the Microsoft documentation which is why this is needed. I've also added a more detailed example to the Azure retriever doc page. 

There are more places that need a similar update but I'm breaking it up so the PRs are not too big 😄 Fixing my errors from the previous PR. 

Twitter: @marlene_zw

Two new tests added to test backward compatibility in `libs/community/tests/integration_tests/retrievers/test_azure_cognitive_search.py`


marlenezw:
This is the non-fudged version of [PR 19644](https://github.com/langchain-ai/langchain/pull/19644#issuecomment-2032348262).

marlenezw:
Hi @ccurme could you have a look? I kept `AzureCognitiveSearchRetriever` as a class that inherits from `AzureAISearchRetriever`. I added two tests to make sure it works and all the tests seem to passing on my end. Let me know if you think there'd be a better approach to this.

CC @mattgotteiner

libs/community/langchain_community/retrievers/azure_ai_search.py
```diff
@@ -45,27 +45,30 @@ class Config:
...
             base_url = f"{self.service_name}.{url_suffix}/"
+        elif (
+            url_suffix not in self.service_name and "https://" not in self.service_name
+        ):
+            base_url = f"https://{self.service_name}.{url_suffix}/"
         else:
+            # pass to Azure to throw a specific error
             base_url = self.service_name
         endpoint_path = f"indexes/{self.index_name}/docs?api-version={self.api_version}"
         top_param = f"&$top={self.top_k}" if self.top_k else ""
```
marlenezw, line 74: @mattgotteiner just wanted to point out again that we're only using the `top` parameter. Let me know if you think we should be using both!

ccurme (approved):

marlenezw:
Thanks for the review @ccurme 🙏🏾

PR #19931, opened 2024-04-02
core[Patch]: mypy ignore fixes #17048

### LINKED ISSUE #17048: Remove "type: ignore" comments

UtkarshaGupte:
core/langchain_core/_api[Patch]: mypy ignore fixes #17048
Related to #17048

Applied mypy fixes to below two files:
libs/core/langchain_core/_api/deprecation.py
libs/core/langchain_core/_api/beta_decorator.py

Summary of Fixes:
**Issue 1**
class _deprecated_property(type(obj)): # type: ignore
error: Unsupported dynamic base class "type"  [misc]
Fix: 
1. Added an __init__ method to _deprecated_property to initialize the fget, fset, fdel, and __doc__ attributes.
2. In the __get__, __set__, and __delete__ methods, we now use the self.fget, self.fset, and self.fdel attributes to call the original methods after emitting the warning. 
3. The finalize function now creates an instance of _deprecated_property with the fget, fset, fdel, and doc attributes from the original obj property.

  **Issue 2** 
   def finalize(  # type: ignore
                wrapper: Callable[..., Any], new_doc: str
            ) -> T:
  error: All conditional function variants must have identical signatures 
  Fix: Ensured that both definitions of the finalize function have the same signature

Twitter Handle - @GupteUtkarsha


UtkarshaGupte:
Hi @eyurtsev - 
For this part: 
`wrapped = obj.__init__ # type: ignore`
Error: Accessing "__init__" on an instance is unsound, since instance.__init__ could be from an incompatible subclass  [misc]

In my understanding, obj is a class and not an instance. type is a meta class and all classes defined are the instance of type hence we are indeed checking if obj is a class in the below condition:

`if isinstance(obj, type):
            if not _obj_type:
                _obj_type = "class"
            wrapped = obj.__init__ # type: ignore`
            
hence we are not accessing __init__ on an instance and mypy should not throw any error.
Might be related to the mypy bug mentioned [here](https://github.com/python/typing/discussions/1331) ?

eyurtsev (approved):

eyurtsev:
Thank you!

PR #19942, opened 2024-04-03
Chromium

hinthornw:


libs/community/langchain_community/document_loaders/chromium.py
```diff
@@ -63,6 +62,33 @@ async def ascrape_playwright(self, url: str) -> str:
             await browser.close()
         return results
 
+    def scrape_playwright(self, url: str) -> str:
+        """
+        Sychronously scrape the content of a given URL using Playwright's async API.
```
cbornet, line 67:
```suggestion
        Synchronously scrape the content of a given URL using Playwright's async API.
```

cbornet (approved):

libs/community/langchain_community/document_loaders/chromium.py
```diff
@@ -63,6 +62,33 @@ async def ascrape_playwright(self, url: str) -> str:
             await browser.close()
         return results
 
+    def scrape_playwright(self, url: str) -> str:
```
eyurtsev, line 65: @hinthornw could you document behavior if the url is: 

1) prefixed with `file:///`
2) What happens on redirects?

PR #19954, opened 2024-04-03
langchain: Add a default description to the retriever tool

cbornet:


eyurtsev:
What's the rational? I think this one is maybe OK without a default value so yours have to add a description? If multiple retrievers are used they will likely need to have different descriptions?

cbornet:
I believe most of the time you'll have only one retriever and I felt that this description was a quite good default for that retriever, isn't it ?

cbornet:
I saw that while assembling this tool with a self-query-retriever where the definition of how the docs were retrieved didn't seem important. The important thing is that this tool retrieves docs base on a query.

eyurtsev (approved):

baskaryan:
my sense is in most production use cases it's worth setting a retriever-specific description that explains what types of documents are being retrieved. and i generally prefer not adding defaults that are usually overridden

PR #19967, opened 2024-04-03
Docs: Update custom chat model

eyurtsev:
* Clean up in the existing tutorial
* Add model_name to identifying params
* Add table to summarize messages


eyurtsev:
I have a mistake in the documentation --need to fix prior to landing

PR #19978, opened 2024-04-03
groq: handle streaming tool call case

efriis:


libs/partners/groq/langchain_groq/chat_models.py
```diff
@@ -225,19 +225,16 @@ def _generate(
         messages: List[BaseMessage],
         stop: Optional[List[str]] = None,
         run_manager: Optional[CallbackManagerForLLMRun] = None,
-        stream: Optional[bool] = None,
         **kwargs: Any,
     ) -> ChatResult:
-        should_stream = stream if stream is not None else self.streaming
-        if should_stream:
+        if self.streaming:
```
efriis, line 230: these are because BaseChatModel handles the other case now

PR #19980, opened 2024-04-03
core: mustache prompt templates

nfcampos:


libs/core/pyproject.toml
```diff
@@ -17,6 +17,7 @@ jsonpatch = "^1.33"
 PyYAML = ">=5.3"
 packaging = "^23.2"
 jinja2 = { version = "^3", optional = true }
+chevron = "^0.14.0"
```
eyurtsev, line 20: chevron hasn't been updated in 3 years. I don't think we this should be a required dependency, especially since I doubt most users will use mustache for templates. https://github.com/noahmorrison/chevron?tab=readme-ov-file#chevron-is-python-2-and-3-compatible

libs/core/langchain_core/prompts/string.py
```diff
@@ -85,8 +88,70 @@ def _get_jinja2_variables_from_template(template: str) -> Set[str]:
...
+    while fields:
+        field = fields.pop()
+        current = defs
+        for part in field[:-1]:
+            current = current.setdefault(part, {})
+        current[field[-1]] = {}
+    return _create_model_recursive("PromptInput", defs)
+
+
+def _create_model_recursive(name: str, defs: Defs) -> Type:
```
eyurtsev, line 142: Fun will review this fully tomorrow :)

libs/core/tests/unit_tests/prompts/test_prompt.py
```diff
@@ -38,6 +38,135 @@ def test_prompt_from_template() -> None:
...
+    assert prompt.input_variables == ["foo"]
+    assert prompt.input_schema.schema() == {
+        "title": "PromptInput",
+        "type": "object",
+        "properties": {"foo": {"$ref": "#/definitions/foo"}},
+        "definitions": {
+            "foo": {
+                "title": "foo",
+                "type": "object",
+                "properties": {"bar": {"title": "Bar", "type": "string"}},
```
eyurtsev, line 164: Should be string or array i think? The array is also needed to support inverted sections?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -1113,6 +1123,7 @@ def _create_template_from_message_type(
 
 def _convert_to_message(
     message: MessageLikeRepresentation,
+    template_format: Literal["f-string", "mustache"] = "f-string",
```
eyurtsev, line 1126: doc-string out of date

eyurtsev (commented):
Looks good! 

Two main changes:
1) Could we make optional dependency? at least to start with?
2) Likely an issue with schema inference for sections, as it needs to support arrays?

libs/core/tests/unit_tests/prompts/test_prompt.py
eyurtsev: cc @nfcampos this might still be an issue

PR #20012, opened 2024-04-04
Community : Add audio-parser "faster-whisper" in audio.py

hulitaitai:
faster-whisper is a reimplementation of OpenAI's Whisper model using CTranslate2, which is up to 4 times faster than enai/whisper for the same accuracy while using less memory. The efficiency can be further improved with 8-bit quantization on both CPU and GPU.

It can automatically detect the following 14 languages and transcribe the text into their respective languages: en, zh, fr, de, ja, ko, ru, es, th, it, pt, vi, ar, tr.

The gitbub repository for faster-whisper is :
    https://github.com/SYSTRAN/faster-whisper


hulitaitai:
I am sorry, I have never passed the lint test successfully. I really don't know what is wrong.

faster-whisper is an open-source parser, and it is a low-cost parser which can run with less than 5GB of GPU memory.

libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -319,3 +319,96 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
...
+    ar, tr.
+
+    The gitbub repository for faster-whisper is :
+    https://github.com/SYSTRAN/faster-whisper
+
+    """
+
+    def __init__(
+        self,
+        device: Optional[str] = "0",
```
eyurtsev, line 343:
```suggestion
        *,
        device: Optional[str] = "0",
```

eyurtsev, line 343: device "0" doesn't look like correct value given that it can only end up being cuda or cpu with logic below.

libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -319,3 +319,96 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
...
+    ):
+        try:
+            import torch
+        except ImportError:
+            raise ImportError(
+                "torch package not found, please install it with " "`pip install torch`"
+            )
+
+        # Determine the device to use
+        if device == "cpu":
```
eyurtsev, line 354: logic looks like it's ignoring input from user

libs/community/langchain_community/document_loaders/parsers/audio.py
eyurtsev, line 343: could you document the arguments via an init?

libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -319,3 +319,96 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
...
+    less memory. The efficiency can be further improved with 8-bit quantization on both 
+    CPU and GPU.
+
+    It can automatically detect the following 14 languages and transcribe the text 
+    into their respective languages: en, zh, fr, de, ja, ko, ru, es, th, it, pt, vi, 
+    ar, tr.
+
+    The gitbub repository for faster-whisper is :
+    https://github.com/SYSTRAN/faster-whisper
+
```
eyurtsev, line 338: If possible it'll be great to include a usage Example with the full import using a ..code-block: python

libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -319,3 +319,96 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
...
+                self.model_size = "medium"
+            else:
+                self.model_size = "large-v3"
+        if model_size is not None:
+            self.model_size = model_size
+        
+    def lazy_parse(self, blob: Blob) -> Iterator[Document]:
+        """Lazily parse the blob."""
+
+        import io
```
eyurtsev, line 379: can be in global namespace

libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -319,3 +319,96 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
...
+        try:
+            from faster_whisper import WhisperModel
+        except ImportError:
+            raise ImportError(
+                "faster_whisper package not found, please install it with "
+                "`pip install faster-whisper`"
+            )
+
+        # Audio file from disk
+        audio = AudioSegment.from_file(blob.path)
```
eyurtsev, line 397: Could you check that a .path is set on the blob? Blobs do not have to be associated with a file in the file system as they can live just in memory.


Is it possible to handle using `blob.as_bytes()`.

libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -319,3 +319,96 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
...
+            device=self.device, 
+            compute_type="float16"
+        )
+
+        segments, info = model.transcribe(file_obj, beam_size=5)
+
+        for segment in segments:
+            yield Document(
+                page_content=segment.text,
+                metadata={"source": blob.source},
```
eyurtsev, line 413: {
  'source': blob.source,
  **blob.metadata, # To use existing metadata
}

libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -319,3 +319,96 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
...
+    def __init__(
+        self,
+        device: Optional[str] = "0",
+        model_size: Optional[str] = None,
+    ):
+        try:
+            import torch
+        except ImportError:
+            raise ImportError(
+                "torch package not found, please install it with " "`pip install torch`"
```
eyurtsev, line 350:
```suggestion
                "torch package not found, please install it with `pip install torch`"
```

eyurtsev:
i can help with the linter -- i mostly care about some of the other functional changes 

https://python.langchain.com/docs/contributing/code/#formatting-and-linting

`make format` and `make lint` will help

libs/community/langchain_community/document_loaders/parsers/audio.py
hulitaitai: Yes, here, I presumed I would reassign the definite value later, so I casually placed a value here. 

Additionally, I admit I copied a lot of code and idees from the parsers above in this file. The code I've written runs well on my PC. I will meticulously re-examine various transmission and variables as you reminde, for the sake of overall system. I'll come back to you after I've finished checking.

hulitaitai:
> i can help with the linter -- i mostly care about some of the other functional changes
> 
> https://python.langchain.com/docs/contributing/code/#formatting-and-linting
> 
> `make format` and `make lint` will help

I asked ChatGPT how to use 'make format' and 'make lint'. I will use it before my next Pull Request.

libs/community/langchain_community/document_loaders/parsers/audio.py
hulitaitai: In the blob sent by ffmpeg, there is only the path of the audio file and nothing in its metadata.

Faster-whisper does not transcribe the bytes, it transcribes only mp3 files. So it should ensure that the file sent to it is an mp3 file.

eyurtsev: The parsing is decoupled from the code that loads the blob. To be correct, the parser cannot assume how the blob was generated instead it should assume blob against the blob API (see link below) -- the blob may be associated with metadata and the blob may not exist on file but instead in memory.

Here's the API for the blob

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/document_loaders/blob_loaders.py#L40-L40

Let me know if you have any questions!

So things that need to be done here:

1. Propagate metadata from the blob
2. Make sure that the parser keeps working if the blob was only specified as binary data (using the data attribute) rather than generated from an existing file

eyurtsev (changes requested):

libs/community/langchain_community/document_loaders/parsers/audio.py
eyurtsev: @hulitaitai if you're willing, I can make the needed changes, but I'll need your help to help test it

hulitaitai:
I have tested the changes with the following example, it works.

from pathlib import Path
from langchain_community.document_loaders.blob_loaders import Blob
from langchain_community.document_loaders.parsers.audio import FasterWhisperParser

m4a_path = Path("example.m4a")

with open(m4a_path, "rb") as f:
    m4a_bytes = f.read()

# blob built with binary data
blob = Blob.from_data(data=m4a_bytes)

p = FasterWhisperParser()
docs = p.lazy_parse(blob)
for doc in docs:
    print(doc)

hulitaitai:
I have tested the changes with the following example, it works.

from pathlib import Path
from langchain_community.document_loaders.blob_loaders import Blob
from langchain_community.document_loaders.parsers.audio import FasterWhisperParser

m4a_path = Path("example.m4a")

with open(m4a_path, "rb") as f:
m4a_bytes = f.read()

blob = Blob.from_data(data=m4a_bytes)

p = FasterWhisperParser()
docs = p.lazy_parse(blob)
for doc in docs:
    print(doc)

hulitaitai:
I have tested the changes with with a blob built with "Blob.from_data(data=m4a_bytes)", and it works.

hulitaitai:
I have tested the changes with with a blob built with "Blob.from_data(data=m4a_bytes)", and the parser works.

eyurtsev (approved):

eyurtsev:
Looks great! I'll resolve the linting issue tomorrow.

If you end up working with `blob` again, it has a convenience interface called `blob.as_bytes_io`

```
    @contextlib.contextmanager
    def as_bytes_io(self) -> Generator[Union[BytesIO, BufferedReader], None, None]:
        """Read data as a byte stream."""
        if isinstance(self.data, bytes):
            yield BytesIO(self.data)
        elif self.data is None and self.path:
            with open(str(self.path), "rb") as f:
                yield f
        else:
            raise NotImplementedError(f"Unable to convert blob {self}")
```

PR #20027, opened 2024-04-04
community[patch]: Invoke callback prior to yielding token #16913

UtkarshaGupte:
**Description**: Invoke callback prior to yielding token in stream 
**Issue**: https://github.com/langchain-ai/langchain/issues/16913

Twitter Handle - @GupteUtkarsha


UtkarshaGupte:
@baskaryan - Can you take a look please ?

PR #20035, opened 2024-04-04
core: Implement aformat_prompt and ainvoke in BasePromptTemplate

cbornet:


libs/core/langchain_core/prompts/base.py
```diff
@@ -122,10 +132,29 @@ def invoke(
             run_type="prompt",
         )
 
+    async def ainvoke(
```
cbornet, line 135: @eyurtsev at last 😉

eyurtsev: 💪

eyurtsev (approved):

PR #20036, opened 2024-04-04
core: Implement aformat_messages for _StringImageMessagePromptTemplate

cbornet:


libs/core/langchain_core/prompts/chat.py
```diff
@@ -526,6 +529,34 @@ def format(self, **kwargs: Any) -> BaseMessage:
                 content=content, additional_kwargs=self.additional_kwargs
             )
 
+    async def aformat(self, **kwargs: Any) -> BaseMessage:
```
eyurtsev, line 532: is there a unit test we can hook into for this function to prevent the implementations from drifting?

eyurtsev (approved):

libs/core/langchain_core/prompts/chat.py
cbornet: Yes, this code path is invoked by the async tests in `tests/unit_tests/prompts/test_few_shot.py`

cbornet: But I'll also add async tests in `test_chat.py` for the multipart text.

cbornet: done.
@eyurtsev can you take a look again ?

eyurtsev: Thank you!

PR #18928, opened 2024-03-11
community[patch]: speed up import times in the community package

leo-gan:
This PR speeds up import times in the community package

PR #20050, opened 2024-04-05
community[patch]: Improve import callbacks to make it IDE friendly

eyurtsev:
* declares __all__ as a list of strings (instead of dynamically computing it)
* import type definitions when TYPE_CHECKING is true

### DIFF
--- a/libs/community/tests/unit_tests/callbacks/test_imports.py
+++ b/libs/community/tests/unit_tests/callbacks/test_imports.py
@@ -1,4 +1,4 @@
-from langchain_community.callbacks import __all__
+from langchain_community.callbacks import __all__, _module_lookup
 
 EXPECTED_ALL = [
     "AimCallbackHandler",
@@ -29,4 +29,6 @@ EXPECTED_ALL = [
 
 
 def test_all_imports() -> None:
+    """Test that __all__ is correctly set."""
     assert set(__all__) == set(EXPECTED_ALL)
+    assert set(__all__) == set(_module_lookup.keys())
--- a/libs/community/langchain_community/callbacks/__init__.py
+++ b/libs/community/langchain_community/callbacks/__init__.py
@@ -9,6 +9,34 @@
 import importlib
 from typing import Any
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from langchain_community.callbacks.aim_callback import AimCallbackHandler
+    from langchain_community.callbacks.argilla_callback import ArgillaCallbackHandler
+    from langchain_community.callbacks.arize_callback import ArizeCallbackHandler
+    from langchain_community.callbacks.arthur_callback import ArthurCallbackHandler
+    from langchain_community.callbacks.clearml_callback import ClearMLCallbackHandler
+    from langchain_community.callbacks.comet_ml_callback import CometCallbackHandler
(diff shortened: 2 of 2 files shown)


pierreveron:
@eyurtsev This solves my [issue](https://github.com/langchain-ai/langchain/pull/18928#issuecomment-2037464828) too in VSCode but a new one appeared. Pylint complains about the imports not being exported despite the `__all__ `variable.

image

My guess is that the linter doesn't compute the list from the dict. My fix to that is to put back the original list.
```python
__all__ = [
    "AimCallbackHandler",
    "ArgillaCallbackHandler",
    "ArizeCallbackHandler",
    "PromptLayerCallbackHandler",
    "ArthurCallbackHandler",
    "ClearMLCallbackHandler",
    "CometCallbackHandler",
    "ContextCallbackHandler",
    "HumanApprovalCallbackHandler",
    "InfinoCallbackHandler",
    "MlflowCallbackHandler",
    "LLMonitorCallbackHandler",
    "OpenAICallbackHandler",
    "LLMThoughtLabeler",
    "StreamlitCallbackHandler",
    "WandbCallbackHandler",
    "WhyLabsCallbackHandler",
    "get_openai_callback",
    "wandb_tracing_enabled",
    "FlyteCallbackHandler",
    "SageMakerCallbackHandler",
    "LabelStudioCallbackHandler",
    "TrubricsCallbackHandler",
    "FiddlerCallbackHandler",
]
```

eyurtsev:
@pierreveron got it -- it sounds like we should not compute the __all__ list dynamically and instead rely on unit tests to catch any drift

eyurtsev:
@pierreveron made required changes, we'll merge and please confirm that it all works as expected

pierreveron:
> @pierreveron made required changes, we'll merge and please confirm that it all works as expected

Super cool thanks @eyurtsev. I think you should do the same for all modules in `langchain_community` that were changed by the [PR](https://github.com/langchain-ai/langchain/pull/18928). There can be cascading errors, for example when importing `from langchain.memory import ChatMessageHistory` as `ChatMessageHistory` export is not caught by the linter in `libs/community/langchain_community/chat_message_histories/__init__.py`

libs/community/tests/unit_tests/callbacks/test_imports.py
```diff
@@ -29,4 +29,6 @@
 
 
 def test_all_imports() -> None:
+    """Test that __all__ is correctly set."""
     assert set(__all__) == set(EXPECTED_ALL)
```
leo-gan, line 33: I guess, we don't need the EXPECTED_ALL anymore, no?

baskaryan (approved):

libs/community/tests/unit_tests/callbacks/test_imports.py
eyurtsev: We usually maintain a second list of hard-coded imports in unit tests to catch accidental breaking changes (e.g., removal of an import)

eyurtsev: We could remove in this case and just make sure that set(__all__) == the keys from the module. I don't have a strong preference

leo-gan:
@eyurtsev I'll refactor other `langchain_community` modules whenever applicable.

PR #20059, opened 2024-04-05
community: add ChatOctoAI

sfvaroglu:
This PR adds ChatOctoAI, a chat model integration for OctoAI.


sfvaroglu:
@baskaryan

libs/langchain/langchain/chat_models/octoai.py
baskaryan, line 1: could we remove all the changes to libs/langchain? existing imports from community are only there for backwards compatibility and will be removed in 0.2

sfvaroglu: Done

baskaryan (approved):

PR #20074, opened 2024-04-05
update agents to use tool call messages

ccurme:
```python
from langchain.agents import AgentExecutor, create_tool_calling_agent, tool

    [
        ("system", "You are a helpful assistant"),
        MessagesPlaceholder("chat_history", optional=True),
        ("human", "{input}"),
        MessagesPlaceholder("agent_scratchpad"),
    ]
)
model = ChatAnthropic(model="claude-3-opus-20240229")

@tool
    return input + 2

tools = [magic_function]

agent = create_tool_calling_agent(model, tools, prompt)

agent_executor.invoke({"input": "what is the value of magic_function(3)?"})
```
```
> Entering new AgentExecutor chain...

Invoking: `magic_function` with `{'input': 3}`
responded: [{'text': '<thinking>\nThe user has asked for the value of magic_function applied to the input 3. Looking at the available tools, magic_function is the relevant one to use here, as it takes an integer input and returns an integer output.\n\nThe magic_function has one required parameter:\n- input (integer)\n\nThe user has directly provided the value 3 for the input parameter. Since the required parameter is present, we can proceed with calling the function.\n</thinking>', 'type': 'text'}, {'id': 'toolu_01HsTheJPA5mcipuFDBbJ1CW', 'input': {'input': 3}, 'name': 'magic_function', 'type': 'tool_use'}]

5
Therefore, the value of magic_function(3) is 5.

> Finished chain.
{'input': 'what is the value of magic_function(3)?',
 'output': 'Therefore, the value of magic_function(3) is 5.'}
```


libs/langchain/langchain/agents/output_parsers/openai_tools.py
baskaryan, line 1: should we rename things to not be openai-specific? ToolCallAgentAction, parse_tool_call_message_to_agent_action

baskaryan: could perhaps even move to core

libs/core/langchain_core/load/mapping.py
```diff
@@ -126,12 +126,12 @@
...
     ),
-    ("langchain", "schema", "agent", "OpenAIToolAgentAction"): (
+    ("langchain", "schema", "agent", "ToolAgentAction"): (
         "langchain",
         "agents",
         "output_parsers",
-        "openai_tools",
-        "OpenAIToolAgentAction",
+        "tools",
+        "ToolAgentAction",
```
jacoblee93, line 134: Don't think it matters too much in this particular case since actions aren't really serialized but still technically breaking

libs/langchain/langchain/agents/tools_agent/base.py
```diff
@@ -0,0 +1,96 @@
...
+from langchain.agents.format_scratchpad.tools import (
+    format_to_tool_messages,
+)
+from langchain.agents.output_parsers.tools import ToolsAgentOutputParser
+
+
+def create_tools_agent(
+    llm: BaseLanguageModel, tools: Sequence[BaseTool], prompt: ChatPromptTemplate
+) -> Runnable:
+    """Create an agent that uses tools.
```
jacoblee93, line 17: All agents use tools - how about `tool_calling_agent`? Still bad but at least a little bit more distinguished

libs/langchain/langchain/agents/tools_agent/base.py
```diff
@@ -0,0 +1,96 @@
...
+from langchain_core.tools import BaseTool
+
+from langchain.agents.format_scratchpad.tools import (
+    format_to_tool_messages,
+)
+from langchain.agents.output_parsers.tools import ToolsAgentOutputParser
+
+
+def create_tools_agent(
+    llm: BaseLanguageModel, tools: Sequence[BaseTool], prompt: ChatPromptTemplate
```
jacoblee93, line 15: `BaseLanguageModelLike` would be better to account for `.bind()`

jacoblee93: Oh wait, never mind you wouldn't be able to check for the presence of `bind_tools()` below

libs/core/langchain_core/load/mapping.py
baskaryan: yea probably shouldn't change namespace

baskaryan: or should support both

baskaryan: updated

libs/langchain/langchain/agents/tools_agent/base.py
ccurme: sure, made this change

PR #20075, opened 2024-04-05
community: Fix branch not being considered when using GithubFileLoader

mehdiosa:
- **Description:** Added `ref` query parameter so data is not loaded only from the default branch but any branch passed


mehdiosa:
@baskaryan could you have a look at this fix? Thanks :)

Issue #20053, opened 2024-04-05
Error : Stop sequence key name for {meta or mistral or any other mode} is not supported with AWS Bedrock


david7joy:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

I am trying to use AWS Bedrock models such as Llama / Mistral with Langchain Libraries such as SQLDatabaseToolkit. 

```
model = Bedrock(credentials_profile_name="my-profile",
                model_id="meta.llama2-70b-chat-v1",
                model_kwargs={"temperature": 0.5},
                streaming=True,
                callbacks=[StreamingStdOutCallbackHandler()])
                
db = SQLDatabase.from_uri('database url')
toolkit = SQLDatabaseToolkit(llm=model,db=db)
agent_executor = create_sql_agent(llm=model, toolkit=toolkit,verbose=True, handle_parsing_errors=True)

# Query Module 
result = agent_executor.invoke(prompt)
```

### Error Message and Stack Trace (if applicable)

This errors out with the following.
```
  File "/opt/homebrew/lib/python3.11/site-packages/langchain_community/llms/bedrock.py", line 833, in _call
    for chunk in self._stream(
  File "/opt/homebrew/lib/python3.11/site-packages/langchain_community/llms/bedrock.py", line 613, in _prepare_input_and_invoke_stream
    raise ValueError(
ValueError: Stop sequence key name for meta is not supported.
```

### Description

I have tried the same code with `OpenAI` / `Ollama Mistral/Lamma` as well as google GenAI models and they don't seem to show this error. This seems like something with the way the bedrock library works in Langchain or the bedrock service. 

Is there a workaround I can use to get this to work. 

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023
> Python Version:  3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)]

Package Information
-------------------
> langchain_core: 0.1.40
> langchain: 0.1.14
> langchain_community: 0.0.31
> langsmith: 0.1.40
> langchain_experimental: 0.0.56
[...]

PR #20115, opened 2024-04-07
community: Fix Bedrock Mistral stop sequence request key

jonathancaevans:
- **Description:** Change Bedrock's Mistral stop sequence key mapping to "stop" rather than "stop_sequences" which is the correct key [Bedrock docs link](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral.html) `{
    "prompt": string,
    "max_tokens" : int,
    "stop" : [string],    
    "temperature": float,
    "top_p": float,
    "top_k": int
}`
- **Issue:** #20053 
- **Dependencies:** N/A
- **Twitter handle:** N/a

### DIFF
--- a/libs/community/langchain_community/llms/bedrock.py
+++ b/libs/community/langchain_community/llms/bedrock.py
@@ -338,7 +338,7 @@ class BedrockBase(BaseModel, ABC):
         "amazon": "stopSequences",
         "ai21": "stop_sequences",
         "cohere": "stop_sequences",
-        "mistral": "stop_sequences",
+        "mistral": "stop",
     }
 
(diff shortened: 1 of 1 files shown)


t-mac81:
looking forward to this one being merged!

baskaryan (approved):
thank you @jonathancaevans! if you're interested it'd be great to make this fix in the langchain-aws package as well, which is where the currently maintained version of the LangChain Bedrock llm lives (the version in langchain-community is deprecated) https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py

cc @3coins

PR #20123, opened 2024-04-07
Fix return value type of dumpd

XinyueZ:
The return type of `json.loads` is `Any`.

In fact, the return type of `dumpd` must be based on `json.loads`, so the correction here is understandable.


baskaryan (commented):
hm given how `dumps` works i beliee we know this will always be a dict

PR #20130, opened 2024-04-07
The predict_and_parse is deprecated, instead pass an output parser directly to LLMChain.

xzy103:
The `predict_and_parse` method is deprecated, instead pass an output parser directly to LLMChain.

- [x] **PR title**: "langchain: update chain_extract.py"

image

### DIFF
--- a/libs/langchain/langchain/retrievers/document_compressors/chain_extract.py
+++ b/libs/langchain/langchain/retrievers/document_compressors/chain_extract.py
@@ -64,7 +64,9 @@ class LLMChainExtractor(BaseDocumentCompressor):
         compressed_docs = []
         for doc in documents:
             _input = self.get_input(query, doc)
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/retrievers/document_compressors/chain_extract.py
```diff
@@ -64,7 +64,9 @@ def compress_documents(
         compressed_docs = []
         for doc in documents:
             _input = self.get_input(query, doc)
-            output = self.llm_chain.predict_and_parse(**_input, callbacks=callbacks)
+            output = self.llm_chain.predict(**_input, callbacks=callbacks)
```
baskaryan, line 67: you should be able to replace predict and parse calls with just `invoke(_input, config={"callbacks": callbacks})`

ccurme (approved):

PR #20134, opened 2024-04-08
fix openai embedding avoiding frequently load encoder and support base64 encoding format

rogerdehe:
1. document embedding process should be optimized by avoiding frequent loading of the TikToken encoder or tokenizer. This loading operation consumes significant time, ranging from 100 to 300 milliseconds, which slows down the embedding process.

2. `openai` package support `base64` encoding_format, it should be compatible, for example:

 ```
documents = ["foo bar"]
embedding = OpenAIEmbeddings(model_kwargs={"encoding_format": "base64"})
output = embedding.embed_documents(documents)
```


efriis (commented):
I like the encoder optimization

could you explain the use case of the base64 kwarg? I believe the result should be identical, and because the langchain abstraction for an embedding is a list of floats, it makes more sense to request the list of floats

efriis:
closing for now. feel free to reopen and reference this PR with an answer to ^!

PR #20151, opened 2024-04-08
langchain-pinecone: pinecone vectorstore does not work when async_req is set to False

Nantero1:
When `async_req` is set to False, running `[res.get() for res in async_res]` produces an exception. We do no need to call `get()` at all, as it is already present, when `async_req` is set to False.

Fixes: {TypeError}TypeError("ModelNormal.get() missing 1 required positional argument: 'name'")


efriis:
#19793

Issue #20085, opened 2024-04-05
Standardized model init arg names


baskaryan:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

There's some discrepancies in the init args different models use to set the same params. It'd be a much nicer UX if common params could be set with a common set of init args

Suggesting that if a param in this list is present in a model integration, the below name should either be the actual attribute name or an init arg that can be used to set the attribute:

```python
model: str  # model name
api_key: str  # api key
temperature: float  # temperature sampling
timeout: ...  # request timeout
max_tokens: int  # max tokens
stop: ...  # stop sequences
max_retries: int  # max num retries
base_url: str # base URL to send requests to
```

Importantly we should also use the above init args in the docs

baskaryan:
> Do we need to refactor every llm?

Ideally yes but we can do so incrementally, doesn't need to be all at once

liugddx:
> > Do we need to refactor every llm?
> 
> 
> 
> Ideally yes but we can do so incrementally, doesn't need to be all at once

I see.

sepiatone:
@baskaryan

Do you think it's worth to set the above names as the actual attribute name (and have an alias for the current one)? eg., for OpenAI we have 
```
param openai_api_key: Optional[SecretStr] = None (alias 'api_key')
```

which could be changed to
```
param api_key: Optional[SecretStr] = None (alias 'openai_api_key')
```

This will bring a common look and feel to the API references since sphinx(I think) arranges them in alphabetical order.

sepiatone:
Also perhaps `base_url` -

Cohere
```
param base_url: Optional[str] = None
```
Anthropic
```
param anthropic_api_url: str = 'https://api.anthropic.com'
```
OpenAI
```
param openai_api_base: Optional[str] = None (alias 'base_url')
```

liugddx:
> @baskaryan @liugddx
> 
> Do you think it's worth to set the above names as the actual attribute name (and have an alias for the current one)? eg., for OpenAI we have
> 
> ```
> param openai_api_key: Optional[SecretStr] = None (alias 'api_key')
> ```
> 
> which could be changed to
> 
> ```
> param api_key: Optional[SecretStr] = None (alias 'openai_api_key')
> ```
> 
> This will bring a common look and feel to the API references since sphinx(I think) arranges them in alphabetical order.

We need to be compatible, we try not to modify the original code.

[...]

PR #20161, opened 2024-04-08
anthropic[patch]: standardize init args

baskaryan:
Related to #20085

PR #20166, opened 2024-04-08
community:(baidu_qianfan)[patch]: standardize init args 

liugddx:
Related to https://github.com/langchain-ai/langchain/issues/20085

@baskaryan

### DIFF
--- a/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
+++ b/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
@@ -103,7 +103,7 @@ class QianfanChatEndpoint(BaseChatModel):
     streaming: Optional[bool] = False
     """Whether to stream the results or not."""
 
-    request_timeout: Optional[int] = 60
+    timeout: Optional[int] = 60
     """request timeout for chat http requests"""
 
     top_p: Optional[float] = 0.8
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
```diff
@@ -103,7 +103,7 @@ class QianfanChatEndpoint(BaseChatModel):
     streaming: Optional[bool] = False
     """Whether to stream the results or not."""
 
-    request_timeout: Optional[int] = 60
+    timeout: Optional[int] = 60
```
baskaryan, line 106: need to make sure we do this in a backwards compatible way:
```python
request_timeout: Optional[int] = Field(default=60, alias="timeout")
```
see https://github.com/langchain-ai/langchain/pull/20161 for an example

and we should add a unit test to make sure current and new init args work

liugddx: > need to make sure we do this in a backwards compatible way:
> 
> ```python
> request_timeout: Optional[int] = Field(default=60, alias="timeout")
> ```
> 
> see #20161 for an example
> 
> and we should add a unit test to make sure current and new init args work

I see.Thanks for review.

baskaryan (commented):
need to add `class Config`

liugddx:
@baskaryan ptal

PR #20170, opened 2024-04-08
core,community,partners[minor]: add optional env var OPENAI_BASE_URL

ego:
`OPENAI_BASE_URL`  as an alternative to `OPENAI_API_BASE`

*Why?* 
There is some confusion with this `env var`, so it is would not be bad to support original one
https://github.com/openai/openai-python/blob/main/src/openai/_client.py#L108

and it can also make life a little better for developers =)


efriis:
Hey there! I believe setting that environment variable will actually work already (if you don't set openai_api_base) because the integration package uses the openai sdk.

PR #20182, opened 2024-04-08
standard-tests: a standard unit and integration test set

efriis:
just chat models for now


libs/basetests/langchain_basetests/unit_tests.py
```diff
@@ -0,0 +1,87 @@
...
+    age: int = Field(..., description="The age of the person.")
+
+
+@tool
+def my_adder_tool(a: int, b: int) -> int:
+    """Takes two integers, a and b, and returns their sum."""
+    return a + b
+
+
+class BaseUnitTests(ABC):
```
eyurtsev, line 21: Could we rename this so that it's obvious it's for chat models?

efriis: I was originally thinking of having just one class for everything - but ya makes sense to split by type

PR #20110, opened 2024-04-06
community: import flattening fix 

leo-gan:
See #20050 as a template for this PR.
- As a byproduct: Added 3 missed `test_imports`.
- Added missed `SolarChat` in to __init___.py Added it into test_import ut.
- Added `# type: ignore` to fix linting. It is not clear, why linting errors appear after ^ changes.

@eyurtsev

PR #20184, opened 2024-04-09
Testing imports

eyurtsev:
Should break typing tests in test_outline due to updated type information


eyurtsev:
Verifying reason for typing failures from https://github.com/langchain-ai/langchain/pull/20110

PR #20219, opened 2024-04-09
experimental: return from HuggingGPT task executor task.run() exception

mjschock:
**Description:** Fixes a bug in the HuggingGPT task execution logic here:

      except Exception as e:
          self.status = "failed"
          self.message = str(e)
      self.status = "completed"
      self.save_product()

where a caught exception effectively just sets `self.message` and can then throw an exception if, e.g., `self.product` is not defined.

**Issue:** None that I'm aware of.
**Dependencies:** None
**Twitter handle:** https://twitter.com/michaeljschock


mjschock:
@baskaryan, @efriis, @eyurtsev, @hwchase17 - okay to merge this change?

baskaryan (approved):

PR #20226, opened 2024-04-09
docs: fix JsonOutputToolsParser import in use_cases/tool_use/quickstart

calvinfo:
- **Description:** fixes incorrect import in the tools quickstart
 - **Dependencies:** none
 
Hey team, I was reading through the tool use quickstart and stumbled across some outdated
docs when it comes to creating your own tools and using the JsonOutputToolsParser.

This should fix it by referencing the right package.


calvinfo:
@ccurme hmm, strange. I must've been on an old version, sorry for the noise.

Issue #20093, opened 2024-04-06
DirectoryLoader with use_multithreading=True loads each doc as a list of docs


roman-y:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code
```
from langchain_community.document_loaders import DirectoryLoader, TextLoader    
loader = DirectoryLoader(
        path,
        glob="**/*.txt",
        use_multithreading=True,
        show_progress=True,
        sample_size=1000,
        loader_cls=TextLoader)
docs: List[Document] = loader.load()
```

### Description

Expecting `List[Document]` but getting back `List[List[Document]]` which cannot be directly fed into `Chroma.from_documents` for example

### System Info

langchain==0.1.14
langchain-community==0.0.31
langchain-core==0.1.38
langchain-openai==0.1.1
langchain-text-splitters==0.0.1
langchainhub==0.1.14

Windows

Python 3.9.13

PR #20230, opened 2024-04-09
community: fixed multithreading returning List[List[Documents]] instead of List[Documents]

chip-davis:
Description: When multithreading is set to True and using the DirectoryLoader, there was a bug that caused the return type to be a double nested list. This resulted in other places upstream not being able to utilize the from_documents method as it was no longer a `List[Documents]` it was a `List[List[Documents]]`. The change made was to just loop through the `future.result()` and yield every item.
Issue: #20093
Dependencies: N/A
Twitter handle: N/A

### DIFF
--- a/libs/community/langchain_community/document_loaders/directory.py
+++ b/libs/community/langchain_community/document_loaders/directory.py
@@ -174,7 +174,8 @@ class DirectoryLoader(BaseLoader):
                         )
                     )
                 for future in concurrent.futures.as_completed(futures):
-                    yield future.result()
+                    for item in future.result():
+                        yield item
         else:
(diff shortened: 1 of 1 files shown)


eyurtsev:
@chip-davis would you be willing to add a unit test?

eyurtsev (approved):

eyurtsev:
Merging in @chip-davis if you're willing to follow up with a unit test to catch this would be great!

chip-davis:
> Merging in @chip-davis if you're willing to follow up with a unit test to catch this would be great!

Sure thing! I'll add one tomorrow😀

PR #20246, opened 2024-04-09
community: add Cassandra Database Toolkit

pmcfadin:
**Description**: ToolKit and Tools for accessing data in a Cassandra Database primarily for Agent integration. Initially, this includes the following tools:
 - `cassandra_db_schema` Gathers all schema information for the connected database or a specific schema. Critical for the agent when determining actions.
 - `cassandra_db_select_table_data` Selects data from a specific keyspace and table. The agent can pass paramaters for a predicate and limits on the number of returned records.
 - `cassandra_db_query` Expiriemental alternative to `cassandra_db_select_table_data` which takes a query string completely formed by the agent instead of parameters. May be removed in future versions. 

Includes unit test and two notebooks to demonstrate usage. 

**Dependencies**: cassio
**Twitter handle**: @PatrickMcFadin


libs/community/langchain_community/tools/cassandra_database/prompt.py
```diff
@@ -0,0 +1,53 @@
+"""Tools for interacting with an Apache Cassandra database."""
```
cbornet, line 1: I don't see these prompts used somewhere. What are they for ?

libs/community/langchain_community/utilities/cassandra_database.py
```diff
@@ -0,0 +1,705 @@
...
+        self,
+        query: str,
+        fetch: str = "all",
+        include_columns: bool = False,
+        **kwargs: Any,
+    ) -> Union[str, Sequence[Dict[str, Any]], ResultSet]:
+        """Execute a CQL query and return the results or an error message."""
+        try:
+            return self.run(query, fetch, include_columns, **kwargs)
+        except Exception as e:
```
cbornet, line 70: Maybe this exception catching logic could be put in a utility method in the tools themselves ?
It seems the "no throw" is not a behavior directly related to the DB management.
So in the tool you would do something like ` no_throw(self.db.run(...))`

libs/community/langchain_community/utilities/cassandra_database.py
```diff
@@ -0,0 +1,705 @@
...
+            return s
+
+        # Try to init and return cassio session
+        if cassio_init_kwargs:
+            if isinstance(cassio_init_kwargs, dict):
+                cassio.init(**cassio_init_kwargs)
+                s = cassio.config.check_resolve_session()
+                return s
+            else:
+                raise ValueError("cassio_init_kwargs must be a keyword dictionary")
```
cbornet, line 460: Isn't this already mandated by the type hints ?

libs/community/langchain_community/utilities/cassandra_database.py
```diff
@@ -0,0 +1,705 @@
...
+
+        if self._indexes:
+            output += "- Indexes\n"
+            for name, kind, options in self._indexes:
+                output += f"  - {name} : kind={kind}, options={options}\n"
+
+        return output
+
+    @property
+    def table_name(self) -> str:
```
cbornet, line 590: NIT: some of this boilerplate code could probably be removed by using pydantic.

libs/community/langchain_community/utilities/cassandra_database.py
```diff
@@ -0,0 +1,705 @@
+"""Apache Cassandra database wrapper."""
+from __future__ import annotations
+
+import re
+import traceback
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
+
+import cassio.config
```
cbornet, line 8: It seems the dependency to cassio could be made optional when users provide the session by their own means.

cbornet: Also the cassio should be imported dynamically as it's not required in the poetry deps.

libs/community/langchain_community/tools/cassandra_database/prompt.py
```diff
@@ -0,0 +1,53 @@
...
+
+QUERY_CHECKER = """
```
cbornet, line 3: I don't see this prompt being used somewhere. What is it for ?

cbornet (approved):
LGTM, great work !

cbornet:
Hi folks. Could this be reviewed ?
This is a very nice feature for Cassandra users.

baskaryan (approved):

PR #20256, opened 2024-04-10
docs: `providers` updates 1

leo-gan:
- Proviers pages: added missed integrations; fixed format
- `mistralai` converted from notebook to .mdx format


sepiatone:
@leo-gan @hwchase17 @efriis 
Why do we have the folder `platforms` at all? why not move all providers to the folder `providers`? I was [thinking](https://github.com/langchain-ai/langchain/discussions/19795) of doing this, but wanted to check if there was some Chesterton's Fence that I was coming up against.

Of course, I don't think it will just be a move + delete (I see the effort for #19881).

leo-gan:
@sepiatone It is a classification for a large group. "Platforms" are the "providers with many important integrations". This grouping is intuitive, I hope. Of course, it is not ideal, but Pros >> Cons, for now.

sepiatone:
> @sepiatone It is a classification for a large group. "Platforms" are the "providers with many 

Thanks for the explanation!

docs/docs/integrations/platforms/anthropic.mdx
```diff
@@ -1,43 +1,44 @@
...
+from langchain_anthropic import AnthropicLLM
+
+model = AnthropicLLM(model='claude-2.1')
 ```
 
-Read more in the [ChatAnthropic documentation](/docs/integrations/chat/anthropic).
+## Chat Models
 
-## [Legacy] `AnthropicLLM`
+### ChatAnthropic
```
eyurtsev, line 34: The chat model shold probably be surfaced prior to the legacy documentation?

eyurtsev (approved):

docs/docs/integrations/platforms/anthropic.mdx
leo-gan: @eyurtsev I've added a note that ChatAnthropic should be used instead of the deprecated AnthropicLLM.
Do you mean, that we need to add `### ChatAnthropic` section under the `## LLMs` section by copy-past it from `## Chat models` section?

leo-gan:
@eyurtsev Could it be merged?

docs/docs/integrations/providers/mistralai.mdx
efriis, line 1: why is this beneficial? If anything, might be nice to move more of the mdx ones to notebooks so people can try easier!

Will also make it testable in the future

leo-gan: We keep only the "Provider card" in the `integrations/providers` and `inegrations/platforms`. The purpose of it to present a company and a list of the provider integrations. We keep the real code examples in other subfolders of the `integrations`. Hence, most of the docs there are .mdx.
The first-timers go to `provider` page then walk to the example pages from there. 
For example [here](https://python.langchain.com/docs/integrations/platforms/huggingface/) we reader uses as the "company" hub. That is the whole idea of the `integrations/providers` and `inegrations/platforms`.
The only code is these references
`from langchain_community.chat_models.huggingface import ChatHuggingFace`

leo-gan:
Thanks!

PR #20272, opened 2024-04-10
openai[patch]: use tool_calls in request

baskaryan:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -172,11 +172,25 @@ def _convert_message_to_dict(message: BaseMessage) -> dict:
             # If function call only, content is None not empty string
             if message_dict["content"] == "":
                 message_dict["content"] = None
-        if "tool_calls" in message.additional_kwargs:
+        if message.tool_calls:
+            message_dict["tool_calls"] = [
+                {
+                    "type": "function",
```
baskaryan, line 178: check if "index" is required

eyurtsev:
This makes sense -- I think it'll make it possible to support fallback behavior with tool calling models

PR #20281, opened 2024-04-10
community[test]: directory loader multithreaded

chip-davis:
This is a unit test for #20230 which was a fix for using multithreaded mode with directory loader @eyurtsev


chip-davis:
Pinging @eyurtsev let me know if anything needs to be changed!

baskaryan (approved):

PR #20291, opened 2024-04-10
core[patch]: include tool_calls in ai msg chunk serialization

baskaryan:


libs/core/langchain_core/messages/ai.py
```diff
@@ -88,6 +96,14 @@ def get_lc_namespace(cls) -> List[str]:
         """Get the namespace of the langchain object."""
         return ["langchain", "schema", "messages"]
 
+    @property
+    def lc_attributes(self) -> Dict:
+        """Attrs to be serialized even if they are derived from other init args."""
+        return {
+            "tool_calls": self.tool_calls,
```
ccurme, line 103: why are `tool_call_chunks` serialized properly without being in here?

ccurme (approved):

libs/core/langchain_core/messages/ai.py
baskaryan: default behavior is to serialize (only) attributes that are passed in as init args

nfcampos (approved):

PR #20294, opened 2024-04-10
core[patch]: fix ChatGeneration.text with content blocks

baskaryan:


eyurtsev (approved):

libs/core/tests/unit_tests/outputs/test_chat_generation.py
```diff
@@ -0,0 +1,36 @@
+from langchain_core.messages import AIMessage
+from langchain_core.outputs import ChatGeneration
+
+
+def test_msg_with_text() -> None:
```
eyurtsev, line 5: pytest parameterize is better here

eyurtsev (approved):

PR #20299, opened 2024-04-10
mistral: add IDs to tool calls

ccurme:
Mistral gives us one ID per response, no individual IDs for tool calls.

```python
from langchain.agents import AgentExecutor, create_tool_calling_agent, tool
from langchain_mistralai import ChatMistralAI


    [
        ("system", "You are a helpful assistant"),
        ("human", "{input}"),
        MessagesPlaceholder("agent_scratchpad"),
    ]
)
model = ChatMistralAI(model="mistral-large-latest", temperature=0)

@tool
    return input + 2

tools = [magic_function]

agent = create_tool_calling_agent(model, tools, prompt)

agent_executor.invoke({"input": "what is the value of magic_function(3)?"})
```

### DIFF
--- a/libs/core/langchain_core/output_parsers/openai_tools.py
+++ b/libs/core/langchain_core/output_parsers/openai_tools.py
@@ -44,7 +44,7 @@ def parse_tool_call(
         "args": function_args or {},
     }
     if return_id:
-        parsed["id"] = raw_tool_call["id"]
+        parsed["id"] = raw_tool_call.get("id")
     return parsed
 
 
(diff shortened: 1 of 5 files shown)


libs/core/langchain_core/output_parsers/openai_tools.py
```diff
@@ -44,7 +44,7 @@ def parse_tool_call(
         "args": function_args or {},
     }
     if return_id:
-        parsed["id"] = raw_tool_call["id"]
+        parsed["id"] = raw_tool_call.get("id")
```
ccurme, line 47: this is breaking if one expects a KeyError when using `JsonOutputToolsParser(return_id=True)` and there are no IDs.

eyurtsev (approved):

PR #20303, opened 2024-04-10
langchain[patch]: agents check prompt partial vars

baskaryan:


hinthornw (commented):
Any tests we should add to catch this?

PR #20307, opened 2024-04-11
core[patch]: Fix JsonOutputParser returns empty json when text ends with "\".

### LINKED ISSUE #20204: JsonOutputParser returns empty json when text ends with "\"
### Checked other resources

### Example Code

Code

```
from langchain.schema import Generation
from langchain_core.output_parsers import JsonOutputParser

print(JsonOutputParser().parse_result([Generation(text='{"key":"value\\')]))
```

Actual result:

```{}```

Expected result: not sure

_No response_

### Description

This problem occurred when I was streaming a JSON response containing new lines "\n". The chunk stopped right after the backslash, leading to an empty json after the parser.

### System Info

langchain==0.1.14
langchain-core==0.1.40

plaform: linux

Python 3.11.0rc1

### PR DESCRIPTION
**Pull request by liugddx**
- **Issue:** Close #20204
- @baskaryan, @efriis, @eyurtsev, @hwchase17.

### DIFF
--- a/libs/core/tests/unit_tests/output_parsers/test_json.py
+++ b/libs/core/tests/unit_tests/output_parsers/test_json.py
@@ -235,13 +235,13 @@ TEST_CASES_PARTIAL = [
     ('{"foo": "bar", "bar": "foo', '{"foo": "bar", "bar": "foo"}'),
     ('{"foo": "bar", "bar": "foo}', '{"foo": "bar", "bar": "foo}"}'),
     ('{"foo": "bar", "bar": "foo[', '{"foo": "bar", "bar": "foo["}'),
-    ('{"foo": "bar", "bar": "foo\\"', '''{"foo": "bar", "bar": "foo\\\\"}'''),
+    ('{"foo": "bar", "bar": "foo\\"', """{"foo": "bar", "bar": "foo\\\\"}"""),
     ('{"foo": "bar", "bar":', '{"foo": "bar"}'),
     ('{"foo": "bar", "bar"', '{"foo": "bar"}'),
     ('{"foo": "bar", ', '{"foo": "bar"}'),
-    ('{"key":"val\\ue', '''{"key": "val\\\\ue"}'''),
-    ('{"key":"value\\', '''{"key": "value\\\\"}'''),
-    ('{"key":"\\value', '''{"key": "\\\\value"}'''),
+    ('{"key":"val\\ue', """{"key": "val\\\\ue"}"""),
+    ('{"key":"value\\', """{"key": "value\\\\"}"""),
+    ('{"key":"\\value', """{"key": "\\\\value"}"""),
 ]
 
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by liugddx** [comment]
@eyurtsev This is a bug, please take a look, thanks.

File: libs/core/tests/unit_tests/output_parsers/test_json.py [hunk 1]
```diff
@@ -235,10 +235,12 @@ def test_parse_json_with_python_dict() -> None:
     ('{"foo": "bar", "bar": "foo', '{"foo": "bar", "bar": "foo"}'),
     ('{"foo": "bar", "bar": "foo}', '{"foo": "bar", "bar": "foo}"}'),
     ('{"foo": "bar", "bar": "foo[', '{"foo": "bar", "bar": "foo["}'),
-    ('{"foo": "bar", "bar": "foo\\"', '{"foo": "bar", "bar": "foo\\""}'),
+    ('{"foo": "bar", "bar": "foo\\"', """{"foo": "bar", "bar": "foo\\\\"}"""),
```
baskaryan, line 238: can we add a test case instead of changing an existing one

liugddx: > can we add a test case instead of changing an existing one

I have to modify this test case to make it pass. In my opinion the user should need "\\" instead of ignoring it.

image

The following is before modification, it ignores "\\"


image

liugddx:
@eyurtsev PTAL

libs/core/tests/unit_tests/output_parsers/test_json.py
```diff
@@ -235,10 +235,12 @@ def test_parse_json_with_python_dict() -> None:
     ('{"foo": "bar", "bar": "foo', '{"foo": "bar", "bar": "foo"}'),
     ('{"foo": "bar", "bar": "foo}', '{"foo": "bar", "bar": "foo}"}'),
```
eyurtsev, line 236: The original test was correct, the updated test modifies the output incorrectly.

You can test it with this version which is a bit easier to understand:

```
def test_streaming_variant() -> None:
    obj = {
        "a": "b\\",
        "c": "d",
    }

    parts = json.dumps(obj)
    sub_strings = [parts[:idx] for idx in range(1, len(parts))]
    results = [parse_partial_json(sub_string) for sub_string in sub_strings]

    assert results == [
        {},
        {},
        {},
        {},
        {},
        {},
        {"a": ""},
        {"a": "b"},
        {"a": "b\\"},
        # Output below is wrong, there should only be a single `\`
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\"},
        {"a": "b\\\\", "c": ""},
        {"a": "b\\\\", "c": "d"},
        {"a": "b\\\\", "c": "d"},
    ]
```

eyurtsev (changes requested):

eyurtsev:
Restore the original test, and then create a new test based on the streaming version and make sure that the JSON that's being streamed doesn't get mutated with additional `\`

hwchase17:
closing due to inactivity

PR #20360, opened 2024-04-12
core: bind_tools interface on basechatmodel

efriis:


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -896,6 +905,18 @@ def dict(self, **kwargs: Any) -> Dict:
...
         return starter_dict
 
+    @beta()
+    def bind_tools(
+        self,
+        tools: Sequence[Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool]],
+        *,
+        tool_choice: Optional[
+            Union[Dict, str, Literal["auto", "any", "none"], bool]
+        ] = None,
```
efriis, line 915: also game to leave this explicit kwarg off. But feels like we'll start recommending it for models that support it and wouldn't want re-use in another way.

baskaryan: don't think tool_choice should be on base implementation, anthropic doesn't support

efriis: But it won't fail them because it's an optional kwarg.

What it will fail is someone defining `bind_tools(..., tool_choice: int)`

Thoughts on keeping?

baskaryan: imo weird to have it as part of signature on implementations that don't support it. also feels like a still pretty beta feature for most providers. also seems easier to add later if things change than to try to remove if things change. but don't feel super strongly here

efriis: removed. Can add later easily.

PR #20361, opened 2024-04-12
core: use qualname in beta message

efriis:


libs/core/langchain_core/_api/beta_decorator.py
```diff
@@ -189,10 +190,10 @@ def finalize(wrapper: Callable[..., Any], new_doc: str) -> Any:
                 )
 
         else:
+            _name = _name or obj.__qualname__
             if not _obj_type:
-                _obj_type = "function"
+                _obj_type = "function" if "." not in _name else "method"
```
efriis, line 195: edge case this won't work on: a function defined within another function within a test

baskaryan:
lgtm, would add to deprecated

efriis:
Was added to deprecated in #20578 also

PR #20378, opened 2024-04-12
community: Add arguments to whisper parser

roiperlman:
**Description:** Added a few additional arguments to the whisper parser, which can be consumed by the underlying API.
The prompt is especially important to fine-tune transcriptions.


libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -31,12 +31,32 @@ def __init__(
         *,
         chunk_duration_threshold: float = 0.1,
         base_url: Optional[str] = None,
+        language: Union[str, None] = None,
```
baskaryan, line 34: can we document all of these in docstring above

hwchase17 (approved):

PR #20392, opened 2024-04-12
core: adding Sentence Window Retrieval to VectorStoreRetriever

rsk2327:

Description : 
The PR implements the Sentence Window Retrieval method as one of the search types for VectorStoreRetriever.

Issue : No issues. Is a feature addition

Dependencies : No dependencies 

Additional guidelines:


rsk2327:
@eyurtsev @baskaryan Would be great to get your inputs on how to scale this up to other databases as its only done for FAISS right. Should I include shell implementations for the other databases that raise NotImplementedError for now?

Also, could you help with debugging the linting errors here? 

`langchain_community/vectorstores/faiss.py:823: error: Incompatible types in assignment (expression has type "ndarray[Any, dtype[floating[_32Bit]]]", variable has type "List[float]")  [assignment]`
This is the exact same line of code at line 196 as well but that doesnt throw an error

`langchain_community/vectorstores/faiss.py:839: error: Item "str" of "Union[str, Document]" has no attribute "page_content"  [union-attr]`
Not sure why this throws an error either. page_content is an attribute of Document

rsk2327:
@eyurtsev @hwchase17 @baskaryan @efriis Could I get a review on this?

Undertone0809 (approved):

Undertone0809:
need to add test to assert it.

rsk2327:
> need to add test to assert it.

@Undertone0809 Are you referring to general unit tests for this or for asserting the FAISS database requirement for the current implementation? For the latter, I do implement a check in the validate_search_type root_validator for the VectorStoreRetriever class.

baskaryan (commented):
could we add some docs explaining how this works / why it's useful? can add to /docs/docs/integrations/vectorstores/faiss.ipynb

rsk2327:
> could we add some docs explaining how this works / why it's useful? can add to /docs/docs/integrations/vectorstores/faiss.ipynb

Added detailed description of the method and how its useful, along with examples on how to run it

rsk2327:
@baskaryan @eyurtsev @efriis Can i get a review on this?

rsk2327:
@baskaryan  @eyurtsev @efriis @hwchase17

eyurtsev (changes requested):
Hi @rsk2327, thank you for the contribution!

We're very cautious about adding additional functionality to the vectorstores interfaces. The interface is already very large and any additional methods just complicate it and make it more confusing.

In this case, the technique can be implemented generally without touching the vectorstore interface by combining a document store with a vectorstore.

eyurtsev:
You could try to implement using: https://python.langchain.com/docs/modules/data_connection/retrievers/parent_document_retriever/

rsk2327:
@eyurtsev Im aware that this introduces a change to a key component and thats why I was extra cautious of the implementation, making sure to use existing functions and only add methods in the vector database's code and not in the VectorStore's implementation. 

As you could see, most of the implementation is in the FAISS.py file and not so much in VectorStore. Im happy to work with you to ensure that the VectorStore implementation isnt affected by this change. I feel that this is the right place to add this functionality because it is alternative to Similarity Search and MMR, so it should ideally be included as an option at the same level. 

But if you still have concerns then I can implement it separately as a separate Retriever class instead.

eyurtsev:
Hi @rsk2327, we'll likely going to move in the opposite direction in the future and remove methods like `mmr` from the vectorstores. We still haven't speced out to do this, so I don't have good advice for that right now, but we want to move in direction of reducing vectorstore interface.

What's needed for the implementation of this search technique is access to a document store + a way to implement nearest neighbor search by embedding. I believe this can be done using a parent retriever, and it immediately applies across all vectorstore implementations.

rsk2327:
@eyurtsev Okay, if thats the case then it makes more sense to implement this as a separate retriever class. I will set that up instead. Thanks for reviewing this.

PR #20396, opened 2024-04-12
community[patch]: add test to catch bad imports

eyurtsev:
Add test to catch bad imports


eyurtsev:
cc @leo-gan code still in the works, but I think it's catching some issues with community imports, we'll probably want to write a generic unit test to make sure that there's no drift

PR #20399, opened 2024-04-12
docs: `integrations/providers` updates

leo-gan:
Added missed provider pages. Updated descriptions; added links; update to the consistent format


efriis (commented):
Could you revert the titles and sidebar labels? Those are helpful for search, and it makes the page title the class that's actually documented (e.g. it documents `ChatFireworks`).

Main use is so when people search "ChatFireworks" it goes to the right page.

leo-gan:
> Could you revert the titles and sidebar labels? Those are helpful for search, and it makes the page title the class that's actually documented (e.g. it documents `ChatFireworks`).
> 
> Main use is so when people search "ChatFireworks" it goes to the right page.

I see. 
When I need to search for a class, I go to the API Reference. Integrations pages are more about trade names and technology names. The integration pages are used by the first-time readers not by developers working on the LC project. They need to choose the right integration/technology to a new project, hence the class names are irrelevant.
For example, the [`Google Cloud Vertex AI`](https://python.langchain.com/docs/integrations/chat/google_vertex_ai_palm/) page is not about the class name, even not about the LLM model name, it is about the registered technology name. 
OK We have the `Google Cloud Vertex AI` in the navbar and `ChatVertexAI` as a title. 
Now we have an additional problem, what the heck is the `ChatVaetexAI`? I'm the first-time reader (because the first-time reader is the main reader of the Integration pages), and I'm confused. Not good.
To summarize:
1. Titles must be equal menu items. The sidebar labels that is not equal to the Titles are confusing.
2. Titles that are not equal to menu items, they require the massive manual page update to add sidebar labels. Then it would require constant effort to keep it up-to-date. Do you have the resources and plans to do this damage?

efriis:
Doesn't need to be perfect. The problem of "I saw `ChatVertexAI` somewhere and wanted to see the docs, but when I searched for it I got the wrong page" is one I would rather not bring back.

Alternative is maintaining keywords on each page, but that is less visible than when someone by mistake forgets to add the sidebar label.

In general, people seem very good at mimicking nearby pages when putting up new chat integrations. Happy to review a PR fixing any that missed it.

leo-gan:
@efriis  The problem of "I saw ChatVertexAI somewhere and wanted to see the docs, but when I searched for it I got the wrong page" is not solved in principal by using the integration page title as the class name. We are trying to solve one problem by creating a confusion.
I can always search any class in docs by using the `Search` field in the top-menu or the `Search` field in the API Ref top-menu.
Almost any class.
I've just tried to search `ConversationBufferMemory`, `BaseStepContainer` (presented only in the API Ref), `TextSplitter` (over-represented in docs and in API Ref). API Ref Search is always perfect for this question. Doc Search could be clogged with irrelevant answers or be empty.
Here is a design proposal: add the API Ref Search results into the Doc Search result. This solution is fully automated and does not require any changes in the existing Docs.

docs/docs/integrations/providers/fireworks.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+  ```bash
+  pip install langchain-fireworks
+  ```
+
+- Get a Fireworks API key by following [instructions here](https://readme.fireworks.ai/docs/quickstart).
+- Authenticate by setting the `FIREWORKS_API_KEY` environment variable.
+
+
+## LlMs
```
efriis, line 21:
```suggestion
## LLMs
```

efriis (commented):
Same comment as a past PR if I recall correctly. Please don't rename the heading from `ChatFireworks` to `Fireworks` because it interferes with search results.

efriis:
Closing because searches with classnames still make up a large portion of the queries

Adding the api docs to docusaurus in python.langchain.com would be a bigger change that I'd support but is tricky!

PR #20402, opened 2024-04-13
core: forward config params to default

efriis:
nuno's fault not mine


libs/core/tests/unit_tests/runnables/test_configurable.py
```diff
@@ -83,3 +104,86 @@ def test_field_alias_set_configurable() -> None:
...
+    with pytest.raises(AttributeError):
+        configurable_runnable.my_other_custom_function()  # type: ignore[attr-defined]
+    with pytest.raises(AttributeError):
+        configurable_runnable.my_other_custom_function_w_config(  # type: ignore[attr-defined]
+            {"configurable": {"my_other_property": "b"}}
+        )
+    with pytest.raises(AttributeError):
+        configurable_runnable.with_config(
+            configurable={"my_other_property": "c", "which": "other"}
+        ).my_other_custom_function()  # type: ignore[attr-defined]
```
baskaryan, line 189: should test test that my_other_custom_function does work if you configure to use alternative? would that work? (either way a test would help make expected behavior explicit)

nfcampos: that's what this test is, see `which: other`. and no it doesn't work

baskaryan (approved):

PR #20404, opened 2024-04-13
docs: update tool use case

baskaryan:


docs/docs/use_cases/tool_use/agents.ipynb
```diff
@@ -17,13 +17,13 @@
     "jp-MarkdownHeadingCollapsed": true
    },
    "source": [
-    "## Agents\n",
+    "## Repeated tool use with agents\n",
```
ccurme, line 20: Maybe a nit but repeated tool use is only one element of this right? There is also the decision of what tool to use, or whether to use any tools at all. I don't have a great alternative title, maybe something like "Agents and model-driven tool use"

docs/docs/use_cases/tool_use/agents.ipynb
```diff
@@ -17,13 +17,13 @@
...
     "\n",
     "LangChain comes with a number of built-in agents that are optimized for different use cases. Read about all the [agent types here](/docs/modules/agents/agent_types/).\n",
     "\n",
-    "As an example, let's try out the OpenAI tools agent, which makes use of the new OpenAI tool-calling API (this is only available in the latest OpenAI models, and differs from function-calling in that the model can return multiple function invocations at once).\n",
+    "We'll use the [tool calling agent](/docs/modules/agents/agent_types/tool_calling/), which is generally the most reliable kind and the recommended one for most use cases. \"Tool calling\" in this case refers to a specific type of model API that allows for explicitly passing tool definitions to models and getting explicit tool invocations out. For more on tool calling models see [this guide](/docs/modules/model_io/chat/function_calling/)\n",
```
ccurme, line 26:
```suggestion
    "We'll use the [tool calling agent](/docs/modules/agents/agent_types/tool_calling/), which is generally the most reliable kind and the recommended one for most use cases. \"Tool calling\" in this case refers to a specific type of model API that allows for explicitly passing tool definitions to models and getting explicit tool invocations out. For more on tool calling models see [this guide](/docs/modules/model_io/chat/function_calling/).\n",
```

docs/docs/use_cases/tool_use/agents.ipynb
```diff
@@ -173,27 +170,45 @@
...
    "source": [
-    "## Create agent"
+    "## Create agent\n",
+    "\n",
+    "We'll need to use a model with tool calling capabilities. You can see which models support tool calling [here](/docs/integrations/chat/).\n",
+    "\n",
+    "```{=mdx}\n",
+    "import ChatModelTabs from \"@theme/ChatModelTabs\";\n",
+    "\n",
+    "<ChatModelTabs customVarName=\"llm\"/>\n",
```
ccurme, line 180: we might `hideCohere` pending https://github.com/langchain-ai/langchain-cohere/pull/15

ccurme (approved):
Thank you!

docs/docs/use_cases/tool_use/agents.ipynb
baskaryan: true, but was thinking that's handled by the "Choosing between multiple tools" section, what makes agents diff from just a routing chain is it chooses how many steps to take

PR #20434, opened 2024-04-14
core: improve comma separated list output parser to handle non-space separated list

anish749:
- **Description:** Changes `lanchain_core.output_parsers.CommaSeparatedListOutputParser` to handle `,` as a delimiter alongside the previous implementation which used `, ` as delimiter.
  - **Issue:** Started noticing that some results returned by LLMs were not getting parsed correctly when the output contained `,` instead of `, `.
  - **Dependencies:** No
  - **Twitter handle:** not active on twitter.


anish749:
hey @efriis, not sure if you are the right person, do you think this change is a good idea, or would you know who would be the right person for a review?

anish749:
ping @efriis @baskaryan

baskaryan (approved):

Issue #20445, opened 2024-04-14
PGVector deprecated create_filter_clause uses incorrect method. Can't filter by OR or AND conditions


MartinGotelli:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
import os
os.environ["TEST_PGVECTOR_PORT"] = "5432"
os.environ["TEST_PGVECTOR_DATABASE"] = "langchain_tests"
os.environ["TEST_PGVECTOR_USER"] = "postgres"
os.environ["TEST_PGVECTOR_PASSWORD"] = "postgres"


import os
from typing import List

from langchain_core.embeddings import Embeddings
from langchain_community.vectorstores.pgvector import PGVector


class FakeEmbeddings(Embeddings):
    def embed_documents(self, texts: List[str]) -> List[List[float]]:
        return [[float(1.0)] * 9 + [float(i)] for i in range(len(texts))]
    def embed_query(self, text: str) -> List[float]:
        return [float(1.0)] * 9 + [float(0.0)]

CONNECTION_STRING = PGVector.connection_string_from_db_params(
    driver=os.environ.get("TEST_PGVECTOR_DRIVER", "psycopg2"),
    host=os.environ.get("TEST_PGVECTOR_HOST", "localhost"),
    port=int(os.environ.get("TEST_PGVECTOR_PORT", "6024")),
    database=os.environ.get("TEST_PGVECTOR_DATABASE", "langchain"),
    user=os.environ.get("TEST_PGVECTOR_USER", "langchain"),
    password=os.environ.get("TEST_PGVECTOR_PASSWORD", "langchain"),
)

texts = ["foo", "bar", "baz"]
metadatas = [{"page": str(i)} for i in range(len(texts))]
docsearch = PGVector.from_texts(
    texts=texts,
    collection_name="test_collection_filter",
    embedding=FakeEmbeddings(),
    metadatas=metadatas,
    connection_string=CONNECTION_STRING,
    pre_delete_collection=True,
)

output = docsearch.similarity_search_with_score(
    "foo", k=3, filter={"page": {"OR": [{"EQ": "0"}, {"EQ": "2"}]}}
)

# Alternatively:
[...]

PR #20446, opened 2024-04-14
community: Fix pgvector deprecated filter clause usage with OR and AND conditions

MartinGotelli:
**Description**: Support filter by OR and AND for deprecated PGVector version
**Issue**: #20445 
**Dependencies**: N/A
**Twitter** handle: @martinferenaz


eyurtsev (approved):

eyurtsev:
Thanks! Going to merge the fix, though I recommend to folks to move to `langchain_postgres` which has a more efficient implementation (that has other issues resolved with it as well)

PR #20460, opened 2024-04-15
Community patch clickhouse make it possible to not specify index

mneedham:
Vector indexes in ClickHouse are experimental at the moment and can sometimes break/change behaviour. So this PR makes it possible to say that you don't want to specify an index type.

Any queries against the embedding column will be brute force/linear scan, but that gives reasonable performance for small-medium dataset sizes.

Previous PRs:

https://github.com/langchain-ai/langchain/pull/19527
https://github.com/langchain-ai/langchain/pull/18381


mneedham:
@efriis  can you get the build to run on this version of the PR?

mneedham:
@efriis did you get a chance to look at this?

mneedham:
@efriis would you be able to kick off the build so that we can see if these changes work?

mneedham:
@efriis any chance you can take a look at this?

efriis (approved):
looks great! Thanks for fixing the comment from #19527

PR #20468, opened 2024-04-15
Add semantic info to metadata, classified by pebblo-server.

rahul-trip:
Description: Add support for Semantic topics and entities. Classification done by pebblo-server is not used to enhance metadata of  Documents loaded by document loaders.
Dependencies: None
Documentation: Updated.

### DIFF
--- a/libs/community/langchain_community/document_loaders/pebblo.py
+++ b/libs/community/langchain_community/document_loaders/pebblo.py
@@ -5,9 +5,9 @@ import logging
 import os
 import uuid
 from http import HTTPStatus
-from typing import Any, Dict, Iterator, List, Optional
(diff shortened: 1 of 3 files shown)


libs/community/langchain_community/document_loaders/pebblo.py
```diff
@@ -43,22 +44,25 @@ def __init__(
         owner: str = "",
         description: str = "",
         api_key: Optional[str] = None,
+        load_semantic: Optional[bool] = False,
```
baskaryan, line 47: doesn't need to be Optional[bool] can just be bool

baskaryan (approved):

PR #20474, opened 2024-04-15
Fix pgembedding

aminghani:
in the PGEmbedding class in pgembedding, the Embedding function is required and can't be None, but in some of functions such as add_embeddings which accepts text as 
a list and embeddings and a list of lists, meaning that the embeddings are constructed outside of this class, so sometime we do not need to initialize the class with embedding function. there are other functions with do not require embedding function as well such as similarity_search_with_score_by_vector,  similarity_search_by_vector and other functions, so it is better to set a default of None to embedding function and there is also a function added with name of set_embedding_function for setting embedding function after initialization. plus when using functions which require using an embedding function, user is asked to provide the function in the initialization or using set_embedding_function.


eyurtsev:
@aminghani could you resolve conflict and merge against master? The PR shows a lot of changes. Also consider seeing if the vectorstor in langchain-postgres works for you: https://github.com/langchain-ai/langchain-postgres/blob/main/examples/vectorstore.ipynb

if you're up for the task, you could also move this vectorstore to langchain-postgres. The package has testing against postgres as part of CI, so it's much easier to maintain

eyurtsev:
Closing this PR for now due to lack of response

PR #20477, opened 2024-04-15
patch: deprecate (a)get_relevant_documents

ccurme:
- `.get_relevant_documents(query)` -> `.invoke(query)`
- `.get_relevant_documents(query=query)` -> `.invoke(query)`
- `.get_relevant_documents(query, callbacks=callbacks)` -> `.invoke(query, config={"callbacks": callbacks})`
- `.get_relevant_documents(query, **kwargs)` -> `.invoke(query, **kwargs)`


libs/core/langchain_core/retrievers.py
```diff
@@ -262,6 +263,7 @@ async def _aget_relevant_documents(
             run_manager=run_manager.get_sync(),
         )
 
+    @deprecated("0.1.43", alternative="invoke", removal="0.2.0")
```
ccurme, line 266: what should go here?

libs/community/langchain_community/retrievers/milvus.py
```diff
@@ -59,8 +59,8 @@ def _get_relevant_documents(
         run_manager: CallbackManagerForRetrieverRun,
         **kwargs: Any,
     ) -> List[Document]:
-        return self.retriever.get_relevant_documents(
```
ccurme, line 62: is there any magic with callbacks that would make these the same for a BaseRetriever?
Case 1:
```python
self._get_relevant_documents(query, run_manager=run_manager.get_child())
```
Case 2:
```python
callback_manager = CallbackManager.configure(callbacks=run_manager.get_child(), ...)
run_manager = callback_manager.on_retriever_start(...)
result = self._get_relevant_documents(query)
```

libs/core/langchain_core/retrievers.py
efriis: Best way to check is just to run it and see the warning gets printed. I tend to add this but otherwise lgtm
```suggestion
    @deprecated(since="0.1.43", alternative="invoke", removal="0.2.0")
```

libs/community/langchain_community/retrievers/milvus.py
efriis: Can call it in `def get_relevant_documents()` probably if `run_manager` is passed in as a kwarg? Seems similar to how we do it in chat model token generation

efriis: Sorry - in *invoke

libs/core/langchain_core/retrievers.py
```diff
@@ -325,6 +327,7 @@ def get_relevant_documents(
             )
             return result
 
+    @deprecated("0.1.43", alternative="invoke", removal="0.2.0")
```
efriis, line 330:
```suggestion
    @deprecated(since="0.1.43", alternative="invoke", removal="0.2.0")
```

libs/core/langchain_core/retrievers.py
efriis: I think you also need to wrap the `self.get_relevant_documents()` call in `invoke` and `ainvoke` with some kind of warning suppressor (until we swap logic to be implemented in invoke and remove get_relevant_documents in 0.2)

efriis (commented):
my guess is there's a bunch of find and replace in docs and cookbooks too

libs/core/langchain_core/retrievers.py
ccurme: I'm not getting warnings on .invoke and I think it's because of the check for `is_caller_internal` in the decorator, but happy to add the warning suppressor if I'm mistaken (or if it's just prudent)

https://github.com/langchain-ai/langchain/blob/c909ae01523bdbfb32b5266d0de617643a82f355/libs/core/langchain_core/_api/deprecation.py#L145

libs/community/langchain_community/retrievers/milvus.py
ccurme: ah makes sense thank you!

ccurme:
> my guess is there's a bunch of find and replace in docs and cookbooks too

thanks, changed docs here: https://github.com/langchain-ai/langchain/pull/20698

PR #20480, opened 2024-04-15
core: Add concept of conditional edge to graph rendering

nfcampos:
- implement for mermaid, graphviz and ascii
- this is to be used in langgraph

### DIFF
--- a/libs/core/langchain_core/runnables/graph.py
+++ b/libs/core/langchain_core/runnables/graph.py
@@ -19,7 +19,6 @@ from typing import (
 from uuid import UUID, uuid4
 
 from langchain_core.pydantic_v1 import BaseModel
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/runnables/graph.py
```diff
@@ -220,13 +220,21 @@ def remove_node(self, node: Node) -> None:
             if edge.source != node.id and edge.target != node.id
         ]
 
-    def add_edge(self, source: Node, target: Node, data: Optional[str] = None) -> Edge:
+    def add_edge(
+        self,
+        source: Node,
+        target: Node,
+        data: Optional[str] = None,
+        conditional: bool = False,
```
baskaryan, line 228: for @eyurtsev's sanity would be nice to add `*` before conditional

PR #20485, opened 2024-04-15
experimental: Makes regex customizable in text_splitter.py (SemanticChunker class)

GustavoSept:
- **Description:** Currently, the regex is static (`r"(?<=[.?!])\s+"`), which is only useful for certain use cases. The current change only moves this to be a parameter of split_text(). Which adds flexibility without making it more complex (as the default regex is still the same).
  - **Issue:** Not applicable (I searched, no one seems to have created this issue yet). 
  - **Dependencies:** None.

_If no one reviews your PR within a few days, please @-mention one of baskaryan, efriis, eyurtsev, hwchase17._


GustavoSept:
Optionally, i think it's a good idea to implement the same parameter in `create_documents()`, `split_documents()` and `transform_documents()`. Even if by the usage of kwargs.

eyurtsev (dismissed):
Thanks for the contribution! We can make the change, but need to pass the config through __ init __. `split_text` should only accept `text`

GustavoSept:
@eyurtsev Thanks, that actually makes much more sense :D

I've updated the PR with a new commit, moving `regex` to __init__ as `sentence_split_regex`.

PR #20487, opened 2024-04-15
community[patch]: Use uuid4 not uuid1

eyurtsev:
Using UUID1 is incorrect since it's time dependent, which makes it easy to generate the exact same uuid


liugddx (approved):
LGTM

Issue #20493, opened 2024-04-16
Callbacks need some TLC


sepiatone:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Callbacks documentation and code is a little ragged around the edges (imo).

### Error Message and Stack Trace (if applicable)

See the Callbacks [page](https://python.langchain.com/docs/modules/callbacks/).

### Description

The documentation has to be brought to the LCEL era, updated to include events like `on_retriever_start` and given a similar look and feel to other pages (e.g. Agents).

The code for the built-in handlers `StdOutCallbackHandler` has to be (lightly) modified for the LCEL era. `FileCallbackHandler` has to be moved from `community` to `core`

### System Info

NA

sepiatone:
* PR [#19845](https://github.com/langchain-ai/langchain/pull/19845), update to the Callbacks landing page as part of issue [#19361](https://github.com/langchain-ai/langchain/issues/19361) (also fixes [#16379](https://github.com/langchain-ai/langchain/issues/16379))
  - concepts - table of all events, included tags here
  - quickstart with LCEL examples
  - similar look and feel to Agents page

* PR [#20494](https://github.com/langchain-ai/langchain/pull/20494), update to the custom callback handler page

* PR [#20495](https://github.com/langchain-ai/langchain/pull/20495), move the `FileCallbackHandler` from community to core

* PR [#20496](https://github.com/langchain-ai/langchain/pull/20496), update to the `FileCallbackHandler` documentation

sepiatone:
If some (or all) of these prs are accepted sidebars.js may need to be modified.

sepiatone:
Closing, all issues fixed.

PR #20494, opened 2024-04-16
docs[callbacks]: custom callback handlers page

sepiatone:
**Description:** Update to the Callbacks page on custom callback handlers
**Issue:** #20493 
**Dependencies:** None


eyurtsev:
Perhaps we could use ChatOpenAI the chat models are much more popular than the non chat variants?

sepiatone:
> Perhaps we could use ChatOpenAI the chat models are much more popular than the non chat variants?

ok, done. though I'm not sure why the deployment is failing...

baskaryan (approved):

PR #20507, opened 2024-04-16
anthropic: env ANTHROPIC_API_URL not work

fayfox:
enviroment variable ANTHROPIC_API_URL will not work if anthropic_api_url has default value


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -251,7 +251,7 @@ class Config:
     max_retries: int = 2
     """Number of retries allowed for requests sent to the Anthropic Completion API."""
 
-    anthropic_api_url: str = "https://api.anthropic.com"
+    anthropic_api_url: Optional[str] = Field(default=None)
```
eyurtsev, line 254:
```suggestion
    anthropic_api_url: Optional[str] = None
```

eyurtsev (approved):

PR #20518, opened 2024-04-16
Add missing method similarity_search_with_relevance_scores to AzureCosmosDBVectorSearch

Zythoxy:
community: Add missing method similarity_search_with_relevance_scores() to AzureCosmosDBVectorSearch in azure_cosmos_db.py 

**Change**: 
Add missing method to AzureCosmosDBVectorSearch in azure_cosmos_db.py. This method should already already exist according to the documentation here, but doesn't: https://api.python.langchain.com/en/latest/vectorstores/langchain_community.vectorstores.azure_cosmos_db.AzureCosmosDBVectorSearch.html

**Issue:** 
When used as a retriever with search_type="similarity_score_threshold", AzureCosmosDBVectorSearch does throw a NotImplemented() error, because the method similarity_search_with_relevance_scores() is not there.


Zythoxy:
I just noticed that the method should be inherited from VectorStore. For those having the same issue as I am having though, the following fixes it for me: 

VECTOR_STORE = AzureCosmosDBVectorSearch.from_connection_string(
    connection_string=MONGODB_CONNECTION_STRING,
    namespace=f"{MONGODB_DATABASE_NAME}.{COLLECTION_NAME_VECTOR_DB}",
    embedding=EMBEDDINGS_CLIENT,
)
**VECTOR_STORE.similarity_search_with_relevance_scores = (
    VECTOR_STORE.similarity_search_with_score
)**

PR #20519, opened 2024-04-16
anthropic: support tool messages

ccurme:


eyurtsev (approved):

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -282,3 +283,30 @@ def test_with_structured_output() -> None:
     response = structured_llm.invoke("what's the weather in san francisco, ca")
     assert isinstance(response, dict)
     assert response["location"]
+
+
+def test_tool_message() -> None:
```
eyurtsev, line 288: Just a test to show that it's broken right now?

efriis: Can we add this to standard integration tests instead if the tool calling fixture is true?

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -196,7 +196,22 @@ def _format_messages(messages: List[BaseMessage]) -> Tuple[Optional[str], List[D
                         f"Content items must be str or dict, instead was: {type(item)}"
                     )
         else:
-            content = message.content
+            if (
+                isinstance(message, AIMessage)
+                and message.content == ""
+                and message.tool_calls
+            ):
+                content = [
```
nfcampos, line 204: can we support putting string content, if not empty, as an additional part of type text?

baskaryan (approved):

Issue #20523, opened 2024-04-16
MAX_TOKENS in MistralAIEmbeddings is incorrect


ReinforcedKnowledge:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain_mistralai.embeddings import MistralAIEmbeddings

```python
embeddings = MistralAIEmbeddings()

pass_test_str = "hello world" * 4094
embedded_pass_test_str = embeddings.embed_documents([pass_test_str])

print(f"Maximum number of tokens that pass: {len(embeddings.tokenizer.encode(pass_test_str))}")  # 8190
print(f"Embedding dimension: {len(embedded_pass_test_str[0])}")  # 1024

fail_test_str = "hello world" * 4095

print(f"Number of tokens: {len(embeddings.tokenizer.encode(fail_test_str))}")  # 8192

embedded_fail_test_str = embeddings.embed_documents([fail_test_str])
```

### Error Message and Stack Trace (if applicable)

An error occurred with MistralAI: 'data'
Traceback (most recent call last):
  File "/Users/y.tahtah/test_langchain_mistralai_embeddings/test.py", line 15, in <module>
    embedded_fail_test_str = embeddings.embed_documents([fail_test_str])
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/y.tahtah/test_langchain_mistralai_embeddings/venv/lib/python3.12/site-packages/langchain_mistralai/embeddings.py", line 135, in embed_documents
    for embedding_obj in response.json()["data"]
                         ~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'data'

### Description

I'm trying to embed some text using `MistralAIEmbeddings` and I have split my text according to the `MAX_TOKENS` variable in `libs/partners/mistralai/langchain_mistralai` but it's not working. Further investigation with the example code provided with this issue led me to find that the embedding models fails to embed a document way before the 16k token indicated by `MAX_TOKENS`.

[MistralAI's official page on endpoints](https://docs.mistral.ai/platform/endpoints/) doesn't specify a context window size for the embedding model.

[...]

PR #20525, opened 2024-04-16
Changed MAX_TOKENS to the correct value of 8190

ReinforcedKnowledge:

- **Changed MAX_TOKENS to the correct value of 8190**: "langchain-mistralai"

- **Description:** Changed the value of `MAX_TOKENS`, context window size of the MistralAI embed model) in the embeddings

- **Issue:** #20523


ccurme:
Hi @ReinforcedKnowledge, thanks for this. I'm finding that after this change we still get an error for the `"hello world" * 4095` case described in the [issue](https://github.com/langchain-ai/langchain/issues/20523), although it is different:
`{'object': 'error', 'message': 'Please provide at least one input element.', 'type': 'invalid_request_error', 'param': None, 'code': None}`
Do you find the same?

efriis:
Commented on the issue!

PR #20527, opened 2024-04-16
docs: response metadata

baskaryan:


docs/docs/modules/model_io/chat/token_usage_tracking.ipynb
baskaryan, line 1: @hinthornw get_openai_callback doesn't seem to work for agent

PR #20529, opened 2024-04-16
Runnable graph viz improvements

nfcampos:
- Add conditional: bool property to json representation of the graphs
- Add option to generate mermaid graph stripped of styles (useful as a text representation of graph)

### DIFF
--- a/libs/core/langchain_core/runnables/graph.py
+++ b/libs/core/langchain_core/runnables/graph.py
@@ -170,6 +170,17 @@ class Graph:
             node.id: i if is_uuid(node.id) else node.id
             for i, node in enumerate(self.nodes.values())
         }
+        edges: List[Dict[str, Any]] = []
+        for edge in self.edges:
+            edge_dict = {
+                "source": stable_node_ids[edge.source],
+                "target": stable_node_ids[edge.target],
+            }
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/runnables/graph.py
```diff
@@ -344,6 +343,7 @@ def draw_png(
 
     def draw_mermaid(
         self,
+        with_styles: bool = True,
```
eyurtsev, line 346: This is technically a breaking change, could we use `*` in this API to force calling by named argument?

libs/core/langchain_core/runnables/graph_mermaid.py
```diff
@@ -16,6 +16,7 @@ def draw_mermaid(
     edges: List[Edge],
     first_node_label: Optional[str] = None,
     last_node_label: Optional[str] = None,
+    with_styles: bool = True,
```
eyurtsev, line 19: Same here -- it's a breaking change in the API

I'd love if we used `/` and `*` operators on all function signatures

libs/core/langchain_core/runnables/graph.py
nfcampos: done

PR #20554, opened 2024-04-17
mistral: read tool calls from AIMessage

ccurme:


libs/partners/mistralai/langchain_mistralai/chat_models.py
```diff
@@ -196,6 +199,34 @@ def _convert_delta_to_message_chunk(
...
+        }
+    }
+    if _id := tool_call.get("id"):
+        result["id"] = _id
+
+    return result
+
+
+def _format_invalid_tool_call_for_mistral(tool_call: InvalidToolCall) -> dict:
+    """Format Langchain ToolCall to dict expected by Mistral."""
```
eyurtsev, line 217: ToolCall should InvalidToolCall

libs/partners/mistralai/langchain_mistralai/chat_models.py
```diff
@@ -216,8 +247,15 @@ def _convert_message_to_mistral_chat_message(
                 if _id := tc.get("id"):
                     chunk["id"] = _id
                 tool_calls.append(chunk)
+        elif message.tool_calls or message.invalid_tool_calls:
```
eyurtsev, line 250: Should this be prioritized over additional_kwargs? The latter should be the source of truth if swapping providers etc

eyurtsev (approved):

libs/partners/mistralai/langchain_mistralai/chat_models.py
ccurme: ah yeah if we have a provider that (1) formats tool_calls into additional_kwargs, and (2) does not use OpenAI-style formatting then this will break. can update.

ccurme: @efriis see above - is it OK to lose the behavior maintained in unit tests, in which applying `_convert_message_to_mistral_chat_message` and `_convert_mistral_chat_message_to_message` returns the original message identically?

PR #20555, opened 2024-04-17
mistralai[patch]: Surface http errors

eyurtsev:
Do not swallow errors when streaming with httpx.

Update affected code if this PR gets merged to httpx: https://github.com/florimondmanca/httpx-sse/pull/25/files


eyurtsev:
Looks like we actually should consume the text in the response since that contains the full error message from the server. I'll update the code here in a bit.

PR #20557, opened 2024-04-17
mistral[patch]: Support both model and model_name

eyurtsev:


eyurtsev:
@baskaryan / @efriis 

Options:
1) support both model and model name
2) force all models to forbid extras (using standard test client) 
3) do both (1) and (2)

I'm strongly in favor of 2. And don't care about 1 if we do (2).

ccurme (approved):

PR #20572, opened 2024-04-17
anthropic[patch]: fix msg mutation

baskaryan:


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -95,11 +96,12 @@ def _format_image(image_url: str) -> Dict:
 
 
 def _merge_messages(
-    messages: List[BaseMessage],
+    messages: Sequence[BaseMessage],
 ) -> List[Union[SystemMessage, AIMessage, HumanMessage]]:
     """Merge runs of human/tool messages into single human messages with content blocks."""  # noqa: E501
     merged: list = []
     for curr in messages:
+        curr = curr.__class__(**deepcopy(curr.dict()))
```
ccurme, line 104: ooc why not `curr.copy(deep=True)`?

ccurme (approved):

libs/partners/anthropic/langchain_anthropic/chat_models.py
baskaryan: didn't know that was an option 🌝

PR #20575, opened 2024-04-17
Community: Fix MLX LLM Stream

### LINKED ISSUE #20561: Mistype issue using MLX model via MLXPipeline
### Checked other resources

### Example Code

The following code 

```
from langchain_community.llms import OpenAI, Ollama, MLXPipeline

model, tokenizer = load("mlx-community/dolphin-2.8-mistral-7b-v02")
self.mlx = MLXPipeline(
    model=model,
    tokenizer=tokenizer,
    pipeline_kwargs={"temp":0.7, "max_tokens":10}
)
```

on the following prompt

```
Collect and summarize recent news articles, press releases, and market analyses related to the company. Pay special attention to any significant events, market, sentiments, and analysts' opinions.
Your final answer MUST be a report that includes a comprehensive identification of key points makrting oriented following the Marketing 5Ps (Product, Place, Price, Promotion, People)
If you do your BEST WORK, I will give you a $10,000 commission!
Make sure to use the most recent data as possible.
Selected company by the customer is Tesla
```

Lead to an error during execution.

```
  File "/opt/homebrew/lib/python3.10/site-packages/langchain_community/llms/mlx_pipeline.py", line 189, in _stream
    text = self.tokenizer.decode(token.item())
AttributeError: 'int' object has no attribute 'item'
```

### Description

Hi

* I am trying to use Langchain for loading a MLX model (cf code) on a given prompt.
* I face the error available in the error section: `AttributeError: 'int' object has no attribute 'item'`

Removing the `.items()` on the line 182 unlock the issue however I have nothing as result.

So my idea is not correct.

The file  `libs/community/langchain_community/llms/mlx_pipeline.py` has been added last week so it is very new.
Could you take a look @Blaizzy  ?

Thank you 

### System Info

here is the version I use:

Python 3.10

```
pip freeze | grep langchain
langchain==0.1.12
langchain-community==0.0.33
langchain-core==0.1.43
langchain-openai==0.0.5
```

Blaizzy:
Closes #20561

This PR fixes MLX LLM stream `AttributeError`. 

Recently, `mlx-lm` changed the token decoding logic, which affected the LC+MLX integration. 

Additionally, I made minor fixes such as: docs example broken link and enforcing pipeline arguments (max_tokens, temp and etc) for invoke.
   
- **Issue:** #20561
    
- **Twitter handle:** @Prince_Canuma


libs/community/langchain_community/llms/mlx_pipeline.py
```diff
@@ -135,9 +135,32 @@ def _call(
                 "Please install it with `pip install mlx_lm`."
             )
 
-        pipeline_kwargs = kwargs.get("pipeline_kwargs", {})
+        pipeline_kwargs = kwargs.get("pipeline_kwargs", self.pipeline_kwargs)
 
-        return generate(self.model, self.tokenizer, prompt=prompt, **pipeline_kwargs)
+        temp: float = pipeline_kwargs.get("temp", 0.0)
+        max_tokens: int = pipeline_kwargs.get("max_tokens", 100)
```
baskaryan, line 141: 100 seems very small — is that number coming from somewhere? does max_tokens need to be specified to generate at all, or would it be better to leave it as None if user doesn't set it?

Blaizzy: Hey @baskaryan 

Thanks for the quick review!

100 is small so the generation can be quick and is also the default value also used in [MLX-LM](https://github.com/Blaizzy/mlx-examples/blob/main/llms/mlx_lm/utils.py#L201C1-L212C10). 

If the user wants they can pass a bigger number this ways:
```
from langchain_community.llms.mlx_pipeline import MLXPipeline

pipe = MLXPipeline.from_model_id(
    "mlx-community/dolphin-2.8-mistral-7b-v02-4bit",
    pipeline_kwargs={"max_tokens": 5000},
)
```

baskaryan (commented):
can we clearly document these default values somewhere? otherwise user may never know how they're being set / how to update thme

Blaizzy:
> can we clearly document these default values somewhere? otherwise user may never know how they're being set / how to update thme

Hey @baskaryan 

I got swamped last couple weeks,

Where would you like me to do it. Directly on the file level or there is a designated place?

hwchase17:
> > can we clearly document these default values somewhere? otherwise user may never know how they're being set / how to update thme
> 
> Hey @baskaryan
> 
> I got swamped last couple weeks,
> 
> Where would you like me to do it. Directly on the file level or there is a designated place?

just inline is probably fine!

Blaizzy:
> just inline is probably fine!

Thanks @hwchase17 and @baskaryan!

It's done ✅

Blaizzy:
Hey @hwchase17 @baskaryan 

Is there anything else missing?

baskaryan (approved):

PR #20579, opened 2024-04-18
docs: fact checking chain example

leo-gan:
Added notebook with the `LLMCheckerChain` example. There are also pictures for this example.


efriis (commented):
Hey there! The featured use case docs should use our recommended interfaces, such as the chain constructor methods. This uses a [legacy chain](https://python.langchain.com/docs/modules/chains/#legacy-chains), so I don't think it makes sense to feature here like this.

Would you be interested in migrating it to an LCEL chain and documenting that?

leo-gan:
> Hey there! The featured use case docs should use our recommended interfaces, such as the chain constructor methods. This uses a [legacy chain](https://python.langchain.com/docs/modules/chains/#legacy-chains), so I don't think it makes sense to feature here like this.
> 
> Would you be interested in migrating it to an LCEL chain and documenting that?

@efriis OK I'll do this.

PR #20617, opened 2024-04-18
fix: rephraser dropping retriever kwargs

buroa:
Simple fix :)


libs/langchain/langchain/retrievers/re_phraser.py
```diff
@@ -62,6 +62,7 @@ def _get_relevant_documents(
         query: str,
         *,
         run_manager: CallbackManagerForRetrieverRun,
+        **kwargs: Any,
```
eyurtsev, line 65: Base interface does not accept **kwargs:

```
    @abstractmethod
    def _get_relevant_documents(
        self, query: str, *, run_manager: CallbackManagerForRetrieverRun
    ) -> List[Document]:
        """Get documents relevant to a query.
        Args:
            query: String to find relevant documents for
            run_manager: The callbacks handler to use
        Returns:
            List of relevant documents
        """
```

Can you provide configuration through initializer -- this is the usual strategy

buroa: > Base interface does not accept **kwargs:
> 
> ```
>     @abstractmethod
>     def _get_relevant_documents(
>         self, query: str, *, run_manager: CallbackManagerForRetrieverRun
>     ) -> List[Document]:
>         """Get documents relevant to a query.
>         Args:
>             query: String to find relevant documents for
>             run_manager: The callbacks handler to use
>         Returns:
>             List of relevant documents
>         """
> ```
> 
> Can you provide configuration through initializer -- this is the usual strategy

Erm... https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/retrievers/contextual_compression.py#L15-L34

PR #20624, opened 2024-04-18
community[patch]: Fix Ollama for LLaMA3

rlancemartin:
We see verbose generations w/ LLaMA3 and Ollama - 
https://smith.langchain.com/public/88c4cd21-3d57-4229-96fe-53443398ca99/r

--- 

Fix here implies that when stop was being set to an empty list, the stream had no conditions under which to stop, which could lead to excessive or unintended output.

Test LLaMA2 - 
https://smith.langchain.com/public/57dfc64a-591b-46fa-a1cd-8783acaefea2/r

Test LLaMA3 - 
https://smith.langchain.com/public/76ff5f47-ac89-4772-a7d2-5caa907d3fd6/r
https://smith.langchain.com/public/a31d2fad-9094-4c93-949a-964b27630ccb/r

Test Mistral -
https://smith.langchain.com/public/a4fe7114-c308-4317-b9fd-6c86d31f1c5b/r


libs/community/langchain_community/llms/ollama.py
```diff
@@ -203,8 +203,8 @@ def _create_stream(
             raise ValueError("`stop` found in both the input and default params.")
         elif self.stop is not None:
             stop = self.stop
-        elif stop is None:
-            stop = []
+        # elif stop is None:
+        #    stop = []
```
efriis, line 207: should delete instead of comment
```suggestion
```

efriis (commented):
are we certain this won't break some ollama models while helping others? if statements like this are usually introduced to fix other bugs.

Should also fix in `_acreate_stream` if we proceed with this

rlancemartin:
> are we certain this won't break some ollama models while helping others? if statements like this are usually introduced to fix other bugs.
> 
> Should also fix in `_acreate_stream` if we proceed with this

I don't have context on why this was added.

I tested llama2, mistral, and llama3.

I do see unit test failure:
https://github.com/langchain-ai/langchain/actions/runs/8745263844/job/23999934675?pr=20624

Unit test expects stop to be an empty list:
```
Expected Output:
The stop key in the options dictionary is expected to be an empty list [].
```

libs/community/langchain_community/llms/ollama.py
rlancemartin: Fix is also made in `_acreate_stream`.

efriis (approved):
should get CI passing before merge

jacoblee93:
Don't know why it was there to begin with, my guess is an oversight? Empty array does evaluate to `False` in Python.

It's not present in the JS invocation params and nobody has complained, so I think pretty safe:

https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/chat_models/ollama.ts#L218

rlancemartin:
> should get CI passing before merge

ya, doc imports issue appears unrelated?

https://github.com/langchain-ai/langchain/actions/runs/8745862984/job/24001675709?pr=20624

efriis:
Fixed in #20629 - will wait for that to auto merge and then will update branch here

jmorganca (approved):

PR #20630, opened 2024-04-18
docs: API Reference inherited members fix

leo-gan:
I'm trying to fix a bug in the API Ref generation when the members of the superclass go into the documentation. Here is an [example](https://api.python.langchain.com/en/latest/llms/langchain_community.llms.aleph_alpha.AlephAlpha.html#langchain_community.llms.aleph_alpha.AlephAlpha). Here the `abatch`, `agenerate` methods are not presented in the `AlephAlpha` class but inherited from -> LLM --> BaseLLM class. As the result, we have bloated, unreadable pages.
I'm following this [description](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html). But it could not work. There were complaints about bugs in this part of the Sphinx generation.

@baskaryan Please, verify it, since I cannot verify it locally, nor can Vercel generate the PR API Ref for now.

### DIFF
--- a/docs/api_reference/templates/pydantic.rst
+++ b/docs/api_reference/templates/pydantic.rst
@@ -11,7 +11,7 @@
     :field-signature-prefix: param
     :members:
     :undoc-members:
-    :inherited-members:
+    :inherited-members: False
     :member-order: groupwise
     :show-inheritance: True
     :special-members: __call__


baskaryan:
mm i'm pretty sure we want to show all methods, including inherited ones. otherwise a user needs to know which parent classes to look at for any given class, and that seems much more confusing. if i'm looking at AlephAlpha i just want to know all the methods i can use, and that includes inherited ones

leo-gan:
@baskaryan The reading profile for API Ref is developer. And there are two main use cases:
1. A dev that wants to update the class or create a new one similar to this class. Then it is more logical to go by links to the base classes to get the inherited methods because it gives me idea of the class hierarchy and how to deal with this hierarchy. Probably, not a big deal with current view, because this developer will go directly to the code and API Ref is just a helper.
2. If I'm a class user, writing the client for this class. Then I need all the methods in one place. But the current huge page is unbearable. We need a ToC (right-side navbar) with the class methods/members or the collapsable view for these methods (click to collapse/expand).

What do you prefer? Maybe you have other ideas?
- ToC (right-side navbar) with the class methods/members
- the collapsable view for all methods (click to collapse/expand)

baskaryan:
> @baskaryan The reading profile for API Ref is developer. And there are two main use cases:
> 
> 1. A dev that wants to update the class or create a new one similar to this class. Then it is more logical to go by links to the base classes to get the inherited methods because it gives me idea of the class hierarchy and how to deal with this hierarchy. Probably, not a big deal with current view, because this developer will go directly to the code and API Ref is just a helper.
> 2. If I'm a class user, writing the client for this class. Then I need all the methods in one place. But the current huge page is unbearable. We need a ToC (right-side navbar) with the class methods/members or the collapsable view for these methods (click to collapse/expand).
> 
> What do you prefer? Maybe you have other ideas?
> 
> * ToC (right-side navbar) with the class methods/members
> * the collapsable view for all methods (click to collapse/expand)

a ToC would be great! would need to figure out how to do it

leo-gan:
@baskaryan Now we have inheritent members in the Pydantic template but not in the Class template. Do wee need to add it to Class template for symmetry?

leo-gan:
@baskaryan scikit-learn can do this (the ToC with the class methods)
image

sepiatone:
@leo-gan Sorry to impose myself on this conversation

I agree that the API reference documentation is difficult to browse (I've shifted to just loading up the latest LangChain sources in my IDE) and that a ToC for class methods would be a welcome addition.

But wouldn't it be better to get the API reference documentation to be displayed by Quarto / Docasaurus (or some combination). This will bring in a uniform look and feel and reduce the cognitive dissonance when moving from the Tutorials and Use-Cases to the API references.

Building documentation is a little bit out of my wheelhouse, so there is probably a very good reason why things are done this way.

leo-gan:
@sepiatone Thank you for your thoughts and feedback!
Uniformity in API Ref regarding other documentation (let's call it Docs, for brevity) could be good or not. If it is not uniform, how big is the non-uniformity issue? 

The requirements for API and Docs are different. 

**API** is a **tool for developers.** The most important requirements: the fast search for specific things (classes, functions, signatures); uniformity (everything is in the same format); full coverage. As a result of these requirements, we **must** generate API automatically. The beauty requirement is not important; the easy/intuitive navigation and the easy/intuitive formats are very important.

**Docs** is a tool **for learning and selecting** the right tools and technologies. Doc pages are created manually by different people at different times and are constantly evolving. Doc's ToCs and page hierarchies are hard and biased. As a result, Docs ToCs and the page formats are hard to automate, and they are manually curated.

So, IMO, it is OK to have different formats for API and Docs. Docs are like the teaching materials and advertising brochures for the boat, and API is like the user manuals for all boat parts. They are rarely used together at the same time.

sepiatone:
@leo-gan Thanks for taking the trouble to respond with a detailed explanation!

> So, IMO, it is OK to have different formats for API and Docs. Docs are like the teaching materials and advertising brochures for the boat, and API is like the user manuals for all boat parts. They are rarely used together at the same time.

This makes sense - perhaps I've been biased by my situation. I started off looking at the Docs and then moved on to the API Ref when I thought something needed to be changed. So in my case there was (and is) a fair bit of back n' forth. This may not apply to most folks.

A ToC for the API ref would be very helpful and would help make it more organized!

Issue #13749, opened 2023-11-22
UTF-8 content is not loaded correctly with RecursiveUrlLoader 


Alejandro-M-Cruz:
### System Info

Langchain version: 0.0.339
Python version: 3.11.5
Running on Ubuntu 22.04.3 LTS via WSL 2

### Information

- [X] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

1. Running the steps described in the guide for RecursiveUrlLoader:  [https://python.langchain.com/docs/integrations/document_loaders/recursive_url](url)

2. Loading some content with UTF-8 encoding, the Python docs for example: [https://docs.python.org/3/](url)
Exact code used: 
```python
from pprint import pprint

from langchain.document_loaders import RecursiveUrlLoader
from bs4 import BeautifulSoup


def load_python_docs():
    url = "https://docs.python.org/3/"
    loader = RecursiveUrlLoader(
        url=url, max_depth=2, extractor=lambda x: BeautifulSoup(x, "html.parser").text
    )
    return loader.load()


pprint([doc.metadata for doc in load_python_docs()])
```

3. If you print the loaded documents, you should be able to see these kind of encoding issues:
```
{'description': 'Editor, Adam Turner,. This article explains the new features '
                 'in Python 3.12, compared to 3.11. Python 3.12 was released '
                 'on October 2, 2023. For full details, see the changelog. '
                 'Summary â\x80\x93 Release hi...',
  'language': None,
  'source': 'https://docs.python.org/3/whatsnew/3.12.html',
  'title': 'Whatâ\x80\x99s New In Python 3.12 — Python 3.12.0 documentation'}
```

### Expected behavior

Should load content correctly, using the right encoding to parse the document. I suppose the issue is due to the fact that the `_get_child_links_recursive` method is calling `requests.get` and not specifying the encoding for the response. A quick fix that worked for me was to include the following line, just after the GET request:  `response.encoding = response.apparent_encoding`

PR #20632, opened 2024-04-19
Using the right encoding to parse the web page in RecursiveUrlLoader

fubuki8087:
As shown in #13749 , `RecursiveUrlLoader` has encoding issue. This PR is to solve this.


baskaryan (approved):

libs/community/langchain_community/document_loaders/recursive_url_loader.py
```diff
@@ -169,6 +169,7 @@ def _get_child_links_recursive(
         visited.add(url)
         try:
             response = requests.get(url, timeout=self.timeout, headers=self.headers)
+            response.encoding = response.apparent_encoding
```
eyurtsev, line 172: I don't think that apparent encoding is bullet proof. It's an inferred encoding and the inferred encoding can be incorrect while the actual encoding could be correct. Is that not the case?

It looks like the logic needs to be a bit more complex or opt-in rather than the default?

fubuki8087: Maybe we can add arguments `encoding` and `autoset_encoding` just like `WebBaseLoader`.
https://github.com/langchain-ai/langchain/blob/940242c1ecacfa85a5de85a261c68bb23174bdad/libs/community/langchain_community/document_loaders/web_base.py#L49-L50

If `encoding` value is given by users, we use the given one. Otherwise, if `autoset_encoding` is `True` (by default), we use apparent encoding.
https://github.com/langchain-ai/langchain/blob/940242c1ecacfa85a5de85a261c68bb23174bdad/libs/community/langchain_community/document_loaders/web_base.py#L232-L235

How does it sound like?

eyurtsev (approved):

libs/community/langchain_community/document_loaders/recursive_url_loader.py
```diff
@@ -169,6 +178,12 @@ def _get_child_links_recursive(
         visited.add(url)
         try:
             response = requests.get(url, timeout=self.timeout, headers=self.headers)
+
+            if self.encoding is not None:
```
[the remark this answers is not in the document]
fubuki8087: `response.text` actually do what you said, and this is the current way to use the `response` in `RecursiveUrlLoader`
https://github.com/langchain-ai/langchain/blob/804390ba4bcc306b90cb6d75b7f01a4231ab6463/libs/community/langchain_community/document_loaders/recursive_url_loader.py#L198
https://github.com/langchain-ai/langchain/blob/804390ba4bcc306b90cb6d75b7f01a4231ab6463/libs/community/langchain_community/document_loaders/recursive_url_loader.py#L202
https://github.com/langchain-ai/langchain/blob/804390ba4bcc306b90cb6d75b7f01a4231ab6463/libs/community/langchain_community/document_loaders/recursive_url_loader.py#L206-L207
In `response.text`, it decodes `response.content` by
```python
content = str(self.content, encoding, errors="replace")
return content
```
It's just like your `text = response.content.decode(encoding=encoding)`

PR #20654, opened 2024-04-19
community: Add async methods to CassandraCache and CassandraSemanticCache

cbornet:


libs/community/langchain_community/cache.py
```diff
@@ -1123,17 +1149,43 @@ def delete_through_llm(
         )[1]
         return self.delete(prompt, llm_string=llm_string)
 
+    async def adelete_through_llm(
```
eyurtsev, line 1152: Don't think this makes sense anymore for ChatModels. This probably worked with non chat models when the `prompt` was just a plain string. Now that the prompt is usually a list of messages, users won't be able to guess the prompt either.

Feel free to keep the async implementation here, but don't think that the `deletion` works for chat models.

eyurtsev (approved):

eyurtsev:
looks good to me, if you're able to update i can merge whenever

libs/community/langchain_community/cache.py
cbornet: I removed the functions. We can add them later if there's a need for them.

cbornet:
I rebased and fixed conflicts.

PR #20665, opened 2024-04-19
step1. adding langchain_google_community to the docs

lkuligin:

- [ ] **PR title**: "docs: step1. adjusting langchain_community -> langchain_google_community"

- [ ] 
    - **Description:** step1. adjusting langchain_community -> langchain_google_community


baskaryan (commented):
should we be adding the installation command as wel `pip install langchain-google-community`

baskaryan (approved):

PR #20666, opened 2024-04-19
community: Remove usage of deprecated StoredBlobHistory in CassandraChatMessageHistory

cbornet:


cbornet:
cc @hemidactylus

eyurtsev:
@cbornet is this backwards compatible? Do we need to warn users or OK as is?

cbornet:
Yes it’s backward compatible and seamless for the users.

eyurtsev (approved):

PR #20667, opened 2024-04-19
core[minor],langchain[patch]: Move base indexing interface and logic to core

eyurtsev:
This PR moves the interface and the logic to core.

The following changes to namespaces:

`indexes` -> `indexing`
`indexes._api` -> `indexing.api`

Testing code is intentionally duplicated for now since it's testing different
implementations of the record manager (in-memory vs. SQL).

Common logic will need to be pulled out into the test client.

A follow up PR will move the SQL based implementation outside of LangChain.


libs/core/langchain_core/indexing/api.py
```diff
@@ -24,12 +24,14 @@
...
 
-from langchain_community.document_loaders.base import BaseLoader
+from langchain_core.document_loaders.base import BaseLoader
 from langchain_core.documents import Document
+from langchain_core.indexing.base import RecordManager
 from langchain_core.pydantic_v1 import root_validator
 from langchain_core.vectorstores import VectorStore
 
-from langchain.indexes.base import NAMESPACE_UUID, RecordManager
+NAMESPACE_UUID = uuid.UUID(int=1984)
```
efriis, line 33: i know this is just a migration but a comment of what this is doing would be great

libs/core/tests/unit_tests/indexing/test_hashed_document.py
efriis, line 1: might be worth keeping some form of the `langchain` ones here just to confirm we don't break anything for existing users for now (maybe with a note at top to remove test file for 0.2)

libs/langchain/tests/unit_tests/indexes/test_indexing.py
efriis, line 1: why are we removing these

efriis (approved):
implementation good - some questions about tests

libs/langchain/tests/unit_tests/indexes/test_indexing.py
eyurtsev: Good catch, I think I meant to remove some of the unit tests but not all of them (duplicated coverage with test_indexing in core). I can revert for now and just keep the duplicated code

PR #20673, opened 2024-04-19
docs[minor]: Show last update

bracesproul:


bracesproul:
Closing:
The docusaurus plugin will not work for any files which are converted notebooks, since those are auto-generated files & not tracked by git.

bracesproul:
Closing until a better solution can be found. Current issue: Non authenticated gh api requests hit rate limits pretty quick

Issue #20685, opened 2024-04-20
convert_to_openai_tool returns self as required param


hmnfalahi:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code
```
class MemberTool:
    def search_member(
            self,
            keyword: str,
            *args,
            **kwargs,
    ):
        """Search on members with any keyword like first_name, last_name, email

        Args:
            keyword: Any keyword of member
        """

        headers = dict(authorization=kwargs['token'])
        members = []
        try:
            members = request_(
                method='SEARCH',
                url=f'{service_url}/apiv1/members',
                headers=headers,
                json=dict(query=keyword),
            )

        except Exception as e:
            logger.info(e.__doc__)

        return members
```

### Description

I have MemberTool class that contains my tools.
if i run convert_to_openai_tool(MemberTool.search_member) i get this:
```
{'type': 'function', 'function': {'name': 'search_member', 'description': 'Search on members with any keyword like first_name, last_name, username, email', 'parameters': {'type': 'object', 'properties': {'keyword': {'type': 'string', 'description': 'Any keyword of member'}}, 'required': ['self', 'keyword']}}}
```

### System Info

pip freeze | grep langchain:
```
langchain==0.1.16
langchain-community==0.0.33
langchain-core==0.1.43
langchain-openai==0.1.3
langchain-text-splitters==0.0.1
```

PR #20691, opened 2024-04-20
Ignore self/cls from required args of class functions in convert_to_openai_tool

hmnfalahi:
Removed redundant self/cls from required args of class functions in _get_python_function_required_args:

```python
class MemberTool:
    def search_member(
            self,
            keyword: str,
            *args,
            **kwargs,
    ):
        """Search on members with any keyword like first_name, last_name, email

        Args:
            keyword: Any keyword of member
        """

        headers = dict(authorization=kwargs['token'])
        members = []
        try:
            members = request_(
                method='SEARCH',
                url=f'{service_url}/apiv1/members',
                headers=headers,
                json=dict(query=keyword),
            )

        except Exception as e:
            logger.info(e.__doc__)

        return members

convert_to_openai_tool(MemberTool.search_member)
```
expected result:
```
{'type': 'function', 'function': {'name': 'search_member', 'description': 'Search on members with any keyword like first_name, last_name, username, email', 'parameters': {'type': 'object', 'properties': {'keyword': {'type': 'string', 'description': 'Any keyword of member'}}, 'required': ['keyword']}}}
```

#20685


baskaryan (approved):

hmnfalahi:
Hi @baskaryan,

In your last changes, you used ``Dummy().dummy_function`` for testing. It should be ``Dummy.dummy_function`` because we don't get an instance of the class at this step. We obtain the class instance when it's necessary to call the function, not during the conversion step.
The tests failed, which is why.

baskaryan:
> > Hi @baskaryan,
> > In your last changes, you used `Dummy().dummy_function` for testing. It should be `Dummy.dummy_function` because we don't get an instance of the class at this step. We obtain the class instance when it's necessary to call the function, not during the conversion step. The tests failed, which is why.
> 
> good catch!

though i wonder if it'd be better to be able to handle methods from the instance as well as from the class

hmnfalahi:
> > > Hi @baskaryan,
> > > In your last changes, you used `Dummy().dummy_function` for testing. It should be `Dummy.dummy_function` because we don't get an instance of the class at this step. We obtain the class instance when it's necessary to call the function, not during the conversion step. The tests failed, which is why.
> > 
> > 
> > good catch!
> 
> though i wonder if it'd be better to be able to handle methods from the instance as well as from the class

Given that all functions of classes have self/cls as the first argument, we can do:

```
if '.' in function.__qualname__:
    required = required[1:]
```
Note that only class methods can have '.' in ``__qualname__``.

Issue #20514, opened 2024-04-16
Qdrant `from_documents` does not load existing collection


a172166:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain_community.vectorstores import Qdrant
from langchain_openai import AzureOpenAIEmbeddings

embeddings = AzureOpenAIEmbeddings(model="text-embedding-3-small",
                                   azure_endpoint="",
                                   api_key="")

qdrant = Qdrant.from_documents(
    docs,
    embeddings,
    path="local_qdrant",
    collection_name="my_documents",
)

### Description

When using `Qdrant.from_documents` to create a collection of documents. It seems that the collection is being recreated each time
```
from langchain_community.vectorstores import Qdrant

qdrant = Qdrant.from_documents(
    docs,
    embeddings,
    path="local_qdrant",
    collection_name="my_documents",
)
```

On  the first run, the embeddings are saving correctly, to a local file.
However, when running again, I had assumed the collection would be reused, but it seems that the collection is be recreated.

My collection is very large, and takes about 10 minutes to complete. If i run the add_documents call again, it still takes the same time to complete, so my assumtion is that it is not being read from the collection on disk

### System Info

langchain==0.1.16
langchain-community==0.0.32
langchain-core==0.1.42
langchain-openai==0.1.2
langchain-text-splitters==0.0.1
Platform: OSX
Python Version: 3.11

spike-spiegel-21:
The collection is not recreated but, every time you run, `Qdrant.from_documents(...)` it will re-create embeddings. If you want to reuse your collection, create a Qdrant client like this, in a separate script:

```
from qdrant_client import QdrantClient, models
client = QdrantClient(path="local_qdrant")
```
Ref: [Qdrant docs](https://qdrant.tech/documentation/)

spike-spiegel-21:
@a172166 more example

```from qdrant_client import QdrantClient, models
from langchain_community.vectorstores import Qdrant
from langchain_cohere import CohereEmbeddings

client = QdrantClient(path="local_qdrant")
embeddings = CohereEmbeddings(model="embed-english-light-v3.0", cohere_api_key = "<api_key>")
[...]

PR #20705, opened 2024-04-21
Added a function `from_existing_index` in `Qdrant` vector database.

spike-spiegel-21:
Issue: #20514 
The current implementation of `construct_instance` expects a `texts: List[str]` that will call the embedding function. This might not be needed when we already have a client with collection and `path,  you don't want to add any text.

This PR adds a class method that returns a qdrant instance with an existing client.

Here everytime https://github.com/langchain-ai/langchain/blob/cb6e5e56c29477c6da5824c17f1b70af11352685/libs/community/langchain_community/vectorstores/qdrant.py#L1592 `construct_instance` is called, this line sends some text for embedding generation.


spike-spiegel-21:
Please check if it is relevant to add this feature @hwchase17 @baskaryan

PR #20711, opened 2024-04-21
core(patch):fix partial_variables not working with SystemMessagePromptTemplate

### LINKED ISSUE #17560: partial_variables not working with ChatPromptTemplate (langchain v0.1.9)
### Checked other resources

### Example Code

```
   async def partial_test(self):
        prompt_template = "Tell me details about {name}. The output should be in {lang}"
        llm = LLMSelector(self.model).get_language_model()
            HumanMessagePromptTemplate.from_template(prompt_template, partial_variables={"lang": "Spanish"}),
        ])
        result = await chain.ainvoke({'name': 'Open AI'})
        print(result)
```

Errror: `Input to ChatPromptTemplate is missing variables {'lang'}.  Expected: ['lang', 'name'] Received: ['name']\"`

### Description

But this works with `PromptTemplate` also work with `ChatPromptTemplate` in previous version

### System Info

langchain v0.1.7

liugddx:
- **Issue:**  close #17560
- @baskaryan, @eyurtsev


hwchase17 (commented):
generally lgtm, can we test this when template is a LIST of templates?

i think then these partial variables will be passed to each template. is that what we want? seems a bit wasteful? we could just raise an error if you pass in list of templates and partial variables? or can keep as is

libs/core/langchain_core/prompts/chat.py
```diff
@@ -406,22 +406,27 @@ def get_lc_namespace(cls) -> List[str]:
...
         """
         if isinstance(template, str):
             prompt: Union[StringPromptTemplate, List] = PromptTemplate.from_template(
-                template, template_format=template_format
+                template,
+                template_format=template_format,
+                partial_variables=partial_variables,
             )
             return cls(prompt=prompt, **kwargs)
         elif isinstance(template, list):
```
eyurtsev, line 432: Could we verify this code path

liugddx: > Could we verify this code path

Sorry, I don't quite understand what you mean.

eyurtsev: I meant to add unit tests to that that the code path that belongs to the test case when template is a list

libs/core/langchain_core/prompts/chat.py
```diff
@@ -406,22 +406,27 @@ def get_lc_namespace(cls) -> List[str]:
     def from_template(
         cls: Type[_StringImageMessagePromptTemplateT],
         template: Union[str, List[Union[str, _TextTemplateParam, _ImageTemplateParam]]],
+        *,
```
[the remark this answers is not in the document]
liugddx: > This is a breaking change since template_format was a positional argument.
> 
> Let's move the `*` after the `template_format` variable, so it doesn't break any existing code.

thanks for the explanation. Done.

libs/core/langchain_core/prompts/chat.py
```diff
@@ -407,21 +407,26 @@ def from_template(
...
         """
         if isinstance(template, str):
             prompt: Union[StringPromptTemplate, List] = PromptTemplate.from_template(
-                template, template_format=template_format
+                template,
+                template_format=template_format,
+                partial_variables=partial_variables,
             )
             return cls(prompt=prompt, **kwargs)
         elif isinstance(template, list):
```
hwchase17, line 432: if you pass in a list of templates and partial variables, they will be silently ignored. imo we should not just silently ignore. we can (a) raise a nice error saying this isnt supported, (b) try to only pass relevant things to each template, (c) something else?

i think id prefer (a)? should also test this codepath more, whatever we decide

liugddx: > if you pass in a list of templates and partial variables, they will be silently ignored. imo we should not just silently ignore. we can (a) raise a nice error saying this isnt supported, (b) try to only pass relevant things to each template, (c) something else?
> 
> i think id prefer (a)? should also test this codepath more, whatever we decide

Sorry for the slow reply, I have solved it.

hwchase17 (approved):

PR #20725, opened 2024-04-22
community: Implement `bind_tools` for ChatTongyi

cheese-git:
## Description

Implement `bind_tools` in ChatTongyi. Usage example:

```py
from langchain_community.chat_models.tongyi import ChatTongyi

@tool
def multiply(first_int: int, second_int: int) -> int:
    """Multiply two integers together."""
    return first_int * second_int

llm = ChatTongyi(model="qwen-turbo")

llm_with_tools = llm.bind_tools([multiply])

msg = llm_with_tools.invoke("What's 5 times forty two")

print(msg)
```

Streaming is also supported.

## Dependencies

No Dependency is required for this change.


libs/community/langchain_community/chat_models/tongyi.py
```diff
@@ -88,8 +96,14 @@ def convert_dict_to_message(
                     )
         else:
             additional_kwargs = {}
+
         return (
-            AIMessageChunk(content=content)
+            AIMessageChunk(
+                content=content,
+                additional_kwargs=additional_kwargs,
+                tool_calls=tool_calls,
```
ccurme, line 104: are these tool calls fully formed in a streaming context?

if not you can specify `tool_call_chunks` on AIMessageChunk instead, with `args` a (partial json) string. See example here: https://github.com/langchain-ai/langchain/blob/c010ec8b71771dc3f54dc148475c90070b6a7c0b/libs/partners/openai/langchain_openai/chat_models/base.py#L264

libs/community/langchain_community/chat_models/tongyi.py
```diff
@@ -117,11 +148,25 @@ def convert_message_chunk_to_message(message_chunk: BaseMessageChunk) -> BaseMes
     if isinstance(message_chunk, HumanMessageChunk):
         return HumanMessage(content=message_chunk.content)
     elif isinstance(message_chunk, AIMessageChunk):
-        return AIMessage(content=message_chunk.content)
+        # assert message_chunk is None
+        return (
+            AIMessage(
+                content=message_chunk.content,
+                tool_calls=message_chunk.additional_kwargs["tool_calls"],
```
ccurme, line 155: this is confusing, because in `convert_dict_to_message` we are parsing tool calls (e.g., in call to `parse_tool_call`) but here we are passing them in although they are already parsed.

at this point, is the following true?
```python
if message_chunk.additional_kwargs["tool_calls"]:
    item = message_chunk.additional_kwargs["tool_calls"][0]

assert isinstance(item["args"], dict)  # not a string
```

libs/community/langchain_community/chat_models/tongyi.py
cheese-git: Thanks a lot, I have made some updates about this.

libs/community/langchain_community/chat_models/tongyi.py
cheese-git: Sorry for delayed response, having a really tough time with my job 😫 
This was a mistake, I've rewritten the method referring langchain_openai's `base` module.

cheese-git:
I also did some refactoring.

The main challenge of supporting tools for Tongyi, is that Tongyi does not support using tools with `incremental_output=True` yet, which means every new chunk contains the content of previous chunks. However, some internal methods of  Langchain assume the opposite.

That's why I end up writing a `subtract_client_response` method, to chop off duplicated content when streaming without `incremental_output=True`.

libs/community/langchain_community/chat_models/tongyi.py
```diff
@@ -256,11 +320,57 @@ def stream_completion_with_retry(self, **kwargs: Any) -> Any:
...
+                # we need to calculate the delta response manually
+                if _kwargs.get("stream") and not _kwargs.get(
+                    "incremental_output", False
+                ):
+                    if prev_resp is None:
+                        delta_resp = resp
+                    else:
+                        delta_resp = self.subtract_client_response(resp, prev_resp)
+                    prev_resp = resp
+                    yield check_response(delta_resp)
```
ccurme, line 336: @cheese-git would this represent a breaking change for streaming generally?

cheese-git: No, it would not.

Before this PR, `incremental_output` is always `True` when streaming:

https://github.com/langchain-ai/langchain/blob/64c47224a0d91b06e0ed04a635c03db4c6760b71/libs/community/langchain_community/chat_models/tongyi.py#L416-L417


To support tools when streaming, I updated the condition to [this](https://github.com/cheese-git/langchain/blob/336f3b852237119c9591e762535bc37095f1d949/libs/community/langchain_community/chat_models/tongyi.py#L326-L330):

```python
# According to the Tongyi official docs,
# `incremental_output` with `tools` is not supported yet
if params.get("stream") and not params.get("tools"):
    params["incremental_output"] = True
```

So "streaming without `incremental_output` being `True`" is a new circumstance.

Arunshmily:
> When i use ChatTongyi, bind_tools provide an error about "NotImplementedError (note: full exception trace is shown but execution is paused at: _run_module_as_main) , exception: no description". I am sure i have correct prompt and tools' description.

OK FINE, this method has not been implemented yet. ->langchain_core/language_models/chat_models.py"
langchain_core == 0.1.52

Arunshmily:
I Use ChatTongyi to bulid a Agent.
Which outputparser I could use?
ToolsAgentOutputParser() could not available to ChatTongyi.
LangChain-core==0.2.1

jlcoo:
`from langchain_core.tools import tool
from langchain_community.chat_models.tongyi import ChatTongyi

@tool
def multiply(first_int: int, second_int: int) -> int:
    """Multiply two integers together."""
    return first_int * second_int

llm = ChatTongyi(model="qwen-turbo")

llm_with_tools = llm.bind_tools([multiply])

msg = llm_with_tools.invoke("What's 5 times forty two")

print(msg)`
 
this code can't call multiply in langchain 0.2.6?
image

PR #20747, opened 2024-04-22
Use lstv2

hinthornw:


baskaryan (approved):

libs/core/langchain_core/tracers/context.py
baskaryan, line 1: any langchain docs we can/should update?

hinthornw: Maybe slowly! We'll keep this in for a long time, and it's less important in the langchain context

PR #20749, opened 2024-04-22
core[minor],community[minor],langchain[patch]: Move entity interface and implementations

eyurtsev:
* Move entity interface to langchain core
* Move entity implementations to community

----

The path for the interface in core is under `memory` rather than `memory.entity`. It seems like these are all base abstractions that somehow support memory use cases, so didn't want to create another namespace for it. 

I'm OK with the extra namespace for the implementations in community.


eyurtsev:
Going to likely handle in a separate refactor

PR #20750, opened 2024-04-22
core[minor],langchain[patch],community[patch]: Move storage interfaces to core

eyurtsev:
* Move storage interface to core
* Move in memory and file system implementation to core


libs/core/langchain_core/stores.py
```diff
@@ -123,3 +129,292 @@ async def ayield_keys(
 
 
 ByteStore = BaseStore[str, bytes]
+
+
+class LocalFileStore(ByteStore):
```
efriis, line 134: feel pretty strongly this shouldn't be in core. Maybe a `langchain-fs` package. Or just in community.

Feels too easy of a target for vulnerabilities.

libs/core/langchain_core/stores.py
```diff
@@ -123,3 +129,292 @@ async def ayield_keys(
...
+            Iterator[str]: An iterator over keys that match the given prefix.
+        """
+        prefix_path = self._get_full_path(prefix) if prefix else self.root_path
+        for file in prefix_path.rglob("*"):
+            if file.is_file():
+                relative_path = file.relative_to(self.root_path)
+                yield str(relative_path)
+
+
+class InMemoryBaseStore(BaseStore[str, V], Generic[V]):
```
efriis, line 288: this is fine in core too. But if we're doing in memory implementations of everything, could also put all these in a `langchain-inmemory` implementation package

eyurtsev: IMO core might make more sense since the in memory abstraction is useful for testing of core itself

libs/core/langchain_core/stores.py
eyurtsev: langchain-fs works

libs/core/langchain_core/stores.py
```diff
@@ -123,3 +126,138 @@ async def ayield_keys(
...
+
+    Attributes:
+        store (Dict[str, Any]): The underlying dictionary that stores
+            the key-value pairs.
+
+    Examples:
+
+        .. code-block:: python
+
+            from langchain.storage import InMemoryStore
```
baskaryan, line 142: core

baskaryan (approved):

PR #20753, opened 2024-04-22
community[patch],core[minor]: Move in memory cache implementation to core

eyurtsev:
This PR moves the InMemoryCache implementation from community to core.


baskaryan (approved):

baskaryan:
should update imports in docs as well

eyurtsev:
@baskaryan let's take care of docs separately -- we need a script to do the docs in one go so that it's a bit more efficient

PR #20756, opened 2024-04-22
infra: rm sep repo partner dirs

baskaryan:
so you can `poetry run pip install -e libs/partners/*/` to your hearts content


efriis (approved):
Also depending what ya wanna achieve might be better to use similar logic to api docs build, which also pulls in own-repo partner packages https://github.com/langchain-ai/langchain-api-docs-build/blob/master/.github/workflows/api_doc_build.yml

efriis:
And gonna have to update ^ to reflect the directories that don't need to be rm'd, as well as the pip install command

baskaryan:
> Also depending what ya wanna achieve might be better to use similar logic to api docs build, which also pulls in own-repo partner packages https://github.com/langchain-ai/langchain-api-docs-build/blob/master/.github/workflows/api_doc_build.yml

https://github.com/langchain-ai/langchain-api-docs-build/pull/6

PR #20759, opened 2024-04-22
langchain[major]: breaks some chains to remove hidden defaults

eyurtsev:
Breaks some chains in langchain to remove hidden chat model / llm
instantiation.


eyurtsev:

PR #20762, opened 2024-04-22
core, openai: support custom token encoders

ccurme:


libs/core/langchain_core/language_models/base.py
```diff
@@ -97,6 +98,10 @@ class BaseLanguageModel(
     """Tags to add to the run trace."""
     metadata: Optional[Dict[str, Any]] = Field(default=None, exclude=True)
     """Metadata to add to the run trace."""
+    token_encoder: Optional[Callable[[str], List[int]]] = Field(
```
baskaryan, line 101: should we call this something that mirror get_token_ids (eg custom_get_token_ids)

baskaryan (approved):

PR #20772, opened 2024-04-23
Add driver config param for neo4j graph

tomasonjo:


libs/community/langchain_community/graphs/neo4j_graph.py
```diff
@@ -173,6 +174,7 @@ def __init__(
         timeout: Optional[float] = None,
         sanitize: bool = False,
         refresh_schema: bool = True,
+        driver_config: Dict = {},
```
baskaryan, line 177: can't have mutable default values:) https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

baskaryan (approved):

PR #20795, opened 2024-04-23
langchain[patch],community[minor]: Move graph index creator

eyurtsev:
Move graph index creator to community


libs/langchain/langchain/indexes/prompts/__init__.py
```diff
@@ -1 +1,12 @@
 """Relevant prompts for constructing indexes."""
+from langchain_core._api import warn_deprecated
+
+warn_deprecated(
+    since="0.1.47",
+    message=(
+        "langchain.indexes.prompts will likely be removed in the future."
```
eyurtsev, line 7:
```suggestion
        "langchain.indexes.prompts will be removed in the future."
```

baskaryan (approved):

PR #20800, opened 2024-04-23
[draft]: propagate score through retriever

ccurme:


libs/core/langchain_core/vectorstores.py
```diff
@@ -690,7 +691,11 @@ def validate_search_type(cls, values: Dict) -> Dict:
...
 
     def _get_relevant_documents(
-        self, query: str, *, run_manager: CallbackManagerForRetrieverRun
+        self,
+        query: str,
+        *,
+        run_manager: CallbackManagerForRetrieverRun,
+        include_score: bool = False,
     ) -> List[Document]:
         if self.search_type == "similarity":
```
efriis, line 700: this case feels like it could lead to some confusion. Should we throw an error if `include_score` is set to true in this case with a message to switch search type?

Also do we want to use relevance scores for this, or do we want to just return raw scores? My goal was to deprecate the relevance score stuff as-is in #20302

PR #20815, opened 2024-04-24
partners: AI21 Labs Jamba Support

Josephasafg:
Description: Added support for AI21 new model - Jamba


baskaryan (approved):

efriis (approved):

libs/partners/ai21/tests/integration_tests/test_standard.py
```diff
@@ -20,24 +25,18 @@ def chat_model_params(self) -> dict:
             "model": "j2-ultra",
         }
 
-    @pytest.mark.xfail(reason="Emits AIMessage instead of AIMessageChunk.")
```
efriis, line 23: Hey @Josephasafg ! These 2 integration tests are failing for both model types. Would it be possible to fix these?

If it's too difficult, you can re-introduce the expected failures in both sets of integration tests.

efriis (changes requested):

libs/partners/ai21/tests/integration_tests/test_standard.py
Josephasafg: We don't currently support streaming (we will in a little bit). Once we do, i'll fix these

efriis (approved):

PR #20833, opened 2024-04-24
[community] add more data types support to ipex-llm llm integration

shane-huang:
- **Description**:  
    -  **add support for more data types**: by default `IpexLLM` will load the model in int4 format. This PR adds more data types support such as `sym_in5`, `sym_int8`, etc. Data formats like NF3, NF4, FP4 and FP8 are only supported on GPU and will be added in future PR. 
    - Fix a small issue in saving/loading, update api docs
- **Dependencies**: `ipex-llm` library
- **Document**: In `docs/docs/integrations/llms/ipex_llm.ipynb`, added instructions for saving/loading low-bit model.
- **Tests**: added new test cases to `libs/community/tests/integration_tests/llms/test_ipex_llm.py`, added config params. 
- **Contribution maintainer**: @shane-huang


libs/community/langchain_community/llms/bigdl_llm.py
```diff
@@ -22,6 +22,9 @@ class BigdlLLM(IpexLLM):
     def from_model_id(
         cls,
         model_id: str,
+        tokenizer_id: Optional[str] = None,
+        load_in_4bit: bool = True,
+        load_in_low_bit: Optional[str] = None,
```
baskaryan, line 27: this changes order of positional args which is a breaking change. could we do this instead (for all function signatures being updated in this pr)

```python
        cls,
        model_kwargs: Optional[dict] = None,
        *,
        tokenizer_id: Optional[str] = None,
        load_in_4bit: bool = True,
        load_in_low_bit: Optional[str] = None,
        **kwargs: Any,
```

shane-huang: updated.

baskaryan (approved):

PR #20845, opened 2024-04-24
langchain[patch],community[minor]: Migrate memory implementations to community

eyurtsev:
Migrates memory implementations to community


libs/community/langchain_community/memory/motorhead_memory.py
```diff
@@ -0,0 +1,100 @@
+from typing import Any, Dict, List, Optional
+
+import requests
+from langchain_core.messages import get_buffer_string
+
+try:
+    # Temporarily tuck import in a conditional import until
+    # community pkg becomes dependent on langchain core
+    from langchain_core.memory.chat_memory import BaseChatMemory
```
baskaryan, line 9: why the try/except if this is importing from core

eyurtsev: good catch it was langchain rather than langchain_core

eyurtsev:

PR #20849, opened 2024-04-24
core[minor], langchain[patch], community[patch]: mv StructuredQuery

baskaryan:
mv StructuredQuery to core


libs/core/langchain_core/structured_query.py
```diff
@@ -0,0 +1,143 @@
+"""Internal representation of a structured query language."""
```
eyurtsev, line 1: Sure about changing the namespaces?

`query_constructor.ir` seems reasonable for me

eyurtsev (approved):

libs/core/langchain_core/structured_query.py
jacoblee93: `ir` is meaningless to me

Issue #20600, opened 2024-04-18
`AzureSearch` vectorstore should be converted to `AzureSearchVectorStoreRetriever` when calling `as_retriever`


mspronesti:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```py
openai_api_version = "2024-02-01"

embeddings = AzureOpenAIEmbeddings(
    deployment=os.getenv('EMBEDDING_DEPLOYMENT_NAME'),
    openai_api_version=openai_api_version,
)

index_name: str = "my-index-name"

fields = [...]

vector_store: AzureSearch = AzureSearch(
    azure_search_endpoint=os.getenv('VECTOR_STORE_ADDRESS'),
    azure_search_key=os.getenv('VECTOR_STORE_PASSWORD'),
    index_name=index_name,
    embedding_function=embedding_function,
    fields=fields,
    # needed for semantic ranking
    semantic_configuration_name = 'my-config',
)

retriever = vector_store.as_retriever(search_kwargs={"k": 5}, search_type="semantic_hybrid")
```

### Description

I'm trying to use "semantic_hybrid" as `search_type` but it is not supported, as `as_retriever` doesn't return an instance of the class `AzureSearchVectorStoreRetriever` but a `VectorStoreRetriever`.

### System Info

-

chrislrobert:
If anybody comes across this fix and has a similar issue, this is my updated override code to essentially undo the fix:

```
class ExtendedAzureSearch(AzureSearch):
    """Extended AzureSearch class with support for similarity_score_threshold searches."""

    def as_retriever(self, **kwargs: Any) -> VectorStoreRetriever:
        """Return VectorStoreRetriever initialized from this VectorStore. Hard-wired to return base class version."""

        # (the latest version in AzureSearch triggers an invalid search_type error)
        return VectorStore.as_retriever(self, **kwargs)

    def _select_relevance_score_fn(self) -> Callable[[float], float]:
        return self._azure_relevance_score_fn

    @staticmethod
    def _azure_relevance_score_fn(score: float) -> float:
        """Returns Azure's similarity score as-is."""

[...]

Issue #9985, opened 2023-08-30
ImportError: cannot import name 'HnswVectorSearchAlgorithmConfiguration' from 'azure.search.documents.indexes.models'


kennywongck:
### System Info

I tried to use langchain with Azure Cognitive Search as vector store and got the following Import Error.

langchain version: 0.0.276
azure documents: 11.4.0b8
python version: 3.8

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [X] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

I tried to use the Azure Cognitive Search as vector store

```
    model: str = "text-embedding-ada-002"
    search_service = os.environ["AZURE_COGNITIVE_SEARCH_SERVICE_NAME"]
    search_api_key = os.environ["AZURE_COGNITIVE_SEARCH_API_KEY"]
    vector_store_address: str = f"https://{search_service}.search.windows.net"
    vector_store_password: str = search_api_key

    # define embedding model for calculating the embeddings
    model: str = "text-embedding-ada-002"
    embeddings: OpenAIEmbeddings = OpenAIEmbeddings(deployment=model, chunk_size=1)
    embedding_function = embeddings.embed_query

    # define schema of the json file stored on the index
    fields = [
            SimpleField(
                name="id",
                type=SearchFieldDataType.String,
                key=True,
                filterable=True,
            ),
            SearchableField(
                name="content",
                type=SearchFieldDataType.String,
                searchable=True,
            ),
            SearchField(
                name="content_vector",
                type=SearchFieldDataType.Collection(SearchFieldDataType.Single),
                searchable=True,
                vector_search_dimensions=len(embedding_function("Text")),
                vector_search_configuration="default",
            ),
            SearchableField(
                name="metadata",
                type=SearchFieldDataType.String,
                searchable=True,
            ),
        ]    
    
    vector_store: AzureSearch = AzureSearch(
        azure_search_endpoint=vector_store_address,
        azure_search_key=vector_store_password,
        index_name=index_name,
        embedding_function=embedding_function,
        fields=fields,
    )


```

[...]

PR #20852, opened 2024-04-24
feat(community): add support for similarity_score_threshold search in…

mspronesti:
See https://github.com/langchain-ai/langchain/issues/20600#issuecomment-2075569338 for details.

@chrislrobert

### DIFF
--- a/libs/community/langchain_community/vectorstores/azuresearch.py
+++ b/libs/community/langchain_community/vectorstores/azuresearch.py
@@ -479,7 +479,7 @@ class AzureSearch(VectorStore):
     def hybrid_search_with_score(
         self, query: str, k: int = 4, filters: Optional[str] = None
     ) -> List[Tuple[Document, float]]:
-        """Return docs most similar to query with an hybrid query.
+        """Return docs most similar to query with a hybrid query.
 
         Args:
             query: Text to look up documents similar to.
@@ -558,7 +558,7 @@ class AzureSearch(VectorStore):
     def semantic_hybrid_search_with_score_and_rerank(
         self, query: str, k: int = 4, filters: Optional[str] = None
     ) -> List[Tuple[Document, float, float]]:
-        """Return docs most similar to query with an hybrid query.
+        """Return docs most similar to query with a hybrid query.
 
         Args:
             query: Text to look up documents similar to.
@@ -702,7 +702,12 @@ class AzureSearchVectorStoreRetriever(BaseRetriever):
         if "search_type" in values:
             search_type = values["search_type"]
             if search_type not in (
-                allowed_search_types := ("similarity", "hybrid", "semantic_hybrid")
+                allowed_search_types := (
+                    "similarity",
+                    "similarity_score_threshold",
+                    "hybrid",
+                    "semantic_hybrid",
+                )
             ):
                 raise ValueError(
                     f"search_type of {search_type} not allowed. Valid values are: "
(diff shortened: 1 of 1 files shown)


baskaryan (approved):

chrislrobert:
@mspronesti Thanks for this. I think, though, that you still need `_select_relevance_score_fn()` and `similarity_search_with_score()` implementations in `AzureSearch` for this to work. I've been working happily with this version:

```
class ExtendedAzureSearch(AzureSearch):
    """Extended AzureSearch class with support for similarity_score_threshold searches."""

    def _select_relevance_score_fn(self) -> Callable[[float], float]:
        return self._azure_relevance_score_fn

    @staticmethod
    def _azure_relevance_score_fn(score: float) -> float:
        """Returns Azure's similarity score as-is."""

        # see https://learn.microsoft.com/en-us/azure/search/vector-search-ranking#scores-in-a-vector-search-results
        return score

    def similarity_search_with_score(
        self, *args: Any, **kwargs: Any
    ) -> List[Tuple[Document, float]]:
        """Run similarity search with distance."""

        # bridge into the AzureSearch class to run similarity search with distance
        return self.vector_search_with_score(*args, **kwargs)
```

It's been a while since I put that together and, unfortunately, I didn't keep good notes. I think that I based the implementation on the Pinecone class, plus the Azure Search scoring discussion at https://learn.microsoft.com/en-us/azure/search/vector-search-ranking#scores-in-a-vector-search-results. (Also had to contend with the LangChain-SDK incompatibility discussed at https://github.com/langchain-ai/langchain/issues/9985#issuecomment-1809631413.)

Finally, I don't know if it's fully okay for `AzureSearch.similarity_search()` to raise a not-allowed exception once the new search type is supported. (It's maybe okay if that function never gets called for that search type?)

mspronesti:
@chrislrobert Not sure I see your point here :/ This (supposed) patch relies on `similarity_search_with_relevance_scores` from `AzureSearch`, filtering according to the threshold. Why do you think this wouldn't work? 

Btw, I wanted to open this PR as a WIP, sorry for that. I hope that @baskaryan tested this before merging. It seemed to work fine for me.

chrislrobert:
Well, @mspronesti, I'm not sure just now, but:

1. There's no `similarity_search_with_score()` implementation in `AzureSearch`, and it's not clear to me when that gets called vs. `similarity_search_with_relevance_scores()`.
2. There's also no `_asimilarity_search_with_relevance_scores()` implementation in `AzureSearch`, and the base class version requires `_select_relevance_score_fn()`. I use async calls for my overall chain, so even if the search part ended up synchronous I'd want the rest to stay async.

So I think it's just a matter of which contexts will or won't be supported. I don't think `similarity_search_with_relevance_scores()` happens to cover my `ConversationalRetrievalChain` use case, but I believe you that there are at least a nonzero number of workflows that would be supported by this change here.

chrislrobert:
@mspronesti I don't know, I'd have to test it end-to-end. I didn't add that stuff randomly, but as a result of trial-and-error, as I tried to get the `ConversationalRetrievalChain` working with Azure. I had an old implementation with Pinecone that was working fine, there were a number of examples out there, but when I tried to swap in the `AzureSearch` implementation I ended up with a bunch of errors to resolve. The thing is that I'm not directly calling the retriever at all: it's being used as a retriever called by other chains. To the extent that it fails to implement various endpoints, it trips up those chains. (Perhaps also, in the past few months, some of those issues themselves have gone away, I just don't know.)

And FWIW, I agree with you: it looks like true `async` support is needed, and my hack is just working around the issue there. I honestly don't understand how so many LangChain bits have been able to stay synchronous-only for so long. async support seems pretty fundamental for production use cases.

mspronesti:
@chrislrobert I will have a look at this in the following days -- unless someone gets ahead of me :-) -- and I will tag you on the PR if I were to find a solution.

chrislrobert:
@mspronesti, the latest version (0.2.10) seems to work now, following a bunch of work to add async support (like, namely, https://github.com/langchain-ai/langchain/pull/22075). Unfortunately, there's a new regression in issue #24740. But things are veeery close to fully working now in the base `AzureSearch` implementation. (You can see in my `ExtendedAzureSearch` class, in #24740, that I now only need to override to fix the regression, not for the `similarity_score_threshold` support.)

PR #20856, opened 2024-04-24
community[minor]: Add indexing via locality sensitive hashing to the Yellowbrick vector store

markcusack:
- **Description:** Add LSH-based indexing to the Yellowbrick vector store module
- **Twitter handle:** @markcusack


libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -24,19 +27,26 @@
...
     """Yellowbrick as a vector database.
-
     Example:
         .. code-block:: python
             from langchain_community.vectorstores import Yellowbrick
             from langchain_community.embeddings.openai import OpenAIEmbeddings
             ...
     """
 
+    LSH_INDEX_TABLE: str = "_lsh_index"
```
eyurtsev, line 37: Could we remove class level attributes? You can move them into global scope or tuck them into the __init__

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -24,19 +27,26 @@
...
+    LSH_HYPERPLANE_TABLE: str = "_lsh_hyperplane"
+    CONTENT_TABLE: str = "_content"
+
     def __init__(
         self,
         embedding: Embeddings,
         connection_string: str,
         table: str,
+        use_lsh: Optional[bool] = False,
```
eyurtsev, line 46:
```suggestion
        *,
        use_lsh: Optional[bool] = False,
```

eyurtsev, line 46: There are different ways to implement nearest neighbor searches.

even lsh can be parameterized in different ways. Is it possible to parameterize in a better way than a boolean?

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -46,67 +56,144 @@ def __init__(
...
+                    self._table + self.CONTENT_TABLE + "_pk_doc_id_embedding_id"
+                ),
+                c2=sql.Identifier(self._table + self.CONTENT_TABLE + "_fk_doc_id"),
+            )
+        )
+        self._get_connection().commit()
         cursor.close()
 
-    def drop(self, table: str) -> None:
+    def _drop_table(self, table: str) -> None:
```
eyurtsev, line 174: this is a breaking change in the API

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -46,67 +56,144 @@ def __init__(
...
+
+    @property
+    def hamming_distance(self) -> bool:
+        return self._hamming_distance
+
+    @hamming_distance.setter
+    def hamming_distance(self, hamming_distance: int) -> None:
+        self._hamming_distance = hamming_distance
+
+    def _get_connection(self):
```
eyurtsev, line 111: I think it would be useful to introduce a new method that provides a context manager for the cursor so that cursor can be closed properly on exception

```
from contextlib import contextmanager

@contetmanager
def get_cursor():
   with connection... as conn: # <-- Can do this or the logic that re-uses the connection etc
      with conn.cusor() as cursor: 
        yield cursor
```

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -24,19 +27,26 @@
...
     """
 
+    LSH_INDEX_TABLE: str = "_lsh_index"
+    LSH_HYPERPLANE_TABLE: str = "_lsh_hyperplane"
+    CONTENT_TABLE: str = "_content"
+
     def __init__(
         self,
         embedding: Embeddings,
         connection_string: str,
```
eyurtsev, line 44: (not required for PR) init should be expanded beyond a connection string, so folks can pass either an existing connection or a connection pool

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -124,51 +211,69 @@ def add_texts(
...
-                            for _ in range(len(embeddings[id]))
-                        ]
-                    )
-                ),
-            )
-            cursor.execute(insert_query, flattened_input)
-            self._connection.commit()
-        return results
+
+        conn = self._get_connection()
```
eyurtsev, line 222: Why the change to remove `INSERT INTO` behavior?

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -46,67 +56,144 @@ def __init__(
...
-                "CREATE TABLE IF NOT EXISTS {} ( \
-            id UUID, \
-            embedding_id INTEGER, \
-            text VARCHAR(60000), \
-            metadata VARCHAR(1024), \
-            embedding FLOAT)"
-            ).format(sql.Identifier(self._table))
+                """
+                CREATE TABLE IF NOT EXISTS {t} (
+                doc_id UUID NOT NULL,
```
eyurtsev, line 140: this is a breaking change in the schema? Is there a migration for users to use?

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: It's significantly faster to use COPY to insert data than SQL INSERT.

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Yes, I've moved into init

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Done

markcusack: Great feedback. I've generalized the way of setting up indexes.

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Good suggestion. I've implemented a context manager.

eyurtsev:
@markcusack tag me or @baskaryan when you're ready for a re-review

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Yep, added this. Thanks!

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Good point. I've introduced a new function to migrate from the prior schema to the updated schema.

markcusack:
@eyurtsev @baskaryan -- I've made the recommended changes. Really appreciate the input. Ready for review.

markcusack:
@eyurtsev @baskaryan  I've improved the transaction, connection and cursor handling. I've also added support for the delete API. I'm not planning to make any extra additions to functionality, unless you suggest otherwise. Thanks!

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -1,204 +1,523 @@
...
+            self.index_type = index_type
+            self.params = params or {}
+
+        def get_param(self, key: str, default: Any = None) -> Any:
+            return self.params.get(key, default)
+
     def __init__(
         self,
         embedding: Embeddings,
-        connection_string: str,
```
baskaryan, line 38: this is a breaking change — are we sure we want to make it?

markcusack: I've decided to remove the breaking change here and leave adding the ability to pass in an existing connection to a later PR.

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -19,157 +26,374 @@
...
     def __init__(
         self,
         embedding: Embeddings,
         connection_string: str,
         table: str,
+        *,
+        schema: Optional[str] = None,
+        logger: Optional[logging.Logger] = None,
+        seed: Optional[float] = 0.42,
+        drop: Optional[bool] = False,
```
baskaryan, line 74: nit these probably dont need to be typed as Optional

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -178,27 +402,104 @@ def from_texts(
...
+        schema: Optional[str] = None,
+        drop: Optional[bool] = False,
         **kwargs: Any,
     ) -> Yellowbrick:
         """Add texts to the vectorstore index.
         Args:
             texts: Iterable of strings to add to the vectorstore.
             metadatas: Optional list of metadatas associated with the texts.
-            connection_string: URI to Yellowbrick instance
+            connection_opt: URI to Yellowbrick instance or a pg connection
```
baskaryan, line 413: undo

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -19,157 +26,374 @@
...
+        schema: Optional[str] = None,
+        logger: Optional[logging.Logger] = None,
+        seed: Optional[float] = 0.42,
+        drop: Optional[bool] = False,
     ) -> None:
         """Initialize with yellowbrick client.
         Args:
             embedding: Embedding operator
             connection_string: Format 'postgres://username:password@host:port/database'
+            or pg connection
```
baskaryan, line 80: delete?

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -178,27 +402,104 @@ def from_texts(
         metadatas: Optional[List[dict]] = None,
         connection_string: str = "",
         table: str = "langchain",
+        schema: Optional[str] = None,
+        drop: Optional[bool] = False,
```
baskaryan, line 406: could just be typed as `bool`

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -178,27 +402,104 @@ def from_texts(
...
+    def delete(self, ids: Optional[List[str]] = None, **kwargs: Any) -> Optional[bool]:
+        """Delete vectors by uuids.
+
+        Args:
+            ids: List of ids to delete, where each id is a uuid string.
+        """
+        from psycopg2 import sql
+
+        if not ids:
+            return None
```
baskaryan, line 437: out of curiosity why None instead of False

baskaryan (approved):

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Makes sense. Changed. Thanks!

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Done

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: I've taken the opportunity to expand the behavior to allow delete_all as well as delete by id. Returns None now. Actually, to your comment, return None is a nod to my DB background: delete successful, failed, or didn't happen :)

libs/community/langchain_community/vectorstores/yellowbrick.py
markcusack: Done

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -19,157 +25,368 @@
...
+        def get_param(self, key: str, default: Any = None) -> Any:
+            return self.params.get(key, default)
+
     def __init__(
         self,
         embedding: Embeddings,
         connection_string: str,
         table: str,
+        *,
+        schema: Optional[str] = None,
```
eyurtsev, line 70: undocumented parameters

libs/community/langchain_community/vectorstores/yellowbrick.py
```diff
@@ -19,157 +25,368 @@
...
+                sql.SQL(
+                    """
+                    CREATE SCHEMA IF NOT EXISTS {s}
+                """
+                ).format(
+                    s=sql.Identifier(self._schema),
+                )
+            )
 
-    def create_table_if_not_exists(self) -> None:
```
eyurtsev, line 71: fyi this is a breaking change since it was a public method

eyurtsev (approved):

eyurtsev:
Left a few comment -- feel free to address in a separate PR if you think they're important. Merging as soon as tests pass.

PR #20862, opened 2024-04-24
support messages in messages out

hwchase17:


baskaryan (approved):

libs/core/langchain_core/runnables/history.py
```diff
@@ -364,10 +368,19 @@ def get_input_schema(
             return super_schema
 
     def _get_input_messages(
-        self, input_val: Union[str, BaseMessage, Sequence[BaseMessage]]
+        self, input_val: Union[str, BaseMessage, Sequence[BaseMessage], dict]
```
eyurtsev, line 371: you can use TypedDict here if you care -- it's extra code, but more precise in terms of what's expected

```
class InputDict(TypedDict):
  input: Sequence[BaseMessage] ? 
```

eyurtsev (approved):

PR #20889, opened 2024-04-25
langchain: Allow passsing run_id from config when invoking the chain

HoangNguyen689:
- **Description:** When passing run_id to a chain invoke, it has no effects. I want to use the pre-defined run_id when invoking the agent executor.
- **Issue:** none
- **Dependencies:** none
- **Twitter handle:** none


libs/langchain/langchain/chains/base.py
```diff
@@ -127,6 +128,7 @@ def invoke(
         tags = config.get("tags")
         metadata = config.get("metadata")
         run_name = config.get("run_name") or self.get_name()
+        run_id = config.get("run_id")
```
hinthornw, line 131: We should do the same in `ainvoke` !

hinthornw: Before approving, we need:
1. add support in `ainvoke`, `stream`, `astream`
2. Add unit tests for each method to confirm they work

HoangNguyen689: @hinthornw 
Thank you. 
I added the run_id in the `ainvoke`, `stream` and `astream` here https://github.com/langchain-ai/langchain/pull/20889/commits/2386d31382282dd1e28615ea3cf8ce5b6d1a7cf3
I also took a look at unit tests, for example: 
- https://github.com/langchain-ai/langchain/blob/master/libs/langchain/tests/unit_tests/chains/test_base.py
- https://github.com/langchain-ai/langchain/blob/master/libs/langchain/tests/unit_tests/agents/test_agent.py
- https://github.com/langchain-ai/langchain/blob/master/libs/langchain/tests/unit_tests/agents/test_agent_iterator.py
But I don't see the relative test about config attributes like `run_name`. Could you guide me on the test?
Our project needs this feature now, so I would be very happy if you could help us soon.

HoangNguyen689:
@hinthornw cc: @baskaryan 
Thanks for your comments.
Regarding the async function, I added the code to it.

I want to discuss the unit tests.
This pull request doesn't implement the new things.
The single `llm.invoke` still works well because it is already defined here:
- https://github.com/langchain-ai/langchain/blob/b8db73233c0e01e283115b0f5b069aec25bf1384/libs/core/langchain_core/tools.py#L265

I pass the `run_id` from config to the callback manager in the chain, so I believe that the unit test about the `run_id` in the callback manager is already covered, and all I need is to pass the test cases.
The test cases are written here:
- https://github.com/langchain-ai/langchain/blob/8c71f92cb2f3cca21f2ca02657c8f2906d6a91d7/libs/core/tests/unit_tests/tracers/test_base_tracer.py#L160

Correct me if I am wrong. Thank you.

HoangNguyen689:
@hinthornw 
Sorry for mentioning you multiple times.
I fixed the lint error. Could you approve the CI workflow?

HoangNguyen689:
@hinthornw cc: @baskaryan 
Please let me know if you ignore this PR or plan to support setting up the specific run_id for agent invoking in another PR.
I want to set the run_id to set up feedback for Langsmith manually. Without it, the process can not go forward.

libs/langchain/langchain/chains/base.py
HoangNguyen689: @hinthornw 
https://github.com/langchain-ai/langchain/pull/20889#issuecomment-2084768994
Could you read this comment? This problem is blocking our process so I would appreciate your help.

hinthornw:
Sorry about it!

libs/langchain/langchain/chains/base.py
hinthornw: Sorry about this! notifications have been inundated from other things. Will review now

PR #20907, opened 2024-04-25
community[patch]: support hybrid search with threshold in Azure AI Search Retriever

mspronesti:
Support hybrid search with a score threshold -- similar to what we do for similarity search.


mspronesti:
@baskaryan

baskaryan (approved):

PR #20911, opened 2024-04-25
docs: top-menu draft

leo-gan:
This is a draft of the higher-level documentation menu.
It is based on the previous discussion about the reading roles.

Top-menu:

`Start`  / `Learn`  / `Develop` / `Integrate`  /  `API`  /  `More`

Menu based on the 3 major reading roles:
- "1st-timer": 1st time reader. It is a reader who wants the bare minimum information to start and try the LC (for different reasons). 
- "Starter": A developer that want to start to develop with LC. 
- "DevUser": A developer that works with LC regularly.

1st-timer:
- starts with "Start"
- wants everything to start immediately, as fast as it can. No details needed.

Starter:
- start with "Learn"
- Starter has also finished "Start" section
- follow the sequential learning path. Some parts could be skipped if they are known.

DevUser:
- uses "Develop", "Integrations", and "API"
- DevUser also finished with "Learn" section.
- reads only necessary documents
- frequently uses "Search"

NOTE: only top-menu and navbar are drafted. Links to the individual documents are placeholders.

CC @baskaryan @efriis @eyurtsev


baskaryan:
there's a big docs refactor already being worked on here https://github.com/langchain-ai/langchain/pull/20363

sepiatone:
@leo-gan imo, It may be better to think of two roles - 'starter' and 'dev_user'. the 'learner' moves to the 'starter' role after doing the installation and the initial quick start section.

Additionally the 'tutorials' could map to what the 'starter' needs and the 'how-to guides' to what the 'dev_user' is looking for. so it's clear to someone looking through the documentation where they need to go for a specific topic.

PR #20930, opened 2024-04-26
Update python.py（experimental：Added code for PythonREPL）

SheldonLiu0412:
Update python.py（experimental：Added code for PythonREPL）

Added code for PythonREPL, defining a static method 'sanitize_input' that takes the string 'query' as input and returns a sanitizing string. The purpose of this method is to remove unwanted characters from the input string, Specifically:

1. Delete the whitespace at the beginning and end of the string (' \s').
2. Remove the quotation marks (`` ` ``) at the beginning and end of the string.
3. Remove the keyword "python" at the beginning of the string (case insensitive) because the user may have typed it.

This method uses regular expressions (regex) to implement sanitizing.

It all started with this code：
from langchain.agents import Tool
from langchain_experimental.utilities import PythonREPL

python_repl = PythonREPL()
repl_tool = Tool(
    name="python_repl",
    description="Remove redundant formatting marks at the beginning and end of source code from input.Use a Python shell to execute python commands. If you want to see the output of a value, you should print it out with `print(...)`.",
    func=python_repl.run,
)

When I call the agent to write a piece of code for me and execute it with the defined code, I must get an error: SyntaxError('invalid syntax', ('<string>', 1, 1,'In', 1, 2))

After checking, I found that pythonREPL has less formatting of input code than the soon-to-be deprecated pythonREPL tool, so I added this step to it, so that no matter what code I ask the agent to write for me, it can be executed smoothly and get the output result.
I have tried modifying the prompt words to solve this problem before, but it did not work, and by adding a simple format check, the problem is well resolved.
image


libs/experimental/langchain_experimental/utilities/python.py
```diff
@@ -22,6 +23,22 @@ class PythonREPL(BaseModel):
     globals: Optional[Dict] = Field(default_factory=dict, alias="_globals")
     locals: Optional[Dict] = Field(default_factory=dict, alias="_locals")
 
+    @staticmethod
+    def sanitize_input(query: str) -> str:
```
baskaryan, line 27: could we add some unit tests for this

SheldonLiu0412: > could we add some unit tests for this

Hello, I have added a test example for it (the local test passed), but I keep reporting errors when I format and detect the local code
ModuleNotFoundError: No module named 'rapidfuzz'
make: *** [format_diff] Error 1
This is a problem caused by my own environment and has nothing to do with the committed code, but I don't know how to fix it @baskaryan

baskaryan (approved):

PR #20936, opened 2024-04-26
community: improve error handling for missing choices and API errors

ssocean:
### PR Title:
community: improve error handling for missing choices and API errors

### PR Message:
**Description:** This update introduces clearer error handling in the response processing logic. Specifically, it adds explicit exceptions for cases where the API response lacks the `choices` field or contains error messages, enhancing the debuggability of the integration. 

For instance, if there is an issue with the API response (possibly due to excessively long tokens, insufficient balance, or service unavailability, etc.), it previously triggered a "TypeError: 'NoneType' object is not iterable" error. Now, it generates a ValueError stating "This model's maximum context length is 16385 tokens. However, your messages resulted in 19223 tokens. Please reduce the length of the messages. (request id: xxx)."

**Issue:** N/A

**Dependencies:** None

**Twitter handle:** N/A


baskaryan:
this is a deprecated implementation, main implementation now lives in libs/partners/openai/langchain_openai/chat_models/base.py. can we make all changes there?

ssocean:
> this is a deprecated implementation, main implementation now lives in libs/partners/openai/langchain_openai/chat_models/base.py. can we make all changes there?

Thank you for pointing out the correct location for the implementation. I apologize for the oversight and for making changes in a deprecated file. I have now applied the necessary modifications to libs/partners/openai/langchain_openai/chat_models/base.py as suggested. Please let me know if there are any further adjustments needed~

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -589,9 +589,11 @@ def _create_chat_result(
...
         # will throw a "TypeError: 'NoneType' object is not iterable" error
         # to mask the true error. Because 'response["choices"]' is None.
-        if response.get("error"):
-            raise ValueError(response.get("error"))
-
+        if 'error' in response:
+            error_message = response['error']['message']
+            raise ValueError(error_message)
+        if 'choices' not in response or response['choices'] is None:
+            raise ValueError("No choices available in the response, please check the input and model configuration.")
```
baskaryan, line 596: why is raising an error here better than returning an empty list of Generations?

ssocean: Indeed, your point is well-taken.  But, the absence of 'choices' might not only be due to user errors but also service interruptions. Returning an empty list may potentially lead to errors in downstream processes, e.g. some users might take an empty list as an empty string. I think it's generally better to alert users directly about the problem so they can troubleshoot effectively. This ensures clarity and helps maintain system integrity.

baskaryan: if no choices were returned due to a server error wouldn't there be 'error' key in the response (which is already checked)

PR #20951, opened 2024-04-26
langchain: enhance `LocalFileStore` to offer `update_atime` parameter that updates access times on read

chrispy-snps:
**Description:**
The `LocalFileStore` class can be used to create an on-disk `CacheBackedEmbeddings` cache. The number of files in these embeddings caches can grow to be quite large over time (hundreds of thousands) as embeddings are computed for new versions of content, but the embeddings for old/deprecated content are not removed.

A *least-recently-used* (LRU) cache policy could be applied to the `LocalFileStore` directory to delete cache entries that have not been referenced for some time:

```bash
# delete files that have not been accessed in the last 90 days
find embeddings_cache_dir/ -atime 90 -print0 | xargs -0 rm
```

However, most filesystems in enterprise environments disable access time modification on read to improve performance. As a result, the access times of these cache entry files are not updated when their values are read.

To resolve this, this pull request updates the `LocalFileStore` constructor to offer an `update_atime` parameter that causes access times to be updated when a cache entry is read.

For example,

```python
file_store = LocalFileStore(temp_dir, update_atime=True)
```

The default is `False`, which retains the original behavior.

**Testing:**
I updated the LocalFileStore unit tests to test the access time update.

### DIFF
--- a/libs/langchain/langchain/storage/file_system.py
+++ b/libs/langchain/langchain/storage/file_system.py
@@ -1,5 +1,6 @@
 import os
 import re
+import time
 from pathlib import Path
 from typing import Iterator, List, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


eyurtsev (approved):

eyurtsev:
Thank you @chrispy-snps

PR #20956, opened 2024-04-26
langchain[patch]: add async methods to ConversationSummaryBufferMemory

meikethomas:
Added asynchronously callable methods according to the ConversationSummaryBufferMemory API documentation.


libs/langchain/langchain/memory/summary_buffer.py
```diff
@@ -18,6 +18,35 @@ class ConversationSummaryBufferMemory(BaseChatMemory, SummarizerMixin):
     def buffer(self) -> List[BaseMessage]:
         return self.chat_memory.messages
 
+    async def abuffer(self) -> Any:
```
baskaryan, line 21: should this match typing/functionality of buffer

ccurme (approved):

PR #20971, opened 2024-04-27
community: DuckDB VS - expose similarity, improve performance of from_texts

### LINKED ISSUE #20969: DuckDB: distance/similarity property not reported to documents returned by similarity_search
### Checked other resources

### Example Code

```python
from dotenv import load_dotenv
import duckdb
from langchain_openai import OpenAIEmbeddings
from langchain_community.vectorstores import DuckDB
from time import time

load_dotenv()
TABLE_NAME = "embeddings"
documents = [Document(page_content="Jacek is the best software engineer in the world", metadata={"id": "1"})]

db_conn = duckdb.connect('./test.DUCKDB')

try:
    start_exists = time()
    print("Checking table exists")
    table = db_conn.table(TABLE_NAME)
    table.show()
    vector_store = DuckDB(connection=db_conn, table_name=TABLE_NAME, embedding=OpenAIEmbeddings(), vector_key="embedding")
    print(f"Table exists check took {time() - start_exists} seconds")
except Exception as e:
    start_not_exists = time()
    print(f"Table does not exist, create it from documents")
    vector_store = DuckDB.from_documents(documents, connection=db_conn, table_name=TABLE_NAME, embedding=OpenAIEmbeddings(), vector_key="embedding")
    print(f"Table does not exist, took {time() - start_not_exists} seconds")

start_search = time()
query = "Who is the best software engineer in the world?"
docs = vector_store.similarity_search(query)
print(f"Search result: {docs}")
print(f"Search took {time() - start_search} seconds")
```

_No response_

### Description

I use DuckDB as vector store.
When I execute similarity_search, I expect `distance` property is returned as a (metadata) part of result documents.
I discussed this issue in DuckDB community and we agreed that it is a bug and it should be returned.
I am going to fix it.

### System Info

- Ubuntu 22.04
- python 3.10.12
- langchain==0.1.16
- openai==1.13.3
- langchain_openai==0.1.1
- duckdb==0.10.2

jaceksan:
3 fixes of DuckDB vector store:
- unify defaults in constructor and from_texts (users no longer have to specify `vector_key`).
- include search similarity into output metadata (fixes #20969)
- significantly improve performance of `from_documents`

Dependencies: added Pandas to speed up `from_documents`.
I was thinking about CSV and JSON options, but I expect trouble loading JSON values this way and also CSV and JSON options require storing data to disk.
Anyway, the poetry file for langchain-community already contains a dependency on Pandas.


libs/community/langchain_community/vectorstores/duckdb.py
```diff
@@ -153,14 +161,21 @@ def add_texts(
...
-                [ids[idx], text, embedding, metadata],
-            )
+            data.append({
+                self._id_key: ids[idx],
+                self._text_key: text,
+                self._vector_key: embedding,
+                "metadata": metadata,
+            })
+        df = pd.DataFrame.from_dict(data)
+        print(df)
```
hkad98, line 171: Forgotten print.

libs/community/langchain_community/vectorstores/duckdb.py
```diff
@@ -181,20 +196,20 @@ def similarity_search(
...
             .fetchdf()
         )
         return [
             Document(
                 page_content=docs[self._text_key][idx],
-                metadata=json.loads(docs["metadata"][idx])
+                metadata={
+                    **json.loads(docs["metadata"][idx]),
+                    **{f"_{SIMILARITY_ALIAS}": docs[SIMILARITY_ALIAS][idx]}
+                }
```
hkad98, line 212: This could be simplified to

```python
metadata={f"_{SIMILARITY_ALIAS}": docs[SIMILARITY_ALIAS][idx], **json.loads(docs["metadata"][idx])}
```

libs/community/langchain_community/vectorstores/duckdb.py
jaceksan: removed

libs/community/langchain_community/vectorstores/duckdb.py
jaceksan: simplified as suggested

libs/community/langchain_community/vectorstores/duckdb.py
```diff
@@ -72,14 +79,14 @@ class DuckDB(VectorStore):
...
     def __init__(
-        self,
-        *,
-        connection: Optional[Any] = None,
-        embedding: Embeddings,
-        vector_key: str = "embedding",
-        id_key: str = "id",
-        text_key: str = "text",
-        table_name: str = "vectorstore",
+            self,
```
leo-gan, line 82: I'm not sure about this new ident formatting. Did you run [`make format`](https://python.langchain.com/docs/contributing/code/#formatting-and-linting) for changes?

jaceksan: I executed `make format` and it made 423 changes in many files :-)
I committed only the change in duckdb.py and rolled back the rest.

jaceksan:
@baskaryan @eyurtsev should I try to fix the failing test? 
To be honest, I don't know why it started to fail. The import in the notebook is identical to what I use in my code base...
Any suggestion would be appreciated.

libs/community/langchain_community/vectorstores/duckdb.py
```diff
@@ -153,10 +168,19 @@ def add_texts(
...
+                    self._id_key: ids[idx],
+                    self._text_key: text,
+                    self._vector_key: embedding,
+                    "metadata": metadata,
+                }
             )
+        # noinspection PyUnusedLocal
+        df = pd.DataFrame.from_dict(data)  # noqa: F841
+        self._connection.execute(
+            f"INSERT INTO {self._table_name} SELECT * FROM df",
```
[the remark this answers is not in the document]
jaceksan: Hey, 
I am afraid we need it.
DuckDB does not provide any other way to import data efficiently.
Respectively, they provide import from file(CSV, PARQUET, JSON), but it would require writing the file to disk, which is IMO risky (which folder? privileges? ...).
If the new dependency becomes a blocker, I can implement a file approach, but then I would appreciate guidance on how to do it properly.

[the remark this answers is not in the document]
baskaryan: what if we try imprting pandas and if its available we do the updated approach, and if its not available we fallback to the current approach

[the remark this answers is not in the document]
jaceksan: I tried to do it.
But when I tested it end-to-end, it failed here:
https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/duckdb.py#L192
The similarity_search() method uses `fetchdf()` function which requires Pandas.
So the hard dependency was there before.
 
Do you want me to update the similarity_search() method too?

I noticed that you added a new code raising an exception if the import of Pandas fails.
Can we keep the current code in the PR and merge it?

libs/community/langchain_community/vectorstores/duckdb.py
```diff
@@ -76,10 +82,10 @@ def __init__(
...
         connection: Optional[Any] = None,
         embedding: Embeddings,
-        vector_key: str = "embedding",
-        id_key: str = "id",
-        text_key: str = "text",
-        table_name: str = "vectorstore",
+        vector_key: str = DEFAULT_VECTOR_KEY,
+        id_key: str = DEFAULT_ID_KEY,
+        text_key: str = DEFAULT_TEXT_KEY,
+        table_name: str = DEFAULT_TABLE_NAME,
```
hwchase17, line 88: this is a change from before - used to be `vectorstore`, now is `embeddings`, is that intentional?

jaceksan: Well, yes.
They were not unified and it caused issues - you created the table with one name and tried to search in the table using the second name.

libs/community/langchain_community/vectorstores/duckdb.py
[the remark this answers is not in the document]
jaceksan: I suggest to keep it as it is. 
If we need to remove the dependency, let's create a new issue and solve it separately.

jaceksan:
@hwchase17 can we merge it now?

libs/community/langchain_community/vectorstores/duckdb.py
```diff
@@ -2,13 +2,23 @@
...
 from langchain_core.embeddings import Embeddings
 from langchain_core.vectorstores import VST, VectorStore
 
+logger = logging.getLogger(__name__)
+
+DEFAULT_VECTOR_KEY = "embedding"
+DEFAULT_ID_KEY = "id"
+DEFAULT_TEXT_KEY = "text"
+DEFAULT_TABLE_NAME = "embeddings"
+SIMILARITY_ALIAS = "similarity"
```
ccurme, line 20: (nit)
```suggestion
SIMILARITY_ALIAS = "similarity_score"
```

baskaryan (approved):

PR #20975, opened 2024-04-27
community[mini]: Add stop_reason and stop_sequence to Bedrock Anthropic model output

Neet-Nestor:
**Description:**
Currently, the output of Amazon Bedrock llms will ignore the `stop_reason` and `stop_sequence` fields in the output of Anthropic models, making it hard for users to rely on these fields to build agents.

Description of these two fields are described here:
https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html

This PR adds these two fields in the "metadata" output field  if Bedrock returns them in the response so the downstream users could read and use the accordingly.

**Dependencies:**
n/a


Neet-Nestor:
Seems this approach is not feasible as the llm model eventually will only return "text" back in `_call`

PR #20987, opened 2024-04-28
[add] integrate Semantic text splitter for better performance

rgupta2508:
text_splitter: Added Semantic text splitter (https://docs.rs/text-splitter/0.12.1/text_splitter/) for better performance and more option of splitter to user
Adding SemanticCharacterTextSplitter, SemanticTiktokenTextSplitter


libs/community/langchain_community/text_splitter/semantic_text_splitter.py
```diff
@@ -0,0 +1,64 @@
...
+            from semantic_text_splitter import TiktokenTextSplitter
+        except ImportError as e:
+            raise ImportError(
+                "Unable to import from semantic-text-splitter."
+                "Please install semantic-text-splitter with "
+                "`pip install -U semantic-text-splitter`."
+            ) from e
+        # Optionally can also have the splitter not trim whitespace for you
+        splitter = TiktokenTextSplitter(self.model_name, trim_chunks=self.trim_chunks)
+        return splitter.chunks(text, self.max_characters)
```
mspronesti, line 38: Doesn't langchain already support a semantic text splitter based on tiktoken [[source](https://github.com/langchain-ai/langchain/blob/master/libs/text-splitters/langchain_text_splitters/base.py#L220-L265)]?

rgupta2508: This is based on sementic-text-splitter package referr [https://docs.rs/text-splitter/0.12.1/text_splitter/](https://docs.rs/text-splitter/0.12.1/text_splitter/#method) .

Actually this library also using same method name TiktokenTextSplitter. but the how it is working and giving some better performance 
[TextSplitter Semantic Levels](https://docs.rs/text-splitter/0.12.1/text_splitter/#textsplitter-semantic-levels)

- Characters
- [Unicode Grapheme Cluster Boundaries](https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries)
- [Unicode Word Boundaries](https://www.unicode.org/reports/tr29/#Word_Boundaries)
- [Unicode Sentence Boundaries](https://www.unicode.org/reports/tr29/#Sentence_Boundaries)
- Ascending sequence length of newlines. (Newline is \r\n, \n, or \r) Each unique length of consecutive newline sequences is treated as its own semantic level. So a sequence of 2 newlines is a higher level than a sequence of 1 newline, and so on.

hwchase17:
imo this is too overlapping with existing langchain functionality where i dont think there is a ton of benefit in including a wrapper integration

PR #20988, opened 2024-04-28
langchain_community: GigaChat model selection fix

BlackSamorez:
Fixed the error that the model name is never actually put into GigaChat request payload, always defaulting to `GigaChat-Lite`.

With this fix, model selection through
```python
import os
from langchain.chat_models.gigachat import GigaChat

chat = GigaChat(
    name="GigaChat-Pro", # <- HERE!!!!!
    ...
)
```
should actually work, as intended in [here](https://github.com/langchain-ai/langchain/blob/804390ba4bcc306b90cb6d75b7f01a4231ab6463/libs/community/langchain_community/llms/gigachat.py#L36).


libs/community/langchain_community/chat_models/gigachat.py
```diff
@@ -133,6 +133,7 @@ def _build_payload(self, messages: List[BaseMessage], **kwargs: Any) -> gm.Chat:
         )
 
         payload.functions = kwargs.get("functions", None)
+        payload.model = self.name
```
baskaryan, line 136: should this be `payload.model = self.model`?

BlackSamorez: Yes, you're right. `self.model` is the correct thing. I got confused by the constructor pulling the `name` kwarg from the deepest parent class. Fixed.

PR #20995, opened 2024-04-28
community: import fix

leo-gan:
Issue: When the third-party package is not installed, whenever we need to `pip install <package>` the ImportError is raised.
But sometimes, the `ValueError` or `ModuleNotFoundError` is raised. It is bad for consistency.
Change: replaced the  `ValueError` or `ModuleNotFoundError` with `ImportError` when we raise an error with the `pip install <package>` message.
Note: Ideally, we replace all `try: import... except... raise ... `with helper functions like `import_aim` or just use the existing [langchain_core.utils.utils.guard_import](https://api.python.langchain.com/en/latest/utils/langchain_core.utils.utils.guard_import.html#langchain_core.utils.utils.guard_import) But it would be much bigger refactoring. @baskaryan Please, advice on this.

### DIFF
--- a/libs/community/langchain_community/cache.py
+++ b/libs/community/langchain_community/cache.py
@@ -323,7 +323,7 @@ class UpstashRedisCache(BaseCache):
         try:
             from upstash_redis import Redis
         except ImportError:
-            raise ValueError(
+            raise ImportError(
                 "Could not import upstash_redis python package. "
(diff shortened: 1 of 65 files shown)


baskaryan (approved):

leo-gan:
@baskaryan Thanks!

PR #21023, opened 2024-04-29
doc(chatbots) Adapt ChatModelTabs

liugddx:
- @baskaryan


docs/docs/use_cases/chatbots/quickstart.ipynb
baskaryan, line 1: dont need to import twice in same notebook, think that's what's causing issues

liugddx: @baskaryan PTAL. If this needs to be done, I will maintain other doc.

hwchase17:
we have updated our documentation structure, this is a great change but will need to be added to the new 0.2 docs

Issue #20909, opened 2024-04-25
Getting "HTTP Error 404: Not Found" Error Using ArxivLoader


Ahmetyasin:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain_community.document_loaders import ArxivLoader

docs = ArxivLoader(query = "Advanced regression models for real estate price prediction", load_max_docs=100, load_all_available_meta=True).load()

### Error Message and Stack Trace (if applicable)

---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
Cell In[9], [line 7](vscode-notebook-cell:?execution_count=9&line=7)
      [1](vscode-notebook-cell:?execution_count=9&line=1) from langchain_community.document_loaders import ArxivLoader
      [3](vscode-notebook-cell:?execution_count=9&line=3) #ArxivLoader
      [4](vscode-notebook-cell:?execution_count=9&line=4) #HTTPError: HTTP Error 404: Not Found
      [5](vscode-notebook-cell:?execution_count=9&line=5) 
      [6](vscode-notebook-cell:?execution_count=9&line=6) # Load documents using ArxivLoader with the concise_query
----> [7](vscode-notebook-cell:?execution_count=9&line=7) docs = ArxivLoader(query = "Advanced regression models for real estate price prediction", load_max_docs=100, load_all_available_meta=True).load()

File [/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/langchain_community/document_loaders/arxiv.py:27](https://file+.vscode-resource.vscode-cdn.net/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/langchain_community/document_loaders/arxiv.py:27), in ArxivLoader.load(self)
     [26](https://file+.vscode-resource.vscode-cdn.net/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/langchain_community/document_loaders/arxiv.py:26) def load(self) -> List[Document]:
[...]

PR #21042, opened 2024-04-29
community: Skip unexpected 404 HTTP Error in Arxiv download

jkpawlowski96:
### Description:
When attempting to download PDF files from arXiv, an unexpected 404 error frequently occurs. This error halts the operation, regardless of whether there are additional documents to process. As a solution, I suggest implementing a mechanism to ignore and communicate this error and continue processing the next document from the list.

Proposed Solution: To address the issue of unexpected 404 errors during PDF downloads from arXiv, I propose implementing the following solution:

- Error Handling: Implement error handling mechanisms to catch and handle 404 errors gracefully.
- Communication: Inform the user or logging system about the occurrence of the 404 error.
- Continued Processing: After encountering a 404 error, continue processing the remaining documents from the list without interruption.

This solution ensures that the application can handle unexpected errors without terminating the entire operation. It promotes resilience and robustness in the face of intermittent issues encountered during PDF downloads from arXiv.

### Issue:
#20909 
### Dependencies:
none


libs/community/langchain_community/utilities/arxiv.py
```diff
@@ -225,6 +227,12 @@ def lazy_load(self, query: str) -> Iterator[Document]:
             except (FileNotFoundError, fitz.fitz.FileDataError) as f_ex:
                 logger.debug(f_ex)
                 continue
+            except HTTPError as f_ex:
```
baskaryan, line 230: would it be better to instead add a `continue_on_failure: bool = False` argument to the loader the way a few other document loaders do (eg https://github.com/langchain-ai/langchain/blob/90f19028e57c5053888ad213f78d0c035169f4b9/libs/community/langchain_community/document_loaders/gcs_directory.py#L28)

jkpawlowski96: thanks a lot @baskaryan for the review. actually. I agree, it will be more practical to use such an attribute. Corrected

baskaryan (approved):

PR #21043, opened 2024-04-29
**PR title**: "community: Avoid cl100k_base encoding for Text embedding engines compatibility with LM Studio"

karkaratz:
- **Description:** With this update, users can set as False tiktoken_enabled when they ned it. By default, the value is set to True (so, not changing current behavior).
    - **Issue:** LM studio provide, by default, nomic-ai/nomic-embed-text-v1.5-GGUF for embedding. When calling the exposed API, with tiktoken_enabled=True, the client sends integers as embedding requests. The API expects Strings and it returns errors: [ERROR] 'input' field must be a string or an array of strings
    - **Dependencies:** 
    - **Twitter handle:** 

- [x] **Add tests and docs**: 
[test.txt](https://github.com/langchain-ai/langchain/files/15153763/test.txt)

### DIFF
--- a/libs/community/langchain_community/embeddings/openai.py
+++ b/libs/community/langchain_community/embeddings/openai.py
@@ -219,7 +219,7 @@ class OpenAIEmbeddings(BaseModel, Embeddings):
     """Timeout for requests to OpenAI completion API. Can be float, httpx.Timeout or 
         None."""
     headers: Any = None
-    tiktoken_enabled: bool = True
+    tiktoken_enabled: Optional[bool] = Field(default=True)
     """Set this to False for non-OpenAI implementations of the embeddings API, e.g.
     the `--extensions openai` extension for `text-generation-webui`"""
     tiktoken_model_name: Optional[str] = None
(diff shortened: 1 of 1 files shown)


baskaryan (commented):
what does typing this as Optional change? you can already set value to False (Optional gives you ability to set it to None)

karkaratz:
> what does typing this as Optional change? you can already set value to False (Optional gives you ability to set it to None)

You are right, it  has no point to make it optional.

baskaryan:
> > what does typing this as Optional change? you can already set value to False (Optional gives you ability to set it to None)
> 
> You are right, it has no point to make it optional. What you suggest to do? Close this PR and Open a New One?

what would new pr do? is a pr needed at all?

karkaratz:
> > > what does typing this as Optional change? you can already set value to False (Optional gives you ability to set it to None)
> > 
> > 
> > You are right, it has no point to make it optional. What you suggest to do? Close this PR and Open a New One?
> 
> what would new pr do? is a pr needed at all?

The conversation is too asynchronous. 😄 
I had modified the PR and I had removed the Optional.

hwchase17:
im not really sure what this does, i think ` = True` is the same as ` = Field(default=True)`

if i'm wrong, feel free to reopen

PR #21049, opened 2024-04-30
[Core] Tracing: update parent run_tree's child_runs

hinthornw:


libs/core/langchain_core/tracers/base.py
```diff
@@ -135,7 +137,7 @@ def _end_trace(self, run: Run) -> None:
                 logger.debug(f"Parent run with UUID {run.parent_run_id} not found.")
             elif (
                 run.child_execution_order is not None
-                and parent_run.child_execution_order is not None
+                and getattr(parent_run, "child_execution_order", None) is not None
```
agola11, line 140: why do we need the getattrs/hasattrs?

agola11 (approved):

libs/core/langchain_core/tracers/base.py
hinthornw: We don't actually have the child execution order in the base run / run tree objects

PR #21065, opened 2024-04-30
community: Extend AzureSearch with `maximal_marginal_relevance`,  `from_embeddings`

pokotylo:
**Description:**
- Extend AzureSearch with `maximal_marginal_relevance` (for vector and hybrid search)
- Add construction `from_embeddings` - if the user has already embedded the texts
- Add `add_embeddings` 
- Refactor common parts (`_simple_search`, `_results_to_documents`, `_reorder_results_with_maximal_marginal_relevance`)
- Add `vector_search_dimensions` as a parameter to the constructor to avoid extra calls to `embed_query` (most of the time the user applies the same model and knows the dimension)

**Issue:** none
**Dependencies:** none

- [x] **Add tests and docs**: The docstrings have been added to the new functions, and unified for the existing ones. The example notebook is great in illustrating the main usage of AzureSearch, adding the new methods would only dilute the main content. 
- [x] **Lint and test**


pokotylo:
Hi @baskaryan, could you please review the PR? We use a copy of this class with similar changes in production and would like to use langchain directly.

baskaryan:
> Hi @baskaryan, could you please review the PR? We use a copy of this class with similar changes in production and would like to use langchain directly.

made few small changes, can land if it looks good to you

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -555,12 +633,13 @@ def semantic_hybrid_search(
         Args:
             query (str): The query text for which to find similar documents.
             k (int): The number of documents to return. Default is 4.
+            filters: Filtering expression.
```
pokotylo, line 636: @baskaryan here and in 661: there is no argument `filters` that you added to the documentation. Is it intended?

baskaryan: think its useful to document since you can pass in via **kwargs

baskaryan (approved):

PR #21070, opened 2024-04-30
docs:Added documentation on Anthropic models on vertex

Adi8885:
Description:Added documentation on Anthropic models on Vertex
@lkuligin for review


docs/docs/integrations/llms/google_vertex_ai_palm.ipynb
baskaryan, line 1: can we set `model_name` in the ChatAnthropicVertex object init, instead of in the invocation? that's the more canonical way to do it

Adi8885: Agreed, demonstrated both ways of setting model_name

baskaryan (approved):

PR #21075, opened 2024-04-30
community: Refactor CassandraDatabase wrapper

cbornet:
* Introduce individual `fetch_` methods for easier typing.
* Rework some docstrings to google style
* Move some logic to the tool
* Merge the 2 cassandra utility files

### DIFF
--- a/libs/community/langchain_community/tools/cassandra_database/tool.py
+++ b/libs/community/langchain_community/tools/cassandra_database/tool.py
@@ -1,6 +1,7 @@
 """Tools for interacting with an Apache Cassandra database."""
 from __future__ import annotations
 
+import traceback
 from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence, Type, Union
 
 from langchain_core.callbacks import CallbackManagerForToolRun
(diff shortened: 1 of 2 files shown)


cbornet:
It would be nice if this can be reviewed and merged before the 0.17 release to not introduce breaking changes (the cassandra tool is not yet part of a release).

eyurtsev:
Are any users importing from casandra_database.py? If so, should we re-import there for backwards compatibility?

eyurtsev (approved):

cbornet:
> Not yet as casandra_database.py has not been part of a release yet.

Well, now that 1.17 has been released, it's not true anymore 😞 .
So I can keep things in cassandra_database.py. But now we have 2 utility modules for cassandra...
What do you prefer ?

baskaryan:
> > Not yet as casandra_database.py has not been part of a release yet.
> 
> Well, now that 1.17 has been released, it's not true anymore 😞 . So I can keep things in cassandra_database.py. But now we have 2 utility modules for cassandra... What do you prefer ?

we should keep around a version of `cassandra_database.py` that re-exports everything it does in current version, something like
```python
from langchain_community.utilities.cassandra import CassandraDatabase

__all__ = ["CassandraDatabase"]
```

cbornet:
I've rolled back to keep both files for the moment. We can see what to do with that later.

PR #21078, opened 2024-04-30
langchain[patch]: Upgrade prompts to optional imports

eyurtsev:
Upgrades prompts module to use optional imports.

This code was generated with a migration script, but had to be adjusted manually a bit.

Testing in preparation for applying this code modification across the rest of the modules in langchain package to reverse the dependency
between langchain community and langchain.

### DIFF
--- a/libs/langchain/langchain/prompts/__init__.py
+++ b/libs/langchain/langchain/prompts/__init__.py
@@ -27,9 +27,8 @@ from multiple components. Prompt classes and functions make constructing
                     ChatPromptValue
 
 """  # noqa: E501
-from langchain_community.example_selectors.ngram_overlap import (
-    NGramOverlapExampleSelector,
-)
+from typing import TYPE_CHECKING, Any
+
 from langchain_core.example_selectors import (
     LengthBasedExampleSelector,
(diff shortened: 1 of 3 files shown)


eyurtsev:

libs/langchain/langchain/prompts/__init__.py
```diff
@@ -53,8 +52,29 @@
...
 
+if TYPE_CHECKING:
+    from langchain_community.example_selectors.ngram_overlap import (
+        NGramOverlapExampleSelector,
+    )
+
+# Create a way to dynamically look up deprecated imports.
+# Used to consolidate logic for raising deprecation warnings and
+# handling optional imports.
+MODULE_LOOKUP = {
```
baskaryan, line 66: ooc any reason not to make this private _MODULE_LOOKUP

baskaryan (approved):

libs/langchain/langchain/prompts/__init__.py
eyurtsev: Ah no reason, just inertia -- we can follow up with another refactor to make this private

Issue #21080, opened 2024-04-30
Convert imports in langchain to langchain community to use optional imports


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

No help is required here. Creating an issue to link associated PRs.

eyurtsev:
https://github.com/langchain-ai/langchain/pull/21117/files

eyurtsev:
https://github.com/langchain-ai/langchain/pull/21258

liugddx:
What work are you doing on this issue?

eyurtsev:
@liugddx all of it in this case wasn't a good issue to parallelize since it involved deciding where to make breaking changes

eyurtsev:
Issue is now complete

PR #21090, opened 2024-04-30
langchain[patch]: Migrate chat models to optional community imports

eyurtsev:
Migrate chat models to optional community imports


eyurtsev:

PR #21092, opened 2024-04-30
langchain[patch]: Migrate document loaders to use optional community imports

eyurtsev:
Migrate document loaders to use optional community imports


eyurtsev:
Closing in favor of more recent one with simplified imports

PR #21099, opened 2024-04-30
langchain[patch]: Migrate embeddings to optional imports

eyurtsev:


eyurtsev:

PR #21108, opened 2024-04-30
Upgrade to Ruff v0.4.2

charliermarsh:
## Summary

No new diagnostics (given that the set of enabled rules hasn't changed), but gains access to our new parser (much faster) and reduced false positives all around.


charliermarsh:
More changes than I would've expected in the Poetry lock, but I ran with the same version (1.6.1) and `poetry lock --no-update`.

eyurtsev (approved):

Issue #6200, opened 2023-06-15
AzureChatOpenAI.get_num_tokens_from_messages  does not work


yoooyle:
### System Info

langchain==0.0.198 and current HEAD

AzureChat inherits from OpenAIChat
![image](https://github.com/hwchase17/langchain/assets/877883/11b94aef-6dd0-4f75-882a-9558b6550c1b)

Which throws on Azure's model name
![image](https://github.com/hwchase17/langchain/assets/877883/fa995f5c-ba8b-4a3b-a877-9981484893dd)

Azure's model name is gpt-35-turbo, not 3.5
![image](https://github.com/hwchase17/langchain/assets/877883/72d716e6-3402-47a3-a3cc-262b3720d8ed)

### Who can help?

@hwchase17 

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

1. deploy private GPT3.5 on Azure
2. Initialize an AzureChatOpenAI object
3. call get_num_tokens_from_messages
4. observe the exception

### Expected behavior

no exception

Hchyeria:
I encountered the same issue.
`OPENAI_MODEL_NAME = os.getenv("OPENAI_MODEL_NAME", "gpt-35-turbo")`
![image](https://github.com/langchain-ai/langchain/assets/30401991/d3d7980a-f30f-4ed3-9aa4-3e6eba67520d)

ellemcfarlane:
Hi, I ran into this issue with the latest version of langchain (v0.0.335). Please unmark as stale.

Hemario:
This issue still persists in version 0.1.11 of langchain.

ties-freeday:
This issue still exists in version 0.1.45 of langchain:
model "gpt-35-turbo" won't match if statement `elif model.startswith("gpt-3.5-turbo") or model.startswith("gpt-4"):`

efriis:
[...]

PR #21114, opened 2024-04-30
langchain_openai[patch]: Fix Azure OpenAI get_num_token

mjvankampen:
Azure has different model names making get_num_tokens not workable for gpt 3.5, ref #6200

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -719,12 +719,14 @@ class ChatOpenAI(BaseChatModel):
         if sys.version_info[1] <= 7:
             return super().get_num_tokens_from_messages(messages)
         model, encoding = self._get_encoding_model()
-        if model.startswith("gpt-3.5-turbo-0301") or model.startswith("gpt-35-turbo-0301"):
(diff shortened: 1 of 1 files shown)


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -719,12 +719,12 @@ def get_num_tokens_from_messages(self, messages: List[BaseMessage]) -> int:
         if sys.version_info[1] <= 7:
             return super().get_num_tokens_from_messages(messages)
         model, encoding = self._get_encoding_model()
-        if model.startswith("gpt-3.5-turbo-0301"):
+        if model.startswith("gpt-3.5-turbo-0301") or model.startswith("gpt-35-turbo-0301"):
```
mspronesti, line 722: How about `model = model.replace(".", "")` before the if-elif sequence (so that both the cases are covered)?

mjvankampen: Indeed better, did that and added a comment for future reference.

efriis:
Commented on the issue! I don't think this fix is necessary.

efriis:
(and happy to reopen if my recommendation of passing `model='gpt-3.5-turbo'` is unfeasible for any reason!)

PR #21120, opened 2024-04-30
community[chat_message_histories]: added MSSQL chat history

chip-davis:
This change adds another `chat_history` object dedicated to MSSQL. I created it after I saw there was already a `sql` one :(. I had to migrate from postgres (was using langchain/postgres) to MSSQL due to org policies. I quickly made this which is almost exactly the same as the postgres package just different drivers.

This does add two python dependencies, `pyodbc` and `aioodbc`. Additionally,  you will need MSSQL tools installed and odbc drivers on the host machine.

I'm not sure how useful this even is due to the `sql` class that already exists but this does have some advantages for folks who only care about MSSQL. For one, there's no sql-alchemy overhead. Connections are made directly using pyodbc. Second, I also implemented asynchronous methods using `aioodbc` which the `sql` class currently does not.

I have added two new unit tests and a unit test util. These pass but I have marked them as skip due to the dependencies involved as well as the need for a running mssql server. If requested, I can make some mock tests but didn't want to spend the time if this isn't needed.

Again, I don't even know if this class is neccecary due to the `sql` class but just figured I would throw it up if it helps anyone else :)


libs/community/langchain_community/chat_message_histories/mssql.py
```diff
@@ -0,0 +1,234 @@
...
+
+        values = [
+            (self._session_id, json.dumps(message_to_dict(message)))
+            for message in messages
+        ]
+        query = _insert_message_query(self._table_name)
+        cursor = await self._aconnection.cursor()
+        await cursor.executemany(query, values)
+        await cursor.commit()
+        await cursor.close()
```
eyurtsev, line 180: Could you create a context manager to handle the closing of the cursor

chip-davis: I must have really missed something. When I was implementing it, I got an error saying that context managers were not implemented for aioodbc and just went on my way. But according to the front page of their repo, they do https://github.com/aio-libs/aioodbc

Will implement and push up

chip-davis:
Does this need any changes for langchain 0.2?

chip-davis:
@eyurtsev does this need any changes?

efriis:
closing as stale/failing CI, and if you want to take another go would recommend contributing it to the `langchain-sqlserver` package here! https://github.com/langchain-ai/langchain-azure/tree/main/libs/sqlserver/langchain_sqlserver

In general, we have moved towards storing message histories in checkpointers instead of in these ChatMessageHistory wrappers though

PR #20981, opened 2024-04-28
[Community] : Added SentenceWindowRetriever

rsk2327:
Thank you for contributing to LangChain!

- [x] **PR title**: "package: description"

Description : Updated TextSplitter to include a new add_chunk_id argument to add a chunk_id variable into document metadata

This is a prerequisite for implementing the Sentence Window Retriever broadly across all databases as it helps easily identify neighboring chunks of text from the same text source

- [x] **Add tests and docs**
This is a simple argument addition thats False by default and has basic logic. Specific unit tests might not be necessary

- [x] **Lint and test**: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified. 

If no one reviews your PR within a few days, please @-mention one of baskaryan, efriis, eyurtsev, hwchase17.

PR #21130, opened 2024-04-30
text-splitters: Extend TextSplitter:keep_separator functionality

### LINKED ISSUE #20908: `CharacterTextSplitter` with `keep_separator=True` sets the separator to the beginning of each chunk instead of an end
### Checked other resources

### Example Code

```python
from langchain.text_splitter import CharacterTextSplitter

splitter = CharacterTextSplitter(
    chunk_size=10,
    chunk_overlap=0,
    separator=". ",
    keep_separator=True,
)

for d in splitter.create_documents(["Text 1. Text 2. Text 3. Text 4."]):
    print(d.page_content)
    print("---")
```

```
Text 1
---
. Text 2
---
. Text 3
---
. Text 4.
---
```

### Description

I'm trying to split text by sentence, while keeping end-of-sentence punctuation. Instead of putting the punctuation back at the end of the corresponding chunk, the library adds it to the front of the following chunk.

This problem is quite critical if the output is used for text-to-speech input.

### System Info

```
langchain==0.1.16
langchain-community==0.0.34
langchain-core==0.1.46
```

MacOS, Python 3.10.13

xbouroseu:
**Description:** Added extra functionality to `CharacterTextSplitter`, `TextSplitter` classes. 
The user can select whether to append the separator to the previous chunk with `keep_separator='end' ` or else prepend to the next chunk. Previous functionality prepended by default to next chunk.
  
**Issue:** Fixes #20908


libs/text-splitters/langchain_text_splitters/base.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         chunk_overlap: int = 200,
         length_function: Callable[[str], int] = len,
         keep_separator: bool = False,
+        separator_kept_at_end: bool = False,
```
baskaryan, line 39: instead of adding another parameter thoughts on expanding the type of `keep_separator` to something like: `Union[bool, Literal["beginning", "end"]]`

xbouroseu: Did so! Let me know if this looks okay :)

rsk2327:
@baskaryan I have another PR for Textsplitter that adds a new argument. Could I get a review on that?

https://github.com/langchain-ai/langchain/pull/20981

libs/text-splitters/langchain_text_splitters/base.py
baskaryan: nice! could we add a simple unit test?

xbouroseu: Sure! Do you want only for the scenario where `keep_separator=="end"`?

baskaryan: one for "start" and one for "end" would be great

xbouroseu: @baskaryan Done, please let me know if there is anything else to be done!

xbouroseu: @baskaryan Following up to see if we could close this?

hwchase17 (commented):
can we update the documentation for `keep_separator` in the docstring?

xbouroseu:
> can we update the documentation for `keep_separator` in the docstring?

Yes I updated! Please let me know if it needs something more.

hwchase17 (approved):

PR #21133, opened 2024-05-01
core: `utils.guard_import` fix

leo-gan:
Issues (nit): 
1. `utils.guard_import` prints wrong error message when there is an import `error.` It prints the whole `module_name` but should be only the first part as the pip package name. E.i. `langchain_core.utils` -> print not `langchain-core` but `langchain_core.utils`. Also replace '_' with '-' in the pip package name.
2. it does not handle the `ModuleNotFoundError` which raised if `guard_import("wrong_module")`

Fixed issues; added ut-s. Controversial: I've reraised `ModuleNotFoundError` as `ImportError`, since in case of the error, the proposed action is the same - we need to install a missed package.

### DIFF
--- a/libs/core/langchain_core/utils/utils.py
+++ b/libs/core/langchain_core/utils/utils.py
@@ -89,7 +89,7 @@ def guard_import(
     installed."""
     try:
         module = importlib.import_module(module_name, package)
-    except ImportError:
+    except (ImportError, ModuleNotFoundError):
         pip_name = pip_name or module_name.split(".")[0].replace("_", "-")
         raise ImportError(
             f"Could not import {module_name} python package. "
--- a/libs/core/langchain_core/utils/utils.py
+++ b/libs/core/langchain_core/utils/utils.py
@@ -1,4 +1,5 @@
 """Generic utility functions."""
+
 import contextlib
 import datetime
 import functools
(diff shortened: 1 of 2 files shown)


eyurtsev:
See comment here: https://github.com/langchain-ai/langchain/pull/21156#issuecomment-2089164930

I'm opposed to this strategy unless we can fix type information

leo-gan:
@eyurtsev Please, see comments to #21156

eyurtsev:
Moving discussion from here https://github.com/langchain-ai/langchain/pull/21156#issuecomment-2089164930 into here since all the PRs are linking this.

I don't think it's possible for the IDEs / static typing tools to behave correctly -- as type information gets washed out by the `Any`

```

def guard_import(
    module_name: str, *, pip_name: Optional[str] = None, package: Optional[str] = None
) -> Any:
```

---

image

leo-gan:
Interesting. In some cases, the static typing works, in others cases it didn't (Disclosure: I've checked PyCharm only). For example, 
Screenshot from 2024-05-02 10-16-38 - works fine.

I've checked both variants and they work the same. If static typing has worked before refactoring it works after refactoring. If it didn't' work before, it doesn't after.
In my case: `import_comet_ml` didn't work; `import_clearml`, `import_pandas`, `import_fiddler` work fine.

Anyway, the custom import functions also return `Any` type. I.e. `def import_comet_ml() -> Any:`
So, they return `Any` and `guard_import` returns `Any`. Refactoring does not change this behaviour but it only unifies the error handling.

eyurtsev (approved):

eyurtsev:
@leo-gan if we're updating existing code that's using this approach i think it's fine, but I think we should over time remove usage of guard import rather than adopt it. Because of the `Any` return type it's impossible for static typing analyzers to do things correctly. The IDE will sometimes be able to guess (especially if you're using github copilot), but ti can also guess quite incorrectly if using slightly unexpected names etc.

leo-gan:
@eyurtsev Maybe we can update `quard_import` return type with generic type or something like this? It looks like bootstrapping but it may work.
Thanks you for discussion and review!!!!

PR #21137, opened 2024-05-01
Remove remaining unused suppression comments

charliermarsh:
## Summary

Just running `ruff check --extend-select RUF100 -n --fix` to remove any remaining, unused `# noqa` violations.

Many of these are `E501` violations that Ruff got smarter about. For example, if you have a `# type: ignore` at the end of a line, and the `# type: ignore` is what _causes_ the line to be too long, Ruff won't require you to put an `# noqa: E501` at the end.


hwchase17 (approved):
thanks charlie :)

baskaryan:
landed in #22049, thanks for sharing the fix @charliermarsh!

PR #21138, opened 2024-05-01
Use `ruff check` in Makefile

charliermarsh:
## Summary

`ruff /path/to/file.py` works but is deprecated, and we now recommend `ruff check /path/to/file.py` (to match `ruff format /path/to/file.py`).


eyurtsev (approved):

eyurtsev:
Thank you @charliermarsh

PR #21142, opened 2024-05-01
core: bump packaging dep to v24

dokterbob:
Allow installing core with packaging@^24.

### DIFF
--- a/libs/core/pyproject.toml
+++ b/libs/core/pyproject.toml
@@ -15,7 +15,7 @@ langsmith = "^0.1.0"
 tenacity = "^8.1.0"
 jsonpatch = "^1.33"
 PyYAML = ">=5.3"
-packaging = "^23.2"
+packaging = ">=23.2,^24.0"
 jinja2 = { version = "^3", optional = true }
 
 [tool.poetry.group.lint]
--- a/libs/core/poetry.lock
+++ b/libs/core/poetry.lock
@@ -1620,13 +1620,13 @@ files = [
 
 [[package]]
 name = "packaging"
-version = "23.2"
+version = "24.0"
 description = "Core utilities for Python packages"
 optional = false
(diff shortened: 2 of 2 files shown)


libs/core/pyproject.toml
```diff
@@ -15,7 +15,7 @@ langsmith = "^0.1.0"
 tenacity = "^8.1.0"
 jsonpatch = "^1.33"
 PyYAML = ">=5.3"
-packaging = "^23.2"
+packaging = ">=23.2,^24.0"
```
eyurtsev, line 18: >=23.2 implies 24 and 25 if i'm not mistaken?

dokterbob: `>=` equal or higher than.

AFAIK this implies higher than or equal to 23.2 and lower than 25. 

Ref: https://python-poetry.org/docs/dependency-specification/

eyurtsev (approved):

eyurtsev:
Thanks! We'll try to see if we can remove packaging as a dependency to avoid dependency conflicts for users

dokterbob:
> Thanks! We'll try to see if we can remove packaging as a dependency to avoid dependency conflicts for users

That's great to hear but in the mean time, perhaps we could use my code to fix the issue for the time being?

Just rebased my PR to master.

libs/core/pyproject.toml
baskaryan: afaik the `,` indicates a logical AND not a logical OR, so `>=23.2,^24.0` ends up just being `^24.0`. should it isntead be `>=23.2,<25`?

mbastian:
Hi folks, I would appreciate if we could merge that as my project currently has conflicts due to this library as another dependency has >=24.0. Thanks in advance!

hwchase17:
fixed in other pr

PR #21155, opened 2024-05-01
langchain[patch]: Migrate retrievers to use optional langchain community imports

eyurtsev:
Upgrade retrievers


eyurtsev:

PR #21156, opened 2024-05-01
community[patch]: `tools` imports

leo-gan:
Issue: we have several helper functions to import third-party libraries like tools.gmail.utils.import_google in [community.tools](https://api.python.langchain.com/en/latest/community_api_reference.html#id37). And we have core.utils.utils.guard_import that works exactly for this purpose.
The import_<package> functions work inconsistently and rather be private functions.
Change: replaced these functions with the guard_import function.

Related to #21133

### DIFF
--- a/libs/community/langchain_community/tools/gmail/utils.py
+++ b/libs/community/langchain_community/tools/gmail/utils.py
@@ -1,9 +1,12 @@
 """Gmail tool utils."""
+
 from __future__ import annotations
 
 import logging
 import os
 from typing import TYPE_CHECKING, List, Optional, Tuple
+
+from langchain_core.utils import guard_import
 
 if TYPE_CHECKING:
     from google.auth.transport.requests import Request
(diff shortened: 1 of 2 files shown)


eyurtsev:
How does this affect developer experience w/ respect to typing and IDE completion? It could be that we improve an error message but make it overall more difficult to develop?

leo-gan:
@eyurtsev Using `guard_import` will decrease typing, since all error handling is done. I don't see how it makes more difficult for development. It adds one more import from core. But the `community` code also depends on core, so no problem here.
But this change is more about consistency. I see many variants of this dynamic imports for no reason.
PyCharm makes completion without problem.

eyurtsev:
guard_import returns an Any type. which prevents auto-completion in the IDE from being useful.

```python
    from langchain_core.utils.utils import guard_import

    anthropic = guard_import("langchain_text_splitters.TextSplitter")()
```

Try seeing what `anthropic.` will suggest on tab.

---

I suspect this will lead to a worse developer experience and more bugs.

leo-gan:
@eyurtsev 
The guard_import used to import a module:
`anthropic = guard_import("langchain_text_splitters").TextSplitter`
Then all IDE suggestions work just fine.

eyurtsev (approved):

eyurtsev:
@leo-gan if the guards are scoped like this:

```
def import_installed_app_flow() -> InstalledAppFlow: # <-- type annotation here
	@@ -39,14 +41,9 @@ def import_installed_app_flow() -> InstalledAppFlow:
    Returns:
        InstalledAppFlow: InstalledAppFlow class.
    """
    return guard_import(
        module_name="google_auth_oauthlib.flow", pip_name="google-auth-oauthlib"
    ).InstalledAppFlow
```

Then everything works as long as users use `import_installed_app_flow()` -- I still think it's a bit indirect, but there won't be a lot off type information as the type will still be available fro the top level function

PR #21166, opened 2024-05-01
langchain[patch]: Migrate more code in utils to use optional langchain import

eyurtsev:
Moving is interactive util to avoid circular deps


eyurtsev:

PR #21169, opened 2024-05-01
community: `vectorstores` import update

leo-gan:
Issue: we have several helper functions to import third-party libraries like lancedb.import_lancedb in [community.vectorstores](https://api.python.langchain.com/en/latest/vectorstores/langchain_community.vectorstores.lancedb.import_lancedb.html#langchain_community.vectorstores.lancedb.import_lancedb). And we have core.utils.utils.guard_import that works exactly for this purpose.
The import_<package> functions work inconsistently and rather be private functions.
Change: replaced these functions with the guard_import function.

Related to #21133

### DIFF
--- a/libs/community/langchain_community/vectorstores/usearch.py
+++ b/libs/community/langchain_community/vectorstores/usearch.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, Iterable, List, Optional, Tuple
 import numpy as np
 from langchain_core.documents import Document
 from langchain_core.embeddings import Embeddings
+from langchain_core.utils import guard_import
 from langchain_core.vectorstores import VectorStore
 
 from langchain_community.docstore.base import AddableMixin, Docstore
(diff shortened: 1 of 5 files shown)


libs/community/langchain_community/vectorstores/usearch.py
```diff
@@ -170,7 +164,7 @@ def from_texts(
             documents.append(Document(page_content=text, metadata=metadata))
 
         docstore = InMemoryDocstore(dict(zip(ids, documents)))
-        usearch = dependable_usearch_import()
+        usearch = guard_import("usearch.index")
```
hwchase17, line 167: does this mean we dont need the `dependable_usearch_import` function at all?

same for the above functions

leo-gan: @hwchase17 I'd prefer to completely replace them. But it is a breaking change, right? So, the whole deprecation ceremony should be in place. If you agree on this I can add deprecation decorators to all custom_import function (whenever it is applicable). But I need instructions about the decorator values.

eyurtsev (approved):

leo-gan:
Thanks!

PR #21170, opened 2024-05-01
langchain[patch]: Update more code to use langchain community as an optional dependency

eyurtsev:
More code to use langchain community as an optional dependency


libs/langchain/langchain/tools/ainetwork/app.py
```diff
@@ -1,7 +0,0 @@
-from langchain_community.tools.ainetwork.app import (
```
eyurtsev, line 1: Deleted since ainetwork wasn't registered as a proper package (not __init__.py)

eyurtsev:

PR #21174, opened 2024-05-02
community: langkit dependency

leo-gan:
Issue: the `langkit` package is not presented in the `pyproject.toml` but it is a requirement for the `WhyLabsCallbackHandler`
Change: added `langkit`


ccurme (approved):

efriis:
Hey team - going to revert this because it doesn't seem necessary for testing. We should only be adding optional + `extended_testing` dependencies for deps that have extended tests.

otherwise it just increases probability of dependency conflicts in the community lockfile.

PR #21183, opened 2024-05-02
Hotfix RetrievalQA Docs: docs: Fix formatting

SecurityViber:
# Newline Characters breaking formatting 

**Description**: 
As you can see in the image below, the formatting in the documentation is broken. As far as I can see the two added `\n` characters are breaking the documentation. Therefore I would propose to remove those
image

**Dependencies**:
None

**Twitter Handle**
- epu9byj


ccurme:
Great catch thanks.

ccurme (approved):

Issue #20923, opened 2024-04-26
[BUG] langchain-openai 0.1.2+ breaks async generation


robcaulk:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_openai import ChatOpenAI
llm = ChatOpenAI(model="gpt-3.5-turbo-16k", temperature=0)
prompt = ChatPromptTemplate.from_messages(
    [
        (
            "system",
            "You are a helpful assistant that translates {input_language} to {output_language}.",
        ),
        ("human", "{input}"),
    ]
)
agent = prompt | llm
response = ""
async for token in agent.astream({
        "input_language": "English",
        "output_language": "German",
        "input": "I love programming.",
    }):
    response += token.content
```

### Error Message and Stack Trace (if applicable)

  File "/home/rcaulk/projects/flowdapt_new_schema/chat-service/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py", line 2900, in astream
    async for chunk in self.atransform(input_aiter(), config, **kwargs):
  File "/home/rcaulk/projects/flowdapt_new_schema/chat-service/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py", line 2883, in atransform
    async for chunk in self._atransform_stream_with_config(
  File "/home/rcaulk/projects/flowdapt_new_schema/chat-service/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py", line 1980, in _atransform_stream_with_config
    chunk: Output = await asyncio.create_task(  # type: ignore[call-arg]
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rcaulk/projects/flowdapt_new_schema/chat-service/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py", line 2853, in _atransform
    async for output in final_pipeline:
  File "/home/rcaulk/projects/flowdapt_new_schema/chat-service/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py", line 1334, in atransform
    async for output in self.astream(final, config, **kwargs):
[...]

PR #21184, opened 2024-05-02
openai: Revert "wrap stream code in context manager blocks"

snopoke:
Reverts langchain-ai/langchain#18013

Issue: https://github.com/langchain-ai/langchain/issues/20923

There have been a number reports of the `'async_generator' object does not support the asynchronous context manager protocol` error since this code was changed.

Until it's clear what the issue is I suggest we revert the changes. The only possible downside is that if anyone is relying on terminating the stream early, that will no longer happen. But there should be no changes to the output.


efriis:
Hey @snopoke ! From the reports in the linked issue, I suspect folks might be using libraries that patch openai, such that the response object is different.

Given this works with an unpatched `openai`, I'll close this PR and keep the context management code for now. If you see any issues from folks that are using unmodified versions of langchain, let me know! Happy to reopen and merge in that case.

PR #21189, opened 2024-05-02
[community][fix](DeepInfraEmbeddings): Implement chunking for large batches

ovuruska:
**Description:**
This PR introduces chunking logic to the `DeepInfraEmbeddings` class to handle large batch sizes without exceeding maximum batch size of the backend. This enhancement ensures that embedding generation processes large batches by breaking them down into smaller, manageable chunks, each conforming to the maximum batch size limit.

**Issue:**
Fixes #21189

**Dependencies:**
No new dependencies introduced.


libs/community/langchain_community/embeddings/deepinfra.py
```diff
@@ -6,6 +6,7 @@
 from langchain_core.utils import get_from_dict_or_env
 
 DEFAULT_MODEL_ID = "sentence-transformers/clip-ViT-B-32"
+MAX_BATCH_SIZE = 1024
```
eyurtsev, line 9: Could you expose this as part of the initializer with a default value? Presumably this can vary over time or folks may want to optimize it in a different way.

Embeddings is not the most obvious place to control the batch size. Usually embedding is tied to an indexing process which will be associated with a smaller batch size.

eyurtsev (approved):
This looks good, would love to see as part of initializer

libs/community/langchain_community/embeddings/deepinfra.py
ovuruska: Thanks for your feedback @eyurtsev . PR was updated as desired.

hwchase17 (approved):

PR #21190, opened 2024-05-02
langchain[patch],community[minor]: Move some unit tests from langchain to community, use core for fake models

eyurtsev:
Updates in some tests


libs/community/scripts/lint_imports.sh
```diff
@@ -6,7 +6,6 @@ set -eu
 errors=0
 
 # make sure not importing from langchain or langchain_experimental
-git --no-pager grep '^from langchain\.' . && errors=$((errors+1))
```
eyurtsev, line 9: Adjusting this lint rule in preparation for flipping the dependency between community and langchain

community will be allowed to dependent on langchain
but langchain will no longer be allowed to dependent on community

eyurtsev:

PR #21192, opened 2024-05-02
Update lancedb.py

raghavdixit99:
very minor update in LanceDB integration, 'metric' argument was missing.

### DIFF
--- a/libs/community/langchain_community/vectorstores/lancedb.py
+++ b/libs/community/langchain_community/vectorstores/lancedb.py
@@ -162,6 +162,7 @@ class LanceDB(VectorStore):
 
     def create_index(
         self,
+        metric : Optional[str] = 'L2',
         col_name: Optional[str] = None,
         vector_col: Optional[str] = None,
         num_partitions: Optional[int] = 256,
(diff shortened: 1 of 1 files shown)


raghavdixit99:
@baskaryan, tagging you for quick merge as its a very small change. please review

libs/community/langchain_community/vectorstores/lancedb.py
```diff
@@ -162,6 +162,7 @@ def get_table(self, name: Optional[str] = None) -> Any:
 
     def create_index(
         self,
+        metric: Optional[str] = "L2",
```
eyurtsev, line 165: This is a breaking change because users may be invoking this code using positional arguments. 

Please put the argument at end and precede it with a `*` operator.

raghavdixit99: Adding it as a `*arg `limits clarity and usage as our user can only specify the arg as a simple string without being able to pass it as a `kwarg (metric = 'cosine' )`, I just shifted this argument to the end of precedence. 

Since its an optional argument and is shifted to the end, it doesnt break previous usage, I have tested it on my end and there is no breakage. Please have a look, Thank you

raghavdixit99:
@eyurtsev  / @baskaryan have updated the branch

baskaryan (approved):

PR #21194, opened 2024-05-02
langchain[patch]: chat histories to handle optional community dependence

eyurtsev:


libs/langchain/langchain/memory/chat_message_histories/in_memory.py
```diff
@@ -1,3 +1,21 @@
-from langchain_community.chat_message_histories.in_memory import ChatMessageHistory
+from typing import TYPE_CHECKING, Any
+
+from langchain._api import create_importer
+
+if TYPE_CHECKING:
+    from langchain_community.chat_message_histories.in_memory import ChatMessageHistory
```
ccurme, line 6: Could do in another PR, but `langchain_community` just imports this from core, so it may make sense to do the same here?

ccurme (approved):

eyurtsev:

Issue #21193, opened 2024-05-02
FastEmbedEmbeddings Abstraction has limited params, unable to set important params which then causes a timeout


Waffleboy:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
class FastEmbedEmbeddings(BaseModel, Embeddings):
    """Qdrant FastEmbedding models.
    FastEmbed is a lightweight, fast, Python library built for embedding generation.
    See more documentation at:
    * https://github.com/qdrant/fastembed/
    * https://qdrant.github.io/fastembed/

    To use this class, you must install the `fastembed` Python package.

    `pip install fastembed`
    Example:
        from langchain_community.embeddings import FastEmbedEmbeddings
        fastembed = FastEmbedEmbeddings()
    """

    model_name: str = "BAAI/bge-small-en-v1.5"
    """Name of the FastEmbedding model to use
    Defaults to "BAAI/bge-small-en-v1.5"
    Find the list of supported models at
    https://qdrant.github.io/fastembed/examples/Supported_Models/
    """

    max_length: int = 512
    """The maximum number of tokens. Defaults to 512.
    Unknown behavior for values > 512.
    """

    cache_dir: Optional[str]
    """The path to the cache directory.
    Defaults to `local_cache` in the parent directory
    """

    threads: Optional[int]
    """The number of threads single onnxruntime session can use.
    Defaults to None
    """

    doc_embed_type: Literal["default", "passage"] = "default"
    """Type of embedding to use for documents
    The available options are: "default" and "passage"
    """

    _model: Any  # : :meta private:

    class Config:
        """Configuration for this pydantic object."""

        extra = Extra.forbid

    @root_validator()
    def validate_environment(cls, values: Dict) -> Dict:
        """Validate that FastEmbed has been installed."""
        model_name = values.get("model_name")
        max_length = values.get("max_length")
        cache_dir = values.get("cache_dir")
        threads = values.get("threads")

       # ----------- the below is the problem  -----------  #
        try:
            # >= v0.2.0
            from fastembed import TextEmbedding

            values["_model"] = TextEmbedding(
                model_name=model_name,
                max_length=max_length,
[...]

PR #21197, opened 2024-05-02
community: Fix abstraction in langchain for fastembed embeddings.

Waffleboy:
linked accepted PR in the fastembed library https://github.com/qdrant/fastembed/pull/223

  **Description:** 
FastEmbedEmbeddings does not work without internet, even if the model is already present locally. This is despite providing cache_dir. This is because under the hood it calls huggingface api, which then makes a network call to check the repo before checking if the model exists locally.

This behavior can be fixed if you specify local_files_only to True. However, this was not previously allowed in fastembed.
  **Issue:** 

This PR updates the langchain abstraction to allow the user to pass params like `local_files_only` to fastembed. As mentioned in the issue, I dont see why we limit the params that can be passed on to fastembed. There are min params yes, but the others should be passed along as well. TextEmbedding allows kwargs. The abstraction should too.

Linked issue here https://github.com/langchain-ai/langchain/issues/21193

   **Dependencies:** 

Nil


hwchase17 (approved):

hwchase17 (changes requested):
1) does this work? if we are passing all values to TextEmbedding class, i see `doc_embed_type` as a value that will be present that previously we were NOT passing to TextEmbedding. so.... whats that about?

2) there are other classes, like FlagEmbedding - shouldnt we do the same there?

ccurme:

PR #21216, opened 2024-05-02
openai[patch]: support tool_choice="required"

baskaryan:


libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -563,6 +572,21 @@ def test_manual_tool_call_msg() -> None:
         llm_with_tool.invoke(msgs)
 
 
+def test_bind_tools_tool_choice() -> None:
+    """Test passing in manually construct tool call message."""
+    llm = ChatOpenAI(model="gpt-3.5-turbo-0125", temperature=0)
+    for tool_choice in ("any", "required"):
```
ccurme, line 578: why not test the other possible values?

ccurme (approved):

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
baskaryan: this is specifically testing that a tool is called even when the prompt is irrelevant to any tool

PR #21218, opened 2024-05-02
[community][graph]: Update KuzuQAChain and docs

prrao87:
This PR makes some small updates for `KuzuQAChain` for graph QA.

- Updated Cypher generation prompt (we now support `WHERE EXISTS`) and generalize it more
- Support different LLMs for Cypher generation and QA
- Update docs and examples


libs/langchain/langchain/chains/graph_qa/prompts.py
```diff
@@ -78,9 +78,9 @@
 
 Generate the Kùzu dialect of Cypher with the following rules in mind:
 
-1. Do not use a `WHERE EXISTS` clause to check the existence of a property.
-2. Do not omit the relationship pattern. Always use `()-[]->()` instead of `()->()`.
-3. Do not include any notes or comments even if the statement does not produce the expected result.
+1. Do not omit the relationship pattern. Always use `()-[]->()` instead of `()->()`.
```
ccurme, line 81: We try to avoid changing prompts like this as it can silently change behavior for users. What are your thoughts?

hwchase17 (commented):
you likely need to merge master - this has been moved to community

libs/langchain/langchain/chains/graph_qa/prompts.py
prrao87: I think it's fine in this case (though I total get your point, changing prompts that may affect larger behavior isn't ideal by any means). However, here, I think the earlier prompt isn't suitable and I've tested the behaviour subjectively using multiple proprietary and open source LLMs, and it seems to not change the behaviour at all. Additionally, we have a rather small user base that's actively using LangChain right now, but this may grow in the coming months, so the sooner we get this in from a Kùzu perspective, the better 🙂.

Hope that makes sense!

prrao87:
@hwchase17 did you mean I have to merge master to this branch on my fork? I believe it all checks out and there aren't any conflicts.

hwchase17:
Screenshot 2024-05-08 at 2 31 26 PM

prrao87:
@eyurtsev should be good to go now 👍🏽

hwchase17 (approved):

PR #21219, opened 2024-05-02
community: Updating Multi-modal in Ollama to match OpenAI API 

0ssamaak0:
# Description
[Ollama Multimodal](https://python.langchain.com/docs/integrations/chat/ollama/#multi-modal) is different from OpenAI API Format, which is used also by other providers like Fireworks, I changed it to match them

# Ollama Multimodal Before Changing
```python
def prompt_func(data):
    ...
    image_part = {
        "type": "image_url",
        "image_url": f"data:image/jpeg;base64,{image}",
    }

    content_parts = []

    text_part = {"type": "text", "text": text}
    ...
```
`image_url` contains the encoded image directly, however in [OpenAI API](https://platform.openai.com/docs/guides/vision/uploading-base-64-encoded-images) it's under `url` key:

# OpenAI API Vision Guide
```python
{
    "role": "user",
    "content": [
        {"type": "text", "text": "What’s in this image?"},
        {
            "type": "image_url",
            "image_url": {"url": f"data:image/jpeg;base64,{base64_image}"},
        },
    ],
}
```

# Fireworks Multimodal
[Fireworks](https://readme.fireworks.ai/docs/querying-vision-language-models#chat-completions-api) Follows the same scheme as OpenAI
```python
{
    "role": "user",
    "content": [{
      "type": "text",
      "text": "Can you describe this image?",
    }, {
      "type": "image_url",
      "image_url": {
        "url": f "data:image/jpeg;base64,{image_base64}"
      },
    }, ],
  }
```

Twitter: @0ssamaak0

### DIFF
--- a/libs/community/langchain_community/chat_models/ollama.py
+++ b/libs/community/langchain_community/chat_models/ollama.py
@@ -120,8 +120,12 @@ class ChatOllama(BaseChatModel, _OllamaCommon):
                     if content_part.get("type") == "text":
                         content += f"\n{content_part['text']}"
                     elif content_part.get("type") == "image_url":
-                        if isinstance(content_part.get("image_url"), str):
-                            image_url_components = content_part["image_url"].split(",")
+                        if isinstance(
+                            content_part.get("image_url"), dict
+                            ) and isinstance(content_part.get("image_url").get("url"), str):
+                            image_url_components = content_part["image_url"][
+                                "url"
+                            ].split(",")
                             # Support data:image/jpeg;base64,<image> format
                             # and base64 strings
                             if len(image_url_components) > 1:
--- a/libs/community/langchain_community/chat_models/ollama.py
+++ b/libs/community/langchain_community/chat_models/ollama.py
@@ -120,6 +120,7 @@ class ChatOllama(BaseChatModel, _OllamaCommon):
                     if content_part.get("type") == "text":
                         content += f"\n{content_part['text']}"
                     elif content_part.get("type") == "image_url":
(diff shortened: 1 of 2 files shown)


0ssamaak0:
@baskaryan Can you check this? 😁

nathansebhastian:
Hi @0ssamaak0 can you actually send a multimodal message to Ollama?

I'm trying to send one with the "image_url" in string format:

```py
{
      "type": "image_url",
      "image_url": f"data:image/jpeg;base64,{image}",
  }
```

But I get ValueError: Only string image_url content parts are supported

For some reason the "url" key is added to the "image_url" key even though I send a string.

More detail on the issue: https://github.com/langchain-ai/langchain/issues/22460

libs/community/langchain_community/chat_models/ollama.py
```diff
@@ -120,8 +120,13 @@ def _convert_messages_to_ollama_messages(
                     if content_part.get("type") == "text":
                         content += f"\n{content_part['text']}"
                     elif content_part.get("type") == "image_url":
-                        if isinstance(content_part.get("image_url"), str):
```
baskaryan, line 123: we should still handle the case where image_url is a string right?

ccurme:
Thanks for flagging this. Closing as I believe this is handled properly in the current version of `langchain-ollama`:

https://github.com/langchain-ai/langchain/blob/3e296e39c8739cffe91d117e28d5fa99b97ec129/libs/partners/ollama/langchain_ollama/chat_models.py#L414-L429

PR #21224, opened 2024-05-02
community: Updated Chroma version range to include 0.5.0 release

tazarov:
- Updated Chroma version range to allow releases in 0.5.x.
- Bumped mypy version as linting was failing


tazarov:
cc: @jeffchuber and @HammadB

hwchase17 (approved):

PR #21229, opened 2024-05-02
docs: add tool choice to tool calling

baskaryan:


docs/docs/modules/model_io/chat/function_calling.ipynb
efriis, line 1: nit: should we rename file and redirect to `tool_calling.ipynb`

efriis (approved):

PR #21238, opened 2024-05-03
community: Instantiate GPT4AllEmbeddings with parameters

Alex-J-W:
### GPT4AllEmbeddings parameters
---

**Description:** 
As of right now the **Embed4All** class inside _GPT4AllEmbeddings_ is instantiated as it's default which leaves no room to customize the chosen model and it's behavior. Thus:

- GPT4AllEmbeddings can now be instantiated with custom parameters like a different model that shall be used.

### DIFF
--- a/libs/community/langchain_community/embeddings/gpt4all.py
+++ b/libs/community/langchain_community/embeddings/gpt4all.py
@@ -14,6 +14,12 @@ class GPT4AllEmbeddings(BaseModel, Embeddings):
 
             from langchain_community.embeddings import GPT4AllEmbeddings
 
+            model_name = "all-MiniLM-L6-v2.gguf2.f16.gguf"
+            gpt4all_kwargs = {'allow_download': 'True'}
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/embeddings/gpt4all.py
```diff
@@ -19,14 +19,32 @@ class GPT4AllEmbeddings(BaseModel, Embeddings):
 
     client: Any  #: :meta private:
 
+    def __init__(
```
hwchase17, line 22: whats this needed for?

Alex-J-W: - I am not used to pydantic. Updated the PR to adopt the pydantic `BaseModel` convention and removed the  `__init__`. 
- Edited the docstring based on what i have seen in `huggingface.py`.

hwchase17 (approved):

PR #21239, opened 2024-05-03
mark NemoEmbeddings as deprecated

raspawar:
The NemoEmbeddings is deprecated, instead use langchain-nvidia-ai-endpoints NVIDIAEmbeddings interface.

cc: @mattf


libs/community/langchain_community/embeddings/nemo.py
```diff
@@ -40,14 +41,22 @@ def is_endpoint_live(url: str, headers: Optional[dict], payload: Any) -> bool:
         # Handle any exceptions (e.g., connection errors)
         raise Exception(f"Error querying the endpoint: {e}")
 
-
 class NeMoEmbeddings(BaseModel, Embeddings):
```
baskaryan, line 44: could we use the @deprecated decorator instead? see [example](https://github.com/langchain-ai/langchain/blob/64e17bd793483b63092407cf79d434179436b250/libs/community/langchain_community/chat_models/openai.py#L147)

libs/community/langchain_community/embeddings/nemo.py
```diff
@@ -40,14 +41,22 @@ def is_endpoint_live(url: str, headers: Optional[dict], payload: Any) -> bool:
...
     """NeMo embedding models."""
 
     batch_size: int = 16
     model: str = "NV-Embed-QA-003"
     api_endpoint_url: str = "http://localhost:8088/v1/embeddings"
 
+    @root_validator(pre=True)
+    def raise_deprecation(cls, values: Dict) -> Dict:
+        print(values)
```
mattf, line 53: don't include print() statements

libs/community/langchain_community/embeddings/nemo.py
```diff
@@ -6,6 +6,7 @@
 
 import aiohttp
 import requests
+import warnings
```
mattf, line 9: make sure you `poetry install --with dev,lint,typing,test` and run `make format lint test`

mattf (commented):
thank you for the contribution

i've verified the deprecation notice is visible from
- an app (`python use-nemo-embeddings.py`)
- python repl
- ipython repl

ccurme (approved):

PR #21260, opened 2024-05-03
[community] Added SentenceWindowRetriever

rsk2327:
  - Added appropriate title

Description 
- Adds  a new type of retriever called Sentence Window Retriever 
- Also adds a modification to TextSplitter to help implement the retriever

No tests added yet. Let me know if any specific tests are required. 
Plan to add documentation on how to run the retriever

- [x] **Lint and test**: Tests ran successfully


libs/community/langchain_community/retrievers/sentence_window_retriever.py
```diff
@@ -0,0 +1,184 @@
...
+
+        vector = self.store.embeddings.embed_query(query)  # type: ignore
+
+        top_results = self.store._collection.query(
+            query_embeddings=vector, n_results=self.k
+        )
+
+        doc_list = []
+
+        for id, metadata in zip(top_results["ids"][0], top_results["metadatas"][0]):  # type: ignore
```
eyurtsev, line 80: Let's update the code to handle the case of an empty list of results.

libs/community/langchain_community/retrievers/sentence_window_retriever.py
```diff
@@ -0,0 +1,184 @@
...
+            raise ValueError("The value of window_size must be greater than 0")
+
+        return values
+
+    def _get_relevant_documents(
+        self, query: str, *, run_manager: CallbackManagerForRetrieverRun
+    ) -> List[Document]:
+        """Sync implementations for retriever."""
+
+        if type(self.store) == Chroma:
```
eyurtsev, line 56: MAJOR: Are you able to implement a vectostore agnostic solution? If we can't make it agnostic, we will likely not add this technique to the code base.

My understanding was that this could be solved by using two abstractions:

1) retriever: base retriever
2) document store: an entity that given a document id can return the document content


Would composing or sub-classing `ParentRetriever` work?

libs/text-splitters/langchain_text_splitters/base.py
```diff
@@ -37,6 +37,7 @@ def __init__(
         length_function: Callable[[str], int] = len,
         keep_separator: bool = False,
         add_start_index: bool = False,
+        add_chunk_id: bool = False,
```
eyurtsev, line 40: Named arguments should be introduced at end of list to avoid breaking changes (some users may be invoking by position). If you place it after `strip_whitespace` it'll be an additive change.

I'll need a bit more time to review the change here, but let's see if we can get the code to be vectorstore agnostic?

eyurtsev (changes requested):

libs/community/langchain_community/retrievers/sentence_window_retriever.py
rsk2327: The issue with forcing a vectorstore agnostic implementation is that each vectorstore has its own unique way of retrieving specific texts (either based on their ID or the new 'chunk_id' metadata variable). 

One way we can make it mostly agnostic is by forcing users to add a specific ID variable to chunks ( as the database's ID and not the chunk_id in the metadata). So for the databases that support filtering based on ID, we can have a general solution. 

But for the databases, that dont support such filtering, we might have need a specific implementation. If its crucial to not have custom solutions, then we could decide not to support those databases for this retriever.

libs/community/langchain_community/retrievers/sentence_window_retriever.py
rsk2327: Got it. Will update for that.

libs/text-splitters/langchain_text_splitters/base.py
rsk2327: Got it. Will change the ordering

libs/community/langchain_community/retrievers/sentence_window_retriever.py
eyurtsev: We generally want the implementations to be generic otherwise users will not be able to use them, they become hard to maintain and are hard to test / debug. 

It sounds like before this can be implemented, we need to improve the vectorstore interface first. 

What functionality do you need from vectorstores?

My understanding is that either one of two techniques will work:

```python
AND = {
   'source_id': source_id,
   'chunk_id' : {
     'between': (chunk_id-window_size, chunk_id+window_size)
   }
}
```

OR just a function to get documents by ids: `get_documents_by_ids`

---

The second case requires populating the metadata of each document with information about its nearby neighbors (including distance and doc ids).

---

By the way, you can work around this right now by combining BaseStore and VectorStore abstractions.

rsk2327: So similar to `embeddings.embed_query` which is agnostic of the vectorstore implementation, if there is a common vectorstore functionality that allows for filtering on chunk ID (either the database id or the chunk_id in metadata) wherever available. 

So the `get_documents_by_ids` would be something that could ideally achieve that. The first option depends on the filtering syntax for each database. Unless we implement our own filtering syntax that converts filter arguments to the database specific filtering syntax. That would allow us to have an agnostic implementation for this retriever. 

Im not sure how the BaseStore + VectorStore would help work around the database specific syntax requirements? If you could share more details on that I can try implementing that.

eyurtsev: `BaseStore` allows getting content by ID -- so it provides a way of doing `get_documents_by_id`

rsk2327: Thanks. Let me take a look at that and see if I can implement `get_documents_by_id` for vectorstore.

rsk2327:
@eyurtsev So I have modified the implementation of SWR to be datastore agnostic. 

I went with approach of defining a `get_document_by_ids` function at the vectorstores which enables a common method for querying vectorstore based on IDs. 

One of the issues with the implementation of SWR is that the search functionality is not standardized across vectorstores. 

Chroma : has `similarity_search_by_vector` and 'similarity_search_by_vector_with_score'
Pinecone : does not have `similarity_search_by_vector` but only 'similarity_search_by_vector_with_score'
Milvus : has `similarity_search_by_vector` but instead of  'similarity_search_by_vector_with_score' has 'similarity_search_with_score_by_vector' which is probably a typo


I can work around the the different search method names for now, but it might be helpful for Pinecone to also have a 'similarity_search_by_vector' implementation and for Milvus to use the same standardized function names like the other vectorstores.

libs/text-splitters/langchain_text_splitters/base.py
rsk2327: Updated this.

libs/community/langchain_community/retrievers/sentence_window_retriever.py
rsk2327: The new implementation should be able to handle empty results from the vectorstore.

rsk2327:
@eyurtsev @efriis Could I get a review on this? 

Let me know if I need to add any additional details to explain the changes made. 

I did have a question on how to include the langchain_pinecone as a dependency within community. The unit tests throw an error when I import PineconeVectorStore from langchain_pinecone

rsk2327:
@eyurtsev @efriis Can I get a review on this?

rsk2327:
@eyurtsev @efriis @hwchase17 @baskaryan Can I get a review on this?

libs/partners/pinecone/langchain_pinecone/vectorstores.py
```diff
@@ -200,6 +200,7 @@ def similarity_search_by_vector_with_score(
         k: int = 4,
         filter: Optional[dict] = None,
         namespace: Optional[str] = None,
+        include_id: Optional[bool] = False,
```
eyurtsev, line 203: We don't want to the search API right now.

eyurtsev:
@rsk2327 You'll need to standby for ~1 month. We'll be focusing on the vectorstore abstractions after the 0.2 release. 

The main things so far:

* Addition of  `get_documents_by_ids` to the base abstraction
* Potentially addition of an `id` attribute on a document (so the ID is not randomly in the metadata). 

Text splitters:

* Determine what if any kind of metadata we should be propagating in the text splitter for provenance purposes

---

I'll leave some comments in the PR itself as well

libs/community/langchain_community/vectorstores/chroma.py
```diff
@@ -33,21 +33,34 @@
...
-        )
-    ]
+def _results_to_docs(
+    results: Any, include_id: Optional[bool] = False
+) -> List[Document]:
+    return [doc for doc, _ in _results_to_docs_and_scores(results, include_id)]
+
+
+def _results_to_docs_and_scores(
+    results: Any, include_id: Optional[bool] = False
```
eyurtsev, line 43: nit: code is not properly typed, what is results?

libs/community/langchain_community/vectorstores/chroma.py
```diff
@@ -357,13 +370,15 @@ def similarity_search_by_vector(
         k: int = DEFAULT_K,
         filter: Optional[Dict[str, str]] = None,
         where_document: Optional[Dict[str, str]] = None,
+        include_id: Optional[bool] = False,
```
eyurtsev, line 373: We'll need to be careful in terms of how we deal with the ID, so it can be rolled out throughout the various integrations.

libs/community/langchain_community/vectorstores/chroma.py
```diff
@@ -33,21 +33,34 @@
...
+
+    for result in zip(
+        results["documents"][0],
+        results["metadatas"][0],
+        results["distances"][0],
+        results["ids"][0],
+    ):
+        metadata = result[1] or {}
+        if include_id:
+            metadata["id"] = result[3]
```
eyurtsev, line 57: search logic should not be modifying metadata.

It's OK if it's present during indexing, but shouldn't be mutated on the search path, as the vectorstore should be returning the document as it was indexed.

libs/community/langchain_community/vectorstores/milvus.py
```diff
@@ -1081,3 +1087,40 @@ def upsert(
                 "Failed to upsert entities: %s error: %s", self.collection_name, exc
             )
             raise exc
+
+    def get_documents_by_ids(self, ids: int | str | List[int | str]) -> List[Document]:
+        # Generating filtering expr for passing to query function
```
eyurtsev, line 1092: nit: map into the same code path, or force users to always think in terms of batch (it's good bias since the code involves round trips between client code and server)

if not isinstance(ids, (list, tuple)):
   ids = [ids]

libs/partners/pinecone/langchain_pinecone/vectorstores.py
rsk2327: You might have missed a word in there. Are you suggesting not to include the 'include_id' argument? 

Is this related to you first comment which said that we need to add an 'id' attribute to Document? Which I guess would make the include_id argument unnecessary?

libs/community/langchain_community/vectorstores/chroma.py
rsk2327: Yeah, this was an iffy change. I didnt want to create an entirely new key to store ID so went with just adding it to the metadata. But if we go with your suggestion from the first comment and setup a new ID attribute for documents, that would resolve such issues.

efriis:
closing as stale

icaroryan:
+1 on this PR

PR #21264, opened 2024-05-03
azure-dynamic-sessions: add Python REPL tool

anthonychu:
Adds a Python REPL that executes code in a code interpreter session using Azure Container Apps dynamic sessions.


libs/partners/azure-dynamic-sessions/langchain_azure_dynamic_sessions/tools/sessions.py
```diff
@@ -0,0 +1,256 @@
...
+            size_in_bytes=properties.get("size"),
+        )
+
+
+class SessionsPythonREPLTool(BaseTool):
+    """A tool for running Python code in an Azure Container Apps dynamic sessions
+    code interpreter."""
+
+    name: str = "Sessions_Python_REPL"
+    description: str = (
```
baskaryan, line 93: maybe worth mentioning something about which packages are available? e.g. "You have access to all Python packages included in anaconda xyz"

libs/partners/azure-dynamic-sessions/langchain_azure_dynamic_sessions/tools/sessions.py
```diff
@@ -0,0 +1,256 @@
...
+            f"stdout:\n{properties['stdout']}\n\n"
+            f"stderr:\n{properties['stderr']}"
+        )
+
+    async def _arun(self, python_code: str) -> Any:
+        """Use the tool asynchronously."""
+        if self.sanitize_input:
+            python_code = _sanitize_input(python_code)
+
+        return await run_in_executor(None, self.run, python_code)
```
baskaryan, line 160: think this doesn't need to be implemented, there's a default implementation of arun on BaseTool that call run (and input will be sanitized in run)

libs/partners/azure-dynamic-sessions/langchain_azure_dynamic_sessions/tools/sessions.py
```diff
@@ -0,0 +1,256 @@
...
+                "code": python_code,
+            }
+        }
+
+        response = requests.post(api_url, headers=headers, json=body)
+        response.raise_for_status()
+        response_json = response.json()
+        properties = response_json.get("properties", {})
+        return (
+            f"result:\n{properties['result']}\n\n"
```
baskaryan, line 150: should we filter out images or anything else that doesn't really make sense to have in str?

libs/partners/azure-dynamic-sessions/langchain_azure_dynamic_sessions/tools/sessions.py
```diff
@@ -0,0 +1,256 @@
...
+                "codeInputType": "inline",
+                "executionType": "synchronous",
+                "code": python_code,
+            }
+        }
+
+        response = requests.post(api_url, headers=headers, json=body)
+        response.raise_for_status()
+        response_json = response.json()
+        properties = response_json.get("properties", {})
```
baskaryan, line 148: could we make L126-148 its own public method called something like `def execute(self, python_code) -> dict` so folks have a way of getting structured results if they way? and then `_run` could just be a wrapper that converts structured output to string
```python
    def _run(self, python_code: str) -> str:
        result = self.execute(python_code)
        result = {k: result[k] for k in ("result", "stdout", "stderr")}
        return json.dumps(result, indent=2)
```

libs/partners/azure-dynamic-sessions/langchain_azure_dynamic_sessions/tools/sessions.py
baskaryan: also maybe simpler/better to just `json.dumps(..., indent=2)` the dict instead of custom formatting it? most models do pretty well with reading json formatted outputs

libs/partners/azure-dynamic-sessions/docs/provider.ipynb
baskaryan, line 1: note when time comes we'll want to put this in main `docs/docs/integrations/providers/...` section for it to render in python docs

libs/partners/azure-dynamic-sessions/langchain_azure_dynamic_sessions/tools/sessions.py
```diff
@@ -0,0 +1,256 @@
...
+        properties = data.get("properties", {})
+        return RemoteFileMetadata(
+            filename=properties.get("filename"),
+            size_in_bytes=properties.get("size"),
+        )
+
+
+class SessionsPythonREPLTool(BaseTool):
+    """A tool for running Python code in an Azure Container Apps dynamic sessions
+    code interpreter."""
```
baskaryan, line 90: simple example in main docstring would be nice
```
"""...

Example:
  .. code-block:: python
    
    from langchain_azure_dynamic_sessions import SessionsPythonREPLTool

    tool = SessionsPythonREPLTool(...)
    result = tool.execute("...")
    ...
"""
```

baskaryan (commented):
nice! few small comments. i'm assuming we don't want to add any docs for now?

anthonychu:
> nice! few small comments. i'm assuming we don't want to add any docs for now?

@baskaryan Good question. Let's add proper docs in a separate PR mid-May.

libs/partners/azure-dynamic-sessions/langchain_azure_dynamic_sessions/tools/sessions.py
anthonychu: We're planning to publish a list of packages, once that's exists, maybe we can link to it from here.

efriis (approved):

PR #21266, opened 2024-05-03
langchain: `summarize` chain fix

leo-gan:
Issue: `load_summarize_chain` is placed in the __init__.py file. As a result, it doesn't listed in the API Reference docs.
Change: moved code from __init__.py into a new file.


ccurme (approved):

leo-gan:
@ccurme Thanks!

PR #21279, opened 2024-05-03
langchain: `qa_chain` fix

leo-gan:
Issue: `load_qa_chain` is placed in the __init__.py file. As a result, it is not listed in the API Reference docs.
BTW `load_qa_chain` is heavily presented in the doc examples, but is missed in API Ref.
Change: moved code from init.py into a new file. Related: #21266


ccurme (approved):

leo-gan:
@ccurme Thanks!

PR #21285, opened 2024-05-04
docs: compact the API Reference links

leo-gan:
This PR is opinionated. 
Issue: the `API Reference` sections in the examples hold too much vertical space and make us scroll the page too much. See an [example](https://python.langchain.com/docs/get_started/quickstart/#conversation-retrieval-chain). These sections are **important**. So, the compacting should not make these sections less noticeable.
Change: compacting the `API Reference` sections. See the [same example after change applied](https://langchain-j6nya46lf-langchain.vercel.app/docs/get_started/quickstart/#conversation-retrieval-chain). It is more compact and now looks like references (footnotes).
Note: I would also change the section style, so it would be more noticeable (maybe to look like the footnotes. Smaller wider font?)


baskaryan:
im a big fan! updated slightly to avoid the first `|`, let me know what you think

PR #21300, opened 2024-05-05
added optional 'tool_choice' argument to openai function calling and …

treisfeld:
…tool calling agents

- [ ] **PR message**: 
    - **Description:** enable using `tool_choice` or `function_call` with `create_openai_tool_calling_agent` and `create_openai_function_calling_agent`, respectively

Additional guidelines:

### DIFF
--- a/libs/langchain/langchain/agents/openai_functions_agent/base.py
+++ b/libs/langchain/langchain/agents/openai_functions_agent/base.py
@@ -230,7 +230,10 @@ class OpenAIFunctionsAgent(BaseSingleActionAgent):
 
 
 def create_openai_functions_agent(
-    llm: BaseLanguageModel, tools: Sequence[BaseTool], prompt: ChatPromptTemplate, tool_choice: Optional[BaseTool] = None
+    llm: BaseLanguageModel,
+    tools: Sequence[BaseTool],
+    prompt: ChatPromptTemplate,
+    tool_choice: Optional[BaseTool] = None,
 ) -> Runnable:
     """Create an agent that uses OpenAI function calling.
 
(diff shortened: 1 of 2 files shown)


ccurme:
Can you clarify the use-case for this? A snippet demonstrating its intended use would be helpful.

A common flow for agents (e.g., via AgentExecutor) is for tool execution to loop until no tool calls are generated. If we force tool calling then the agent loop won't naturally terminate.

hwchase17:
> > Can you clarify the use-case for this? A snippet demonstrating its intended use would be helpful.
> > A common flow for agents (e.g., via AgentExecutor) is for tool execution to loop until no tool calls are generated. If we force tool calling then the agent loop won't naturally terminate.
> 
> Hi @ccurme, Thank you for your reply. I believe there are other potential use cases but let me clarify ours: Our team uses an agent to classify requests by populating the relevant arguments of a function designed to handle different request types and extract some associated properties. Here's a simplified snippet of our implementation:
> 
> ```python
> from enum import Enum
> 
> class RequestType(Enum, str):
>     TYPE1 = "type1"
>     TYPE2 = "type2"
> 
> def func_to_choose(
>     request_type: RequestType,
>     first_request_property: str,
>     second_request_property: str
> ):
>     # Function logic here
>     pass
> ```
> 
> In this context, the agent classifies a request into one of several potential types and extracts specific properties from the plain text of the request. OpenAI's specialization in function calling causes this method to significantly improve the agent performance for some use cases compared to the alternatives (i.e, requesting the model to answer in a similar structure without forcing a function call).

for this use case you should use `with_structured_output`. agents are explicitly when the llm chooses the action to take

PR #21304, opened 2024-05-05
core: Extract a generic document store not relying on vector embeddings

cbornet:
The current VectorStore depends on Embeddings in some of its methods (embeddings(), similarity_search_by_vector(), from_texts(), ...)
But most LangChain components using the VectorStore don't use these methods and only do similarity search on strings (eg. SemanticSimilarityExampleSelector, SelfQueryRetriever, ...)
And a number of retrievers use non-vector similarity search and have `add_documents` methods (eg. ElasticSearchBM25Retriever, PineconeHybridSearchRetriever, WeaviateHybridSearchRetriever, ...).. They could be converted to "store" components that could be used with the former components if those components accepted a more generic class.
It is also needed to implement a ColbertStore as the Colbert embedding doesn't fit into the `Embeddings` model.

This PR introduces a `DocumentStore` API from which `VectorStore` derives that doesn't depend on `Embeddings`.
And this new class is used in the components that only do similarity search with a string.

I used `DocumentStore` but document store sometimes also refers to `BaseStore[str, Document]`. So maybe there would be another better name ? Maybe `SimilaritySearchStore` ? Please suggest.


libs/core/langchain_core/document_stores.py
```diff
@@ -0,0 +1,553 @@
...
+
+
+class DocumentStore(ABC):
+    """Interface for document store."""
+
+    @abstractmethod
+    def add_texts(
+        self,
+        texts: Iterable[str],
+        metadatas: Optional[List[dict]] = None,
```
epinzur, line 38: should this also have a list of optional ids, to match the delete method?

eyurtsev:
@cbornet we'll need a bit of time before we're ready to integrate this.

@baskaryan and I are planning on reviewing vectorstore and retriever abstractions to determine what we want to have longer-term.

We're preparing for a 0.2 release soon, so we likely won't have many free cycles for the next ~2 weeks for reviewing this carefully

libs/core/langchain_core/document_stores.py
cbornet: This PR is about moving the method to a super class.
If modifications to the API are to be done, I think they should be done in another PR.
Also, since not all vector stores support passing explicit IDs when inserting documents, I'm not sure it is a good thing to add it to the core API.

PR #21307, opened 2024-05-05
chore(community): reduce number of list creations and calls in Azure AI Search

mspronesti:
This PR aims at making the `AzureSearch` vectorstore a bit more efficient by reducing the number of calls and lists being created by the corresponding retriever. 

The current version operates as follows:
- a client wrapper returns a `List[Document, float]` containing all the matching documents with the corresponding score
- then a second method is called to create a new list after filtering out according to a score threshold
- finally a third list is created dropping the score "column".

Here's an example:
image

However one could avoid the intermediate step by filtering directly in the first step (if a threshold is passed). One could even skip the third step specifying whether to include the scores or not, but this would not be consistent with the naming conventions of many methods (starting from the base class), thus I decided not to do that.

### DIFF
--- a/libs/community/langchain_community/vectorstores/azuresearch.py
+++ b/libs/community/langchain_community/vectorstores/azuresearch.py
@@ -398,9 +398,9 @@ class AzureSearch(VectorStore):
         self, query: str, k: int = 4, **kwargs: Any
     ) -> List[Document]:
         search_type = kwargs.pop("search_type", self.search_type)
-        if search_type == "similarity":
+        if search_type in ("similarity", "similarity_score_threshold"):
             docs = self.vector_search(query, k=k, **kwargs)
-        elif search_type == "hybrid":
+        elif search_type in ("hybrid", "hybrid_score_threshold"):
             docs = self.hybrid_search(query, k=k, **kwargs)
         elif search_type == "semantic_hybrid":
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -397,9 +398,9 @@ def similarity_search(
         self, query: str, k: int = 4, **kwargs: Any
     ) -> List[Document]:
         search_type = kwargs.get("search_type", self.search_type)
-        if search_type == "similarity":
+        if search_type in ("similarity", "similarity_score_threshold"):
             docs = self.vector_search(query, k=k, **kwargs)
-        elif search_type == "hybrid":
+        elif search_type in ("hybrid", "hybrid_score_threshold"):
```
mspronesti, line 403: the `*_score_threshold` search type is now automatically handled passing the `score_threshold` kwarg to the search method. Otherwise, we don't filter the scores (as expected).

mspronesti (commented):
Left a few comments to (hopefully) help the review process -- GitHub's diff seems to "overcomplicate" the changes of this PR.

mspronesti:
@ccurme @baskaryan

mspronesti:
This PR is way too outdated at this point and resolving the conflicts could take time. I will be more than happy to do so

PR #21314, opened 2024-05-06
[Community][Minor]: Fetch loader_source of GoogleDriveLoader in PebbloSafeLoader.

rahul-trip:
Description: This PR includes fix for loader_source to be fetched from metadata in case of GdriveLoaders.
Documentation: NA
Unit Test: NA

### DIFF
--- a/libs/community/langchain_community/utilities/pebblo.py
+++ b/libs/community/langchain_community/utilities/pebblo.py
@@ -236,6 +236,27 @@ def get_loader_full_path(loader: BaseLoader) -> str:
             location = "in-memory"
         elif isinstance(loader, NotionDBLoader):
             location = f"notiondb://{loader.database_id}"
+        elif loader.__class__.__name__ == "GoogleDriveLoader":
+            if loader_dict.get("folder_id"):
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/utilities/pebblo.py
```diff
@@ -236,6 +236,27 @@ def get_loader_full_path(loader: BaseLoader) -> str:
             location = "in-memory"
         elif isinstance(loader, NotionDBLoader):
             location = f"notiondb://{loader.database_id}"
+        elif loader.__class__.__name__ == "GoogleDriveLoader":
```
hwchase17, line 239: why is this checking the class name and not `isinstance`?

rahul-trip: It was to avoid the unnecessary import of GoogleDriveLoader, I am planning to remove other imports and check the names for them too.

Also, particularly in case of GoogleDriveLoader, they are present in langchain_community and langchain_google both the packages.

hwchase17 (approved):

PR #21334, opened 2024-05-06
community: Vectara Integration Update - Streaming, FCS, Chat, updates to documentation and example notebooks

ofermend:

**Description:** update to the Vectara / Langchain integration to integrate new Vectara capabilities:
- Full RAG implemented as a Runnable with as_rag()
- Vectara chat supported with as_chat()
- Both support streaming response
- Updated documentation and example notebook to reflect all the changes
- Updated Vectara templates

**Twitter handle:** ofermend

**Add tests and docs**: no new tests or docs, but updated both existing tests and existing docs


ofermend:
Hey @efriis and @baskaryan - just checking about your thoughts on this PR?

ofermend:
@baskaryan @efriis - just updated reranking, added tests, and merged with master so it's up-to-date.
Any further thoughts or feedback so far?

libs/community/langchain_community/vectorstores/vectara.py
```diff
@@ -28,29 +34,30 @@ class SummaryConfig:
...
-    max_results: int = 7
+    max_results: int = 5
     response_lang: str = "eng"
-    prompt_name: str = "vectara-summary-ext-v1.2.0"
+    stream: bool = True
+    prompt_name: str = "vectara-summary-ext-24-05-sml"
 
 
 @dataclass
-class MMRConfig:
```
efriis, line 37: I believe this is a breaking change for anyone currently using `MMRConfig` - could we add some kind of backwards compatibility version that allows the old interface as well?

libs/community/langchain_community/vectorstores/vectara.py
```diff
@@ -317,69 +363,101 @@ def add_texts(
             )
         return [doc_id]
 
-    def vectara_query(
```
efriis, line 320: any reason someone would have used this before?

libs/community/langchain_community/vectorstores/vectara.py
```diff
@@ -317,69 +363,101 @@ def add_texts(
...
+    def _get_query_body(
```
efriis, line 366: thanks for underscore! Will prevent changes to this interface being breaking in future

libs/community/langchain_community/vectorstores/vectara.py
```diff
@@ -66,21 +73,47 @@ class VectaraQueryConfig:
     score_threshold: minimal score threshold for the result.
         If defined, results with score less than this value will be
         filtered out.
-    n_sentence_context: number of sentences before/after the matching segment
+    n_sentence_before: number of sentences before the matching segment
```
efriis, line 76: some of these attribute changes are breaking [removals + changed defaults]. Is there a way to do this in a nonbreaking way?

efriis (commented):
would it be better to just implement this as a `vectara_v2` retriever instead? There's a few breaking changes in here that will cause issues if anyone is using this currently.

Also - am I correct in my understanding that this only implements streaming for the generic VectaraRAG `Runnable` and not for a retriever?

libs/community/langchain_community/vectorstores/vectara.py
ofermend: Not at all. That's why I realized it's best to rename as such.

libs/community/langchain_community/vectorstores/vectara.py
ofermend: Great point and thanks for pointing that out. 
I made some changes to make everything backwards compatible. In the process also added test_cases to also test it works in the new form and in the old form, and also updated default values to ensure nothing is a breaking change.

libs/community/langchain_community/vectorstores/vectara.py
ofermend: Not at all. I think this one is safe to change.

libs/community/langchain_community/vectorstores/vectara.py
ofermend: See your broader comment on breaking change. New update should reflect support for backwards compatibility.

ofermend:
Thanks @efriis for all the great comments and suggestions to making this backwards compatible.
I did a lot of work to make it so today, including adding tests for the same. Please review and let me know if I missed any other point there.
I don't think we need this to be retriever_v2 as these changes address backwards compatibility, but please let me know if you think otherwise.

And yes streaming is only for VectaraRAG. I am unsure how streaming from Vectara could/should work for the retriever alone (which just acts as semantic search) - did you have anything specific in mind here?

damianoneill:
@ofermend, thank you for this PR looking forward to using these updates, glad to see that you made these changes backwards compatible.  

@efriis is there anything additional required in this PR? Apologies, I'm not familiar with the merge process/release schedule, just looking to understand when this would be available in a release?

baskaryan (approved):

PR #21337, opened 2024-05-06
together: add chat models, use openai base

Nutlope:
**Description:** Adding chat completions to the Together AI package, which is our most popular API. Also staying backwards compatible with the old API so folks can continue to use the completions API as well. Also moved the embedding API to use the OpenAI library to standardize it further.

**Twitter handle:** @nutlope


libs/partners/together/langchain_together/__init__.py
efriis, line 1: Can we add back a `__version__`? Can copy it to use metadata lib like other partner packages if helpful

libs/partners/together/langchain_together/completions.py
```diff
@@ -109,7 +108,7 @@ def _format_output(self, output: dict) -> str:
 
     @staticmethod
     def get_user_agent() -> str:
-        return f"langchain-together/{__version__}"
+        return "langchain-together/0.3"
```
efriis, line 111: let's rely on `__version__` if that works for you. Otherwise more difficult to maintain

libs/partners/together/langchain_together/chat_models.py
```diff
@@ -0,0 +1,103 @@
...
+    Optional,
+)
+
+import openai
+from langchain_core.pydantic_v1 import Field, SecretStr, root_validator
+from langchain_core.utils import (
+    convert_to_secret_str,
+    get_from_dict_or_env,
+)
+from langchain_openai import ChatOpenAI
```
efriis, line 17: Should inherit from `BaseChatOpenAI` instead

libs/partners/together/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [tool.poetry]
 name = "langchain-together"
-version = "0.1.0"
-description = "An integration package connecting Together and LangChain"
+version = "0.3.0"
```
efriis, line 3: Is this a breaking change? Could we do `0.1.1` instead? Any reason we're skipping 0.2?

libs/partners/together/tests/integration_tests/test_chat_models.py
efriis, line 1: Let's rename these to be `together` instead of `upstage`

libs/partners/together/tests/integration_tests/test_llms.py
efriis, line 1: could we keep these to confirm backwards compatibility?

efriis (commented):
Looks good! Main point is about whether this is a breaking change or not (I think it should not be), in which case we can make this a 0.1.1?

High level:
- find and replace "upstage" with "together"

libs/partners/together/pyproject.toml
Nutlope: You're right, not a breaking change! Will change to 0.1.1

libs/partners/together/tests/integration_tests/test_chat_models.py
Nutlope: oops forgot about those, done!

libs/partners/together/tests/integration_tests/test_llms.py
Nutlope: Good call, added it back under `test_completions.py`!

efriis: Great! If we can keep it as `test_llms` that would be best (that's the langchain name for text completions endpoints, as opposed to chat models)

libs/partners/together/langchain_together/__init__.py
Nutlope: Couldn't find `__version` in other partner packages, checked OpenAI, groq, nomic, ect... How do I add this properly?

Nutlope: Done!

libs/partners/together/langchain_together/completions.py
Nutlope: Makes sense, done!

Nutlope: Should we add this in all methods? It's currently only there for completions

libs/partners/together/langchain_together/chat_models.py
Nutlope: Says langchain_openai doesn't have a module called that. Plus I saw this was the way upstage was doing it. Where should I import that from?

libs/partners/together/tests/integration_tests/test_llms.py
Nutlope: Done!

libs/partners/together/langchain_together/completions.py
efriis: Up to you all! If you're not using this data on your side, feel free to remove this.

libs/partners/together/langchain_together/chat_models.py
efriis: `from langchain_openai.chat_models.base import BaseChatOpenAI`

Will ping the upstage folks to do the same. Basically if you rely on `ChatOpenAI` it will throw an error if you don't pass/have an environment variable for `OPENAI_API_KEY`. Using the base one removes that restriction, such that you can read `together_api_key` instead.

libs/partners/together/langchain_together/completions.py
Nutlope: We're not using it so I'll just remove, thanks!

libs/partners/together/langchain_together/chat_models.py
Nutlope: Ahh thanks for the explanation! Done

efriis (approved):

PR #21342, opened 2024-05-06
fix: core: Include in json output also fields set outside the constructor

nfcampos:


libs/core/langchain_core/load/serializable.py
```diff
@@ -128,8 +128,9 @@ def to_json(self) -> Union[SerializedConstructor, SerializedNotImplemented]:
         # Get latest values for kwargs if there is an attribute with same name
         lc_kwargs = {
             k: getattr(self, k, v)
-            for k, v in self._lc_kwargs.items()
```
hinthornw, line 131: Are we deprecating `self. _lc_kwargs` then? Do we need it around?

libs/core/langchain_core/load/serializable.py
```diff
@@ -186,6 +187,23 @@ def to_json_not_implemented(self) -> SerializedNotImplemented:
...
+        key: The key.
+        value: The value.
+
+    Returns:
+        Whether the field is useful.
+    """
+    field = inst.__fields__.get(key)
+    if not field:
+        return False
+    return field.required is True or field.get_default() != value
```
hinthornw, line 204: Hmmm for the second condition, if we do something like:
```
ChatOpenAI(model="gpt-3.5-turbo").to_json()
```
the model name is elided.

We really shouldn't have put default model names in these classes, but now that that bridge has been crossed it seems not great to then lose that info in the serialized output

hinthornw (approved):
TY!

PR #21349, opened 2024-05-06
Add source metadata to bedrock retriever response

rozerarenu:

- [X] **PR title**: "community: Add source metadata to bedrock retriever response"

- [X] **PR message**: 
    - **Description:** Bedrock retrieve API returns extra metadata in the response which is currently not returned in the retriever response
    - **Issue:** The change adds the metadata from bedrock retrieve API response to the bedrock retriever in a backward compatible way. Renamed metadata to sourceMetadata as metadata term is being used in the Document already. This is in sync with what we are doing in llama-index as well.
    - **Dependencies:** No

- [X] **Add tests and docs**:
  1. Added unit tests
  2. Notebook already exists and does not need any change
  3. Response from end to end testing, just to ensure backward compatibility: `[Document(page_content='Exoplanets.', metadata={'location': {'s3Location': {'uri': 's3://bucket/file_name.txt'}, 'type': 'S3'}, 'score': 0.46886647, 'source_metadata': {'x-amz-bedrock-kb-source-uri': 's3://bucket/file_name.txt', 'tag': 'space', 'team': 'Nasa', 'year': 1946.0}})]`

Additional guidelines:


libs/community/langchain_community/retrievers/bedrock.py
```diff
@@ -115,13 +115,16 @@ def _get_relevant_documents(
         results = response["retrievalResults"]
         documents = []
         for result in results:
+            content = result["content"]["text"]
+            result.pop("content")
+            if "score" not in result:
+                result["score"] = 0
+            if "metadata" in result:
+                result["sourceMetadata"] = result.pop("metadata")
```
3coins, line 123: Can you use snake case instead of camel case here.

```suggestion
                result["source_metadata"] = result.pop("metadata")
```

3coins (commented):
@rozerarenu 
Thanks for making the change. One minor comment, can you also make sure this PR is in alignment with changes here.
https://github.com/langchain-ai/langchain-aws/pull/36

libs/community/langchain_community/retrievers/bedrock.py
rozerarenu: With the approach that we have taken to populate the metadata here, if we add a new field in the [retrieve API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_Retrieve.html#API_agent-runtime_Retrieve_ResponseElements) in the future that will get added to the metadata automatically and that field name will be in camelCase. So having sourceMetadata in snake case will bring an inconsistency in the namings. We would like this integration to work out of the box for new additions in the bedrock APIs. Hence we should keep camelCase in my opinion.

3coins (approved):
Code looks good! Can you please fix the lint errors.

3coins:
@baskaryan 
Can you help merge this PR. Seems like @rozerarenu has attempted to fix the CI issues, but the CI did not start after her changes, can you help with this. Thanks.

ccurme (approved):

ccurme:
added back the unit tests here: https://github.com/langchain-ai/langchain/pull/21485

PR #21353, opened 2024-05-07
together: fix chat model and embedding classes

efriis:


libs/partners/together/langchain_together/embeddings.py
efriis, line 1: @Nutlope just want to make sure I'm not messing anything up by removing the logic appending `-query` and `-passage` to the models! It doesn't seem to work with the default `togethercomputer` models.

Nutlope: You're right! These shouldn't have been there, thanks

Nutlope (approved):
LGTM!

PR #21394, opened 2024-05-08
[Bugfix] Make ObjectId JSON-serializable on generation

Jibola:

- **Description:** MongoDB ObjectId being non-JSON Serializable creates errors in our code generation.
- **Issue:** [#535](https://github.com/langchain-ai/langserve/issues/535)
- **Dependencies:** dbx/python team (@caseyclements, @blink1073 , @shaneharvey, @NoahStapp)
- **Twitter handle:** @mongodb

- [x] **Add tests and docs**: Added a component to the unit test that ensures the resulting metadata from an Vector Serach result is JSON Serializable. 

### DIFF
--- a/libs/partners/mongodb/langchain_mongodb/vectorstores.py
+++ b/libs/partners/mongodb/langchain_mongodb/vectorstores.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import math
 from importlib.metadata import version
 from typing import (
     Any,
@@ -30,8 +31,6 @@ MongoDBDocumentType = TypeVar("MongoDBDocumentType", bound=Dict[str, Any])
 VST = TypeVar("VST", bound=VectorStore)
 
 logger = logging.getLogger(__name__)
-
-DEFAULT_INSERT_BATCH_SIZE = 100
 
 
 class MongoDBAtlasVectorSearch(VectorStore):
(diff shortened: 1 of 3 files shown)


hwchase17:
would love a review from Casey!

libs/partners/mongodb/tests/unit_tests/test_vectorstores.py
```diff
@@ -75,6 +76,9 @@ def _validate_search(
         output = vectorstore.similarity_search("", k=1)
         assert output[0].page_content == page_content
         assert output[0].metadata.get("c") == metadata
+        # Validate the ObjectId provided is json serializable
+        assert loads(dumps(output[0].page_content)) == output[0].page_content
+        assert loads(dumps(output[0].metadata)) == output[0].metadata
```
caseyclements, line 81: And then here you can show how this would be used in practice.
Notice that I'm even using bson.json_util.loads on a string dumped using plain json.dumps

```
        from bson import json_util, ObjectId
        json_metadata = dumps(output[0].metadata). # normal json.dumps 
        assert isinstance(json_util.loads(json_metadata)["_id"], ObjectId)
```

caseyclements (changes requested):
This looks good. It solves the reported problem.

libs/partners/mongodb/tests/unit_tests/test_vectorstores.py
Jibola: I think most folks in the LangChain workflow would end up using either `orjson` or `json` library especially as the underlying LangChain "pipes" leverage either of those two. 

I'll add this in as I think it's still valuable.

libs/partners/mongodb/langchain_mongodb/vectorstores.py
```diff
@@ -213,6 +214,10 @@ def _similarity_search_with_score(
         for res in cursor:
             text = res.pop(self._text_key)
             score = res.pop("score")
+            # Make every ObjectId found JSON-Serializable
```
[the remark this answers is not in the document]
Jibola: It's unlikely many people would want to store a RAG Chain object that has an additional ObjectId outside of the Id as an item. With that being said, I'll still go ahead and add the ability to serialize nested values.

caseyclements (approved):
LGTM

ShaneHarvey (approved):

efriis (approved):

Issue #21354, opened 2024-05-07
module 'playwright.async_api' has no attribute 'AsyncBrowser' 


mehrdad-shokri:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_community.agent_toolkits import PlayWrightBrowserToolkit
from langchain_community.tools.playwright.utils import create_async_playwright_browser

PlayWrightBrowserToolkit.from_browser(
        async_browser=create_async_playwright_browser(headless=True, args=["--disable-dev-shm-usage"])).get_tools()
```

### Error Message and Stack Trace (if applicable)

"module 'playwright.async_api' has no attribute 'AsyncBrowser'"

### Description

After 1 hour of investigation here is my findings:  
- The issue is introduced by [this commit](https://github.com/langchain-ai/langchain/commit/9639457222afac372de7ef8fa722434e7692935a)
-  As you [can](https://github.com/microsoft/playwright-python/blob/release-1.43/playwright/async_api/__init__.py) [see](https://github.com/microsoft/playwright-python/blob/release-1.43/playwright/async_api/__init__.py) from playwright-python source code, there is no AsyncBrowser or SyncBrowser exports. They were correctly imported ``Browser as AsyncBrowser`` ``Browser as SyncBrowser``  in the past but [this](https://github.com/langchain-ai/langchain/pull/21156/commits/ec8df9e2de997d4cf3a8f0633166b3634bdaaeed#diff-a98f55f30d9f1733f6e399517dd5fbc4d74e9a16ad0840d3b51e3e3fd1ca59b9) PR broke it.
- As a side note to the mentioned PR, the ``lazy_import_playwright_browsers`` uses ``guard_import`` for both of the browsers(although the name of the class called from the module is incorrect), but still ``guard_import`` is called inside ``validate_browser_provided`` and ``from_browser`` again. These 2 functions can just call ``lazy_import_playwright_browsers`` instead of repeating your self 6 times. Just call ``lazy_import_playwright_browsers`` inside the mentions methods.
- The ``lazy_import_playwright_browsers`` body should look like:
```python
[...]

PR #21395, opened 2024-05-08
bugfix(community): fix Playwright import paths.

mehrdad-shokri:
- **Description:**  Fix import class name exporeted from 'playwright.async_api' and 'playwright.sync_api' to match the correct name in playwright tool. Change import from inline guard_import to helper function that calls guard_import to make code more readable in gmail tool. Upgrade playwright version to 1.43.0
- **Issue:** #21354
 - **Dependencies:** upgrade playwright version(this is not required for the bugfix itself, just trying to keep dependencies fresh. I can remove the playwright version upgrade if you want.)


hwchase17 (approved):

hwchase17:
thanks!

PR #21404, opened 2024-05-08
community: Support ChatZhipuAI with agent tools call

migo-robben:
Description: This update fixed the problem that the ChatZhipuAI module can't call tools, now it can support tool calls in the agent.
```python
from langchain_community.chat_models import ChatZhipuAI
# from ... import ...

# Add, Divide is your tools
tools = [Add(), Divide()]
openai_tools_agent_prompt = hub.pull("hwchase17/openai-tools-agent")

chat = ChatZhipuAI(model="glm-3-turbo")
agent = create_openai_tools_agent(chat, tools, openai_tools_agent_prompt)
result = agent_executor.invoke(
    {
        "input": "What is the result of adding 12.5 to the division of 3 by 15?"
    }
)
```

### DIFF
--- a/libs/community/langchain_community/chat_models/zhipuai.py
+++ b/libs/community/langchain_community/chat_models/zhipuai.py
@@ -107,7 +107,10 @@ def _convert_dict_to_message(dct: Dict[str, Any]) -> BaseMessage:
         additional_kwargs = {}
         if "name" in dct:
             additional_kwargs["name"] = dct["name"]
-        return ToolMessage(content=content, tool_call_id=dct["tool_call_id"], additional_kwargs=additional_kwargs)
+        return ToolMessage(
+            content=content,
+            tool_call_id=dct["tool_call_id"],
+            additional_kwargs=additional_kwargs)
     return ChatMessage(role=role, content=content)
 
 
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/chat_models/zhipuai.py
```diff
@@ -133,19 +151,22 @@ def _convert_delta_to_message_chunk(
     role = dct.get("role")
     content = dct.get("content", "")
     additional_kwargs = {}
-    tool_calls = dct.get("tool_call", None)
+    tool_calls = dct.get("tool_calls", None)
```
ccurme, line 154: was `dct.get("tool_call", None)` a bug earlier?

migo-robben: Yes, when the return is a tool call, zhipuai use "tool_calls" as the key

efriis:
Sorry for the delay on this one! I think the bug may have been fixed in a different PR and if not, would recommend moving this integration over to a dedicated integration package.

If the bug still remains in latest community, feel free to reopen against latest master with just the bugfix!

Issue #11945, opened 2023-10-17
Streaming support with ConversationChain


marcocello:
### System Info

langchain          0.0.316
langserve          0.0.10
python               3.11.4 on darwin

### Who can help?

@agola11 @hwchase17 

### Information

- [X] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [X] Async

### Reproduction

Langserve example here (https://github.com/langchain-ai/langserve-launch-example/blob/main/langserve_launch_example/chain.py) in which I want to use ConversationChain instead of ChatOpenAI.

server.py

    #!/usr/bin/env python
    """A server for the chain above."""

    from fastapi import FastAPI
    from langserve import add_routes

    from chain import conversation_chain

    app = FastAPI(title="My App")

    add_routes(app, conversation_chain)

    if __name__ == "__main__":
        import uvicorn

        uvicorn.run(app, host="localhost", port=8000)

chain.py
```
from langchain.llms import OpenAI`
from langchain.chains import ConversationChain
from langchain.prompts import PromptTemplate
from langchain.memory import ConversationBufferMemory

template = """Assistant is a large language model trained by OpenAI.

Assistant is designed to be able to assist with a wide range of tasks, from answering simple questions to providing in-depth explanations and discussions on a wide range of topics. As a language model, Assistant is able to generate human-like text based on the input it receives, allowing it to engage in natural-sounding conversations and provide responses that are coherent and relevant to the topic at hand.

{history}
Human: {input}
Assistant:"""

prompt = PromptTemplate(input_variables=["history", "input"], template=template)

conversation_chain = ConversationChain(
    llm=OpenAI(temperature=0),
    prompt=prompt,
    verbose=True,
    memory=ConversationBufferMemory(),
)
```

### Expected behavior

I was expecting to have a streaming response as the ChatOpenAI behavior. It seems to me that ConversationChain doesn't support streaming response.

dmahlow:
Are there any updates on this? Is ConversationChain not being able to be streamed from a bug or as expected? If the former, what's a good way of fixing this?

[...]

PR #21413, opened 2024-05-08
core: add a `no_collect: bool=False` param to RunnableLambda, enables per-chunk processing instead of whole collected input

rhighs:
This PR introduces a new param `no_collect: bool` to `RunnableLambda.__init__`. This flags allows for the `_transform` and `_atransform` methods to not collect input before processing but rather to process each individual chunk via `self.func/self.afunc`. This is particularly useful when we want to keep the input stream flowing as it gets transformed. One specific issue this implementation has solved on my part is being able to have a streamed conversation chain whilst being able to save the final chain output in a context memory. This is a problem some other ppl have encountered in this issue: #11945

Example usage:
```python
def generate_values(input: Iterator[Any]) -> Iterator[int]:
    yield 1
    yield 2
    yield 3
    yield 4

def square(input: Input) -> int:
    n = cast(int, input)
    return n * n

expected_values = [1, 4, 9, 16]
runnable = RunnableGenerator(generate_values) | RunnableLambda(
   square, no_collect=True
)
for i, value in enumerate(runnable.stream({"input": "some_input"})):
    assert value == expected_values[i]
```

twitter: @rhighs_

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -640,8 +640,7 @@ class Runnable(Generic[Input, Output], ABC):
         *,
         return_exceptions: Literal[False] = False,
         **kwargs: Any,
-    ) -> Iterator[Tuple[int, Output]]:
-        ...
+    ) -> Iterator[Tuple[int, Output]]: ...
 
     @overload
     def batch_as_completed(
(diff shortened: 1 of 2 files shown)


ccurme:
cc @nfcampos

hwchase17:
We have runnable generator for this, we will add a how to guide

rhighs:
Sounds good

PR #21420, opened 2024-05-08
community: Chroma Adding create_collection_if_not_exists flag to Chroma constructor

tazarov:
- **Description:** Adds the ability to either `get_or_create` or simply `get_collection`. This is useful when dealing with read-only Chroma instances where users are constraint to using `get_collection`. Targeted at Http/CloudClients mostly.
- **Issue:** chroma-core/chroma#2163
- **Dependencies:** N/A
- **Twitter handle:** `@t_azarov`

| Collection Exists | create_collection_if_not_exists | Outcome                                                        | test                                                     |
|-------------------|---------------------------------|----------------------------------------------------------------|----------------------------------------------------------|
| True              | False                           | No errors, collection state unchanged                          | `test_create_collection_if_not_exist_false_existing`     |
| True              | True                            | No errors, collection state unchanged                          | `test_create_collection_if_not_exist_true_existing`      |
| False             | False                           | Error, `get_collection()` fails                                | `test_create_collection_if_not_exist_false_non_existing` |
| False             | True                            | No errors, `get_or_create_collection()` creates the collection | `test_create_collection_if_not_exist_true_non_existing`  |


libs/partners/chroma/tests/integration_tests/test_vectorstores.py
```diff
@@ -414,3 +420,73 @@ def test_chroma_legacy_batching() -> None:
...
+            collection_name="test_collection",
+            embedding_function=FakeEmbeddings(),
+            create_collection_if_not_exists=False,
+        )
+
+
+def test_create_collection_if_not_exist_true_non_existing(
+    client: chromadb.ClientAPI,
+) -> None:
+    """Tests create_collection_if_not_exists=True and collection non-existing. ."""
```
ccurme, line 482: Is this test correct? we have `create_collection_if_not_exists=False` below, and we create the collection.

tazarov: I think so, yes. if `create_collection_if_not_exists=False`, we'll do `get_collection` so the collection must exist; otherwise, Chroma will throw an error.

tazarov: This is easier to implement with hypothesis, but perhaps on the next iteration.

ccurme: Isn't this test identical to `test_create_collection_if_not_exist_false_existing`?

Sorry if I'm missing something.

ccurme (approved):

libs/partners/chroma/tests/integration_tests/test_vectorstores.py
tazarov: Nope, you're not missing anything. You are right that the test is identical even though its intended test scope (docstring) differs. Let me fix that.

tazarov:
@ccurme, I've added truth table with test case mapping in description to make it a little easier to understand

ccurme (approved):

harshal-cuminai:
@efriis  We are taking dependency on langchain-chroma and have to publish a package to pypi that needs this change. Can we expedite the next release of this partner package to pypi ? When is the planned release ?

ccurme:
@harshal-cuminai langchain-chroma 0.1.1 was released yesterday.

PR #21427, opened 2024-05-08
Community: Fix in MultiQuery Retriever to parse the generated queries in the List[str]

jsourabh1:
- **Description:**  Fixes parsing of generated queries in the MultiQuery Retriever.

- **Issue:** The MultiQueryRetriever was not correctly parsing the generated queries returned by the language model, which should be a list of strings (List[str]).

- **Fix:** This PR updates the code to handle cases where the language model response might not have a "text" key directly. It now uses getattr to access the "text" key or an alternative key specified by the parser_key attribute. This ensures the retrieved queries are always a list of strings.


hwchase17:
```
"""DEPRECATED. parser_key is no longer used and should not be specified."""
```
Not sure why you are getting this error, but if you use `from_llm` it should construct the LLMChain in the correct way

PR #21436, opened 2024-05-08
langchain-robocorp: remove toolkit return content max length

tonnitommi:
Robocorp (action server) toolkit had a limitation that the content length returned by the tool was always cut to max 5000 chars. This was from the time when context windows were much more limited.

This PR removes the limitation. Whatever the underlying tool provides gets sent back to the agent.

As the robocorp toolkit no longer restricts the content, the implication is that either the Action (tool) developer or the agent developer needs to be aware of potentially oversized tool responses. Our point of view is this should be the agent developer's responsibility, them being in control of the use case and aware of the context window the LLM has.

### DIFF
--- a/libs/partners/robocorp/langchain_robocorp/toolkits.py
+++ b/libs/partners/robocorp/langchain_robocorp/toolkits.py
@@ -27,7 +27,6 @@ from langchain_robocorp._prompts import (
     API_CONTROLLER_PROMPT,
 )
 
-MAX_RESPONSE_LENGTH = 5000
 LLM_TRACE_HEADER = "X-action-trace"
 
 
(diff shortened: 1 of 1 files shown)


ccurme (approved):

ccurme:
Thanks @tonnitommi. It looks like we haven't released `langchain-robocorp` in about [a month](https://pypi.org/project/langchain-robocorp/#history). I'll plan to release in an hour to pick up this change, but let me know if you have additional changes you want to get in ahead of that (can also cut additional releases).

tonnitommi:
Thanks @ccurme, we are happy with this change for now. :)

Swift turnaround is very much appreciated here in our camp! 🙇

ccurme:
Hi @tonnitommi, this is now released in `langchain-robocorp==0.0.6`. Would you mind testing to confirm the issue is resolved? Thanks.

tonnitommi:
@ccurme sorry took a while, but the PR indeed fixed what it was supposed to do. :)

PR #21438, opened 2024-05-08
docs: v0.2 docs in master

efriis:
current python.langchain.com is building from branch `v0.1`. Iterate on v0.2 docs here.


efriis:
Changes are only in docs. This is failing CI because change log is too large. Ran tests locally, and will monitor for failures in master.

PR #21445, opened 2024-05-08
community: Fix malformed output in `GitHubAPIWrapper.get_files_from_directory`

joshuasundance-swca:
# Description
Resolves an issue where the `get_files_from_directory` method produced malformed outputs due to incorrect handling of string concatenation during recursive directory traversal. The function has been redesigned to ensure correct type handling and output formatting.

# Changes Made
1. **Implementation of an Inner Function (`_inner`)**:
   - A new inner function `_inner` manages the recursion, handling the traversal through directories and the accumulation of file paths.
   - This change prevents the improper extension of a list with string characters, which was previously causing malformed outputs.

2. **Output Formatting**:
   - The output from the `_inner` function is now consistently a list of file paths, converted to a string after all paths are collected, ensuring the function's return value is consistent with its documentation.

## Example of Previous Malfunction
Previously, if the function was called on a directory structure like this:
```
/root
  /subdir
    file1.txt
```
The expected output would ideally be `['root/subdir/file1.txt']`. However, due to the misuse of list extension with a string, the output was malformed, resembling something like `['[', 'r', 'o', 'o', 't', '/', ...]` — effectively breaking down the file path into individual characters.


ccurme:

PR #21448, opened 2024-05-08
Harrison/move flashrank rerank

hwchase17:
third party integration, should be in community


mspronesti (commented):
~Out of curiosity, why is this being moved in `community` -which makes sense to me- but  `CohereReranker` isn't?~

EDIT: my bad, I've just seen you have deprecated `CohereReranker` in favor of `langchain-cohere`'s.

baskaryan (approved):

PR #21459, opened 2024-05-09
docs: spell correction

efriis:


hinthornw:
Is this something we could add to our docs:
```
# We do this to avoid spellchecking cell outputs
          - name: Codespell Notebooks
            run: |
              find . -name "*.ipynb" | head -n 1 | xargs cat $1 | jupytext --from ipynb --to py:percent | codespell -
```

PR #21469, opened 2024-05-09
Pass through Run ID Explicitly

hinthornw:


HoangNguyen689:
@hinthornw 
May I ask that when these changes will be release?
I saw the release of langchain 0.1.20 but didn't found the fix in that.
There isn't any release notes, it is difficult.

HoangNguyen689:
@hinthornw 
I found it on 0.2.0rc.
I guess it may take time.

PR #21496, opened 2024-05-09
docs: api reference build fix

leo-gan:
Issues: 1. The `create_api_rst.py` fails on the `standard-tests` directory. 2. An error in building one package stops all build. Build should proceed with building the next package.
Fix: added `standard-tests` directory to the exception list. Added error handling for #2


efriis (commented):
Agreed we should add standard tests to the build, but I don't think the whole build should proceed if a package fails? Isn't it better it fails loudly so we can go and fix it?

This runs daily to build api.python.langchain.com, and we get notifications when the build fails

leo-gan:
@efriis Your comments addressed.

leo-gan:
outdated

PR #21509, opened 2024-05-09
langchain, community: remove cap on sqlalchemy and bump duckdb

ccurme:


libs/community/pyproject.toml
```diff
@@ -115,7 +115,7 @@ optional = true
 pytest = "^7.3.0"
 pytest-cov = "^4.1.0"
 pytest-dotenv = "^0.5.2"
-duckdb-engine = "^0.9.2"
+duckdb-engine = "^0.11.0"
```
ccurme, line 118: test fails with 0.10.0 as well

PR #21515, opened 2024-05-10
community: updated pubmed wrapper

Landstein:
Description:  Added the PMCID to the response of the pubmedapiwrapper.  The PMCID can be used to retrieve the full text of a paper of it is available on Pub Med Central.  

There are many other ways that the pubmed wrapper can be updated if this is something others would be interested in I can work on it further.  I can also update the documentation.


jigsawlabs-student:
Hi @eyurtsev, do you know who is the correct person to take a look at this?

Landstein:
@eyurtsev Is it possible to merge this in?

PR #21527, opened 2024-05-10
feat(community): support semantic hybrid score threshold in Azure AI Search

mspronesti:
Support semantic hybrid search with a score threshold -- similar to what we do for similarity search and for hybrid search (#20907).

### DIFF
--- a/libs/community/langchain_community/vectorstores/azuresearch.py
+++ b/libs/community/langchain_community/vectorstores/azuresearch.py
@@ -584,6 +584,29 @@ class AzureSearch(VectorStore):
         )
         return [(doc, score) for doc, score, _ in docs_and_scores]
 
+    def semantic_hybrid_search_with_rerank(
+        self, query: str, k: int = 4, **kwargs: Any
+    ) -> List[Tuple[Document, float]]:
+        """
+        Returns the most similar indexed documents to the query text.
+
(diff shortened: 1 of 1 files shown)


mspronesti:
@baskaryan @hwchase17

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -584,6 +584,29 @@ def semantic_hybrid_search_with_score(
         )
         return [(doc, score) for doc, score, _ in docs_and_scores]
 
+    def semantic_hybrid_search_with_rerank(
```
ccurme, line 587: what are your opinions on modifying `semantic_hybrid_search_with_score` to:

- accept `score_type: Literal["score", "reranker_score"] = "score"` to control which score is used in `docs_and_scores`;
- check for `score_threshold` as you do here.

ccurme (approved):

libs/community/langchain_community/vectorstores/azuresearch.py
mspronesti: Makes sense to me -- It would also make the vector store retriever more versatile. I've just appended one commit.

ccurme (approved):

PR #21529, opened 2024-05-10
text-splitters: Fix/recursive json splitter data persistence issue

bilkoh:

**Description:** Noticed an issue with when I was calling `RecursiveJsonSplitter().split_json()` multiple times that I was getting weird results. I found an issue where `chunks` list in the `_json_split` method. If chunks is not provided when _json_split (which is the case when split_json calls _json_split) then the same list is used for subsequent calls to `_json_split`.

You can see this in the test case i also added to this commit.

Output should be: 
```
[{'a': 1, 'b': 2}]
[{'c': 3, 'd': 4}]
```

Instead you get:
```
[{'a': 1, 'b': 2}]
[{'a': 1, 'b': 2, 'c': 3, 'd': 4}]
```


isahers1:
Bug is not reproducible by me, and I don't see bug in the code. @baskaryan

nfcampos (approved):
Thanks

PR #21534, opened 2024-05-10
fix(community): allow support for disabling max_tokens args

thehapyone:
This PR fixes an issue with not able to use unlimited/infinity tokens from the respective provider for the LiteLLM provider.

This is an issue when working in an agent environment that the token usage can drastically increase beyond the initial value set causing unexpected behavior.

### DIFF
--- a/libs/community/langchain_community/chat_models/litellm.py
+++ b/libs/community/langchain_community/chat_models/litellm.py
@@ -191,7 +191,7 @@ class ChatLiteLLM(BaseChatModel):
     n: int = 1
     """Number of chat completions to generate for each prompt. Note that the API may
        not return the full n completions if duplicates are generated."""
-    max_tokens: int = 256
+    max_tokens: Optional[int] = None
 
     max_retries: int = 6
 


thehapyone:
cc @hwchase17 @baskaryan @efriis

libs/community/langchain_community/chat_models/litellm.py
```diff
@@ -191,7 +191,7 @@ class ChatLiteLLM(BaseChatModel):
     n: int = 1
     """Number of chat completions to generate for each prompt. Note that the API may
        not return the full n completions if duplicates are generated."""
-    max_tokens: int = 256
+    max_tokens: Optional[int] = None
```
ccurme, line 194: Alternative is `Optional[int] = 256`.

Do you think anyone's workflow is assuming a default of 256 and will break as a result of this change?

This looks like a more reasonable default and appears consistent with litellm [as well](https://github.com/BerriAI/litellm/blob/80d8bf5d8f954b357d695fe77975764ffcf735e3/litellm/main.py#L233).

ccurme (approved):

PR #21536, opened 2024-05-10
community: Add alazy_load to AsyncHtmlLoader

cbornet:
Also fixes a bug that `_scrape` was called and was doing a second HTTP request synchronously.

**Twitter handle:** cbornet_

### DIFF
--- a/libs/community/langchain_community/document_loaders/async_html.py
+++ b/libs/community/langchain_community/document_loaders/async_html.py
@@ -1,8 +1,18 @@
 import asyncio
 import logging
 import warnings
-from concurrent.futures import ThreadPoolExecutor
-from typing import Any, Dict, Iterator, List, Optional, Union, cast
+from concurrent.futures import Future, ThreadPoolExecutor
+from typing import (
+    Any,
+    AsyncIterator,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+    cast,
+)
 
 import aiohttp
 import requests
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/document_loaders/async_html.py
```diff
@@ -186,21 +171,43 @@ async def fetch_all(self, urls: List[str]) -> Any:
...
+            parser = "xml"
+        else:
+            parser = self.default_parser
+        self._check_parser(parser)
+        soup = BeautifulSoup(text, parser)
+        metadata = _build_metadata(soup, url)
+        return Document(page_content=text, metadata=metadata)
 
+    def lazy_load(self) -> Iterator[Document]:
+        """Lazy load text from the url(s) in web_path."""
```
cbornet, line 210: I think it's problematic that load()/lazy_load() uses asyncio.
A thread-pool with a thread-safe synchronous http lib should be used instead.

eyurtsev: Yeah it's forcing users to use nest_asyncio which is not a good solution

eyurtsev (approved):

eyurtsev (approved):

Nazareka:
@cbornet After your changes, `_fetch_valid_connection_docs` is no longer used, is that a coincidence?

cbornet:
Indeed. It should probably be removed.

Nazareka:
Actually, I need functionality like that, so it's easier to skip invalid links:
```
except aiohttp.InvalidURL:
    if self.ignore_load_errors:
        logger.error(f"Invalid URL: {url}")
        return ""
```
instead of validating links before passing them to the loader. I will create a PR for this.

PR #21554, opened 2024-05-11
docs: Fix pixelation in stack graphic

gkermit:
This change updates the stack graphic displayed in the top-level README. The LangChain tile is pixelated in the current graphic.


gkermit:
@efriis

gkermit:
@hwchase17

gkermit:
@efriis

efriis:
Which browser are you using, and why does this fix it for that browser?

On macos chrome this doesn't change anything

efriis:
Doesn't look different in my version (17.4.1) of safari either :(

Would love to understand the change, and generally not opposed as long as it won't create issues in other browsers! Appreciate the contribution

efriis (approved):
oh fascinating - in safari it renders differently in github.com (the top-level readme you called out), but not when hosted in the python.langchain.com docs.

Will merge it in, and if you can spare a minute to explain how we can make this not happen in the future, that would be super helpful!

I believe these were just exported from figma.

gkermit:
Thanks for taking a look! I also found it confusing while verifying my change. The svg file was pixelated only when hosted on github. Rendering the original file in a browser by itself showed no issues. I fixed it by opening the file in adobe illustrator and reexporting it. No other changes.

efriis:
Well if it works it works 🤷 

Thanks @gkermit for the fix!

Issue #20469, opened 2024-04-15
OpenAPI agent does not support PATCH and PUT and DELETE, only GET and POST


yair-the-super-wise:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

agent = planner.create_openapi_agent(
    api_spec=api_spec,
    requests_wrapper=requests_wrapper,
    llm=llm,
    allow_dangerous_requests=True,
)

### Description

i want the tool to have full HTTP capabilities and not only GET and POST 

### System Info

latest langchain python installation which is 0.1.16

yair-the-super-wise:
it seems like can just add the requests to _create_api_controller_agent in planner.py at line 251 in langchain_community/agent_toolkits/openapi/planner.py

eyurtsev:
Feel free to enhance the functionality -- it should be opt-in if you do so

pow3rpi:
@yair-the-super-wise Please check this out [pull request](https://github.com/langchain-ai/langchain/pull/22962).
I don't know why this issue wasn't added to be automatically closed on merge despite the fact that I referenced it in the PR description.

PR #21576, opened 2024-05-12
community: support PATCH, PUT and DELETE for OpenAPI agent

ComposeC:
**Description:** add PATCH, PUT and DELETE tool to tool list

**Issue:** #20469


eyurtsev (changes requested):
If you'd like to add them to the API, could you expose them via a variable?

The reason is that:

1) This is a breaking change
2) delete, put, patch are mutations (post is a mutation too), but those are definitely more dangerous to be enabled by default

eyurtsev:
Closing this PR, feel free to re-open with something that exposes the methods via an argument

PR #21578, opened 2024-05-12
community: Allow passing a pg connection to PostgresChatMessageHistory

PolRF:
Enable passing a Postgresql connection to the PostgresChatMessageHistory class. To enhance production efficiency, it can be resource-intensive to open and close multiple Postgresql connections each time the class is instantiated. A more efficient approach would be to reuse the same connection without having to close it.


PolRF:
The improvement is already implemented in the new package: "This class is deprecated and will be removed in a future version. "
        "You can swap to using the `PostgresChatMessageHistory`"
        " implementation in `langchain_postgres`. "
        "Please do not submit further PRs to this class."
        "See https://github.com/langchain-ai/langchain-postgres"

PR #21586, opened 2024-05-13
[experimental][quick-fix]: Fix import errors and docs of experimental prompt injection identifier

nelsonauner:
**Description**

I noticed that code in experimental was not functional due to small errors, and users were complaining at https://stackoverflow.com/questions/78272521/hugging-face-prompt-injection-identifier and https://github.com/langchain-ai/langchain/discussions/19995.

I know folks are busy with `v2.0` now, but I believe this is worthy fix because it is isolated, minor, backwards compatible, and fixes an error with current v0.1 code

**Issue**
https://github.com/langchain-ai/langchain/discussions/19995

**Dependencies**
None

**Twitter handle**
https://twitter.com/nelsonauner

cc @asofter

### DIFF
--- a/libs/experimental/langchain_experimental/prompt_injection_identifier/hugging_face_identifier.py
+++ b/libs/experimental/langchain_experimental/prompt_injection_identifier/hugging_face_identifier.py
@@ -6,8 +6,8 @@ from typing import TYPE_CHECKING, Union
 from langchain.pydantic_v1 import Field, root_validator
 from langchain.tools.base import BaseTool
 
-if TYPE_CHECKING:
-    from transformers import Pipeline
+# This import needs to be toplevel to avoid error for refs on HuggingFaceInjectionIdentifier.model 
+from transformers import Pipeline
 
 
 class PromptInjectionException(ValueError):
@@ -98,3 +98,5 @@ class HuggingFaceInjectionIdentifier(BaseTool):
             raise PromptInjectionException("Prompt injection attack detected", score)
 
         return query
+
+HuggingFaceInjectionIdentifier.update_forward_refs()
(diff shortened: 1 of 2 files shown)


nelsonauner:
Thanks @baskaryan! I think should be OK to merge now, LMK if anything else

libs/experimental/langchain_experimental/prompt_injection_identifier/hugging_face_identifier.py
```diff
@@ -59,7 +59,7 @@ class HuggingFaceInjectionIdentifier(BaseTool):
         "Useful for when you need to ensure that prompt is free of injection attacks. "
         "Input should be any message from the user."
     )
-    model: Union[Pipeline, str, None] = Field(default_factory=_model_default_factory)
+    model: Any = Field(default_factory=_model_default_factory)
```
nelsonauner, line 62: The downside of using `Any` is that we actually _do_ want to force a `Pipeline`, `str`, or `None` - feeding in, say, an integer (`74`) would result in an error.

baskaryan: agreed that'd be ideal, but pydantic type validation causes issues with our conditional imports. if we really wanted we could add an explicit validator/root_validator` to check the model type

baskaryan: eg after line 86 we could check type

nelsonauner: Good points. That said are you OK merging this as-is? I'm not the author of this code or anything, I just found someone complaining that langchain was broken on stack overflow and figured I could do a quick patch

baskaryan: yep, and appreciate the fix!

PR #21598, opened 2024-05-13
community: add some modifications to bypass human verification from document_loaders on url_playwright.py

Kev744:
I add a context variable for mimicking human control on website to url_playwright.py code in order to bypass human verification websites (e.g. Cloudfare) when loading websites


eyurtsev:
Could you expose user agent as a parameter?

eyurtsev (changes requested):
could you expose as a parameter to the initializer

baskaryan:
is this same as #21911? if so can i close this one?

Kev744:
> is this same as #21911? if so can i close this one?

Yes almost the same because in my newer version I added an additional functionnal functionnality.

PR #21603, opened 2024-05-13
upstage : add merge_and_split function for document loader

junkeon0412:
- Introduce the `merge_and_split` function in the `UpstageLayoutAnalysisLoader`.
- The `merge_and_split` function takes a list of documents and a splitter as inputs.
- This function merges all documents and then divides them using the `split_documents` method, which is a proprietary function of the splitter.
- If the provided splitter is `None` (which is the default setting), the function will simply merge the documents without splitting them.

### DIFF
--- a/libs/partners/upstage/langchain_upstage/layout_analysis.py
+++ b/libs/partners/upstage/langchain_upstage/layout_analysis.py
@@ -1,7 +1,7 @@
 import os
 import warnings
 from pathlib import Path
-from typing import Iterator, List, Literal, Optional, Union
+from typing import Any, Dict, Iterator, List, Literal, Optional, Union
 
 from langchain_core.document_loaders import BaseLoader, Blob
 from langchain_core.documents import Document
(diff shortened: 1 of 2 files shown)


libs/partners/upstage/langchain_upstage/layout_analysis.py
```diff
@@ -204,3 +204,45 @@ def lazy_load(self) -> Iterator[Document]:
                 exclude=self.exclude,
             )
             yield from parser.lazy_parse(blob)
+
+    def merge_and_split(
```
eyurtsev, line 208: Would suggest to stick to the public API of Base Loader. This looks a bit superfluous  -- i.e., can use a stand alone function or else accept a splitting or merging logic as part of the initializer to the logic.

libs/partners/upstage/langchain_upstage/layout_analysis.py
```diff
@@ -204,3 +204,45 @@ def lazy_load(self) -> Iterator[Document]:
...
+        self, documents: List[Document], splitter: Optional[object] = None
```
eyurtsev, line 209: fyi if you want type hints you could add a dev dependency on text splitters and use the type

eyurtsev (approved):

PR #21606, opened 2024-05-13
docs: gh feedback link

baskaryan:


ccurme (approved):

docs/src/theme/Feedback.js
```diff
@@ -240,6 +240,7 @@ export default function Feedback() {
           </div>
         </>
       )}
+      <h4>You can leave detailed feedback <a href="https://github.com/langchain-ai/langchain/issues/new?assignees=&labels=03+-+Documentation&projects=&template=documentation.yml&title=DOC%3A+%3CPlease+write+a+comprehensive+title+after+the+%27DOC%3A+%27+prefix%3E">on GitHub</a>.</h4>
```
eyurtsev, line 243: do we have a way to add a URL to the given page?

baskaryan: im sure someone with greater js abilities than i does have a way, cc @bracesproul

bracesproul: Yes this is possible, can get on it today

baskaryan: i tried and i failed 🌝

bracesproul (approved):
you were so close @baskaryan ! Just forgot to use ` instead of " for template string

PR #21609, opened 2024-05-13
docs: fix path for state_of_the_union sample file

ccurme:


eyurtsev (approved):

eyurtsev:
I wonder whether we should provide some standard fixtures in a separate package `langchain_sample_data` or something like that

PR #21622, opened 2024-05-13
core[major]: only use function description

eyurtsev:
Do not prefix function signature

---

* Reason for this is that information is already present with tool calling models. 
* This will save on tokens for those models, and makes it more obvious what the description is!
* The @tool can get more parameters to allow a user to re-introduce the the signature if we want


hinthornw:
I like this move. OpenAI's max description length is actually fairly low.

hinthornw (approved):

PR #21637, opened 2024-05-13
community: Support gpt-4o in the Chat Model for OpenAI

Sumukh:
- **Description:** Supports other gpt-4 models here: 
  
  - **Issue:** Fixes OpenAI crashes 
  
  Calls to `llm.get_num_tokens_from_messages(buffer)`  would otherwise fail with 
  
  ```
  get_num_tokens_from_messages() is not presently implemented for model cl100k_base.See https://github.com/openai/openai-python/blob/main/chatml.md for information on how messages are converted to tokens.
```

  - **Dependencies:** N/A 
  - **Twitter handle:** @SumukhSridhara

### DIFF
--- a/libs/community/langchain_community/chat_models/openai.py
+++ b/libs/community/langchain_community/chat_models/openai.py
@@ -587,7 +587,7 @@ class ChatOpenAI(BaseChatModel):
                 # gpt-3.5-turbo may change over time.
                 # Returning num tokens assuming gpt-3.5-turbo-0301.
                 model = "gpt-3.5-turbo-0301"
-            elif model == "gpt-4":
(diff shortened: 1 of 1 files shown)


Haptein:
You don't need this ugly hack, just upgrade tiktoken.

derino:
This is indeed incorrect. The encoding model for gpt-4o is [different](https://github.com/openai/tiktoken/blob/main/tiktoken/model.py) than gpt-4. Instead, tiktoken should be bumped to 0.7.0.

Sumukh:
Yup, confirming that worked, thanks for taking a look.

PR #21638, opened 2024-05-13
core[minor]: Add v2 implementation of astream events

eyurtsev:
This PR introduces a v2 implementation of astream events that removes intermediate abstractions and fixes some issues with v1 implementation. 

The v2 implementation significantly reduces relevant code that's associated with the astream events implementation together with overhead.

After this PR, the astream events implementation:

- Uses an async callback handler
- No longer relies on BaseTracer
- No longer relies on json patch

As a result of this re-write, a number of issues were discovered with the existing implementation.

### on_chat_model_end `output`

The outputs associated with `on_chat_model_end` changed depending on whether it was within a chain or not.

As a root level runnable the output was: 

```python
"data": {"output": AIMessageChunk(content="hello world!", id='some id')}
```

As part of a chain the output was:

```
            "data": {
                "output": {
                    "generations": [
                        [
                            {
                                "generation_info": None,
                                "message": AIMessageChunk(
                                    content="hello world!", id=AnyStr()
                                ),
                                "text": "hello world!",
                                "type": "ChatGenerationChunk",
                            }
                        ]
                    ],
                    "llm_output": None,
                }
            },
```

After this PR, we will always use the simpler representation:

```python
"data": {"output": AIMessageChunk(content="hello world!", id='some id')}
```

**NOTE** Non chat models (i.e., regular LLMs) are still associated with the more verbose format.

### Remove some `_stream` events

`on_retriever_stream` and `on_tool_stream` events were removed -- these were not real events, but created as an artifact of implementing on top of astream_log. 

The same information is already available in the `x_on_end` events.

### Propagating Names

Names of runnables have been updated to be more consistent

```python
  model = GenericFakeChatModel(messages=infinite_cycle).configurable_fields(
        messages=ConfigurableField(
            id="messages",
            name="Messages",
            description="Messages return by the LLM",
        )
    )
```

Before:
```python
"name": "RunnableConfigurableFields",
```

After:
```python
"name": "GenericFakeChatModel",
```

### on_retriever_end

on_retriever_end will always return `output` which is a list of documents (rather than a dict containing a key called "documents")

### Retry events

Removed the `on_retry` callback handler. It was incorrectly showing that the failed function being retried has invoked `on_chain_end`

https://github.com/langchain-ai/langchain/pull/21638/files#diff-e512e3f84daf23029ebcceb11460f1c82056314653673e450a5831147d8cb84dL1394


libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -0,0 +1,793 @@
...
+            },
+            run_info["run_type"],
+        )
+
+    async def on_llm_end(
+        self, response: LLMResult, *, run_id: UUID, **kwargs: Any
+    ) -> None:
+        """End a trace for an LLM run."""
+        # "chat_model" is only used for the experimental new streaming_events format.
+        # This change should not affect any existing tracers.
```
nfcampos, line 263: this comment shouldn't be here?

libs/core/langchain_core/runnables/base.py
```diff
@@ -1936,7 +1800,8 @@ async def _atransform_stream_with_config(
         """Helper method to transform an Async Iterator of Input values into an Async
         Iterator of Output values, with callbacks.
         Use this to implement `astream()` or `atransform()` in Runnable subclasses."""
-        from langchain_core.tracers.log_stream import LogStreamCallbackHandler
+        # Mixing that is used by both astream log and astream events implementation
```
jacoblee93, line 1803: Typo

eyurtsev:
@jacoblee93  / @nfcampos 

`on_llm_end` still has a very verbose output. 

Options are:

1) Keep as is
2) output only the text (but makes it impossible to pluck out metadata)
3) output only the text, and start feeding metadata into a separate key in `data` 
4) do a hack and start outputting `AIMessageChunk`

libs/core/langchain_core/runnables/base.py
eyurtsev, line 1803:
```suggestion
        # Mixin that is used by both astream log and astream events implementation
```

jacoblee93:
> @jacoblee93 / @nfcampos
> 
> `on_llm_end` still has a very verbose output.
> 
> Options are:
> 
> 1. Keep as is
> 2. output only the text (but makes it impossible to pluck out metadata)
> 3. output only the text, and start feeding metadata into a separate key in `data`
> 4. do a hack and start outputting `AIMessageChunk`

I'd opt for making as few transformations as possible, which I think would be 1 or 4?

eyurtsev:
We can keep as (1) for now until we decide what to do. (4) Might be a bad idea since we don't use AIMessage in regular LLMs currently.

libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -0,0 +1,790 @@
...
+                event["data"]["input"] = input
+                first_event_run_id = event["run_id"]
+                yield event
+                continue
+
+            if event["run_id"] == first_event_run_id and event["event"].endswith(
+                "_end"
+            ):
+                # If it's the end event corresponding to the root runnable
+                # we want to include the input in the event since it's guaranteed
```
eyurtsev, line 779:
```suggestion
                # we dont include the input in the event since it's guaranteed
```

Issue #21425, opened 2024-05-08
Input variables is ignored even though passed in when MessagesPlaceholder.optional=True


LinboYuan:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```llm = init_chat_openai()

memory_key = "chat_context"

prompt = ChatPromptTemplate(
    messages=[
        MessagesPlaceholder(variable_name=memory_key, optional=True),
        HumanMessagePromptTemplate.from_template("{question}")
    ]
)

memory = ConversationBufferMemory(memory_key=memory_key, return_messages=True)
conversation = LLMChain(
    llm=llm,
    verbose=True,
    memory=memory,
    prompt=prompt
)


user_input = input("Me: ")
while user_input != "exit":
    print(conversation.invoke({"question": user_input}))
    user_input = input("Me: ")

### Error Message and Stack Trace (if applicable)

![1](https://github.com/langchain-ai/langchain/assets/12031633/419f2b04-a259-4cec-b66a-2d9ce2d5e112)
![2](https://github.com/langchain-ai/langchain/assets/12031633/462381c0-83dc-4ec2-9d79-5662c3c7d292)


### Description

as the title described, when MessagesPlaceholder.optional=True, the input variables will be ignored, even though I passed the arguments in. **I suppose this to be a bug, because it makes no sense when init a MessagesPlaceholder and optional set True.**

or **if this is by design, could you pls share some original design intentions or use cases.**

![3](https://github.com/langchain-ai/langchain/assets/12031633/893cf930-7cdb-40d9-8341-6fa043ce90ed)


### System Info

Name: langchain-core
Version: 0.1.45
Summary: Building applications with LLMs through composability
Home-page: https://github.com/langchain-ai/langchain
Author: 
Author-email: 
License: MIT
Location: /Users/linbo.yuan/Library/Python/3.9/lib/python/site-packages
Requires: PyYAML, tenacity, pydantic, langsmith, jsonpatch, packaging
Required-by: langchain, langchain-text-splitters, langchain-openai, langchain-community

keenborder786:
[...]

PR #21640, opened 2024-05-13
[Core]: MessagePlaceholder Optional Input Variable Fix

keenborder786:

  - **Description:**  The problem is that when MessagePlaceholder is set as optional, it is excluded from the input variables of the PromptsTemplate. This exclusion means that the messages in the MessagePlaceholder are not utilized, even if the user has explicitly provided them in the input variables. I see no rationale for this behavior in [this code](https://github.com/langchain-ai/langchain/blob/0541e06e21146419af226fa4fa35cfd6e601c267/libs/core/langchain_core/prompts/chat.py#L203). The input variables should be returned by MessagePlaceholder even when set as optional, because format_messages already accounts for this by returning an empty list when necessary.
  - **Issue:** #21425


keenborder786:
@baskaryan, @efriis, @eyurtsev, @hwchase17 please review

eyurtsev:
Hi @keenborder786 given that this is in core would you be able to add a unit test for the fix and inline into the PR summary what this PR is fixing?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -206,7 +206,7 @@ def input_variables(self) -> List[str]:
         Returns:
             List of input variable names.
         """
-        return [self.variable_name] if not self.optional else []
+        return [self.variable_name]
```
eyurtsev, line 209: Is `input_variables` supposed to represent required input variables or the names of all input variables?

keenborder786: @eyurtsev 
This should **always** represent the input variable associated with `MessagePlaceholder`.

Please see [this line](https://github.com/langchain-ai/langchain/blob/54e003268eada488ff007d9afb1974ef9c5535cd/libs/langchain/langchain/chains/llm.py#L196). This line shows that `LLMChain` creates the prompt using the `input_variables` attribute.

The bug is that `LLMChain` won't be able to use the messages if `MessagePlaceholder` is marked as optional. Regardless of whether messages are stored in the placeholder or not, the `input_variables` property will return an empty list. Therefore, the input won't be populated with the messages from `MessagePlaceholder`.

keenborder786:
> Hi @keenborder786 given that this is in core would you be able to add a unit test for the fix and inline into the PR summary what this PR is fixing?

@eyurtsev  no need of adding tests as this is a bug fix.

keenborder786:
@eyurtsev  please see

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev

eyurtsev:
@keenborder786 thanks i'll try to review next week -- we generally want unit tests everywhere in core. bug fixes are especially important -- i.e., it indicates that a logical issue was already made

keenborder786:
Closing this PR in favour of #22851

PR #21665, opened 2024-05-14
text-splitters: Handles cases where the separator is at the beginning of the string

ymh823680483:
**Description:** 
 When I use the `RecursiveCharacterTextSplitter` to split a string, if the separator is not at the beginning of the string, the  result will be abnormal, such as this example:
```
from langchain.text_splitter import RecursiveCharacterTextSplitter
SPLIT_TAGS = [",", "."]
splitter = RecursiveCharacterTextSplitter(chunk_size=10, chunk_overlap=0, separators=SPLIT_TAGS)
text = "Apple,banana,orange and tomato."
res = splitter.split_text(text)
print(res)
```
output: `['Apple', ',banana', ',orange and tomato', '.']`
The separator is attached to the beginning of the text block, which is strange when dividing the text.
My desired result is this:
`['Apple,', 'banana,', 'orange and tomato.']`
This PR solves this problem by modifying the method `_split_text_with_regex`

### DIFF
--- a/libs/text-splitters/langchain_text_splitters/character.py
+++ b/libs/text-splitters/langchain_text_splitters/character.py
@@ -36,10 +36,15 @@ def _split_text_with_regex(
         if keep_separator:
             # The parentheses in the pattern keep the delimiters in the result.
             _splits = re.split(f"({separator})", text)
-            splits = [_splits[i] + _splits[i + 1] for i in range(1, len(_splits), 2)]
-            if len(_splits) % 2 == 0:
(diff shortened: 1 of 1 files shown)


baskaryan (commented):
could we add some unit tests

ccurme:
Thanks for this. This is now supported via `keep_separator="end"`, although the type hint does not reflect this. I've updated the type hint and added your example as a test case here: https://github.com/langchain-ai/langchain/pull/25706.

PR #21669, opened 2024-05-14
core[patch]: Add unit test to catch ordering

eyurtsev:
Add unit test to catch ordering issues


eyurtsev:
unit tests for: https://github.com/langchain-ai/langchain/pull/21638

PR #21671, opened 2024-05-14
community: Add Zep Cloud components + docs + examples

paul-paliychuk:

- [x] **PR title**: community: Add Zep Cloud components + docs + examples

- [x] **PR message**: 
    We have recently released our new zep-cloud sdks that are compatible with Zep Cloud (not Zep Open Source). We have also maintained our Cloud version of langchain components (ChatMessageHistory, VectorStore) as part of our sdks. This PRs goal is to port these components to langchain community repo, and close the gap with the existing Zep Open Source components already present in community repo (added ZepCloudMemory,ZepCloudVectorStore,ZepCloudRetriever).
Also added a ZepCloudChatMessageHistory components together with an expression language example ported from our repo. We have left the original open source components intact on purpose as to not introduce any breaking changes.
    - **Issue:** -
    - **Dependencies:** Added optional dependency of our new cloud sdk `zep-cloud`
    - **Twitter handle:** @paulpaliychuk51

- [x] **Add tests and docs**

Additional guidelines:


paul-paliychuk:
Looking into linter errors

baskaryan (commented):
just double checking, there's no nice way to merge the local and cloud integrations is there? think it'd make for a slightly nicer devx but shouldn't force it if the local and cloud interfaces are totally different

paul-paliychuk:
@baskaryan Our Cloud APIs have diverged from Open Source, both from an API and functionality perspective. The API will likely be harmonized soon, but the functionality will stay differentiated. We have two sets of SDKs as a result.

We don't want to break the existing components for the open source users. 
Additionally, we want to avoid confusion for the cloud users about the sdk arguments, and just point them to the cloud components.

libs/community/langchain_community/retrievers/zep_cloud.py
```diff
@@ -0,0 +1,165 @@
...
+from langchain_core.documents import Document
+from langchain_core.pydantic_v1 import root_validator
+from langchain_core.retrievers import BaseRetriever
+
+try:
+    from zep_cloud import MemorySearchResult, SearchScope, SearchType
+    from zep_cloud.client import AsyncZep, Zep
+
+    class ZepCloudRetriever(BaseRetriever):
+        """`Zep Cloud` MemoryStore Retriever.
```
baskaryan, line 18: this type of definition will prevent this class from being properly documented in the [API reference](https://api.python.langchain.com/en/latest/community_api_reference.html). Would it be better to just remove the typing from the pydantic attributes (eg `zep_client: Any` instead of `zep_client: Zep`)?

paul-paliychuk: Roger that, addressed. Also did the same for async client.

baskaryan: sorry should've been clearer, we need to remove the class definition from the try/except clause for it to be documented

paul-paliychuk: @baskaryan my bad, removed the try/except clause.

baskaryan (approved):

PR #21680, opened 2024-05-14
community: Terminal tool

panasenco:
### Description
Allow AI agents to interact with a terminal window through tmux.

There's already a shell tool but it's limited to just executing commands and returning the output.
What if an AI agent could have full access to a terminal window and be able to see the contents of the window, interrupt long-running processes with Ctrl+C, and even use text editors like Vim?

It's worth noting that the current generation of LMs is completely incapable of utilizing full terminal access to do anything impressive. However at the rate progress is being made, it's safe to say that this limitation will be gone by the end of the year.

### Dependencies
Depends on the `tmux` system package being installed and the `libtmux` Python package. However, I think I succeeded in making the tests I added skip if `libtmux` is not installed, so I didn't have to modify pyproject.yml.


eyurtsev:
@panasenco thank you for the PR --  code looks clean and thanks for including tests.

We don't have a good opinion yet on where to place tools that are similar to this one (e.g., terminal tools or python interpreter tools). 

The problem is that we've been getting CVEs filed against the code base due to the more dangerous tools.

@baskaryan / @efriis should we consider cutting a separate package for tools that involve dealing with terminal or python access etc?

fehtemam:
I think excluding terminal and Python and similar tools will limit the capabilities of langchain. Sooner or later someone else will come along and include these tools in their project and will gain popularity. Organizations that are concerned about these generic and widespread tools becoming available in langchain are likely so risk-averse that they are not using langchain to begin with. Open source software comes with no warrantees and that's part of the deal. Just my two cents.

lucasrothman:
Hi any updates, this seems like a great tool!

Issue #21422, opened 2024-05-08
OllamaFunctions returning type Error when using with_structured_output


SatouKuzuma1:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

I'm using this simple code and its returning an error.
I cheked the documentation and the example is something similar to what I'm trying to do.
```python

from langchain_experimental.llms.ollama_functions import OllamaFunctions

class RelatedSubjects(BaseModel):
    topics: List[str] = Field(
        description="Comprehensive list of related subjects as background research.",
    )

ollama_functions_llm = OllamaFunctions(model="llama3",format='json')
expand_chain = gen_related_topics_prompt | ollama_functions_llm.with_structured_output(
    RelatedSubjects
)

related_subjects = await expand_chain.ainvoke({"topic": example_topic})
related_subjects
```

### Error Message and Stack Trace (if applicable)

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[537], [line 1](vscode-notebook-cell:?execution_count=537&line=1)
----> [1](vscode-notebook-cell:?execution_count=537&line=1) related_subjects = await expand_chain.ainvoke({"topic": example_topic})
      [2](vscode-notebook-cell:?execution_count=537&line=2) related_subjects

File [~/Desktop/python/lang-last/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py:2536](https://file+.vscode-resource.vscode-cdn.net/Users/chrollolucifer/Desktop/python/lang-last/~/Desktop/python/lang-last/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py:2536), in RunnableSequence.ainvoke(self, input, config, **kwargs)
   [2534](https://file+.vscode-resource.vscode-cdn.net/Users/chrollolucifer/Desktop/python/lang-last/~/Desktop/python/lang-last/.venv/lib/python3.11/site-packages/langchain_core/runnables/base.py:2534) try:
[...]

Issue #2222, opened 2023-03-31
TypeError: Object of type Document is not JSON serializable


sergerdn:
**Any comments would be appreciated.**

The issue is that the json module is unable to serialize the Document object, which is a custom class that inherits from BaseModel. The error message specifically says that the Document object is not JSON serializable, meaning it cannot be converted into a JSON string. This is likely because the json module does not know how to serialize the BaseModel class or any of its child classes. To fix the issue, we may need to provide a custom encoder or implement the jsonable_encoder function from the FastAPI library, which is designed to handle pydantic models like BaseModel.

```python
def query_chromadb():
    client_settings = chromadb.config.Settings(
        chroma_db_impl="duckdb+parquet",
        persist_directory=DB_DIR,
        anonymized_telemetry=False
    )

    embeddings = OpenAIEmbeddings()

    vectorstore = Chroma(
        collection_name="langchain_store",
        embedding_function=embeddings,
        client_settings=client_settings,
        persist_directory=DB_DIR,
    )

    result = vectorstore.similarity_search_with_score(query="FREDERICK", k=4)
    print(result)
    print(json.dumps(result, indent=4, sort_keys=False)) // ERROR


def main():
    # init_chromadb()
    query_chromadb()
```

```python
import json
from pydantic import BaseModel, Field

class Document(BaseModel):
    """Interface for interacting with a document."""

    page_content: str
    metadata: dict = Field(default_factory=dict)

doc = Document(page_content="Some page content", metadata={"author": "John Doe"})

print(json.dumps(doc)) // ERROR
```

Possible fixes:

```python
import json

from pydantic import BaseModel, Field


class Document(BaseModel):
    """Interface for interacting with a document."""

    page_content: str
    metadata: dict = Field(default_factory=dict)

    def to_dict(self):
        return self.dict(by_alias=True, exclude_unset=True) # just an example!

    def to_json(self):
        return self.json(by_alias=True, exclude_unset=True) # just an example!


doc = Document(page_content="Some page content", metadata={"author": "John Doe"})

# Convert to dictionary and serialize
doc_dict = doc.to_dict()
doc_json = json.dumps(doc.to_dict())

## {"page_content": "Some page content", "metadata": {"author": "John Doe"}}
print(doc_json)

# Or use the custom to_json() method
doc_json = doc.to_json()
## {"page_content": "Some page content", "metadata": {"author": "John Doe"}}
print(doc_json)
```

Another approach:
```python
import json
[...]

PR #21682, opened 2024-05-14
[Experimental]: Async agenerate method ollama functions

keenborder786:
- **Description:** :
          Added Async method for Generate for OllamaFunctions which was missing and was raising errors for the users.
   
- **Issue:** 
#21422


keenborder786:
@baskaryan, @efriis, @eyurtsev, @hwchase17 please review

libs/experimental/langchain_experimental/llms/ollama_functions.py
```diff
@@ -354,6 +357,86 @@ def _generate(
             generations=[ChatGeneration(message=response_message_with_functions)]
         )
 
+    async def _agenerate(
```
eyurtsev, line 360: Isn't this duplicating the implementation of the sync method? I thought BaseModel already takes care of this automatically under the hood?

eyurtsev (commented):
I can't tell if this code is doing anything that the default base implementation isn't doing already.

It looks like the original issue has to do with serialization and that something is not serializable. Have you verified that adding an async implementation makes a difference?

libs/experimental/langchain_experimental/llms/ollama_functions.py
```diff
@@ -354,6 +357,86 @@ def _generate(
...
+            functions.append(DEFAULT_RESPONSE_FUNCTION)
+        if _is_pydantic_class(functions[0]):
+            functions = [convert_to_ollama_tool(fn) for fn in functions]
+        system_message_prompt_template = SystemMessagePromptTemplate.from_template(
+            self.tool_system_prompt_template
+        )
+        system_message = system_message_prompt_template.format(
+            tools=json.dumps(functions, indent=2)
+        )
+        response_message = await super()._agenerate(
```
keenborder786, line 390: @eyurtsev this is the difference, I am calling the async generate method which solves the serialization issue.

keenborder786: Also I tested the async method on my machine and it solves the issue: #2222

keenborder786: You can try it on your machine as well!!!

keenborder786:
@eyurtsev please see

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev

falmanna:
what is the release schedule for this?

PR #21684, opened 2024-05-14
core[patch]: Update mypy config

eyurtsev:
Update mypy config to ignore checking deps from numpy (which is optional)

### DIFF
--- a/libs/core/pyproject.toml
+++ b/libs/core/pyproject.toml
@@ -81,7 +81,8 @@ disallow_untyped_defs = "True"
 exclude = ["notebooks", "examples", "example_data", "langchain_core/pydantic"]
 
 [[tool.mypy.overrides]]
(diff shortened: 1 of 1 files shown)


eyurtsev:
fix for: https://github.com/langchain-ai/langchain/actions/runs/9082098369/job/24957678369

baskaryan (approved):

libs/core/pyproject.toml
```diff
@@ -80,6 +80,10 @@ select = [
 disallow_untyped_defs = "True"
 exclude = ["notebooks", "examples", "example_data", "langchain_core/pydantic"]
 
+[[tool.mypy.overrides]]
+module = "numpy"
```
baskaryan, line 84: should we skip all of langsmith

PR #21687, opened 2024-05-14
docs: algolia api key update

efriis:


efriis:
not it - seems to be something with facets /contextual search

PR #21709, opened 2024-05-15
langchain.vectorstores.azuresearch: Allow AzureSerach Vector Store to update index on schema changes

thenewnano:
**Description:**
When changes are made to the index schema the vector store shall be updated. 
The way it is now it is just created on the first call and doesn't reflect changes later on, which mean manual update or delete has  to be done.


baskaryan (commented):
cc @lz-chen @lawadam (know you've updated this integration relatively recently)

PR #21715, opened 2024-05-15
How to: Streaming

eyurtsev:
Update the how to on streaming

### DIFF
--- a/docs/docs/how_to/streaming.ipynb
+++ b/docs/docs/how_to/streaming.ipynb
@@ -649,8 +649,7 @@
     "\n",
     ":::{.callout-note}\n",
     "\n",
-    "* `astream_events V1` was introduced in langchain-core **0.1.14**.\n",
-    "* `astream_events V2` was introduced in langchain-core **0.2.0**.\n",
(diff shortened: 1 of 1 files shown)


docs/docs/how_to/streaming.ipynb
```diff
@@ -599,27 +648,18 @@
     "Event Streaming is a **beta** API. This API may change a bit based on feedback.\n",
     "\n",
     ":::{.callout-note}\n",
-    "Introduced in langchain-core **0.1.14**.\n",
+    "\n",
+    "* `astream_events V1` was introduced in langchain-core **0.1.14**.\n",
+    "* `astream_events V2` was introduced in langchain-core **0.2.0**.\n",
```
ccurme, line 653: What do you think about something like "This guide demonstrates the V2 API and requires langchain-core >= 0.2. For the V1 API compatible with older versions of LangChain, see [here](https://python.langchain.com/v0.1/docs/expression_language/streaming/#using-stream-events)."

not sure if the APIs are sufficiently different to warrant this.

ccurme (approved):

PR #21717, opened 2024-05-15
docs: add deprecations for astream events v1

eyurtsev:
Add deprecations


docs/docs/versions/overview.mdx
```diff
@@ -7,6 +7,10 @@ sidebar_label: Overview
 
 ## What’s new in LangChain?
 
+The following features have been added during the development of 0.2.x:
+
+- Rolled out `v2` of the [Event Streaming API](https://python.langchain.com/docs/expression_language/streaming/#using-stream-events)
```
ccurme, line 12: I think if we're adding this section we should align it with the features we publicize in blogs. It looks strange if this is the only bullet point IMO.

docs/docs/versions/v_02/index.mdx
```diff
@@ -172,4 +172,9 @@ We have two main types of deprecations:
 
 2. Code that has better alternatives available and will eventually be removed, so there’s only a single way to do things. (e.g., `predict_messages` method in ChatModels has been deprecated in favor of `invoke`).
 
-    Many of these were marked for removal in 0.2. We have bumped the removal to 0.3.
\ No newline at end of file
+    Many of these were marked for removal in 0.2. We have bumped the removal to 0.3.
+
+### Notable Deprecations
```
ccurme, line 177: could also call out deprecating

- `create_structured_output_runnable` in favor of `with_structured_output` (this deprecation was recent; there are other extraction chains that were deprecated a month or two back, but this would draw attention to `with_structured_output` generally)
- `BaseTool.__call__` in favor of `invoke`
- LLMChain, RetrievalQA, and ConversationalRetrievalChain (docstrings explain replacements)
- `BaseRetriever.(a)get_relevant_documents` in favor of `(a)invoke`

these were all added in the last month or so and are targeted for removal in 0.3 in the decorator, although we can update that

might also be worth a reminder that `__call__`, `predict`, `predict_messages` and async variants remain deprecated for LLMs and chat models, with target for removal being extended to 0.3.

eyurtsev: added a list

eyurtsev:
Merged in another PR

PR #21730, opened 2024-05-15
upstage: Support batch input in embedding request.

JuHyung-Son:
**Description:** upstage embedding now supports batch input.

### DIFF
--- a/libs/partners/upstage/langchain_upstage/embeddings.py
+++ b/libs/partners/upstage/langchain_upstage/embeddings.py
@@ -51,7 +51,7 @@ class UpstageEmbeddings(BaseModel, Embeddings):
 
     client: Any = Field(default=None, exclude=True)  #: :meta private:
     async_client: Any = Field(default=None, exclude=True)  #: :meta private:
-    model: str
+    model: str = Field(default=None)
     """Embeddings model name to use. Do not add suffixes like `-query` and `-passage`.
     Instead, use 'solar-embedding-1-large' for example.
     """
--- a/libs/partners/upstage/langchain_upstage/embeddings.py
+++ b/libs/partners/upstage/langchain_upstage/embeddings.py
@@ -34,6 +34,7 @@ logger = logging.getLogger(__name__)
 DEFAULT_EMBED_BATCH_SIZE = 10
 MAX_EMBED_BATCH_SIZE = 100
 
+
 class UpstageEmbeddings(BaseModel, Embeddings):
     """UpstageEmbeddings embedding model.
 
@@ -203,7 +204,6 @@ class UpstageEmbeddings(BaseModel, Embeddings):
         params["model"] = params["model"] + "-passage"
         embeddings = []
 
-        
         batch_size = min(self.embed_batch_size, len(texts))
         for i in range(0, len(texts), batch_size):
             batch = texts[i : i + batch_size]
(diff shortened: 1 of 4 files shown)


libs/partners/upstage/langchain_upstage/embeddings.py
```diff
@@ -48,7 +51,7 @@ class UpstageEmbeddings(BaseModel, Embeddings):
 
     client: Any = Field(default=None, exclude=True)  #: :meta private:
     async_client: Any = Field(default=None, exclude=True)  #: :meta private:
-    model: str = "solar-1-mini-embedding"
+    model: str = "solar-embedding-1-large"
```
efriis, line 54: in general we recommend against setting this default and forcing people to define `UpstageEmbeddings(model="solar-embedding-1-large")` because this is technically a breaking change.

It might be worth shipping that breaking change now and not having to ship more breaking in the future
```suggestion
    model: str
```

libs/partners/upstage/langchain_upstage/embeddings.py
```diff
@@ -193,16 +197,19 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
         Returns:
             List of embeddings, one for each text.
         """
-        embeddings = []
+        assert (
+            len(texts) <= MAX_EMBED_BATCH_SIZE
+        ), f"The batch should not be larger than {MAX_EMBED_BATCH_SIZE}."
```
efriis, line 202: for most other integrations, we handle this batching (i.e. split it into multiple batches and return them all as if it were one request)

libs/partners/upstage/langchain_upstage/embeddings.py
JuHyung-Son: Thanks. Fixed it.

libs/partners/upstage/langchain_upstage/embeddings.py
JuHyung-Son: The original intent was to ensure that self.embed_batch_size does not exceed MAX_EMBED_BATCH_SIZE.

JuHyung-Son: Is it better to manage with Field?

libs/partners/upstage/langchain_upstage/embeddings.py
```diff
@@ -48,9 +51,9 @@ class UpstageEmbeddings(BaseModel, Embeddings):
 
     client: Any = Field(default=None, exclude=True)  #: :meta private:
     async_client: Any = Field(default=None, exclude=True)  #: :meta private:
-    model: str = "solar-1-mini-embedding"
+    model: str = Field(default=None)
```
efriis, line 54: I don't think you want this. Should we update the test to pass one of the available models instead?

efriis: With this implementation, you'll be able to create `UpstageEmbeddings()` but when you call it, it'll try to call the model `None-passage` which users probably don't want

libs/partners/upstage/langchain_upstage/embeddings.py
efriis: Ah makes sense - I misunderstood! This looks good

libs/partners/upstage/langchain_upstage/embeddings.py
```diff
@@ -232,16 +239,18 @@ async def aembed_documents(self, texts: List[str]) -> List[List[float]]:
         Returns:
             List of embeddings, one for each text.
         """
-        embeddings = []
+        assert (
+            self.embed_batch_size > MAX_EMBED_BATCH_SIZE
```
chosh0615, line 243:
```suggestion
            self.embed_batch_size <= MAX_EMBED_BATCH_SIZE
```

libs/partners/upstage/langchain_upstage/embeddings.py
JuHyung-Son: Yes right. I think this Field should be  `Field(...)`

efriis (approved):

PR #21736, opened 2024-05-16
openai[patch]: fix embedding float precision issue

efriis:
also clean up + comment some of the embedding batching code

### DIFF
--- a/libs/partners/openai/langchain_openai/embeddings/base.py
+++ b/libs/partners/openai/langchain_openai/embeddings/base.py
@@ -37,6 +37,68 @@ from langchain_core.utils import (
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -37,6 +37,67 @@
...
+        if len(_result) == 0:
+            # this will be populated with the embedding of an empty string
+            # in the sync or async code calling this
+            embeddings.append(None)
+            continue
+
+        elif len(_result) == 1:
+            # if only one embedding was produced, use it
+            embeddings.append(_result[0])
+            continue
```
efriis, line 77: @baskaryan this was the meat of solving the precision issue (just needed to add a len==1 case)

PR #21743, opened 2024-05-16
Add is_error

hinthornw:


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -109,6 +109,7 @@ def _merge_messages(
                             "type": "tool_result",
                             "content": curr.content,
                             "tool_use_id": curr.tool_call_id,
+                            "is_error": getattr(curr, "is_error", False),
```
hinthornw, line 112: I should probably just not add if it's none

PR #21762, opened 2024-05-16
community: enable SupabaseVectorStore to support extended table fields

xingwanying:

- [x] **PR title**: "community: enable SupabaseVectorStore to support extended table fields"

- [x] **PR message**: 
    - Added extension fields to the function  _add_vectors so that users can add other custom fields when insert a record into the database. eg:
    
   image


hwchase17 (approved):

xingwanying:
image
🤔

libs/community/langchain_community/vectorstores/supabase.py
```diff
@@ -333,10 +336,10 @@ def _add_vectors(
                 "content": documents[idx].page_content,
                 "embedding": embedding,
                 "metadata": documents[idx].metadata,  # type: ignore
+                **kwargs,
```
baskaryan, line 339: out of curiosity what's value of inserting same thing in every row? do you have an example of how this would be used?

xingwanying: I need to content from multiple GitHub repositories and embed it into a database. For version control, it is necessary to record the commit Id and commit sha. So, I'm thinking of adding some extra fields for this purpose. Here's how I plan to use it:
```
SupabaseVectorStore.from_documents(
            docs,
            embeddings,
            client=get_client(),
            table_name=TABLE_NAME,
            query_name=QUERY_NAME,
            chunk_size=CHUNK_SIZE,
            repo_name=repo_name,
            commit_id=commit_id
            commit_sha=commit_sha
        )
```
Result table：
image

PR #21763, opened 2024-05-16
docs: Docs (sample notebook) for Vertex Check Grounding Wrapper + Update Google Provider mdx docs

Abhishekbhagwat:
Adding 2 documentation updates:

1. [Vertex Check Grounding API ](https://cloud.google.com/generative-ai-app-builder/docs/check-grounding) Wrapper - The check grounding API determines how grounded a given piece of text is in a given set of reference texts. The API can generate supporting citations from the reference text to indicate where the given text is supported by the reference texts.
2. Update the Google Provider mdx docs with the Rerank and Check Grounding API


hwchase17 (commented):
i dont think `tools` is the right place for this - tools should be for things that subclass the `tool` class. @efriis where should we put integrations like this?

Abhishekbhagwat:
+1, I did not think that `tools` was the right place, but I saw some existing integrations (which are not tools) like [Upstage](https://python.langchain.com/v0.1/docs/integrations/tools/upstage_groundedness_check/) (which also does grounding check) and [Wolfram Alpha](https://python.langchain.com/v0.1/docs/integrations/tools/wolfram_alpha/) under tools, so I placed it there as well.

Do let me know what would be a more appropriate place and I'll send out a change for that! 🙏 
Thanks 

cc: @RajeshThallam @lkuligin

Abhishekbhagwat:
Hi @hwchase17 @efriis, just wanted to follow up on this to see if there is any resolution on where to place the docs ? Thank you!

efriis:
Could you

1. add a small example to the docstring in langchain-ai/langchain-google, so that it renders in the api docs: https://api.python.langchain.com/en/latest/vertex_check_grounding/langchain_google_community.vertex_check_grounding.VertexAICheckGroundingWrapper.html#langchain_google_community.vertex_check_grounding.VertexAICheckGroundingWrapper
2. reopen this pr with just the provider changes, linking to the api doc with the example code

Thanks!

PR #21770, opened 2024-05-16
langchain: default to Runnable in MultiQueryRetriever

ccurme:
- `llm_chain` becomes `Union[LLMChain, Runnable]`
- `.from_llm` creates a runnable

tested by verifying that docs/how_to/MultiQueryRetriever.ipynb runs unchanged with sync/async invoke (and that it runs if we specifically instantiate with LLMChain).


docs/docs/how_to/MultiQueryRetriever.ipynb
```diff
@@ -189,14 +180,16 @@
...
-   "id": "2eca2d96-8057-4ed9-873d-fa1064c09acf",
+   "id": "59c75c56-dbd7-4887-b9ba-0b5b21069f51",
    "metadata": {},
    "outputs": [
     {
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "INFO:langchain.retrievers.multi_query:Generated queries: ['1. Can you provide insights on regression from the course material?', '2. How is regression discussed in the course content?', '3. What information does the course offer about regression analysis?', '4. What are the teachings of the course regarding regression?', '5. In what manner is regression covered in the course curriculum?']\n"
+      "/Users/chestercurme/repos/langchain/libs/core/langchain_core/_api/deprecation.py:119: LangChainDeprecationWarning: The class `LLMChain` was deprecated in LangChain 0.1.17 and will be removed in 0.3.0. Use RunnableSequence, e.g., `prompt | llm` instead.\n",
```
ccurme, line 190: Would like input here: I think this is related to Pydantic and the union type. The warning is emitted if you instantiate directly with a Runnable. If I change the type hint to `Runnable` the warning is not emitted.

eyurtsev: I think you have to do this otherwise pydantic will coerce the runnable to an LLMChain (and the warning will be emitted)

libs/langchain/langchain/retrievers/multi_query.py
```diff
@@ -49,7 +50,7 @@ class MultiQueryRetriever(BaseRetriever):
     """
 
     retriever: BaseRetriever
-    llm_chain: LLMChain
+    llm_chain: Union[LLMChain, Runnable]
```
baskaryan, line 53: LLMChain is Runnable, should we just type as Runnable?

PR #21790, opened 2024-05-16
[ai21] configuration fix

leo-gan:
added "repository" and "Source Code" parameters (these parameters are missed only in this partner package configuration).


baskaryan (approved):

leo-gan:
@baskaryan Thanks, Bagatur!

PR #21792, opened 2024-05-16
docs: scripts: `package list` generation

leo-gan:
**Issue**: the current documentation does not support a list of all integration packages. If a provider creates a package outside of the `langchain-ai` GitHub organization, this package is "lost". There is no automatic way to announce this package. The provider can add an example of this integration to the `Integrations` part of the LC doc but an example is about the integration class, it is not about the integration package.

**PR**: Implemented the `Package Discovery` protocol. All LC packages (including the partner integration packages) should be included in the `packages.yaml` file. The `package dicsovery` code searches for the package information and collects this information in one table. `Package Discovery` protocol allows enumerate not only the integration packages but all "langchain_related" packages, which is helpful in many cases.

**Implementation**:
Code collects the package information and saves it in the table format.
Code updates the `integrations/platforms/index.mdx` file with the `packages` table.

**Note:** Right now, this script is started manually.
I didn't add the generation script to the `Makefile: generate-files` since the current documentation generation is in refactoring.
NOTE: All packages should be listed in the `packages.yaml` file. See the [`Discovery protocol`](https://github.com/langchain-ai/langchain/discussions/21768). Right now, only the packages are presented from the repositories in the `LangChain-ai` and from `googleapis` organizations. See [result here](https://langchain-2b3h0x1h5-langchain.vercel.app/v0.2/docs/integrations/platforms/).


efriis (commented):
I like it! Can't merge this until either
- some sort of discovery protocol is up to pull those in
- this is modified to hard code partner repos.

Out of curiosity, how it it currently discovering external repo partners like [cohere, elastic] and not others [google]?

leo-gan:
> Out of curiosity, how it it currently discovering external repo partners like [cohere, elastic] and not others [google]?
@efriis Those are discovered only locally. They will be gone when the script runs in the build or in the current `partners/` directory.

leo-gan:
> I like it! Can't merge this until either
> 
> * some sort of discovery protocol is up to pull those in
@efriis Let me know if you want me to draft the Discovery protocol. It will take a couple of days.

leo-gan:
@efriis Your recommendations have been addressed.

leo-gan:
@efriis Please let me know if you want other changes? I've added profiles for the packages:

- in the 'langchain` repo
- in the `langchain-ai` organization but outside of the `longchain` repo
- intrgration packages ouside of the `langchain-ai` organization (now I aware only about the Google packages. It could be something else)

I've added the correspondent instructions. They are also could be organized as a separate "Integration Package specification" doc.

packages.yaml
```diff
@@ -0,0 +1,168 @@
+version: v1
```
eyurtsev, line 1: Just popping in quickly to say that I love this!

efriis (commented):
Still sorry for the delay on this one. I'm planning on modifying to make the above suggested changes, as well as:

- make the generation script happen during the docs build
- split the generated package listing by type (base ones, shared repo ones, monorepo ones, then third party ones)
- try to standardize the paths that the packages point to (e.g. the googleapis ones point to a different relative parent than the langchain ones)

leo-gan:
> Still sorry for the delay on this one. I'm planning on modifying to make the above suggested changes, as well as:
> 
> * make the generation script happen during the docs build
> * split the generated package listing by type (base ones, shared repo ones, monorepo ones, then third party ones)
> * try to standardize the paths that the packages point to (e.g. the googleapis ones point to a different relative parent than the langchain ones)

Re: `split the generated package listing by type (base ones, shared repo ones, monorepo ones, then third party ones)` - it is also here. See the `slug` parameter in the `packages.yaml` It is not explicit but it is f(slug)

efriis:
I think e.g. all the google third party ones actually point to the wrong directory. All the langchain ones point to the package root (that has pyproject). All the google ones point to the src (code) root (that has root __init__.py)

leo-gan:
> I think e.g. all the google third party ones actually point to the wrong directory. All the langchain ones point to the package root (that has pyproject). All the google ones point to the src (code) root (that has root **init**.py)

I'm not sure I get you. We are talking about the `path` parameter in the `packages.yaml` file, right?

For example, the `langchain-google-firestore-python` package, `packages.yaml`: L156:160 The path: `path: src/langchain_google_firestore` is [OK](https://github.com/googleapis/langchain-google-firestore-python/tree/main/src/langchain_google_firestore). It is the source code that IS the python package, copied here `site-packages/langchain_google_firestore` in the python environment when the package is installed. 
Using `packages/path` allows us to parse the source directly from GitHub **without** installing the package. It is not required. It is used freely in the `METADATA` package file without any standard. 
It helps us to find (and parse) the package source code **without** installing the package. But we can just remove this `path` parameter and add a requirement to install the package to allow parsing its files (I'm not fun of it because many packages are incompatible in dependencies and we should deal with it, with the install-parse-uninstall cycle?). The `path` parameter is also helpful when we have multiple packages in one repo (all packages under `partner/`). LangChain repo has a standard on the source code placement but other vendors have different standards (like Google with `src/` prefix), so the `path` helps everybody with simple navigation in the repo code.

efriis:
closing because these are now populated in a slightly different way in FeatureTables

on the package front still open to a yaml file outlining them, but don't think it has value until it powers a feature

leo-gan:
@efriis 
> on the package front still open to a yaml file outlining them, but don't think it has value until it powers a feature
IMO, having a separate file with a list of packages is much cleaner than ingesting a package list inside code. It is better is support.  New packages go to this file and don't touch the code.

PR #21794, opened 2024-05-17
community:perplexity[patch]: standardize init args

jjesp123:
updated request_timeout default alias value per related docstring.

Related to [20085](https://github.com/langchain-ai/langchain/issues/20085)


eyurtsev (dismissed):
I think the original issue is about updating the names of the parameters rather than their values.

cc @baskaryan

eyurtsev:
@baskaryan could you update the original issue to clarify whether it's about updating default values or updating the names of the arguments?

ccurme:
Hi @jjesp123, thanks for noticing the inconsistency with the docstring. I've edited it instead of updating the default.

ccurme (approved):

PR #21843, opened 2024-05-17
LMStudioLLM class added. It may not be perfect but it is working. All…

AlonAshken:
…ows users to easily interact using LangChain with a hosted Local Inference Server in Meta's LMStudio. Feel free to corrent/add things to it. it is working well but may not be complete

- [x] **PR title**: "package: lmstudio. It is a package to allow interactions with LMStudio hosted LLMs."

- [x] **PR message**:
    - **Description:** Added an LLM to allow interacting with LMStudio hosted LLMs.
    - **Dependencies:** Only depends on BaseLLM
    - **Twitter handle:** Don't mind much about that but here's my handle: @AlonAshkenazi69
.


baskaryan (commented):
can we add a page to the docs at `docs/docs/integrations/llms/...`

AlonAshken:
> can we add a page to the docs at `docs/docs/integrations/llms/...`

Sure, I've added another piece for LMStudioEmbeddings and added docs for both of them.

PR #21846, opened 2024-05-17
docs: added template to `arxiv` page

leo-gan:
Updated `arXiv` page with the arxiv references from Templates (were references from Docs and API Refs, not Templates).
Re #21450 
CC @eyurtsev


leo-gan:
@baskaryan Thanks!

PR #21865, opened 2024-05-18
partners: Remove unnecessary print from voyageai embeddings

fzowl:

Remove unnecessary print from voyageai embeddings


libs/partners/voyageai/langchain_voyageai/embeddings.py
```diff
@@ -45,7 +45,6 @@ def default_values(cls, values: dict) -> dict:
         model = values.get("model")
         batch_size = values.get("batch_size")
         if batch_size is None:
-            print("batch size", batch_size)
```
ccurme, line 48: we import a logger in this module, should we emit the batch size with the logger?

ccurme (approved):

PR #21875, opened 2024-05-19
qdrant[patch]: Make path optional in from_existing_collection()

Anush008:
## Description

The `path` param is used to specify the local persistence directory, which isn't required if using Qdrant server.

~~This is a breaking but necessary change.~~

### DIFF
--- a/libs/partners/qdrant/langchain_qdrant/vectorstores.py
+++ b/libs/partners/qdrant/langchain_qdrant/vectorstores.py
@@ -1353,7 +1353,8 @@ class Qdrant(VectorStore):
     def from_existing_collection(
         cls: Type[Qdrant],
         embedding: Embeddings,
-        collection_name: str,
+        path: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


efriis (commented):
This is a breaking change for anyone currently using this function. Should we make all the params keyword instead of positional in case some others become optional in the future too? Alternatively could make both this and collection name optional.

Anush008:
This was always supposed to be optional.

This bug got carried over from the community implementation. Just that we caught it early in the partner package.

Anush008:
Hi @efriis. Just bumping this PR.

baskaryan (approved):

mariosajdini:
Hello!Do we have any update in this pr? I am currently using this function and i manualy removed path parameter in order for this to work.

Anush008:
@mariosajdini, we're waiting on a release.
In the meantime, you can pass `None` as the value to path. It'll work fine too.

PR #21885, opened 2024-05-19
community[minor]: Add UpstashRatelimitHandler

CahidArda:
Adding `UpstashRatelimitHandler` callback for rate limiting based on number of chain invocations or LLM token usage.

For more details, see [upstash/ratelimit-py repository](https://github.com/upstash/ratelimit-py) or the notebook guide included in this PR.

Twitter handle: @cahidarda


libs/community/langchain_community/callbacks/upstash_ratelimit_callback.py
```diff
@@ -0,0 +1,199 @@
...
+    every time you invoke.
+    """
+
+    raise_error = True
+    _checked = False
+
+    def __init__(
+        self,
+        identifier: str,
+        token_ratelimit: Optional[Ratelimit] = None,
```
eyurtsev, line 61: I generally recommend using `*` operator to force named arguments with defaults to be invoked as named arguments, it reduces breaking changes on future refactors

libs/community/langchain_community/callbacks/upstash_ratelimit_callback.py
```diff
@@ -0,0 +1,199 @@
...
+            request_ratelimit Optional[Ratelimit]: Ratelimit to limit the
+                number of requests
+            include_output_tokens bool: Whether to count output tokens when
+                rate limiting based on number of tokens. Only used when
+                `token_ratelimit` is passed. False by default.
+
+        Example:
+            .. code-block:: python
+
+            from upstash_redis import Redis
```
eyurtsev, line 87: missing indentation code block example

libs/community/langchain_community/callbacks/__init__.py
```diff
@@ -72,6 +72,10 @@
     from langchain_community.callbacks.trubrics_callback import (
         TrubricsCallbackHandler,
     )
+    from langchain_community.callbacks.upstash_ratelimit_callback import (
+        UpstashRatelimitError,  # noqa: F401
```
eyurtsev, line 76: Could you remove the F401 please to match the other callbacks?

eyurtsev (commented):
looks good overall, some minor comments only

libs/community/pyproject.toml
```diff
@@ -258,6 +258,7 @@ extended_testing = [
  "timescale-vector",
  "anthropic",
  "upstash-redis",
+ "upstash-ratelimit",
```
eyurtsev, line 261: Could you remove this import? I don't see it appearing in any extended tests

eyurtsev:
After thinking a bit more about this -- I am not sure that this is a good design for rate limiting. 

Why is it implemented via a callback handler? It would ideally just be a part in the chain that can wait until it can issue a request?

CahidArda:
> After thinking a bit more about this -- I am not sure that this is a good design for rate limiting.
> 
> Why is it implemented via a callback handler? It would ideally just be a part in the chain that can wait until it can issue a request?

I wanted to use callbacks because I felt like it would make adding request or [token based](https://langchain-6o2n61vth-langchain.vercel.app/v0.2/docs/integrations/callbacks/upstash_ratelimit/#ratelimiting-per-token) ratelimiting very easy.

I guess something like this would work for request based rate limiting:
```py
# request based
request_limiter = UpstashRatelimit("ip")
other_step = RunnableLambda(str)

chain = request_limiter | other_step
chain.invoke()
```

But I think token based would be more complex. We would need a step before LLM starts to stop the chain and another step after the LLM to count the tokens. Or somehow wrap the model step to do both but I don't know if this is possible in LangChain

```py
# token based
other_step = RunnableLambda(str)
model = ChatOpenAI()
model_with_ratelimit = UpstashRatelimit("ip", model=model)

chain = other_step | model_with_ratelimit
chain.invoke()
```

libs/community/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
eyurtsev, line 1: could you undo the changes in the lock file?

eyurtsev:
We generally don't want to assume that callbacks must be blocking for execution.

What use case is this callback handler helping to solve given that it's raising an exception?

Is the goal to apply different (lower) rate limits on a given deployment then the ones specified by the model provider?

CahidArda:
> We generally don't want to assume that callbacks must be blocking for execution.
> 
> What use case is this callback handler helping to solve given that it's raising an exception?
> 
> Is the goal to apply different (lower) rate limits on a given deployment then the ones specified by the model provider?

Yes, with the callback, it's becomes possible to allow n number of requests from an ip address or some user per minute/hour/day. It's also possible to rate limit based on the number of tokens.

eyurtsev:
I suspect a better design would be to create a chat model wrapper, potentially a bit more work for the user, but won't have any unexpected issues associated with the callback not being blocking

@CahidArda Anyway, let me know if you'd still like to merge

eyurtsev (approved):

CahidArda:
Hi @eyurtsev,

I think we can go ahead with callback if it's okay.

As for the lockfile, I have tried to remove it but when I remove it linter gets an error saying that the lock file is not compatible with the toml file. If I remove the changes in the toml file, tests get an error saying that upstash_ratelimit was not found. So I added upstash_ratelimit and bumped upstash_redis version while I am at it.

CahidArda:
Hi @eyurtsev,

Have you had the chance to review the changes?

CahidArda:
Hi again @eyurtsev,

[JavaScript version of this PR](https://github.com/langchain-ai/langchainjs/pull/5474) was merged recently. Have you had a chance to review the latest changes in this PR? 😄

eyurtsev:
@CahidArda apologies was on vacation until yesterday! merging

libs/community/tests/unit_tests/callbacks/test_upstash_ratelimit_callback.py
```diff
@@ -0,0 +1,239 @@
...
+
+from langchain_community.callbacks import UpstashRatelimitError, UpstashRatelimitHandler
+
+logger = logging.getLogger(__name__)
+
+try:
+    from upstash_ratelimit import Ratelimit, Response
+except ImportError as error:
+    Ratelimit, Response = None, None
+    logger.error(
```
eyurtsev, line 16: Could you remove side-effects on optional imports? (i.e., don't log an error here)

libs/community/langchain_community/callbacks/upstash_ratelimit_callback.py
```diff
@@ -0,0 +1,211 @@
...
+from typing import Any, Dict, List, Literal, Optional
+
+from langchain_core.callbacks import BaseCallbackHandler
+from langchain_core.outputs import LLMResult
+
+logger = logging.getLogger(__name__)
+try:
+    from upstash_ratelimit import Ratelimit
+except ImportError as error:
+    logger.error(
```
eyurtsev, line 13: Optional imports should have no side-effects for folks that are just importing the file

Side-effects only OK at run time if someone is using this callback (not at import time)

eyurtsev:
@CahidArda could you address the side-effects for the optional imports and we can merge then?

CahidArda:
Hope you had a great holiday! 🌴

I fixed the side effects.

eyurtsev:
Taking over to resolve merge conflicts

PR #21897, opened 2024-05-20
Fix ChatDatabricsk in case that streaming response doesn't have role field in delta chunk

WeichenXu123:

- [X] **PR title**: "package: description"

**Description:**
 Fix ChatDatabricsk in case that streaming response doesn't have role field in delta chunk

Additional guidelines:

### DIFF
--- a/libs/community/langchain_community/chat_models/mlflow.py
+++ b/libs/community/langchain_community/chat_models/mlflow.py
@@ -174,9 +174,14 @@ class ChatMlflow(BaseChatModel):
         )
         # TODO: check if `_client.predict_stream` is available.
         chunk_iter = self._client.predict_stream(endpoint=self.endpoint, inputs=data)
+        first_chunk_role = None
         for chunk in chunk_iter:
             choice = chunk["choices"][0]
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/chat_models/mlflow.py
```diff
@@ -174,9 +174,16 @@ def _stream(
...
         # TODO: check if `_client.predict_stream` is available.
         chunk_iter = self._client.predict_stream(endpoint=self.endpoint, inputs=data)
+        first_chunk_role = None
         for chunk in chunk_iter:
             choice = chunk["choices"][0]
-            chunk = ChatMlflow._convert_delta_to_message_chunk(choice["delta"])
+
+            chunk_delta = choice["delta"]
+            if first_chunk_role is None:
+                first_chunk_role = chunk_delta["role"]
```
WeichenXu123, line 183: `Mistral 7b on Databricks LLM` returns delta chunk which only the first chunk has "role" field.

efriis, line 183: should we do this in case it's not present?
```suggestion
                first_chunk_role = chunk_delta.get("role")
```

baskaryan (approved):

PR #21911, opened 2024-05-20
community : Update url_playwright from document_loaders to maximize efficiency retrieval

Kev744:
**Description :**
I added some modification to keep only some selectors in order to maximize execution time when loading documents and an another feature is to bypass human verification from some websites which has this ability (e.g. Cloudflare)

### DIFF
--- a/libs/community/langchain_community/document_loaders/url_playwright.py
+++ b/libs/community/langchain_community/document_loaders/url_playwright.py
@@ -60,9 +60,10 @@ class UnstructuredHtmlEvaluator(PlaywrightEvaluator):
     """Evaluate the page HTML content using the `unstructured` library."""
 
     def __init__(
-        self, 
-        remove_selectors: Optional[List[str]] = None, 
-        scraped_selectors: Optional[List[str]] = None
+        self,
+        remove_selectors: Optional[List[str]] = None,
+        *,
+        scraped_selectors: Optional[List[str]] = None,
     ):
         """Initialize UnstructuredHtmlEvaluator."""
         try:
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/document_loaders/url_playwright.py
```diff
@@ -59,7 +59,7 @@ async def evaluate_async(
 class UnstructuredHtmlEvaluator(PlaywrightEvaluator):
     """Evaluate the page HTML content using the `unstructured` library."""
 
-    def __init__(self, remove_selectors: Optional[List[str]] = None):
+    def __init__(self, remove_selectors: Optional[List[str]] = None, keep_only_selectors: Optional[List[str]] = None):
```
jacoblee93, line 62: This name is misleading - I was expecting it to be a boolean.

How about `scraped_selectors` or something similar?

libs/community/langchain_community/document_loaders/url_playwright.py
```diff
@@ -80,6 +81,15 @@ def evaluate(self, page: "Page", browser: "Browser", response: "Response") -> st
             for element in elements:
                 if element.is_visible():
                     element.evaluate("element => element.remove()")
+        
+        selectors = ','.join(self.keep_only_selectors) or 'body'
```
jacoblee93, line 85: Will this technically miss out on things like `<head>` or `<meta>`?

libs/community/langchain_community/document_loaders/url_playwright.py
```diff
@@ -135,8 +153,11 @@ def __init__(
         continue_on_failure: bool = True,
         headless: bool = True,
         remove_selectors: Optional[List[str]] = None,
+        keep_only_selectors: Optional[List[str]] = None,
         evaluator: Optional[PlaywrightEvaluator] = None,
         proxy: Optional[Dict[str, str]] = None,
+        header: Optional[str] = ('Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) '
```
jacoblee93, line 159: This is unrelated, correct?

Let's call this `user_agent`.

Kev744: Yes it's un related. It's done.

libs/community/langchain_community/document_loaders/url_playwright.py
Kev744: `<head>` and `<meta>` stays in the document because I only change body and not the head, so unstructured can partition these two tags even if he will not extract some informations from it (I think)

Kev744:
@baskaryan Now he must work :) Thanks

Kev744:
Can you update it, now it will work. Thanks you in advance @baskaryan @jacoblee93

Issue #21905, opened 2024-05-20
DOC: `langchain-community` will be not installed with `pip install langchain`


OpenVINO-dev-contest:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

`pip install langchain-community`

### Description

as my test on LangChain 0.2, `langchain-community` will be not installed with `pip install langchain`, which is conflicted with [document](https://python.langchain.com/v0.2/docs/how_to/installation/#langchain-community)

### System Info

Ubuntu

eyurtsev:
This is an intentional change. Please review the release notes: https://python.langchain.com/v0.2/docs/versions/overview/#what-changed

langchain should not dependency on langchain-community going forward.

eyurtsev:
fixing docs 
https://github.com/langchain-ai/langchain/pull/21920

PR #21917, opened 2024-05-20
[Langchain]: Added Back langchain-community as core dependency for langchain

keenborder786:
As pointed out in issue: #21905, `langchain-community` was no longer being installed with langchain while the official docs stated that it does get installed. Therefore I updated the poetry dependency for `langchain` so that `langchain-community` gets installed.


eyurtsev (changes requested):
Please review the changes here:

https://python.langchain.com/v0.2/docs/versions/overview/#what-changed

PR #21932, opened 2024-05-20
rfc: use langsmith style docs

baskaryan:


jacoblee93:
I think fine in a vacuum, some concern about confusion around a drastic visual change this soon after a big announcement

PR #21935, opened 2024-05-20
infra: fix CI on text-splitters

ccurme:


.github/workflows/_lint.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         # It doesn't matter how you change it, any change will cause a cache-bust.
         working-directory: ${{ inputs.working-directory }}
         run: |
-          poetry install --with lint,typing
```
baskaryan, line 72: ooc why is this needed? we want to lint without the testing dependencies since they might mask an invalid import

PR #21938, opened 2024-05-20
langchain: In `LocalFileStore`, enhance `update_atime` to support files owned by other users

chrispy-snps:
**Description:**

The `LocalFileStore` file-based cache supports an `update_atime=True` parameter that specifies to update the access time when a cache entry is read. This allows a *least-recently-used* (LRU) cache policy to be applied to the `LocalFileStore` directory to delete cache entries that have not been referenced for some time.

However, the `os.chmod()` operation used by `update_atime` fails if the cache file was written by a different user. (Files owned by one user cannot have permissions or access times changed by another user.)

To handle this case, the pull request implements a copy-then-modify fallback for files written by other users. This is slightly slower than a metadata-only change, but it is still plenty fast (~200 files per second on an NFS filesystem) thanks to small files and modern filesystem caching.

**Testing:**

I could not think of a way to test multi-user situations in the unit tests. However, I am using this code in our environment so it is tested here.

@eyurtsev has reviewed my previous contributions in this area.

### DIFF
--- a/libs/langchain/langchain/storage/file_system.py
+++ b/libs/langchain/langchain/storage/file_system.py
@@ -1,5 +1,6 @@
 import os
 import re
+import shutil
 import time
 from pathlib import Path
 from typing import Iterator, List, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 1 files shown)


baskaryan (commented):
could we add some unit tests?

chrispy-snps:
@baskaryan - I wasn't sure how to create a unit test that could reproduce the specific permissions situation where a file is owned by another user. Do you have any thoughts on this?

chrispy-snps:
@baskaryan - do you have any thoughts on how a unit test could be constructed for a feature that relies on file ownership behaviors?

ccurme:
Closing but will re-open if anyone wants to continue iterating on this.

PR #21975, opened 2024-05-21
community: Add async methods to CassandraChatMessageHistory

cbornet:


libs/community/langchain_community/chat_message_histories/cassandra.py
```diff
@@ -19,27 +22,37 @@
...
+class CassandraChatMessageHistory(BaseChatMessageHistory):
     def __init__(
         self,
         session_id: str,
         session: Optional[Session] = None,
         keyspace: Optional[str] = None,
         table_name: str = DEFAULT_TABLE_NAME,
         ttl_seconds: Optional[int] = DEFAULT_TTL_SECONDS,
+        *,
+        setup_mode: Optional[SetupMode] = SetupMode.SYNC,
```
eyurtsev, line 41: What's your perspective on `setup_mode` vs. having a dedicated `async_session` ?

cbornet: The cassandra `Session` is used for both sync and async so there's no "async session"

PR #21976, opened 2024-05-21
community[patch]: Improve tips in abnormal request for BaichuanTextEmbeddings

maang-h:
- **Description:** 
   - When using 'BaichuanEmbedding', the content of the prompt is almost the same regardless of the problem. For example, for 'api_key error' and 'Incorrect proxy settings', there are only differences in the status code, which is not very user-friendly. We should further explain the common status codes! 
 
    - Additionally, it turns out that for ` status_code= 200 ` all directly print statements and return None, but here I think we can directly more clearly prompt the user to discover the problem! (there is no conflict with the previous one)

### DIFF
--- a/libs/community/langchain_community/embeddings/baichuan.py
+++ b/libs/community/langchain_community/embeddings/baichuan.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Dict, List, Optional
 
 import requests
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/embeddings/baichuan.py
```diff
@@ -77,17 +78,42 @@ def _embed(self, texts: List[str]) -> Optional[List[List[float]]]:
                 sorted_embeddings = sorted(embeddings, key=lambda e: e.get("index", 0))
                 # Return just the embeddings
                 return [result.get("embedding", []) for result in sorted_embeddings]
+            # Common Error 1: Invalid `api_key`
```
eyurtsev, line 81: why not using: 

```python
response.raise_for_status()
```

The warnings.warn seem like a pretty unusual pattern. You can improve the error message in the raised exception?

libs/community/langchain_community/embeddings/baichuan.py
```diff
@@ -77,17 +78,42 @@ def _embed(self, texts: List[str]) -> Optional[List[List[float]]]:
...
-                    "embedding API"
+                    "`BaichuanEmbedding` API"
                 )
-                return None
+            return None
         except Exception as e:
             # Log the exception or handle it as needed
-            print(f"Exception occurred while trying to get embeddings: {str(e)}")  # noqa: T201
-            return None
+            raise Exception(  # noqa: T201
```
eyurtsev, line 114: Should not be raising an `Exception` type, either re-raise the original exception or else make a dedicated exception type?

eyurtsev (commented):
Exception handling is unusual in this code. It's better to raise the exception in general to avoid losing information that might be relevant for retries and logging

libs/community/langchain_community/embeddings/baichuan.py
maang-h: get, i will modify it.

libs/community/langchain_community/embeddings/baichuan.py
maang-h: get

maang-h:
@eyurtsev hi, i modified the code as you suggested.

PR #21977, opened 2024-05-21
core[patch]: In astream_events(version=v2) tap output of root run

nfcampos:
- if tap_output_iter/aiter is called multiple times for the same run issue events only once
- if chat model run is tapped don't issue duplicate on_llm_new_token events
- if first chunk arrives after run has ended do not emit it as a stream event

### DIFF
--- a/libs/core/langchain_core/runnables/config.py
+++ b/libs/core/langchain_core/runnables/config.py
@@ -124,7 +124,6 @@ def ensure_config(config: Optional[RunnableConfig] = None) -> RunnableConfig:
         metadata={},
         callbacks=None,
         recursion_limit=25,
-        run_id=None,
     )
     if var_config := var_child_runnable_config.get():
         empty.update(
--- a/libs/core/langchain_core/tracers/event_stream.py
+++ b/libs/core/langchain_core/tracers/event_stream.py
@@ -129,19 +129,20 @@ class _AstreamEventsCallbackHandler(AsyncCallbackHandler, _StreamingCallbackHand
         run_info = self.run_map.get(run_id)
         if run_info is None:
             # run has finished, don't issue any stream events
-            yield first
+            yield cast(T, first)
             return
         if tap is sentinel:
             # if we are the first to tap, issue stream events
-            event = {
+            event: StreamEvent = {
                 "event": f"on_{run_info['run_type']}_stream",
                 "run_id": str(run_id),
                 "name": run_info["name"],
                 "tags": run_info["tags"],
                 "metadata": run_info["metadata"],
+                "data": {},
             }
             self._send({**event, "data": {"chunk": first}}, run_info["run_type"])
-            yield first
+            yield cast(T, first)
             # consume the rest of the output
             async for chunk in output:
                 self._send(
@@ -151,7 +152,7 @@ class _AstreamEventsCallbackHandler(AsyncCallbackHandler, _StreamingCallbackHand
                 yield chunk
         else:
             # otherwise just pass through
-            yield first
+            yield cast(T, first)
             # consume the rest of the output
             async for chunk in output:
                 yield chunk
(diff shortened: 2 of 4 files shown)


libs/core/langchain_core/runnables/config.py
```diff
@@ -124,7 +124,6 @@ def ensure_config(config: Optional[RunnableConfig] = None) -> RunnableConfig:
         metadata={},
         callbacks=None,
         recursion_limit=25,
-        run_id=None,
```
nfcampos, line 127: this ensures we never have None run_id after calling ensure_config

libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -116,41 +118,83 @@ async def tap_output_aiter(
...
-            )
-            yield chunk
+        sentinel = object()
+        # atomic check and set
+        tap = self.is_tapped.setdefault(run_id, sentinel)
+        # wait for first chunk
+        first = await py_anext(output, sentinel)
+        if first is sentinel:
+            return
+        # get run info
```
nfcampos, line 128: run info may not be available before first chunk is emitted

libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -116,41 +118,83 @@ async def tap_output_aiter(
...
+        run_info = self.run_map.get(run_id)
+        if run_info is None:
+            # run has finished, don't issue any stream events
+            yield first
```
nfcampos, line 132: no need to raise exception here, if something not working we already raise exception in the end events

libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -244,6 +288,8 @@ async def on_llm_new_token(
 
         if run_info is None:
             raise AssertionError(f"Run ID {run_id} not found in run map.")
+        if self.is_tapped.get(run_id):
```
nfcampos, line 291: avoid duplicate stream events when chat model is tapped (which could already happen before this pr)

libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -748,6 +794,7 @@ async def _astream_events_implementation_v2(
 
     # Assign the stream handler to the config
     config = ensure_config(config)
+    run_id = config.setdefault("run_id", uuid4())
```
nfcampos, line 797: this ensures we have a known id for this run (instead of assigning it inside start callback

eyurtsev:
Taking over to add unit-tests!

Thanks @nfcampos

libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -122,7 +122,7 @@ async def tap_output_aiter(
         # atomic check and set
         tap = self.is_tapped.setdefault(run_id, sentinel)
         # wait for first chunk
-        first = await py_anext(output, sentinel)
+        first = await py_anext(output, default=sentinel)
```
nfcampos, line 125: hmm why?

eyurtsev: Convention that i always follow with python, if defined as named argument, then invoke as a named argument, if defined as positional argument then invoke as a positional argument. This is from before `*` days.

```
def py_anext(
    iterator: AsyncIterator[T], default: Union[T, Any] = _no_default
) -> Awaitable[Union[T, None, Any]]:
```

libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -116,41 +118,85 @@ async def tap_output_aiter(
...
+        sentinel = object()
+        # atomic check and set
+        tap = self.is_tapped.setdefault(run_id, sentinel)
+        # wait for first chunk
+        first = await py_anext(output, sentinel)
+        if first is sentinel:
+            return
+        # get run info
+        run_info = self.run_map.get(run_id)
+        if run_info is None:
```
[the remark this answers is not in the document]
nfcampos: if on_run_end callback was called before this one, as we pop the run out of the map there

PR #21980, opened 2024-05-21
community:javelin[patch]: standardize init args, update for javelin sdk release.

MSubik:
Related to [20085](https://github.com/langchain-ai/langchain/issues/20085) Updated the Javelin chat model to standardize the initialization argument.  Also fixed an existing bug, where code was initialized with incorrect call to the JavelinClient defined in the javelin_sdk, resulting in an initialization error.  See related [Javelin Documentation](https://docs.getjavelin.io/docs/javelin-python/quickstart).

### DIFF
--- a/libs/community/tests/unit_tests/chat_models/test_javelin_ai_gateway.py
+++ b/libs/community/tests/unit_tests/chat_models/test_javelin_ai_gateway.py
@@ -4,6 +4,7 @@ import pytest
 from langchain_core.pydantic_v1 import SecretStr
 
 from langchain_community.chat_models import ChatJavelinAIGateway
+from typing import cast
 
 
 @pytest.mark.requires("javelin_sdk")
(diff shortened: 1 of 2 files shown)


libs/community/tests/unit_tests/chat_models/test_javelin_ai_gateway.py
```diff
@@ -30,3 +31,17 @@ def test_api_key_masked_when_passed_via_constructor() -> None:
...
     assert "secret-api-key" not in repr(llm)
+
+
+@pytest.mark.requires("javelin_sdk")
+def test_api_key_alias() -> None:
+    for model in [
+        ChatJavelinAIGateway(
+            route="<javelin-ai-gateway-chat-route>",
+            javelin_api_key=SecretStr("test_key"),
+        ),
```
baskaryan, line 42: you shouldn't need to pass in the api key as SecretStr, can just pass in str

baskaryan (approved):

PR #21992, opened 2024-05-21
core[minor]: Add InMemoryRateLimiter

eyurtsev:
This PR introduces the following Runnables:

1. BaseRateLimiter: an abstraction for specifying a time based rate limiter as a Runnable
2. InMemoryRateLimiter: Provides an in-memory implementation of a rate limiter

## Example

```python

from langchain_core.runnables import InMemoryRateLimiter, RunnableLambda
from datetime import datetime

foo = InMemoryRateLimiter(requests_per_second=0.5)

def meow(x):
    print(datetime.now().strftime("%H:%M:%S.%f"))
    return x

chain = foo | meow

for _ in range(10):
    print(chain.invoke('hello'))
```

Produces:

```
17:12:07.530151
hello
17:12:09.537932
hello
17:12:11.548375
hello
17:12:13.558383
hello
17:12:15.568348
hello
17:12:17.578171
hello
17:12:19.587508
hello
17:12:21.597877
hello
17:12:23.607707
hello
17:12:25.617978
hello
```

image

## Interface

The rate limiter uses the following interface for acquiring a token:

```python
class BaseRateLimiter(Runnable[Input, Output], abc.ABC):
  @abc.abstractmethod
  def acquire(self, *, blocking: bool = True) -> bool:
      """Attempt to acquire the necessary tokens for the rate limiter.```
```

The flag `blocking` has been added to the abstraction to allow supporting streaming (which is easier if blocking=False).

## Limitations

- The rate limiter is not designed to work across different processes. It is an in-memory rate limiter, but it is thread safe.
- The rate limiter only supports time-based rate limiting. It does not take into account the size of the request or any other factors.
- The current implementation does not handle streaming inputs well and will consume all inputs even if the rate limit has been reached. Better support for streaming inputs will be added in the future.
- When the rate limiter is combined with another runnable via a RunnableSequence, usage of .batch() or .abatch() will only respect the average rate limit.  There will be bursty behavior as .batch() and .abatch() wait for each step to complete before starting the next step. One way to mitigate this is to use batch_as_completed() or abatch_as_completed().

## Bursty behavior in `batch` and `abatch`

When the rate limiter is combined with another runnable via a RunnableSequence, usage of .batch() or .abatch() will only respect the average rate limit.  There will be bursty behavior as .batch() and .abatch() wait for each step to complete before starting the next step.

This becomes a problem if users are using `batch` and `abatch` with many inputs (e.g., 100). In this case, there will be a burst of 100 inputs into the batch of the rate limited runnable.

1. Using a RunnableBinding

The API would look like:

```python
from langchain_core.runnables import InMemoryRateLimiter, RunnableLambda

rate_limiter = InMemoryRateLimiter(requests_per_second=0.5)

def meow(x):
    return x

rate_limited_meow = RunnableLambda(meow).with_rate_limiter(rate_limiter)
```

2. Another option is to add some init option to RunnableSequence that changes `.batch()`  to be depth first (e.g., by delegating to `batch_as_completed`)

```python
RunnableSequence(first=rate_limiter, last=model, how='batch-depth-first')
```

Pros: Does not require Runnable Binding
Cons: Feels over-complicated


hinthornw:
> model.with_tools([]).with_rate_limit works, but model.with_rate_limit.with_tools doesn't work (order sensitive)


Did we ever settle on LLM / proxy objects for binding? Feels like we should solve this generically

eyurtsev:
image

> Did we ever settle on LLM / proxy objects for binding? Feels like we should solve this generically

No, still unsolved, and a frequent problem

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,210 @@
...
+        """Consume the given amount of tokens if possible.
+
+        Returns:
+            True means that the tokens were consumed, and the caller can proceed to
+            make the request. A False means that the tokens were not consumed, and
+            the caller should try again later.
+        """
+
+    @abc.abstractmethod
+    async def _aconsume(self) -> bool:
```
hinthornw, line 57: does this need to be in the base class? or is this specific to token counting rate limiting methods?

hinthornw:
seems reasonable to me... should we implement 2 or 3 rate limiters to get more confident in the base abstraction?

libs/core/langchain_core/runnables/base.py
```diff
@@ -1337,6 +1337,17 @@ def with_types(
             kwargs={},
         )
 
+    def with_rate_limit(self, *, rate_limiter) -> Runnable[Input, Output]:
+        """Create a new Runnable that retries the original runnable on exceptions."""
```
baskaryan, line 1341: docstring

libs/core/langchain_core/runnables/rate_limiter.py
eyurtsev: Base class useful if we want to support rate limit between different processes, could be overkill since it'll probably cause serialization complications

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def _consume(self) -> bool:
+        """Consume the given amount of tokens if possible.
+
+        Returns:
+            True means that the tokens were consumed, and the caller can proceed to
+            make the request. A False means that the tokens were not consumed, and
+            the caller should try again later.
+        """
+        with self._consume_lock:
```
nfcampos, line 125: why not use a semaphore?

eyurtsev: do you think it will help here? i think of a semaphore as helping to control for concurrency rather than for limiting across time

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,217 @@
...
+                return True
+
+            return False
+
+    def sync_wait(self) -> None:
+        """Blocking call to wait until the given number of tokens are available."""
+        while not self._consume():
+            time.sleep(self.check_every_n_seconds)
+
+    async def _aconsume(self) -> bool:
```
baskaryan, line 204: ooc why's this needed

eyurtsev: It's not -- was convenient for testing -- it's a non blocking version of consume

eyurtsev: We could change the abstraction to match this: 

https://docs.python.org/3/library/threading.html#threading.Lock.acquire

A non-blocking implementation could perhaps allow users to switch to another LLM based on the rate limits

eyurtsev: I changed to match acquire semantics so we can differentiate between blocking vs. non-blocking (useful to support streaming)

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,326 @@
...
+    using a blocking call.
+
+    Current limitations:
+
+    - The rate limiter is not designed to work across different processes. It is
+      an in-memory rate limiter, but it is thread safe.
+    - The rate limiter only supports time-based rate limiting. It does not take
+      into account the size of the request or any other factors.
+    - The current implementation does not handle streaming inputs well and will
+      consume all inputs even if the rate limit has been reached. Better support
```
eyurtsev, line 48:
```suggestion
      consume all inputs even if the rate limit has not been reached. Better support
```

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,326 @@
...
+    a way to keep track of how many requests can be made at a given time.
+
+    Current limitations:
+
+    - The rate limiter is not designed to work across different processes. It is
+      an in-memory rate limiter, but it is thread safe.
+    - The rate limiter only supports time-based rate limiting. It does not take
+      into account the size of the request or any other factors.
+    - The current implementation does not handle streaming inputs well and will
+      consume all inputs even if the rate limit has been reached. Better support
```
eyurtsev, line 175:
```suggestion
      consume all inputs even if the rate limit has not been reached. Better support
```

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,326 @@
...
+
+        async def _ainvoke(input: Input) -> Output:
+            """Invoke the rate limiter. Internal function."""
+            await self.aacquire(blocking=True)
+            return cast(Output, input)
+
+        return await self._acall_with_config(_ainvoke, input, config, **kwargs)
+
+
+@beta(message="Introduced in 0.2.24. API subject to change.")
```
baskaryan, line 147: should the base class be what's marked as beta? if the api is subject to change

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,326 @@
...
+
+        Args:
+            blocking: If True, the method will block until the tokens are available.
+                If False, the method will return immediately with the result of
+                the attempt. Defaults to True.
+
+        Returns:
+           True if the tokens were successfully acquired, False otherwise.
+
+        .. versionadded:: 0.2.24
```
baskaryan, line 98: feels unnecessary given the whole class was added in 0.2.24?

libs/core/langchain_core/runnables/rate_limiter.py
```diff
@@ -0,0 +1,326 @@
...
+
+            # As long as we have at least one token, we can proceed.
+            if self.available_tokens >= 1:
+                self.available_tokens -= 1
+                return True
+
+            return False
+
+    def acquire(self, *, blocking: bool = True) -> bool:
+        """Attempt to acquire a token from the rate limiter.
```
baskaryan, line 277: nit dont strictly need to copy docstring here

baskaryan (approved):

PR #21998, opened 2024-05-21
langchain[patch]: Add unit test to detect changes to community imports

eyurtsev:
Add unit tests for community imports

### DIFF
--- a/libs/langchain/tests/unit_tests/test_imports.py
+++ b/libs/langchain/tests/unit_tests/test_imports.py
@@ -124,8 +124,8 @@ def extract_deprecated_lookup(file_path: str) -> Optional[Dict[str, Any]]:
     return None
 
 
-def _dict_from_ast(node) -> Dict[str, str]:
-    """Convers an AST dict node to a Python dictionary, assuming str to str format.
+def _dict_from_ast(node: ast.AST) -> Dict[str, str]:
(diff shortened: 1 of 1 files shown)


libs/langchain/tests/unit_tests/test_imports.py
```diff
@@ -60,3 +62,103 @@ def test_import_all_using_dir() -> None:
...
+
+        deprecated_lookup = extract_deprecated_lookup(str(path))
+        if deprecated_lookup is None:
+            continue
+
+        # This uses a very simple hash, so it's not foolproof, but it should catch
+        # most cases.
+        hash_ += len(str(sorted(deprecated_lookup.items())))
+
+    evil_magic_number = 38620
```
baskaryan, line 91: are we sure evil magic number is better than just dumping all the current lookup dicts somewhere 🌝

eyurtsev: never been sure of anything more in my life -- less work and easy to catch modifications

baskaryan: easier to understand if the actual dicts are there, and you've already got all the functionality in the tests to make a script to extract current dicts — but i defer to ya on this one 👍

baskaryan (approved):

PR #22002, opened 2024-05-21
core[patch]: RFC: Allow concatenation of messages with multi part content

jacoblee93:
Anthropic's streaming treats tool calls as different content parts (streamed back with a different index) from normal content in the `content`.

This means that we need to update our chunk-merging logic to handle chunks with multi-part content. The alternative is coerceing Anthropic's responses into a string, but we generally like to preserve model provider responses faithfully when we can. This will also likely be useful for multimodal outputs in the future.

This current PR does unfortunately make `index` a magic field within content parts, but Anthropic and OpenAI both use it at the moment to determine order anyway. To avoid cases where we have content arrays with holes and to simplify the logic, I've also restricted merging to chunks in order.

TODO: tests

CC @baskaryan @ccurme @efriis


libs/core/langchain_core/messages/base.py
```diff
@@ -95,6 +95,38 @@ def merge_content(
...
+            raise Exception(
+                f"Invalid index value {second_index} found when concatenating chunks"
+            )
+        if isinstance(first_content, List):
+            if second_index > len(first_content):
+                raise Exception(
+                    f"Failed to concatenate a chunk with index {second_index}",
+                    f"to a chunk that only had {len(first_content)} parts.\n"
+                    "You must concatenate message content chunks in order.",
+                )
```
baskaryan, line 112: why couldn't second chunk refer to an item after all the items in first chunk?

jacoblee93: It can, but it must be the very next one. Otherwise you get holey arrays, which sounded ugly. We *could* do it

libs/core/langchain_core/messages/base.py
```diff
@@ -95,6 +95,38 @@ def merge_content(
         else:
             return_list: List[Union[str, Dict]] = [first_content]
             return return_list + second_content
+    elif isinstance(second_content, dict) and isinstance(
```
baskaryan, line 98: contents are only supposed to be strs or lists, no?

jacoblee93: No? `{"type":"text","text":" San"}`

libs/core/langchain_core/messages/base.py
baskaryan: doesn't feel like something we should be enforcing imo

baskaryan (commented):
some unit tests would help understand expected behavior

jacoblee93:
Tests added

libs/core/langchain_core/messages/base.py
```diff
@@ -95,9 +95,33 @@ def merge_content(
...
+            if not isinstance(part, Dict):
+                return first_content + second_content
+            part_index = part.get("index")
+            if isinstance(part_index, int) and part_index >= 0:
+                if part_index > len(final_content) - 1:
+                    final_content.extend([None] * (part_index - len(final_content) + 1))
+                if final_content[part_index] is None:
+                    final_content[part_index] = part
+                elif isinstance(final_content[part_index], Dict):
+                    final_content[part_index] = merge_dicts(
```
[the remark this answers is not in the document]
jacoblee93: Ah cool, will explore.

libs/core/tests/unit_tests/test_messages.py
```diff
@@ -162,6 +162,42 @@ def test_chat_message_chunks() -> None:
...
+        == AIMessageChunk(content=[{"index": 0, "text": "I am indeed."}])
+    ), "Concatenating when both content arrays are dicts with the same index should merge"  # noqa: E501
+
+    assert (
+        AIMessageChunk(content=[{"index": 0, "text": "I am"}])
+        + AIMessageChunk(content=[{"index": 2, "text": " indeed."}])
+        == AIMessageChunk(
+            content=[{"index": 0, "text": "I am"}, {"index": 2, "text": " indeed."}]
+        )
+    ), "Concatenating when both content arrays are dicts with a gap between indexes should not result in a holey array"  # noqa: E501
```
jacoblee93, line 190: Are we ok with this behavior? Holey arrays cause type-checking errors.

We could also sort each time we merge in case chunks come out of order (if chunk 1 comes after chunk 2 for some reason), but this seems unlikely.

libs/core/langchain_core/messages/base.py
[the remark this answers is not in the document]
jacoblee93: Seems like the right thing to do.

libs/core/langchain_core/messages/base.py
```diff
@@ -95,9 +95,13 @@ def merge_content(
...
             return_list: List[Union[str, Dict]] = [first_content]
             return return_list + second_content
-    # If both are lists, merge them naively
     elif isinstance(second_content, List):
-        return first_content + second_content
+        # If both are lists
+        merged_list = merge_lists(first_content, second_content)
+        # Satisfy type-checker
+        if merged_list is None:
+            raise Exception("Failed to merge message chunk contents.")
```
ccurme, line 103: (nit) if we know this will be a list can also import `cast` from `typing` and do `return cast(list, merged_list)`

libs/core/langchain_core/messages/base.py
```diff
@@ -95,9 +95,13 @@ def merge_content(
         else:
             return_list: List[Union[str, Dict]] = [first_content]
             return return_list + second_content
-    # If both are lists, merge them naively
     elif isinstance(second_content, List):
-        return first_content + second_content
+        # If both are lists
+        merged_list = merge_lists(first_content, second_content)
```
ccurme, line 100: only break I can think of is if one list has a dict item with key `"index"` and the other doesn't. e.g., `merge_lists([{"a": 3}], [{"a": 3, "index": 2}])` will raise, whereas before we would just add the lists. but unclear to me if this condition happens.

jacoblee93: Yeah unclear what would make sense - I think ok to raise? I can add a test case

libs/core/langchain_core/messages/base.py
jacoblee93: Done

libs/core/langchain_core/messages/base.py
jacoblee93: Ah it's the opposite right now. I think implicitly appending is ok too and actually may make more sense

jacoblee93: Unresolve if you disagree

baskaryan (approved):

PR #22007, opened 2024-05-22
docs: fix partner api ref build

baskaryan:


efriis (approved):
If we switch to pathlib these are more intuitive imo but if it works it works

PR #22011, opened 2024-05-22
community:jinachat[patch]: standardize init args

cahughes95:
Description:
updated jinachat_api_key and request_timeout so that they aliased to api_key, and timeout respectively. Added test that both continue to set the same underlying attributes in a new file titled test_jinachat.py

Issue: 20085
community:javelin[patch]: standardize init args

updated jinachat_api_key so that it aliased to api_key.

Related to [20085](https://github.com/langchain-ai/langchain/issues/20085)

Dependencies: None

### DIFF
--- a/libs/community/langchain_community/chat_models/javelin_ai_gateway.py
+++ b/libs/community/langchain_community/chat_models/javelin_ai_gateway.py
@@ -18,7 +18,7 @@ from langchain_core.outputs import (
     ChatGeneration,
     ChatResult,
 )
-from langchain_core.pydantic_v1 import BaseModel, Extra, SecretStr
+from langchain_core.pydantic_v1 import BaseModel, Extra, SecretStr, Field
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


libs/community/tests/unit_tests/chat_models/test_jinachat.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+
+import pytest
+
+from langchain_community.chat_models import JinaChat
+
+os.environ["JINACHAT_API_KEY"] = "foo"
+
+
+@pytest.mark.requires("openai")
```
baskaryan, line 12: is this supposed to be javelin_sdk?

libs/community/langchain_community/chat_models/javelin_ai_gateway.py
```diff
@@ -65,7 +65,7 @@ class ChatJavelinAIGateway(BaseChatModel):
     client: Any
     """javelin client."""
 
-    javelin_api_key: Optional[SecretStr] = None
```
baskaryan, line 68: think you'll need to add this to the ChatJavelinAIGateway class
```python
    class Config:
        """Configuration for this pydantic object."""
        allow_population_by_field_name = True
```

ccurme:

PR #22030, opened 2024-05-22
community[patch]: Add unit test to verify that init is correctly defined

eyurtsev:
Fix some __init__ files and add a unit test


leo-gan:
@eyurtsev It is not clear what is the correct way to make the "__init__.py" files. 

Can you, please, add the rules, reasons for the rules, and the ways to test them into the ut comments ?

Issue #21983, opened 2024-05-21
Standardize docstrings and improve coverage


baskaryan:
# Issue
Every public module, class, method and attribute should have a docstring. 

# Requirements
- All docstrings should follow the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods).
- Examples should use [RST code-block format](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block) so that they render in the API reference correctly.

### NOTE!
RST code block must have a newline between `.. code-block:: python` and the code example, and code example must be tabbed, to render correctly!

# Examples
## Module
`langchain/foo/__init__.py`
```python
""""One line summary.

More detailed paragraph.
"""
```

### Not Pydantic class
```python
class Foo:
  """One line summary.

  More detailed paragraph.
  
  Attributes:
    first_attr: does first thing.
    second_attr: does second thing.

  Example:
    .. code-block:: python
    
      from langchain.foo import Foo
  
      f = Foo(1, 2, "bar")
      ...
  """
  
  def __init__(self, a: int, b: int, c: str) -> None:
    """Initialize using a, b, c.
   
    Args:
      a: ...
      b: ...
      c: ...
   """
    self.first_attr = a + b
    self.second_attr = c
```

#### NOTE
If the object attributes and init args are the same then you can just document the init args for non-Pydantic classes and just document the attributes for Pydantic classes.

### Pydantic class
```python
from typing import Any

from langchain_core.base_models import BaseModel

class FooPydantic(BaseModel):
  """One line summary.

  More detailed paragraph.

  Example:
    .. code-block:: python
    
      from langchain.foo import Foo
  
      f = Foo(1, 2, "bar")
      ...
  """

  first_attr: int
  """Does first thing."""
  second_attr: str
  """Does second thing.

  Additional info if needed.
  """

    def __init__(self, a: int, b: int, c: str, **kwargs: Any) -> None:
    """Initialize using a, b, c.
   
    Args:
      a: ...
      b: ...
      c: ...
      **kwargs: ...
   """
    first_attr = a + b
    second_attr = c
    super().__init__(first_attr=first_attr, second_attr=second_attr, **kwargs)
```

## Function/method
```python
def bar(a: int, b: str) -> float:
  """One line description.

  More description if needed.

  Args:
    a: ...
    b: ...

  Returns:
    A float of ...

  Raises:
    ValueError: If a is negative.

  Example:
[...]

PR #22031, opened 2024-05-22
community[patch]: add detailed paragraph and example for BaichuanTextEmbeddings

maang-h:
- **Description:** add detailed paragraph and example for BaichuanTextEmbeddings
   - **Issue:** the issue #21983

### DIFF
--- a/libs/community/langchain_community/embeddings/baichuan.py
+++ b/libs/community/langchain_community/embeddings/baichuan.py
@@ -23,11 +23,23 @@ BAICHUAN_API_URL: str = "http://api.baichuan-ai.com/v1/embeddings"
 # NOTE!! BaichuanTextEmbeddings only supports Chinese text embedding.
 # Multi-language support is coming soon.
 class BaichuanTextEmbeddings(BaseModel, Embeddings):
-    """Baichuan Text Embedding models."""
+    """Baichuan Text Embedding models.
+
+    To use, you should set the environment variable ``BAICHUAN_API_KEY`` to
+    your API key or pass it as a named parameter to the constructor.
+
+    Example:
+        .. code-block:: python
+
+            from langchain_community.embeddings import BaichuanTextEmbeddings
+
+            baichuan = BaichuanTextEmbeddings(baichuan_api_key="my-api-key")
+    """
 
     session: Any  #: :meta private:
     model_name: str = "Baichuan-Text-Embedding"
     baichuan_api_key: Optional[SecretStr] = None
+    """Automatically inferred from env var `BAICHUAN_API_KEY` if not provided."""
 
     @root_validator(allow_reuse=True)
     def validate_environment(cls, values: Dict) -> Dict:
--- a/libs/community/langchain_community/embeddings/baichuan.py
+++ b/libs/community/langchain_community/embeddings/baichuan.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/embeddings/baichuan.py
```diff
@@ -79,15 +94,14 @@ def _embed(self, texts: List[str]) -> Optional[List[List[float]]]:
...
-                print(  # noqa: T201
+                # Handle 100 <= status_code < 400, not include 200
+                raise RequestException(
                     f"Error: Received status code {response.status_code} from "
-                    "embedding API"
+                    "`BaichuanEmbedding` API"
                 )
-                return None
-        except Exception as e:
+        except Exception:
```
eyurtsev, line 102: can remove the try and except now since it doesn't do anything?

libs/community/langchain_community/embeddings/baichuan.py
```diff
@@ -79,15 +94,14 @@ def _embed(self, texts: List[str]) -> Optional[List[List[float]]]:
                 return [result.get("embedding", []) for result in sorted_embeddings]
             else:
                 # Log error or handle unsuccessful response appropriately
-                print(  # noqa: T201
+                # Handle 100 <= status_code < 400, not include 200
+                raise RequestException(
```
eyurtsev, line 98: This will raise exceptions for all 2xx except for 200 as well, which isn't correct behavior

Why not remove this exception entirely since `response.raise_for_status()` already raises?

maang-h: the method `response.raise_for_status()` just handle status_code from 400 to 600

```python
# requests -> models.py -> Response

def raise_for_status(self):
    """Raises :class:`HTTPError`, if one occurred."""

    http_error_msg = ""
    if isinstance(self.reason, bytes):
        # We attempt to decode utf-8 first because some servers
        # choose to localize their reason strings. If the string
        # isn't utf-8, we fall back to iso-8859-1 for all other
        # encodings. (See PR #3538)
        try:
            reason = self.reason.decode("utf-8")
        except UnicodeDecodeError:
            reason = self.reason.decode("iso-8859-1")
    else:
        reason = self.reason

    if 400 <= self.status_code < 500:
        http_error_msg = (
            f"{self.status_code} Client Error: {reason} for url: {self.url}"
        )

    elif 500 <= self.status_code < 600:
        http_error_msg = (
            f"{self.status_code} Server Error: {reason} for url: {self.url}"
        )

    if http_error_msg:
        raise HTTPError(http_error_msg, response=self)
```

libs/community/langchain_community/embeddings/baichuan.py
maang-h: done

libs/community/langchain_community/embeddings/baichuan.py
maang-h: @eyurtsev Can you give me some suggestions for improvement?

eyurtsev (approved):

libs/community/langchain_community/embeddings/baichuan.py
eyurtsev: I'm merging this version. If you want feel free to follow up with a PR. The conditional if response.status==200 can be changed to just accept 2xx codes. Not super critical here, but code looks a bit odd where an error message is rasied from a 2xx code that isn't 200

PR #22039, opened 2024-05-22
langchain[minor]: add universal init_model

baskaryan:
decisions to discuss
- only chat models
- model_provider isn't based on any existing values like llm-type, package names, class names
- implemented as function not as a wrapper ChatModel
- function name (init_model)
- in langchain as opposed to community or core
- marked beta


ccurme:
langchain feels like a natural place for this. but if we (a) put this in langchain, and (b) encourage this method of initializing chat models, then users will need to install langchain where they previously just needed core.

ccurme:
currently we aim to have all chat model integrations support the same interface. a wrapper ChatModel would allow us some flexibility if a single provider diverges wildly from LangChain conventions (e.g., we can let the integration chat model do its own thing, but enforce conventions on the ChatModel version).

baskaryan:
> langchain feels like a natural place for this. but if we (a) put this in langchain, and (b) encourage this method of initializing chat models, then users will need to install langchain where they previously just needed core.

imo fine to require langchain now that it doesn't depend on community

> currently we aim to have all chat model integrations support the same interface. a wrapper ChatModel would allow us some flexibility if a single provider diverges wildly from LangChain conventions (e.g., we can let the integration chat model do its own thing, but enforce conventions on the ChatModel version).

imo if an integration isn't implementing BaseChatModel it doesn't belong here anyways

the pro of not doing a wrapper ChatModel is that this way any custom functionality (like bind_tools) is preserved on the returned object

eyurtsev:
Are we sure we want to add? When users use `langchain_fireworks` or some other partner, they know exactly where to go to see documentation (APIReference) for model parameters, they can also use tab completion in an IDE setting to read the doc-string associated with the model.

If we do it this way, users won't know:

1) which package code comes from
2) what parameters it supports etc

and they don't have a way of listing the models that are available through this init method

libs/langchain/langchain/chat_models/base.py
```diff
@@ -10,4 +14,179 @@
...
+
+            gpt_3 = init_model("gpt-4o", model_provider="openai", temperature=0)
+            claude_opus = init_model("claude-3-opus-20240229", model_provider="anthropic", temperature=0)
+            gemini_15 = init_model("gemini-1.5-pro", model_provider="google_vertexai", temperature=0)
+
+            gpt_3.invoke("what's your name")
+            claude_opus.invoke("what's your name")
+            gemini_15.invoke("what's your name")
+    """  # noqa: E501
+    model_provider = model_provider or _attempt_infer_model_provider(model_name)
```
eyurtsev, line 69: What qualifications must a chat model meet to be included here?

libs/langchain/langchain/chat_models/base.py
```diff
@@ -10,4 +14,179 @@
...
+        return "google-vertexai"
+    elif model_name.startswith("amazon."):
+        return "bedrock"
+    else:
+        return None
+
+
+def _check_pkg(pkg: str) -> None:
+    try:
+        importlib.import_module(pkg)
```
eyurtsev, line 186: importlib.util.find_spec ?

baskaryan:
> Are we sure we want to add? When users use `langchain_fireworks` or some other partner, they know exactly where to go to see documentation (APIReference) for model parameters, they can also use tab completion in an IDE setting to read the doc-string associated with the model.
> 
> If we do it this way, users won't know:
> 
> 1. which package code comes from
> 2. what parameters it supports etc
> 
> and they don't have a way of listing the models that are available through this init method

goal of this is to enable developers of other packages who want to let their users easily access many models via simple configs. just bc we add this method doesn't mean we need to recommend it as default way to instantiate model

libs/langchain/langchain/chat_models/base.py
eyurtsev: recognized model_provider below

baskaryan: any ChatModel can be added. can update with more if we want

baskaryan: until requested will stick only to models with their own integration packages, not langchain-community ones

PR #22052, opened 2024-05-23
cli: switch migration CLI to GritQL

morgante:
**Description**
Replaces the libCST integration with GritQL. This has a few advantages:
* There's less code to maintain, as Grit handles the complexities of actually replacing imports and remapping correctly (notice the big red reduction).
* Grit includes advanced features like `--interactive` to allow interactive review of changes.

**Dependencies**
- Added gritQL

**Twitter handle**
https://x.com/morgantepell

**Testing**
Existing tests continue to run correctly for the cli_runner.


eyurtsev:
how do i run it?

```
(cli) ➜  langchain git:(c50273684) ✗ langchain-cli migrate .
Running GritQL pattern with args: /home/eugene/.cache/openai-python/.install/bin/marzano ['apply', 'langchain_all_migrations()', '.', '--grit-dir', PosixPath('/home/eugene/src/langchain/libs/cli/langchain_cli/namespaces/migrate/.grit')]
error: unexpected argument '--grit-dir' found

  tip: to pass '--grit-dir' as a value, use '-- --grit-dir'

Usage: grit apply <PATTERN_OR_WORKFLOW> <PATHS>...
```

Could we update the cache dir to either be library netural?

```
    dir_name = _cache_dir() / "openai-python"
    install_dir = dir_name / ".install"
```

As a user the the name of the `marzano` binary is a bit confusing, unclear what relationship it has to grit

morgante:
> how do i run it?

I will need to cut a release on the Grit side before this is runnable - I'd love feedback on the code / confirmation of this approach to make sure it's aligned before doing that.

> Could we update the cache dir to either be library netural?
> 
> ```
>     dir_name = _cache_dir() / "openai-python"
>     install_dir = dir_name / ".install"
> ```

Yes I'll change it to `~/.grit`.

> As a user the the name of the `marzano` binary is a bit confusing, unclear what relationship it has to grit

This ideally shouldn't be user facing at all.

eyurtsev:
> I will need to cut a release on the Grit side before this is runnable - I'd love feedback on the code / confirmation of this approach to make sure it's aligned before doing that.

Overall looks good. I need we need the following prior to merging:

1) I need to be able to run it to verify correctness -- I'll double check correctness side by side with existing migration script.
2) ability to run on ipython notebooks since we rely on the migration script internally as well to migrate our own docs.
3) Update documentation for how to use the migration script the CLI
4) Upgrade documentation [here](https://python.langchain.com/v0.2/docs/versions/v0_2/) -- I can help with this

eyurtsev:
I wanted to check in are you still planning on working on this? Anything else needed from me?

morgante:
> I wanted to check in are you still planning on working on this? Anything else needed from me?

Yeah, sorry for the delay! Parsing Jupyter proved a bit more complicated than I expected, but I should be able to wrap it up over the weekend.

eyurtsev:
Does grit.io have any solution to escape into imperative mode that allows writing the modifications in python? I now worked with ast-grep a lot -- the pattern matching (which is likely similar to grit.io) is very convenient, but re-writes can still be challenging if they require changing more than one kind of thing. For example, update the type signature from Iterable to List in matching methods + add `from typing import List` into the global namespace.

morgante:
> Does grit.io have any solution to escape into imperative mode that allows writing the modifications in python?  

Not currently. We would consider building it, but honestly only if I saw genuine interest and usage of Grit already. If you're already preferring alternatives there's no reason for me to keep building.

> I now worked with ast-grep a lot -- the pattern matching (which is likely similar to grit.io) is very convenient

Would love to understand why you prefer ast-grep to Grit. They are very similar and I don't think you should need to use two tools. If there's something in our UX you aren't understanding / finding simpler with AST-grep that's great to know.

> For example, update the type signature from Iterable to List in matching methods + add from typing import List into the global namespace.

This is straightforward to do with GritQL.

```
`def matching_method($_): $type:
  $_` where {
    $type <: `Iterable` => `List`,
    add_import(source="typing", name="List")
```

eyurtsev:
Hi @morgante,

There are two different end users here: (1) langchain users for code mods, (2) me for doing internal refactors in the code base.

If I understand correctly, your goal is to roll out code mods to LangChain users that use grit as the underlying technology. I'm in support as it also reduces maintenance burden for core developers. The requirements for adopting the solution haven't changed: https://github.com/langchain-ai/langchain/pull/22052#issuecomment-2127747190. but it seems like the work has been deprioritized as it's been 2 months?

I'm looking into code-modding tools for internal refactors rather than distributing code mods to users -- so here the end user is just me. If we adopt grit for distributing code mods to users, then we'll use gritql for that purpose.

morgante:
> The requirements for adopting the solution haven't changed: https://github.com/langchain-ai/langchain/pull/22052#issuecomment-2127747190. but it seems like the work has been deprioritized as it's been 2 months?

Yeah, it has been on the backburner thus far.

> so here the end user is just me

Understood. I'd still love your feedback / to understand why ast-grep is more compelling for you. If we're putting in the time to integrate Grit into langchain, I'd appreciate at least understanding why you wouldn't use it for everything - anything you do with ast-grep should be equally doable with Grit.

eyurtsev:
> Understood. I'd still love your feedback / to understand why ast-grep is more compelling for you. If we're putting in the time to integrate Grit into langchain, I'd appreciate at least understanding why you wouldn't use it for everything - anything you do with ast-grep should be equally doable with Grit.

I don't have a preference -- just exploring available tools. Started with ast-grep b/c I've used it a bit in the past.

In terms of things that **might** be nice for internal refactors:

1. Being able to plug in an LLM for certain kinds of re-writes (e.g., update doc-strings). I can imagine doing this with the python API (https://ast-grep.github.io/guide/api-usage/py-api.html)

2. As far as I can tell none of the tools allows incorporating information collected during run time. I'd imagine running a test suite to collect type information and make it easier to roll out some kinds of re-writes (e.g., replace deprecated method with the newer method name). It feels like pattern matching this could be quite difficult in some cases (e.g., missing / incorrect type information in function signature, or accessing a class attribute from a method and not knowing the type of the attribute.)

morgante:
> Being able to plug in an LLM for certain kinds of re-writes (e.g., update doc-strings). I can imagine doing this with the python API (https://ast-grep.github.io/guide/api-usage/py-api.html)

We actually have built-in support for LLM rewrites using the `ai_transform` function. You can see an example [here](https://github.com/getgrit/stdlib/blob/542b4de5d98e3116f38dad5d0e11bdca95c4ccd3/.grit/patterns/universal/ai/_ai_transform.md#L4).

baskaryan:
closing since this migration has occurred, but really appreciate the help @morgante! happy to try out GritQL for future migrations

eyurtsev:
comandeering

eyurtsev:
i don't have edit permissions here so continuing here: https://github.com/langchain-ai/langchain/pull/26427

Issue #22021, opened 2024-05-22
It's impossible to use and **async** ChatMessageHistory with langchain-core. 


pprados:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
import asyncio
import uuid
from pprint import pprint

import psycopg
from langchain_core.chat_history import BaseChatMessageHistory
from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
from langchain_core.runnables.history import RunnableWithMessageHistory
from langchain_openai import ChatOpenAI

from langchain_postgres import PostgresChatMessageHistory

model = ChatOpenAI()
prompt = ChatPromptTemplate.from_messages(
    [
        (
            "system",
            "You're an assistant who's good at {ability}. Respond in 20 words or fewer",
        ),
        MessagesPlaceholder(variable_name="history"),
        ("human", "{input}"),
    ]
)
runnable = prompt | model

table_name = "chat_history"

async_connection = None


async def init_async_connection():
    global async_connection
    async_connection = await psycopg.AsyncConnection.connect(
        user="postgres",
        password="password_postgres",
        host="localhost",
        port=5432)


def aget_session_history(session_id: str) -> BaseChatMessageHistory:
    return PostgresChatMessageHistory(
        table_name,
        session_id,
        async_connection=async_connection
    )


awith_message_history = RunnableWithMessageHistory(
    runnable,
    aget_session_history,
    input_messages_key="input",
    history_messages_key="history",
)


async def amain():
    await init_async_connection()
    result = await awith_message_history.ainvoke(
        {"ability": "math", "input": "What does cosine mean?"},
        config={"configurable": {"session_id": str(uuid.uuid4())}},
    )
    pprint(result)


asyncio.run(amain())
```

### Error Message and Stack Trace (if applicable)

Error in RootListenersTracer.on_chain_end callback: ValueError('Please initialize the PostgresChatMessageHistory with a sync connection or use the aadd_messages method instead.')

### Description

# It's impossible to use and async ChatMessageHistory with langchain-core.

The `ChatMessageHistory` class is synchronous and doesn't have an async counterpart. 
[...]

PR #13200, opened 2023-11-10
core[minor]: Adds an in-memory implementation of RecordManager

pprados:
**Description:**
langchain offers three technologies to save data:
- [vectorstore](https://python.langchain.com/docs/modules/data_connection/vectorstores/)
- [docstore](https://js.langchain.com/docs/api/schema/classes/Docstore)
- [record manager](https://python.langchain.com/docs/modules/data_connection/indexing)

If you want to combine these technologies in a sample persistence stategy you need a common implementation for each. `DocStore` propose `InMemoryDocstore`. 

We propose the class `MemoryRecordManager` to complete the system.

This is the prelude to another full-request, which needs a consistent combination of persistence components.

**Tag maintainer:**
@baskaryan

**Twitter handle:**
@pprados

PR #22065, opened 2024-05-23
community[minor]: Add native async support to SQLChatMessageHistory

pprados:
# package community: Fix SQLChatMessageHistory

## Description
Here is a rewrite of `SQLChatMessageHistory` to properly implement the asynchronous approach. The code circumvents [issue 22021](https://github.com/langchain-ai/langchain/issues/22021) by accepting a synchronous call to `def add_messages()` in an asynchronous scenario. This bypasses the bug.

For the same reasons as in [PR 22](https://github.com/langchain-ai/langchain-postgres/pull/32) of `langchain-postgres`, we use a lazy strategy for table creation. Indeed, the promise of the constructor cannot be fulfilled without this. It is not possible to invoke a synchronous call in a constructor. We compensate for this by waiting for the next asynchronous method call to create the table.

The goal of the `PostgresChatMessageHistory` class (in `langchain-postgres`) is, among other things, to be able to recycle database connections. The implementation of the class is problematic, as we have demonstrated in [issue 22021](https://github.com/langchain-ai/langchain/issues/22021).

Our new implementation of `SQLChatMessageHistory` achieves this by using a singleton of type (`Async`)`Engine` for the database connection. The connection pool is managed by this singleton, and the code is then reentrant.

We also accept the type `str` (optionally complemented by `async_mode`. I know you don't like this much, but it's the only way to allow an asynchronous connection string).

In order to unify the different classes handling database connections, we have renamed `connection_string` to `connection`, and `Session` to `session_maker`.

Now, a single transaction is used to add a list of messages. Thus, a crash during this write operation will not leave the database in an unstable state with a partially added message list. This makes the code resilient.

We believe that the `PostgresChatMessageHistory` class is no longer necessary and can be replaced by:
```
PostgresChatMessageHistory = SQLChatMessageHistory
```
This also fixes the bug.

## Issue
- [issue 22021](https://github.com/langchain-ai/langchain/issues/22021)
  - Bug in _exit_history()
  - Bugs in PostgresChatMessageHistory and sync usage
  - Bugs in PostgresChatMessageHistory and async usage
 - [issue 36](https://github.com/langchain-ai/langchain-postgres/issues/36)
 ## Twitter handle:
pprados

## Tests
- libs/community/tests/unit_tests/chat_message_histories/test_sql.py (add async test)

@baskaryan, @eyurtsev or @hwchase17 can you check this PR ?
And, I've been waiting a long time for validation from other PRs. Can you take a look?
- [PR 32](https://github.com/langchain-ai/langchain-postgres/pull/32)
- [PR 15575](https://github.com/langchain-ai/langchain/pull/15575)
- [PR 13200](https://github.com/langchain-ai/langchain/pull/13200)


libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -80,36 +92,85 @@ def get_sql_model_class(self) -> Any:
...
+    def Session(self) -> Union[scoped_session, async_sessionmaker]:
+        logger.warning("Session is deprecated, use session_maker instead")
+        return self.session_maker
+
     def __init__(
         self,
         session_id: str,
-        connection_string: str,
+        connection_string: Optional[str] = None,
+        connection: Union[None, DBConnection] = None,
```
eyurtsev, line 110: This would be a breaking change for any users that used positional arguments to define the table_name.

Instead, could you place all the new named arguments at the end of the function signature and force them to be invoked as named arguments by users using the `*` operator.

Hopefully ruff ends a lint rule for this at some point. :)

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -80,36 +92,85 @@ def get_sql_model_class(self) -> Any:
...
+        if not self._table_created:
+            async with self.async_engine.begin() as conn:
+                await conn.run_sync(self.sql_model_class.metadata.create_all)
+            self._table_created = True
 
     @property
     def messages(self) -> List[BaseMessage]:  # type: ignore
         """Retrieve all messages from db"""
-        with self.Session() as session:
+        assert not self.async_mode, "This method must be called without async_mode"
```
eyurtsev, line 172: Could you push this assert statement inside `session_maker`?

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -123,18 +184,85 @@ def messages(self) -> List[BaseMessage]:  # type: ignore
                 messages.append(self.converter.from_sql_model(record))
             return messages
 
+    async def aget_messages(self) -> List[BaseMessage]:
+        """Retrieve all messages from db"""
+        assert self.async_mode, "This method must be called with async_mode"
+        await self._acreate_table_if_not_exists()
+        async with self.session_maker() as session:
```
eyurtsev, line 191: Could you change this to create an async session maker, and place relevant asserts inside it?

```python
async with self.asession_maker() as session:
```

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -80,36 +92,85 @@ def get_sql_model_class(self) -> Any:
...
 
+DBConnection = Union[AsyncEngine, Engine, str]
+
+
 class SQLChatMessageHistory(BaseChatMessageHistory):
     """Chat message history stored in an SQL database."""
 
+    @property
+    def Session(self) -> Union[scoped_session, async_sessionmaker]:
+        logger.warning("Session is deprecated, use session_maker instead")
```
eyurtsev, line 103: Could you use @deprecated decorator on top of the property if you want to deprecate this. It's in `langchain_core._api`. 

It has logic for:
1) Not raising warnings for internal calls to this code that the user has no control over (i.e., from within langchain code)
2) only showing the warning once
3) It uses a warning rather than logger.warning because in this case moving away from deprecated code is under control over the user

4) We can use it to auto-generate list of deprecation tables when publishing release notes (e.g., https://python.langchain.com/v0.2/docs/versions/v0_2/deprecations/)

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -80,36 +92,85 @@ def get_sql_model_class(self) -> Any:
...
+        engine_args: Optional[Dict[str, Any]] = None,
+        async_mode: Optional[bool] = None,  # Use only if connection is a string
         table_name: str = "message_store",
         session_id_field_name: str = "session_id",
         custom_message_converter: Optional[BaseMessageConverter] = None,
     ):
-        self.connection_string = connection_string
-        self.engine = create_engine(connection_string, echo=False)
+        if connection_string:
+            logger.warning("connection_string is deprecated, use connection instead")
```
eyurtsev, line 118: Could you swap to use `warn_deprecated` (git grep for usage of `warn_deprecated`)

You may need to memoize it so it doesn't flood the user with warnings on repeated instantiations.

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -80,36 +92,85 @@ def get_sql_model_class(self) -> Any:
...
+        connection: Union[None, DBConnection] = None,
+        engine_args: Optional[Dict[str, Any]] = None,
+        async_mode: Optional[bool] = None,  # Use only if connection is a string
         table_name: str = "message_store",
         session_id_field_name: str = "session_id",
         custom_message_converter: Optional[BaseMessageConverter] = None,
     ):
-        self.connection_string = connection_string
-        self.engine = create_engine(connection_string, echo=False)
+        if connection_string:
```
eyurtsev, line 117: Could you raise an exception if a user provides both connection and connection_string?

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -123,18 +184,85 @@ def messages(self) -> List[BaseMessage]:  # type: ignore
...
+        async with self.session_maker() as session:
+            session.add(self.converter.to_sql_model(message, self.session_id))
+            await session.commit()
+
+    def add_messages(self, messages: Sequence[BaseMessage]) -> None:
+        # The method RunnableWithMessageHistory._exit_history() call
+        #  add_message method by mistake and not aadd_message.
+        # See https://github.com/langchain-ai/langchain/issues/22021
+        if self.async_mode:
+            loop = asyncio.get_event_loop()
```
eyurtsev, line 229: We generally raise exceptions here instead of kicking off another event loop.

We need to resolve a design issue with `RunnableWithMessageHistory` before I'd feel OK breaking our rule of generally not invoking async code from sync code

libs/core/langchain_core/chat_history.py
```diff
@@ -138,6 +155,23 @@ def add_ai_message(self, message: Union[AIMessage, str]) -> None:
         else:
             self.add_message(AIMessage(content=message))
 
+    async def aadd_ai_message(self, message: Union[AIMessage, str]) -> None:
```
eyurtsev, line 158: add_ai_message should be treated as deprecated -- we'll likely deprecate it sometime soon, so let's not introduce `add_ai_message` here

libs/core/langchain_core/chat_history.py
```diff
@@ -154,6 +188,22 @@ def add_message(self, message: BaseMessage) -> None:
                 "Please implement add_message or add_messages."
             )
 
+    async def aadd_message(self, message: BaseMessage) -> None:
```
eyurtsev, line 191: Not sure about this... I'd rather discourage use of non batched variants in chat history -- the non batched variant is a design mistake since it encourages extra latency due to round-trips

eyurtsev (commented):
Almost ready for merging. 

Most important things to address:

1) Undo changes in core
2) Undo the code that calls async code from a sync function using a new event loop 
3) Create a separate async session maker in chat history, so we can consolidate the asserts in just two places

libs/community/langchain_community/chat_message_histories/sql.py
pprados: done

libs/core/langchain_core/chat_history.py
pprados: done

libs/core/langchain_core/chat_history.py
pprados: I agree. Done

libs/community/langchain_community/chat_message_histories/sql.py
pprados: Ok, but you must fix the bug describe [here](https://github.com/langchain-ai/langchain/issues/22021), in `_exit_history()`.
I'd be delighted to remove this ugly code ;-)

but for the moment it's an acceptable temporary solution.

libs/community/langchain_community/chat_message_histories/sql.py
pprados: No. Because the 
```
with self.session_maker() as session:
```
is different with
```
async with self.session_maker() as session:
```

libs/community/langchain_community/chat_message_histories/sql.py
pprados: done

libs/community/langchain_community/chat_message_histories/sql.py
pprados: No. Because the 
```
with self.session_maker() as session:
```
is different with
```
async with self.session_maker() as session:
```

libs/community/langchain_community/chat_message_histories/sql.py
pprados: done

libs/community/langchain_community/chat_message_histories/sql.py
pprados: And, the self.session_maker is a *variable*. It's may be changed by the user. If the user changes the value, the assertion is no longer present

libs/community/langchain_community/chat_message_histories/sql.py
eyurtsev: Could you add something like this, so the asserts are always applied?


```python
    @contextlib.contextmanager
    def _make_sync_session(self) -> Session:
        """Make an async session."""
        if self.async_mode:
            raise ValueError(
                "Attempting to use a sync method in when async mode is turned on. "
                "Please use the corresponding async method instead."
            )
        with self.session_maker() as session:
            yield cast(Session, session)

    @contextlib.asynccontextmanager
    async def _make_async_session(self) -> AsyncSession:
        """Make an async session."""
        if not self.async_mode:
            raise ValueError(
                "Attempting to use an async method in when sync mode is turned on. "
                "Please use the corresponding async method instead."
            )
        async with self.session_maker() as session:
            yield cast(AsyncSession, session)
```

libs/community/langchain_community/chat_message_histories/sql.py
eyurtsev: ok that's reasonable

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -80,36 +93,93 @@ def get_sql_model_class(self) -> Any:
...
+        engine_args: Optional[Dict[str, Any]] = None,
+        async_mode: Optional[bool] = None,  # Use only if connection is a string
     ):
-        self.connection_string = connection_string
-        self.engine = create_engine(connection_string, echo=False)
+        assert not (
+            connection_string and connection
+        ), "connection_string and connection are mutually exclusive"
+        if connection_string:
+            warn_deprecated(
```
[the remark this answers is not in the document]
pprados: done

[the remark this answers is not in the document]
pprados: done

eyurtsev:
@pprados back from vacation -- this looks ready to merge, I'll take a final look tomorrow and resolve merge conflict

libs/community/langchain_community/chat_message_histories/sql.py
```diff
@@ -80,36 +108,98 @@ def get_sql_model_class(self) -> Any:
...
+
     def __init__(
         self,
         session_id: str,
-        connection_string: str,
+        connection_string: Optional[str] = None,
         table_name: str = "message_store",
         session_id_field_name: str = "session_id",
         custom_message_converter: Optional[BaseMessageConverter] = None,
+        connection: Union[None, DBConnection] = None,
```
eyurtsev, line 131: (unrelated to this PR)

---

@pprados this pattern could be useful if you needed a transaction (e.g., as was the case in the indexing API if I recall correctly)

Allows passing a session maker as part of the __init__ instead of initializing and then mutating a class attribute (the latter pattern is more likely to break in threaded scenarios)

eyurtsev (approved):

alew3:
Can you add in the documentation readme how to use async connections? Even with the PR, I'm getting errors.

pprados:
Use:
```
history_engine = create_async_engine(url="postgresql+psycopg://user:password@localhost:5432)

def get_session_history(session_id: str) -> BaseChatMessageHistory:
    return SQLChatMessageHistory( 
        session_id=session_id,
        connection=history_engine,
        engine_args={"echo": False},
    )
```

alew3:
@pprados thank you! You saved me a lot of head scratching! This should be in the docs!  
One more thing, do you know if its possible to pass metadata to go to the logs? I found it strange that it didn't save in the history when the messages happened so I tried:

 async for chunk in conversational_rag_chain.astream(
            {"input": question},
            config=RunnableConfig(configurable={"session_id": session_id },metadata={ "timestamp": datetime.now()}, callbacks=[])
        )

But the metadata is not persisted.

alew3:
Also, any reason why using it could 
makes thing take a long time when retrieving chat history from db?

Issue #21726, opened 2024-05-15
AZURE_OPENAI_ENDPOINT env var conflicts with OPENAI_API_BASE, while OPENAI_API_BASE is being used in MLFlow to point to Azure OpenAI endpoint.


loomlike:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Environment variables:
```
# for langchain
OPENAI_API_VERSION='...'
AZURE_OPENAI_API_KEY='my-key'
AZURE_OPENAI_ENDPOINT='my-azure-openai-endpoint'

# for mlflow -- this conflicts
OPENAI_API_BASE='my-azure-openai-endpoint'
```
Code to cause the error:
```
from langchain_openai import AzureOpenAIEmbeddings, AzureChatOpenAI

model = AzureChatOpenAI(
    azure_deployment="my-deployment-name",
)

# or 
model = AzureOpenAIEmbeddings(
    azure_deployment="my-deployment-name",
)
```

### Error Message and Stack Trace (if applicable)

```
ValidationError: 1 validation error for AzureChatOpenAI
__root__
  As of openai>=1.0.0, Azure endpoints should be specified via the `azure_endpoint` param not `openai_api_base` (or alias `base_url`). (type=value_error)
```

### Description

`AzureOpenAIEmbeddings` and `AzureChatOpenAI` automatically pick up the env vars and throwing errors complaining about having both `azure_endpoint` param and `openai_api_base`. 

`OPENAI_API_BASE` env var is being used in other packages, like MLFlow.

### System Info

```
#pip freeze | grep 'langchain\|mlflow'
langchain==0.1.20
langchain-community==0.0.38
langchain-core==0.1.52
langchain-openai==0.1.6
langchain-text-splitters==0.0.1
mlflow==2.12.2
```

chip-davis:
Have you tried specifying everything when you create the model? 
For example:
```
model = AzureChatOpenAI(
            openai_api_version="2023-05-15",
            azure_deployment=os.getenv("AZURE_OPENAI_DEPLOYMENT_NAME"),
            temperature=0,
            azure_endpoint=os.getenv("AZURE_OPENAI_ENDPOINT"),
[...]

PR #22068, opened 2024-05-23
partners[azure]: fix having openai_api_base set for other packages

chip-davis:
This fix is for #21726. When having other packages installed that require the `openai_api_base` environment variable, users are not able to instantiate the AzureChatModels or AzureEmbeddings.

This PR adds a new value `ignore_openai_api_base` which is a bool. When set to True, it sets `openai_api_base` to `None`

Two new tests were added for the `test_azure` and a new file `test_azure_embeddings`

A different approach may be better for this. If you can think of better logic, let me know and I can adjust it.


libs/partners/openai/langchain_openai/chat_models/azure.py
```diff
@@ -548,8 +549,10 @@ def validate_environment(cls, values: Dict) -> Dict:
         values["openai_api_key"] = (
             convert_to_secret_str(openai_api_key) if openai_api_key else None
         )
-        values["openai_api_base"] = values["openai_api_base"] or os.getenv(
-            "OPENAI_API_BASE"
+        values["openai_api_base"] = (
+            values["openai_api_base"]
+            if "openai_api_base" in values
+            else os.getenv("OPENAI_API_BASE")
```
efriis, line 555: thanks! Just switched logic here to respect `openai_api_base=None` instead

efriis (approved):

libs/partners/openai/langchain_openai/chat_models/azure.py
chip-davis: Nice! I like that logic better.

PR #22071, opened 2024-05-23
[partners/pinecone]: Add id_prefix as a parameter to PineconeVectorStore

cwaddingham:
This change adds a new parameter, `id_prefix` when creating a PineconeVectorStore object. This allows the `list` operation to filter on ID prefixes and the option to delete all records with a common parent document. 

See the following Pinecone documentation entries for more details:

- [List all records with a common prefix](https://docs.pinecone.io/guides/data/list-record-ids#list-the-ids-of-records-with-a-common-prefix)
- [Delete all records with a common parent document](https://docs.pinecone.io/guides/data/manage-rag-documents#delete-all-records-for-a-parent-document)


eyurtsev:
I don't think we want to support this functionality. This is doable on the user side right now

eyurtsev (changes requested):
I don't think this is a feature we want to add

eyurtsev (approved):

efriis:
Hey @cwaddingham is this something you still want merged? Looks like there have been some persistent CI issues if you could take a look at those and the merge conflicts, and reopen!

PR #22089, opened 2024-05-23
cli: model name substitution fix, release 0.0.23

efriis:


libs/cli/langchain_cli/integration_template/integration_template/llms.py
```diff
@@ -41,7 +42,7 @@ def _generate(
     ) -> LLMResult:
         raise NotImplementedError
 
-    # TODO: Implement if __model_name__ supports async generation. Otherwise
```
baskaryan, line 44: unrelated but should we also add a TODO for _generate, since that's the only required TODO

baskaryan (approved):

libs/cli/langchain_cli/integration_template/integration_template/llms.py
efriis: added

PR #22095, opened 2024-05-23
core[minor]: Add renderer

eyurtsev:
Add renderer


eyurtsev:
This would enable introducing a code in langchain_community for rendering the graphs in an interactive environment

https://github.com/langchain-ai/langchain/pull/23258/files#diff-ca82ff4c3a88bfb98e0336c8421064c7c7788e17fad58535684e4539cd00debe

PR #22102, opened 2024-05-24
retrieval scratch

baskaryan:

- [ ] **PR title**: "package: description"

Additional guidelines:


libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,120 @@
+####################
+# Interfaces
+####################
+
+
+class Index(ABC):
```
baskaryan, line 6: currently just a BaseStore that accepts kwargs. should it just be BaseStore?

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+GetResponse = List[Document]
+
+
+class RetrievalResponse(TypedDict, total=False):
+    hits: List[Hit]
+    metadata: dict
+
+    @property
+    def documents(self) -> List[Document]:
+        return [h.document for h in hits]
```
ccurme, line 113:
```suggestion
        return [h.source for h in hits]
```

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class Hit(TypedDict):
+    source: Document
+    snippet: Optional[Tuple[float, float]]
```
ccurme, line 118: why float? is this intended to represent a span?

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+    id: str
+    score: float
```
ccurme, line 120: might make sense for some of these to be optional (or set total=False)

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+            params = {**input, **kwargs}
+        else:
+            params = {"query": input, **kwargs}
+        result = self.retrieve(**params)
+        # TODO: callback manager stuff
+        return result
+
+    # QUESTION: Anything to do with streaming?
+
+    def as_runnable(
```
ccurme, line 58: (nit) `with_output_format`?

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+            raise ValueError
+
+
+class VectorStoreV2(Index, RetrieverV2[Union[str, Sequence[float]]]):
+    @abstractmethod
+    def retrieve(
+        self,
+        query: Union[str, Sequence[float]],
+        *,
+        method: str = "similarity",
```
ccurme, line 82: what will be captured by `method`?

baskaryan: mmr, hybrid, (anything else provider specific)

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+        metric: str = "cosine_similarity",
+        **kwargs: Any,
+    ) -> RetrievalResponse:
+        """Retrieve documents by query"""
+
+
+####################
+# TYPES
+####################
+
```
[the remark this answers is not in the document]
baskaryan: as in they don't error if you try double-adding/deleting?

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+
+class AddResponse(TypedDict):
```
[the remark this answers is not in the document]
baskaryan: iirc bulk operations on eg elasticsearch aren't guaranteed to fully succeed/fail, they can partially succeed. tho probably if we wanted to support this we should return ids of successes and/or failures instead of counts (so you can retry)

libs/core/langchain_core/vectorstores_v2.py
```diff
@@ -0,0 +1,122 @@
...
+    hits: List[Hit]
+    metadata: dict
+
+    @property
+    def documents(self) -> List[Document]:
+        return [h.document for h in hits]
+
+
+class Hit(TypedDict):
+    source: Document
```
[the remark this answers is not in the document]
baskaryan: does multimodality come down to making the source multimodal

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,179 @@
...
+                    prefix = prefix or ""
+                    for key in self.store:
+                        if key.startswith(prefix):
+                            yield key
+    """  # noqa: E501
+
+    @abstractmethod
+    def add(
+        self,
+        documents: Sequence[Document],
```
baskaryan, line 67: make accept Iterable

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,179 @@
...
+            ids: IDs of the documents to delete.
+
+        Returns:
+           A dict ``{"succeeded": [...], "failed": [...]}`` with the IDs of the
+           documents that were successfully deleted and the ones that failed to be
+           deleted.
+        """
+
+    @abstractmethod
+    def get_by_ids(self, ids: Union[List[str], Tuple[str]]) -> List[Document]:
```
baskaryan, line 88: add lazy version

libs/partners/qdrant/langchain_qdrant/vectorstores/v2.py
```diff
@@ -0,0 +1,120 @@
...
+                prefer_grpc=prefer_grpc,
+                https=https,
+                api_key=api_key,
+                prefix=prefix,
+                timeout=timeout,
+                host=host,
+                path=path,
+            )
+
+        if not async_client or location == ":memory:" or path is not None:
```
baskaryan, line 62: parentheses after first `not`

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,222 @@
...
+        """Lazily get documents by id.
+
+        Args:
+            ids: IDs of the documents to get.
+
+        Yields:
+           Document
+        """
+
+    def get_by_ids(self, ids: Iterable[str]) -> List[Document]:
```
eyurtsev, line 105: Sequence instead of Iterable perhaps to require a user to think about pagination

eyurtsev:
Issues with existing vectorstores that I think we should figure out whether we want to resolve or not

1. ability to fetch documents by id (without forcing a search) or scan by id
> Are we committed to make a vectorstore always store the document in addition to the vector?
2. standardize interface for search including filters
> what filtering syntax is supported?
> Can we the API to fail loudly if a user is attempting to use it incorrectly or with unsupported functionality? (e.g., not have unrecognized arguments be absorbed by **kwargs and result in silent failure)
3. expose `config` (`RunnableConfig`) In base abstractions that are runnables to allow passing run time information to child runnables

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Raises:
+            ValueError: if ids are not provided.
+        """
+        if ids is None:
+            raise ValueError("Must provide ids to delete.")
+        if filters:
+            kwargs = {"filters": filters, **kwargs}
+        if kwargs:
+            warnings.warn(
```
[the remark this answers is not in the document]
baskaryan: think there's a tradeoff between interoperability and how narrow the api is for each integration

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+)
+
+from langchain_core.documents import Document
+from langchain_core.indexes.types import DeleteResponse, UpsertResponse
+from langchain_core.runnables import run_in_executor
+from langchain_core.stores import BaseStore
+from langchain_core.structured_query import StructuredQuery
+
+
+class Index(BaseStore[str, Document], ABC):
```
eyurtsev, line 23: Should we add any ability to add unique hash to identify a document or information about time of modification / creation

eyurtsev:
https://github.com/langchain-ai/langchain/pull/21304

libs/core/langchain_core/indexes/base.py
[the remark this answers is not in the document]
eyurtsev: I'd be in favor of raising an exception here

libs/core/langchain_core/indexes/__init__.py
```diff
@@ -0,0 +1,4 @@
+from langchain_core.indexes.base import Index
```
eyurtsev, line 1: @baskaryan we have an `indexing` namespace in langchain-core. Let's keep in the same namespace, otherwise, it's going to be confusing?

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+
+        Default implementation, runs sync () in async executor.
+
+        Args:
+            ids: IDs of the documents to get.
+
+        Yields:
+           Document
+        """
+        return await run_in_executor(None, self.lazy_get_by_ids, ids)
```
cbornet, line 149: run_in_executor returns an Iterable, not an AsyncIterable.
I think run_in_executor(None, iterator, next) should be called in a loop

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+            warnings.warn(
+                "Only deletion by ids is supported for this integration, all other "
+                f"arguments are ignored. Received {kwargs=}"
+            )
+        return await self.adelete_by_ids(ids)
+
+    def lazy_get(
+        self,
+        *,
+        ids: Optional[Iterable[str]] = None,
```
eyurtsev, line 243: Thinking it makes sense to decouple lazy_get_by_ids from lazy_get with filters

```suggestion
```

(filters can support ids, if we add special syntax for the field?)

libs/core/langchain_core/indexes/types.py
```diff
@@ -0,0 +1,11 @@
+from typing import List, TypedDict
+
+
+class UpsertResponse(TypedDict):
+    succeeded: List[str]
```
eyurtsev, line 5:
```suggestion
    upserted: List[str]
```

libs/core/langchain_core/indexes/types.py
```diff
@@ -0,0 +1,11 @@
...
+    failed: List[str]
+
+
+class DeleteResponse(TypedDict):
```
eyurtsev, line 9: Let's get rid of this. It doesn't make sense on deletion operations

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+            raise ValueError("Must provide ids to get.")
+        if filters:
+            kwargs = {"filters": filters, **kwargs}
+        if kwargs:
+            warnings.warn(
+                "Only deletion by ids is supported for this integration, all other "
+                f"arguments are ignored. Received {kwargs=}"
+            )
+        return await self.alazy_get_by_ids(ids)
+
```
eyurtsev, line 307: Considering breaking into `get_by_id` and `get_by_filter`. `get_by_id` is a subset of `get_by_filter`. Use two distinct ones for ease of implementation and testing.

Later we could use `$ID` or `$.ID` to denote special cased fields in filters (if we ever cared)

---

There's a more general get by query that could support sorting by metadata fields (if we really cared to add that). It's generally useful for working with documents to sort by publication datetime. Seems like an unnecessary complication right now and not part of a basic vector index

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+    def delete(
+        self,
+        *,
+        ids: Optional[Iterable[str]] = None,
+        filters: Union[
+            StructuredQuery, Dict[str, Any], List[Dict[str, Any]], None
+        ] = None,
+        **kwargs: Any,
+    ) -> DeleteResponse:
```
eyurtsev, line 176: The `DeleteResponse` doesn't works -- it's impossible to know the IDs if using filters (especially if it fails). 

We should think about the kinds of failure cases that may occur:

1. network connection / bad API key -- most frequent failure
2. less frequent would be an API that raises an exception if the underlying ID doesn't exist -- but this isn't super common

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+                    prefix = prefix or ""
+                    for key in self.store:
+                        if key.startswith(prefix):
+                            yield key
+    """  # noqa: E501
+
+    @abstractmethod
+    def upsert(
+        self,
+        # TODO: Iterable or Iterator?
```
eyurtsev, line 86: Definitely Iterable rather than Iterator

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+        documents: Iterable[Document],
+        *,
+        ids: Optional[Iterable[str]] = None,
+        **kwargs: Any,
+    ) -> UpsertResponse:
```
eyurtsev, line 91: If we're making the input an Iterable, then the output should be an iterator over UpsertResponse

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def lazy_get(
+        self,
+        *,
+        ids: Optional[Iterable[str]] = None,
+        filters: Union[
+            StructuredQuery, Dict[str, Any], List[Dict[str, Any]], None
+        ] = None,
+        **kwargs: Any,
+    ) -> Iterable[Document]:
```
eyurtsev, line 248: We need to recover the `ID` of the document on the output if we're fetching by filter (i.e., we're fetching by filter b/c we don't know the ID).

We either:

1. Update `Document` to include id
2. Use a 2-tuple, dict etc more complex type that can include id
3. Subclass Document and return Document With ID

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+        if filters:
+            kwargs = {"filters": filters, **kwargs}
+        if kwargs:
+            warnings.warn(
+                "Only deletion by ids is supported for this integration, all other "
+                f"arguments are ignored. Received {kwargs=}"
+            )
+        return await self.adelete_by_ids(ids)
+
+    def lazy_get(
```
eyurtsev, line 240: What are semantics of lazy_get() -> is this a scan over all available documents?

libs/core/langchain_core/indexes/base.py
```diff
@@ -0,0 +1,369 @@
...
+    # FOR CONTRIBUTORS: Overwrite this in Index child classes that support native async upsertions.
+    async def aupsert(
+        self,
+        documents: AsyncIterable[Document],
+        *,
+        ids: Optional[AsyncIterable[str]] = None,
+        **kwargs: Any,
+    ) -> UpsertResponse:
+        """Upsert documents to index. Default implementation, runs sync upsert() in async executor."""
+        # TODO: how to convert AsyncIterable -> Iterable
```
cbornet, line 103: Here is an example: https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/document_loaders/base.py#L76-L84
Maybe there could be a utility method extracted for this.

cbornet, line 103: Such a good question for a coding test interview 😄 !
I would say:
* with materialization (simple but not the most efficient): `[item async for item in async_iterable]`
* without materialization: I think a queue is needed. We can't use an async queue because it must be accessed from different threads. So using a sync queue in non-blocking mode, something like:
```python
async def async_it():
    for i in range(10):
        yield i

async def test():
    q = queue.Queue(2)
    sentinel = object()

    async def put_into_queue(item):
        while True:
            try:
                q.put(item, block=False)
                break
            except queue.Full:
                await asyncio.sleep(0)

    async def put_all_into_queue():
        async for item in async_it():
            await put_into_queue(item)

        await put_into_queue(sentinel)

    def sync_it():
        while True:
            try:
                item = q.get(block=False)
                if item is sentinel:
                    break
                yield item
            except queue.Empty:
                time.sleep(0)

    coro = run_in_executor(None, sum, sync_it())

    result = await asyncio.gather(put_all_into_queue(), coro)
    print(result[1])
```

PR #22118, opened 2024-05-24
docs: Add explanation on how to use Hugging Face embeddings

tomaarsen:
- **Description:** I've added a tab on embedding text with LangChain using Hugging Face models to here: https://python.langchain.com/v0.2/docs/how_to/embed_text/. HF was mentioned in the running text, but not in the tabs, which I thought was odd.
- **Issue:** N/A
- **Dependencies:** N/A
- **Twitter handle:** No need, this is tiny :) 

Also, I had a ton of issues with the poetry docs/lint install, so I haven't linted this. Apologies for that. 

cc @Jofthomas 

- Tom Aarsen

### DIFF
--- a/docs/docs/how_to/embed_text.mdx
+++ b/docs/docs/how_to/embed_text.mdx
@@ -78,6 +78,31 @@ embeddings_model = CohereEmbeddings()
 ```
 
   </TabItem>
+  <TabItem value="huggingface" label="Hugging Face">
+
+To start we'll need to install the Hugging Face partner package:
+
+```bash
+pip install langchain-huggingface
+```
+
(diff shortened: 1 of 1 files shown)


tomaarsen:
I see that you've recently updated your documentation! It looks quite clean, well done. In my field, I see a lot of demand for 1) local models and 2) custom models. I'm kind of struggling with finding this in your documentation now, sadly. Do you think that you'll add such sections to your docs?

- Tom Aarsen

Jofthomas:
You are indeed right,
Documentation will be a priority for the next release imo. That and making multi-turn function calling work.

Thanks a lot @tomaarsen

ccurme (approved):

PR #22136, opened 2024-05-24
community[patch]: Update the default api_url and reqeust_body of sparkllm embedding

maang-h:
- **Description:** When I was running the SparkLLMTextEmbeddings, app_id, api_key and api_secret are all correct, but it cannot run normally using the current URL.

    ```python
    # example
    from langchain_community.embeddings import SparkLLMTextEmbeddings

    embedding= SparkLLMTextEmbeddings(
        spark_app_id="my-app-id",
        spark_api_key="my-api-key",
        spark_api_secret="my-api-secret"
    )
    embedding= "hello"
    print(spark.embed_query(text1))
    ```
    sparkembedding
   
    So I updated the url and request body parameters according to [Embedding_api](https://www.xfyun.cn/doc/spark/Embedding_api.html), now it is runnable.

### DIFF
--- a/libs/community/langchain_community/embeddings/sparkllm.py
+++ b/libs/community/langchain_community/embeddings/sparkllm.py
@@ -16,9 +16,6 @@ from langchain_core.pydantic_v1 import BaseModel, Field, SecretStr, root_validat
 from langchain_core.utils import convert_to_secret_str, get_from_dict_or_env
 from numpy import ndarray
 
-# Used for embedding
-EMBEDDING_API_URL: str = "https://emb-cn-huabei-1.xf-yun.com/"
-
 # SparkLLMTextEmbeddings is an embedding model provided by iFLYTEK Co., Ltd.. (https://iflytek.com/en/).
 
 # Official Website: https://www.xfyun.cn/doc/spark/Embedding_api.html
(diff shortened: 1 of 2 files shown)


maang-h:
@eyurtsev @baskaryan  cc

libs/community/langchain_community/embeddings/sparkllm.py
```diff
@@ -5,25 +5,23 @@
...
+from langchain_core.pydantic_v1 import BaseModel, Field, SecretStr, root_validator
 from langchain_core.utils import convert_to_secret_str, get_from_dict_or_env
 from numpy import ndarray
 
-# Used for document and knowledge embedding
-EMBEDDING_P_API_URL: str = "https://cn-huabei-1.xf-yun.com/v1/private/sa8a05c27"
-# Used for user questions embedding
-EMBEDDING_Q_API_URL: str = "https://cn-huabei-1.xf-yun.com/v1/private/s50d55a16"
+# Used for embedding
+EMBEDDING_API_URL: str = "https://emb-cn-huabei-1.xf-yun.com/"
```
baskaryan, line 20: should we make this configurable as well?

maang-h: Should I add 'base_url' as a class attribute?

maang-h: Done

baskaryan (approved):

PR #22151, opened 2024-05-24
core[minor]: Add an async root listener and with_alisteners method

nicolasnk:
- [x] **Adding AsyncRootListener**: "langchain_core: Adding AsyncRootListener"

- **Description:** Adding an AsyncBaseTracer, AsyncRootListener and `with_alistener` function. This is to enable binding async root listener to runnables. This currently only supported for sync listeners.
- **Issue:** None
- **Dependencies:** None

- [x] **Add tests and docs**: Added units tests and example snippet code within the function description of `with_alistener`

- [x] **Lint and test**: Run make format_diff, make lint_diff and make test

### DIFF
--- a/libs/core/langchain_core/tracers/abstract_base_tracer.py
+++ b/libs/core/langchain_core/tracers/abstract_base_tracer.py
@@ -1,0 +1,565 @@
+"""Utilities for the root listener."""
+
+from __future__ import annotations
+
+import logging
+import sys
+import traceback
+from abc import ABC, abstractmethod
+from datetime import datetime, timezone
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Coroutine,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Union,
+    cast,
+)
+from uuid import UUID
+
+from tenacity import RetryCallState
+
+from langchain_core.exceptions import TracerException
+from langchain_core.load import dumpd
+from langchain_core.messages import BaseMessage
+from langchain_core.outputs import (
+    ChatGeneration,
+    ChatGenerationChunk,
+    GenerationChunk,
+    LLMResult,
+)
+from langchain_core.tracers.schemas import Run
+
+if TYPE_CHECKING:
+    from langchain_core.documents import Document
+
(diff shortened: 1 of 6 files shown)


libs/core/langchain_core/tracers/abstract_base_tracer.py
```diff
@@ -0,0 +1,562 @@
...
+
+if TYPE_CHECKING:
+    from langchain_core.documents import Document
+
+logger = logging.getLogger(__name__)
+
+SCHEMA_FORMAT_TYPE = Literal["original", "streaming_events"]
+
+
+class AbstractBaseTracer(ABC):
```
eyurtsev, line 47: nit: base tracer is already abstract. I don't have a good suggestion here since the name is already occupied so probably fine.

Is this for internal use only if so, could we document it as private?

eyurtsev (commented):
Looks good overall. 

Consider combining `with_alisteners` and `with_listeners` to avoid bugs when folks implement custom runnables like `RunnableEach` otherwise default implementation is inherited

libs/core/langchain_core/tracers/abstract_base_tracer.py
nicolasnk: Correct. Ideally it would be BaseTracer. 
Documenting as private, do you mean adding a note in the class description ?

nicolasnk: Changed to _TracerCore

eyurtsev (approved):

libs/core/langchain_core/tracers/base.py
```diff
@@ -2,38 +2,26 @@
...
-    Tuple,
     Union,
-    cast,
 )
 from uuid import UUID
 
 from tenacity import RetryCallState
 
-from langchain_core.callbacks.base import BaseCallbackHandler
-from langchain_core.exceptions import TracerException
```
eyurtsev, line 28: Hmm keep this import here with a noqa. Hitting some de-serializatoin issue in the unit tests. I can address this in a separate PR later on?

nicolasnk: sounds good. Thanks

PR #22166, opened 2024-05-25
community: create_sql_agent Resolve BadRequestError for API requests with pre-filled assistant

HyunjunJeon:
**Description**: 

Fixes the issue where **API requests with a pre-filled assistant message in the final position were causing a BadRequestError with code 400**. The error message indicated that pre-filling the assistant response is not supported when using tools. The changes in this commit ensure that API requests no longer include an assistant message in the final position, allowing the API to handle the request correctly and prevent the invalid request error from occurring. With this fix, users can now make API requests smoothly without encountering the BadRequestError, improving the overall user experience and reliability of the API.

### DIFF
--- a/libs/community/langchain_community/agent_toolkits/sql/base.py
+++ b/libs/community/langchain_community/agent_toolkits/sql/base.py
@@ -13,7 +13,7 @@ from typing import (
     cast,
 )
 
-from langchain_core.messages import AIMessage, SystemMessage
+from langchain_core.messages import SystemMessage
 from langchain_core.prompts import BasePromptTemplate, PromptTemplate
 from langchain_core.prompts.chat import (
     ChatPromptTemplate,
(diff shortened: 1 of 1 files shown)


baskaryan:
which model is this causing an error for?

HyunjunJeon:
ChatAnthropic caused a problem.

hrshdhgd:
I made the suggested changes in the PR and it seems to be the solution to the same error I experienced in one of my projects.

hrshdhgd:
> I made the suggested changes in the PR and it seems to be the solution to the same error I experienced in one of my projects.

Alternatively, if I do not set the `agent_type` parameter , it works just fine.

gkulasik-modal:
+1 having this issue with ChatAnthropic

efriis:
Hey team! Could you file an issue with a reproducable example (ideally the direct call of ChatAnthropic)? I think this PR won't solve the root issue

Alternatively could add a unit test showing what this fixes

hwchase17:
this is ultimately a prompting issue, and our prompting strategy isn't compatible with ChatAnthropic. That being said, this isnt the best solution - the previous AI message is writen from the perspective of the assistant, so just putting it as a human message doesnt make sense

we are moving less and less towards built in prompts, so id recommend just passing in a prompt here

Issue #22175, opened 2024-05-25
HTMLSectionSplitter errors when passed a path to an xslt file


tClelford:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code:
```python
from langchain_text_splitters import HTMLSectionSplitter

some_html = "..."
xslt_path = "./this_exists.xslt"
headers_to_split_on = [("h1", "Header 1"), ("h2", "Header 2")]

html_splitter = HTMLSectionSplitter(headers_to_split_on=headers_to_split_on, xslt_path=xslt_path)
html_header_splits = html_splitter.split_text(some_html)
```

Or

```python
from langchain_text_splitters import HTMLSectionSplitter

some_html = "..."
xslt_path = "/path/to/this_exists.xslt"
headers_to_split_on = [("h1", "Header 1"), ("h2", "Header 2")]

html_splitter = HTMLSectionSplitter(headers_to_split_on=headers_to_split_on, xslt_path=xslt_path)
html_header_splits = html_splitter.split_text(some_html)
```

In both cases assuming `this_exists.xslt` is a valid xslt file that exists at the location passed.

### Error Message and Stack Trace (if applicable)

File src/lxml/parser.pxi:743, in lxml.etree._handleParseResult()

File src/lxml/parser.pxi:670, in lxml.etree._raiseParseError()

OSError: Error reading file '{app_dir}/.venv/lib/python3.12/site-packages/langchain_text_splitters/this_exists.xslt': failed to load external entity "{app_dir}/.venv/lib/python3.12/site-packages/langchain_text_splitters/this_exists.xslt"

### Description

There are a couple of bugs here:

1. if you pass a relative file path - `./this_exists.xslt` then `HTMLSectionSplitter` tries to turn it to an absolute path but uses the path to the langchain module (`{app_dir}/.venv/lib/python3.12/site-packages/langchain_text_splitters` in my case) rather than the path to the current working directory.  
[...]

PR #22176, opened 2024-05-25
text-splitters: fix HTMLSectionSplitter parsing of xslt paths

tClelford:
## Description
This PR allows passing the HTMLSectionSplitter paths to xslt files. It does so by fixing two trivial bugs with how passed paths were being handled. It also changes the default value of the param `xslt_path` to  `None` so the special case where the file was part of the langchain package could be handled.

## Issue
#22175


libs/text-splitters/tests/unit_tests/test_text_splitters.py
```diff
@@ -1619,6 +1619,37 @@ def test_happy_path_splitting_based_on_header_with_whitespace_chars() -> None:
...
+@pytest.mark.requires("lxml")
+@pytest.mark.requires("bs4")
+def test_section_splitter_accepts_a_relative_path() -> None:
+    html_string = """<html><body><p>Foo</p></body></html>"""
+    test_file = Path("tests/test_data/test_splitter.xslt")
+    assert test_file.is_file()
+
+    sec_splitter = HTMLSectionSplitter(
+        headers_to_split_on=[("h1", "Header 1"), ("h2", "Header 2")],
+        xslt_path=test_file.as_posix(),
```
baskaryan, line 1631: out of curiosity why do we need to add as_posix here and in 1647

tClelford: Hmmm, I guess we don't need to anymore. Initially, we were assigning either a string or a `Path` to `xslt_path`/`self.xslt_path` so I used `.as_posix()` to give us a cross-platform compatible string. Now we don't have the default arg as a string, it could always be an instance of `Path` and we can simplify our logic. I'll push up an improvement, thanks for the review :)

baskaryan (approved):

tClelford:
@baskaryan - Is there anything else I need to do to get this merged? I can't approve the workflows :/

Issue #19997, opened 2024-04-04
DOC: How to create a ChatModel from Huggingface using local llm


jnchen:
### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

`langchain_community.llms.huggingface_hub.HuggingFaceHub` was deprecated, and `langchain_community.chat_models.huggingface.ChatHuggingFace` only Works with HuggingFaceTextGenInference, HuggingFaceEndpoint, and HuggingFaceHub
So how to create a ChatHuggingFace instance from local huggingface Model Now??

source :https://api.python.langchain.com/en/latest/chat_models/langchain_community.chat_models.huggingface.ChatHuggingFace.html#langchain_community.chat_models.huggingface.ChatHuggingFace
https://api.python.langchain.com/en/latest/llms/langchain_community.llms.huggingface_hub.HuggingFaceHub.html#langchain_community.llms.huggingface_hub.HuggingFaceHub

### Idea or request for content:

document says HuggingFaceHub was deprecated, i should use the HuggingFaceEndPoint, but that is two difference integrate method.

DiaQusNet:
I am alse confused by this method, LangChain can't use huggingface chat model locally.
Did you find a way to use it?

jnchen:
> I am alse confused by this method, LangChain can't use huggingface chat model locally. Did you find a way to use it?

in 0.2.0, they still haven't fixed this problem

jnchen:
and then, the all huggingface llm related class was deprecated in 0.2.0, so how to run huggingface chat model locally in future

jnchen:
> I am alse confused by this method, LangChain can't use huggingface chat model locally. Did you find a way to use it?

there's a library called langchain-huggingface in https://github.com/langchain-ai/langchain/tree/master/libs/partners/huggingface, but can't find any docs about it.

michalgregor:
Is it deprecated because it now got moved to the separate library? That's what I would expect. 

But the chat in the library still only supports the API endpoints according to its docstring:

Works with `HuggingFaceTextGenInference`, `HuggingFaceEndpoint`,
    and `HuggingFaceHub` LLMs.

PR #22194, opened 2024-05-27
huggingface: Support for HuggingFacePipeline in ChatHuggingFace.

michalgregor:
- **Description:** Added support for using HuggingFacePipeline in ChatHuggingFace (previously it was only usable with API endpoints, probably by oversight).
- **Issue:** #19997 
- **Dependencies:** none
- **Twitter handle:** none


baskaryan (commented):
this is great @michalgregor! could we also add an example of instantiating via a pipeline here https://python.langchain.com/v0.2/docs/integrations/chat/huggingface/#1-instantiate-an-llm (in `docs/docs/integrations/chat/huggingface.ipynb`)

michalgregor:
That is a good idea. I added an example, including also a short excerpt of how one can pass in a quantization config. That should be helpful, I think.

baskaryan (approved):

PR #22250, opened 2024-05-28
docs[patch]: Add canonical URLs pointing at 0.2 docs for popular pages

jacoblee93:
@efriis


efriis (commented):
I think should also override `rel=alternate`. Right now they're pointing at different things - not sure if that's bad / a rule we shouldn't break?

view-source:https://langchain-git-jacob-canonicalurls-langchain.vercel.app/v0.1/docs/expression_language/get_started/

Issue #22256, opened 2024-05-29
MarkdownHeaderTextSplitter flattens Paragraphs separators into single line breaks


relston:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code to Reproduce
```python
from langchain.textsplitters import MarkdownHeaderTextSplitter

# Sample Markdown input
markdown_text = """
# My Heading

This is a paragraph with some detailed explanation.

This is another separate paragraph.
"""

# Initialize and apply the text splitter
splitter = MarkdownHeaderTextSplitter(headers_to_split_on=[('#', 'Header 1')])
result = splitter.split_text(markdown_text)
print(result)
```

### Expected Behavior
The expected behavior would be to keep the paragraph breaks as they are crucial for subsequent text manipulation tasks that may rely on the structure conveyed by separate paragraphs:
```python
[Document(page_content='This is a paragraph with some detailed explanation.\n\nThis is another separate paragraph.', metadata={'Header 1': 'My Heading'})]
```

### Actual Behavior
Currently, the text after being processed by `MarkdownHeaderTextSplitter` loses paragraph distinctions, flattening into line breaks:
```python
[Document(page_content='This is a paragraph with some detailed explanation.\nThis is another separate paragraph.', metadata={'Header 1': 'My Heading'})]
```

This issue affects not only readability but also the downstream processing capabilities that require structured and clearly delineated text for effective analysis and feature extraction.

### Description

The current implementation of `MarkdownHeaderTextSplitter` in LangChain notably [splits on text on `/n`](https://github.com/relston/langchain/blob/d268587cbadded8cf1a3d67546afafe5db0690c5/libs/text-splitters/langchain_text_splitters/markdown.py#L96) and [strips out white space](https://github.com/relston/langchain/blob/d268587cbadded8cf1a3d67546afafe5db0690c5/libs/text-splitters/langchain_text_splitters/markdown.py#L111) from each line when processing Markdown text. This removal of white spaces and paragraph separators (`\n\n`) directly impacts further text splitting and processing strategies, as it disrupts the natural paragraph structure integral to most textual analyses and transformations.

### Other Examples
[...]

Issue #20823, opened 2024-04-24
Markdown text splitter will remove code block indentation


pierre-yv:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
MarkdownHeaderTextSplitter(headers_to_split_on=headers_to_split_on)

doc = """
    ```yaml
    my: 
       foo: bar
       a: b
   ```
"""
md_header_splits = markdown_splitter.split_text(markdown_document)
```

expected should be 
```yaml
my: 
   foo: bar
   a: b
```

actual is 
```yaml
my: 
foo: bar
a: b
```

### Description

Here is a fix that works for whitespaces indents, not sure it covers the tabs though

markdown.py: split_text
```python
        in_code_block = False
        opening_fence = ""
        code_block_indent = -1

        for line in lines:
            stripped_line = line.strip()

            if not in_code_block:
                # Exclude inline code spans
                if stripped_line.startswith("```") and stripped_line.count("```") == 1:
                    in_code_block = True
                    opening_fence = "```"
                    code_block_indent = len(line) - len(stripped_line)
                elif stripped_line.startswith("~~~"):
                    in_code_block = True
                    opening_fence = "~~~"
                    code_block_indent = len(line) - len(stripped_line)
            else:
                if stripped_line.startswith(opening_fence):
                    in_code_block = False
                    opening_fence = ""

            if in_code_block:
                total_line_indent = len(line) - len(stripped_line)
                stripped_line = " "*max(0, total_line_indent - code_block_indent) + stripped_line
                current_content.append(stripped_line)
                continue
```

### System Info

- Installing typing-extensions (4.11.0)
  - Downgrading urllib3 (2.2.1 -> 1.26.18)
  - Installing exceptiongroup (1.2.1)
  - Installing frozenlist (1.4.1)
  - Installing h11 (0.14.0)
  - Installing jsonpointer (2.4)
  - Installing multidict (6.0.5)
  - Installing mypy-extensions (1.0.0)
  - Installing orjson (3.10.1)
  - Downgrading packaging (24.0 -> 23.2)
  - Installing pydantic (1.10.15)
  - Installing six (1.16.0)
[...]

Issue #19436, opened 2024-03-22
MarkdownHeaderTextSplitter removing format of nested lists / sublists


migcosta:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

``` python
from langchain_text_splitters import MarkdownHeaderTextSplitter
markdown_document = """
# Heading 1

## Heading 2

This is a Markdown List with a nested List:

-   Item 1

    -   Sub Item 1.1

-   Item 2

    -   Sub Item 2.1

    -   Sub Item 2.2

-   Item 3
"""

headers_to_split_on = [
    ("#", "Header 1"),
    ("##", "Header 2"),
    ("###", "Header 3"),
]

markdown_splitter = MarkdownHeaderTextSplitter(headers_to_split_on=headers_to_split_on)
md_header_splits = markdown_splitter.split_text(markdown_document)
print(md_header_splits[0].page_content)

```

**Expected output**
```markdown
This is a Markdown List with a nested List:

-   Item 1

    -   Sub Item 1.1

-   Item 2

    -   Sub Item 2.1

    -   Sub Item 2.2

-   Item 3
```

Actual Output
```markdown
This is a Markdown List with a nested List:
-   Item 1
-   Sub Item 1.1
-   Item 2
-   Sub Item 2.1
-   Sub Item 2.2
-   Item 3
```

### Description

**MarkdownHeaderTextSplitter.split_text()** is removing the format from the nested lists / sublists.
The issue is caused in this part of the [code](https://github.com/langchain-ai/langchain/blob/53ac1ebbbccbc16ce57badf08522c6e59256fdfe/libs/text-splitters/langchain_text_splitters/markdown.py#L109C13-L109C41). Stripping the line removes the Markdown sublist / nested list indentation ( [Markdown Lists Docs](https://www.markdownguide.org/basic-syntax/#lists-1) ).

The same issue is also expected on [paragraphs ](https://www.markdownguide.org/basic-syntax/#paragraphs) [blockquotes ](https://www.markdownguide.org/basic-syntax/#blockquotes) and so on.

### System Info

Package Information
-------------------
> langchain_core: 0.1.28
> langchain: 0.0.350
[...]

PR #22257, opened 2024-05-29
text-splitters: Introduce Experimental Markdown Syntax Splitter

relston:
#### Description
This MR defines a `ExperimentalMarkdownSyntaxTextSplitter` class. The main goal is to replicate the functionality of the original `MarkdownHeaderTextSplitter` which extracts the header stack as metadata but with one critical difference: it keeps the whitespace of the original text intact.

This draft reimplements the `MarkdownHeaderTextSplitter` with a very different algorithmic approach. Instead of marking up each line of the text individually and aggregating them back together into chunks, this method builds each chunk sequentially and applies the metadata to each chunk. This makes the implementation simpler. However, since it's designed to keep white space intact its not a full drop in replacement for the original.  Since it is a radical implementation change to the original code and I would like to get feedback to see if this is a worthwhile replacement, should be it's own class, or is not a good idea at all.

Note: I implemented the `return_each_line` parameter but I don't think it's a necessary feature. I'd prefer to remove it.

This implementation also adds the following additional features:
- Splits out code blocks and includes the language in the `"Code"` metadata key
- Splits text on the horizontal rule `---` as well
- The `headers_to_split_on` parameter is now optional - with sensible defaults that can be overridden.

#### Issue
Keeping the whitespace keeps the paragraphs structure and the formatting of the code blocks intact which allows the caller much more flexibility in how they want to further split the individuals sections of the resulting documents. This addresses the issues brought up by the community in the following issues:
- https://github.com/langchain-ai/langchain/issues/20823
- https://github.com/langchain-ai/langchain/issues/19436
- https://github.com/langchain-ai/langchain/issues/22256

#### Dependencies
N/A

#### Twitter handle
@RyanElston

### DIFF
--- a/libs/text-splitters/langchain_text_splitters/markdown.py
+++ b/libs/text-splitters/langchain_text_splitters/markdown.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import re
-from typing import Any, Dict, List, Tuple, TypedDict
+from typing import Any, Dict, List, Tuple, TypedDict, Union
 
 from langchain_core.documents import Document
 
@@ -223,13 +223,14 @@ class HeaderType(TypedDict):
     name: str
     data: str
 
+
 class ExperimentalMarkdownSyntaxTextSplitter:
     """
     An experimental text splitter for handling Markdown syntax.
 
-    This splitter aims to retain the exact whitespace of the original text while 
-    extracting structured metadata, such as headers. It is a re-implementation of the 
-    MarkdownHeaderTextSplitter with notable changes to the approach and 
+    This splitter aims to retain the exact whitespace of the original text while
+    extracting structured metadata, such as headers. It is a re-implementation of the
+    MarkdownHeaderTextSplitter with notable changes to the approach and
     additional features.
 
     Key Features:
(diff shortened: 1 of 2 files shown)


libs/text-splitters/langchain_text_splitters/markdown.py
```diff
@@ -221,3 +222,119 @@ class HeaderType(TypedDict):
     level: int
     name: str
     data: str
+
+
+class ExperimentalMarkdownSyntaxTextSplitter:
```
baskaryan, line 227: could we add some docs on the behavior of this textsplitter, especially as opposed to existing mardkown one

relston: Sure thing! Added some documentation, let me know what you think 😄

relston: Fixed linting errors

isahers1:
@baskaryan I think I fixed all lint error, and code lgtm lmk if I can merge

relston:
Thanks a bunch @isahers1!

relston:
@baskaryan let me know if there is any other feedback

luc42ei:
great, 100% was looking for this! thanks!

note: in the docstring you wrote "splitter.split(text)" using a `split()` method that doesn't exist

relston:
> great, 100% was looking for this! thanks!
> 
> note: in the docstring you wrote "splitter.split(text)" using a `split()` method that doesn't exist

Ah, thanks! Happy to do a follow up. Let me know if you run into any bugs or issues with it @luc42ei

PR #22274, opened 2024-05-29
Adding Managed Identity for Azure AI Search 

marlenezw:
This PR adds managed identity functionality to Azure AI Search and the Search Retriever. Keyless authentication is part of our efforts to help make AI development more secure. More on managed identity with Search [here](https://learn.microsoft.com/en-us/azure/search/search-howto-managed-identities-data-sources?tabs=portal-sys%2Cportal-user). This PR also updates the retriever notebook to show an example of using keyless authentication with the retriever. 

Twitter: @marlene_zw

Working on adding tests.


marlenezw:
Just a heads up on this @mattgotteiner and @farzad528. This adds MI to search so it works with default credential. I need to add tests and will fix linting issues.

farzad528:
@chuwik fyi

marlenezw:
Marking this ready for review, checks are passing for me and I've added two new integration tests as well. @ccurme if you have time please take a look!

libs/community/langchain_community/retrievers/azure_ai_search.py
```diff
@@ -30,6 +30,8 @@ class AzureAISearchRetriever(BaseRetriever):
     recommended to use a Query key."""
     api_version: str = "2023-11-01"
     """API version"""
+    azure_ad_token_provider: Optional[Callable] = None
+    """Your Azure credential."""
```
pamelafox, line 34:
```suggestion
    """A function that returns an Azure OpenAI authentication token."""
```

A bit more accurate? Based on https://github.com/openai/openai-python/blob/54a5911f5215148a0bdeb10e2bcfb84f635a75b9/src/openai/lib/azure.py#L157C38-L157C127

libs/community/langchain_community/retrievers/azure_ai_search.py
```diff
@@ -52,9 +54,14 @@ def validate_environment(cls, values: Dict) -> Dict:
...
         )
-        values["api_key"] = get_from_dict_or_env(
-            values, "api_key", "AZURE_AI_SEARCH_API_KEY"
-        )
+        try:
+            values["api_key"] = get_from_dict_or_env(
+                values, "api_key", "AZURE_AI_SEARCH_API_KEY"
+            )
+        except ValueError:
+            values["azure_ad_token_provider"] = get_from_dict_or_env(
```
pamelafox, line 62: How did you decide the order of things to check? i.e. how did you decide to check api_key first versus azure_ad_token_provider?

libs/community/langchain_community/retrievers/azure_ai_search.py
```diff
@@ -79,10 +86,17 @@ def _build_search_url(self, query: str) -> str:
 
     @property
     def _headers(self) -> Dict[str, str]:
-        return {
-            "Content-Type": "application/json",
-            "api-key": self.api_key,
-        }
+        if self.azure_ad_token_provider:
+            token = "Bearer " + f"{self.azure_ad_token_provider()}"
```
pamelafox, line 90:
```suggestion
            token = f"Bearer {self.azure_ad_token_provider()}"
```

The whole thing can be an f-string, no?

libs/community/langchain_community/retrievers/azure_ai_search.py
```diff
@@ -79,10 +86,17 @@ def _build_search_url(self, query: str) -> str:
...
+            return {
+                "Content-Type": "application/json",
```
pamelafox, line 92: To avoid repeating the Content-Type, you could do something like:

headers = {
  "Content-Type": "application/json"
}

if self.azure_ad_token_provider:
   headers["authorization"] = ...
else:
   headers["api-key"] = ...
return headers


Kind of nice especially if headers increase.

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -72,8 +72,9 @@
 
 def _get_search_client(
     endpoint: str,
-    key: str,
+    key: Optional[str],
```
pamelafox, line 75: Does mypy let you have an Optional before a non-Optional? I would think that it'd need a default in that case or something.

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -72,8 +72,9 @@
...
     index_name: str,
+    azure_credential: Optional[Callable] = None,
```
pamelafox, line 77: Is this an azure_credential or is it a token provider?

pamelafox: Ah nevermind, it looks like it truly is a credential.

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -236,14 +240,15 @@ def fmt_err(x: str) -> str:
...
-    """`Azure Cognitive Search` vector store."""
+    """`Azure AI Search` vector store."""
 
     def __init__(
         self,
         azure_search_endpoint: str,
-        azure_search_key: str,
-        index_name: str,
         embedding_function: Union[Callable, Embeddings],
+        azure_search_key: str = "",
```
pamelafox, line 249: If the parameter order changed, that'd be a backwards compatible issue for library users, right?

libs/community/tests/integration_tests/retrievers/test_azure_ai_search.py
```diff
@@ -68,3 +68,32 @@ async def test_azure_cognitive_search_ainvoke() -> None:
     for doc in documents:
         assert isinstance(doc, Document)
         assert doc.page_content
+
+
+def test_azure_ai_search_invoke_managed_identity() -> None:
+    """Test valid call to Azure AI Search using managed identity.
+
+    In order to run this test, you should provide
+    `service_name`,index_name`, and azure_ad_token_provider,
```
pamelafox, line 77: Inconsistent use of backticks

libs/community/tests/integration_tests/retrievers/test_azure_ai_search.py
```diff
@@ -68,3 +68,32 @@ async def test_azure_cognitive_search_ainvoke() -> None:
     for doc in documents:
         assert isinstance(doc, Document)
         assert doc.page_content
+
+
+def test_azure_ai_search_invoke_managed_identity() -> None:
```
pamelafox, line 73: So all of the tests are actually testing with real credentials, versus using mocks?

libs/community/langchain_community/retrievers/azure_ai_search.py
marlenezw: I didn't think too much about this decision. My assumption was that most people will use keys compared to MI. Wasn't sure as well how to handle the checking for the api_key env variable without try/except. Open to other suggestions!

libs/community/langchain_community/retrievers/azure_ai_search.py
marlenezw: Sure! Will update

libs/community/tests/integration_tests/retrievers/test_azure_ai_search.py
marlenezw: Yes, the person running the tests has to add in their own. I'm following what the rest of the tests went with

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: It does, but I could add an empty string as default here? I think it doesn't let you pass something with a default in before non-default.

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: It would be. My problem was getting `azure_search_key: Optional[str]` to work here. I'd get an error unless I explicitly added the default. My ideal would be to not have to. Do you have any thoughts on how to fix this? I might try refactor in a bit

libs/community/tests/integration_tests/retrievers/test_azure_ai_search.py
marlenezw: Back ticks added

libs/community/langchain_community/retrievers/azure_ai_search.py
marlenezw: Done!

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: Let me actually change this into an if else statement and change the order to first check for the token and the else out to api key.

libs/community/langchain_community/retrievers/azure_ai_search.py
marlenezw: Let me actually change this into an if else statement and change the order to first check for the token and the else out to api key. This is consistent with what we're doing later, so makes sense to change the order.

marlenezw: @pamelafox changed this code to use if/else and check to see if the token provider is given in the values dictionary first. I think this works because it's rare for people to set the token provider as an env variable.

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: @pamelafox adding a lambda as the default for embeddings and then maintaining the order of parameters worked 😄 Mypy is happy and the code should be backwards compatible now!

marlenezw:
cc @baskaryan in case you have a second to look!

marlenezw:
@ccurme I realise this is quite a large PR. Let me know if you'd prefer for me to break it into two smaller ones. I could have one for the retriever and one for the vector store.

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -236,14 +240,15 @@ def fmt_err(x: str) -> str:
...
-    """`Azure Cognitive Search` vector store."""
+    """`Azure AI Search` vector store."""
 
     def __init__(
         self,
-        azure_search_endpoint: str,
-        azure_search_key: str,
-        index_name: str,
-        embedding_function: Union[Callable, Embeddings],
+        azure_search_endpoint: str = "",
```
ccurme, line 247: are these no longer required?

if so, I'd suggest adding a docstring that explains what parameters (or combinations of parameters) must be passed in, and validating that early in \_\_init\_\_.

IMO it's more clear to have something like
```python
embedding_function: Union[Callable, Embeddings, None] = None,
```
and then raise an error if it's None. ideally if they are required we keep them positional though.

libs/community/tests/integration_tests/vectorstores/test_azuresearch.py
```diff
@@ -84,3 +85,35 @@ def test_semantic_hybrid_search() -> None:
     time.sleep(1)
     res = vector_store.semantic_hybrid_search(query="What's Azure Search?", k=3)
     assert len(res) == 3
+
+
+def test_similarity_search_test_managed_identity() -> None:
```
ccurme, line 90: Can you confirm you've run existing integration tests? (CI doesn't run them.)

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -110,11 +111,14 @@ def _get_search_client(
...
+    if azure_credential:
+        credential = azure_credential
+    elif key is None:
         credential = DefaultAzureCredential()
     elif key.upper() == "INTERACTIVE":
-        credential = InteractiveBrowserCredential()
-        credential.get_token("https://search.azure.com/.default")
+        credential = InteractiveBrowserCredential().get_token(
+            "https://search.azure.com/.default"
+        )
```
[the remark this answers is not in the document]
marlenezw: Yes, at least to me it looked like a bug since credential wasn't actually set. I can run the tests to double check and confirm though

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: azure_search_endpoint, index_name and the embedding_function are all still required. We need to make the azure_search_key optional, since with managed identity the user passes in a token instead. 

We could leave azure_search_endpoint as positional, but my main concern was that if I left index_name  and embedding_function as positional then the order of the positional arguments would change because index_name would need to come before azure_search_key, which could break things for current users. Making things optional means that we can make this change backwards compatible. Not sure how else to do this without breaking things for users. What do you think @ccurme?

libs/community/tests/integration_tests/vectorstores/test_azuresearch.py
marlenezw: Yes, I ran the existing tests with both Managed Identity (using the token) and API keys and they were passing!

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: I think for this I was trying to weigh the what makes more sense, keeping required arguments positional or prioritising maintaining the order of the arguments for the sake of backwards compatibility. Open to thoughts on what is preferred, or other ideas in case I'm missing something.

ccurme: Yeah that makes sense. I'd prioritize backwards compatibility here. My first thought is:

- `azure_search_endpoint` stays positional (why not)
- `azure_search_key`, `index_name`, `embedding_function` default to None similar to `azure_credential`
- We add a short function to validate parameters and call it early in `__init__`
  - tests that either `azure_search_key` or `azure_credential` are provided
  - tests that `index_name` and `embedding_function` are provided
- add a docstring explaining required arguments

Let me know what you think.

libs/community/langchain_community/vectorstores/azuresearch.py
```diff
@@ -236,14 +240,15 @@ def fmt_err(x: str) -> str:
...
         self,
-        azure_search_endpoint: str,
-        azure_search_key: str,
-        index_name: str,
-        embedding_function: Union[Callable, Embeddings],
+        azure_search_endpoint: str = "",
+        azure_search_key: str = "",
+        index_name: str = "",
+        embedding_function: Union[Callable, Embeddings] = lambda x: x,
+        azure_credential: Optional[Callable] = None,
```
ccurme, line 251: I'd put this down below `*,` in required keyword arguments in case users are passing these in positionally.

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: This makes sense to me!!! I'll start working on these changes!

libs/community/langchain_community/vectorstores/azuresearch.py
marlenezw: Sounds good! I'll make sure that we can test things are working!

docs/docs/integrations/retrievers/azure_ai_search.ipynb
baskaryan, line 1: doc has been updated since pr was opened, need to merge existing and whats in the pr

johnayoub-wtw:
@marlenezw @ccurme any possibility to extend this to `AzureOpenAIEmbeddings` as well? maybe in a separate PR.

ccurme:
@marlenezw should we keep iterating on this?

I think it'd be great to get Azure AI Search into `langchain-azure` as well, might make sense to port there.

PR #22286, opened 2024-05-29
langchain: fix: example in system prompt for KG extraction was wrong

thiswillbeyourgithub:
- **Description:** The example given in the system prompt to extract knowledge triplets is wrong. The sentence is not "Descartes plays the mandolin" but "Descartes likes to play the mandolin". I think this seemingly minor issue can have dramatic consequences for KG generation.

### DIFF
--- a/libs/langchain/langchain/indexes/prompts/knowledge_triplet_extraction.py
+++ b/libs/langchain/langchain/indexes/prompts/knowledge_triplet_extraction.py
@@ -25,7 +25,7 @@ _DEFAULT_KNOWLEDGE_TRIPLE_EXTRACTION_TEMPLATE = (
     "END OF EXAMPLE\n\n"
     "EXAMPLE\n"
     "Oh huh. I know Descartes likes to drive antique scooters and play the mandolin.\n"
-    f"Output: (Descartes, likes to drive, antique scooters){KG_TRIPLE_DELIMITER}(Descartes, plays, mandolin)\n"
+    f"Output: (Descartes, likes to drive, antique scooters){KG_TRIPLE_DELIMITER}(Descartes, likes to play, the mandolin)\n"
     "END OF EXAMPLE\n\n"
     "EXAMPLE\n"
     "{text}"


ccurme:
Thank you for this. We're moving away from optimizing these prompts because it can lead to silent changes in behavior for users. Closing the PR but let me know if you think this is safe to do (or if it's a particularly egregious case) and can re-open.

thiswillbeyourgithub:
Thanks for the quick response. I gave it some more thoughts and, while I understand and agree why it would be a bad idea to "optimize these prompts" regularly, I now think this is actually just a fix and not an optimization.

If my PR is accepted, I would be reallly surprised if users found the KG extractor misbehave, on the other hand with the current prompt I can totally imagine users behind disappointed by KG extraction quality but for the wrong reason, and probably attribute it to a particular LLM's weakness instead of a silly prompt mistake.

Globally, this system prompt is *fine* but regarding the 3 examples it uses:
* the 2nd example is just an example of what to ignore (and I do disagree with that example but as I said I'm fine with not optimizing)
* the 3rd example with my fix has the additional property of showing how to parse a multi clause sentence correctly (it takes the "likes" from the first clause and the rest from the last). The current state is to incorrectly parse multi clause sentence, which I can imagine snow balls into worst KG quality down the chain.
* The 1st example about nevada, after rereading it seems actually wrong too! In `It's also the number 1 producer of gold in the US` the predicate is not `is the number 1 producer of` but `is the number 1 US producer of`. It could be argued that given that we talk about Nevada it could be seen as implicit that we are talking about US states but to me KG extraction only makes sense if it's explicit! Given that KG extraction removes a triplet from its context, being implicit rather than explicit seems worst all around to me! And again, this to me looks like a fix and not an optimization. To illustrate this point, imagine this prompt roaming wikipedia and then comparing the subject that have as predicate "is the number 1 producer of" and have as subjets US states, whole countries, continents, human body organs and companies. Give that the previous triplet shows how to turn "it's a state in the US" as talking about "Nevada", it seems weird to then tell the LLM to be that unspecific for the other triplet.

In conclusion:

>  but let me know if you think this is safe to do (or if it's a particularly egregious case) and can re-open.

I do think it's safe to do.
I do think it's a warranted fix, borderline egregious.

But I'm open for debate.

Thank for your attention.

thiswillbeyourgithub:
> Closing the PR but let me know if you think this is safe to do (or if it's a particularly egregious case) and can re-open.

Hi @ccurme , as you said you were fine with reopening this issue and didn't reply to my last message nor reopened, I'm thinking you might have forgotten so here's a reminder.

PR #22295, opened 2024-05-29
community: remove redundant default arguments in HuggingFaceEmbeddings class

Soratz:
- **Description:** Two members of the class HuggingFaceEmbeddings, model_kwargs and encode_kwargs, encompasses arguments of cache_folder and show_progress_bar. These arguments can be directly put in kwargs dicts and there is no need to pass them as arguments. Doing so may result in a TypeError exception if the same keyword argument is inside the dictionary. 
    The default value of cache_folder is backward compatible. To ensure backward compatibility with show_progress_bar keyword argument, the key-value pair is added to kwargs dict if it doesn't exist in the first place.
- **Issue:** 
- **Dependencies:** None
- **Twitter handle:**


Soratz:
Hey @baskaryan , is it possible to review this?

baskaryan (commented):
hi @Soratz, thanks for contribution! this implementation has actually been deprecated in favor of the one in libs/partners/huggingface/embeddings (updating deprecation decorators to note this https://github.com/langchain-ai/langchain/pull/22444). could we update that implementation?

Soratz:
Thanks for the notice! I see that your implementation is on master now. I also rebased my branch.

efriis:
as mentioned above this should be against the partner package in libs/partners/huggingface!

Closing, and feel free to reopen there!

PR #22318, opened 2024-05-30
[Core] Update Tracing Interops

hinthornw:
LangSmith and LangChain context var handling evolved in parallel since originally we didn't expect people to want to interweave the decorator and langchain code.

Once we get a new langsmith release, this PR will let you seemlessly hand off between @traceable context and runnable config context so you can arbitrarily nest code.

It's expected that this fails right now until we get another release of the SDK


libs/core/tests/unit_tests/runnables/test_tracing_interops.py
```diff
@@ -0,0 +1,165 @@
...
+    return posts
+
+
+def test_config_traceable_handoff() -> None:
+    mock_session = MagicMock()
+    mock_client_ = Client(session=mock_session, api_key="test")
+    tracer = LangChainTracer(client=mock_client_)
+
+    @traceable
+    def my_great_great_grandchild_function(a: int) -> int:
```
nfcampos, line 32: lol

nfcampos (approved):

libs/core/tests/unit_tests/runnables/test_tracing_interops.py
hinthornw: image

PR #22321, opened 2024-05-30
Docs: Ollama (LLM, Chat Model & Text Embedding)

seanmavley:
- [x] Docs Update: Ollama
  - llm/ollama 
      - Switched to using llama3 as model with reference to templating and prompting
      - Added concurrency notes
  - chat_models/ollama
      - Added concurrency notes
  - text_embedding/ollama
     - include example for specific embedding models from Ollama


ccurme (approved):
Thank you!

Issue #10714, opened 2023-09-17
Custom Tool Return Direct BUT only if it doesnt raise a Tool Exception


Hsgngr:
### System Info

0.0.287 langchain

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [X] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [X] Agents / Agent Executors
- [X] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Currently I have a Custom Tool like this:
```
def _handle_error(error: ToolException) -> str:
    return (
        "The following errors occurred during tool execution:"
        + error.args[0]
        + "please try another tool."
    )
    tools = [
      Tool.from_function(
             name="GetCurrentWeather",
             func=get_current_weather,
             description="useful for getting the current weather for a location",
             return_direct=True,
             handle_tool_error=_handle_error)]
```
It works unless I add the return_direct=True.

When I do return_direct= True instead of trying other tools it just returns this:

```
> Entering new AgentExecutor chain...
16 AgentAction --> Thought: The question is asking about the current weather in Uludağ.
17 Action: GetCurrentWeather
18 Action Input: Uludağ
19 Thought: The question is asking about the current weather in Uludağ.
20 Action: GetCurrentWeather
21 Action Input: Uludağ
22 Observation:The following errors occurred during tool execution:the location is not found. please try another tool.
23 > Finished chain.
24 The following errors occurred during tool execution:location is not found.please try another tool.
```

### Expected behavior

Let's see the same thing without the `return_direct=True`:
```
> Entering new AgentExecutor chain...
Thought: The question is asking about the current weather in Uludağ.
Action: GetCurrentWeather
Action Input: Uludağ

Observation:The following errors occurred during tool execution:location is not found. please try another tool.
I couldn't find the current weather for Uludağ. I should try another tool.
Action: DuckDuckSearch
Action Input: "Uludağ current weather"

[...]

PR #22327, opened 2024-05-30
langchain: support forced termination on single action agents

pooriaPoorsarvi:
### Description 
Being able to force the chain to stop is a very important feature. It helps with the flow of the chain, controls costs, and potentially controls what can and can not go into the chain for privacy. 
Right now, two solutions exist, each with its own trade-offs, as discussed on https://github.com/langchain-ai/langchain/discussions/22149#discussioncomment-9608308. This MR tries to give chain designers a third option: explicitly forcing a chain to return the AgentFinish action on the next observation they see. 

This is done through the following steps:

1. An agent executor (for single action agents), will have an endpoint that will keep track of run_ids that have been forced to terminate
2. A tool can potentially be given access to both the run manager and the agent executing it 
3. The tool can have explicit code deciding when to terminate and when to return a result. Said result can now be a normal result or an error string that would be used by the agent to try and fix the issue that caused the error (no trade-off with error strings and return_direct any more).

### Issues
This MR will try to address problems seen on https://github.com/langchain-ai/langchain/discussions/22149  and https://github.com/langchain-ai/langchain/issues/10714 .

### Sample code
Here is a sample dummy code for using this feature. This is a dummy example that forces the chain to stop for whatever reason the programmer sees fit (e.g., if they don't want other tools to be called based on an arbitrary flag). It can be improved if tools and agents are defined in a class (to prevent the use of things like "global").

```
from langchain.pydantic_v1 import BaseModel, Field
from langchain.tools import BaseTool



from typing import Optional, Type

from langchain.callbacks.manager import (
    AsyncCallbackManagerForToolRun,
    CallbackManagerForToolRun,
)

class PowerInput(BaseModel):
    base: float = Field(description="base")
    exponent: float = Field(description="exponent")

    

class CustomPowerTool(BaseTool):
    name = "Power"
    description = "useful for when you need to do power operations"
    args_schema: Type[BaseModel] = PowerInput

    def _run(
        self, base: float, exponent: float, run_manager: Optional[CallbackManagerForToolRun] = None
    ) -> str:
        """Use the tool."""
        return base ** exponent

    async def _arun(
        self,
        base: float, exponent: float,
        run_manager: Optional[AsyncCallbackManagerForToolRun] = None,
    ) -> str:
        """Use the tool asynchronously."""
        raise NotImplementedError("Multiplier does not support async")

class MultiplierInput(BaseModel):
    a: float = Field(description="first number")
    b: float = Field(description="second number")

    
should_stop_flag = False
class CustomMultiplierTool(BaseTool):
    name = "Multiplier"
    description = "useful for when you need to answer questions about multiplication"
    args_schema: Type[BaseModel] = MultiplierInput

    def _run(
        self, a: float, b: float, run_manager: Optional[CallbackManagerForToolRun] = None
    ) -> str:
        """Use the tool."""
        global agent_executor
        global should_stop_flag
        if should_stop_flag:
            print("Forcing termination")
            agent_executor.terminate(run_manager=run_manager)
        return a * b

    async def _arun(
        self,
        a: float,
        b: float,
        run_manager: Optional[AsyncCallbackManagerForToolRun] = None,
    ) -> str:
        """Use the tool asynchronously."""
        raise NotImplementedError("Multiplier does not support async")
    

llm = ChatOpenAI(model="gpt-4o", temperature=0)

from langchain import hub

prompt = hub.pull("hwchase17/structured-chat-agent")

from langchain.agents import create_structured_chat_agent

tools = [
    CustomMultiplierTool(), 
    CustomPowerTool()
]

agent = create_structured_chat_agent(llm, tools, prompt)
from langchain.agents import AgentExecutor

agent_executor.invoke({"input": "multiply 3 by 5 and then take the result and raise it to the 4th power."})


```


hwchase17:
i would prefer to add support for this into the langgraph version of agent executor (https://python.langchain.com/v0.2/docs/how_to/migrate_agent/).

reason: this is a non-small feature and the langgraph version will be the main supported version going forward, so it seems better to me to invest more there

pooriaPoorsarvi:
@hwchase17 I'm a bit more new to LangGraph compared to LangChain, but that makes sense to me, however I had some questions:

1. Would we support this on both and have feature parity? (i.e., create a sister MR on the LangGraph repo)
2. From my understanding, create_react_agent (and the older deprecated create_agent_executor) both create a CompiledStateGraph, and not any class that is only dedicated to agents. Since the final object needs to have an endpoint that can be called by the developer we need to either: Add the endpoint to CompiledStateGraph, or have a class that extends CompiledStateGraph that has the ability to terminate. Since CompiledStateGraph is probably used by more things than just the agents, separating this would make sense, but I'm very new LangGraph, so I would really appreciate your insight into this. 
3. Also after the endpoint is called, do we want to handle the actual termination (i.e. exactly what piece of code is responsible for stopping) on the CompiledStateGraph (or its extension) or should we just use the endpoint to setup the correct flag for inner functions of create_react_agent like "should_continue"?

pooriaPoorsarvi:
@hwchase17  After reading up on LangGraph a bit, I think I have another potential solution that could easily be supported on both LangChain agent executors and compiled state graphs that are in LangGraph.



1. Add an additional callback that can be called to retrieve whether or not the chain should continue (defaults to True).
2. Use said call-back in def _get_tool_return for LangChain agents to handle whether or not the agent should stop
3. Use said call-back on should_continue's conditional edge to check whether or not the graph should continue

This would be on top of the call-back system, which is part of langchain_core. Thus, the routine to use it would be the same for both langchain and langgraph. 


If that's something that we can do, please let me know so that I close this MR and make two MRs with this system for both langchain and langgraph.

pooriaPoorsarvi:
will mention https://github.com/langchain-ai/langchain/issues/22441 here for reference on other tickets that need a solution for a proper way to terminate the agent

ccurme:
My reading of https://github.com/langchain-ai/langchain/issues/22441 suggests that the solution we're adopting is to implement via langgraph. Let me know if I've misunderstood and can re-open.

PR #22331, opened 2024-05-30
core[perf]: optimize chatprompttemplate `from_messages` constructor

marcoscannabrava:
**Description:** 
Runs loop over messages only once instead of twice when instantiating `ChatPromptTemplate` using the `from_messages` constructor.

**Twitter handle:**
@cannabrav

### DIFF
--- a/libs/core/langchain_core/prompts/chat.py
+++ b/libs/core/langchain_core/prompts/chat.py
@@ -966,14 +966,13 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
         Returns:
             a chat prompt template
         """
-        _messages = [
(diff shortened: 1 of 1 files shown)


marcoscannabrava:
Hi langchain team! @baskaryan @efriis. I'm hoping this is an easy review and yes/no change. Thanks!

ccurme:
Hi, thanks for this. It's not clear to me that this represents a substantial improvement in efficiency or readability. Closing for now but happy to re-open if needed.

marcoscannabrava:
Thanks, I did run into a problem where this caused high memory usage though. The patch reduced the memory usage in half.

PR #22336, opened 2024-05-30
docs: script `Component lists` generation

leo-gan:
**Issue:** 
The current API Reference is build on the "package" perspective navigation. Each package is represented with a separate API Reference. 
The API Ref does not have the "component" perspective navigation. For example, how do you get a list of all embeddings, LLMs, tools, etc.?

With more and more components extracted in separate packages, the API Docs is also splitting into unrelated pages. As a result, it is harder and harder to overview all components from **all packages** in one go.

**Idea:** Collect classes from all existing LC and partner packages; create the lists of classes by "component" (by the component base class inheritance. Say, by the `langchain_core.embeddings.Embeddings` class for the `embedding_model` components.). Use these lists for the "component_perspective" navigation.

**PR:** created the index page generation script. It creates index pages with a table with existing component implementation classes including links to the API Reference class pages. See [result page here](https://langchain-18a4lhmdp-langchain.vercel.app/v0.2/docs/integrations/text_embedding/).
It differs from the API Reference, because it lists all component classes in one table. It does not include any helper classes/functions. But it includes classes from **all packages** in one table!

**UPD:** saved all found classes in a Yaml file. This file can be used for many helpful things. :)


efriis (commented):
Hey @leo-gan ! This feels difficult to maintain.

Do you know if sphinx has any features that would let us show the list of "implementation classes" of `BaseCallbackHandler` e.g. here: https://api.python.langchain.com/en/latest/callbacks/langchain_core.callbacks.base.BaseCallbackHandler.html

Given this is just a list of things in the api ref, I would prefer to have the list in the API ref

leo-gan:
> Hey @leo-gan ! This feels difficult to maintain.
The pages are generated in fully automated fashion. The generation script is not included in `make docs_build` for ^ reason. I can add it, if you want, but you can do it better. I'm not sure I understand, what is "difficult to maintain" exactly.
 
> Do you know if sphinx has any features that would let us show the list of "implementation classes" of `BaseCallbackHandler` e.g. here: https://api.python.langchain.com/en/latest/callbacks/langchain_core.callbacks.base.BaseCallbackHandler.html

I've found [this](https://matplotlib.org/3.5.3/api/axis_api.html) inheritance diagram. But it doesn't work with many packages. I didn't how to config it.

> Given this is just a list of things in the api ref, I would prefer to have the list in the API ref

Yes, it makes sense. But we need a page with a list of all component, then we could link each component to the "derived classes" table/page. We don't have it now. That's why I included the tables in the `integrations/Components/<component>/index` pages.

isahers1 (commented):
@leo-gan I think getting better tables is a good idea, and I appreciate you helping out. There are a few questions I have.

1. Can you put your script in docs/scripts and follow the way model_feat_table.py is used?
2. Can we add more info than just the class? - see the chat models table for an example of a table that has more columns.

leo-gan:
> @leo-gan I think getting better tables is a good idea, and I appreciate you helping out. There are a few questions I have.
> 
> 1. Can you put your script in docs/scripts and follow the way model_feat_table.py is used?
> 2. Can we add more info than just the class? - see the chat models table for an example of a table that has more columns.

@isahers1 
1. Do you mean to add it to `makefile` like this line?
`$(PYTHON) scripts/document_loader_feat_table.py $(INTERMEDIATE_DIR)`
I didn't do this because running this script in each doc build is not obvious. 
2. I can do this but not in this PR. It requires a big refactoring and big "rethinking" :) Reasons for that: a) Chat Model table is build manually (all rows are hard-coded). The whole idea of this PR is the fully-automated table building. b) Chat Model table columns work only for the chat model component. Other components require different features/columns.
So, I'd prefer to finish this PR as is (it is still helpful), then follow with your proposal (#2)

leo-gan:
> 2. Can we add more info than just the class? - see the chat models table for an example of a table that has more columns.

@isahers1 BTW [Current Chat chat model](https://python.langchain.com/v0.2/docs/integrations/chat/) table holds **16** classes.
This PR code extracts **84** chat model classes. This fact push us against hard-coding this kind of tables.

isahers1:
@leo-gan I think we are going to hold off on this PR for now as we plan better ways to display the index pages for integrations.

leo-gan:
> @leo-gan I think we are going to hold off on this PR for now as we plan better ways to display the index pages for integrations.

@isahers1  Yeah, I'm thinking about a special menu for these tables in the API Reference. 

We have several major dimensions of the API Reference:
1. `Package dimension` (which is not in a good condition right now)
   - major LanChain packages: LangChain, Core, Community, TextSplitters (it is questionable for this item, to be in the main menu.
   -  partner packages.
2. each package presented by the 2nd level namespaces

Historically, the 2nd level namespaces in API Ref give us the `Component dimension`. But not now. 
In Docs, we have the `Component dimension` [here](https://python.langchain.com/v0.2/docs/integrations/components/). 

**The whole top-level navigation in the API Ref is broken now with the last changes with splitting LangChain into many packages.** That's the fact.

I'd change this in two points:
1. `Partner packages`: It could be more navigable if `Partners` opens a page where left nav presents a list of packages in one page. It kind of this [page](https://api.python.langchain.com/en/latest/langchain_api_reference.html), but left nav holds list of packages.
2. `Components`: It kind of this [page](https://api.python.langchain.com/en/latest/langchain_api_reference.html), but left nav holds list of individual component. Individual component page is the Component index (generated by this PR script)

CC @baskaryan

PR #22337, opened 2024-05-31
openai[patch]: refactor embeddings, add tests

baskaryan:


libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -586,3 +543,73 @@ async def aembed_query(self, text: str) -> List[float]:
...
+    ) -> List[List[float]]:
+        chunk_size = chunk_size or self.chunk_size
+        responses = await asyncio.gather(
+            *(
+                self.async_client.create(
+                    input=input[start : start + chunk_size], **self._invocation_params
+                )
+                for start in range(0, len(input), chunk_size)
+            )
+        )
```
baskaryan, line 590: @nfcampos does this look right

libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -518,24 +485,19 @@ def embed_documents(
...
+            return self._get_embeddings(texts, chunk_size=chunk_size)
 
-        # NOTE: to keep things simple, we assume the list may contain texts longer
-        #       than the maximum context and use length-safe embedding function.
-        engine = cast(str, self.deployment)
-        return self._get_len_safe_embeddings(texts, engine=engine)
+        num_texts = len(texts)
+        split_tokens, embedding_to_text_indices = self._tokenize_and_split(texts)
+        if len(split_tokens) == num_texts:
+            return self._get_embeddings(texts, chunk_size=chunk_size)
```
baskaryan, line 493: skip len safe embeddings if they're not needed

libs/partners/openai/langchain_openai/embeddings/base.py
nfcampos: This does all requests in parallel without a limit, not sure if that's intended?

libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -307,106 +285,106 @@ def _invocation_params(self) -> Dict[str, Any]:
...
-    # please refer to
+        tokenizer = AutoTokenizer.from_pretrained(
+            pretrained_model_name_or_path=model_name
+        )
+        for i, text in enumerate(texts):
+            # Tokenize the text using HuggingFace transformers
+            tokenized: List[int] = tokenizer.encode(text, add_special_tokens=False)
+
+            # Split tokens into chunks respecting the embedding_ctx_length
+            for j in range(0, len(tokenized), self.embedding_ctx_length):
```
efriis, line 336: I _think_ if you pass in an empty text, this will do weird things with the number of entries. Not 100% confident though

libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -518,24 +485,19 @@ def embed_documents(
...
-            return embeddings
+            return self._get_embeddings(texts, chunk_size=chunk_size)
 
-        # NOTE: to keep things simple, we assume the list may contain texts longer
-        #       than the maximum context and use length-safe embedding function.
-        engine = cast(str, self.deployment)
-        return self._get_len_safe_embeddings(texts, engine=engine)
+        num_texts = len(texts)
+        split_tokens, embedding_to_text_indices = self._tokenize_and_split(texts)
+        if len(split_tokens) == num_texts:
```
efriis, line 492: Can we add a test where one of the texts is `""`, and one of the texts is worth 2 splits? I think this will do funky things.

libs/partners/openai/langchain_openai/embeddings/base.py
efriis: ++ do we want to set a limit on parallelism? I think this could end up rate limiting pretty quick

libs/partners/openai/tests/integration_tests/embeddings/test_base.py
```diff
@@ -1,59 +1,88 @@
...
+    """Test openai embeddings."""
+    long_text = " ".join(["foo bar"] * embeddings.embedding_ctx_length)
+    token_splits, _ = embeddings._tokenize_and_split([long_text])
+    assert len(token_splits) > 1
+    output = embeddings.embed_documents([long_text])
+    assert len(output) == 1
+    assert len(output[0]) > 0
+
+
+async def test_embed_documents_long_async(
```
efriis, line 79: add a test with an empty text

libs/partners/openai/tests/unit_tests/embeddings/test_base.py
```diff
@@ -1,18 +1,53 @@
...
+
+@pytest.mark.parametrize(
+    ("vector", "expected"),
+    [
+        ([0.1], [1]),
+        ([1, 1, 1], [3**0.5 / 3] * 3),
+        ([27, 95, 13], [0.2710455418938115, 0.9536787585152627, 0.13050340905998334]),
+    ],
+)
+def test__vector_norm(vector: List[float], expected: List[float]) -> None:
```
efriis, line 51: could unit test to make sure an empty text is still set as an index

libs/partners/openai/langchain_openai/embeddings/base.py
efriis: @nfcampos is the right way to do this using an `asyncio.Semaphore(paralellism)` like this?

Would be great to land this

```suggestion
        semaphore = asyncio.Semaphore(parallelism)
        async def create_limit_parallelism(inputs: List) -> List:
            async with semaphore:
                return await self.async_client.create(input=inputs, **self._invocation_params)
        responses = await asyncio.gather(
            *(
                create_limit_parallelism(input[start : start + chunk_size])
                for start in range(0, len(input), chunk_size)
            )
        )
```

libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -251,6 +233,11 @@ class OpenAIEmbeddings(BaseModel, Embeddings):
     check_embedding_ctx_length: bool = True
     """Whether to check the token length of inputs and automatically split inputs 
         longer than embedding_ctx_length."""
+    max_concurrency: Optional[int] = None
```
baskaryan, line 236: @nfcampos what's a reasonable default value here?

PR #22367, opened 2024-05-31
docs: add Microsoft Azure to ChatModelTabs

mainred:


bracesproul (commented):
This looks great! Thanks for taking initiative on this. We would really appreciate it if you ported this change to the JS repo as well, in [this file](https://github.com/langchain-ai/langchainjs/blob/main/docs/core_docs/src/theme/ChatModelTabs.js)

If not, no worries, and thanks again for putting up this PR

docs/src/theme/ChatModelTabs.js
```diff
@@ -35,13 +36,15 @@ export default function ChatModelTabs(props) {
     mistralParams,
     googleParams,
     togetherParams,
+    azureParams,
```
mainred, line 39: @ccurme I did not defined `azureParams` as the others because AzureChatOpenAI does not accept a model as an input variable. I wanted to add a comment in L92 to describe the default model gpt-3.5-turbo with version 0125 but deleted considering this comment is visible in any places consuming ChatModelTabs while gpt-3.5-turbo is a the best choice, though there's no such case yet.

ccurme: Thanks for the clarification @mainred. I think it's OK to keep `azureParams`. When contributors add documentation it will be on them to ensure that the params are valid (e.g., they specify `azure_endpoint`).

Let me know if I've misunderstood your concern and we can open another PR.

PR #22378, opened 2024-05-31
Update chat model standard tests

baskaryan:
- Refactor standard test classes to make them easier to configure
- Update openai to support stop_sequences init param
- Update groq to support stop_sequences init param
- Update fireworks to support max_retries init param
- Update ChatModel.bind_tools to type tool_choice
- Update groq to handle tool_choice="any". **this may be controversial**


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -946,6 +946,8 @@ def dict(self, **kwargs: Any) -> Dict:
     def bind_tools(
         self,
         tools: Sequence[Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool]],
+        *,
+        tool_choice: Union[Literal["any", "auto", "none"], str, None] = None,
```
ccurme, line 950: (nit) might make sense to do this in its own PR

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -345,6 +345,8 @@ class BaseChatOpenAI(BaseChatModel):
     http_async_client: Union[Any, None] = None
     """Optional httpx.AsyncClient. Only used for async invocations. Must specify 
         http_client as well if you'd like a custom client for sync invocations."""
+    stop: Optional[Union[List[str], str]] = Field(default=None, alias="stop_sequences")
```
ccurme, line 348: If I initialize ChatOpenAI with `stop` now, I get a warning:
```
UserWarning: WARNING! stop is not default parameter.
stop was transferred to model_kwargs.
Please confirm that stop is what you intended.
```
Are we un-doing a change here?

ccurme (commented):
changes to standard tests LGTM but would like @efriis's opinion

libs/core/langchain_core/language_models/chat_models.py
baskaryan: yea i got a bit carried away

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: i think this should have opposite effect?

libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
```diff
@@ -1,74 +1,31 @@
...
-from typing import Type
 
 import pytest
 from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import AIMessage, AIMessageChunk, HumanMessage, ToolMessage
-from langchain_core.pydantic_v1 import BaseModel, Field
-from langchain_core.tools import tool
 
+from langchain_standard_tests.unit_tests.chat_models import (
+    ChatModelTests,
```
efriis, line 8: why subclass? Can't we treat them separately?

If anything should be shared, maybe do a base class?

libs/standard-tests/langchain_standard_tests/unit_tests/chat_models.py
```diff
@@ -18,81 +21,109 @@ def my_adder_tool(a: int, b: int) -> int:
...
+
     @pytest.fixture
-    def chat_model_has_tool_calling(
-        self, chat_model_class: Type[BaseChatModel]
+    def model(self) -> BaseChatModel:
+        return self.chat_model_class(
+            **{**self.standard_chat_model_params, **self.chat_model_params}
+        )
+
+    @property
```
efriis, line 50: also open to making these abstract, so user _has_ to override them

efriis (commented):
I think having some extra duplicate stuff between the stubs for each is probably good / verbosity is good in tests like this

Happy to discuss though. main thing is that we probably shouldn't copy all the unit tests into the integration tests.

libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
baskaryan: that is what this is

baskaryan: see ChatModelUnitTests for unit tests

efriis: lol facepalm

libs/partners/ai21/tests/integration_tests/test_standard.py
```diff
@@ -10,98 +10,38 @@
...
-            chat_model_class,
-            chat_model_params,
-        )
+    def test_usage_metadata(self, model: BaseChatModel) -> None:
+        super().test_usage_metadata(model)
 
-    @pytest.fixture
+
+class TestAI21J2(BaseTestAI21):
+    @property
```
efriis, line 36: other option would be a pytest.mark.parameterize on the two diff models in one class

efriis (approved):
works for me. Only downside of property vs pytest.fixture is it's harder to parameterize (see comment on AI21). IMO property is and duplicated class is more readable so I like this

This will immediately break tests in other repos so should fix em on the quick end

libs/partners/groq/langchain_groq/chat_models.py
```diff
@@ -461,8 +464,6 @@ def _create_message_dicts(
     ) -> Tuple[List[Dict[str, Any]], Dict[str, Any]]:
         params = self._default_params
         if stop is not None:
-            if "stop" in params:
-                raise ValueError("`stop` found in both the input and default params.")
```
ccurme, line 465: Do we need this for openai as well?
```python

model = ChatOpenAI()

model.invoke("hi", stop=["you"])
```
```
ValueError: `stop` found in both the input and default params.
```

libs/standard-tests/langchain_standard_tests/unit_tests/chat_models.py
```diff
@@ -18,81 +21,109 @@ def my_adder_tool(a: int, b: int) -> int:
...
+class ChatModelTests(ABC):
+    @property
     @abstractmethod
-    @pytest.fixture
     def chat_model_class(self) -> Type[BaseChatModel]:
         ...
 
-    @pytest.fixture
+    @property
     def chat_model_params(self) -> dict:
```
ccurme, line 31: do these need to be fixtures? I'm getting `fixture 'chat_model_class' not found` on `test_stop_sequence`.

PR #19894, opened 2024-04-02
community[minor] : updated openllm api

baskaryan:
Updated #12968

PR #22380, opened 2024-06-01
community: update support for OpenLLM 0.5

aarnphm:
This PR updates the internal wrapper for OpenLLM

Since OpenLLM v0.5 is considered breaking, I have also taken this opportunity to separate the  API and local inference component into two separate class:
- `OpenLLMAPI` will be responsible for handing remote server going forward (async, sync, streaming supported)
- `OpenLLM` will now only for local inference, and only support batching and synchronous generations.

Hopefully this would make it a bit more future-proof. In terms of testing, I believe I have tested through some manual workflow and will be include the tests on our end.

updates from https://github.com/langchain-ai/langchain/pull/19894, cc @baskaryan


efriis:
CI failing - feel free to reopen as a draft to debug, and full PR when it's passing!

PR #22391, opened 2024-06-01
feat: Implement MiniMaxChat interface

maang-h:
- **Description:** Implement MiniMaxChat interface, include:
    - No longer inherits the LLM class (like other chat model)
    - Update request parameters (v1 -> v2)
        - update `base url`
        - update message role (system, user, assistant)
        - add `stream` function
        - no longer use `group id`
    - Implement the `_stream`, `_agenerate`, and `_astream` interfaces

 [minimax v2 api document](https://platform.minimaxi.com/document/guides/chat-model/V2?id=65e0736ab2845de20908e2dd)

### DIFF
--- a/libs/community/langchain_community/chat_models/minimax.py
+++ b/libs/community/langchain_community/chat_models/minimax.py
@@ -1,61 +1,212 @@
 """Wrapper around Minimax chat models."""
+import json
 import logging
-from typing import Any, Dict, List, Optional, cast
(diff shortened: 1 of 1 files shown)


maang-h:
@baskaryan @eyurtsev  cc.  And should the class name be changed to `ChatMiniMax`?

baskaryan (approved):

PR #22393, opened 2024-06-01
core: In RunnableSequence pass kwargs to the first step

nfcampos:
- This is a pattern that shows up occasionally in langgraph questions, people chain a graph to something else after, and want to pass the graph some kwargs (eg. stream_mode)

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -2553,6 +2553,7 @@ class RunnableSequence(RunnableSerializable[Input, Output]):
                             )
                             for rm, config in zip(run_managers, configs)
                         ],
+                        return_exceptions=return_exceptions,
                         **(kwargs if i == 0 else {}),
                     )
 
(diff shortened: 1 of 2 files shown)


efriis (dismissed):
Just being clear, this basically make the following easier (and equivalent):

Possible previously:
```
chain = r1.bind(**my_kwargs) | r2.bind(**my_kwargs) | r3.bind(**my_kwargs)
chain.invoke(input)
```

Possible now:
```
chain = r1 | r2 | r3
chain.invoke(input, **my_kwargs)
```

nfcampos:
No, kwargs are passed only to first step in sequence

efriis:
That feels unintuitive to me. I have a slight preference for requiring people to `bind()` in this case. 

For the specific case of graph into something else, why not
```
chain = graph.bind(**my_kwargs) | my_next_step
chain.invoke(input)
```

An equivalently valid use case feels like having some sort of "pre" step before the graph that computes/gathers inputs, which wouldn't work for this interface

efriis:
Or maybe a `RunnableSequence.bind_step(step_num, **kwargs)`?

So would be
```
chain = graph | my_next_step
chain.bind_step(0, **my_kwargs).invoke(input)
```

nfcampos:
No I think this is quite natural? After all the input is also passed only to the first step, not to all others

efriis (approved):
Fair. Should we add kwarg-forwarding to `RunnableParallel` too?

PR #22397, opened 2024-06-02
community[patch]: update dynamodb chat history to update instead of overwrite

Davi-Schumacher:
**Description:**
The current implementation of `DynamoDBChatMessageHistory` updates the `History` attribute for a given chat history record by first extracting the existing contents into memory, appending the new message, and then using the `put_item` method to put the record back. This has the effect of overwriting any additional attributes someone may want to include in the record, like chat session metadata.

This PR suggests changing from using `put_item` to using `update_item` instead which will keep any other attributes in the record untouched. The change is backward compatible since 
1. `update_item` is an "upsert" operation, creating the record if it doesn't already exist, otherwise updating it
2. It only touches the db insert call and passes the exact same information. The rest of the class is left untouched

**Dependencies:**
None

**Tests and docs:**
No unit tests currently exist for the `DynamoDBChatMessageHistory` class. This PR adds the file `libs/community/tests/unit_tests/chat_message_histories/test_dynamodb_chat_message_history.py`  to test the `add_message` and `clear` methods. I wanted to use the moto library to mock DynamoDB calls but I could not get poetry to resolve it so I mocked those calls myself in the test. Therefore, no test dependencies were added.

The change was tested on a test DynamoDB table as well. The first three images below show the current behavior. First a message is added to chat history, then a value is inserted in the record in some other attribute, and finally another message is added to the record, destroying the other attribute.
using_put_1_first_message
using_put_2_add_attribute
using_put_3_second_message

The next three images show the new behavior. Once again a value is added to an attribute other than the History attribute, but now when the followup message is added it does not destroy that other attribute. The History attribute itself is unaffected by this change.
using_update_1_first_message
using_update_2_add_attribute
using_update_3_second_message

The doc located at `docs/docs/integrations/memory/aws_dynamodb.ipynb` required no changes and was tested as well.

### DIFF
--- a/libs/community/langchain_community/chat_message_histories/dynamodb.py
+++ b/libs/community/langchain_community/chat_message_histories/dynamodb.py
@@ -167,16 +167,19 @@ class DynamoDBChatMessageHistory(BaseChatMessageHistory):
                 import time
 
                 expireAt = int(time.time()) + self.ttl
-                self.table.put_item(
-                    Item={
-                        **self.key,
-                        self.history_messages_key: messages,
-                        self.ttl_key_name: expireAt,
-                    }
+                self.table.update_item(
+                    Key={**self.key},
+                    UpdateExpression=(
+                        f"set {self.history_messages_key} = :h, "
+                        f"{self.ttl_key_name} = :t"
+                    ),
+                    ExpressionAttributeValues={":h": messages, ":t": expireAt},
                 )
             else:
-                self.table.put_item(
-                    Item={**self.key, self.history_messages_key: messages}
+                self.table.update_item(
+                    Key={**self.key},
+                    UpdateExpression=f"set {self.history_messages_key} = :h",
+                    ExpressionAttributeValues={":h": messages},
                 )
         except ClientError as err:
             logger.error(err)
--- a/libs/community/tests/unit_tests/chat_message_histories/test_dynamodb_chat_message_history.py
+++ b/libs/community/tests/unit_tests/chat_message_histories/test_dynamodb_chat_message_history.py
@@ -1,0 +1,153 @@
+from typing import Any
+
+import pytest
+from langchain_core.messages import AIMessage, HumanMessage, messages_to_dict
+from pytest_mock import MockerFixture
+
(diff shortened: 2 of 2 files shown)


Davi-Schumacher:
Hi @eyurtsev, can I get some guidance on what it would take to get this PR in shape for review/approval?

libs/community/langchain_community/chat_message_histories/dynamodb.py
```diff
@@ -167,16 +167,19 @@ def add_message(self, message: BaseMessage) -> None:
...
-                self.table.put_item(
-                    Item={
-                        **self.key,
-                        self.history_messages_key: messages,
-                        self.ttl_key_name: expireAt,
-                    }
+                self.table.update_item(
+                    Key={**self.key},
+                    UpdateExpression=(
+                        f"set {self.history_messages_key} = :h, "
```
eyurtsev, line 173: Does this value need to be sanitized to avoid query injection?  We can validate it at the __init__ or here

same for ttl_key_name and anything else that might require escaping?

eyurtsev:
@Davi-Schumacher looks good, could you confirm whether we need to escape any of the values to avoid sql injection?

eyurtsev (approved):

libs/community/langchain_community/chat_message_histories/dynamodb.py
Davi-Schumacher: Thanks for the review!

For query injection, I am confident the Boto API here will prevent it. The API builds the query, we don't pass one in, so you can see the `UpdateExpression` just sets the update action and parameterizes the fields to apply the action to, and the `ExpressionAttributeValues` maps the values to the parameters. While the update expression could be tampered with, the most you could do is pass a `self.history_messages_key` that is composed of additional actions and attribute names. But the value of `self.history_messages_key` is set by the developer and is not exposed to end-user data. Only the chat history items passed in `ExpressionAttributeValues` are exposed to user data and that's already parameterized, so having a DynamoDB query in your chat history for example will just send that query as a value in your chat history, it has no way of being executed. Same goes for TTL.

Issue #13368, opened 2023-11-14
module 'openai' has no attribute 'error'


ryanpeach:
### System Info

openai==1.2.4
langchain==0.0.325
llama_index==0.8.69

### Who can help?

@hwchase17 @agola11 

### Information

- [ ] The official example notebooks/scripts
- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [X] Embedding Models
- [ ] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

```
/usr/local/lib/python3.10/site-packages/llama_index/indices/base.py:102: in from_documents
    return cls(
/usr/local/lib/python3.10/site-packages/llama_index/indices/vector_store/base.py:49: in __init__
    super().__init__(
/usr/local/lib/python3.10/site-packages/llama_index/indices/base.py:71: in __init__
    index_struct = self.build_index_from_nodes(nodes)
/usr/local/lib/python3.10/site-packages/llama_index/indices/vector_store/base.py:254: in build_index_from_nodes
    return self._build_index_from_nodes(nodes, **insert_kwargs)
/usr/local/lib/python3.10/site-packages/llama_index/indices/vector_store/base.py:235: in _build_index_from_nodes
    self._add_nodes_to_index(
/usr/local/lib/python3.10/site-packages/llama_index/indices/vector_store/base.py:188: in _add_nodes_to_index
    nodes = self._get_node_with_embedding(nodes, show_progress)
/usr/local/lib/python3.10/site-packages/llama_index/indices/vector_store/base.py:100: in _get_node_with_embedding
    id_to_embed_map = embed_nodes(
/usr/local/lib/python3.10/site-packages/llama_index/indices/utils.py:137: in embed_nodes
    new_embeddings = embed_model.get_text_embedding_batch(
/usr/local/lib/python3.10/site-packages/llama_index/embeddings/base.py:250: in get_text_embedding_batch
    embeddings = self._get_text_embeddings(cur_batch)
/usr/local/lib/python3.10/site-packages/llama_index/embeddings/langchain.py:82: in _get_text_embeddings
    return self._langchain_embedding.embed_documents(texts)
[...]

PR #22399, opened 2024-06-02
community: upgrade openai>=1.30 for LocalAIEmbedings 

mkhludnev:

- **Description:** migrate LocalAIEmbeddings code to OpenAI 1.30
- **Issue:** LocalAIEmbeddings fails due to old OpenAI 0.28 code openai.error see #13368
- **Dependencies:** OpenAI is already  bumped 1.30

- [x] **Add tests and docs**: integration test with vcr

- [x] **Lint and test**: 

Additional guidelines:


libs/community/langchain_community/embeddings/localai.py
```diff
@@ -240,21 +253,25 @@ def validate_environment(cls, values: Dict) -> Dict:
...
             **self.model_kwargs,
         }
         if self.openai_proxy:
             import openai
 
-            openai.proxy = {
-                "http": self.openai_proxy,
-                "https": self.openai_proxy,
-            }  # type: ignore[assignment]
+            # TODO: The 'openai.proxy' option isn't read in the client API. You will need to pass it when you instantiate the client, e.g. 'OpenAI(proxy={
```
mkhludnev, line 267: still not yet done

libs/community/langchain_community/embeddings/localai.py
```diff
@@ -228,7 +237,20 @@ def validate_environment(cls, values: Dict) -> Dict:
...
             import openai
 
-            values["client"] = openai.Embedding
+            client_params = {
+                "api_key": values["openai_api_key"],
+                "organization": values["openai_organization"],
+                "base_url": values["openai_api_base"],
+                "timeout": values["request_timeout"],
+                "max_retries": values["max_retries"],
+                # "default_headers": values["default_headers"],
```
mkhludnev, line 246: we'd rather wipe it right?

libs/community/langchain_community/embeddings/localai.py
```diff
@@ -240,21 +262,14 @@ def validate_environment(cls, values: Dict) -> Dict:
     def _invocation_params(self) -> Dict:
         openai_args = {
             "model": self.model,
-            "request_timeout": self.request_timeout,
-            "headers": self.headers,
-            "api_key": self.openai_api_key,
-            "organization": self.openai_organization,
-            "api_base": self.openai_api_base,
-            "api_version": self.openai_api_version,
+            # "request_timeout": self.request_timeout,
```
mkhludnev, line 265: also wipe it TODO

libs/community/langchain_community/embeddings/localai.py
```diff
@@ -135,15 +135,30 @@ class LocalAIEmbeddings(BaseModel, Embeddings):
...
 
+    Specifying proxy:
+        .. code-block:: python
+
+            from langchain_community.embeddings import LocalAIEmbeddings
+            import openai
+            import httpx
+            openai = LocalAIEmbeddings(
+                openai_api_key="random-string",
+                client=openai.OpenAI(
```
baskaryan, line 146: shouldn't need to pass in client directly

mkhludnev: I think it should be 
https://github.com/openai/openai-python/issues/825#issuecomment-1813730980
https://github.com/openai/openai-python#configuring-the-http-client

mkhludnev: eg. I don't see actual proxy usage in https://github.com/langchain-ai/langchain/blob/master/libs/partners/openai/langchain_openai/embeddings/base.py

mkhludnev:
I got back to the problem. 
Let me ask very basic question: which version of openai we expect for `langchain_community` 1.x or 0.x? 
We can assume that 90% of users have both  `langchain_community`  and  `langchain-openai` in their dependencies. 
Although I don't know whether  `langchain-openai` is able to work with openai 0.x.

libs/community/langchain_community/embeddings/localai.py
mkhludnev: > shouldn't need to pass in client directly

I copied `openai_proxy` handling from OpenAIEmbedding

mkhludnev:
@baskaryan kindly asking to check this PR out. I resolved the question regarding proxy. Note, after this PR is merged. I want to improve #22666 localai embeddings performance. And then update LocalAI LLM completion/chat classes for OpenAI 1.x.

mkhludnev:
ahhhh it's a bummer 

```
       we can conclude that friendli-client>=1.2.4,<2 depends on
      httpx>=0.24.1,<0.25.0.
      And because you require friendli-client>=1.2.4,<2 and httpx>=0.27.0, we
      can conclude that your requirements are unsatisfiable.
```

mkhludnev:
ping @efriis pleaase

dawu415:
Curious if there's any update on this or alternative solutions in the meantime?

mkhludnev:
> Curious if there's any update on this or alternative solutions in the meantime?

I can only rebase it from time to time. Let me know if it's necessary. I'll refresh.

efriis:
Unfortunately there isn't a great way to update these old community integrations that rely on openai sdk v0 in a way that's both backwards compatible + supports v1 without being a huge mess to maintain. This is one of the main motivations behind moving towards dedicated integration packages, which have replaced the community package as the right path forward. I'll close this PR, and would recommend reopening with just docs updates, as well as registering your package in `libs/packages.yml`! We'll be able to review simple PRs that only modify these two things much faster :)

Would also love to @deprecated this integration pointing to your new integration package, which can even lock to openai v1 sdk to avoid support for v0!

mkhludnev:
Thanks @efriis 
I read it as a need to copy LocalAI Embeddings to the separate repo that I did https://github.com/mkhludnev/langchain-localai/

The first question is which account this integration package will belong to:

- can you adopt it into  https://github.com/langchain-ai/langchain-localai ?
- or @mudler you can kindly keep it under https://github.com/mudler/langchain-localai ? 
- I think `langchain/libs/partners/langchain-localai ` isn't a preferable option anymore?

Then, more serious issue. I can't manage pytest-vcr to work in this standalone integration repo. And the worse after I moved folders to root pytest breaks on asyncio config option. That's sad. I'm glad to any help or advice. Thanks! 

PS. The next (real target) is a performance via #22666

efriis:
we won't create one under langchain-ai, and one under `mudler` would probably be best as an integration package.

However, that being said, because it relies on the `openai` sdk anyways, I would strongly recommend just rewriting the Local AI docs page in langchain to use the `ChatOpenAI` integration instead. That way it benefits from all fixes folks make against that package, and we can deprecate `ChatLocalAI` in favor of just using that.

mkhludnev:
> However, that being said, because it relies on the `openai` sdk anyways, I would strongly recommend just rewriting the Local AI docs page

aaaghh I've got your point after all. Let me check if embeddings work this way!

mkhludnev:
@efriis 
I tried to invoke LocalAI embeddings via OpenAIEmbeddings as far as I've got your suggestion. 
```
embeddings = OpenAIEmbeddings(
    model="bge-m3",
    openai_api_base="https://localai.my/v1",
    api_key="foo bar")
vectors = embeddings.embed_documents(["lorem ipsum","dolor sit amet"])
```
I noticed that underneath OpenAIEmbeddings tokenizes text by [cl100k_base](https://github.com/langchain-ai/langchain/blob/c2a3021bb0c5f54649d380b42a0684ca5778c255/libs/partners/openai/langchain_openai/embeddings/base.py#L422l) and submits int array per the given string. LocalAI handles it somehow and yields vectors. Probably they are meaningful, but their values are different from what I've got from LocalAIEmbeddings, which submits texts as-is. 
So, even if we got confirmation that `"intput":[[1,2,3,..],[4,5,6]]` a valid way to call https://localai.io/features/embeddings/ I don't think users were encouraged by need to reindex all vectors just due to client upgrade. 
So, wdyt? Is there a country for LocalAIEmbeddings?

mkhludnev:
> `ChatLocalAI`

Let me clarify. There's no chat model for LocalAI here, there's only `LocalAIEmbeddings`

efriis:
I see makes sense. Feel free to publish a langchain-localai then!

Issue #22389, opened 2024-06-01
REGRESSION: ChatLiteLLM: ValidationError only when using cache


thiswillbeyourgithub:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

``` python
import os
from pathlib import Path

from langchain.globals import set_verbose, set_debug, set_llm_cache
from langchain_community.chat_models import ChatLiteLLM
from langchain_community.cache import SQLiteCache
from langchain_core.output_parsers.string import StrOutputParser

os.environ["OPENAI_API_KEY"] = Path("OPENAI_API_KEY.txt").read_text().strip()
set_verbose(True)
set_debug(True)
Path("test_cache.db").unlink(missing_ok=True)
set_llm_cache(SQLiteCache(database_path="test_cache.db"))

llm = ChatLiteLLM(
    model_name="openai/gpt-4o",
    cache=True,
    verbose=True,
    temperature=0,
    )
print(llm.predict("this is a test"))  # works fine because cache empty
print("Success 1/2")
print(llm.predict("this is a test"))  # fails
print("Success 2/2")
```

### Error Message and Stack Trace (if applicable)
```
Success 1/2
[llm/start] [llm:ChatLiteLLM] Entering LLM run with input:
{
  "prompts": [
    "Human: this is a test"
  ]
}
Retrieving a cache value that could not be deserialized properly. This is likely due to the cache being in an older format. Please recreate your cache to avoid this error.
[llm/error] [llm:ChatLiteLLM] [3ms] LLM run errored with error:
[...]

PR #22408, opened 2024-06-02
[Core,Community]: Fixed SQLAlchemyCache serialization exception to return ChatGeneration and added `type` in llm_string

keenborder786:
**Description** 

- In SQLAlchemyCache, an exception is raised due to a `NotImplementedError` during the conversion of a looked-up JSON object to a Langchain object. This caused the function to return a `Generation` object instead of the expected `ChatGeneration` object, as specified [here](https://github.com/langchain-ai/langchain/blob/acaf214a4516a2ffbd2817f553f4d48e6a908695/libs/core/langchain_core/language_models/chat_models.py#L634). The discrepancy resulted in an error because `ChatResult` expects a `ChatGeneration` object.
- To fix this issue, I ensured that the object returned from the `lookup` function is of type `ChatGeneration` if the underlying LLM belongs to `BaseChatModel`. If the underlying model belongs to `BaseLLM`, the function returns an object of type `Generation`. I also modified the `core` code to encode `type` information in the underlying LLM string in order to check the type of LLM.
- This fix should be applied to the `lookup` functions of other cache types as well because all caches currently return objects of only type `Generation`, which is incorrect if the underlying model is of type `BaseChatModel`. I will implement these changes once I have confirmation that this approach is correct.
- **Note:** Another possible fix was to not raise the [exception](https://github.com/langchain-ai/langchain/blob/acaf214a4516a2ffbd2817f553f4d48e6a908695/libs/core/langchain_core/load/load.py#L67) and simply return the `not_implemented` object in the `Revivier`. However, since this change could affect core functionality and introduce unintended or breaking changes, I decided to isolate the fix to the `SQLAlchemyCache` in the community package.

**Issue**: #22389


keenborder786:
@eyurtsev please check

keenborder786:
@eyurtsev please check

keenborder786:
@eyurtsev please check

keenborder786:
@eyurtsev @ccurme

libs/community/langchain_community/cache.py
```diff
@@ -272,6 +273,18 @@ def lookup(self, prompt: str, llm_string: str) -> Optional[RETURN_VAL_TYPE]:
                     )
                     # In a previous life we stored the raw text directly
                     # in the table, so assume it's in that format.
+                    if "('type', 'chat')" in llm_string:
```
eyurtsev, line 276: Is it possible to check whether the data looks like a ChatGeneration and if so return it otherwise return Generation?

I'd like to avoid the code checking the contents of llm_string to reason about what model is being used

eyurtsev (changes requested):
Left a few comments here. This solution has dynamic module imports that we definitely don't want in the code as it's a security vulnerability

keenborder786:
@eyurtsev will check your comments

Issue #21482, opened 2024-05-09
AttributeError on calling LLMGraphTransformer.convert_to_graph_documents


mintomohan:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code
```
from langchain_experimental.graph_transformers.llm import LLMGraphTransformer
from langchain.llms.bedrock import Bedrock
from langchain.text_splitter import RecursiveCharacterTextSplitter
from langchain_community.retrievers import WikipediaRetriever
from langchain_community.graphs.neo4j_graph import Neo4jGraph

bedrock=boto3.client(service_name='bedrock-runtime')

def prepare_graph(wiki_keyword):
    wiki_retriever = WikipediaRetriever(doc_content_chars_max=2000, top_k_results=1)
    docs = wiki_retriever.invoke(wiki_keyword)
    
    text_splitter = RecursiveCharacterTextSplitter(chunk_size=500, chunk_overlap=50)
    doc_chunks = text_splitter.split_documents(docs)
    
    llm=Bedrock(model_id='mistral.mistral-7b-instruct-v0:2', client=bedrock)
    llm_transformer = LLMGraphTransformer(llm=llm)    
    
    graph_documents = llm_transformer.convert_to_graph_documents(doc_chunks)
    graph = Neo4jGraph()
    graph.add_graph_documents(graph_documents)
```

### Error Message and Stack Trace (if applicable)

```
Traceback (most recent call last):
  File "...\...\load_data_graph.py", line 46, in <module>
    prepare_graph('Paris')
  File "...\...\load_data_graph.py", line 38, in prepare_graph
    graph_documents = llm_transformer.convert_to_graph_documents(doc_chunks)
  File "...\...\venv\lib\site-packages\langchain_experimental\graph_transformers\llm.py", line 646, in convert_to_graph_documents   
    return [self.process_response(document) for document in documents]
  File "...\...\venv\lib\site-packages\langchain_experimental\graph_transformers\llm.py", line 646, in <listcomp>
    return [self.process_response(document) for document in documents]
  File "...\...\venv\lib\site-packages\langchain_experimental\graph_transformers\llm.py", line 595, in process_response
[...]

PR #22417, opened 2024-06-03
experimental: Fix graph_transformers llms #21482

liugz18:
Fix AttributeError on calling LLMGraphTransformer.convert_to_graph_documents #21482

 since raw_schema is always a str

@baskaryan


baskaryan:
@tomasonjo

tomasonjo:
I don't think it's always a string as this is how it's defined:

```
self.chain = prompt | llm
```

I've also tested this approach with Ollama.
Perhaps there are some inconsistencies between different LLM providers...
Can you change the code so that it checks if raw_schema isn't a string it extracts content attribute, otherwise it doesn't do anything?

tomasonjo:
Thanks @liugz18 ... cc @baskaryan

PR #22419, opened 2024-06-03
Fix for AzureSearch delete method not using FIELDS_ID (Generated by Ana - AI SDE)

### LINKED ISSUE #22314: AzureSearch delete method does not use the variable FIELDS_ID therefore it does not override the value
### Checked other resources

### Example Code

The following code from AzureSearch class
```
def delete(self, ids: Optional[List[str]] = None, **kwargs: Any) -> bool:
        """Delete by vector ID.

        Args:
            ids: List of ids to delete.

        Returns:
            bool: True if deletion is successful,
            False otherwise.
        """
        if ids:
            res = self.client.delete_documents([{"id": i} for i in ids])
            return len(res) > 0
        else:
            return False
```
Do not use the FIELDS_ID variable defined at the top of the script therefore it doesn't allow complete override of the key field in Azure AI Search.

Simply replacing by:

```
def delete(self, ids: Optional[List[str]] = None, **kwargs: Any) -> bool:
        """Delete by vector ID.

        Args:
            ids: List of ids to delete.

        Returns:
            bool: True if deletion is successful,
            False otherwise.
        """
        if ids:
            res = self.client.delete_documents([{FIELDS_ID: i} for i in ids])
            return len(res) > 0
        else:
            return False
```
Will do the trick.

_No response_

### Description

I'm using langchain to interact with Azure AISearch , as I want to remove documents based on their id (the key field in AISearch) I would like Langchain to offer override capabilities for this field

### System Info

No specific info

ana-ai-sde:
### Description
This PR addresses a bug in the `AzureSearch` class where the `delete` method did not use the `FIELDS_ID` variable, which prevented the complete override of the key field in Azure AI Search. By modifying the `delete` method, we ensure that `FIELDS_ID` is used instead of the hard-coded "id" string, providing the necessary flexibility for different document identifier field names.

### Fixes
This patch fixes [Issue 22314](https://github.com/langchain-ai/langchain/issues/22314).

This patch was generated by [Ana - AI SDE](https://openana.ai/), an AI-powered software development assistant.


d4rk-lucif3r:
@ccurme 
Hi
I am the author of Ana. Sorry for the wrong commit, due to some bug, Ana committed the wrong fix.

We have fixed it and Ana has created another PR here: https://github.com/langchain-ai/langchain/pull/22436

PR #22421, opened 2024-06-03
community: fix missing `apify_api_token` field in ApifyWrapper

charl3sj:
- **Description:** The `ApifyWrapper` class expects `apify_api_token` to be passed as a named parameter or set as an environment variable. But the corresponding field was missing in the class definition causing the argument to be ignored when passed as a named param. This patch fixes that.


ccurme (approved):
thanks!

PR #22438, opened 2024-06-03
docs: fix api ref link generation

efriis:


baskaryan (approved):

docs/api_reference/guide_imports.json
baskaryan, line 1: wait added bunch of ipynb_checkpoints

baskaryan: ok think i fixed

PR #22439, opened 2024-06-03
chore(ULIDs): Changing uuids to uuid1 for database performance

franz101:
UUIDs should be uuid1,uuid6,uuid7 or even better snowflakeIDs

image


hwchase17:
our convention is uuid4, we'll likely stick with that

PR #22442, opened 2024-06-03
feat(0.5): update support for OpenLLM 0.5

aarnphm:
This PR updates the internal wrapper for OpenLLM

Since OpenLLM v0.5 is considered breaking, I have also taken this opportunity to separate the  API and local inference component into two separate class:
- `OpenLLMAPI` will be responsible for handing remote server going forward (async, sync, streaming supported)
- `OpenLLM` will now only for local inference, and only support batching and synchronous generations.

Hopefully this would make it a bit more future-proof. In terms of testing, I believe I have tested through some manual workflow and will be include the tests on our end.

updates from https://github.com/langchain-ai/langchain/pull/19894, cc @baskaryan


libs/community/langchain_community/llms/openllm.py
```diff
@@ -6,247 +6,121 @@
...
     def __init__(
         self,
-        model_name: Optional[str] = None,
-        *,
-        model_id: Optional[str] = None,
-        server_url: Optional[str] = None,
+        server_url: str,
         timeout: int = 30,
-        server_type: Literal["grpc", "http"] = "http",
-        embedded: bool = True,
```
baskaryan, line 132: shoudl we keep around the old args and raise a warning if they're passed in, so that this isn't a breaking change?

aarnphm: sure I think we can do that.

aarnphm: oh, this is for the new class OpenLLMAPI, but for OpenLLM we can keep backward compat.

aarnphm:
@baskaryan since the OpenLLM implementation here won't support `_acall` or async anymore, it would be a breaking change regardless.

libs/community/langchain_community/llms/openllm.py
baskaryan: ah missed the rename :+1

docs/docs/integrations/providers/openllm.mdx
```diff
@@ -15,40 +15,61 @@ Install the OpenLLM package via PyPI:
 pip install openllm
 ```
 
+> [!NOTE]
+> OpenLLM will require GPU to run locally. If you already have a OpenLLM server running elsewhere, you might want to install `openllm-client` instead.
```
aarnphm, line 19:
```suggestion
> `OpenLLM` will require GPU to run locally. If you already have a OpenLLM server running elsewhere, you might want to install `openllm-client` and use `OpenLLMAPI` instead.
```

docs/docs/integrations/llms/openllm.ipynb
```diff
@@ -84,8 +84,7 @@
     "from langchain_community.llms import OpenLLM\n",
     "\n",
     "llm = OpenLLM(\n",
-    "    model_name=\"dolly-v2\",\n",
-    "    model_id=\"databricks/dolly-v2-3b\",\n",
+    "    model_id=\"microsoft/Phi-3-mini-4k-instruct\",\n",
```
aarnphm, line 87:
```suggestion
    "    model_id=\"microsoft/Phi-3-mini-4k-instruct\",\n",
    "    trust_remote_code=True,\n",  
```

aarnphm:
bump @baskaryan when you have bandwidth. sorry for the ping :)

PR #22451, opened 2024-06-04
Update openai.py

danny-zhu:
fix the bug:
File "/home/xxx/.conda/envs/xxx/lib/python3.10/site-packages/langchain_community/chat_models/openai.py", line 493, in _astream
TypeError: object of type 'NoneType' has no len()


baskaryan (commented):
this is a deprecated implementation, the main one lives in libs/partners/openai/langchain_openai/chat_models/base.py

could we make any bugfixes there

ccurme:
@danny-zhu could you provide a snippet to reproduce this?

Issue #18235, opened 2024-02-28
DOC: Missing explanation of lancedb usage with filtering


datainvestor:
### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

Missing explanation how to use LanceDB with filter from langchain api, https://python.langchain.com/docs/integrations/vectorstores/lancedb

Based on the documentation I am trying to use Langchain with LanceDB as vector database. Here is how I instatiate database:

    db = lancedb.connect("./data/lancedb")
    
    table = db.create_table("my_docs", data=[
        {"vector": embeddings.embed_query(chunks[0].page_content), "text": chunks[0].page_content, "id": "1", "file":"bb"}
    ], mode="overwrite")

I then load more documents with different `file` metadata:

`vectordb = LanceDB.from_documents(chunks[1:], embeddings, connection=table)`

Then another batch with also a different `file` metadata value

`vectordb = LanceDB.from_documents(chunks_ma, embeddings, connection=table)`

I can see they were loaded succesfully and my vector db has correct amount of docuemnts:

`print(len(db['my_docs']))`

`11`

Now I want to create a retriever that will be able to pre filter the data based on `file` value:

I tried this 

    retriever = vectordb.as_retriever(search_kwargs={"k": 6, 'filter':{'file': 'bb'}})
    retrieved_docs = retriever.invoke("My query regarding something")

But when I check the outputs of the query invocation its still giving me the documents with wrong file metadata values:

`print(retrieved_docs[0].metadata['file'])`

`'cc'`

But it was supposed to only query the docuemnts in the database matchin the file value.

Is there something I am doing wrong, or what is the correct approach to filter the values before running retrieval query from LanceDB vector DB using Langchain api? I think the guidelines are missing from the documentation but would reatly help.

brnaba-aws:
@datainvestor can you check that filtering is now working based on this pull request? #22461 
Thanks

PR #22461, opened 2024-06-04
community[minor]: Added filter search for LanceDB

brnaba-aws:
- [ ] **community**: "vectorstore: added filtering support for LanceDB vector store"

- [ ] **This PR adds filtering capabilities to LanceDB**:
    - **Description:** In LanceDB filtering can be applied when searching for data into the vectorstore. It is using the SQL language as mentioned in the LanceDB documentation.
    - **Issue:** #18235 
    - **Dependencies:** No

### DIFF
--- a/libs/community/langchain_community/vectorstores/lancedb.py
+++ b/libs/community/langchain_community/vectorstores/lancedb.py
@@ -113,7 +113,7 @@ class LanceDB(VectorStore):
         Args:
             texts: Iterable of strings to add to the vectorstore.
             metadatas: Optional list of metadatas associated with the texts.
-            ids: Optional list of ids to associate w    ith the texts.
+            ids: Optional list of ids to associate with the texts.
 
         Returns:
             List of ids of the added texts.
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/lancedb.py
```diff
@@ -218,14 +218,42 @@ def similarity_search(
...
+                                                    })
         """
         embedding = self._embedding.embed_query(query)  # type: ignore
         tbl = self.get_table(name)
+        filters = kwargs.pop("filter", {})
+        sql_filter = filters.pop("sql_filter", None)
+        prefilter = filters.pop("prefilter", False)
         docs = (
             tbl.search(embedding, vector_column_name=self._vector_key)
+            .where(sql_filter, prefilter=prefilter)
```
eyurtsev, line 256: There's no unit test associated with this change. Could you either add a unit test or help confirm that the default case where sql_filter = None works? Should the query be built conditionally?

brnaba-aws: if filter is not mentioned, then it works. I tried. That's why I decided to not put any condition around the query.

eyurtsev (approved):

PR #22465, opened 2024-06-04
community: Improve Cassandra VectorStore as_retriever

cbornet:
The Vectorstore's API  `as_retriever` doesn't expose explicitly the parameters `search_type` and `search_kwargs` and so these are not well documented.
This PR improves  `as_retriever` for the Cassandra VectorStore by making these parameters explicit.

NB: An alternative would have been to modify `as_retriever` in `Vectorstore`. But there's probably a good reason these were not exposed in the first place ? Is it because implementations may decide to not support them and have fixed values when creating the VectorStoreRetriever ?

### DIFF
--- a/libs/community/langchain_community/vectorstores/cassandra.py
+++ b/libs/community/langchain_community/vectorstores/cassandra.py
@@ -24,7 +24,7 @@ if typing.TYPE_CHECKING:
 
 from langchain_core.documents import Document
 from langchain_core.embeddings import Embeddings
-from langchain_core.vectorstores import VectorStore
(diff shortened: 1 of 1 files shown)


eyurtsev (approved):

eyurtsev:
> NB: An alternative would have been to modify as_retriever in Vectorstore. But there's probably a good reason these were not exposed in the first place ? Is it because implementations may decide to not support them and have fixed values when creating the VectorStoreRetriever ?

Likely no one attempted to standardize the interface and the code evolved organically

PR #22475, opened 2024-06-04
community: Remove MiniMax(embeddings) group_id

maang-h:
- **Description:** Remove `group_id` in MimimaxChat and MiniMaxEmbeddings
    -  MimimaxChat 
    Remove reading of environment variables -- `MINIMAX_API_KEY`
    - MiniMaxEmbeddings
        -  Remove reading of environment variables -- `MINIMAX_API_KEY`
        - Update API request parameters


libs/community/langchain_community/chat_models/minimax.py
```diff
@@ -127,7 +127,6 @@ def _default_params(self) -> Dict[str, Any]:
             **self.model_kwargs,
         }
 
-    _client: Any
```
baskaryan, line 130: why this change

maang-h: Previously I thought that `_client` was not used in this code, but now I think it may be used as a connection object in the future to reduce the number of socket connections (no need to recreate the socket connection every time `invoke`)----If this is necessary, I will submit a PR later to complete this function?

maang-h: `_client` is not used here, so I deleted it.

PR #22479, opened 2024-06-04
community: Add huaweicloud css vector store

ElvisLiu1994:
Description:
Add HuaweiCloud CSSVectorSearch in langchain vectorstores


libs/community/langchain_community/vectorstores/huaweicloud_vector_search.py
```diff
@@ -0,0 +1,487 @@
...
+from langchain_core.embeddings import Embeddings
+from langchain_core.vectorstores import VectorStore
+
+if TYPE_CHECKING:
+    from elasticsearch import Elasticsearch
+
+logger = logging.getLogger(__name__)
+
+
+class CSSVectorStore(VectorStore):
```
baskaryan, line 25: can we call this something like HuaweiElasticsearchVectorStore? CSS usually means something unrelated to vector stores

baskaryan: can we call is HuaweiCSSVectorStore? not obvious to an average user that this would refer to a huawei service

ElvisLiu1994: thanks for your advice, I've renamed it

PR #22512, opened 2024-06-04
core[patch]: async support for runnable with message history

eyurtsev:
Async support for runnable with message history


eyurtsev:
Can likely leverage async root tracer here: https://github.com/langchain-ai/langchain/pull/22151

libs/core/tests/unit_tests/runnables/test_history.py
```diff
@@ -404,3 +405,48 @@ def get_session_history(user_id: str, conversation_id: str) -> ChatMessageHistor
...
+
+    def get_session_history(session_id: str, **kwargs: Any) -> AsyncChatMessageHistory:
+        if session_id not in chat_history_store:
+            chat_history_store[session_id] = AsyncChatMessageHistory()
+        return chat_history_store[session_id]
+
+    return get_session_history
+
+
+async def test_async_input_messages() -> None:
```
eyurtsev, line 440: Demonstrates that test currently fails

libs/core/langchain_core/runnables/history.py
```diff
@@ -306,8 +306,23 @@ def get_session_history(
             history_chain = RunnablePassthrough.assign(
                 **{messages_key: history_chain}
             ).with_config(run_name="insert_history")
-        bound = (
+        bound_sync = (
```
eyurtsev, line 309: This patter is overly complex. 
It swaps between sync and async depending on whether the invoke was .invoke or .ainvoke

It may be easier to accept a parameter in the RunnableWithMessageHistory to specify whether we're using sync or async methods for the store

Issue #22510, opened 2024-06-04
HuggingfacePipeline - ValueError: alternative_import must be a fully qualified module path


ErikTromp:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```# Directly from the documentation
from langchain_community.llms.huggingface_pipeline import HuggingFacePipeline
from transformers import AutoModelForCausalLM, AutoTokenizer, pipeline

model_id = 'meta-llama/Meta-Llama-3-8B-Instruct'
tokenizer = AutoTokenizer.from_pretrained(model_id)
model = AutoModelForCausalLM.from_pretrained(model_id)
pipe = pipeline("text-generation", model=model, tokenizer=tokenizer, max_new_tokens=10)
hf = HuggingFacePipeline(pipeline=pipe)```

### Error Message and Stack Trace (if applicable)

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/e/anaconda3/envs/rag/lib/python3.11/site-packages/langchain_core/_api/deprecation.py", line 182, in warn_if_direct_instance
    emit_warning()
  File "/home/e/anaconda3/envs/rag/lib/python3.11/site-packages/langchain_core/_api/deprecation.py", line 119, in emit_warning
    warn_deprecated(
  File "/home/e/anaconda3/envs/rag/lib/python3.11/site-packages/langchain_core/_api/deprecation.py", line 345, in warn_deprecated
    raise ValueError("alternative_import must be a fully qualified module path")
ValueError: alternative_import must be a fully qualified module path

### Description

Documentation shows how to use HuggingFacePipeline but using that code leads to an error. HF Pipeline can no longer be used.

### System Info

Windows
Langchain 0.2

Aillian:
temp fix
`pip install langchain-community==0.2.1 langchain-core==0.2.1`

ErikTromp:
I was already on 0.2.1 so that isn't the solution, the docs are just wrong. What did work was:
- Install langchain-huggingface==0.0.2
- Change the imports:
```from langchain_huggingface.embeddings import HuggingFaceEmbeddings
from langchain_huggingface.llms import HuggingFacePipeline```

[...]

PR #22513, opened 2024-06-04
core[patch]: fix validation of @deprecated decorator

eyurtsev:
This PR moves the validation of the decorator to a better place to avoid creating bugs while deprecating code.

Prevent issues like this from arising: https://github.com/langchain-ai/langchain/issues/22510

we should replace with a linter at some point that just does static analysis


eyurtsev:
Prevent issues like this from arising: https://github.com/langchain-ai/langchain/issues/22510

eyurtsev:
community tests should fail with this PR probably (until the deprecated decorator is updated in all places)

PR #22515, opened 2024-06-04
partner: add optimum intel ipex backend to huggingface pipelines

rbrugaro:
Description: Add optimum intel ipex backend to huggingface pipelines
Dependencies: optimum.intel.ipex
Example notebook showing its use: /docs/docs/integrations/llms/huggingface_pipelines.ipynb

- [ X] **Lint and test**: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified. See contribution guidelines for more: https://python.langchain.com/docs/contributing/


libs/partners/huggingface/langchain_huggingface/llms/huggingface_pipeline.py
```diff
@@ -125,10 +126,41 @@ def from_model_id(
...
+                        config = AutoConfig.from_pretrained(model_id)
+                        export = not getattr(config, "torchscript", False)
+                    except RuntimeError:
+                        logger.warning(
+                            "We will use IPEXModel with export=True to export the model"
+                        )
+                        export = True
+                    model = IPEXModelForCausalLM.from_pretrained(
+                        model_id,
+                        export=export,
```
jiqing-feng, line 154: The IPEXModel only support export=True for now, it means we will load the model in HF and do ipex jit trance.

rbrugaro: @jiqing-feng is your recommendation to remove the export setting logic or leave it even if for now it will render export=True? I followed same logic as in[ /optimum/intel/pipelines/pipeline_base.py](https://github.com/huggingface/optimum-intel/blob/2c79d987b685ef5e87b2062518598e9c1cf64c78/optimum/intel/pipelines/pipeline_base.py#L117 )

jiqing-feng: I see, yes you can detect export in this way.

rbrugaro:
@efriis can you please help review and provide any feedback? Thx!

rbrugaro:
hi @efriis, this is my first time contributing to langchain so I am trying to understand the process. is there anything else i can do now to help get this PR merged? is there ETA to get it reviewed? thanks :)

efriis (dismissed):
lgtm! cc @Jofthomas for a little review on the huggingface side if they'd prefer to keep alternative backends in community as opposed to the partner package

rbrugaro:
Thanks for the reviews! @efriis @Jofthomas 
Please allow me to share my design thoughts: 
1) having these optimized backends integrated into HF Pipelines is the best way to benefit more Hugging Face users since these are optimimum-intel optimizations are only useful for HF models/pipelines users and it would be unlikely/counterintuitive that users think of finding other HF pipeline implementations in a different package than Hugging Face. 
2) In the current implementation (which already supports openvino backend) and in this PR, if user wants to use an optimized backend they get a message to install the optimum[ipex/openvino] package to their environment so the HF partner package remains light for other users. We are not adding anything into the pyproject.toml, just some additional lines of code. 
Please let me know your thoughts!
cc @yao-matrix @kding1

yao-matrix:
> Thanks for the reviews! @efriis @Jofthomas Please allow me to share my design thoughts:
> 
> 1. having these optimized backends integrated into HF Pipelines is the best way to benefit more Hugging Face users since these are optimimum-intel optimizations are only useful for HF models/pipelines users and it would be unlikely/counterintuitive that users think of finding other HF pipeline implementations in a different package than Hugging Face.
> 2. In the current implementation (which already supports openvino backend) and in this PR, if user wants to use an optimized backend they get a message to install the optimum[ipex/openvino] package to their environment so the HF partner package remains light for other users. We are not adding anything into the pyproject.toml, just some additional lines of code.
>    Please let me know your thoughts!
>    cc @yao-matrix @kding1

@Jofthomas. Yes, we really seek a design for such cases. Why we are doing so is because we can see optimum-intel openvino backend already integrated into hf partner package, based on this fact, we are thinking your design is using HuggingFacePipeline as HF's contract interface to langchain, and make optimum-family hw accelerated HF libraries as soft plug-in backends. That's our takeaway by interpreting the code.

Jofthomas:
The HuggingFace pipeline class in community is the same as this one ATM. Futhermore, I can't guarantee all pipeline parameters will work as expected with alternative models. 

But, we can try to integrate it.

overall looks good, but the logic here is getting messy. would It be possible to move the backend logic in a new utils.py or utils folder ?

rbrugaro:
@Jofthomas, please have a look at my last commit, I moved the logic to load model based on task and backend to a utils.py. Is this what you had in mind to simplify the logic in huggingface_pipeline.py? Please let me know if you have any feedback!

efriis:
Marking this as draft until @Jofthomas indicates huggingface's preference here!

In general, I don't think a specific intel backend belongs in huggingface's partner package, but this is a shortcoming of too many things existing under the current partner package.

efriis:
closing without news - will let HF figure out where they want this directly

PR #22521, opened 2024-06-05
infra: recursive url loader unit tests

isahers1:


baskaryan (approved):
nice! cc @eyurtsev for thoughts

to fix CI just need to run format / lint:
```bash
cd libs/community
poetry install --with lint,typing
make format
make lint
```

libs/community/tests/unit_tests/document_loaders/test_recursive_url_loader.py
```diff
@@ -0,0 +1,91 @@
+import aiohttp
+import requests_mock
```
eyurtsev, line 2: I think `responses` is a more common library for mocking requests

Do we have both in the code base? https://github.com/getsentry/responses

libs/community/tests/unit_tests/document_loaders/test_recursive_url_loader.py
```diff
@@ -0,0 +1,91 @@
...
+    assert len(docs) == 1
+
+
+def test_sync_default_depth():
+    loader = RecursiveUrlLoader("http://test.com")
+    docs = mock_requests(loader)
+    assert len(docs) == 3
+
+
+def test_async_default_depth(mocker):
```
eyurtsev, line 74: This is using use_async=True, but testing in sync code.  

If there's `aload` defined could we also add async tests

``` python
async test_async_load():
  ...
  docs =  await loader.aload()
  assert docs == ...
```

eyurtsev:
Looks pretty good thanks for adding tests!

libs/community/tests/unit_tests/document_loaders/test_recursive_url_loader.py
baskaryan: yea we already have requests-mock as test dep. no preference for which we use

libs/community/tests/unit_tests/document_loaders/test_recursive_url_loader.py
```diff
@@ -0,0 +1,99 @@
...
+    ) -> None:
+        pass
+
+    async def __aenter__(self) -> MockGet:
+        return self
+
+
+@pytest.mark.parametrize(("max_depth", "expected_docs"), [(1, 1), (2, 3), (3, 4)])
+@pytest.mark.parametrize("use_async", [False, True])
+def test_lazy_load(
```
baskaryan, line 49: @nfcampos any idea why this test is failing with use_async=True in CI but passing for me locally
```bash
max_depth = 2, expected_docs = 3, use_async = True

    @pytest.mark.parametrize(("max_depth", "expected_docs"), [(1, 1), (2, 3), (3, 4)])
    @pytest.mark.parametrize("use_async", [False, True])
    def test_lazy_load(
        mocker: Any, max_depth: int, expected_docs: int, use_async: bool
    ) -> None:
>       loader = RecursiveUrlLoader(fake_url, max_depth=max_depth, use_async=use_async)

tests/unit_tests/document_loaders/test_recursive_url_loader.py:52: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
langchain_community/document_loaders/recursive_url_loader.py:173: in __init__
    self._lock = asyncio.Lock() if self.use_async else None
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/asyncio/locks.py:81: in __init__
    self._loop = events.get_event_loop()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <asyncio.unix_events._UnixDefaultEventLoopPolicy object at 0x7f7ecac39340>

    def get_event_loop(self):
        """Get the event loop for the current context.
    
        Returns an instance of EventLoop or raises an exception.
        """
        if (self._local._loop is None and
                not self._local._set_called and
                threading.current_thread() is threading.main_thread()):
            self.set_event_loop(self.new_event_loop())
    
        if self._local._loop is None:
>           raise RuntimeError('There is no current event loop in thread %r.'
                               % threading.current_thread().name)
E           RuntimeError: There is no current event loop in thread 'MainThread'.

/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/asyncio/events.py:642: RuntimeError
```

baskaryan: and potentially only failing on second, third test cases with use_async=True but not the first

PR #22523, opened 2024-06-05
core: throw error on invalid alternative_import in deprecated

efriis:
make the code fail on `@deprecated` instead of when actually throwing a warning


eyurtsev:
see https://github.com/langchain-ai/langchain/pull/22513 

you think remove entirely so we never throw due to typos at run time? (makes sense to me)

PR #22530, opened 2024-06-05
fix: fix bug load jsonl with JSONLoader

pphuc25:
- [ x] **PR title**: "community: fix bug cannot load jsonl file with JSONLoader"

- [ x] **PR message**: ***Delete this entire checklist*** and replace with
    - **Description:** I changed the way of reading file JSONL, in which a not line by line file can also be read
    - **Issue:** JSONDecodeError: Expecting value: line 1 column 1 (char 0), when loading JSON

I would like to cc @efriis and @vbarda to review my PR

### DIFF
--- a/libs/community/langchain_community/document_loaders/json_loader.py
+++ b/libs/community/langchain_community/document_loaders/json_loader.py
@@ -71,21 +71,20 @@ class JSONLoader(BaseLoader):
         """Load and return documents from the JSON file."""
         index = 0
         if self._json_lines:
-            with self.file_path.open(encoding="utf-8") as f:
-                for line in f:
-                    line = line.strip()
-                    if line:
-                        for doc in self._parse(line, index):
-                            yield doc
-                            index += 1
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/document_loaders/json_loader.py
```diff
@@ -71,21 +71,20 @@ def lazy_load(self) -> Iterator[Document]:
...
         index = 0
         if self._json_lines:
-            with self.file_path.open(encoding="utf-8") as f:
-                for line in f:
-                    line = line.strip()
-                    if line:
-                        for doc in self._parse(line, index):
-                            yield doc
-                            index += 1
+            f_json = json.loads(self.file_path.read_text())
```
eyurtsev, line 74: json lines is a streaming variant of json. One cannot apply json.loads to a file that's encoded as JSONL

eyurtsev (changes requested):
The fix is not correct. A corresponding test case should be added for a bug fix

PR #22531, opened 2024-06-05
community[patch]: improve test setup to accurately test filtering of labels in neo4j

leila-messallem:
**Description:** This PR addresses an issue with an existing test that was not effectively testing the intended functionality. The previous test setup did not adequately validate the filtering of the labels in neo4j, because the nodes and relationship in the test data did not have any properties set. Without properties these labels would not have been returned, regardless of the filtering.


libs/community/tests/integration_tests/graphs/test_neo4j.py
```diff
@@ -266,12 +266,17 @@ def test_neo4j_filtering_labels() -> None:
     # Remove all constraints
     graph.query("CALL apoc.schema.assert({}, {})")
     graph.query(
-        "CREATE (:`_Bloom_Scene_`)-[:_Bloom_HAS_SCENE_]->(:`_Bloom_Perspective_`)"
+        """
+        CREATE (:_Bloom_Scene_{property_a: 'a'})
+        -[:_Bloom_HAS_SCENE_{property_b: 'b'}]
```
oskarhane, line 271: I think we want to add a space between the label/type and the property map for clarity (and I'm actually surprised this works).

```suggestion
        CREATE (:_Bloom_Scene_ {property_a: 'a'})
        -[:_Bloom_HAS_SCENE_ {property_b: 'b'}]
```

oskarhane (approved):
LGTM

PR #22555, opened 2024-06-05
community: support Databricks Unity Catalog functions as LangChain tools

mengxr:
This PR adds support for using Databricks Unity Catalog functions as LangChain tools, which runs inside a Databricks SQL warehouse.

* An example notebook is provided.


libs/community/langchain_community/tools/databricks/tool.py
```diff
@@ -0,0 +1,230 @@
...
+    def __init__(
+        self,
+        function_name: str,
+        warehouse_id: str,
+        workspace_client: Optional[WorkspaceClient] = None,
+        **kwargs: Any,
+    ):
+        if workspace_client is None:
+            workspace_client = WorkspaceClient()
+        function = workspace_client.functions.get(function_name)
```
efriis, line 139: Can we move this loading logic into the `UCFunctionTools.get_tool` factory function and make the init method just a setter?

efriis: I would also be supportive of not implementing `UCFunctionTool` as a class and instead having it be a function that returns `Tool(...)` e.g. https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/agent_toolkits/load_tools.py#L180

I think having a single entry point of UCFunctionTools is more intuitive

efriis: As a quick fix could also just prefix this class with an underscore, so we're more flexible to change it later

libs/community/langchain_community/tools/databricks/execution.py
efriis, line 1: Can we rename this file `_execution` to make it clear users aren't using these directly?

mengxr:
@efriis Thanks for the review! This PR is ready for another pass.

efriis (approved):

PR #22365, opened 2024-05-31
huggingface: Skip Login to HuggubgFaceHub when token is not set 

mirkenstein:
The changes in this PR are identical to the ones applied in the the community library  https://github.com/langchain-ai/langchain/pull/21561
. 

- [ ] **HuggingFaceEndpoint**: "Skip Login to HuggingFaceHub when hf token is not set"
  - Where:  langchain, parthners, llm, huggingface_endpoint
 

- [ ] **PR message**: ***Delete this entire checklist*** and replace with
    - **Description:**  Skip login to huggingface hub when when `huggingfacehub_api_token` is not set. This is needed when using custom `endpoint_url` outside of HuggingFaceHub.
    - **Issue:** it fixes https://github.com/langchain-ai/langchain/issues/20342 and https://github.com/langchain-ai/langchain/issues/19685
    - **Related PR :** https://github.com/langchain-ai/langchain/pull/21561
    - **Dependencies:** None

- [ ] **Add tests and docs**: 
  1. Tested with locally available TGI endpoint
  2.  Example Usage
```python
from langchain_huggingface.llms import HuggingFaceEndpoint

llm = HuggingFaceEndpoint(
    endpoint_url='http://localhost:8080',
    server_kwargs={
        "headers": {"Content-Type": "application/json"}
    }
)
resp = llm.invoke("Tell me a joke")
print(resp)
```
 Also tested against HF Endpoints
 ```python
from langchain_huggingface.llms import HuggingFaceEndpoint

huggingfacehub_api_token = "hf_xyz"
repo_id = "mistralai/Mistral-7B-Instruct-v0.2"
llm = HuggingFaceEndpoint(
    huggingfacehub_api_token=huggingfacehub_api_token,
    repo_id=repo_id,
)
resp = llm.invoke("Tell me a joke")
print(resp)
 ```
Additional guidelines:
- Make sure optional dependencies are imported within a function.
- Please do not add dependencies to pyproject.toml files (even optional ones) unless they are required for unit tests.
- Most PRs should not touch more than one package.
- Changes should be backwards compatible.
- If you are adding something to community, do not re-import it in langchain.

If no one reviews your PR within a few days, please @-mention one of baskaryan, efriis, eyurtsev, hwchase17.

PR #22558, opened 2024-06-05
community:removed unnecessary authenticate or fail

JeanBaptiste-dlb:
Authentication should not be forced here, as the client does not require authentication when running in local.

**Description:** When using HuggingFace client with a local deployment of a hf text gen inference server, it is not mandatory to provide a huggingface_api_token, I am aware my proposed fix is a bit dirty, but I want to propose the idea that the code should not just raise an exception if there is no login to huggingface.

### DIFF
--- a/libs/community/langchain_community/llms/huggingface_endpoint.py
+++ b/libs/community/langchain_community/llms/huggingface_endpoint.py
@@ -179,7 +179,7 @@ class HuggingFaceEndpoint(LLM):
             )
             login(token=huggingfacehub_api_token)
         except Exception:
-            huggingfacehub_api_token=None
+            huggingfacehub_api_token = None
 
         from huggingface_hub import AsyncInferenceClient, InferenceClient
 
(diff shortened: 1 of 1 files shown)


isahers1:
This is deprecated - should I ask them to edit the partner lib? @baskaryan 
Also is this not what the guy wants: https://python.langchain.com/v0.2/docs/integrations/llms/huggingface_pipelines/ ? Not sure if using HuggingFaceEndpoint is necessary.

baskaryan:
> This is deprecated - should I ask them to edit the partner lib? @baskaryan Also is this not what the guy wants: https://python.langchain.com/v0.2/docs/integrations/llms/huggingface_pipelines/ ? Not sure if using HuggingFaceEndpoint is necessary.

yea this is a deprecated version of the integration, main implementation lives in libs/partners/huggingface and was update in https://github.com/langchain-ai/langchain/pull/22365. We can still update the community version if we'd like but could we copy the solution in #22365?

JeanBaptiste-dlb:
Hello, sure we can close, I will use the partner version

PR #22559, opened 2024-06-05
core: correctly format ImagePromptTemplate when used with a _StringImageMessagePromptTemplate (e.g. HumanMessagePromptTemplate)

thdesc:
- Description: When formatting an `ImagePromptTemplate`, only the `url`,  `path` and `detail` variables are used (see the [format](https://github.com/langchain-ai/langchain/blob/3280a5b49b0cd997129d341d1ecc97515593f224/libs/core/langchain_core/prompts/image.py#L45) method). However `inputs` (defined [here](https://github.com/langchain-ai/langchain/blob/3280a5b49b0cd997129d341d1ecc97515593f224/libs/core/langchain_core/prompts/chat.py#L551)) cannot have any of these values, since `url`,  `path` and `detail` are not allowed as `input_variables` in `ImagePromptTemplate` (see [here](https://github.com/langchain-ai/langchain/blob/3280a5b49b0cd997129d341d1ecc97515593f224/libs/core/langchain_core/prompts/image.py#L20)).
By passing the `kwargs` directly, it would allow the correct formatting of the `ImagePromptTemplate`. Here is an example of how it would work: 
```python
from langchain_core.prompts import HumanMessagePromptTemplate, PromptTemplate

human_message_prompt_template = HumanMessagePromptTemplate(
    prompt=[PromptTemplate.from_template("Describe the following image"), ImagePromptTemplate()]
)
print(human_message_prompt_template.format(url="https://example.com/image.jpg", detail="auto"))
# Output:
# HumanMessage(content=[{'type': 'text', 'text': 'describe the following image'}, {'type': 'image_url', 'image_url': {'url': 'https://example.com/image.jpg', 'detail': 'auto'}}])
print(human_message_prompt_template.format(path="path/to/image.png", detail="auto"))
# Output:
# HumanMessage(content=[{'type': 'text', 'text': 'describe the following image'}, {'type': 'image_url', 'image_url': {'url': 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAABQAAAB......QAAIABJRE', 'detail': 'auto'}}])
```
- Issues: 
    - fixes #20820
    - related to #22024

- Dependencies: None

- Twitter handle: @DeschampsTho5


thdesc:
At the end, this PR is not useful. I have created another one that is related to the same problem: #22613

thdesc:
At the end, this PR is not useful. I have created another one that is related to the same problem: #22613

PR #22563, opened 2024-06-05
core[minor]: Add parent_ids to astream_events API

eyurtsev:
Include a list of parent ids for each event in astream events.


libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -105,6 +106,22 @@ def __init__(
         self.send_stream = memory_stream.get_send_stream()
         self.receive_stream = memory_stream.get_receive_stream()
 
+    def _get_parent_ids(self, run_id: UUID) -> List[str]:
+        """Get the parent IDs of a run (non-recursively) cast to strings."""
+        parent_ids = []
+
+        run_info = self.run_map.get(run_id)
```
nfcampos, line 113: lets us the dotted order string instead?

eyurtsev: created a parent map that should do essentially the same thing

nfcampos (approved):

Issue #22516, opened 2024-06-04
model_name in ChatResult from BaseChatOpenAI is not sourced from API response


bantucaravan:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
llm = ChatOpenAI(model_name="gpt-4-turbo")
aimessage = llm.invoke([('human', "say hello!!!")])
aimessage.response_metadata['model_name']
```

### Description

In the openai  API, you can specify a model by a generic identifier (e.g. "gpt-4-turbo") which will be matched to a specifc model version by openai for continuous upgrades. The specific model used is returned in the openai API response (see this documentation for details: https://platform.openai.com/docs/models/continuous-model-upgrades).

I would expect the the `model_name` in the `ChatResult.llm_output` returned from `BaseChatOpenAI` to show the specific model returned by the openai API. However, the model_name returned is whatever the model_name that was passed to `BaseChatOpenAI` is (which will often be the generic model name). This makes logging and observability for your invocations difficult. The problem is found here: https://github.com/langchain-ai/langchain/blob/cb183a9bf18505483d3426530cce2cab2e1c5776/libs/partners/openai/langchain_openai/chat_models/base.py#L584 when `self.model_name` is used to populate the model_name key instead of `response.get("model", self.model_name)`.

This should be a simple fix that greatly improves logging and observability.

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 23.2.0: Wed Nov 15 21:53:34 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T8103
> Python Version:  3.11.9 (main, Apr 19 2024, 11:43:47) [Clang 14.0.6 ]

Package Information
-------------------
> langchain_core: 0.2.3
> langchain: 0.2.1
> langsmith: 0.1.69
> langchain_anthropic: 0.1.15
> langchain_openai: 0.1.8
> langchain_text_splitters: 0.2.0

Packages not installed (Not Necessarily a Problem)
--------------------------------------------------
The following packages were not found:

> langgraph
> langserve

isatyamks:
[...]

PR #22569, opened 2024-06-05
openai, azure: update model_name in ChatResult to use name from API response

isatyamks:
The response.get("model", self.model_name) checks if the model key exists in the response dictionary. If it does, it uses that value; otherwise, it uses self.model_name.


isatyamks:
@efriis, please proceed with the merge if the changes are accurate and appropriate.

ccurme (approved):
This will change model names for aliases (e.g., "gpt-4o" -> "gpt-4o-2024-05-13") and for Azure ("gpt-35-turbo" -> "gpt-3.5-turbo").

I think this is an improvement but let's change the title before merging so the impact is clear to users, e.g., `openai, azure: update model_name in ChatResult to use name from API response`

bantucaravan:
resolves #22516

bantucaravan:
Thanks for fixing this and merging this so quickly, this was my first issue and it was so nice to see it acted on so fast @isatyamks @ccurme

PR #22573, opened 2024-06-05
multiple: add `stop` attribute

ccurme:


libs/partners/ai21/langchain_ai21/chat_models.py
```diff
@@ -97,6 +99,8 @@ def _default_params(self) -> Mapping[str, Any]:
             "top_k_return": self.top_k_return,
             "n": self.n,
         }
+        if self.stop:
+            base_params["stop_sequences"] = self.stop
```
baskaryan, line 103: unrelated but the check in L126 doesn't do anything, it couldn't be in kwargs

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -492,6 +492,9 @@ class Config:
     max_retries: int = 2
     """Number of retries allowed for requests sent to the Anthropic Completion API."""
 
+    stop: Optional[List[str]] = Field(None, alias="stop_sequences")
```
baskaryan, line 495: can we make the attribute stop_sequences and the alias stop? since stop_sequences is the Anthropic name for it

libs/partners/ai21/langchain_ai21/chat_models.py
```diff
@@ -35,6 +35,8 @@ class ChatAI21(BaseChatModel, AI21Base):
         You can view the options at https://github.com/AI21Labs/ai21-python?tab=readme-ov-file#model-types"""
     num_results: int = 1
     """The number of responses to generate for a given prompt."""
+    stop: Optional[List[str]] = None
```
baskaryan, line 38: same comment about making this stop_sequences and alias stop

ccurme: missed your comments, will add in another PR

PR #22576, opened 2024-06-05
[Core] Unified Enable/Disable Tracing

hinthornw:


libs/core/tests/unit_tests/runnables/test_tracing_interops.py
```diff
@@ -168,3 +169,31 @@ async def my_parent_function(a: int) -> int:
...
+
+    def my_func(a: int) -> int:
+        return a + 1
+
+    env_on = env == "true"
+    with patch.dict("os.environ", {"LANGSMITH_TRACING": env}):
+        with tracing_context(enabled=enabled):
+            RunnableLambda(my_func).invoke(1)
+
+        time.sleep(0.3)
```
agola11, line 190: rather than use time.sleep in unit tests, should we just turn temporarily turn off auto-batching / use `wait_for_all_tracers`?

hinthornw: ya

agola11 (approved):

PR #22579, opened 2024-06-05
docs: `how-to` index page update

leo-gan:
Removed repeated "How to: ". That makes the [page more readable](https://langchain-9jq1kef7i-langchain.vercel.app/v0.2/docs/how_to/).

### DIFF
--- a/docs/docs/how_to/index.mdx
+++ b/docs/docs/how_to/index.mdx
@@ -13,16 +13,16 @@ For comprehensive descriptions of every class and function see the [API Referenc
 
 ## Installation
 
-- [How to: install LangChain packages](/docs/how_to/installation/)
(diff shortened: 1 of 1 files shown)


eyurtsev:
Suspect this was done intentionally by @hwchase17, probably modeled after: https://docs.djangoproject.com/en/5.0/howto/

leo-gan:
> Suspect this was done intentionally by @hwchase17, probably modeled after: https://docs.djangoproject.com/en/5.0/howto/

@eyurtsev Sure, it was done intentionally. But on this scale, it makes reading comprehension very hard. Just look at both versions and try to quickly comprehend the text.

efriis:
This was discussed prior to the v0.2 docs launch and we decided on keeping the "how to:"! Your feedback is noted.

PR #22586, opened 2024-06-06
Fix a bug with the parameter, path, in langchain_qdrant.vectorstores.…

w22l:
…Qdrant.from_existing_collection.  Make the parameter "path" optional so that the function works as shown in the example: https://python.langchain.com/v0.2/docs/integrations/vectorstores/qdrant/

### DIFF
--- a/libs/partners/qdrant/langchain_qdrant/vectorstores.py
+++ b/libs/partners/qdrant/langchain_qdrant/vectorstores.py
@@ -1034,9 +1034,11 @@ class Qdrant(VectorStore):
             **kwargs,
         )
         embeddings = [
-            result.vector.get(self.vector_name)  # type: ignore[index, union-attr]
-            if self.vector_name is not None
-            else result.vector
+            (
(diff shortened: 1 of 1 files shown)


libs/partners/qdrant/langchain_qdrant/vectorstores.py
```diff
@@ -1353,8 +1357,8 @@ def from_texts(
     def from_existing_collection(
         cls: Type[Qdrant],
         embedding: Embeddings,
-        path: str,
         collection_name: str,
+        path: Optional[str] = None,
```
baskaryan, line 1361: hm this is a breaking change (swapping the order of positional args). an alternative would be to give both path and collection_name defaults and just manually check that collection_name is passed in
```python
  path: Optional[str] = None,
  collection_name: Optional[str] = None,
  ...
  if collection_name is None:
    raise ValueError("Must specify collection_name, received None").
```

baskaryan:
landed in #21875! thank you for the pr anyways!

PR #22595, opened 2024-06-06
[Community] [Text Splitters]: Fix TEI support and html section splitter

plageon:
**[Community]: Fix HuggingFaceHubEmbeddings** 
Modified `langchain_community.embeddings.HuggingFaceHubEmbeddings`, where the API call function does not match current [Text Embeddings Inference API](https://huggingface.github.io/text-embeddings-inference/#/Text%20Embeddings%20Inference/embed).
One example is:
```json
{
  "inputs": "string",
  "normalize": true,
  "truncate": false
}
```
Parameters in `_model_kwargs` are not passed properly in the latest version. By the way, the issue *[why cause 413? #50](https://github.com/huggingface/text-embeddings-inference/issues/50)* might be solved.

**[Text Splitters]: Fix HTMLSectionSplitter** 
Modified `langchain_text_splitters.HTMLSectionSplitter`, where in the latest version `dict` data structure is used to store sections from a html document, in function `split_html_by_headers`. The header/section element names serve as dict keys. This can be a problem when duplicate header/section element names are present in a single html document. Latter ones can replace former ones with the same name. Therefore some contents can be miss after html text splitting is conducted.

Using a list to store sections can hopefully solve the problem.

### DIFF
--- a/libs/community/pyproject.toml
+++ b/libs/community/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "langchain-community"
-version = "0.2.3"
+version = "0.2.4"
 description = "Community contributed LangChain integrations."
 authors = []
 license = "MIT"


libs/text-splitters/langchain_text_splitters/html.py
baskaryan, line 1: could we split this into two prs? afaict the two changes aren't related right

libs/community/langchain_community/embeddings/huggingface_hub.py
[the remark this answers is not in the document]
plageon: New PR: https://github.com/langchain-ai/langchain/pull/22653

libs/text-splitters/langchain_text_splitters/html.py
plageon: New PRs: 

1. https://github.com/langchain-ai/langchain/pull/22653
2. https://github.com/langchain-ai/langchain/pull/22654

PR #22601, opened 2024-06-06
[Partner] trigger `stop` option for HF pipeline

openvino-dev-samples:
Since currently `stop` option is not working in HF pipeline, this PR will help to split the sentence according to `stop` list.

### DIFF
--- a/libs/community/langchain_community/llms/huggingface_pipeline.py
+++ b/libs/community/langchain_community/llms/huggingface_pipeline.py
@@ -10,6 +10,8 @@ from langchain_core.language_models.llms import BaseLLM
 from langchain_core.outputs import Generation, LLMResult
 from langchain_core.pydantic_v1 import Extra
 
+from langchain_community.llms.utils import enforce_stop_tokens
+
 DEFAULT_MODEL_ID = "gpt2"
 DEFAULT_TASK = "text-generation"
 VALID_TASKS = (
@@ -297,6 +299,8 @@ class HuggingFacePipeline(BaseLLM):
                     )
                 if skip_prompt:
                     text = text[len(batch_prompts[j]) :]
+                if stop is not None:
+                    text = enforce_stop_tokens(text, stop)
                 # Append the processed text to results
                 text_generations.append(text)
 
--- a/libs/partners/huggingface/langchain_huggingface/llms/huggingface_pipeline.py
+++ b/libs/partners/huggingface/langchain_huggingface/llms/huggingface_pipeline.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import importlib.util
 import logging
+import re
 from typing import Any, List, Mapping, Optional
 
 from langchain_core.callbacks import CallbackManagerForLLMRun
@@ -293,6 +294,8 @@ class HuggingFacePipeline(BaseLLM):
                     )
                 if skip_prompt:
                     text = text[len(batch_prompts[j]) :]
+                if stop is not None:
+                    text = re.split("|".join(stop), text, maxsplit=1)[0]
                 # Append the processed text to results
                 text_generations.append(text)
 


OpenVINO-dev-contest:
@baskaryan pls help to check

baskaryan (commented):
@Jofthomas is there any native support for stop sequences in pipelines

OpenVINO-dev-contest:
This method will not truly stop generation, but it will help to split the output and remove the stop word.

To truly stop generation, we can pass `stop token` to hf pipeline, however is this case the stop word will still be added in output.

baskaryan:
> This method will not truly stop generation, but it will help to split the output and remove the stop word.
> 
> To truly stop generation, we can pass `stop token` to hf pipeline through `pipeline_kwargs`, however in this case the stop word will still be added in output.

think passing to pipeline makes sense if it supports that

OpenVINO-dev-contest:
Yes, we can pass the `stop token` without using `stop` params, but  `stop` params can help to remove last token from output.

Jofthomas:
Yes, I think the better option is to convert the list of stops into a StoppingCriteria.
See Arthur's comment on how to leverage StoppingCriteria in pipeline:
https://github.com/huggingface/transformers/issues/26280#issuecomment-1737163306

And it is possible to implement custom StoppingCriteria like I did in this educational colab :
https://colab.research.google.com/drive/1j_vsc28FwZEDocDxVxWJ6Fvxd18FK8Gl#scrollTo=a_w-gazERC0_

With both solution it is possible to implement working stops. Hope it helps

OpenVINO-dev-contest:
> Yes, I think the better option is to convert the list of stops into a StoppingCriteria. See Arthur's comment on how to leverage StoppingCriteria in pipeline: [huggingface/transformers#26280 (comment)](https://github.com/huggingface/transformers/issues/26280#issuecomment-1737163306)
> 
> And it is possible to implement custom StoppingCriteria like I did in this educational colab : https://colab.research.google.com/drive/1j_vsc28FwZEDocDxVxWJ6Fvxd18FK8Gl#scrollTo=a_w-gazERC0_
> 
> With both solutions it is possible to implement working stops. Hope it helps

The logic is different. The purpose of this PR is to keep the words before `stop word` (remove the stop word from output), instead of stopping generation.
`StoppingCriteria` can be passed through another attributes which is `pipeline_kwargs`

Jofthomas:
Yes I understood your point. 

Indeed StoppingCriteria can already be passed. 
The definition of StoppingCriteria is very free, hence nothing is stopping you from removing the stop tokens in the __call__of your custom StoppingCriteria.

Still, I think your raised a valid point in stating we should implement stops in the pipeline. 
But if we do implement it should come from a StoppingCriteriaList instead of a basic split.
( I forwarded this PR to transformers for their opinion so let's see )

And about keeping the stop in the generated sequence or not, we should refer to what's standard practice in other langchain LLM's. What's the standard @baskaryan ?

zucchini-nlp:
Hey there! The StoppingCriteria is nice to handle cases when you need to stop after a certain condition is triggered. But if you want to stop before the condition (i.e. remove the stop token thus before it's added to generated text), StoppingCriteria will not be a fit.

In that case it's nice to do a normal stopping criteria and stop after the condition, then postprocess the text by removing the last tokens. This case is nice if you don't want to waste compute generating beyond the custom stop_token, so that you can stop as soon as it's complete in single generations.

In transformers we have an option to pass `stop_string` to generation which will trigger when one of the stop string is generated. See the [docs](https://huggingface.co/docs/transformers/en/internal/generation_utils#transformers.StopStringCriteria) for more details. I believe that is already available through pipeline, if you pass it along with other generation kwargs 😄

Jofthomas:
Thanks @zucchini-nlp,

I think it answers our question, we should rather pass the langchain stop as a stop_string ( was not aware of this parameter ).

And then decide if we keep the stop or not based on langchain's standard practice.

OpenVINO-dev-contest:
I suppose someone will help to standardize this approach, am I right ?
Then shall I close this PR ?

Jofthomas:
I would say you can close It.
I will help in implementing that next week.

I have multiple improvement to do in the library.

OpenVINO-dev-contest:
> I would say you can close It. I will help in implementing that next week.
> 
> I have multiple improvement to do in the library.

Sure. I'd like to track the progress of this improvement.
Thank you.

OpenVINO-dev-contest:
> I would say you can close It. I will help in implementing that next week.
> 
> I have multiple improvement to do in the library.

Hi  @Jofthomas, May I know if there is any progress on how to removing the last(stop) token in reponses ?  
When selecting `Observation` as stop token, It will cause 2 `Observation` flags generated in a ReAct agent's conversatio , for example:

```
Thought: ....
Action: .....
Observation: (stop token)
Observation: (True Observation flag by LangChain agent parser)...
···

Thank you.

PR #22627, opened 2024-06-06
community: update the default hf bge embeddings

JonZeolla:
**Description:** This updates the langchain_community > huggingface > default bge embeddings ([the current default recommends this change](https://huggingface.co/BAAI/bge-large-en))
**Issue:** None
**Dependencies:** None
**Twitter handle:** @jonzeolla


isahers1:
Lgtm @baskaryan

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -7,7 +7,7 @@
 
 DEFAULT_MODEL_NAME = "sentence-transformers/all-mpnet-base-v2"
 DEFAULT_INSTRUCT_MODEL = "hkunlp/instructor-large"
-DEFAULT_BGE_MODEL = "BAAI/bge-large-en"
+DEFAULT_BGE_MODEL = "BAAI/bge-large-en-v1.5"
```
baskaryan, line 10: updating the default model is technically a breaking change, could we stick to just updating all the docs?

JonZeolla: I'm fine with that for now but is there a ballpark plan for eventually doing an update?

baskaryan: in general we've moved away from having default values for the model arg precisely because of this (over time it either becomes outdated or you have to make a breaking change). 

in this case think best first step would be to always explicitly pass in model name in all the documentation, and in a future minor version bump we can remove the default altogether

JonZeolla: That makes sense to me; let me change this to a docs update and a deprecation warning

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -66,6 +66,15 @@ class HuggingFaceEmbeddings(BaseModel, Embeddings):
     def __init__(self, **kwargs: Any):
         """Initialize the sentence_transformer."""
         super().__init__(**kwargs)
+
+        if "model_name" not in kwargs:
+            warn_deprecated(
+                since="0.2.5",
+                removal="0.4.0",
```
JonZeolla, line 73: Open to feedback on the removal versions

libs/community/langchain_community/embeddings/huggingface.py
JonZeolla: @baskaryan this should be all set for another review; I kept the scope narrow to mostly just huggingface embeddings (and related docs) for this first go at it.

JonZeolla:
This hit some kind of issue with all the 3.12 tests added in https://github.com/langchain-ai/langchain/pull/22813 taking a long time; retrying...

JonZeolla:
That's better 🤷

baskaryan (approved):

PR #22635, opened 2024-06-06
updated oracleai_demo.ipynb

isatyamks:
The outer try/except block handles connection errors, and the inner try/except block handles SQL execution errors, providing detailed error messages for both.
try:
    conn = oracledb.connect(user=username, password=password, dsn=dsn)
    print("Connection successful!")

    cursor = conn.cursor()
    try:
        cursor.execute(
            """
            begin
                -- Drop user
                begin
                    execute immediate 'drop user testuser cascade';
                exception
                    when others then
                        dbms_output.put_line('Error dropping user: ' || SQLERRM);
                end;


ccurme (approved):

isatyamks:
thanks for merging my contribution.

PR #22637, opened 2024-06-06
core[patch]: Correctly order parent ids in astream events (from root to immediate parent), add defensive check for cycles

eyurtsev:
This PR makes two changes:

1. Fixes the order of parent IDs to be from root to immediate parent
2. Adds a simple defensive check for cycles


libs/core/langchain_core/tracers/event_stream.py
```diff
@@ -118,13 +118,20 @@ def __init__(
     def _get_parent_ids(self, run_id: UUID) -> List[str]:
         """Get the parent IDs of a run (non-recursively) cast to strings."""
         parent_ids = []
```
eyurtsev, line 120: list of parents ids isn't big so using list instead of deque

nfcampos (approved):

PR #22654, opened 2024-06-07
[Text Splitters]: Fix HTMLSectionSplitter

plageon:
Update former pull request: https://github.com/langchain-ai/langchain/pull/22595.

Modified `langchain_text_splitters.HTMLSectionSplitter`, where in the latest version `dict` data structure is used to store sections from a html document, in function `split_html_by_headers`. The header/section element names serve as dict keys. This can be a problem when duplicate header/section element names are present in a single html document. Latter ones can replace former ones with the same name. Therefore some contents can be miss after html text splitting is conducted.

Using a list to store sections can hopefully solve the problem.


baskaryan:
could we update unit tests to validate the change

plageon:
new PR updated: https://github.com/langchain-ai/langchain/pull/22812

PR #22662, opened 2024-06-07
core: runnables: special handling GeneratorExit because no error

ret2libc:
When using `stream()` with some parts of a chain being stream-able and others not, the results from the non-streamable parts are yielded by langchain, however when a `StopGeneration` exception is raised, this is collected as an error even though it is an "expected" exception.

Reproducer:
```python
from langchain_core.callbacks import StdOutCallbackHandler

llm = ChatOpenAI(model="gpt-4o")

class MyHandler(StdOutCallbackHandler):
    def on_chain_error(self, error: BaseException, *, run_id, parent_run_id = None, **kwargs):
        print("error: ", end="")
        print(str(error))

handler = MyHandler()

    ("user", "produce a poem"),
])


def get_f(x: str) -> str:
    return x

chain = prompt | llm | StrOutputParser() | get_f
chain = chain.with_config({'callbacks': [handler]})

print(next(chain.stream({})))
```

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -1891,6 +1891,8 @@ class Runnable(Generic[Input, Output], ABC):
                             final_input_supported = False
                 else:
                     final_input = ichunk
+        except GeneratorExit:
+            run_manager.on_chain_end(final_output, inputs=final_input)
         except BaseException as e:
             run_manager.on_chain_error(e, inputs=final_input)
             raise


baskaryan (commented):
cc @nfcampos @eyurtsev

libs/core/langchain_core/runnables/base.py
```diff
@@ -1889,6 +1889,8 @@ def _transform_stream_with_config(
                             final_input_supported = False
                 else:
                     final_input = ichunk
+        except GeneratorExit as e:
```
eyurtsev, line 1892:
```suggestion
        except GeneratorExit:
```

eyurtsev: @ret2libc would you be able to make this change? I'm unable to this fork

eyurtsev (approved):

eyurtsev:
Here's a unit test to catch that we can catch afterwards.. seems like an edge case that probably doesn't have many consequences?

```python
async def test_stream_error():
    from langchain_core.callbacks import BaseCallbackHandler
    from langchain_core.language_models.fake_chat_models import GenericFakeChatModel

    on_chain_error_triggered = False

    class MyHandler(BaseCallbackHandler):
        def on_chain_error(
            self, error: BaseException, *, run_id, parent_run_id=None, **kwargs
        ):
            nonlocal on_chain_error_triggered
            on_chain_error_triggered = True

    llm = GenericFakeChatModel(messages=iter(["hi there"]))
    chain = llm | StrOutputParser()
    chain = chain.with_config({"callbacks": [MyHandler()]})
    st = chain.stream("hello")
    next(st)
    st.close()
    # Wait for a bit to make sure that the callback is called.
    time.sleep(0.05)
    assert on_chain_error_triggered is False

```

ret2libc:
BTW I'm not sure why but by using the following pattern I was able to avoid errors showing up in Langfuse.


```python
def _reduce_output(x, y):
   return y

res = functools.reduce(_reduce_output, chain.stream({}), InitialObject())
```

eyurtsev:
> BTW I'm not sure why but by using the following pattern I was able to avoid errors showing up in Langfuse.
> 
> ```python
> def _reduce_output(x, y):
>    return y
> 
> res = functools.reduce(_reduce_output, chain.stream({}), InitialObject())
> ```

haven't used reduce in a while, but I think it consumes the entire iterable before garbage collecting it. the issue was appearing when the iterable is closed before being fully consumed

PR #22682, opened 2024-06-07
community[minor]: implement huggingface show_progress consistently

JonZeolla:
- **Description:** This implements `show_progress` more consistently (i.e. it is also added to the `HuggingFaceBgeEmbeddings` object).
 - **Issue:** This implements `show_progress` more consistently in the embeddings huggingface classes. Previously this could have been set via `encode_kwargs`.
 - **Dependencies:** None
 - **Twitter handle:** @jonzeolla

### DIFF
--- a/libs/community/langchain_community/embeddings/huggingface.py
+++ b/libs/community/langchain_community/embeddings/huggingface.py
@@ -1,7 +1,8 @@
+import warnings
 from typing import Any, Dict, List, Optional
 
 import requests
-from langchain_core._api import deprecated
+from langchain_core._api import deprecated, warn_deprecated
 from langchain_core.embeddings import Embeddings
 from langchain_core.pydantic_v1 import BaseModel, Extra, Field, SecretStr
 
@@ -154,6 +155,8 @@ class HuggingFaceInstructEmbeddings(BaseModel, Embeddings):
     """Instruction to use for embedding documents."""
     query_instruction: str = DEFAULT_QUERY_INSTRUCTION
     """Instruction to use for embedding query."""
+    show_progress: bool = False
+    """Whether to show a progress bar."""
 
     def __init__(self, **kwargs: Any):
         """Initialize the sentence_transformer."""
@@ -166,6 +169,20 @@ class HuggingFaceInstructEmbeddings(BaseModel, Embeddings):
             )
         except ImportError as e:
             raise ImportError("Dependencies for InstructorEmbedding not found.") from e
+
+        if "show_progress_bar" in self.encode_kwargs:
+            warn_deprecated(
+                since="0.2.5",
+                removal="0.4.0",
+                name="encode_kwargs['show_progress_bar']",
+                alternative=f"the show_progress method on {self.__class__.__name__}",
+            )
+            if self.show_progress:
+                warnings.warn(
+                    "Both encode_kwargs['show_progress_bar'] and show_progress are set;"
+                    "encode_kwargs['show_progress_bar'] takes precedence"
+                )
+            self.show_progress = self.encode_kwargs.pop("show_progress_bar")
 
     class Config:
         """Configuration for this pydantic object."""
@@ -182,7 +199,11 @@ class HuggingFaceInstructEmbeddings(BaseModel, Embeddings):
             List of embeddings, one for each text.
         """
         instruction_pairs = [[self.embed_instruction, text] for text in texts]
-        embeddings = self.client.encode(instruction_pairs, **self.encode_kwargs)
+        embeddings = self.client.encode(
+            instruction_pairs,
+            show_progress_bar=self.show_progress,
+            **self.encode_kwargs,
+        )
         return embeddings.tolist()
 
     def embed_query(self, text: str) -> List[float]:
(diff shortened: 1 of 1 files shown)


eyurtsev (approved):

eyurtsev:
@JonZeolla is this a breaking change for users that were passing it via encode_kwargs?

JonZeolla:
@eyurtsev Yes, it would break with the following:

```python
TypeError: sentence_transformers.SentenceTransformer.SentenceTransformer.encode() got multiple values for keyword argument 'show_progress_bar'
```

This should not be merged yet.

JonZeolla:
Is there a specific commit title (`community[major]:` or `!feat:`) to indicate the breaking change?

eyurtsev:
@JonZeolla currently `community[major]`. (we may adopt https://www.conventionalcommits.org/en/v1.0.0/ at a later stage later.)

But we try to avoid making breaking changes like these -- it makes the API a bit prettier, but can cause user code to break without a good reason. 

Instead if you can, make it backwards compatible so if a user passes show_progres=True via the kwargs, it'll still apply that correctly, but also raise a deprecation warning. You can use

`from langchain_core._api import warn_deprecated` for deprecation warnings. And target for remove in release 0.4.0.

There's a bit of information here about release policy that may or may not be useful (https://python.langchain.com/v0.2/docs/versions/release_policy/#langchain-and-langchain-core)

JonZeolla:
No problem, will do

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -79,6 +79,16 @@ def __init__(self, **kwargs: Any):
             self.model_name, cache_folder=self.cache_folder, **self.model_kwargs
         )
 
+        if "show_progress_bar" in self.encode_kwargs:
+            warn_deprecated(
+                since="0.2.5",
+                removal="0.4.0",
+                name="encode_kwargs['show_progress_bar']",
+                alternative="show_progress",
+                # message = "Prefer %(alternative)s over %(name)s since %(since)s; %(alternative)s defaults to False"
```
eyurtsev, line 88:
```suggestion
```

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -79,6 +79,16 @@ def __init__(self, **kwargs: Any):
             self.model_name, cache_folder=self.cache_folder, **self.model_kwargs
         )
 
+        if "show_progress_bar" in self.encode_kwargs:
```
eyurtsev, line 82: @JonZeolla do we need to update the logic for the two implementations below?

JonZeolla: @eyurtsev sorry; moved this to draft. Give me some time to test the requested changes locally. I'll @ and set it ready for review when ready

JonZeolla:
@eyurtsev should be all set now

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -167,6 +170,20 @@ def __init__(self, **kwargs: Any):
         except ImportError as e:
             raise ImportError("Dependencies for InstructorEmbedding not found.") from e
 
+        if "show_progress_bar" in self.encode_kwargs:
+            warn_deprecated(
+                since="0.2.5",
```
JonZeolla, line 175: I'm assuming this is the version this would be released in; not sure how to know?

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -154,6 +155,8 @@ class HuggingFaceInstructEmbeddings(BaseModel, Embeddings):
     """Instruction to use for embedding documents."""
     query_instruction: str = DEFAULT_QUERY_INSTRUCTION
     """Instruction to use for embedding query."""
+    show_progress: bool = False
```
JonZeolla, line 158: This used to default on and now defaults to off, not sure if you'd consider that a breaking change since it's just showing output.

libs/community/langchain_community/embeddings/huggingface.py
```diff
@@ -167,6 +170,20 @@ def __init__(self, **kwargs: Any):
         except ImportError as e:
             raise ImportError("Dependencies for InstructorEmbedding not found.") from e
 
+        if "show_progress_bar" in self.encode_kwargs:
+            warn_deprecated(
+                since="0.2.5",
+                removal="0.4.0",
```
JonZeolla, line 176: Logistical question since this is my first contribution but I expect more - how does the removal actually happen as a part of 0.4.0; is there a backlog of these somehow generated when preparing for releases?

libs/community/langchain_community/embeddings/huggingface.py
eyurtsev: Check current version, and then increment patch version by 1

libs/community/langchain_community/embeddings/huggingface.py
eyurtsev: We'll scan the code base prior to release of 0.4.0 for `warn_deprecated` and `@deprecated` usage to see what's targeted for removal

eyurtsev (approved):

PR #22687, opened 2024-06-07
anthropic[minor]: Adds streaming tool call support for Anthropic

jacoblee93:
Preserves string content chunks for non tool call requests for convenience.

One thing - Anthropic events look like this:

```
RawContentBlockStartEvent(content_block=TextBlock(text='', type='text'), index=0, type='content_block_start')
RawContentBlockDeltaEvent(delta=TextDelta(text='<thinking>\nThe', type='text_delta'), index=0, type='content_block_delta')
RawContentBlockDeltaEvent(delta=TextDelta(text=' provide', type='text_delta'), index=0, type='content_block_delta')
...
RawContentBlockStartEvent(content_block=ToolUseBlock(id='toolu_01GJ6x2ddcMG3psDNNe4eDqb', input={}, name='get_weather', type='tool_use'), index=1, type='content_block_start')
RawContentBlockDeltaEvent(delta=InputJsonDelta(partial_json='', type='input_json_delta'), index=1, type='content_block_delta')
```

Note that `delta` has a `type` field. With this implementation, I'm dropping it because `merge_list` behavior will concatenate strings.

We currently have `index` as a special field when merging lists, would it be worth adding `type` too?

If so, what do we set as a context block chunk? `text` vs. `text_delta`/`tool_use` vs `input_json_delta`?

CC @ccurme @efriis @baskaryan


efriis:
Game to add `type` to that special list

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -146,6 +146,55 @@ async def test_abatch_tags() -> None:
         assert isinstance(token.content, str)
 
 
+async def test_async_tool_use() -> None:
+    llm = ChatAnthropic(  # type: ignore[call-arg]
+        model="claude-3-sonnet-20240229",
```
efriis, line 151:
```suggestion
        model=MODEL_NAME,
```

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -313,7 +362,7 @@ async def test_astreaming() -> None:
 
 def test_tool_use() -> None:
     llm = ChatAnthropic(  # type: ignore[call-arg]
-        model="claude-3-opus-20240229",
+        model="claude-3-sonnet-20240229",
```
efriis, line 365:
```suggestion
        model=MODEL_NAME,
```

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -340,12 +389,15 @@ def test_tool_use() -> None:
 
     # Test streaming
     first = True
+    chunks = []  # type: ignore
```
efriis, line 392: did you mean to add something else here operating on chunks? Could we just `assert not first` instead of chunk length otherwise or remove change?

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1114,27 +1068,76 @@ def _make_message_chunk_from_anthropic_event(
...
-        text = event.delta.text
-        message_chunk = AIMessageChunk(content=text)
+    elif (
+        event.type == "content_block_start"
+        and event.content_block is not None
+        and event.content_block.type == "tool_use"
+    ):
+        if coerce_content_to_string:
+            warnings.warn("Received unexpected tool content block.")
+        # TODO: Pass type through here
```
efriis, line 1097: should we make all the keys line up here (and store a `content_block` field), or better to re-form?

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1114,27 +1068,76 @@ def _make_message_chunk_from_anthropic_event(
     event: anthropic.types.RawMessageStreamEvent,
     *,
     stream_usage: bool = True,
+    coerce_content_to_string: bool = False,
```
efriis, line 1071: is this to avoid unnecessarily turning everything into a list of dicts instead of a str? IMO might be "less is more" and change everything to list of dict

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
jacoblee93: It's checking to make sure we get multiple message chunks via the assert below.

jacoblee93: It's checking to make sure we get multiple message chunks via the assert below

libs/partners/anthropic/langchain_anthropic/chat_models.py
jacoblee93: Wanted to limit the amount of adjustment for users for now - and wasn't sure how StrOutputParser would handle it.

libs/partners/anthropic/langchain_anthropic/chat_models.py
jacoblee93: Purely because of `type` string fields - right now they would concatenate repeatedly. Could change in core but am wary of magic params.

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -667,34 +666,12 @@ def _stream(
...
-            else:
-                yield cast(ChatGenerationChunk, result.generations[0])
-            return
         stream = self._client.messages.create(**params, stream=True)
         for event in stream:
             msg = _make_message_chunk_from_anthropic_event(
-                event, stream_usage=stream_usage
+                event,
+                stream_usage=stream_usage,
+                coerce_content_to_string=(not _tools_in_params(params)),
```
ccurme, line 674: micro-nit: compute `_tools_in_params(params)` outside streaming loop?

libs/partners/anthropic/langchain_anthropic/chat_models.py
ccurme: I think we will need the type field to be able to send these messages back to Anthropic (we also [raise ValueError](https://github.com/langchain-ai/langchain/blob/c1fced926923f7cf9bb982e3d0c610aa9dff6f35/libs/partners/anthropic/langchain_anthropic/chat_models.py#L179) without them).

I'm also on board to add `type` as another magic param in core. The risk would be if providers stream out values from the `"type"` key.

Not 100% on this but another idea is to keep track of a mapping from `index` to `type` within ChatAnthropic and pop out the type key from all chunks after the first one per value of `index`.

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1114,27 +1068,76 @@ def _make_message_chunk_from_anthropic_event(
...
+            "index": event.index,
+            "id": event.content_block.id,
+            "name": event.content_block.name,
+            "input": event.content_block.input,
+        }
+        tool_call_chunk = {
+            "index": event.index,
+            "id": event.content_block.id,
+            "name": event.content_block.name,
+            "args": "",
```
ccurme, line 1108: do we need an empty tool call chunk?

jacoblee93: Great point on passing back to Anthropic, will update and add a test to ensure it works as expected.

jacoblee93:
Blocked on #22750 per above discussions on treating `type` as a special field.

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -146,6 +146,55 @@ async def test_abatch_tags() -> None:
...
+        chunks = chunks + [chunk]
+        if first:
+            gathered = chunk
+            first = False
+        else:
+            gathered = gathered + chunk  # type: ignore
+    assert len(chunks) > 1
+    assert isinstance(gathered, AIMessageChunk)
+    assert isinstance(gathered.tool_call_chunks, list)
+    assert len(gathered.tool_call_chunks) == 1
```
efriis, line 191: why is this only 1 chunk? Couldn't it be multiple that sum together into a single tool call?

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -339,20 +388,54 @@ def test_tool_use() -> None:
...
+    )
+    chunks = []  # type: ignore
+    first = True
+    for chunk in stream:
+        chunks = chunks + [chunk]
+        if first:
+            gathered = chunk
+            first = False
+        else:
+            gathered = gathered + chunk  # type: ignore
```
efriis, line 437: might be good to assert some qualities of the chunks, like content is always a string (albeit sometimes empty) etc

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1114,27 +1060,68 @@ def _make_message_chunk_from_anthropic_event(
...
     Note that not all events will result in a message chunk. In these cases
     we return None.
     """
     message_chunk: Optional[AIMessageChunk] = None
+    # See https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/streaming/_messages.py  # noqa: E501
     if event.type == "message_start" and stream_usage:
         input_tokens = event.message.usage.input_tokens
         message_chunk = AIMessageChunk(
-            content="",
+            content="" if coerce_content_to_string else [],
```
efriis, line 1075: To be clear I'm good with this, but could this be a breaking change if someone was relying on content as a str

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1114,27 +1060,68 @@ def _make_message_chunk_from_anthropic_event(
     event: anthropic.types.RawMessageStreamEvent,
     *,
     stream_usage: bool = True,
+    coerce_content_to_string: bool = False,
```
efriis, line 1063: why does this need a default value?

efriis: If internal helper methods, might be more stable if we remove default, and has to be passed in. Don't feel super strongly though

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
jacoblee93: The chunks here combine into one tool call

libs/partners/anthropic/langchain_anthropic/chat_models.py
jacoblee93: This wouldn't be getting hit if `coerce_content_to_string` were true, which it is unless you're passing tools, which in the present version aren't streamed. But it's unnecessary, I'll remove

libs/partners/anthropic/langchain_anthropic/chat_models.py
jacoblee93: Sure

libs/partners/anthropic/langchain_anthropic/chat_models.py
jacoblee93: Never mind, we need this since it's the first chunk. Otherwise we have to update the core logic again to replace an empty string chunk which seems messy.

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
jacoblee93: We get a few different things

efriis (approved):

PR #22690, opened 2024-06-08
Add function response to graph cypher qa chain

tomasonjo:
LLMs struggle with Graph RAG, because it's different from vector RAG in a way that you don't provide the whole context, only the answer and the LLM has to believe. However, that doesn't really work a lot of the time. However, if you wrap the context as function response the accuracy is much better.

btw... `union[LLMChain, Runnable]` is linting fun, that's why so many ignores

### DIFF
--- a/libs/community/langchain_community/chains/graph_qa/cypher.py
+++ b/libs/community/langchain_community/chains/graph_qa/cypher.py
@@ -261,8 +261,8 @@ class GraphCypherQAChain(Chain):
         qa_llm = qa_llm or llm
         if use_function_response:
             try:
-                qa_llm.bind_tools({})  # noqa: F401
-                qa_chain = response_prompt | qa_llm | StrOutputParser()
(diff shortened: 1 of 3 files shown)


libs/community/langchain_community/chains/graph_qa/cypher.py
```diff
@@ -163,12 +209,14 @@ def from_llm(
         qa_prompt: Optional[BasePromptTemplate] = None,
         cypher_prompt: Optional[BasePromptTemplate] = None,
         cypher_llm: Optional[BaseLanguageModel] = None,
-        qa_llm: Optional[BaseLanguageModel] = None,
+        qa_llm: Optional[Union[BaseLanguageModel, Any]] = None,
```
baskaryan, line 212: nit: could just type as Any at this point (no need for Optional or Union)

baskaryan (approved):

PR #22698, opened 2024-06-08
community: Added memory to OpenAPI planner and allowed optional endpoint parameters

jameslai-dev:
**Description:**

Dear LangChain team.

Thanks for your fantastic framework. LangChain enables us to build innovative LLM applications. While utilizing OpenAPI toolkit from the LangChain community package, I encountered a limitation with the OpenAPI planner agent: it cannot retain information about previous user queries and responses. If users want to add a constraint or requirement to their previous query, they must restate a new query instead of simply mentioning the new condition. I believe the latter approach is more natural and closer to real-world conversations.

For example, let's say I'm searching for laptop products from Klarna, and I initially ask: `Please list 5 laptops in US for me`. Without memory, if I want to add a price constraint, I will need to repeat my initial query and specify the price range: `Please list 5 laptops with a maximum price of $500 in US for me`. However, with memory functionality, I can just continue the conversation by asking: `Please search again with a maximum price of $500.`

This pull request made the following changes:
- Added a `{history}` section to the OpenAPI planner prompts to store chat history.
- If the `shared_memory` parameter is assigned to `create_openapi_agent`:
  - The provided memory object is passed to the root agent of OpenAPI planner to store the chat history with the user.
  - A read-only memory, wrapped from the given memory, is passed to the planner and controller to ensure the chat history remains unchanged during the planning progress.
- Introduced `include_optional_params` as an optional parameter of `reduce_openapi_spec` with a default value of `False`. When set to `True`, optional parameters of OpenAPI endpoints are included in the generated API documentation, allowing LLMs to leverage optional endpoint parameters.
- Added an example showcasing the interaction with memory-enabled OpenAPI planner to the existing OpenAPI toolkit example: `docs/docs/integrations/toolkits/openapi.ipynb`

**Issue:**

None

**Dependencies:**

None

**Twitter handle:**

None

### DIFF
--- a/docs/docs/integrations/toolkits/openapi.ipynb
+++ b/docs/docs/integrations/toolkits/openapi.ipynb
@@ -593,10 +593,338 @@
   },
   {
    "cell_type": "markdown",
+   "id": "75175211-d24b-4aae-a95c-d3a7faba39de",
+   "metadata": {},
+   "source": [
+    "## 2nd example: hierarchical planning agent with memory\n",
+    "\n",
+    "Continuing with the previous example, we'll enable chat history support for the planner. With memory, we can ask the agent to modify the request or make a new request following the conversation we've made with the agent.\n",
+    "\n",
+    "---"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "01155ac9-e189-4c54-b79a-e0f3bd399c70",
+   "metadata": {},
+   "source": [
+    "### Load an OpenAPI spec\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


jamesljlster:
Rebased onto the latest master branch (head at 8f019e91d7206c830f6a32678a9d4497b12c66ec).

isahers1:
@baskaryan I looked through the code and it lgtm, but I'm not 100% because there were a decent amount of changes - you might also have comments on the doc changes

PR #22716, opened 2024-06-09
[Partner]: Add metadata to stream response

ha3:
Adds `response_metadata` to stream responses from OpenAI. This is returned with `invoke` normally, but wasn't implemented for `stream`.


ccurme:
Hi @ha3, thanks for this!

I've made an adjustment where we just put the metadata in `generation_info`. It is a little circuitous but in langchain-core we merge this back into the `.response_metadata` on the message:
https://github.com/langchain-ai/langchain/blob/c4f2bc95401bd41d40c9c7ba2fe959e12a059408/libs/core/langchain_core/language_models/chat_models.py#L996-L1002

This simplifies things slightly. I also took the opportunity to rename some of the various `chunk` variables (since they are different types) and added tests. Will test a bit more on my end before merging. Let me know if you have any more input.

ha3:
@ccurme thanks for the improvements and testing. It looks good on my side!

ccurme (approved):

PR #22721, opened 2024-06-09
anthropic: always add tool_result type to ToolMessage content

ccurme:
Anthropic tool results can contain image data, which are typically represented with content blocks having `"type": "image"`. Currently, these content blocks are passed as-is as human/user messages to Anthropic, which raises BadRequestError as it expects a tool_result block to follow a tool_use.

Here we update ChatAnthropic to nest the content blocks inside a tool_result content block.

Example:
```python
import base64

import httpx
from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
from langchain_core.pydantic_v1 import BaseModel, Field


# Fetch image
image_url = "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
image_data = base64.b64encode(httpx.get(image_url).content).decode("utf-8")


class FetchImage(BaseModel):
    should_fetch: bool = Field(..., description="Whether an image is requested.")


llm = ChatAnthropic(model="claude-3-sonnet-20240229").bind_tools([FetchImage])

messages = [
    HumanMessage(content="Could you summon a beautiful image please?"),
    AIMessage(
        content=[
            {
                "type": "tool_use",
                "id": "toolu_01Rn6Qvj5m7955x9m9Pfxbcx",
                "name": "FetchImage",
                "input": {"should_fetch": True},
            },
        ],
        tool_calls=[
            {
                "name": "FetchImage",
                "args": {"should_fetch": True},
                "id": "toolu_01Rn6Qvj5m7955x9m9Pfxbcx",
            },
        ],
    ),
    ToolMessage(
        name="FetchImage",
        content=[
            {
                "type": "image",
                "source": {
                    "type": "base64",
                    "media_type": "image/jpeg",
                    "data": image_data,
                },
            },
        ],
        tool_call_id="toolu_01Rn6Qvj5m7955x9m9Pfxbcx",
    ),
]

llm.invoke(messages)
```

Trace: https://smith.langchain.com/public/d27e4fc1-a96d-41e1-9f52-54f5004122db/r


baskaryan (commented):
could we update the unit tests for this as well

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -104,18 +104,15 @@ def _merge_messages(
...
-                        }
-                    ]
-                )
-            else:
-                curr = HumanMessage(curr.content)  # type: ignore[misc]
+            curr = HumanMessage(  # type: ignore[misc]
+                [
+                    {
+                        "type": "tool_result",
+                        "content": curr.content,
```
baskaryan, line 111: we should check that curr.content isn't already a list with tool_result in it

ccurme: good call. made this change and added a test case.

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -104,15 +104,21 @@ def _merge_messages(
...
-            curr = HumanMessage(  # type: ignore[misc]
-                [
-                    {
-                        "type": "tool_result",
-                        "content": curr.content,
-                        "tool_use_id": curr.tool_call_id,
-                    }
-                ]
-            )
+            if isinstance(curr.content, list) and any(
```
baskaryan, line 107: uber nit but shouldn't these all be tool_result type (instead of `any`)?

ccurme: good catch thanks.

baskaryan (approved):

PR #22724, opened 2024-06-10
community[patch]: Kinetica Integrations handled error in querying; quotes in table names; updated gpudb API

am-kinetica:
- [ ] **Miscellaneous updates and fixes**: 
    - **Description:** Handled error in querying; quotes in table names; updated gpudb API
    - **Issue:** Threw an error with an error message difficult to understand if a query failed or returned no records
    - **Dependencies:** Updated GPUDB API version to `7.2.0.9`

@baskaryan @hwchase17

### DIFF
--- a/libs/community/langchain_community/vectorstores/kinetica.py
+++ b/libs/community/langchain_community/vectorstores/kinetica.py
@@ -427,7 +427,7 @@ class Kinetica(VectorStore):
         filter: Optional[dict] = None,
     ) -> List[Tuple[Document, float]]:
         from gpudb import GPUdbException
-        
+
         resp: Dict = self.__query_collection(embedding, k, filter)
         if resp and resp["status_info"]["status"] == "OK" and "records" in resp:
             records: OrderedDict = resp["records"]
--- a/libs/community/langchain_community/vectorstores/kinetica.py
+++ b/libs/community/langchain_community/vectorstores/kinetica.py
@@ -834,10 +834,11 @@ class Kinetica(VectorStore):
         config: KineticaSettings = KineticaSettings(),
         dimensions: int = Dimension.OPENAI,
         collection_name: str = _LANGCHAIN_DEFAULT_COLLECTION_NAME,
-        schema_name: str = _LANGCHAIN_DEFAULT_SCHEMA_NAME,
         distance_strategy: DistanceStrategy = DEFAULT_DISTANCE_STRATEGY,
         ids: Optional[List[str]] = None,
         pre_delete_collection: bool = False,
+        *,
+        schema_name: str = _LANGCHAIN_DEFAULT_SCHEMA_NAME,
         **kwargs: Any,
     ) -> Kinetica:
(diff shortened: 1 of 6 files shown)


libs/community/langchain_community/vectorstores/kinetica.py
```diff
@@ -221,6 +221,7 @@ def __from(
         metadatas: Optional[List[dict]] = None,
         ids: Optional[List[str]] = None,
         collection_name: str = _LANGCHAIN_DEFAULT_COLLECTION_NAME,
+        schema_name: str = _LANGCHAIN_DEFAULT_SCHEMA_NAME,
```
eyurtsev, line 224: This is a breaking change because users may technically be invoking this by position

could you add place it as a last parameter and right before it use a `*` operator

So

```python
logger: ...,
*,
schema_name: str = ...,
**kwargs

```

libs/community/langchain_community/vectorstores/kinetica.py
```diff
@@ -757,6 +763,7 @@ def from_texts(
         metadatas: Optional[List[dict]] = None,
         config: KineticaSettings = KineticaSettings(),
         collection_name: str = _LANGCHAIN_DEFAULT_COLLECTION_NAME,
+        schema_name: str = _LANGCHAIN_DEFAULT_SCHEMA_NAME,
```
eyurtsev, line 766: Could you update the ordering of parmeters in function signature here and in remaining places to avoid making a breaking change (see comment before about using `*` and placing at end of function signature)

eyurtsev (changes requested):
Looks good just a few minor changes to avoid making a breaking API change.

When ready to re-view, could you click on the "request re-review button"

libs/community/langchain_community/vectorstores/kinetica.py
am-kinetica: @eyurtsev Thanks for pointing this, I have modified the code to make this keyword only.

libs/community/langchain_community/vectorstores/kinetica.py
am-kinetica: Nade this parameter keyword only as suggested

libs/community/langchain_community/vectorstores/kinetica.py
```diff
@@ -423,11 +426,14 @@ def similarity_search_with_score_by_vector(
...
+            records: OrderedDict = resp["records"]
+            results = list(zip(*list(records.values())))
 
-        records: OrderedDict = resp["records"]
-        results = list(zip(*list(records.values())))
-
-        return self._results_to_docs_and_scores(results)
+            return self._results_to_docs_and_scores(results)
+        else:
+            self.logger.error(resp["status_info"]["message"])
```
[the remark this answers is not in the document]
am-kinetica: Modified, now raises `GPUdbException` to propagate the exact error message from the server

eyurtsev (approved):

PR #22728, opened 2024-06-10
text-splitter[minor]: Declare MarkdownHeaderTextSplitter deprecated. Use MarkdownTextSplitter.

pprados:
Langchain offers two Markdown splitter implementations:
- MarkdownTextSplitter
- MarkdownHeaderTextSplitter

The `MarkdownHeaderTextSplitter` class is not langchain compliant, since it inherits from anyone. It is not possible to use in a function or class that expects a `TextSplitter`.

In my experience, users who come across this class don't know that there is another exists. If they need to be `TextSplitter` compatible, they write a new class.

We propose to declare it deprecated, and referencing `MarkdownTextSplitter`.


eyurtsev:
@baskaryan any idea if we should remove or sub-class?

baskaryan (commented):
are MarkdownHeaderTextSplitter and MarkdownTextSplitter identical in functionality

pprados:
`MarkdownHeaderTextSplitter` is not exactly the same as `MarkdownTextSplitter`.
It's possible to merge the feature in one class
- `headers_to_split_on: List[Tuple[str, str]]`, => `separators`
- `return_each_line: bool = False`,  # To add in `RecursiveCharacterTextSplitter`
- `strip_headers: bool = True`, # To add in `RecursiveCharacterTextSplitter`

Or,`MarkdownTextSplitter()` can inherit `TextSplitter`.

Having two classes for the same use creates confusion.

baskaryan:
> `MarkdownHeaderTextSplitter` is not exactly the same as `MarkdownTextSplitter`. It's possible to merge the feature in one class
> 
> * `headers_to_split_on: List[Tuple[str, str]]`, => `separators`
> * `return_each_line: bool = False`,  # To add in `RecursiveCharacterTextSplitter`
> * `strip_headers: bool = True`, # To add in `RecursiveCharacterTextSplitter`
> 
> Or,`MarkdownTextSplitter()` can inherit `TextSplitter`.
> 
> Having two classes for the same use creates confusion.

agree that two classes is confusing but don't think we can deprecate one until the other has a superset of the functionality. would be down to try and merge the features

pprados:
@baskaryan @eyurtsev 
As a reminder:
- `MarkdownHeaderTextSplitter` IS NOT a `TextSplitter`
- `MarkdownTextSplitter` IS a `TextSplitter`

I try to merge `MarkdownHeaderTextSplitter` and `MarkdownTextSplitter` or make `MarkdownHeaderTextSplitter` extends `TextSplitter`.

But, the problem is the signature of the method `split_text`:
- `MarkdownHeaderTextSplitter.split_text(self, text: str) -> List[Document]`
- `MarkdownTextSplitter.split_text(self, text: str) -> List[str]`

Perhaps we should proceed gradually. First, declare `MarkdownHeaderTextSplitter.split_text(self, text: str)` deprecated and replace it with a new name (`MarkdownHeaderTextSplitter.split_text_to_docs()`). Later, marry the two classes.

What do you think?

efriis:
This PR doesn't do what was described - it adds the wrong method to the implementation the description claims would be deprecated, and deprecates `split_text` which is the one we want :(

Closing

PR #22735, opened 2024-06-10
Experimental: Repair parsed_json errors made by LLMs, e.g Llama3:8b and Mistral:7b when converting documents into graph nodes.

Blueflamefirez:
- [ ] **Repair parsed_json dictionary mistakes made by LLMs, e.g Llama3 and Mistral**: "Experimental: Added a __repair_response_ function for the "parsed_json" output"

- [ ] **PR message**:
    - **Description:** The __repair_response_ function checks if the parsed_json has any nested lists, and flattens them into a singular list, moreover, since the nodes cannot hash "List" type, only the first instance of the list is chosen, and the List is then discarded. Moreover, the function also converts any None types into "None", and any empty strings into "None" as well. This is because the node system None type is invalid and neo4j does not allow for empty strings. (Tested for LLama3:8b and Mistral:7b)
    - **Issue:** The LLM output for "process_response()" is unpredictable and not very robust. Different models yield different output structures. When tested for gemma:2b, a properties key is sometimes added making the dictionary parsing fail. For mistral:7b and Llama3:8b, None types, empty strings, and nested lists are sometimes added. To mitigate the latter part, i created a function that parses the "parsed_json" dictionary and removes any; Nested lists: taking the first value instead, None types: converting into "None", and lastly any empty strings: also converting into "None".

### DIFF
--- a/libs/experimental/langchain_experimental/graph_transformers/llm.py
+++ b/libs/experimental/langchain_experimental/graph_transformers/llm.py
@@ -509,6 +509,54 @@ def _convert_to_graph_document(
     return _format_nodes(nodes), _format_relationships(relationships)
 
 
+def _repair_response(json: list[dict]) -> list[dict]:
+    def flatten_nested_lists(data: Any) -> Any:
+        if isinstance(data, list):
+            while isinstance(data, list):
+                data = data[0]
+        return data
+
+    # If the json is nested
+    while isinstance(json, list) and isinstance(json[0], list):
+        json = json[0]
+
+    # Remove any None types, empty strings, and flatten nested lists
+    for rel in json:
+        rel["head"] = flatten_nested_lists(rel.get("head"))
+        rel["head_type"] = flatten_nested_lists(rel.get("head_type"))
+        rel["tail"] = flatten_nested_lists(rel.get("tail"))
+        rel["tail_type"] = flatten_nested_lists(rel.get("tail_type"))
+        rel["relation"] = flatten_nested_lists(rel.get("relation"))
+
+        rel["tail"] = (
+            rel["tail"]
+            if rel.get("tail") is not None and rel.get("tail") != ""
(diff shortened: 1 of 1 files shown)


baskaryan (commented):
cc @tomasonjo

tomasonjo:
We don't really plan to support current 3b models, and I think this workaround isn't good at all:

```
        rel["tail"] = (
            rel["tail"]
            if rel.get("tail") is not None and rel.get("tail") != ""
            else "None"
        )
 ```

Blueflamefirez:
> We don't really plan to support current 3b models, and I think this workaround isn't good at all:
> 
> ```
>         rel["tail"] = (
>             rel["tail"]
>             if rel.get("tail") is not None and rel.get("tail") != ""
>             else "None"
>         )
> ```

The issue i created the function for stems from the llama3:8b and mistral:7b models, i only tried to use the gemma:2b model but quickly gave up on it. So this is for the 8b and 7b  models, not 3b.

What would be a good workaround? The models themselves output a None type which crashes the node creation since they do not support none types, and some other mistakes like nested lists and so on.

tomasonjo:
Yeah, we just ignore node results if they aren't correct

Blueflamefirez:
> Yeah, we just ignore node results if they aren't correct

Huh interesting, however, there is no such functionality currently that ignores incorrect nodes, it just crashes instead. How do you suppose the proposed function can be modified to achieve what is up to par of expectations?

tomasonjo:
Iterate over each rel and at minimum, node ids and rel type has to be valid, otherwise ignore

baskaryan:
@Blueflamefirez @tomasonjo is this ok to close?

tomasonjo:
Yes

V tor., 3. sep. 2024, 06:37 je oseba Bagatur ***@***.***>
napisala:

PR #22741, opened 2024-06-10
RFC [Core] Tracing - unify tracing

hinthornw:


libs/core/langchain_core/tracers/schemas.py
```diff
@@ -110,14 +111,13 @@ class ToolRun(BaseRun):
 # Begin V2 API Schemas
 
 
-class Run(BaseRunV2):
+# TODO: rm client API key validation
```
hinthornw, line 114: Would want to make the client creation lazy so you don't have to have an api key set in environment just to use this.

Also there's a fun pydantic gotcha where since we mark a couple fields as "excluded" in teh object, `run.copy()` does a deep copy on **other** fields (i think) which isn't the behavior we want.

PR #13318, opened 2023-11-14
Framework for supporting more languages in LanguageParser

ThatsJustCheesy:
## Description

I am submitting this for a school project as part of a team of 5. Other team members are @LeilaChr, @maazh10, @Megabear137, @jelalalamy. This PR also has contributions from community members @Harrolee and @Mario928.

Initial context is in the issue we opened (#11229).

This pull request adds:

- Generic framework for expanding the languages that `LanguageParser` can handle, using the [tree-sitter](https://github.com/tree-sitter/py-tree-sitter#py-tree-sitter) parsing library and existing language-specific parsers written for it
- Support for the following additional languages in `LanguageParser`:
  - C
  - C++
  - C#
  - Go
  - Java (contributed by @Mario928 https://github.com/ThatsJustCheesy/langchain/pull/2)
  - Kotlin
  - Lua
  - Perl
  - Ruby
  - Rust
  - Scala
  - TypeScript (contributed by @Harrolee https://github.com/ThatsJustCheesy/langchain/pull/1)

Here is the [design document](https://docs.google.com/document/d/17dB14cKCWAaiTeSeBtxHpoVPGKrsPye8W0o_WClz2kk) if curious, but no need to read it.

## Issues

- Closes #11229
- Closes #10996
- Closes #8405

## Dependencies

`tree_sitter` and `tree_sitter_languages` on PyPI. We have tried to add these as optional dependencies.

## Documentation

We have updated the list of supported languages, and also added a section to `source_code.ipynb` detailing how to add support for additional languages using our framework.

## Maintainer

- @hwchase17 (previously reviewed https://github.com/langchain-ai/langchain/pull/6486)

Thanks!!

## Git commits

We will gladly squash any/all of our commits (esp merge commits) if necessary. Let us know if this is desirable, or if you will be squash-merging anyway.

PR #22742, opened 2024-06-10
Community[minor]: Add language parser for Elixir

klappradla:
Hi 👋 

First off, thanks a ton for your work on this 💚 Really appreciate what you're providing here for the community.

## Description

This PR adds a basic language parser for the [Elixir](https://elixir-lang.org/) programming language. The parser code is based upon the approach outlined in https://github.com/langchain-ai/langchain/pull/13318: it's using `tree-sitter` under the hood and aligns with all the other `tree-sitter` based parses added that PR.

The `CHUNK_QUERY` I'm using here is probably not the most sophisticated one, but it worked for my application. It's a starting point to provide "core" parsing support for Elixir in LangChain. It enables people to use the language parser out in real world applications which may then lead to further tweaking of the queries. I consider this PR just the ground work.

- **Dependencies:** requires `tree-sitter` and `tree-sitter-languages` from the extended dependencies
- **Twitter handle:**`@bitcrowd`

## Checklist

- [x] **Add tests and docs**
- [x] **Lint and test**: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified.

### DIFF
--- a/libs/community/tests/unit_tests/document_loaders/parsers/language/test_elixir.py
+++ b/libs/community/tests/unit_tests/document_loaders/parsers/language/test_elixir.py
@@ -1,0 +1,57 @@
+import unittest
+
+import pytest
+
+from langchain_community.document_loaders.parsers.language.elixir import ElixirSegmenter
+
+
+@pytest.mark.requires("tree_sitter", "tree_sitter_languages")
+class TestElixirSegmenter(unittest.TestCase):
+    def setUp(self) -> None:
+        self.example_code = """@doc "some comment"
+def foo do
+  i = 0
+end
+
+defmodule M do
+  def hi do
+    i = 2
+  end
+
+  defp wave do
+    :ok
+  end
+end"""
+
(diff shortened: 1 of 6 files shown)


libs/community/tests/unit_tests/document_loaders/parsers/language/test_elixir.py
```diff
@@ -0,0 +1,57 @@
+import unittest
+
+import pytest
+
+from langchain_community.document_loaders.parsers.language.elixir import ElixirSegmenter
+
+
+@pytest.mark.requires("tree_sitter", "tree_sitter_languages")
+class TestElixirSegmenter(unittest.TestCase):
```
eyurtsev, line 9: FWIW we usually use pytest style tests rather than unit test style tests, if you have time, consider updating the test structure in the next PR

eyurtsev (approved):

eyurtsev (approved):

PR #22747, opened 2024-06-10
core: fix mustache falsy cases

efriis:


libs/core/tests/unit_tests/prompts/test_prompt.py
```diff
@@ -499,3 +500,56 @@ async def test_prompt_ainvoke_with_metadata() -> None:
...
+        ),
+        (
+            {},
+            {
+                "mustache": "",
+                "f-string": "{}",
+            },
+        ),
+    ],
+)
```
efriis, line 535: is this the spec we want to follow?

PR #22750, opened 2024-06-10
core[patch]: Treat type as a special field when merging lists

jacoblee93:
Should we even log a warning? At least for Anthropic, it's expected to get e.g. `text_block` followed by `text_delta`.

@ccurme @baskaryan @efriis


isahers1:
@baskaryan lgtm other than the link warning - which I believe is from the "import warnings" without using the package

PR #22752, opened 2024-06-10
core[minor]: message transformer utils

baskaryan:


libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +255,136 @@ def convert_to_messages(
...
+        elif excl_ids and msg.id in excl_ids:
+            continue
+        else:
+            filtered.append(msg)
+    return filtered
+
+
+def abbreviate_messages(
+    messages: Sequence[MessageLikeRepresentation],
+    n_tokens: int,
```
eyurtsev, line 307: What's n tokens

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +255,136 @@ def convert_to_messages(
...
+    messages: Sequence[BaseMessage],
+    n_tokens: int,
+    list_token_counter: Callable[[Sequence[BaseMessage]], int],
+    *,
+    allow_partial_messages: bool = False,
+) -> List[BaseMessage]:
+    ...
+
+
+def _last_n_tokens(
```
eyurtsev, line 370: Do we need an interface for strategy? Messages in -> messages out?

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +255,136 @@ def convert_to_messages(
...
+        elif incl_ids and msg.id not in incl_ids:
+            continue
+        elif excl_ids and msg.id in excl_ids:
+            continue
+        else:
+            filtered.append(msg)
+    return filtered
+
+
+def abbreviate_messages(
```
eyurtsev, line 305: Trim sounds better to me even if there's summarization

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +255,136 @@ def convert_to_messages(
         List of messages (BaseMessages).
     """
     return [_convert_to_message(m) for m in messages]
+
+
+def filter_messages(
```
eyurtsev, line 260: Is this internal or user facing?

libs/core/langchain_core/messages/utils.py
baskaryan: num tokens to abbreviate to. alternate ideas
`max_tokens` — in our codebase this usually refers to max tokens to generate, was worried reuse might confuse
`token_count` — n_tokens just seems more succinct to me but clearly its not obv what it means

libs/core/langchain_core/messages/utils.py
baskaryan: seems like overkill to me since these are all private methods but no strong feelings

libs/core/langchain_core/messages/utils.py
baskaryan: personally disagree (trim really sounds like just shaving off some, not condensing/summarizing) but dont feel strongly

libs/core/langchain_core/messages/utils.py
baskaryan: user facing

libs/core/langchain_core/messages/ai.py
```diff
@@ -152,8 +159,28 @@ def lc_attributes(self) -> Dict:
     @root_validator()
     def init_tool_calls(cls, values: dict) -> dict:
         if not values["tool_call_chunks"]:
-            values["tool_calls"] = []
-            values["invalid_tool_calls"] = []
+            if values["tool_calls"]:
```
baskaryan, line 162: cc @ccurme changing init behavior of chunks to create chunks from tool calls if only the latter is present

baskaryan: should add unit test

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +248,604 @@ def convert_to_messages(
...
+    strategy: Literal["first", "last"] = "last",
+    allow_partial: bool = False,
+    end_on: Optional[
+        Union[str, Type[BaseMessage], Sequence[Union[str, Type[BaseMessage]]]]
+    ] = None,
+    start_on: Optional[
+        Union[str, Type[BaseMessage], Sequence[Union[str, Type[BaseMessage]]]]
+    ] = None,
+    include_system: bool = False,
+    text_splitter: Optional[Callable[[str], List[str]]] = None,
```
baskaryan, line 448: document and test

baskaryan: support TextSplitter? would require conditional imports

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +248,604 @@ def convert_to_messages(
...
+                last_chunk.content += "\n"
+            merged.append(_chunk_to_msg(last_chunk + curr_chunk))
+    return merged
+
+
+def trim_messages(
+    messages: Sequence[MessageLikeRepresentation],
+    *,
+    n_tokens: int,
+    token_counter: Union[
```
baskaryan, line 436: support tokenizers?

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +248,604 @@ def convert_to_messages(
...
+            merged.append(curr)
+        elif isinstance(curr, ToolMessage) or not isinstance(curr, last.__class__):
+            merged.extend([last, curr])
+        else:
+            last_chunk = _msg_to_chunk(last)
+            curr_chunk = _msg_to_chunk(curr)
+            if isinstance(last_chunk.content, str) and isinstance(
+                curr_chunk.content, str
+            ):
+                last_chunk.content += "\n"
```
baskaryan, line 427: note: we add a newline separator when merging two string messages, which is a more opinionated choice

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +248,604 @@ def convert_to_messages(
...
+
+            [
+                SystemMessage("you're a good assistant."),
+                HumanMessage("what's your name", id="foo", name="example_user"),
+            ]
+    """  # noqa: E501
+    messages = convert_to_messages(messages)
+    filtered: List[BaseMessage] = []
+    for msg in messages:
+        if excl_names and msg.name in excl_names:
```
ccurme, line 316: where there are conflicts, here we prefer exclusions generally. would it make sense to instead prefer id > name > type? e.g., so I can say "exclude all human messages but include id 123, which is a human message"

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +248,604 @@ def convert_to_messages(
...
+
+    return filtered
+
+
+def merge_message_runs(
+    messages: Sequence[MessageLikeRepresentation],
+) -> List[BaseMessage]:
+    """Merge runs of Messages of the same type.
+
+    **NOTE**: ToolMessages are not merged, as each has a distinct tool call id that
```
ccurme, line 349: should we merge them if they have the same tool call ID?

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +248,604 @@ def convert_to_messages(
...
+            ]
+
+    """  # noqa: E501
+    if not messages:
+        return []
+    messages = convert_to_messages(messages)
+    merged: List[BaseMessage] = []
+    for msg in messages:
+        curr = msg.copy(deep=True)
+        last = merged.pop() if merged else None
```
ccurme, line 416: micro nit: `previous`?

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +248,604 @@ def convert_to_messages(
...
+        messages: Sequence of Message-like objects to trim.
+        n_tokens: Max token count of trimmed messages.
+        token_counter: Function for counting tokens of a BaseMessage or a list of
+            BaseMessage.
+        strategy: Strategy for trimming.
+            - "first": Keep the first <= n_count tokens of the messages.
+            - "last": Keep the last <= n_count tokens of the messages.
+        allow_partial: Whether to split a message if only part of the message can be
+            included.
+        end_on: The message type to end on. Can be specified as string names (e.g.
```
ccurme, line 462: Suggest explaining that the message list is trimmed to satisfy this (maybe that's obvious :slightly_smiling_face:)

ccurme (approved):

libs/core/langchain_core/messages/utils.py
baskaryan: imo thats a bit too hard to grok as an end user

libs/core/langchain_core/messages/utils.py
baskaryan: mm in my mind thats an invalid list of messages to begin with

libs/core/langchain_core/messages/utils.py
baskaryan: no pref, happy to change

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +255,650 @@ def convert_to_messages(
...
+    end_on: Optional[
+        Union[str, Type[BaseMessage], Sequence[Union[str, Type[BaseMessage]]]]
+    ] = None,
+    start_on: Optional[
+        Union[str, Type[BaseMessage], Sequence[Union[str, Type[BaseMessage]]]]
+    ] = None,
+    include_system: bool = False,
+    text_splitter: Optional[Union[Callable[[str], List[str]], TextSplitter]] = None,
+) -> List[BaseMessage]:
+    """Trim messages to be below a token count.
```
eyurtsev, line 459: Most common need is probably to keep system message, and last n messages while ending on a particular type of message. We should probably document it as prominently as possible?

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +255,650 @@ def convert_to_messages(
...
+            [
+                SystemMessage("This is a 4 token text. The full message is 10 tokens."),
+                HumanMessage("This is a 4 token text. The full message is 10 tokens.", id="first"),
+                AIMessage( [{"type": "text", "text": "This is the FIRST 4 token block."}], id="second"),
+            ]
+
+        First 30 tokens, allowing partial messages, have to end on HumanMessage:
+        .. code-block:: python
+
+            trim_messages(
```
eyurtsev, line 574: not sure if you'd agree but seeing `messages` as a value for each block will make it easier to see what the API does

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +255,650 @@ def convert_to_messages(
...
+    ] = None,
+    include_system: bool = False,
+    text_splitter: Optional[Union[Callable[[str], List[str]], TextSplitter]] = None,
+) -> List[BaseMessage]:
+    """Trim messages to be below a token count.
+
+    Args:
+        messages: Sequence of Message-like objects to trim.
+        n_tokens: Max token count of trimmed messages.
+        token_counter: Function or llm for counting tokens in a BaseMessage or a list of
```
eyurtsev, line 464: Should we avoid advertising `BaseLanguageModel`? is `get_num_tokens_from_messages()` implemented correctly?

---

Is there any ambiguity in how to apply the token counter if the token counter takes in a list of BaseMessage and there are nearest neighbor interactions?

baskaryan: imo via model is the main way ppl will pass this in, nobody wants to think about implementing a token counter. think we should fix our implementations if they're broken

not sure i understand second part, which parts would be ambiguous?

libs/core/langchain_core/messages/utils.py
```diff
@@ -242,3 +258,754 @@ def convert_to_messages(
...
+            continue
+        if include_names and msg.name in include_names:
+            filtered.append(msg)
+            continue
+        elif include_types and _is_message_type(msg, include_types):
+            filtered.append(msg)
+            continue
+        elif include_ids and msg.id in include_ids:
+            filtered.append(msg)
+            continue
```
isahers1, line 377: Are these continues necessary?

baskaryan: nope good catch

Issue #22770, opened 2024-06-11
DocumentDBVectorSearch and metadata filtering


andreadisimone:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

not applicable

### Error Message and Stack Trace (if applicable)

not applicable

### Description

The DocumentDBVectorSearch docs mention it supports metadata filtering:

https://api.python.langchain.com/en/latest/vectorstores/langchain_community.vectorstores.documentdb.DocumentDBVectorSearch.html#langchain_community.vectorstores.documentdb.DocumentDBVectorSearch.as_retriever

However, unless I misunderstand the code, I really don't think it does.

I see that VectorStoreRetriever._get_relevant_documents passes search_kwargs to the similarity search of the underlying vector store.
And nothing in the code of DocumentDBVectorSearch is using search_kwargs at all.

In my project we need to review relevant parts of opensource softwares to make sure they really meet the requirements. So if this is not a bug, and the feature is indeed implemented somewhere else, could anybody please clarify how metadata filtering in DocumentDBVectorSearch is implemented?

### System Info

not applicable

keenborder786:
Nope you are wrong, vectorstore retriever  unzips the `search_kwargs` [here](https://github.com/keenborder786/langchain/blob/27b9ea14a51398e8ff2a4972026178778f85134b/libs/core/langchain_core/vectorstores.py#L797) therefore using the passed parameter such as `k` and `ef_search` [here](https://github.com/keenborder786/langchain/blob/27b9ea14a51398e8ff2a4972026178778f85134b/libs/community/langchain_community/vectorstores/documentdb.py#L353). The `filter` parameter that you are suggesting is not supported in current implementation of `DocumentDB`. Their is no bug whatsoever.

andreadisimone:
@keenborder786 thanks for the quick reply.
so you are saying the docs are wrong when they say DocumentDBVectorSearch supports filtering?

andreadisimone:
re: DocumentDB not supporting the "filter" parameter, that seems to me like an internal (LangChain's) implementation detail.

[...]

PR #22777, opened 2024-06-11
[Community]: Added Metadata filter support for DocumentDB Vector Store

keenborder786:
- **Description:** As pointed out in this issue #22770, DocumentDB `similarity_search` does not support filtering through metadata which this PR adds by passing in the parameter `filter`. Also this PR fixes a minor Documentation error.
- **Issue:** #22770

### DIFF
--- a/libs/community/langchain_community/vectorstores/documentdb.py
+++ b/libs/community/langchain_community/vectorstores/documentdb.py
@@ -357,6 +357,7 @@ class DocumentDBVectorSearch(VectorStore):
         query: str,
         k: int = 4,
         ef_search: int = 40,
+        *,
         filter: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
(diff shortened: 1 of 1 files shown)


keenborder786:
@baskaryan

libs/community/langchain_community/vectorstores/documentdb.py
```diff
@@ -352,10 +357,11 @@ def similarity_search(
         query: str,
         k: int = 4,
         ef_search: int = 40,
+        filter: Optional[Dict[str, Any]] = None,
```
efriis, line 360:
```suggestion
        *,
        filter: Optional[Dict[str, Any]] = None,
```
changing to be a kwarg instead, so order changes aren't a breaking change in the future

efriis (approved):

PR #22780, opened 2024-06-11
community: add safe_query mode to QuerySQLDataBaseTool

D4ve-R:
Add a safe query mode to the QuerySQLDataBaseTool to ensure no database modifying sql commands are run by the agent.

Add `safe_query: bool` to QuerySQLDataBaseTool
Add `forbidden_keywords: List[str]` to QuerySQLDataBaseTool

### DIFF
--- a/libs/community/langchain_community/tools/sql_database/tool.py
+++ b/libs/community/langchain_community/tools/sql_database/tool.py
@@ -33,6 +33,28 @@ class _QuerySQLDataBaseToolInput(BaseModel):
 class QuerySQLDataBaseTool(BaseSQLDatabaseTool, BaseTool):
     """Tool for querying a SQL database."""
 
+    safe_query: bool = Field(
+        True,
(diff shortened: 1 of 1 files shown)


eyurtsev (changes requested):
Hi @D4ve-R! Thanks for the contribution!

There are two issues with this strategy:

1. A strategy based on filtering out queries based on presence of keywords is generally not effective at sanitization iif exposed to untrusted inputs.
2. This is a breaking change for users that might want to rely on the tool to WRITE data to the database (which is a valid use case)

The correct way to handle this is by scoping permissions at the database level.

See instructions here: https://python.langchain.com/v0.2/docs/security/

Or else by have pre-processing logic that gets the ast of the SQL query and uses an allow list rather than a block list approach (but this is a ton of work, and it's much easier to just assign permissions at the database level)

D4ve-R:
Alright! Thank you for the feedback!

PR #22790, opened 2024-06-11
Add RAG to conceptual guide

rlancemartin:


docs/docs/concepts.mdx
```diff
@@ -754,7 +754,98 @@ There are two main use cases for function/tool calling:
...
+:::tip
+
+* See our RAG from Scratch [code](https://github.com/langchain-ai/rag-from-scratch) and [video series](https://youtube.com/playlist?list=PLfaIDFEXuae2LXbO1_PKyVJiQ23ZztA0x&feature=shared).
+* See our documentation focused on RAG [here](/docs/tutorials/#working-with-external-knowledge).
+
+:::
+
+RAG is only as good as the retrieved documents’ relevance and and detail. Fortunatly, an emerging set of techiques can be employed to design and improve RAG systems. We've focused on taxonomizing and summarizing many of these techniques (see below figure) and will share some high-level strategic guidance in the following sections.
+
+![](/img/rag_landscape.png)
```
rlancemartin, line 769: @jacoblee93

jacoblee93:
Overall I like! Left some more granular comments via Vercel but:

- Would be nice to share more guidance on concrete situations when you might use one technique over the other, or which one you would try first - it's ok to be opinionated in this part of the docs.
- Links to videos are really nice but links to other how-to guides/LangGraph pages would be good too

rlancemartin:
> Overall I like! Left some more granular comments via Vercel but:
> 
> * Would be nice to share more guidance on concrete situations when you might use one technique over the other, or which one you would try first - it's ok to be opinionated in this part of the docs.
> * Links to videos are really nice but links to other how-to guides/LangGraph pages would be good too

Addressed!

jacoblee93:
Still some broken links, not sure why link checker isn't catching?

Would be nice to have some kind of example/guidance around fitting various pieces together at the same time. Like Hyde + MultiVector + MMR or something along with ways to evaluate performance of each step, but I think that can be in the future.

PR #22791, opened 2024-06-11
docs[patch]: Add structured output to conceptual docs

jacoblee93:
This downgrades `Function/tool calling` from a h3 to an h4 which means it'll no longer show up in the right sidebar, but any direct links will still work. I think that is ok, but LMK if you disapprove.

CC @hwchase17 @eyurtsev @rlancemartin

### DIFF
--- a/docs/docs/concepts.mdx
+++ b/docs/docs/concepts.mdx
@@ -710,8 +710,7 @@ See [this guide](/docs/how_to/streaming/#using-stream-events) for more detailed 
 
 ### Structured output
 
-LLMs are capable of generating arbitrary text. This is fantastic
-in general, since it enables the model to respond appropriately to a wide
+LLMs are capable of generating arbitrary text. This enables the model to respond appropriately to a wide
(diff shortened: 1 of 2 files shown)


docs/docs/concepts.mdx
```diff
@@ -744,14 +831,25 @@ LangChain provides a standardized interface for tool calling that is consistent
...
 
-There are two main use cases for function/tool calling:
+The following how-to guides are good practical resources for using function/tool calling:
 
 - [How to return structured data from an LLM](/docs/how_to/structured_output/)
 - [How to use a model to call tools](/docs/how_to/tool_calling/)
 
+For a full list of model providers that support tool calling, [see this table](/docs/integrations/chat/#advanced-features).
+
+#### `.with_structured_output()`
```
hwchase17, line 844: thsi should maybe be more prominently featured since its our recomended way?

jacoblee93: Moved to be the first discussed method.

PR #22813, opened 2024-06-12
ci: test with Python 3.12

cbornet:
We need to use a different version of numpy for py3.8 and py3.12 in pyproject.
And so do projects that use that Python version range and import langchain.

    - **Twitter handle:** _cbornet


eyurtsev:
@efriis

eyurtsev (approved):
w00t!

PR #22830, opened 2024-06-12
docs: `ReAct` reference

leo-gan:
The `ReAct` is used all across LangChain but it is not referenced properly. 
Added references to the original paper.


isahers1:
@baskaryan lgtm

leo-gan:
@baskaryan Thanks!

PR #22792, opened 2024-06-12
core[patch]: allowing latest `packaging` versions

jamesbraza:
Allowing version 24 of https://github.com/pypa/packaging

PR #22835, opened 2024-06-12
core: loosen packaging lib version

efriis:


efriis:
do it in #22792 instead

PR #22839, opened 2024-06-13
community[minor]: Fix long_context_reorder.py async

pprados:
Implement `async def atransform_documents( self, documents: Sequence[Document], **kwargs: Any ) -> Sequence[Document]` for  `LongContextReorder`


isahers1:
This is the same implementation as transform_documents, I think atransform_documents is intended to have a different, async implementation @baskaryan

isahers1:
> This is the same implementation as transform_documents, I think atransform_documents is intended to have a different, async implementation @baskaryan

Actually can we just remove the function? atransform_documents is already implemented in BaseDocumentTransformer

pprados:
The `super()` implementation use:
```

eyurtsev (approved):

PR #22842, opened 2024-06-13
community[minor]: Fix json._validate_metadata_func()

pprados:
JSONparse, in _validate_metadata_func(), checks the consistency of the _metadata_func() function. To do this, it invokes it and makes sure it receives a dictionary in response. However, during the call, it does not respect future calls, as shown on line 100. This generates errors if, for example, the function is like this:
```python
        def generate_metadata(json_node:Dict[str,Any],kwargs:Dict[str,Any]) -> Dict[str,Any]:
             return {
                "source": url,
                "row": kwargs['seq_num'],
                "question":json_node.get("question"),
            }
        loader = JSONLoader(
            file_path=file_path,
            content_key="answer",
            jq_schema='.[]',
            metadata_func=generate_metadata,
            text_content=False)
```
To avoid this, the verification must comply with the specifications. This patch does just that.


libs/core/pyproject.toml
```diff
@@ -40,6 +40,7 @@ optional = true
 jupyter = "^1.0.0"
 setuptools = "^67.6.1"
 grandalf = "^0.8"
+aiosqlite = "^0.19.0"
```
baskaryan, line 43: what's this needed for

baskaryan (commented):
technically we don't specify that the metadata_func will always receive the source and seq_num kwargs anywhere, could view that as an implementation detail that your metadata_func shouldn't rely on

pprados:
the return of metadata_func will be check to be sure if it's a dict.
To that, the code "simulate" a legitimate invocation.
But, If I add a special case, if the kwargs is empty, I just bypass the test.
```
def generate_metadata(json_node:Dict[str,Any],kwargs:Dict[str,Any]) -> Dict[str,Any]:
             if not kwargs:
                      return {} # Bypass the test
             return 43 # And return an integer
```

I think it it not correct.

pprados:
The `_validate_metadata_func()` function is there to test the consistency of the implementation only once.
It's an optimization. The non-optimized alternative would be:
```python
    def _get_metadata(
        self, sample: Dict[str, Any], **additional_fields: Any
    ) -> Dict[str, Any]:
        """
        Return a metadata dictionary base on the existence of metadata_func
        :param sample: single data payload
        :param additional_fields: key-word arguments to be added as metadata values
        :return:
        """
        if self._metadata_func is not None:
            result= self._metadata_func(sample, additional_fields)
            # CHECK HERE
            if not isinstance(result, dict):
                raise ValueError(
                    f"Expected the metadata_func to return a dict but got \
                        `{type(result)}`"
                )
            return result
        else:
            return additional_fields
```
Optimization must not change the implementation. We shouldn't write `generate_metadata()` differently just because there's an optimization.

pprados:
@baskaryan
With this fix, I demonstrate the incorrect code with a more simple implementation.
If you want to test the invocation of a function, you have to invoke it properly.

eyurtsev:
Hi @pprados, 

I haven't been able to follow the description in the PR summary -- so hard to make a decision regarding whether we should merge the PR or not.

From reading the code, the newer version of the code makes more sense to me, but without digging through the code in detail I can't tell whether this change is backwards compatible or not. 

1) Is this change backwards compatible or is it going to break anyones code?
2) Could you remove the commented out code from the PR? We don't want to merge commented code.
3) Are you able to provide a unit test?

pprados:
1. Is this change backwards compatible or is it going to break anyones code?
At present, no code can be written correctly, as I try to show in my explanations. The only way to implement this method is to pretend that everything is fine. So code that checks kwargs will continue to work. And above all, it will now be possible to have a real method, without bypassing the test!
3. Could you remove the commented out code from the PR? We don't want to merge commented code.
Of course you can. I just wanted to explain why the implicit specification requires the presence of `source` and `seq_num`.
5. Are you able to provide a unit test?
Yes.

But, @eyurtsev, do you prefer the last implementation, with a test for each iteration or the first version, with only one test ?

pprados:
@eyurtsev 
is it correct for you ?

eyurtsev:
@pprados this looks good to me. Is there an existing unit test that covers this functionality in the code base already if not would it be possible to add one?

eyurtsev:
Hi @pprados just wanted to check in if you were still interested in merging this change? A unit test would be helpful here to verify that things work as expected

pprados:
Yes, i must find the time for that.

pprados:
@eyurtsev 
There are bugs in the validation of other modules. Community is clean.

eyurtsev (approved):

Issue #22832, opened 2024-06-12
Spec out API for all required vs. all possible input variables


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Context: 

* MessagePlaceholder can be optional or non optional 
* Current LangChain API for input variables doesn't distinguish between possible input variables vs. required input variables

See: https://github.com/langchain-ai/langchain/pull/21640

## Requirements

* get_input_schema() should reflect optional and required inputs
* Expose another property to either fetch all required or all possible input variables (with explanation about why this is the correct approach) alternatively delegate to `get_input_schema()`, and make semantics of input_variables clear (e.g., all possible values)

```python
from langchain import LLMChain
prompt = ChatPromptTemplate.from_messages([MessagesPlaceholder("history", optional=True), ('user', '${input}')])
model = ChatOpenAI()
chain = LLMChain(llm=model, prompt=prompt)
chain({'input': 'what is your name'})
prompt.get_input_schema()
```

keenborder786:
@eyurtsev  since I was already working on this issue in #21640, I can tackle this issue.

eyurtsev:
Merged! Closing this issue!

PR #22851, opened 2024-06-13
core[patch]: Accounting for Optional Input Variables in BasePromptTemplate

keenborder786:
**Description**: After reviewing the prompts API, it is clear that the only way a user can explicitly mark an input variable as optional is through the `MessagePlaceholder.optional` attribute. Otherwise, the user must explicitly pass in the `input_variables` expected to be used in the `BasePromptTemplate`, which will be validated upon execution. Therefore, to semantically handle a `MessagePlaceholder` `variable_name` as optional, we will treat the `variable_name` of `MessagePlaceholder` as a `partial_variable` if it has been marked as optional. This approach aligns with how the `variable_name` of `MessagePlaceholder` is already handled [here](https://github.com/keenborder786/langchain/blob/optional_input_variables/libs/core/langchain_core/prompts/chat.py#L991). Additionally, an attribute `optional_variable` has been added to `BasePromptTemplate`, and the `variable_name` of `MessagePlaceholder` is also made part of `optional_variable` when marked as optional.

Moreover, the `get_input_schema` method has been updated for `BasePromptTemplate` to differentiate between optional and non-optional variables.

**Issue**: #22832, #21425


libs/core/langchain_core/prompts/base.py
```diff
@@ -82,9 +82,14 @@ def get_input_schema(
...
+        required_input_variables = {
+            k: (self.input_types.get(k, str), ...) for k in self.input_variables
+        }
+        optional_input_variables = {
+            k: (self.input_types.get(k, str), None) for k in self.partial_variables
+        }
         return create_model(  # type: ignore[call-overload]
-            "PromptInput",
-            **{k: (self.input_types.get(k, str), None) for k in self.input_variables},
+            "PromptInput", **{**required_input_variables, **optional_input_variables}
```
eyurtsev, line 92: partial_variables are variables that have already been provided. So partial variables should not appear as part of the input schema -- they're unrelated to optional and required variables

libs/core/tests/unit_tests/prompts/test_chat.py
```diff
@@ -709,3 +710,15 @@ async def test_messages_prompt_accepts_list() -> None:
 
     with pytest.raises(TypeError):
         await prompt.ainvoke([("user", "Hi there")])  # type: ignore
+
+
+def test_chat_input_schema() -> None:
```
eyurtsev, line 715: Could you add a unit test that shows the input variables for a chat prompt template for both scenarios?

keenborder786: okay

libs/core/langchain_core/prompts/base.py
keenborder786: I am going to add a new attribute to represent the optional variables.

keenborder786:
@eyurtsev please check now

libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -346,296 +346,8 @@ async def typed_async_lambda_impl(x: str) -> int:
     assert chat_prompt.input_schema.schema() == {
         "title": "PromptInput",
         "type": "object",
-        "properties": {
```
eyurtsev, line 349: Why did this change?

keenborder786: Because of the change `get_input_schema`.

keenborder786:
Please let me know if everything is looking good @eyurtsev ?

keenborder786:
@eyurtsev any updates on this one?

libs/core/tests/unit_tests/runnables/test_runnable.py
eyurtsev: I meant more of does the change look correct. In this case it looks like the change is introducing an unintentional bug? (It's usually a bad sign to see such a big change in a unit test.)

get_input_schema() is expected to provideda precise JSON schema description of what the variable `history` is (it's a list of messages for a chat prompt template). After this PR, this looks like it's no longer true -- which means that the change introduces a new bug

keenborder786: Oh yes I incidently removed the `input_types` update in root_validator for `ChatPromptTemplate` which was causing this. But now it is fixed.

libs/core/langchain_core/prompts/base.py
```diff
@@ -105,9 +107,14 @@ def get_input_schema(
         self, config: Optional[RunnableConfig] = None
     ) -> Type[BaseModel]:
         # This is correct, but pydantic typings/mypy don't think so.
+        required_input_variables = {
+            k: (self.input_types.get(k, str), ...) for k in self.input_variables
```
eyurtsev, line 111: This seems to adopt the convention that `input_variables` represent required variables rather than all possible inputs, which I think is inconsistent with the behavior in the message history?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -231,7 +231,7 @@ def input_variables(self) -> List[str]:
         Returns:
             List of input variable names.
         """
-        return [self.variable_name] if not self.optional else []
+        return [self.variable_name]
```
eyurtsev, line 234: Here it looks like `input_variables` represents all possible values into the template rather than only required values (which is opposite: 0.2.11  to the convention above).

libs/core/langchain_core/prompts/base.py
```diff
@@ -44,6 +44,8 @@ class BasePromptTemplate(
 
     input_variables: List[str]
     """A list of the names of the variables the prompt template expects."""
```
eyurtsev, line 46: Could you update the doc-string for input_variables to document how it should be interpreted?

eyurtsev:
Left two comments in the PR where where the code does not endow the same semantics on the same variable. Let's document the `input_variables` attributes well and clarify its semantics and then verify that the semantics are consistent across the different abstractions.

libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -1111,6 +1117,7 @@ def test_configurable_fields() -> None:
             "lang": {"title": "Lang", "type": "string"},
             "name": {"title": "Name", "type": "string"},
         },
+        "required": ["lang", "name"],
```
eyurtsev, line 1120: Great! This looks like it's improving the schema that we get as its more precise about what's required and what's optional!

libs/core/langchain_core/prompts/base.py
keenborder786: Is it better?

keenborder786: Is it any better?

libs/core/langchain_core/prompts/base.py
keenborder786: No, it is consistent. If the message history is declared as optional, then it will only be in `optional_variables` and not in `required_input_variables`. However, if the message history is not declared as optional, then it will be in `input_variables` and therefore in `required_input_variables`. `input_variables` will always show the variables that are REQUIRED in the prompt.

libs/core/langchain_core/prompts/chat.py
keenborder786: Yeap this is fixed. I see your point

keenborder786:
@eyurtsev please check above comments

eyurtsev:
@keenborder786 looks good at high level, i'll dig a bit deeper on Friday to verify that nothing else is affected , and if looks good will merge. Thanks for your help on this!

eyurtsev (approved):

eyurtsev:
@keenborder786 Looks good -- I cleaned up redefinitions of input_variables in favor of the single one that's inherited  -- it'll carry the updated doc-string to all `input_variables`

PR #22854, opened 2024-06-13
openai[patch]: add stream_usage parameter

ccurme:
Here we add `stream_usage` to ChatOpenAI as:

1. a boolean attribute
2. a kwarg to _stream and _astream.

Question: should the `stream_usage` attribute be `bool`, or `bool | None`?

Currently I've kept it `bool` and defaulted to False. It was implemented on [ChatAnthropic](https://github.com/langchain-ai/langchain/blob/e832bbb48627aa9f00614e82e7ace60b7d8957c6/libs/partners/anthropic/langchain_anthropic/chat_models.py#L535) as a bool. However, to maintain support for users who access the behavior via OpenAI's `stream_options` param, this ends up being possible:
```python
llm = ChatOpenAI(model_kwargs={"stream_options": {"include_usage": True}})
assert not llm.stream_usage
```
(and this model will stream token usage).

Some options for this:
- it's ok
- make the `stream_usage` attribute bool or None
- make an \_\_init\_\_ for ChatOpenAI, set a `._stream_usage` attribute and read `.stream_usage` from a property

Open to other ideas as well.


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1463,6 +1468,43 @@ def is_lc_serializable(cls) -> bool:
...
+    ) -> bool:
+        """Determine whether to include usage metadata in streaming output.
+
+        For backwards compatibility, we check for `stream_options` passed
+        explicitly to kwargs or in the model_kwargs and override self.stream_usage.
+        """
+        stream_usage_sources = [  # order of preference
+            stream_usage,
+            kwargs.get("stream_options", {}).get("include_usage"),
+            self.model_kwargs.get("stream_options", {}).get("include_usage"),
```
baskaryan, line 1482: what about self.stream_usage

ccurme: it's below in the return statement. all the values in this list are `bool | None`.

baskaryan: whats wrong with having a bool in the list

ccurme: Added to the list, the linter just gets confused and requires a return statement outside the loop. I felt like `self.stream_usage` is the best default return so it appears there as well.

baskaryan (approved):

PR #22859, opened 2024-06-13
core[minor]: BaseChatModel with_structured_output implementation

bracesproul:


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -950,6 +958,202 @@ def bind_tools(
     ) -> Runnable[LanguageModelInput, BaseMessage]:
         raise NotImplementedError()
 
+    def with_structured_output(
+        self,
+        schema: Union[Dict, Type[BaseModel]],
+        *,
+        method: Literal["function_calling", "json_mode"] = "function_calling",
```
baskaryan, line 965: dunno if this should be on base implementation

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -950,6 +958,202 @@ def bind_tools(
...
+
+                If schema is a Pydantic class then _DictOrPydantic is the Pydantic
+                    class.
+
+                If schema is a dict then _DictOrPydantic is a dict.
+
+        Example: Function-calling, Pydantic schema (method="function_calling", include_raw=False):
+            .. code-block:: python
+
+                from langchain_provider import ChatModel
```
baskaryan, line 1010: this is kinda weird

libs/core/langchain_core/language_models/chat_models.py
bracesproul: dropped

libs/core/langchain_core/language_models/chat_models.py
bracesproul: Dropped in favor of no import and the implementation uses `ChatModel` as the class name

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -976,6 +980,191 @@ def bind_tools(
...
+                    "What's heavier a pound of bricks or a pound of feathers?"
+                )
+                # -> {
+                #     'raw': AIMessage(content='{\n    "answer": "They are both the same weight.",\n    "justification": "Both a pound of bricks and a pound of feathers weigh one pound. The difference lies in the volume and density of the materials, not the weight." \n}'),
+                #     'parsed': {
+                #         'answer': 'They are both the same weight.',
+                #         'justification': 'Both a pound of bricks and a pound of feathers weigh one pound. The difference lies in the volume and density of the materials, not the weight.'
+                #     },
+                #     'parsing_error': None
+                # }
```
baskaryan, line 1127: delete json mode examples

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -976,6 +980,191 @@ def bind_tools(
...
+
+        if self.bind_tools is BaseChatModel.bind_tools:
+            raise NotImplementedError(
+                "with_structured_output is not implemented for this model."
+            )
+        if schema is None:
+            raise ValueError(
+                "schema must be specified when method is 'function_calling'. "
+                "Received None."
+            )
```
baskaryan, line 1145: not needed

libs/experimental/langchain_experimental/llms/ollama_functions.py
```diff
@@ -151,33 +148,13 @@ def bind_tools(
     ) -> Runnable[LanguageModelInput, BaseMessage]:
         return self.bind(functions=tools, **kwargs)
 
-    @overload
```
baskaryan, line 154: why remove the overloads?

in fact, could we move these to base chat model?

baskaryan (commented):
few last nits, overall lgtm!

libs/experimental/langchain_experimental/llms/ollama_functions.py
bracesproul: Adding those overloads to the base class causes tons of lint errors in already implemented classes. Because of that, I'm going to not add it since I'm worried about breaking typing for existing users.

PR #22860, opened 2024-06-13
experimental[major]: Force users to opt-in into code that relies on the python repl

eyurtsev:
This should make it obvious that a few of the agents in langchain experimental rely on the python REPL as a tool under the hood, and will force users to opt-in.


efriis (approved):

weider69weider69 (commented):
[weider69weider69@yandex.ru](**image**)

weider69weider69 (commented):
Go

PR #22867, opened 2024-06-13
infra: lint new docs to match doc loader template

isahers1:


baskaryan (approved):
whoot! should probably wait to land till after we've got the template landed

PR #22870, opened 2024-06-13
(RFC) core: add RemoveMessage

vbarda:
This change adds a new message type `RemoveMessage` (and `RemoveMessageChunk`) to allow the following behaviors:

(1) in `langhcain_core` -- when using model w/ fallbacks, if one of the models fails mid-stream, we would want to roll back chunks already streamed
(2) in `langgraph` -- to allow user or graph node to manually modify the state

Example usage:

(1) `langchain_core` -- `RunnableWithFallbacks`

```python
from langchain_core.language_models.fake_chat_models import FakeListChatModel
from langchain_core.runnables.fallbacks import RunnableWithFallbacks
from langchain_core.callbacks import StdOutCallbackHandler, CallbackManager, BaseCallbackHandler


class CustomCallbackHandler(BaseCallbackHandler):
    def on_chain_end(self, outputs, **kwargs):
        print("Finished chain", outputs)


callback_manager = CallbackManager([StdOutCallbackHandler()])
callbacks = [CustomCallbackHandler()]

fake_list_chat_model_1 = FakeListChatModel(name="model1", responses=["one", "two"], error_on_chunk_number=2)
fake_list_chat_model_2 = FakeListChatModel(name="model2", responses=["three", "four"])

llm_with_fallbacks = fake_list_chat_model_1.with_fallbacks([fake_list_chat_model_2], exceptions_to_handle=(Exception,)).with_config(callbacks=callbacks)

for s in llm_with_fallbacks.stream("hi"):
    print(s.__class__.__name__, s.content, s.id)
```

Output:

```
AIMessageChunk o run-4b83315c-1ffc-4474-818a-2efb3811ac38
AIMessageChunk n run-4b83315c-1ffc-4474-818a-2efb3811ac38
RemoveMessageChunk modifier run-4b83315c-1ffc-4474-818a-2efb3811ac38
AIMessageChunk t run-2208adaf-26a7-43ce-b203-8a279626cacd
AIMessageChunk h run-2208adaf-26a7-43ce-b203-8a279626cacd
AIMessageChunk r run-2208adaf-26a7-43ce-b203-8a279626cacd
AIMessageChunk e run-2208adaf-26a7-43ce-b203-8a279626cacd
AIMessageChunk e run-2208adaf-26a7-43ce-b203-8a279626cacd
Finished chain content='three' id='run-2208adaf-26a7-43ce-b203-8a279626cacd'
```

(2) `langgraph`

- allow users to delete messages from state by calling

```python
graph.update_state(config, values=[RemoveMessage(id=state.values[-1].id)])
```

- allow nodes to delete messages

```python
graph.add_node("delete_messages", lambda state: [RemoveMessage(id=state[-1].id)])
```


eyurtsev:
How is it parametrized / used?

vbarda:
> How is it parametrized / used?

working on the full change! this will be used in two places:

(1) in `langhcain_core` -- when using model w/ fallbacks, if one of the models fails mid-stream, we would want to roll back chunks already streamed
(2) in `langgraph` -- to allow user to manually modify the state

libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -454,43 +455,43 @@ def stream(
...
-                try:
-                    output = output + chunk  # type: ignore
-                except TypeError:
-                    output = None
-        except BaseException as e:
-            run_manager.on_chain_error(e)
-            raise e
         run_manager.on_chain_end(output)
 
     async def astream(
```
vbarda, line 497: note: this still needs updating

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -254,7 +265,14 @@ def stream(
...
                     else:
                         generation += chunk
+
+                    if chunk.message.id != last_chunk_id:
+                        seen_chunk_ids.append(chunk.message.id)
+                        last_chunk_id = chunk.message.id
                 assert generation is not None
+            except self.exceptions_to_handle:
+                for chunk_id in seen_chunk_ids:
+                    yield RemoveMessageChunk(id=chunk_id)  # type: ignore[call-arg]
```
baskaryan, line 275: i think we'd still want to raise an error at the end of this? otherwise the user code has now way of knowing that nothing has actually streamed

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -120,6 +122,13 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
     callback_manager: Optional[BaseCallbackManager] = Field(default=None, exclude=True)
     """[DEPRECATED] Callback manager to add to the run trace."""
 
+    exceptions_to_handle: Tuple[Type[BaseException], ...] = ()
```
baskaryan, line 125: do we need this? could just always yield remove messages on error (assuming we always raise error anyways)

vbarda: yea if we always raise we don't need!

libs/core/langchain_core/language_models/chat_models.py
vbarda: yes, agreed

baskaryan (commented):
dunno if this is a good idea but: we could add the message removal logic to BaseChatModel. ie check for RemoveMessages in the input and perform specified removals before passing messages on. that way user doesn't have to write this logic everywhere

libs/core/langchain_core/messages/modifier.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any, List, Literal
+
+from langchain_core.messages.base import BaseMessage, BaseMessageChunk
+
+
+class ModifierMessage(BaseMessage):
+    """Message responsible for modifying other messages (deleting / updating.)"""
+
+    def __init__(self, id: str, **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
vbarda: i considered this -- there is an issue w/ `content` field during serialization/deserialization. since it's required on the base message, we still need to pass it here at deserialization. i can raise for any other keys though!

libs/core/langchain_core/messages/modifier.py
```diff
@@ -0,0 +1,38 @@
...
+        return super().__init__("modifier", id=id)
```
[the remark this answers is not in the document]
vbarda: yea, can just be an empty string, wasn't sure which is less confusing `""`

vbarda:
> dunno if this is a good idea but: we could add the message removal logic to BaseChatModel. ie check for RemoveMessages in the input and perform specified removals before passing messages on. that way user doesn't have to write this logic everywhere

i don't think this would always be possible? let's say a user already consumed the streamed chunks from the LLM/runnable in some sort of UI -- they would still need to check if the next chunk in the stream is a `RemoveMessage(Chunk)`, and then reset the state in the UI

but maybe i am misreading your proposal

libs/core/langchain_core/messages/modifier.py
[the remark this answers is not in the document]
eyurtsev: If we don't need the content, do we want to change the hierarchy so that content is only for messages that contain content? Or could the content be helpful here?

eyurtsev:
Generally looks good!

A few questions:

* A user that hits a network request while streaming will get the usual exception + callback together with a list of RemoveMessage -- do we think this will be confusing/surprising?
* What happens if the user is streaming: `model | parser`? Will this logic end up creating another exception at the parsing stage and hide the fact that there was a network error? 
* Could document the schema with more details so folks can understand semantics?
* Is there any other entity except for chat model or LLM that we'd want to emit a RemoveMessage?

libs/core/langchain_core/messages/modifier.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any, List, Literal
+
+from langchain_core.messages.base import BaseMessage, BaseMessageChunk
+
+
+class ModifierMessage(BaseMessage):
```
nfcampos, line 6: what's the point of this base class?

libs/core/langchain_core/messages/modifier.py
```diff
@@ -0,0 +1,38 @@
...
+    @classmethod
+    def get_lc_namespace(cls) -> List[str]:
+        """Get the namespace of the langchain object."""
+        return ["langchain", "schema", "messages"]
+
+
+RemoveMessage.update_forward_refs()
+
+
+class RemoveMessageChunk(RemoveMessage, BaseMessageChunk):
```
nfcampos, line 27: what purpose does this one serve? If we do want a chunk variant it should do something smart on `+` ?

libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -454,43 +455,44 @@ def stream(
...
+                            output = chunk
+                        else:
+                            output = output + chunk  # type: ignore
+                    except TypeError:
+                        output = None
             except self.exceptions_to_handle as e:
-                first_error = e if first_error is None else first_error
                 last_error = e
+                output = None
+                if not isinstance(chunk, RemoveMessage):
```
nfcampos, line 489: why do we not raise an error if chunk is remove message?

libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -454,43 +455,44 @@ def stream(
...
-                    input,
-                    patch_config(config, callbacks=run_manager.get_child()),
-                    **kwargs,
-                )
-                chunk = next(stream)
+
+                for chunk in stream:
+                    yield cast(Output, chunk)
+
+                    if isinstance(chunk, RemoveMessage):
```
nfcampos, line 475: I think `__radd__` of RemoveMessageChunk should implement this behavior no? Otherwise what's the point of the chunk variant

libs/core/langchain_core/runnables/fallbacks.py
vbarda: cause in that case we would not get to the next runnable in the fallbacks when streaming, which i think is actually a desired behavior. otherwise fallback will never work on streaming and we'll just raise an error

libs/core/langchain_core/runnables/fallbacks.py
vbarda: which behavior?

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -254,8 +257,14 @@ def stream(
                         generation = chunk
                     else:
                         generation += chunk
+
+                    if chunk.message.id != last_chunk_id:
+                        seen_chunk_ids.append(chunk.message.id)
```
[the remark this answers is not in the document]
vbarda: set makes sense conceptually, but i think we might want to maintain the order?

libs/core/langchain_core/messages/modifier.py
vbarda: mainly to please the `BaseChatModel` typing, as it requires `BaseMessageChunk` in streaming, otherwise doesn't really make sense i agree. i think it's fine to just use `RemoveMessage` and ignore typecheck for that line

vbarda:
> * A user that hits a network request while streaming will get the usual exception + callback together with a list of RemoveMessage -- do we think this will be confusing/surprising?

I think this is fine, as long as it's properly documented. I wonder if we can just control this with a kwarg? Something like `emit_remove_message_on_error` or the like

> * What happens if the user is streaming: `model | parser`? Will this logic end up creating another exception at the parsing stage and hide the fact that there was a network error?

Exceptions work fine w/ chaining parsers, i.e. the original exception is raised. however, if we use `StrOutputParser()`, we end up with a "parsed" version of `RemoveMessage`, which is probably not what we want. I almost wonder if it's just cleaner to roll this logic directly into `RunnableWithFallbacks` (which is what I had originally)

> * Could document the schema with more details so folks can understand semantics?

Definitely!

> * Is there any other entity except for chat model or LLM that we'd want to emit a RemoveMessage?

I could see something like this being useful for Runnables in general, like handle failures of output parsers or something. so perhaps this is another argument for handling this directly in `RunnableWithFallbacks`

vbarda:
closing in favor of #23636

PR #22889, opened 2024-06-14
community[patch]: set tool name for tongyi&qianfan llm

mackong:
- **Description:** The name of ToolMessage is default to None, which makes tool message send to LLM likes
 ```json
{"role": "tool",
   "tool_call_id": "",
   "content": "{\"time\": \"12:12\"}",
   "name": null}
```
But the name seems essential for some LLMs like TongYi Qwen. so we need to set the name use agent_action's tool value.
  - **Issue:** N/A
  - **Dependencies:** N/A

### DIFF
--- a/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
+++ b/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
@@ -65,7 +65,7 @@ def convert_message_to_dict(message: BaseMessage) -> dict:
         message_dict = {
             "role": "function",
             "content": message.content,
-            "name": message.name,
+            "name": message.name or message.additional_kwargs.get("name"),
         }
     else:
         raise TypeError(f"Got unknown type {message}")
--- a/libs/community/langchain_community/chat_models/tongyi.py
+++ b/libs/community/langchain_community/chat_models/tongyi.py
@@ -201,7 +201,7 @@ def convert_message_to_dict(message: BaseMessage) -> dict:
             "role": "tool",
             "tool_call_id": message.tool_call_id,
             "content": message.content,
-            "name": message.name,
+            "name": message.name or message.additional_kwargs.get("name"),
         }
     elif isinstance(message, FunctionMessage):
         message_dict = {


eyurtsev:
@ccurme you probably have the most insight here -- after reviewing would you be willing to update the doc-string https://github.com/langchain-ai/langchain/blob/1fe5f102c491368f707f7ea738f180edf71add01/libs/core/langchain_core/messages/base.py#L34-L34
to define the semantics for name since I'm not sure when it should and should not be used

mackong:
@ccurme please have a look at this PR

ccurme:
Hi @mackong, thanks for the ping.

My first thought would be to populate the `tool_call_id` field on ToolMessages for models like Tongyi Qwen, to keep them consistent with other chat models. One reason this is important is because it would fix the problem across agents-- e.g., for [langgraph](https://github.com/langchain-ai/langgraph) agents and other AgentExecutor agent types.

- Tool calls have a `name` field to tell the runtime what tool to execute;
- Tool calls and tool messages have ID fields so that the tool calls can be matched to their results.

One reason that these are different is to support parallel tool calling (e.g., calling one tool with two different sets of arguments).

How does Tongyi Qwen link tool calls to results? If it is via the `name` field, then can we use this for the value of `tool_call_id`?

Concretely, something like `tool_call_id=_dict.get("tool_call_id") or _dict.get("name")` when creating ToolMessages:
https://github.com/langchain-ai/langchain/blob/daf733b52e6362513fbc2634507a6b2769fad3e0/libs/community/langchain_community/chat_models/tongyi.py#L146

and something like `"name": message.name or message.tool_call_id,` when creating message dicts from ToolMessages:
https://github.com/langchain-ai/langchain/blob/daf733b52e6362513fbc2634507a6b2769fad3e0/libs/community/langchain_community/chat_models/tongyi.py#L198-L204

I'm not as familiar with Tongyi Qwen so let me know if this doesn't make sense.

mackong:
@ccurme thank you for your detailed explanation about the name&id's usage of ToolMessage, I misunderstood before.

You are right, we can just process the tool's name for needed llm, instead process in  langchain core.

PR is changed just for tongyi and qianfan llm, please have a look again.

ccurme (approved):

PR #22895, opened 2024-06-14
SemanticChunker : Feature Addition ("Semantic Splitting with gradient")

rrajp:
```SemanticChunker``` currently provide three methods to split the texts semantically:
- percentile
- standard_deviation
- interquartile

I propose new method ```gradient```. In this method, the gradient of distance is used to split chunks along with the percentile method (technically) . This method is useful when chunks are highly correlated with each other or specific to a domain e.g. legal or medical. The idea is to apply anomaly detection on gradient array so that the distribution become wider and easy to identify boundaries in highly semantic data.
I have tested this merge on a set of 10 domain specific documents (mostly legal).

Details : 
    - **Issue:** Improvement
    - **Dependencies:** NA
    - **Twitter handle:** [x.com/prajapat_ravi](https://x.com/prajapat_ravi)

@hwchase17

### CODE REVIEW

File: docs/docs/how_to/semantic-chunker.ipynb [hunk 1]
```diff
@@ -297,13 +297,85 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "outputs": [],
+   "source": [],
+   "metadata": {
+    "collapsed": false
+   },
+   "id": "256b9e486f54663d"
```
isahers1, line 378: Can you delete the empty code boxes at the bottom of the doc

libs/experimental/langchain_experimental/text_splitter.py
```diff
@@ -127,23 +128,30 @@ def __init__(
...
+            ), distances
         elif self.breakpoint_threshold_type == "interquartile":
             q1, q3 = np.percentile(distances, [25, 75])
             iqr = q3 - q1
 
-            return np.mean(distances) + self.breakpoint_threshold_amount * iqr
+            return np.mean(distances) + self.breakpoint_threshold_amount * iqr, distances
+        elif self.breakpoint_threshold_type == "gradient":
+            # Calculate the threshold based on the distribution of gradient of distance array.
+            distance_gradient = np.gradient(distances, list(range(0, len(distances))))
```
isahers1, line 150: No need for the list(range... part

isahers1:
@baskaryan two small nits, and even though he changed function signature, it only gets called in class and he updated that call

libs/experimental/langchain_experimental/text_splitter.py
rrajp: Yeah, Fixed that. Thanks for suggestion

docs/docs/how_to/semantic-chunker.ipynb
rrajp: Sorry, Missed that.
Deleted!

rrajp:
@baskaryan Please review this. Let me know if there is any conflict or issue.

PR #22903, opened 2024-06-14
community[patch]: SitemapLoader restrict depth of parsing sitemap (CVE-2024-2965)

eyurtsev:
This PR restricts the depth to which the sitemap can be parsed.

Fix for: CVE-2024-2965


efriis (approved):

vmesel:
@eyurtsev sorry to bother you here, but it looks like this approach doesn't solve the issue, shouldn't we check if the URL was already parsed as well?

PR #22914, opened 2024-06-14
[minor]: add parallel tool calls as param to .bind_tools

isahers1:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -817,6 +817,7 @@ def bind_tools(
         tool_choice: Optional[
             Union[dict, str, Literal["auto", "none", "required", "any"], bool]
         ] = None,
+        parallel_tool_calls: Optional[bool] = None,
```
baskaryan, line 820: since we're not doing anything with this param (just plumbing through kwargs) we actually don't even need to add it to function signature

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1283,13 +1287,13 @@ class GetPopulation(BaseModel):
 
         Note that ``openai >= 1.32`` supports a ``parallel_tool_calls`` parameter
         that defaults to ``True``. This parameter can be set to ``False`` to
-        disable parallel tool calls:
+        disable parallel tool calls. You can pass this argument into .bind_tools:
```
baskaryan, line 1290: ah didn't realize we already had documented this!

docs/docs/how_to/tool_calling.ipynb
baskaryan, line 1: shouldn't need to explicitly pip install openai (langchain-openai depends on openai)

baskaryan: instead of accessing additional_kwargs we should access tool_calls

PR #22919, opened 2024-06-14
[docs]: split up tool docs

isahers1:


docs/docs/how_to/tool_results_pass_to_model.ipynb
baskaryan, line 1: nit: top header should have a single `#` (to indicate H1) and for how tos should always have "How to ..." format

docs/docs/how_to/tools_pass_to_model.ipynb
baskaryan, line 1: nit: could probably just leave this file as being called `tool_calling`. that way we also don't have to add a redirect for existing links to https://python.langchain.com/v0.2/docs/how_to/tool_calling/

docs/docs/how_to/tool_calling.ipynb
baskaryan, line 1: - would split tool_choice bit into its own page
- would link to the other tool calling docs at the bottom of this page (since itll be the main tool calling page)

docs/docs/how_to/tool_streaming.ipynb
baskaryan, line 1: nit: capitalization (How to stream tool calls)

docs/docs/how_to/tools_few_shot.ipynb
baskaryan, line 1: nit: might be clearer to title "How to use few-shot prompting with tool calling"

docs/docs/how_to/tools_model_specific.ipynb
baskaryan, line 1: mm this doc is kind of confusing. i think the API ref for each model should make it clear what formats are supported for bind_tools, and we can mention that folks should check the api ref for their models bind_tools in the main tool calling docs. thoughts on deleting this page @ccurme

baskaryan (commented):
nice! last comments

docs/docs/how_to/tools_model_specific.ipynb
isahers1: Happy to do whatever. I've included it in the latest commit but lmk if I should remove.

baskaryan:
one last nit, could we put disabling parallel tool calls in its own how to? otherwise lgtm!

Issue #22804, opened 2024-06-12
ChatHuggingFace using local model with HuggingFacePipeline wrongly checks for available inference endpoints


c-pletinckx:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_huggingface import (
    ChatHuggingFace,
    HuggingFacePipeline,
)

chat_llm = ChatHuggingFace(
    llm=HuggingFacePipeline.from_model_id(
        model_id="path/to/your/local/model",  # I downloaded Meta-Llama-3-8B
        task="text-generation",
        device_map="auto",
        model_kwargs={"temperature": 0.0, "local_files_only": True},
    )
)
```

### Error Message and Stack Trace (if applicable)

```bash
src/resources/predictor.py:55: in load
    self.llm = ChatHuggingFace(
/opt/poetry-cache/virtualenvs/sagacify-example-llm-8EXZSVYp-py3.10/lib/python3.10/site-packages/langchain_huggingface/chat_models/huggingface.py:169: in __init__
    self._resolve_model_id()
/opt/poetry-cache/virtualenvs/sagacify-example-llm-8EXZSVYp-py3.10/lib/python3.10/site-packages/langchain_huggingface/chat_models/huggingface.py:295: in _resolve_model_id
    available_endpoints = list_inference_endpoints("*")
/opt/poetry-cache/virtualenvs/sagacify-example-llm-8EXZSVYp-py3.10/lib/python3.10/site-packages/huggingface_hub/hf_api.py:7081: in list_inference_endpoints
    user = self.whoami(token=token)
/opt/poetry-cache/virtualenvs/sagacify-example-llm-8EXZSVYp-py3.10/lib/python3.10/site-packages/huggingface_hub/utils/_validators.py:114: in _inner_fn
    return fn(*args, **kwargs)
/opt/poetry-cache/virtualenvs/sagacify-example-llm-8EXZSVYp-py3.10/lib/python3.10/site-packages/huggingface_hub/hf_api.py:1390: in whoami
    headers=self._build_hf_headers(
[...]

PR #22925, opened 2024-06-14
[Community [HuggingFace]]: Small Fix for ChatHuggingFace.

keenborder786:
- **Description:** A small fix where I moved the `available_endpoints` in order to avoid the token error in the below issue. Also I have added conftest file and updated the `scripy`,`numpy` versions to support newer python versions in poetry files.
- **Issue:** #22804


efriis (approved):

libs/partners/huggingface/langchain_huggingface/tests/__init__.py
efriis, line 1: Actually we want to keep these as python modules: https://docs.pytest.org/en/stable/explanation/pythonpath.html

keenborder786: Why can't we keep `conftest.py` file?

libs/partners/huggingface/tests/conftest.py
keenborder786, line 1: @efriis I have added this which I feel like is much better for appending huggingface  package to sys path.

keenborder786:
@efriis please review now!!!

keenborder786:
@efriis please review now!!!

libs/partners/huggingface/tests/conftest.py
hwchase17: we have standardized on the init files as erick suggested. if we want to do this, we should do everywhere. and we should open up a discussion item before doing that

keenborder786:
@efriis can you please merge since this error is creeping up, I faced the same error when I was trying to resolve the following issue: #22952

libs/partners/huggingface/pyproject.toml
```diff
@@ -18,6 +18,17 @@ transformers = ">=4.39.0"
 sentence-transformers = ">=2.6.0"
 huggingface-hub = ">=0.23.0"
 
+# Support Python 3.8 and 3.12+.
```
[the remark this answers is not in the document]
keenborder786: It needs to be added here otherwise the build fails when making the python 3.12 environment in CI/CD pipeline

ccurme (approved):

libs/partners/huggingface/pyproject.toml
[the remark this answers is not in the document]
ccurme: Believe this was resolved in https://github.com/langchain-ai/langchain/pull/23197

PR #22942, opened 2024-06-15
community: Add Baichuan Embeddings batch size

maang-h:
- **Support batch size** 
Baichuan updates the document, indicating that up to 16 documents can be imported at a time

- **Standardized model init arg names**
    - baichuan_api_key -> api_key
    - model_name  -> model

### DIFF
--- a/libs/community/langchain_community/embeddings/baichuan.py
+++ b/libs/community/langchain_community/embeddings/baichuan.py
@@ -2,7 +2,7 @@ from typing import Any, Dict, List, Optional
 
 import requests
 from langchain_core.embeddings import Embeddings
-from langchain_core.pydantic_v1 import BaseModel, SecretStr, root_validator
+from langchain_core.pydantic_v1 import BaseModel, Field, SecretStr, root_validator
 from langchain_core.utils import convert_to_secret_str, get_from_dict_or_env
 from requests import RequestException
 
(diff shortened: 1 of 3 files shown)


libs/community/langchain_community/embeddings/baichuan.py
```diff
@@ -37,9 +37,16 @@ class BaichuanTextEmbeddings(BaseModel, Embeddings):
     """
 
     session: Any  #: :meta private:
-    model_name: str = "Baichuan-Text-Embedding"
-    baichuan_api_key: Optional[SecretStr] = None
+    model_name: str = Field(default="Baichuan-Text-Embedding", alias="model")
```
ccurme, line 40: thank you

libs/community/langchain_community/embeddings/baichuan.py
```diff
@@ -37,9 +37,16 @@ class BaichuanTextEmbeddings(BaseModel, Embeddings):
...
+    baichuan_api_key: Optional[SecretStr] = Field(default=None, alias="api_key")
     """Automatically inferred from env var `BAICHUAN_API_KEY` if not provided."""
+    chunk_size: int = 16
```
ccurme, line 43: How was this chosen / compare with limits on the underlying API? Is it possible it's too low (or high)?

libs/community/langchain_community/embeddings/baichuan.py
```diff
@@ -78,26 +85,35 @@ def _embed(self, texts: List[str]) -> Optional[List[List[float]]]:
...
+                BAICHUAN_API_URL, json={"input": chunk, "model": self.model_name}
             )
+            # Raise exception if response status code from 400 to 600
+            response.raise_for_status()
+            # Check if the response status code indicates success
+            if response.status_code == 200:
+                resp = response.json()
+                embeddings = resp.get("data", [])
+                # Sort resulting embeddings by index
+                sorted_embeddings = sorted(embeddings, key=lambda e: e.get("index", 0))
```
ccurme, line 104: Previously, the embeddings were "globally" sorted. Here they are sorted within each chunk. What impact could this have?

ccurme (commented):
Thanks for this!

libs/community/langchain_community/embeddings/baichuan.py
maang-h: The batch size is equal to 16 because according to [Baichuan Embedings docs](https://platform.baichuan-ai.com/docs/text-Embedding), the document specifies that the maximum single input is the number of texts, which is 16.

libs/community/langchain_community/embeddings/baichuan.py
maang-h: I don't think so.
Before:
The maximum number of documents passed in is 16 (any more will result in an error), and sorting is performed for embeddings of less than or equal to 16 documents.
Now:
The sorting of the first 16 documents is also completed, and there is no problem with the extra documents (17-more) being sorted later.

ccurme (approved):

PR #22946, opened 2024-06-15
community[minor]: Add TextEmbed Embedding Integration

kevaldekivadiya2415:
**Description:**

**TextEmbed** is a high-performance embedding inference server designed to provide a high-throughput, low-latency solution for serving embeddings. It supports various sentence-transformer models and includes the ability to deploy image and text embedding models. TextEmbed offers flexibility and scalability for diverse applications.

- **PyPI Package:** [TextEmbed on PyPI](https://pypi.org/project/textembed/)
- **Docker Image:** [TextEmbed on Docker Hub](https://hub.docker.com/r/kevaldekivadiya/textembed)
- **GitHub Repository:** [TextEmbed on GitHub](https://github.com/kevaldekivadiya2415/textembed)

**PR Description**
This PR adds functionality for embedding documents and queries using the `TextEmbedEmbeddings` class. The implementation allows for both synchronous and asynchronous embedding requests to a TextEmbed API endpoint. The class handles batching and permuting of input texts to optimize the embedding process.

**Example Usage:**

```python
from langchain_community.embeddings import TextEmbedEmbeddings

# Initialise the embeddings class
embeddings = TextEmbedEmbeddings(model="your-model-id", api_key="your-api-key", api_url="your_api_url")

# Define a list of documents
documents = [
    "Data science involves extracting insights from data.",
    "Artificial intelligence is transforming various industries.",
    "Cloud computing provides scalable computing resources over the internet.",
    "Big data analytics helps in understanding large datasets.",
    "India has a diverse cultural heritage."
]

# Define a query
query = "What is the cultural heritage of India?"

# Embed all documents
document_embeddings = embeddings.embed_documents(documents)

# Embed the query
query_embedding = embeddings.embed_query(query)

# Print embeddings for each document
for i, embedding in enumerate(document_embeddings):
    print(f"Document {i+1} Embedding:", embedding)

# Print the query embedding
print("Query Embedding:", query_embedding)

### CODE REVIEW

**Comment by dipgabani2204** [comment]
That would be great if textembed will integrate with langchain. It will be very much helpful those people who want to get the high throughput for serving embedding applications.

File: libs/community/langchain_community/embeddings/textembed.py [hunk 1]
```diff
@@ -0,0 +1,390 @@
...
+            from langchain_community.embeddings import TextEmbedEmbeddings
+            embeddings = TextEmbedEmbeddings(
+                model="sentence-transformers/clip-ViT-B-32",
+                textembed_api_url="http://localhost:8000/v1",
+                textembed_api_key="<API_KEY>"
+            )
+    """
+
+    model: str
+    "Underlying TextEmbed model id."
```
isahers1, line 60: type: triple quotes instead of single quotes

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+
+    client: Any = None
+    """TextEmbed client."""
+
+    class Config:
+        """Configuration for this pydantic object."""
+
+        extra = Extra.forbid
+
+    @root_validator(allow_reuse=True)
```
eyurtsev, line 76: Please use one of these

```suggestion
    @root_validator(pre=True)
```

OR

```suggestion
    @root_validator(pre=False, skip_on_failure=True)
```

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+            )
+    """
+
+    model: str
+    "Underlying TextEmbed model id."
+
+    textembed_api_url: str = "http://localhost:7997"
+    """Endpoint URL to use."""
+
+    textembed_api_key: str = "None"
```
eyurtsev, line 65: Do not assign a default for the API key

```suggestion
    textembed_api_key: str
```

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+import requests
+from langchain_core.embeddings import Embeddings
+from langchain_core.pydantic_v1 import BaseModel, Extra, root_validator
+from langchain_core.utils import get_from_dict_or_env
+
+__all__ = ["TextEmbedEmbeddings"]
+
+
+class TextEmbedEmbeddings(BaseModel, Embeddings):
+    """
```
eyurtsev, line 27: Could you include a URL to relevant information about TextEmbedEmbeddings?

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+                model="sentence-transformers/clip-ViT-B-32",
+                textembed_api_url="http://localhost:8000/v1",
+                textembed_api_key="<API_KEY>"
+            )
+    """
+
+    model: str
+    "Underlying TextEmbed model id."
+
+    textembed_api_url: str = "http://localhost:7997"
```
eyurtsev, line 62: Is 7997 the standard port? If yes/no, could you match it against the code block in the example?

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+        embed_query(self, text: str) -> List[float]:
+            Embeds a single query.
+        aembed_query(self, text: str) -> List[float]:
+            Asynchronously embeds a single query.
+
+    Example:
+        .. code-block:: python
+
+            from langchain_community.embeddings import TextEmbedEmbeddings
+            embeddings = TextEmbedEmbeddings(
```
eyurtsev, line 52: Python convention it to have at least one new line after the import blocks

```suggestion

            embeddings = TextEmbedEmbeddings(
```

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Returns:
+            List[List[float]]: List of embeddings for the texts.
+        """
+        perm_texts, unpermute_func = self._permute(texts)
+        perm_texts_batched = self._batch(perm_texts)
+
+        if self.aiosession is None:
+            self.aiosession = aiohttp.ClientSession(
+                trust_env=True, connector=aiohttp.TCPConnector(limit=32)
```
eyurtsev, line 375: why is `trust_env=True` ?

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+            model=self.model,
+            texts=texts,
+        )
+        return embeddings
+
+    def embed_query(self, text: str) -> List[float]:
+        """Call out to TextEmbed's embedding endpoint for a single query.
+
+        Args:
+            text (str): The text to embed.
```
eyurtsev, line 133: we usually don't document the type in the param strings since we use type annotations

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Returns:
+            List[List[float]]: List of embeddings for the request.
+
+        Raises:
+            Exception: If the response status is not 200.
+        """
+        async with session.post(**kwargs) as response:
+            if response.status != 200:
+                raise Exception(
```
eyurtsev, line 352: `response.raise_for_status()` potentially? instead of changing exception class to general Exception?

libs/community/langchain_community/embeddings/textembed.py
eyurtsev, line 65: You can support an alias if you want, but let's keep the value `api_key`

```suggestion
    api_key: str = "None"
```

eyurtsev:
@kevaldekivadiya2415  Could you add unit tests to the integration? There's a lot of code that could be break during any refactors needed for maintenance. 

The server code itself isn't unit tested as far as I can tell: https://github.com/kevaldekivadiya2415/textembed/tree/main

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,390 @@
...
+            **self._kwargs_post_request(model=model, texts=batch_texts)
+        )
+        if response.status_code != 200:
+            raise Exception(
+                f"extEmbed responded with an unexpected status message "
+                f"{response.status_code}: {response.text}"
+            )
+        return [e["embedding"] for e in response.json()["data"]]
+
+    def embed(self, model: str, texts: List[str]) -> List[List[float]]:
```
eyurtsev, line 304: I'd expect a lot of this code to be handled by the server. Why is this on the client?

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: The trust_env=True parameter in the aiohttp.ClientSession initialisation is used to trust and utilise the environment variables related to the proxy configuration. 
This means that if your environment is set up to use proxies, this parameter ensures that the aiohttp client will respect those settings.

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: resolved

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: resolved

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: resolved

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: Hey @eyurtsev 

Thank you for your feedback. I understand that we typically don't include the type in the parameter strings since we use type annotations. However, I believe including the type in the docstring can be beneficial for a few reasons:

Readability: When someone is reading the docstring, they don't need to look at the function signature to understand the expected types.
Documentation Tools: Some documentation tools generate more detailed and user-friendly documentation when the types are included in the docstring.
Consistency: It can provide a consistent format for those who might not be as familiar with type annotations.

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: Thank you for the suggestion! I agree that using response.raise_for_status() is a cleaner and more specific way to handle HTTP errors.

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: Because the server response and LangChain embedding response are different, the server is compatible with OpenAI requests and responses.

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: resolved

libs/community/langchain_community/embeddings/textembed.py
kevaldekivadiya2415: resolved

kevaldekivadiya2415:
All the changes are complete. The PR is now ready for deployment.

kevaldekivadiya2415:
All the changes are complete. The PR is now ready for deployment.

Sudip-SmartSense:
Interesting idea!! can anyone check on this? Looking forward to seeing this in prod.

isahers1 (approved):

eyurtsev (approved):

libs/community/langchain_community/embeddings/textembed.py
eyurtsev: Up to you, but this makes it difficult to retry on failure based on status code since the status code information is lost.

libs/community/langchain_community/embeddings/textembed.py
```diff
@@ -0,0 +1,356 @@
...
+            texts (List[str]): List of texts to embed.
+
+        Returns:
+            List[List[float]]: List of embeddings for the texts.
+        """
+        perm_texts, unpermute_func = self._permute(texts)
+        perm_texts_batched = self._batch(perm_texts)
+
+        async with aiohttp.ClientSession(
+            trust_env=True, connector=aiohttp.TCPConnector(limit=32)
```
eyurtsev, line 342: I'd prefer if this was opt-in and matched the defaults of aiohttp. I don't understand the security implications: https://docs.aiohttp.org/en/stable/client_advanced.html#aiohttp-client-proxy-support

eyurtsev: @kevaldekivadiya2415 I'm going to disable trust_env and merge and if you want to re-enable, please explain why that makes sense and do it via a separate PR

eyurtsev, line 342:
```suggestion
            connector=aiohttp.TCPConnector(limit=32)
```

PR #22954, opened 2024-06-16
 community: add args_schema to SearxSearch 

Nold360:
This change adds args_schema (pydantic BaseModel) to SearxSearchRun for correct schema formatting on LLM function calls

Issue: currently using SearxSearchRun with OpenAI function calling returns the following error "TypeError: SearxSearchRun._run() got an unexpected keyword argument '__arg1' ". 

This happens because the schema sent to the LLM is "input: '{"__arg1":"foobar"}'" while the method should be called with the "query" parameter.

### DIFF
--- a/libs/community/langchain_community/tools/searx_search/tool.py
+++ b/libs/community/langchain_community/tools/searx_search/tool.py
@@ -11,7 +11,7 @@ from langchain_core.tools import BaseTool
 from langchain_community.utilities.searx_search import SearxSearchWrapper
 
 class SearxSearchQueryInput(BaseModel):
-    """input for the SearxSearch tool"""
+    """Input for the SearxSearch tool."""
     query: str = Field(description="query to look up on searx")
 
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/tools/searx_search/tool.py
```diff
@@ -1,15 +1,18 @@
...
     CallbackManagerForToolRun,
 )
-from langchain_core.pydantic_v1 import Extra, Field
+from langchain_core.pydantic_v1 import Extra, Field, BaseModel
 from langchain_core.tools import BaseTool
 
 from langchain_community.utilities.searx_search import SearxSearchWrapper
 
+class SearxSearchQueryInput(BaseModel):
+    """input for the SearxSearch tool"""
```
hwchase17, line 14:
```suggestion
    """Input for the SearxSearch tool."""
```

hwchase17 (approved):

PR #22993, opened 2024-06-17
docs[patch]: Reorder streaming guide, add tags

jacoblee93:
CC @hinthornw


hinthornw (approved):

docs/docs/concepts.mdx
```diff
@@ -607,6 +607,7 @@ For specifics on how to use callbacks, see the [relevant how-to guides here](/do
 ## Techniques
 
 ### Streaming
+<span data-heading-keywords="stream,streaming"></span>
```
hinthornw, line 610: nice - TIL
Is this docusaurus/algolia generic or do we have a custom script here?

jacoblee93: Custom script that elevates specific headers in search results. Use VERY sparingly

PR #23057, opened 2024-06-17
community: add support for 'cloud' parameter in JiraAPIWrapper

quintinoandre:
- **Description:** Enhance JiraAPIWrapper to accept the 'cloud' parameter through an environment variable. This update allows more flexibility in configuring the environment for the Jira API.
 - **Twitter handle:** Andre_Q_Pereira


libs/community/langchain_community/utilities/jira.py
```diff
@@ -14,6 +15,7 @@ class JiraAPIWrapper(BaseModel):
     jira_username: Optional[str] = None
     jira_api_token: Optional[str] = None
     jira_instance_url: Optional[str] = None
+    jira_cloud: Optional[str] = "True"
```
hwchase17, line 18: can we make this be a boolean? this is the instance on the model. if we load jira_cloud from dict_or_env we can convert from string to boolen there

libs/core/langchain_core/utils/env.py
```diff
@@ -31,10 +31,17 @@ def get_from_dict_or_env(
         return get_from_env(key, env_key, default=default)
 
 
-def get_from_env(key: str, env_key: str, default: Optional[str] = None) -> str:
+def get_from_env(
```
hwchase17, line 34: lets not change for core - thats a bigger decicision

lets just do the string check outside this

libs/community/langchain_community/utilities/jira.py
```diff
@@ -14,6 +15,7 @@ class JiraAPIWrapper(BaseModel):
     jira_username: Optional[str] = None
     jira_api_token: Optional[str] = None
     jira_instance_url: Optional[str] = None
+    jira_cloud: Optional[bool] = None
```
hwchase17, line 18: is this allowed to be None? or will it error when passed into Confluence

quintinoandre: Yes, it's allowed, it doesn't give an error.

Inside the library (confluence.py file) they just do the following check several times:

If self.cloud:
    do something...

Which gives the same result if it's False or None.

libs/core/langchain_core/utils/env.py
quintinoandre: I revert the core changes and did the check outside as you suggest.

libs/community/langchain_community/utilities/jira.py
eyurtsev: Perhaps we could remove the ambiguity here and use bool type w/ a default of False?

quintinoandre: If I set a default value here, either True or False, the environment variable's value is never considered. That's why I set it to None.

eyurtsev (approved):

ccurme (approved):

PR #23061, opened 2024-06-18
docs: add trim_messages to chatbot

baskaryan:


isahers1 (commented):
Idk if there is a way to remove the "Parent run ... treating as root run" warnings in the notebook, but other than that lgtm

PR #23080, opened 2024-06-18
community[patch]: add size for message histories

mackong:
- **Description:** 
Currently some ChatMessageHistory (neo4j, zep_cloud) support limit the size of message history, others are not. So this patch add history_size parameter to other ChatMessageHistory. 
I have tried to use native database `sort`&`limit` capability to do this, but some databases seem not easy to handle it, then python list slice used.
NOTE: for deprecated ChatMessageHistory like AstraDB, I will try make PR to correspending repo.
  - **Issue:** N/A
  - **Dependencies:** N/A

### DIFF
--- a/libs/community/langchain_community/chat_message_histories/cassandra.py
+++ b/libs/community/langchain_community/chat_message_histories/cassandra.py
@@ -39,6 +39,7 @@ class CassandraChatMessageHistory(BaseChatMessageHistory):
         ttl_seconds: Optional[int] = DEFAULT_TTL_SECONDS,
         *,
         setup_mode: SetupMode = SetupMode.SYNC,
+        history_size: Optional[int] = None,
     ) -> None:
         """Chat message history that stores history in Cassandra.
 
@@ -52,6 +53,9 @@ class CassandraChatMessageHistory(BaseChatMessageHistory):
             ttl_seconds: time-to-live (seconds) for automatic expiration
                 of stored entries. None (default) for no expiration.
             setup_mode: mode used to create the Cassandra table (SYNC, ASYNC or OFF).
+            history_size: Maximum number fo messages to retrieve. If None then
+                there is no limit. If not None then only the latest `history_size`
+                messages are retrieved.
         """
         try:
             from cassio.table import ClusteredCassandraTable
@@ -62,6 +66,7 @@ class CassandraChatMessageHistory(BaseChatMessageHistory):
             )
         self.session_id = session_id
         self.ttl_seconds = ttl_seconds
+        self.history_size = history_size
         kwargs: Dict[str, Any] = {}
         if setup_mode == SetupMode.ASYNC:
             kwargs["async_setup"] = True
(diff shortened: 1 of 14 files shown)


eyurtsev:
What should happen with the system message? It's usually the first message? @baskaryan do you want to take lead -- I think it's relevant in context of `trim_messages` functionality.

Some nits (pending @baskaryan 's opinion on this)

* Could you add `*` operator prior to any new named argument added to a function signature?
* In places, where history size is interpolated into the SQL query could you do a run time check on it to be more defensive?

mackong:
> What should happen with the system message? It's usually the first message? @baskaryan do you want to take lead -- I think it's relevant in context of `trim_messages` functionality.
> 
> Some nits (pending @baskaryan 's opinion on this)
> 
> * Could you add `*` operator prior to any new named argument added to a function signature?
> * In places, where history size is interpolated into the SQL query could you do a run time check on it to be more defensive?

@eyurtsev 
1. `*` operator and runtime check fixed.
2. Now, system message is not added as history message. just input messages and output messages is processed. See
https://github.com/langchain-ai/langchain/blob/6b46b5e9ce6aa1bdc61a76e7778ec79afeb5691e/libs/core/langchain_core/runnables/history.py#L484

baskaryan (commented):
imo history size should only be managed in ChatMessageHistory if its about reducing storage load. if its about handling how many messages are passed to model that should be handled outside the history, using eg [trim_messages](https://python.langchain.com/v0.2/docs/how_to/trim_messages/)

mackong:
@baskaryan yes, history size in this PR is about reducing storage load, and can be combined with `trim_messages` to handle token-level limitations further.

baskaryan:
> @baskaryan yes, history size in this PR is about reducing storage load, and can be combined with `trim_messages` to handle token-level limitations further.

if its about storage then shouldn't we be using it in add_message, not load_message? ie when we add a new message, we check if our history is at capacity and delete the oldest message if so

mackong:
> > @baskaryan yes, history size in this PR is about reducing storage load, and can be combined with `trim_messages` to handle token-level limitations further.
> 
> if its about storage then shouldn't we be using it in add_message, not load_message? ie when we add a new message, we check if our history is at capacity and delete the oldest message if so

Here, what I have in mind is that in an LLM application, we may need to retain complete user session data. Therefore, no restriction is applied in the add_message operation.
On the other hand, when constructing a prompt to be given to the LLM, we usually only require a limited number of the latest history records. Therefore, we can utilize the database's limit & sort mechanism to reduce the amount of data to be read, thereby improving the performance of loading historical data. Thus, the restriction is applied in the load_message operation.

eyurtsev:
>> On the other hand, when constructing a prompt to be given to the LLM, we usually only require a limited number of the latest history records.

My concern is that constructing the prompt is often more complex than getting the last n messages; e.g., the most common issue is failing to include the `system message` (which is the first message)

mackong:
@eyurtsev @baskaryan based on the disscussion before, add history_size parameter seems not a good idea, I close the PR now.

PR #23118, opened 2024-06-18
fireworks[patch]: fix api_key alias in Fireworks LLM

shaunakgodbole:

**Description**
The current code snippet for `Fireworks` had incorrect parameters.  This PR fixes those parameters.


eyurtsev (dismissed):
This is expected to work with the newest fireworks package. Could you confirm that it's broken after upgrading to the newest version of the package?

shaunakgodbole:
Thanks Eugene, looking at the [code](https://github.com/langchain-ai/langchain/blob/b483bf509520fafb3b5618717de70e1edf5f2ba7/libs/partners/fireworks/langchain_fireworks/llms.py#L41-L42): 
```
    """Base inference API URL."""
    fireworks_api_key: SecretStr = Field(default=None, alias="api_key")
```

The alias `api_key` should work.  However, when I run the snippet I get the following error:
```
pydantic.v1.error_wrappers.ValidationError: 1 validation error for Fireworks
__root__
  Did not find fireworks_api_key, please add an environment variable `FIREWORKS_API_KEY` which contains it, or pass `fireworks_api_key` as a named parameter. (type=value_error)
```

I am using:
* langchain-core: 0.2.8
* langchain-fireworks: 0.1.3

which are the most recent releases.

ccurme:
@eyurtsev would you mind taking another look at this?

baskaryan (approved):

libs/partners/fireworks/tests/unit_tests/test_llms.py
```diff
@@ -18,6 +18,15 @@ def test_fireworks_api_key_is_secret_string() -> None:
     )
     assert isinstance(llm.fireworks_api_key, SecretStr)
 
+    # Test api_key alias
```
efriis, line 21: can we do this as a standard test instead / as well?

efriis (approved):

PR #23131, opened 2024-06-18
core[patch]: Add an example to the Document schema doc-string

eyurtsev:
Add an example to the document schema


eyurtsev:
Accidentally added to this PR some changes from the messages namespace. Changes look fine so retaining

PR #23147, opened 2024-06-18
openai[patch]: image token counting

### LINKED ISSUE #23000: Add image token counting to ChatOpenAI
### Privileged issue

### Issue Content

_No response_

baskaryan:
Resolves #23000


libs/partners/openai/tests/integration_tests/chat_models/test_base_standard.py
```diff
@@ -20,3 +23,36 @@ def chat_model_params(self) -> dict:
...
+        actual = model.get_num_tokens_from_messages([message])
+        assert expected == actual
+
+        image_data = base64.b64encode(httpx.get(image_url).content).decode("utf-8")
+        message = HumanMessage(
+            content=[
+                {"type": "text", "text": "describe the weather in this image"},
+                {
+                    "type": "image_url",
+                    "image_url": {"url": f"data:image/jpeg;base64,{image_data}"},
```
baskaryan, line 50: worth testing with a non-jpeg image

isahers1: Added png tests, lmk if you want more

libs/partners/openai/tests/integration_tests/chat_models/test_base_standard.py
```diff
@@ -20,3 +23,69 @@ def chat_model_params(self) -> dict:
...
+            content=[
+                {"type": "text", "text": "describe the weather in this image"},
+                {
+                    "type": "image_url",
+                    "image_url": {"url": image_url},
+                },
+            ],
+        )
+        expected = cast(AIMessage, model.invoke([message])).usage_metadata[  # type: ignore[index]
+            "input_tokens"
```
ccurme, line 39: nice!

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -753,17 +763,35 @@ def get_num_tokens_from_messages(self, messages: List[BaseMessage]) -> int:
...
         messages_dict = [_convert_message_to_dict(m) for m in messages]
         for message in messages_dict:
             num_tokens += tokens_per_message
             for key, value in message.items():
-                # Cast str(value) in case the message value is not a string
-                # This occurs with function messages
-                num_tokens += len(encoding.encode(str(value)))
+                if isinstance(value, list):
+                    for val in value:
+                        if val["type"] == "text" or isinstance(val, str):
```
ccurme, line 776:
```suggestion
                        if isinstance(val, str) or val["type"] == "text":
```
Would this raise an error as-is if `val` is a str?

libs/partners/openai/tests/integration_tests/chat_models/test_base_standard.py
```diff
@@ -20,3 +23,69 @@ def chat_model_params(self) -> dict:
     @property
     def supports_image_inputs(self) -> bool:
         return True
+
+    def test_image_token_counting_jpeg(self, model: BaseChatModel) -> None:
```
ccurme, line 27: (nit) typically standard tests test methods that are implemented on `ChatModelIntegrationTests`. If we think other providers will support estimating image token counts then I think it makes sense to leave here as a "todo" to eventually migrate to langchain-standard-tests. Otherwise could make more sense in tests/integration_tests/chat_models/test_base.py.

ccurme (commented):
Thanks for this! LGTM pending question about string values.

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: lol yea good catch. i think technically by the time we're here it can't be a string anyways, but seems worth doing this for future proofing

libs/partners/openai/tests/integration_tests/chat_models/test_base_standard.py
baskaryan: not sure any other models have oss tokenizers atm so seems premature to add to standard tests, but down to leave a todo

PR #23152, opened 2024-06-19
prompty: docstring

leo-gan:
Added missed docstrings. Format docstrings to the consistent format (used in the API Reference)


leo-gan:
@efriis Could you, please, check the PR errors? It is something with dependencies and unrelated to the PR changes.
Thanks!

ccurme:
Thanks @leo-gan. Python 3.12 was added to CI recently but some packages (e.g,. partner packages) haven't had CI run yet. A common error is that we need `numpy >=1.26.0` for python 3.12. I believe packages that directly depend on numpy have been updated but for some we may need to re-lock.

ccurme (approved):

PR #23154, opened 2024-06-19
core[patch]: Document messages namespace

eyurtsev:
- Moved doc-strings below attribtues in TypedDicts -- seems to render better on APIReference pages.
* Provided more description and some simple code examples


libs/core/langchain_core/messages/ai.py
```diff
@@ -25,19 +25,36 @@
...
+
+            {
+                "input_tokens": 10,
+                "output_tokens": 20,
+                "total_tokens": 30
+            }
     """
 
     input_tokens: int
+    """Count of input (or prompt) tokens."""
```
eyurtsev, line 42: @baskaryan let me know if OK -- it seems to render better in APIReference

libs/core/langchain_core/messages/human.py
```diff
@@ -4,11 +4,33 @@
...
+
+            # Instantiate a chat model and invoke it with the messages
+            model = ...
+            print(model.invoke(messages))
+    """
 
     example: bool = False
     """Whether this Message is being passed in to the model as part of an example 
-        conversation.
+    conversation.
```
eyurtsev, line 33: @baskaryan do you know if example is actually used / respected in any models?

libs/core/langchain_core/messages/system.py
```diff
@@ -4,8 +4,29 @@
...
+
+    Example:
+
+        .. code-block:: python
+
+            from langchain_core.messages import HumanMessage, SystemMessage
+
+            messages = [
+                SystemMessage(
+                    content="You are a helpful assistant! Your name is Bob."
```
baskaryan, line 20: nit don't need to specify content as kwarg

baskaryan: (and imo cleaner not to)

libs/core/langchain_core/messages/human.py
baskaryan: was for some of the original google palm models but afaict doesn't apply anymore

baskaryan (approved):

libs/core/langchain_core/messages/system.py
eyurtsev: Most of our examples still do -- I'd prefer doing an upgrade a bit later once more folks are on the newest version of core?

PR #23172, opened 2024-06-19
core[patch]: update test to catch circular imports

ccurme:
This raises ImportError due to a circular import:
```python
from langchain_core import chat_history
```

This does not:
```python
from langchain_core import runnables
from langchain_core import chat_history
```

Here we update `test_imports` to run each import in a separate subprocess. Open to other ways of doing this!

### DIFF
--- a/libs/core/langchain_core/runnables/__init__.py
+++ b/libs/core/langchain_core/runnables/__init__.py
@@ -36,7 +36,6 @@ from langchain_core.runnables.config import (
     run_in_executor,
 )
 from langchain_core.runnables.fallbacks import RunnableWithFallbacks
-from langchain_core.runnables.history import RunnableWithMessageHistory
 from langchain_core.runnables.passthrough import (
     RunnableAssign,
     RunnablePassthrough,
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/runnables/__init__.py
```diff
@@ -79,7 +78,6 @@
     "RunnablePick",
     "RunnableSequence",
     "RunnableWithFallbacks",
-    "RunnableWithMessageHistory",
```
eyurtsev, line 82: We want to retain this here

libs/core/tests/unit_tests/test_imports.py
```diff
@@ -1,14 +1,22 @@
...
+import pytest
+
 
 def test_importable_all() -> None:
     for path in glob.glob("../core/langchain_core/*"):
         relative_path = Path(path).parts[-1]
         if relative_path.endswith(".typed"):
             continue
         module_name = relative_path.split(".")[0]
+        result = subprocess.run(
```
eyurtsev, line 15: Perhaps split into another unit test? (Or maybe add a doc-string to explain what it's doing -- i realize I'm guilty of this as well, but as the test picks up complexity it'll be more difficult to remember why the code is what it is

eyurtsev (approved):

libs/core/langchain_core/runnables/__init__.py
eyurtsev: Unless it wasn't released yet, if so, let's remove and add properly

PR #23180, opened 2024-06-19
community[patch]: Prevent unit tests from making network requests

eyurtsev:
* Prevent unit tests from making network requests


libs/community/tests/unit_tests/embeddings/test_ovhcloud.py
```diff
@@ -23,3 +23,4 @@ def test_ovhcloud_empty_access_token_should_not_raise_error() -> None:
         model_name="multilingual-e5-base", region="kepler", access_token=""
     )
     assert isinstance(llm, OVHCloudEmbeddings)
+
```
eyurtsev, line 26:
```suggestion
```

ccurme (approved):

PR #23182, opened 2024-06-19
core[patch]: fix chat history circular import

baskaryan:


libs/core/langchain_core/chat_history.py
```diff
@@ -102,6 +101,8 @@ async def aget_messages(self) -> List[BaseMessage]:
         In general, fetching messages may involve IO to the underlying
         persistence layer.
         """
+        from langchain_core.runnables.config import run_in_executor
```
eyurtsev, line 104: Should be usable in global namespace I think since it's using runnables.config rather than runnables?

libs/core/langchain_core/runnables/history.py
```diff
@@ -25,14 +24,14 @@
 )
 
 if TYPE_CHECKING:
+    from langchain_core.chat_history import BaseChatMessageHistory
```
eyurtsev, line 27: could we use langchain_core.chat_history.base?

baskaryan: there is no base

eyurtsev (approved):

PR #23184, opened 2024-06-19
doc: replace function all with tool call

mainred:


docs/docs/integrations/tools/bing_search.ipynb
```diff
@@ -213,24 +213,24 @@
...
+      "{'snippet': '<b>Shanghai</b> - <b>Weather</b> warnings issued 14-day forecast. <b>Weather</b> warnings issued. Forecast - <b>Shanghai</b>. Day by day forecast. Last updated today at 18:00. Tonight, A clear sky and a gentle breeze. Clear Sky.', 'title': 'Shanghai - BBC Weather', 'link': 'https://www.bbc.com/weather/1796236'}\n"
      ]
     }
    ],
    "source": [
     "# .invoke wraps utility.results\n",
-    "response = tool.invoke(\"What is the weather in Shanghai\")\n",
-    "for item in list(response):\n",
+    "response = tool.invoke(\"What is the weather in Shanghai?\")\n",
+    "for item in eval(response):\n",
```
ccurme, line 233: why `eval` here? maybe a nitpick but I'd avoid it on principle since this will execute any code returned by the tool.

ccurme (commented):
Thanks for this!

docs/docs/integrations/tools/bing_search.ipynb
mainred: Replaced with json.loads instead.
I think we can consider make the response of bingsearch-related to be more exact by returning list of dic as Tavily search.

ccurme (approved):

PR #22546, opened 2024-06-05
community[patch]: restore compatibility with SQLAlchemy 1.x

krassowski:
- **Description:** Restores compatibility with SQLAlchemy 1.4.x that was broken since #18992 and adds a test run for this version on CI (only for Python 3.11)
- **Issue:** fixes #19681
- **Dependencies:** None
- **Twitter handle:** `@krassowski_m`

PR #23186, opened 2024-06-19
infra: add SQLAlchemy to min version testing

efriis:
preventing issues like #22546 

Notes:
- this will only affect release CI. We may want to consider adding running unit tests with min versions to PR CI in some form
- because this only affects release CI, it could create annoying issues releasing while I'm on vacation. Unless anyone feels strongly, I'll wait to merge this til when I'm back


krassowski:
Since this was switched to draft - what needs to be done before merging?

PR #23187, opened 2024-06-19
rfc: lint docstrings

baskaryan:


spazm:
As a random developer reading open issues and PRs, I will find it useful to know that these linter errors are part of the project of adding linters and not a random regression.  

Since you'll be making a bunch of "add linter" and "fix the linter errors" PRs, A link back to the parent #23188 would be helpful (Yes, I see that this is 23187 and you added that directly after this.  :smile: )

> Add docstring linting to core, langchain, partner packages #23188

PR #23191, opened 2024-06-19
infra: add formatting rules

baskaryan:


efriis:
pretty certain this will have to be redone if we want to land

PR #23192, opened 2024-06-19
infra: run CI on large diffs

baskaryan:
currently we skip CI on diffs >= 300 files. think we should just run it on all packages instead

### DIFF
--- a/.github/scripts/check_diff.py
+++ b/.github/scripts/check_diff.py
@@ -15,6 +15,10 @@ LANGCHAIN_DIRS = [
     "libs/experimental",
 ]
 
+def all_package_dirs() -> Set[str]:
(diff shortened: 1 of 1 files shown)


.github/scripts/check_diff.py
```diff
@@ -55,8 +59,9 @@ def add_dependents(dirs_to_eval: Set[str], dependents: dict) -> List[str]:
 
     if len(files) == 300:
```
efriis, line 60:
```suggestion
    if len(files) >= 300:
```

efriis (approved):
small suggestion to future proof if that action ever returns more than 300

.github/scripts/check_diff.py
baskaryan: fwiw this still goes through the regular addition of things to test as well after this check. so if they ever bump limit to > 300 it would actually work correctly and at exactly 300 would be overly eager to CI things

PR #23197, opened 2024-06-19
huggingface[patch]: fix CI for python 3.12

ccurme:


libs/partners/huggingface/pyproject.toml
```diff
@@ -17,6 +17,15 @@ tokenizers = ">=0.19.1"
...
 huggingface-hub = ">=0.23.0"
+# Support Python 3.8 and 3.12+.
+scipy = [
+    {version = "^1", python = "<3.12"},
+    {version = "^1.7.0", python = ">=3.12"}
+]
+numpy = [
+    {version = "^1", python = "<3.12"},
+    {version = "^1.26.0", python = ">=3.12"}
+]
```
baskaryan, line 28: do these need to specified as required deps or could they be optional / in an optional group? i think we just need poetry.lock to update

ccurme: Good call thanks. Moved to test deps.

PR #23198, opened 2024-06-19
RFC: Remove dependency on Pydantic from langchain-core

nfcampos:


hwchase17:
dont think were doing

PR #23204, opened 2024-06-20
Docs: Update Rag tutorial so it includes an additional notebook cell with pip installs of required langchain_chroma and langchain_community.

Random-Black-Coder:
Description: Update Rag tutorial notebook so it includes an additional notebook cell with pip installs of required langchain_chroma and langchain_community.

This fixes the issue with the rag tutorial gives you a 'missing modules' error if you run code in the notebook as is.

### DIFF
--- a/docs/docs/tutorials/rag.ipynb
+++ b/docs/docs/tutorials/rag.ipynb
@@ -124,7 +124,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


ccurme:
Thanks for this! Consolidated with the dependency block farther up the tutorial.

ccurme (approved):

PR #23211, opened 2024-06-20
Community: ConfluenceLoader extensions (fetch all content from all spaces + fetch all pages in space)

GaetanoPrudente:
**Description**
This PR is to extend the functionality of the ConfluenceLoader, specifically:
1. the fetch is not limited only to a specifc space, page id, etc.., it is now possible to also fetch all content from all spaces
2. the max_pages limit is not mandatory, when not specified all pages in the context are retrieved

**Issue**
1. The loader was not as flexible as the Confluence API, in case it was necessary to fetch all the documents from a Confluence workspace it was necessary to use the Confluence API directly to get all the spaces, loop on those with the Loader, etc.. 
Not being the optimal solution (1 extra API only for the spaces, loop execution, multiple API calls with the loader) this extension aims to solve the issue giving the possibility to query all content
2. When requesting pages from a space without knowing how many are there, the only chance to get it right was to set the value to the highest possible number. This extension covers that situation and keeps on looping until no more pages are left

**Dependencies**
No dependencies introduced


baskaryan (approved):

libs/community/langchain_community/document_loaders/confluence.py
```diff
@@ -461,9 +476,11 @@ def paginate_request(self, retrieval_method: Callable, **kwargs: Any) -> List:
         :rtype: List
         """
 
-        max_pages = kwargs.pop("max_pages")
         docs: List[dict] = []
-        while len(docs) < max_pages:
+        max_pages = kwargs.pop("max_pages", -1)
+        pages_limit_condition = (len(docs) < max_pages) if max_pages != -1 else True
```
efriis, line 481: bug

efriis:
this has a bug and some conflicts - if this is still an issue feel free to reopen against latest master!

PR #23214, opened 2024-06-20
core[patch]: set schema format for AsyncRootListenersTracer

mackong:
- **Description:** AsyncRootListenersTracer support on_chat_model_start, it's schema_format should be "original+chat".
  - **Issue:** N/A
  - **Dependencies:**

### DIFF
--- a/libs/core/langchain_core/tracers/root_listeners.py
+++ b/libs/core/langchain_core/tracers/root_listeners.py
@@ -71,7 +71,7 @@ class AsyncRootListenersTracer(AsyncBaseTracer):
         on_end: Optional[AsyncListener],
         on_error: Optional[AsyncListener],
     ) -> None:
-        super().__init__()
+        super().__init__(_schema_format="original+chat")
 
         self.config = config
         self._arg_on_start = on_start


eyurtsev (approved):
@nicolasnk any concerns with this change? I think we want to merge this in, but wanted to check in with you first.

eyurtsev:
@mackong thanks, going to give @nicolasnk a chance to opine prior to merging in

nicolasnk:
> @mackong thanks, going to give @nicolasnk a chance to opine prior to merging in

Hey @eyurtsev. Thanks for letting me know.

I see that the same [code for RootListener]9https://github.com/langchain-ai/langchain/blob/976b45661900e44f28179327dc0c178d6d839191/libs/core/langchain_core/tracers/root_listeners.py#L29). Does it make sense to do it in [the parent class _TracerCore](https://github.com/langchain-ai/langchain/blob/976b45661900e44f28179327dc0c178d6d839191/libs/core/langchain_core/tracers/core.py#L56) ?

eyurtsev:
@nicolasnk Yes, but we'll do it as a part of a separate PR. We need to spec out if anything might break for users that are sub-classing from BaseTracer right now

PR #23224, opened 2024-06-20
**community**: Fix sql_databse.from_databricks issue when ran from Job

lmoros-DB:
**Desscription**: When the ``sql_database.from_databricks`` is executed from a Workflow Job,  the ``context`` object does not have a "browserHostName" property, resulting in an error. This change manages the error so the  "DATABRICKS_HOST" env variable value is used instead of stoping the flow

### DIFF
--- a/libs/community/langchain_community/utilities/sql_database.py
+++ b/libs/community/langchain_community/utilities/sql_database.py
@@ -201,10 +201,12 @@ class SQLDatabase:
             from dbruntime.databricks_repl_context import get_context
 
             context = get_context()
+            default_host = context.browserHostName
+            
         except ImportError:
-            pass
+            default_host = None
 
-        default_host = context.browserHostName if context else None
+        
         if host is None:
             host = get_from_env("host", "DATABRICKS_HOST", default_host)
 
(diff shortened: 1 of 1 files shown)


eyurtsev (approved):

libs/community/langchain_community/utilities/sql_database.py
```diff
@@ -201,10 +201,12 @@ def from_databricks(
...
 
             context = get_context()
+            default_host = context.browserHostName
+            
         except ImportError:
-            pass
+            default_host = None
 
-        default_host = context.browserHostName if context else None
+
```
eyurtsev, line 209:
```suggestion
```

libs/community/langchain_community/utilities/sql_database.py
```diff
@@ -201,10 +201,12 @@ def from_databricks(
             from dbruntime.databricks_repl_context import get_context
 
             context = get_context()
+            default_host = context.browserHostName
+
```
eyurtsev, line 205:
```suggestion
```

eyurtsev (approved):

PR #23225, opened 2024-06-20
docs: API reference remove Prev/Up/Next buttons

eyurtsev:
These do not work anyway. Let's remove them for now for simplicity.


docs/api_reference/themes/scikit-learn-modern/layout.html
```diff
@@ -2,132 +2,129 @@
 {%- set url_root = pathto('', 1) %}
```
eyurtsev, line 2: Reforamtted the code + removed buttons for navigation which do not work

PR #23234, opened 2024-06-20
standard tests: add test for tool calling

ccurme:
Including streaming

### DIFF
--- a/libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
+++ b/libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
@@ -1,13 +1,21 @@
 import base64
 import json
+from typing import Optional
 
 import httpx
 import pytest
 from langchain_core.language_models import BaseChatModel
-from langchain_core.messages import AIMessage, AIMessageChunk, HumanMessage, ToolMessage
+from langchain_core.messages import (
+    AIMessage,
+    AIMessageChunk,
+    BaseMessageChunk,
+    HumanMessage,
+    ToolMessage,
+)
 
 from langchain_standard_tests.unit_tests.chat_models import (
     ChatModelTests,
+    magic_function,
     my_adder_tool,
 )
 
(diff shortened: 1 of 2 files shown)


libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
```diff
@@ -98,6 +106,31 @@ def test_stop_sequence(self, model: BaseChatModel) -> None:
         result = custom_model.invoke("hi")
         assert isinstance(result, AIMessage)
 
+    def test_tool_calling(self, model: BaseChatModel) -> None:
+        if not self.has_tool_calling:
+            pytest.skip("Test requires tool calling.")
+        model_with_tools = model.bind_tools([magic_function])
+
+        def _validate_tool_call_message(message: AIMessage) -> None:
```
baskaryan, line 114: nit does this need to be define inside

libs/standard-tests/langchain_standard_tests/unit_tests/chat_models.py
```diff
@@ -21,6 +21,12 @@ def my_adder_tool(a: int, b: int) -> int:
     return a + b
 
 
+@tool
+def magic_function(input: int) -> int:
```
baskaryan, line 25: whats this for?

baskaryan: oh i see, but why does it have to be defined in unit test?

libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
ccurme: moved outside

libs/standard-tests/langchain_standard_tests/unit_tests/chat_models.py
ccurme: I just put it there because that's where the other tool was. I moved it to the integration test module.

ccurme:
@baskaryan all standard tests are [passing](https://github.com/langchain-ai/langchain/actions/runs/9604427116/job/26489897545) (Cohere is failing for another reason), will merge but lmk if you'd like to see any changes.

PR #23258, opened 2024-06-21
core: added graph visualization method using yfiles_jupyter_graphs

Ge-sa:
- **Description** : added a method similar to draw_mermaid_png, visualizing the underlying graphs using the yfiles_jupyter_graphs widget. This needs a jupyter environment. 
- **Twitter handle:**  @yworks


eyurtsev (changes requested):
Hi @Ge-sa thank you for the contribution. I'll check in with @nfcampos, but I don't think that we'd want to add this into core even as an optional dependency. I don't think that the interactive widgets significantly improve existing capabilities to justify another dependency for langchain-core (even an optional one)

yGuy:
I think the few lines of code (fewer than the mermaid integration, if I'm correct) should be worth it:

It's not just about interactivity (being able to search your graph and better see dependencies in the local view, e.g.), but its also about being able to zoom in and out and seeing all the details even in larger graphs. With the PNG export you get large image files of mediocre quality for large graphs with many items.

And there is no additional dependency. If you want to use it, you need to do the additional pip install, if not, there will be no dependency. For people working within jupyter notebooks or even more so google colab and other colaborative notebooks, being able to view and search the graph interactively is a big plus, I think.

That said - full disclosure - I suggested to Gesa to do this PR, so I am biased. And should your doubts be more about the logo in the widget - we could also start with the side pane collapsed or showing the data view, first.

Thanks for considering this.

lmeyerov:
if there is interest in this, we can do a much more scalable & interactive version with pygraphistry, and establishing a good pattern for follow-ons like these vs hard-coding 1 engine would be of interest, at least as we've seen in other oss libs for this step

yGuy:
@lmeyerov - great idea! Why don't you create a PR for this and be constructive, too? No need to bash and stop someone else's PR and making statements you don't prove ("*much more* scalable & interactive"). 
As a proof-of-concept, create a PR for graphviz! GraphViz is good, too, but It's just the same as with mermaid and just as non-interactive. Then create a version that works with [your paid, proprietary software](https://github.com/graphistry/pygraphistry/blob/53448d4ef153fd262466087a951bc28a44c8fadf/README.md?plain=1#L14), too. And maybe next time, be sure to disclose that this is what it's really all about: Getting people to use your "free" service and sign-up with your plans. Thanks!

That said, I like the idea of a pluggable and swappable option, as long as it's easy to use and benefits the user, which should be our top priority and also was the reason this PR was created. I believe @Ge-sa will be more than happy to make use of your proposed new API and infrastructure, if it gets approved. You can point your paying users to it and the rest can use the free tools that they want to use, too. Ge-Sa called the function `show_jupyter_graph`, after all, so you can plug your pluggable implementation in there and it would be totally backwards-compatible!

Let's be constructive and improve the situation for everyone, instead...

eyurtsev:
```python
from langchain_core.globals import set_renderer
from langchain_community.renderers import yfiles_graphs

set_renderer(yfiles_graphs)
```

lmeyerov:
@eyurtsev Slick! Yes I was thinking similar, thank you


RE:config,

* Maybe also an env var  like `LANGCHAIN_DEFAULT_RENDERER=some.package.name`, given the use of env vars for configuring the project already? Then it's easier to do a default programmatic import in managed envs. And/or some prebaked special forms like `"yfiles"` that are natively recognized.

* If the config path here is the status quo and nothing to hook into near-term, `set_renderer(BaseRenderer)` makes sense given the `main`'s `globals.py`. Note below on how I see config** happen in other OSS libs, so wanted to double check.

RE:proprietary bits,

* Agreed, I would also be against putting proprietary and non-core sw in core, that's become an issue in pygraphistry as well, and our own use of langchain
* AFAICT not relevant to a custom renderer capability, but as apparently it's relevant to Sebastian, PyGraphistry happens to be OSS and the hub server tier has a free GPU tier with many users, but again, I don't think that's relevant to custom renderers. Afaict, the proposed solution by @eyurtsev gives a path for custom `BaseRenderer`s to extend, inject, change defaults & verify custom rendering behavior

- Leo

---

**config: In other OSS I've seen more generic global settings, e.g., dask is pretty clean for swapping in gpu engines via `dask.config.set({k: v})`: https://huggingface.co/docs/datasets-server/cudf , and their underlying settings mechanism does a lot of 12-factor config stuff nice for production settings like env vars and config files: https://docs.dask.org/en/latest/configuration.html . We'd certainly use that in our envs so things Just Work for our users. But my vote remains current `globals.py` hard-coding if new config modes aren't already landing ..

yGuy:
Hi @eyurtsev, thanks for implementing the alternative PR. I think that would work for this PR (which itself would be obsolete, but it would work for the *intent* behind the PR) and others, but in my humble opinion, I think this seems a bit too much for a tool like this. Having a pluggable renderer architecture for a tool where rendering is only used as a tool to improve DX seems a bit overkill to me and your changeset and API is actually quite a bit larger than what was proposed, here. But of course that's not my decision, and I fully respect the decision of the maintainers. Anything that improves the DX is great!

I thought about this, again. Maybe it makes more sense to make the underlying data that would be required for a visualization easily accessible to third party code. Then anyone could do something along the lines of the following:

```
from Something import GraphWidget

display(GraphWidget(graph = chain.get_graph())  # or even "graph = chain"
```

Which to me seems at least as convenient DX, but of course is way less discoverable than "render_in_jupyter":
```
from langchain_core.globals import set_renderer
from langchain_community.renderers import yfiles_graphs

set_renderer(yfiles_graphs)

....

chain.get_graph().render_in_jupyter()
```

This could be improved by showing some example in the docs, of course so that people would become aware of other options.

Right now, AFAICT, the `to_json` is the best way to get a hold of all the relevant data, but this is a bit under-specified and not well documented. I haven't looked into the sources, so I am not sure, whether all the required information is in there.



Then again, if it's really about the OSS: I can see us creating an OSS wrapper around our widget that only internally uses our free (but non-OSS) widget, if that helps. That would bring it on par with the proposed _more scalable & interactive_, free to signup and pay solution with regards to the license terms ;-) .

yGuy:
Thanks, @eyurtsev - we'll look into these options and will also take a closer look at the two different use-cases. @Ge-sa - let's discuss the initial options offline!

PR #23260, opened 2024-06-21
community: Fix OVHcloud 401 Unauthorized on embedding.

Joffref:
They are now rejecting with code 401 calls from users with expired or invalid tokens (while before they were being considered anonymous). Thus, the authorization header has to be removed when there is no token.

Related to: #23178


libs/community/langchain_community/embeddings/ovhcloud.py
```diff
@@ -51,9 +51,11 @@ def _generate_embedding(self, text: str) -> List[float]:
         """
         headers = {
             "content-type": "text/plain",
-            "Authorization": f"Bearer {self.access_token}",
         }
 
+        if self.access_token is not None:
+            headers["Authorization"] = f"Bearer {self.access_token}"
```
ccurme, line 57: what happens to requests if they are missing this header?

eyurtsev:
Hi @Joffref, the approach is typically to not have unit tests issue any network request. We usually use `responses` package to mock the response and allow unit tests to run against the code

libs/community/langchain_community/embeddings/ovhcloud.py
Joffref: They are just rate-limited to 12 calls/min.

Joffref:
We should rather avoid allowing unauthenticated usage. Therefore, I changed a bit my PR.

libs/community/langchain_community/embeddings/ovhcloud.py
Joffref: As it is enforced now, I resolve the conversation.

libs/community/tests/unit_tests/embeddings/test_ovhcloud.py
```diff
@@ -4,22 +4,28 @@
...
+            model_name="multilingual-e5-base", region="", access_token="token"
+        )
 
 
-def test_ovhcloud_empty_access_token_should_not_raise_error() -> None:
-    llm = OVHCloudEmbeddings(
-        model_name="multilingual-e5-base", region="kepler", access_token=""
-    )
-    assert isinstance(llm, OVHCloudEmbeddings)
+def test_ovhcloud_empty_access_token_should_raise_error() -> None:
```
ccurme, line 27: (nit) could parameterize these

ccurme (approved):

PR #23273, opened 2024-06-21
rfc: tool call / message filter

baskaryan:


libs/core/langchain_core/messages/utils.py
```diff
@@ -373,6 +392,71 @@ def filter_messages(
...
+                isinstance(msg, AIMessage)
+                and msg.tool_calls
+                or (
+                    isinstance(msg.content, list)
+                    and any(
+                        block.get("type") == "tool_use"
+                        for block in msg.content
+                        if isinstance(block, dict)
+                    )
+                )
```
ccurme, line 406: can we assume that if there are tool use blocks, then .tool_calls is populated?

libs/core/langchain_core/messages/utils.py
```diff
@@ -373,6 +392,71 @@ def filter_messages(
...
+                msg.tool_calls = filtered_tool_calls
+                if isinstance(msg.content, list):
+                    filtered_content = []
+                    for block in msg.content:
+                        if (
+                            isinstance(block, dict)
+                            and block.get("type") == "tool_use"
+                            and exclude_tool_name_condition(block.get("name"))
+                        ):
+                            continue
```
ccurme, line 424: is this for practical reasons (e.g., having inconsistent tool_use blocks and .tool_calls is problematic) or for aesthetics?

ccurme (commented):
LGTM

PR #23285, opened 2024-06-21
core[minor]: update draw_mermaid node label processing

vbarda:
This fixes processing issue for nodes with numbers in their labels (e.g. `"node_1"`, which would previously be relabeled as `"node__"`, and now are correctly processed as `"node_1"`)

### DIFF
--- a/libs/core/langchain_core/runnables/graph_mermaid.py
+++ b/libs/core/langchain_core/runnables/graph_mermaid.py
@@ -114,7 +114,7 @@ def draw_mermaid(
 
 def _escape_node_label(node_label: str) -> str:
     """Escapes the node label for Mermaid syntax."""
-    return re.sub(r"[^a-zA-Z-_]", "_", node_label)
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/runnables/graph_mermaid.py
```diff
@@ -114,7 +114,7 @@ def draw_mermaid(
 
 def _escape_node_label(node_label: str) -> str:
     """Escapes the node label for Mermaid syntax."""
-    return re.sub(r"[^a-zA-Z-_]", "_", node_label)
+    return re.sub(r"[^a-zA-Z-_0-9]", "_", node_label)
```
eyurtsev, line 117: Do you have a unit test?

eyurtsev (approved):

libs/core/langchain_core/runnables/graph_mermaid.py
vbarda: now i do!

eyurtsev (approved):

PR #23291, opened 2024-06-21
anthropic[patch]: expose payload

baskaryan:
Screenshot 2024-06-21 at 4 56 02 PM


efriis (commented):
should we make it `_get_request_payload`? In the event of sdk changes out of our control, this could be difficult to support long-term

PR #23300, opened 2024-06-22
community: allows to disable `HuggingFaceEndpoint` API-Token validation

AIexanderDicke:
**Description:** Allows to skip the validation of the API-Token in `HuggingFaceEndpoint` by setting `validate_api_token=False`.


ccurme:
What is the motivation for this?

efriis:
closing without a response to ^

if you reopen would recommend either opening against the integration package in libs/partners/huggingface instead of against the deprecated community integration here!

PR #23306, opened 2024-06-22
community: add default value to bing_search_url

mainred:
bing_search_url is an endpoint to requests bing search resource and is normally invariant to users, we can give it the default value to simply the uesages of this utility/tool

### DIFF
--- a/libs/community/langchain_community/utilities/bing_search.py
+++ b/libs/community/langchain_community/utilities/bing_search.py
@@ -5,6 +5,8 @@ import requests
 from langchain_core.pydantic_v1 import BaseModel, Extra, Field, root_validator
 from langchain_core.utils import get_from_dict_or_env
 
+# BING_SEARCH_ENDPOINT is the default endpoint for Bing Search API and is invariant to users.
+BING_SEARCH_ENDPOINT = "https://api.bing.microsoft.com/v7.0/search"
 
 class BingSearchAPIWrapper(BaseModel):
     """Wrapper for Bing Search API."""
@@ -52,7 +54,7 @@ class BingSearchAPIWrapper(BaseModel):
             values,
             "bing_search_url",
             "BING_SEARCH_URL",
-            # default="https://api.bing.microsoft.com/v7.0/search",
+            default=BING_SEARCH_ENDPOINT,
         )
 
         values["bing_search_url"] = bing_search_url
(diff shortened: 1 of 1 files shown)


ccurme:
Thanks @mainred, can you clarify - if that's the default for Bing, why override it instead of letting it be determined by Bing? Is it to enable users to set it with environment variables?

mainred:
> Thanks @mainred, can you clarify - if that's the default for Bing, why override it instead of letting it be determined by Bing? Is it to enable users to set it with environment variables?

@ccurme , Currently there are two bing search services provided on Azure, [bing web custom search](https://learn.microsoft.com/en-us/bing/search-apis/bing-custom-search/overview) and [bing web search](https://learn.microsoft.com/en-us/bing/search-apis/bing-web-search/overview) if we focus on the web search part. And they all have constant endpoint, https://api.bing.microsoft.com/v7.0/custom/search for [bing custom web search](https://learn.microsoft.com/en-us/bing/search-apis/bing-custom-search/reference/endpoints#endpoints) and https://api.bing.microsoft.com/v7.0/search for [bing web search](https://learn.microsoft.com/en-us/bing/search-apis/bing-web-search/reference/endpoints#endpoints)

Before, bing search and custom search are integrated into azure cognitive service, and [the cognitive service endpoint varying from one cognitive service to another](https://github.com/Azure-Samples/cognitive-services-REST-api-samples/blob/master/python/Search/BingCustomSearchv7.py#L19), so our bing wrapper asks for an endpoint from env, but as bing search and bing web search are now standalone services, and we can default constant endpoint instead.
The user can still pick bing custom search through the endpoint env.

> Why we choose bing search by default instead of bing custom search?

Bing custom search requires a customConfig, which is not always the common case when people just make simple search results.

mainred:
Not sure if it's valuable to put these into the code comments.

ccurme (approved):

ccurme:
Thanks for the explanation. Will merge for now but could be helpful to add a comment as you suggest!

mainred:
Created https://github.com/langchain-ai/langchain/pull/23523 to make bing web search as the option and added comment to explain why. 
@ccurme It may be a bit different from what I said earlier, which is to configurable but default value to bing web search considering bing custom search is not a common use case for bing search and requires additional custom search instance, which is so far not well supported. So I think people were just using bing search service.

Issue #22819, opened 2024-06-12
Prepare for pydantic 2 migration by refactoring vanilla @root_validator() usage


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

# Context

Currently, LangChain supports Pydantic 2 only through the v1 namespace. 

The plan is to transition for Pydantic 2 with release 0.3.0 of LangChain, and drop support for Pydantic 1. 

LangChain has around ~1000 pydantic objects across different packages While LangChain uses a number of deprecated features, one of the harder things to update is the usage of a vanilla `@root_validator()` (which is used ~250 times across the code base).

The goal of this issue is to do as much preliminary work as possible to help prepare for the migration from pydantic v1 to pydantic 2.

To help prepare for the migration, we'll need to refactor each occurrence of a vanilla `root_validator()`  to one of the following 3 variants (depending on what makes sense in the context of the model):
1. `root_validator(pre=True)` -- pre initialization validator
2. `root_validator(pre=False, skip_on_failure=True)` -- post initialization validator
3. `root_validator(pre=True)` AND `root_validator(pre=False, skip_on_failure=True)` to include both pre initialization and post initialization validation.

## Guidelines

- Pre-initialization is most useful for **creating defaults** for values, especially when the defaults cannot be supplied per field individually.
- Post-initialization is most useful for doing more complex validation, especially one that involves multiple fields.

## What not to do

* Do **NOT** upgrade to `model_validator`. We're trying to break the work into small chunks that can be done while we're still using Pydantic v1 functionality!
* Do **NOT** create `field_validators` when doing the refactor.

## Simple Example

```python
class Foo(BaseModel):
    @root_validator()
    def validate_environment(cls, values: Dict) -> Dict:
        values["api_key"] = get_from_dict_or_env(
            values, "some_api_key", "SOME_API_KEY", default=""
        )

        if values["temperature"] is not None and not 0 <= values["temperature"] <= 1:
            raise ValueError("temperature must be in the range [0.0, 1.0]")
        return values
```

# After refactor

```python
class Foo(BaseModel):
    @root_validator(pre=True)
    def pre_init(cls, values):
        # Logic for setting defaults goes in the pre_init validator.
        # While in some cases, the logic could be pulled into the `Field` definition
        # directly, it's perfectly fine for this refactor to keep the changes minimal
        # and just move the logic into the pre_init validator.
[...]

Issue #22448, opened 2024-06-03
Support native HuggingFace env vars


baskaryan:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

huggingface_hub has its own environment variables that it reads from: https://huggingface.co/docs/huggingface_hub/main/en/package_reference/environment_variables. Langchain x HuggingFace integrations should be able to read from these, too.

liugddx:
i want try

liugddx:
<img width="1201" alt="image" src="https://github.com/langchain-ai/langchain/assets/48236177/0bd0d864-aae5-4903-9b98-d3f32b2966aa">

I found that the parameters for environment variables have been specified here. Do we still need to specify them separately?

wenngong:
@baskaryan Please review #23309. Let me know if I miss any part, Thx.

zk1tty:
@wenngong I think you might also need to mention @liugddx !

wenngong:
> @wenngong I think you might also need to mention @liugddx !

Loop @liugddx. 

Hi liugddx, please review this PR: #23309 related to this issue.

PR #23309, opened 2024-06-23
partners: support reading HuggingFace params from env

Zizo-Vi:
Description: 
  1. partners/HuggingFace module support reading params from env. Not adjust langchain_community/.../huggingfaceXX modules since they are deprecated.
  2. pydantic 2 @root_validator migration.

Issue: #22448 #22819

### DIFF
--- a/libs/partners/huggingface/langchain_huggingface/embeddings/huggingface_endpoint.py
+++ b/libs/partners/huggingface/langchain_huggingface/embeddings/huggingface_endpoint.py
@@ -1,5 +1,4 @@
 import json
-import os
 from typing import Any, Dict, List, Optional
 
 from langchain_core.embeddings import Embeddings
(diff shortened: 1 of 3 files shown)


libs/partners/huggingface/langchain_huggingface/embeddings/huggingface_endpoint.py
```diff
@@ -46,11 +46,15 @@ class Config:
 
         extra = Extra.forbid
 
-    @root_validator()
+    @root_validator(pre=False, skip_on_failure=True)
```
eyurtsev, line 49: This should be pre=True based on the logic below -- all of it can be done as pre-init logic

Just requires re-writing values['model'] to values.get('model')

eyurtsev (approved):

libs/partners/huggingface/langchain_huggingface/embeddings/huggingface_endpoint.py
eyurtsev: Merging as is, but if you can please re-write this with `pre=True`

Issue #23308, opened 2024-06-23
SQL problem in langchain-community `langchain_community.vectorstores.duckdb`:194


NewJerseyStyle:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code:
```python
from langchain_community.embeddings import HuggingFaceEmbeddings
from langchain_community.vectorstores import DuckDB

db = DuckDB.from_texts(
    ['a', 'v', 'asdfadf', '893yhrfa'],
    HuggingFaceEmbeddings())
```

### Error Message and Stack Trace (if applicable)

```
CatalogException                          Traceback (most recent call last)
Cell In[3], line 2
      1 texts = ['a', 'v', 'asdfadf', '893yhrfa']
----> 2 db = DuckDB.from_texts(texts,
      3                        HuggingFaceEmbeddings())
      4 db.similarity_search('ap', k=5)

File /opt/conda/lib/python3.10/site-packages/langchain_community/vectorstores/duckdb.py:287, in DuckDB.from_texts(cls, texts, embedding, metadatas, **kwargs)
    278 instance = DuckDB(
    279     connection=connection,
    280     embedding=embedding,
   (...)
    284     table_name=table_name,
    285 )
    286 # Add texts and their embeddings to the DuckDB vector store
--> 287 instance.add_texts(texts, metadatas=metadatas, **kwargs)
    289 return instance

File /opt/conda/lib/python3.10/site-packages/langchain_community/vectorstores/duckdb.py:194, in DuckDB.add_texts(self, texts, metadatas, **kwargs)
    191 if have_pandas:
    192     # noinspection PyUnusedLocal
    193     df = pd.DataFrame.from_dict(data)  # noqa: F841
--> 194     self._connection.execute(
    195         f"INSERT INTO {self._table_name} SELECT * FROM df",
    196     )
    197 return ids

CatalogException: Catalog Error: Table with name df does not exist!
Did you mean "pg_am"?
LINE 1: INSERT INTO embeddings SELECT * FROM df
```

### Description

* I am trying to use DuckDB as vector storage
* I expect to get a vector storage instance connected to DuckDB
* Instead it throw error on initialization

### System Info

System Information
[...]

PR #23310, opened 2024-06-23
community: Patch/issue#23308

NewJerseyStyle:
- **Description:** Changed `langchain/libs/community/langchain_community/vectorstores/duckdb.py` line 194 from `self._connection.execute(...)` to `df.to_sql(...)`
    - Actually I tested both version copying the section of code and test them in notebook environment and both of them works fine
    - But integrating to the package, only my version works in the kaggle notebook where I encountered issue#23308
    - > ⚠️ A warning popped up
duckdb.py:194: UserWarning: pandas only supports SQLAlchemy connectable (engine/connection) or database string URI or sqlite3 DBAPI2 connection. Other DBAPI2 objects are not tested. Please consider using SQLAlchemy.
  df.to_sql(name=self._table_name,
    - ℹ️ The warning did not affect the result of writing to database and `similarity_search()` returns expected response
- **Issue:** #23308 
- **Dependencies:** None

### DIFF
--- a/libs/community/langchain_community/vectorstores/duckdb.py
+++ b/libs/community/langchain_community/vectorstores/duckdb.py
@@ -193,7 +193,8 @@ class DuckDB(VectorStore):
             df = pd.DataFrame.from_dict(data)
             df.to_sql(name=self._table_name,
                       con=self._connection,
-                      if_exists='append')
+                      if_exists='append',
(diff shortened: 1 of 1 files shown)


NewJerseyStyle:
An elegant solution is proposed in comment to issue#23310

Issue #23338, opened 2024-06-23
FlashrankRerank validate_environment creates a Ranker client even if a custom client is passed


rhlarora84:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code is used to create a Ranker object which downloads the model into a custom directory and then FlashrankRerank is initialized.
```python
def create_ms_marco_mini_llm():
    logger.info("Download model ms_marco_mini_llm")
    model_name = "ms-marco-MiniLM-L-12-v2"
    ranker = Ranker(model_name=model_name, max_length=1024, cache_dir=model_dir)
    return FlashrankRerank(client=ranker, model=model_name)
 ```
 
 The validate_environment method in FlashrankRerank goes to create another Ranker object causing it to download model into a different directory
 
 ```python
     @root_validator(pre=True)
    def validate_environment(cls, values: Dict) -> Dict:
        """Validate that api key and python package exists in environment."""
        try:
            from flashrank import Ranker
        except ImportError:
            raise ImportError(
                "Could not import flashrank python package. "
                "Please install it with `pip install flashrank`."
            )

        values["model"] = values.get("model", DEFAULT_MODEL_NAME)
        values["client"] = Ranker(model_name=values["model"])
        return values
 ```

### Description

FlashrankRerank should check if the client is already initialized before creating a ranker.

### System Info

langchain==0.2.5
langchain-anthropic==0.1.15
langchain-aws==0.1.6
langchain-community==0.2.5
langchain-core==0.2.7
langchain-experimental==0.0.61
langchain-openai==0.1.8
langchain-text-splitters==0.2.1
openinference-instrumentation-langchain==0.1.19

wenngong:
pr #23350

PR #23350, opened 2024-06-24
community: FlashrankRerank support loading customer client

Zizo-Vi:
Description: FlashrankRerank Document compressor support loading customer client
Issue: #23338

### DIFF
--- a/libs/community/langchain_community/document_compressors/flashrank_rerank.py
+++ b/libs/community/langchain_community/document_compressors/flashrank_rerank.py
@@ -38,17 +38,20 @@ class FlashrankRerank(BaseDocumentCompressor):
     @root_validator(pre=True)
     def validate_environment(cls, values: Dict) -> Dict:
         """Validate that api key and python package exists in environment."""
-        try:
-            from flashrank import Ranker
-        except ImportError:
-            raise ImportError(
-                "Could not import flashrank python package. "
-                "Please install it with `pip install flashrank`."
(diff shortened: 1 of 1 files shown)


rhlarora84 (approved):

Zizo-Vi (commented):
pass

wenngong:
@ccurme Please help review and merge, thx!

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -38,17 +38,20 @@ class Config:
...
+            except ImportError:
+                raise ImportError(
+                    "Could not import flashrank python package. "
+                    "Please install it with `pip install flashrank`."
+                )
 
-        values["model"] = values.get("model", DEFAULT_MODEL_NAME)
-        values["client"] = Ranker(model_name=values["model"])
-        return values
+            values["model"] = values.get("model", DEFAULT_MODEL_NAME)
```
ccurme, line 52: do we also want to run this line if `"client" in values`?

wenngong: No, If client set in values, values.model should not load from default. Because:
1. values.model is just used for creating client
2. And If client set by user through values.client, we cannot set values.model to default because we do not know witch model user used for creating the custom client.

ccurme (approved):

PR #23353, opened 2024-06-24
Small bug fixes

joshc-ai21:
Small bug fixes according to your comments


libs/partners/ai21/langchain_ai21/chat_models.py
```diff
@@ -19,15 +19,17 @@
...
+    to understand which parameters are available.
 
     Example:
         .. code-block:: python
 
             from langchain_ai21 import ChatAI21
 
 
-            model = ChatAI21()
+            model = ChatAI21(api_key=<AI21_API_KEY>)
```
ccurme, line 32: this would also be read from environment variables, no?

joshc-ai21: Indeed yes. But the example doesn't show loading or setting that variable explicitly, like our other code samples do and I wanted to call it out explicitly. We have the odd behavior of allowing api_key to either be set explicitly or via OS environment key "AI21_API_KEY". I've seen a couple ways of doing this:
1. Setting the key explicitly, as here,
2. `export AI21_API_KEY=your-api-key`
3. ```import os...
    import os
    from getpass import getpass
    os.environ["AI21_API_KEY"] = getpass()```

I just wanted to make setting the key explicit, so that users don't fail and get frustrated. WDYT?

joshc-ai21: Updated a bit more, and added more context for this specific question. I hope it's okay.

ccurme: Looks good!

PR #23363, opened 2024-06-24
[Core] Logging: Suppress missing parent warning

hinthornw:


libs/core/langchain_core/tracers/core.py
```diff
@@ -50,6 +50,8 @@ class _TracerCore(ABC):
     This class provides common methods, and reusable methods for tracers.
     """
 
+    log_missing_parent: bool = True
```
eyurtsev, line 53: Why class level attribute?

hinthornw: It's more a function of the type of tracer ya?

Could also add to `__init__` and just set in the `with_(async)_listeners()` call

eyurtsev: i usually rely only on __init__

eyurtsev: Anyway up to you works either way!

ccurme (approved):

PR #23368, opened 2024-06-24
LLM Graph transformer dealing with empty strings

tomasonjo:
Pydantic allows empty strings:

```
from langchain.pydantic_v1 import Field, BaseModel

class Property(BaseModel):
  """A single property consisting of key and value"""
  key: str = Field(..., description="key")
  value: str = Field(..., description="value")

x = Property(key="", value="")
```

Which can produce errors downstream. We simply ignore those records

### DIFF
--- a/libs/experimental/langchain_experimental/graph_transformers/llm.py
+++ b/libs/experimental/langchain_experimental/graph_transformers/llm.py
@@ -525,7 +525,7 @@ def _format_nodes(nodes: List[Node]) -> List[Node]:
     return [
         Node(
             id=el.id.title() if isinstance(el.id, str) else el.id,
-            type=el.type.capitalize(),
+            type=el.type.capitalize() if el.type else None,  # handle empty strings
             properties=el.properties,
         )
         for el in nodes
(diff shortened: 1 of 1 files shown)


ccurme (approved):

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -576,13 +576,17 @@ def _convert_to_graph_document(
     else:  # If there are no validation errors use parsed pydantic object
         parsed_schema: _Graph = raw_schema["parsed"]
         nodes = (
-            [map_to_base_node(node) for node in parsed_schema.nodes]
+            [map_to_base_node(node) for node in parsed_schema.nodes if node.id]
```
jexp, line 579: Should we also test for empty label?

jexp (approved):
Should probably check for empty label too

libs/experimental/langchain_experimental/graph_transformers/llm.py
tomasonjo: If empty label is passed it uses default "Node" label

tomasonjo: So I wouldn't skip records if node labels are missing, as we just use default node label in that case

tomasonjo:
Added explicit check for empty strings in node labels

jexp (approved):
LGTM

ccurme (approved):

PR #23370, opened 2024-06-24
[Core] Unify function schema parsing

hinthornw:
Use pydantic to infer nested schemas and all that fun.
Include bagatur's convenient docstring parser
Include annotation support

Previously we didn't adequately support many typehints in the bind_tools() method on raw functions (like optionals/unions, nested types, etc.)

### DIFF
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -318,9 +318,9 @@ class ChildTool(BaseTool):
     You can use these to eg identify a specific instance of a tool with its use case.
     """
 
-    handle_tool_error: Optional[Union[bool, str, Callable[[ToolException], str]]] = (
-        False
-    )
+    handle_tool_error: Optional[
+        Union[bool, str, Callable[[ToolException], str]]
+    ] = False
     """Handle the content of the ToolException thrown."""
 
     handle_validation_error: Optional[
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -28,7 +28,20 @@ from abc import ABC, abstractmethod
 from contextvars import copy_context
 from functools import partial
 from inspect import signature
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/tools.py
```diff
@@ -89,19 +123,83 @@ def _create_subset_model(
...
     valid_keys = signature(func).parameters
-    return {k: schema[k] for k in valid_keys if k not in ("run_manager", "callbacks")}
+    return {
+        k: schema[k]
+        for i, (k, param) in enumerate(valid_keys.items())
+        if k not in filter_args and (i > 0 or param.name not in ("self", "cls"))
+    }
+
+
+def _parse_python_function_docstring(function: Callable) -> Tuple[str, dict]:
```
eyurtsev, line 151: My opinion for the parser is that the default would've ideally been opt-in from the get-go AND it would've raised exceptions on incomplete doc-string.

Given we can't do that right now are we sure that we want opt-in by default? Instead of adding an option to fail loudly?

eyurtsev (approved):

ccurme (commented):
A few thoughts:

IMO it could be nice to de-couple some of the changes here. Expanding the scope of docstring parsing worries me a little. Consider the scenario where docstrings are outdated, and tool calling was previously working in spite of the outdated docstrings (e.g., relying on type hints, which may be less likely to get outdated due to linting)-- could this change adversely impact users in that situation?

---

Using the tool parsing for functions feels like a straightforward win. There are edge cases (probably undocumented) where it could introduce errors. For example, this used to work, but now raises ValueError:
```python

class float:
    pass

def my_function(x: float) -> None:
    pass

convert_to_openai_function(my_function)
```
I can't think of any errors that I'd consider very realistic scenarios though.

hinthornw:
I'm down to not parse docstrings by default / separate in another pr

hinthornw:
Git rebase 🤦

ccurme (approved):

PR #23375, opened 2024-06-24
core: `callbacks` docstrings

leo-gan:

### DIFF
--- a/libs/core/langchain_core/callbacks/base.py
+++ b/libs/core/langchain_core/callbacks/base.py
@@ -1,4 +1,5 @@
-"""Base callback handler that can be used to handle callbacks in langchain."""
+"""Base callback handler for LangChain."""
+
 from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, TypeVar, Union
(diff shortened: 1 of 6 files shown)


libs/core/langchain_core/callbacks/base.py
```diff
@@ -125,7 +155,13 @@ def on_agent_finish(
         parent_run_id: Optional[UUID] = None,
         **kwargs: Any,
     ) -> Any:
-        """Run on agent end."""
+        """Run on the agent end.
+
+        Args:
+            finish (AgentFinish): The agent finish.
```
eyurtsev, line 161: What do you think if we removed the type information from the args section? That doc-style precedes the usage of type annotations in python.

eyurtsev (approved):

libs/core/langchain_core/callbacks/base.py
leo-gan: @eyurtsev I have mixing feeling about it. Say, this [compress_document](https://api.python.langchain.com/en/latest/retrievers/langchain.retrievers.document_compressors.cohere_rerank.CohereRerank.html#langchain.retrievers.document_compressors.cohere_rerank.CohereRerank.compress_documents) function.
We have types in two places:
1. function signature
2. function arguments (incorrectly named "Parameters" here)

#2 is structured (in a column) and easy to read and comprehend, #1 is messy and hard to comprehend.
So, I would rather remove types from #1 not from #2. But, #1 is still helpful because I still comprehend the function signature as a complete entity.
Resume:
- I like to have both, #1 and #2 type desctiptions, even it breaks the DRY principle. The usability principle is more important.

Note: type information in the ["Methods" table](https://api.python.langchain.com/en/latest/graphs/langchain_community.graphs.graph_store.GraphStore.html#langchain_community.graphs.graph_store.GraphStore) and in the ["Classes" and "Functions"](https://api.python.langchain.com/en/latest/community_api_reference.html#id30) tables is useless. That's for sure.

Issue #23383, opened 2024-06-25
DOC: <Issue related to /v0.2/docs/how_to/extraction_examples/>


mail2mhossain:
### URL

https://python.langchain.com/v0.2/docs/how_to/extraction_examples/

### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

Getting following error:

Traceback (most recent call last):
  File "Z:\llm_images\extract_info.py", line 148, in <module>
    response = chain.invoke(
               ^^^^^^^^^^^^^
  File "Z:\conda_env\ccs\Lib\site-packages\langchain_core\runnables\base.py", line 2504, in invoke
    input = step.invoke(input, config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\conda_env\ccs\Lib\site-packages\langchain_core\output_parsers\base.py", line 169, in invoke
    return self._call_with_config(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "Z:\conda_env\ccs\Lib\site-packages\langchain_core\runnables\base.py", line 1598, in _call_with_config
    context.run(
  File "Z:\conda_env\ccs\Lib\site-packages\langchain_core\runnables\config.py", line 380, in call_func_with_variable_args
    return func(input, **kwargs)  # type: ignore[call-arg]
           ^^^^^^^^^^^^^^^^^^^^^
  File "Z:\conda_env\ccs\Lib\site-packages\langchain_core\output_parsers\base.py", line 170, in <lambda>
    lambda inner_input: self.parse_result(
                        ^^^^^^^^^^^^^^^^^^
  File "Z:\conda_env\ccs\Lib\site-packages\langchain_core\output_parsers\openai_tools.py", line 196, in parse_result
    pydantic_objects.append(name_dict[res["type"]](**res["args"]))
                            ~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'Person'

### Idea or request for content:

_No response_

wulifu2hao:
I can't reproduce your issue following the doc. please give a complete script to reproduce your issue

mail2mhossain:
from typing import List, Optional
import uuid
from typing import Dict, List, TypedDict
from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
from langchain_core.messages import (
    AIMessage,
    BaseMessage,
    HumanMessage,
    SystemMessage,
[...]

PR #23393, opened 2024-06-25
docs[patch]: extraction_examples fix the examples given to the llm

wulifu2hao:
Descriptions: currently in the [doc](https://python.langchain.com/v0.2/docs/how_to/extraction_examples/) it sets "Data" as the LLM's structured output schema, however its examples given to the LLM output's "Person", which causes the LLM to be confused and might occasionally return "Person" as the function to call

issue: #23383


ccurme (approved):
Thank you!

PR #23412, opened 2024-06-25
core: `chat_*` docstrings

leo-gan:


leo-gan:
@eyurtsev Thanks!

PR #23244, opened 2024-06-21
core[minor]: configurable universal model

baskaryan:
```python
from langchain.chat_models import init_chat_model

configurable_model = init_chat_model()
configurable_model.invoke("how are you", config={"configurable": {"model": "gpt-4o"}})
```

Screenshot 2024-06-20 at 9 36 30 PM

PR #23419, opened 2024-06-25
langchain[minor]: Generic configurable model

baskaryan:
alternative to [23244](https://github.com/langchain-ai/langchain/pull/23244). allows you to use chat model declarative methods

Screenshot 2024-06-25 at 1 07 10 PM


libs/langchain/langchain/chat_models/base.py
```diff
@@ -18,10 +25,144 @@
...
+                *args: Any, config: Optional[RunnableConfig] = None, **kwargs: Any
+            ) -> Any:
+                return getattr(self._model(config), name)(
+                    *args, config=config, **kwargs
+                )
+
+            return from_config
+        elif self._default_config and hasattr(self._model(), name):
+            return getattr(self._model(), name)
+        else:
```
baskaryan, line 81: raise error if not defined on BaseChatModel

libs/langchain/langchain/chat_models/base.py
```diff
@@ -18,10 +25,144 @@
...
+        Bind config to a Runnable, returning a new Runnable.
+        """
+        return _ConfigurableModel(
+            default_config={
+                **self._default_config,
+                **((config or {}).get("configurable", {})),
+                **kwargs,
+            },
+            config_prefix=self._config_prefix,
+            configure_any=self._configure_any,
```
baskaryan, line 123: pass in configured_operations

libs/langchain/langchain/chat_models/base.py
```diff
@@ -18,10 +25,144 @@
...
+        self,
+        *,
+        default_config: Optional[dict] = None,
+        config_prefix: Optional[str] = None,
+        configure_any: bool = False,
+    ) -> None:
+        self._default_config = default_config or {}
+        self._config_prefix = config_prefix
+        self._configure_any = configure_any
+        self._configured_operations = []
```
baskaryan, line 66: rename. curried_operations? queued_operations? lazy_operations?

libs/langchain/langchain/chat_models/base.py
```diff
@@ -77,7 +106,120 @@ def init_chat_model(
...
+        .. code-block:: python
+
+            # pip install langchain langchain-openai langchain-anthropic
+            from langchain.chat_models import init_chat_model
+
+            # We don't need to specify if config_prefix or configure_any is specified.
+            configurable_model_with_default = init_chat_model(
+                "gpt-4o",
+                model_provider="openai",
+                config_prefix="foo",
```
ccurme, line 142: When would a user want to provide a prefix?

libs/langchain/langchain/chat_models/base.py
```diff
@@ -77,7 +106,120 @@ def init_chat_model(
...
+                    "configurable": {
+                        "foo_model": "claude-3-5-sonnet-20240620",
+                        "foo_model_provider": "anthropic",
+                        "foo_temperature": 0.6
+                    }
+                }
+            )
+            # Claude-3.5 sonnet response with temperature 0.6
+
+    Bind tools to a configurable model:
```
ccurme, line 162: worth mentioning with_structured_output as well?

libs/langchain/langchain/chat_models/base.py
```diff
@@ -55,19 +68,35 @@ def init_chat_model(
...
+        configurable: Whether model parameters are configurable. Defaults to True if
+            model is None or config_prefix is non-null or configure_any is True.
+        config_prefix: If config_prefix is a non-empty string then model will be
+            configurable at runtime via the
+            ``config["configurable"]["{config_prefix}_model"]`` and
+            ``config["configurable"]["{config_prefix}_model_provider"]`` keys. If
+            config_prefix is an empty string then model will be configurable via
+            ``config["configurable"]["model"]`` and
+            ``config["configurable"]["model_provider"]`` keys.
+        configure_any: If True then any parameter can be passed to model initializer as
```
ccurme, line 80: In what situations would we want `configure_any` to be False?

libs/langchain/langchain/chat_models/base.py
```diff
@@ -200,3 +342,124 @@ def _check_pkg(pkg: str) -> None:
...
+    return s
+
+
+_CHAT_MODEL_METHODS_TAKE_CONFIG = tuple(
+    name
+    for name, func in inspect.getmembers(BaseChatModel, inspect.isfunction)
+    if "config" in inspect.signature(func).parameters
+)
+
+_DECLARATIVE_METHODS = (
```
ccurme, line 359: (nit) is there a (straightforward) way to test this collection or generate it programmatically, as you do for chat model methods?

libs/langchain/langchain/chat_models/base.py
baskaryan: when they've got multiple models in a chain/graph

libs/langchain/langchain/chat_models/base.py
baskaryan: initially tried to but couldn't find a good programmatic way. Runnable has lots of methods that don't take a config and aren't declarative

libs/langchain/langchain/chat_models/base.py
baskaryan: if dev doesn't want to expose certain params, or there's name collisions between model params and params for something else (eg base_url)

eyurtsev: Seems like an easy way to steal someone's API keys if api_url is configurable by mistake

eyurtsev, line 80: config_any=True seems like an easy way to steal someones api keys or potentially do a man in the middle attack by changing the api url

Could we add an option to secify specific field names?

libs/langchain/langchain/chat_models/base.py
```diff
@@ -200,3 +342,124 @@ def _check_pkg(pkg: str) -> None:
...
+    "with_retry",
+    "with_types",
+    "bind_tools",
+    "with_structured_output",
+    "configurable_fields",
+    "configurable_alternatives",
+)
+
+
+class _ConfigurableModel:
```
eyurtsev, line 377: What about actual invoke methods?

libs/langchain/langchain/chat_models/base.py
```diff
@@ -200,3 +342,124 @@ def _check_pkg(pkg: str) -> None:
...
+
+            return from_config
+        elif name in _DECLARATIVE_METHODS:
+
+            def queue(*args: Any, **kwargs: Any) -> _ConfigurableModel:
+                self._queued_operations.append((name, args, kwargs))
+                return self
+
+            return queue
+        elif self._default_config and hasattr(self._model(), name):
```
eyurtsev, line 409: `self._model()` may be expensive to instantiate (e.g., opens network connection), would be good to remove unnecessary instantiations

eyurtsev (approved):

libs/langchain/langchain/chat_models/base.py
```diff
@@ -55,19 +70,37 @@ def init_chat_model(
                 - gemini... -> google_vertexai
                 - command... -> cohere
                 - accounts/fireworks... -> fireworks
+        configurable_fields: Which model parameters are configurable. If
+            - None: No configurable fields.
+            - "any": All fields are configurable.
```
eyurtsev, line 75: Could you add a security note and recommend specifying an explicit list of allowed parameters. If you want you could add the example of the api_url being controllable

libs/langchain/tests/unit_tests/chat_models/test_base.py
```diff
@@ -1,7 +1,16 @@
+import os
+
 import pytest
+from langchain_core.language_models import BaseChatModel
+from langchain_core.messages import HumanMessage
+from langchain_core.runnables import RunnableConfig
 
 from langchain.chat_models.base import __all__, init_chat_model
 
+os.environ["OPENAI_API_KEY"] = os.environ.get("OPENAI_API_KEY", "dummy")
```
eyurtsev, line 10: test code should never set env variables (especially unit test code), could you use mock or a context manager?

eyurtsev (approved):

PR #23500, opened 2024-06-25
core: fix run_manager not being passed to _stream

will-gao42:
Inside `BaseChatModel` class, `stream` function constructs run_manager, but doesn't pass it to `_stream`, rendering the model invocations unable to interact with the callbacks.


nfcampos:
This is not a bug, it's on purpose, BaseChatModel now handles all interaction with run manager

will-gao42:
@nfcampos 

> This is not a bug, it's on purpose, BaseChatModel now handles all interaction with run manager

Exactly, which makes it necessary to pass the run_manager object on as the `_stream` function is invoked, right? It's even in the `_stream` function signature itself. The issue I ran into is **lack of access to run_manager** when my ChatModel invokes its stream function. What are your thoughts?

PR #23504, opened 2024-06-25
langchain[patch]: deprecate ConversationChain

ccurme:
Would like some feedback on how to best incorporate legacy memory objects into `RunnableWithMessageHistory`.

### DIFF
--- a/libs/langchain/langchain/chains/conversation/base.py
+++ b/libs/langchain/langchain/chains/conversation/base.py
@@ -28,7 +28,7 @@ class ConversationChain(LLMChain):
     ``RunnableWithMessageHistory`` offers several benefits, including:
 
     - Stream, batch, and async support;
-    - More flexibile memory handling, including the ability to manage memory
+    - More flexible memory handling, including the ability to manage memory
       outside the chain;
     - Support for multiple threads.
 
--- a/libs/langchain/langchain/chains/conversation/base.py
+++ b/libs/langchain/langchain/chains/conversation/base.py
@@ -1,6 +1,7 @@
 """Chain that carries on a conversation and calls an LLM."""
 from typing import Dict, List
 
+from langchain_core._api import deprecated
 from langchain_core.memory import BaseMemory
 from langchain_core.prompts import BasePromptTemplate
 from langchain_core.pydantic_v1 import Extra, Field, root_validator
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/chains/conversation/base.py
```diff
@@ -10,9 +11,87 @@
...
 class ConversationChain(LLMChain):
     """Chain to have a conversation and load context from memory.
 
+    This class is deprecated in favor of ``RunnableWithMessageHistory``. Please refer
+    to this tutorial for more detail: https://python.langchain.com/v0.2/docs/tutorials/chatbot/
+
+    ``RunnableWithMessageHistory`` offers several benefits, including:
+
+    - Stream, batch, and async support;
+    - More flexibile memory handling, including the ability to manage memory
```
jacoblee93, line 31: Small typo: `flexibile` -> `flexible`

libs/langchain/langchain/chains/conversation/base.py
```diff
@@ -10,9 +11,87 @@
...
+
+
+            store = {}  # memory is maintained outside the chain
+
+            def get_session_history(session_id: str) -> InMemoryChatMessageHistory:
+                if session_id not in store:
+                    store[session_id] = InMemoryChatMessageHistory()
+                    return store[session_id]
+
+                memory = ConversationBufferWindowMemory(
```
jacoblee93, line 76: Do we need to update the memory class after invocation for other types of memory?

jacoblee93: If not this is actually not terrible

libs/langchain/langchain/chains/conversation/base.py
```diff
@@ -10,9 +11,87 @@
...
+    the default ``ConversationBufferMemory``:
+
+        .. code-block:: python
+
+            from langchain_core.chat_history import InMemoryChatMessageHistory
+            from langchain_core.runnables.history import RunnableWithMessageHistory
+            from langchain_openai import ChatOpenAI
+
+
+            store = {}  # memory is maintained outside the chain
```
Noobgam, line 45: At this point it's just safer to say that conversations are not supported by langchain

Introducing a potential endless memory leak just to have an obscure way to be flexible without providing any reasonable default way to pass the chain along with it's memory is just terrible.

How is this an alternative to what conversationChain achieves in 2 lines of code?

ccurme: Hi @Noobgam, I took a look at your [commit](https://github.com/Noobgam/noobgam-monorepo/commit/b1436609a997d65112fa008e1d570c2012f5c2f9) and think there may be a misunderstanding here.

`RunnableWithMessageHistory` will automatically append messages to the conversation history. If you do not need to manage multiple threads, you can do this:
```python
from langchain_core.chat_history import InMemoryChatMessageHistory
from langchain_core.runnables.history import RunnableWithMessageHistory

llm = ChatOpenAI(model="gpt-3.5-turbo-0125")
memory = InMemoryChatMessageHistory()
chain = RunnableWithMessageHistory(llm, lambda x: memory)
```
```python
message = chain.invoke(
    "Hi I'm Bob.",
    config={"configurable": {"session_id": "1"}},
)
print(message.content)
```
```
Hello Bob! How can I assist you today?
```
```python
message = chain.invoke(
    "What is my name?",
    config={"configurable": {"session_id": "1"}},
)
print(message.content)
```
```
Your name is Bob.
```
The only change you must make is to configure each invocation with a session ID representing the thread (although here the ID is ignored and there is effectively one thread).

Let me know if I've misunderstood your concerns.

Noobgam: My concern here is that the `memory` object will tend to leak too much.

As you outline with your example this is (very) likely to be globally scoped for a lot of cases.
This change
1. makes the interface incredibly complicated to use compared to conversational chain. The core idea of `ConversationChain` to begin with to have a super short way to create a conversation bot that would hold it's own memory. This by itself has a lot of usecases, and as an active langchain user at work I would argue this is the majority of cases, whereas `RunnableWithMessageHistory` covers only some obscure usecases where library consumers actually need to actively maintain multiple conversations.

2. Doesn't exactly match what you have outlined in the documentation. That actually seems to be to a certain extent safer solution that has a lot less security implications, so to me it makes a lot more sense to include that example rather than what you've done in the PR.

The reason I originally stumbled upon this is because ConversationChain + `langchain-community` is not compatible with anthropic latest models (I haven't tried exactly with `langchain-community-anthropic` but I presume it might actually work). That made me look for an alternative, but the docstring proposed solution makes my code explode ton of boilerplate of this `"configuration: {...}"`

I can briefly touch how we use it in our work projects without mentioning any details:
1. Insantiating `ConversationChain` once and passing it around as a "single entrypoint to the whole conversation" around the code. So you just call `chain.predict("text prompt")` and get your results you need.
2. Forking `ConversationChain` in order to achieve debate (see this [paper](https://arxiv.org/abs/2305.14325) or this [paper](https://arxiv.org/pdf/2305.10601) to have details). This was easily achievable via just copying `ConversationChain` memory and creatign a new one, just like some deep cloning thing.
3. Our company is very security concerned, because our service is multitenant. Hence any pull request that potentially introduces potential vulnurability as in "conversations of users are still in memory after the chain is discarded" / "multiple user conversations are stored in the same place, so 1 typo can lead to a disasterous customer trust loss"

So TL;DR, `ConversationChain` is bettter because
1. Less boilerplate
2. Seemingly easier flexibility for a lot of State of the art usecases.
3. It has next to zero surface to attack.

If you can outline how `RunnableWithMessageHistory` is better at resolving these issues, I would be happy to actually work on refining `ConversationChain` to accomodate and align more with langchain's mainters point of view.

PR #23512, opened 2024-06-26
core[patch]: add filter for input/output messages

mackong:
**Description:** 
For some LLMs, the output of the last conversation cannot be completely used as the input for the next conversation, especially function calling conversation, so we need a way to ignore unsupported messages got from last conversation when saved to history.

This patch adds input/output messages filter before (a)add_messages.

**Issue:** N/A
**Dependencies:** N/A

### DIFF
--- a/libs/core/langchain_core/runnables/history.py
+++ b/libs/core/langchain_core/runnables/history.py
@@ -15,9 +15,11 @@ from typing import (
 
 from langchain_core.chat_history import BaseChatMessageHistory
 from langchain_core.load.load import load
+from langchain_core.messages import BaseMessage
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/runnables/history.py
```diff
@@ -27,13 +29,14 @@
 
 if TYPE_CHECKING:
     from langchain_core.language_models.base import LanguageModelLike
-    from langchain_core.messages.base import BaseMessage
-    from langchain_core.runnables.config import RunnableConfig
```
mackong, line 31: This change is used to fix ```Pydantic Optional parameters not yet prepared so type is still a ForwardRef```

Issue #23394, opened 2024-06-25
RecursiveCharacterTextSplitter uses regex value instead of original separator when merging and keep_separator is false


loperamos:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain.text_splitter import RecursiveCharacterTextSplitter


if __name__ == "__main__":
    # Wrong behaviour, using \s instead of regular space
    splitter_keep = RecursiveCharacterTextSplitter(
        separators=[r"\s"],
        keep_separator=False,
        is_separator_regex=True,
        chunk_size=15,
        chunk_overlap=0,
        strip_whitespace=False)
    assert splitter_keep.split_text("Hello world")[0] == r"Hello\sworld"

    # Expected behaviour, keeping regular space
    splitter_no_keep = RecursiveCharacterTextSplitter(
        separators=[r"\s"],
        keep_separator=True,
        is_separator_regex=True,
        chunk_size=15,
        chunk_overlap=0,
        strip_whitespace=False)
    assert splitter_no_keep.split_text("Hello world")[0] == r"Hello world"
```

### Description

I am trying to use the `langchain` library to split a test using regex separators. I expect the output strings to contain the original separators, but what happens is that when using the `keep_separator` flag as `False` it uses the regex value instead of the original separator.

Possible code pointer where the problem might be coming from: [libs/text-splitters/langchain_text_splitters/character.py#L98](https://github.com/langchain-ai/langchain/blob/master/libs/text-splitters/langchain_text_splitters/character.py#L98)

### System Info

langchain==0.2.5
langchain-core==0.2.9
langchain-text-splitters==0.2.1

Platform: Apple M1 Pro
macOS: 14.5 (23F79)

python version: Python 3.12.3

loperamos:
Adding some post clarification, I believe that the expected behaviour should be as the following code snippet, since I would expect we want to keep the chunk exactly as the original text. The keep separator flag should only affect when the separator is at the begining of a chunk.
```python
if __name__ == "__main__":
    # Wrong behaviour, using \s instead of regular space
    splitter_keep = RecursiveCharacterTextSplitter(
        separators=[r"\s"],
        keep_separator=False,
[...]

PR #23519, opened 2024-06-26
text-splitters: bug fix for CharacterTextSplitter replacing original separator with regex pattern when merging

wulifu2hao:
- **Description:** 
  bug: currently when using CharacterTextSplitter with regex pattern as separator (e.g. "\s") and keep_separator=False to split a text like "Hello world", it will end up with "Hello\sworld" (assume chunk_size=15 so after the splitting it will need to merge it back)
  
  changes: 
  1. "_split_text" function will use the actual original separators to merge split strings, instead of the regex pattern 
  2. "_split_text_with_regex" function will return not just the split strings but also the original separators 
  3. "_merge_splits" function of TextSplitter in base.py will pick the right separators to join two split strings instead of always using one separator (which could be the regex pattern)
  4. "_join_docs" function of TextSplitter in base.py, similarly, allows passing in a list of separators the facilitate the joining of strings required in the "_merge_splits" function 
    

**Issue:**  #23394 


wulifu2hao:
@baskaryan  pls do help review when you have time , thanks :)

efriis (dismissed):

efriis:
thoughts on these new tests?

efriis:
closing as stale!

PR #23531, opened 2024-06-26
langchain[minor] : Fix embeddings filter if the retriever return an empty list

pprados:
Assume the stateful_documents = get_stateful_documents(documents) may return an empty list.


ccurme:
Thanks for this.

Can you share a traceback you're seeing, just to clarify the error?

One scenario where `stateful_documents` will be empty is if the input `documents` is an empty list. is there another scenario? If not, I'd just check that at the top of the function and return `documents` for clarity.

pprados:
It's not confirmed

PR #23544, opened 2024-06-26
RFC indexing interface

eyurtsev:
# RFC 

Interface for indexing and vectorstores.

## Goals

1. Have a well-defined READ & WRITE interface for indexing
    - Upsert instead of add
    - Support upsert, delete and read by ID
2.  Allow supporting binary data down the road
3. Have a well-defined query API on retrieval
    - Define filters, limit, offset, sort
    - Support similarity search
    - Enforcement can be done by standard unit-tests against the implementation.
    - Implementations can be required to either self-describe supported features
      or else just raise a NotImplementedError for unsupported features, so
      no silent failures occur.
    - We may need to throw in an extra field for disambiguating between
      a query executing using the `raw` interface vs. via a standardized langchain
      interface.

The query API should allow us to support adhoc capabilities like:
    - mmr search
    - setting min score threshold etc

## Discussion points

1. Force all content to have metadata OR only ID?
    - If it's only ID, then we don't support the ability to filter by metadata in the query API.
    Abstraction is more generic, and we can sub-class to add an interface that supports metadata operations.
2. Base Indexing abstraction forced to support metadata?
3. Support `get_with_extras` for indexers or vectorstores? (See below)

4. Two options for how to deal with the retriever interface:
    - Have a standalone interface, and a vectorstore implements both indexer and retriever interfaces.
    - The retriever interface is defined in the vectorstore abstraction itself (which inherits from indexing interface).

5. Need to determine if we try to add an API that will support sync primitives.
    - This is not needed for the initial version of the API.
    - Requires an indexed integer or datetime field to be updatable, and filterable

**Variations**

https://github.com/langchain-ai/langchain/pull/22102


libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+        *,
+        config: Optional[RunnableConfig] = None,
+        **kwargs: Any,
+    ) -> QueryResponse[
+        Document
+    ]:  # kwargs for things that are not captured via the standard query?
+        """Query for items."""
+
+    @abc.abstractmethod
+    def upsert_with_vector(
```
eyurtsev, line 243: It's nice to be able to provide additional data during indexing. For example, for a vectorstore being able to specify the vector is a very obvious thing. What's a good way to represent this while still keeping the upsert interface?

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+
+    This query format is specifically adapted to vectorstores.
+
+    It supports similarity searches together with standard filtering operations.
+
+    We would need to verify that we can accommodate all the existing functionality
+    with it.
+    """
+
+    query: Optional[T]
```
eyurtsev, line 215: This is a very naive parameterization of a search query. They can get much more complex in realistic situations;

e.g., match on both similarity of text as well as presence of specific keywords for a hybrid score

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+    Any,
+    Literal,
+    Dict,
+)
+
+from langchain_core.load.serializable import Serializable
+from langchain_core.runnables import RunnableConfig
+
+
+class IndexableData(
```
baskaryan, line 62: nothing that really makes this Data based on its contents

if we wanted could do 
```python
class IndexableData(Serializable, Generic[D], ABC):
  id: str
  metadata: dict = {}
  data: D

class Document(IndexableData[str]):
  @property
  def content(self) -> str:
    return self.data

  @content.setter
  def content(self, content: str) -> None:
    self.data = content
```

or keep as is and give it a more generic name (IDObject, LCObject, ...)

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+):  # Introduce new base class for content. Name OK?
+    id: str
+    metadata: Optional[dict]  # In a more extreme form we could remove
+
+
+class Document(IndexableData):  # Re-use existing Document model in langchain_core
+    content: str
+
+
+class Blob(IndexableData):  # Re-use existing Blob model in langchain_core
```
baskaryan, line 73: would be great to be able to use this in messages as well (nothing to change just noting)

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+T = TypeVar("T", bound=IndexableData)
+
+
+class UpsertResponse(TypedDict):
+    """An indexing result."""
+
+    failed: Sequence[str]
+    indexed: Sequence[str]
```
baskaryan, line 95: nit why not just succeeded

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class DeleteResponse(TypedDict):
+    """A response to a delete request."""
+
+    num_deleted: NotRequired[int]
+    num_failed: NotRequired[int]
+    failed: NotRequired[Sequence[str]]
+    deleted: NotRequired[Sequence[str]]
```
baskaryan, line 104: same q as above

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class Sort(TypedDict):
+    """A sort object."""
+
+    field: str
+    order: NotRequired[Literal["asc", "desc"]]  # Assume asc by default
```
baskaryan, line 111: if its only every ascending/descending would prefer a bool

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class FilterQuery(TypedDict):
```
baskaryan, line 114: just Query? more than just a filter

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+    """Query for an item.
+
+    This enables querying for an item using similarity search +
+    standard operations on the relational data associated with the item.
+    """
+
+    filter: Optional[dict]  # We'll need to type this # <-- Could stand for raw filter?
```
baskaryan, line 121: guessing we'll want to support a list of dict for usability (and that represents conjunction or disjunction, whichever is more common)

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+    limit: Optional[int]  # equivalent to top-k right now
+    offset: Optional[int]  # e.g., interpret as page offset (maybe a better name?)
+    # e.g., browse through documents by publication date
+    sort: Optional[List[Sort]]  # Optional multiple sort fields
+
+
+C = TypeVar("C")
+
+
+Vector = List[float]
```
baskaryan, line 131: nit: these make the api reference confusing bc the aliases aren't resolved to their actual types and they're not documented

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+        self,
+        ids: Iterable[str],  # Sequence or Iterable
+        /,
+    ) -> Iterable[T]:
+        """Get items by id."""
+
+    @abc.abstractmethod
+    def delete_by_ids(
+        self,
+        ids: Sequence[str],  # Sequence or Iterable
```
baskaryan, line 165: why would get and delete be diff (re sequence or iterable)

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+        **kwargs: Any,
+    ) -> DeleteResponse:
+        """Delete an item by id."""
+
+    # Delete and get are part of the READ/WRITE interface.
+    # They do not take advantage of indexes on the content.
+    # However, all the indexers ARE assumed to have the capability to index
+    # on metadata if they implement the delete_by_query and get_by_query methods.
+    @abc.abstractmethod
+    def lazy_get(self, query: FilterQuery, /, **kwargs: Any) -> Iterable[T]:
```
baskaryan, line 176: why prefix with lazy_ if get_by_ids isn't prefixed with lazy_

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @abc.abstractmethod
+    def lazy_delete(
+        self, query: FilterQuery, /, **kwargs: Any
+    ) -> Iterable[DeleteResponse]:
+        """Delete items by query."""
+        # Careful with halloween problem
+
+
+class Hit(TypedDict):
```
baskaryan, line 187: shoudl be generic? (can a typed dict be generic)

libs/core/langchain_core/indexing/schema_02.py
baskaryan, line 215: why T? are you searching for Documents by Documents?

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+
+    query: Optional[T]
+    # Optionally specify the method to use for similarity search
+    method: NotRequired[Optional[str]]  # e.g., "cosine", "euclidean", "jaccard", "mmr"
+
+
+class UpsertData(TypedDict):
+    """Data to upsert."""
+
+    id: str
```
baskaryan, line 223: doesn't Document have id?

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+        **kwargs: Any,
+    ) -> QueryResponse[
+        Document
+    ]:  # kwargs for things that are not captured via the standard query?
+        """Query for items."""
+
+    @abc.abstractmethod
+    def upsert_with_vector(
+        self,
+        data: Iterable[UpsertData],
```
baskaryan, line 245: seems kinda annoying (needing to zip together vectors and docs)

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,269 @@
...
+        """Upsert vectors by id."""
+
+    # --------------
+    # TODO: Discuss whether we want this
+    class ContentWithExtras(TypedDict, Generic[T]):
+        content: T
+        metadata: dict
+
+    @abc.abstractmethod
+    def get_with_extras(
```
baskaryan, line 258: could query just support get by id

libs/core/langchain_core/indexing/schema_02.py
eyurtsev: expressing my uncertainty -- I'm inclined to use Sequence for both

libs/core/langchain_core/indexing/schema_02.py
eyurtsev: In a vectorstore we could search for documents by text or by vector at minimum. If we add multi modal support, then we could search for them through an image or some other data

libs/core/langchain_core/indexing/schema_02.py
eyurtsev: Yes. I'm not sure about how to do this yet.

Document has an optional ID right now (not in this schema, but in terms of what we're about to introduce in LangChain). 

We could an `id` at top level which is non required that will match the existing vectorstore APIs it'll just keep the data together instead of forcing users to provide parallel lists

libs/core/langchain_core/indexing/schema_02.py
eyurtsev: Not the opposite? They usually appear zipped, and we require users to unzip them right now, and for vectorstore implementations to re-zip them?

i.e., i'm thinking that it's easier to think in terms of a list of indexable items. 

For the more generic interface, the item contains the document + the vector

libs/core/langchain_core/indexing/schema_02.py
eyurtsev: Yeah that's the thing to do -- we'll need to standardize syntax for the ID field. Perhaps `$.ID`

libs/core/langchain_core/indexing/schema_02.py
baskaryan: also for usability maybe we just unpack query into separate args

eyurtsev: I packaged it to make it easier to roll out changes, but yeah we can unpack, and rename to get_by_query?

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,235 @@
...
+
+    This allows providing additional information such as the vector.
+    """
+
+    id: str  # An override of the ID for the document
+    document: Document
+    vector: List[float]
+
+
+class VectorStore(BaseIndex[Document]):
```
eyurtsev, line 212: Vectorstore implementation over Document, but can look differently for a vectorstore over images (i.e., different query)

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,235 @@
...
+
+    1. Storing the content
+    2. Allowing retrieval of the content by id
+    3. Supporting queries against the metadata associated with the content
+
+    The index is **NOT** responsible for:
+
+    1. Support search queries against the content.
+
+    The types of supported queries seems very dependent on the actual
```
eyurtsev, line 130:
```suggestion
    The types of supported queries depends on the implementation (e.g., a vectorstore index)
```

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+
+    content: T
+    vector: List[float]
+
+
+class VectorStoreGeneric(Generic[Q, T], BaseIndex[T]):
+    @abc.abstractmethod
+    def query(
+        self, query: Q, config: Optional[RunnableConfig] = None, **kwargs: Any
```
eyurtsev, line 257: This version parameterizes the query, so we could define something like


```python
class Query(TypedDict):
  query: Union[str, List[float]]
  filter: ...

```
And allow each vectorstore to vary the representation if they need to

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,265 @@
...
+)
+
+from langchain_core.load.serializable import Serializable
+from langchain_core.runnables import RunnableConfig
+
+
+class IndexableContent(
+    ABC, Serializable
+):  # Introduce new base class for content. Name OK?
+    id: str
```
eyurtsev, line 65: cc @cbornet  -- this might not be a good name, but related to the initial version of the node you had

https://github.com/langchain-ai/langchain/pull/23092/files#diff-151dbb86c9cfb0578d1f425781066107852cd0bd4b4caf0077baf605d0d8c632R41

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,265 @@
...
+    """A sort object."""
+
+    field: str
+    ascending: NotRequired[bool]  # Assume asc=True
+
+
+C = TypeVar("C")
+
+
+class BaseIndex(Generic[T]):
```
eyurtsev, line 117: @cbornet this is very similar to @baskaryan's previous PR -- mostly a subset of @baskaryan's proposed interface (https://github.com/langchain-ai/langchain/pull/22102)

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # Delete and get are part of the READ/WRITE interface.
+    # They do not take advantage of indexes on the content.
+    # However, all the indexers ARE assumed to have the capability to index
+    # on metadata if they implement the delete_by_query and get_by_query methods.
+    @abc.abstractmethod
+    def get_by_query(
+        self,
+        *,
+        filter: Optional[Union[List[Dict[str, Any], Dict[str, Any]]]] = None,
```
eyurtsev, line 162: Alternatives:

1. Use a generic for the entire `Query`


```
class Indexer(Generic[Q]):
  def get_by_query(
     self, query: Q
  ):
```

This seems to be the most bullet proof way to handle different types of queries.

2. Use a generic on the filter

```python
  @abc.abstractmethod
    def get_by_query(
        self,
        *,
        filter: Optional[F] = None,
```

---

In each case the goal would be to allow implementations to support both a LangChain representation for filters/queries  as well supporting implementation native functionality (which is usually much richer than what LangChain can accommodate)

---

The main concern is for end-user UX and having to input dictionaries instead of relying on the function signature more directly

libs/core/langchain_core/indexing/schema_02.py
eyurtsev: I think might be easier to not rely on generics for this

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,265 @@
...
+    """A structured representation of a data item to upsert.
+
+    This allows providing additional information such as the vector.
+    """
+
+    content: T
+    vector: List[float]
+
+
+class VectorStoreGeneric(Generic[Q, T], BaseIndex[T]):
```
cbornet, line 254: It seems there could be an intermediate "queryable index" that is independent of vector.
For instance in ColBERT, the similarity search is not done on a vector/`list[float]`, it's a done on a `list[list[float]`.

eyurtsev: Yeah I agree. The main issue that's blocking that is whether we try to specify the full function signature in the abstraction or try to capture via a generic. 

My main concern is that Unpack isn't well supported: https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.Unpack

libs/core/langchain_core/indexing/schema_02.py
lkuligin, line 65: a minor question - what if id doesn't exist and it's created on the vectorstore side?

eyurtsev: Check out this PR for `BaseMedia`

Ended up merging an Optional ID to allow vectorstores to assign uuids for now. We may make it required in the future so vector stores don't have to implement extra conditional logic on the ID field

libs/core/langchain_core/indexing/schema_02.py
```diff
@@ -0,0 +1,265 @@
...
+class Document(IndexableContent):  # Re-use existing Document model in langchain_core
+    content: str
+
+
+class Blob(IndexableContent):  # Re-use existing Blob model in langchain_core
+    """Use for images / audio / video"""
+
+    data: bytes
+    encoding: str  # base64, utf-8, URI, etc.
+    mimetype: str  # image/png, text/plain, etc.
```
bjchambers, line 78: Random thought: Could all indexable content by the same? Isn't a Document just a `Blob` with `encoding = "utf-8"` and `mimetype="text/plain"`? It may simplify things if the hierarchy is collapsed.

libs/core/langchain_core/indexing/schema_02.py
bjchambers: It seems like it would be better for this to be part of the `BaseMedia`, so that the ID generation is consistent. Specifically, while it would be easy to use a random ID or timestamp, it may be better to hash the content and metadata. That would ensure the ID is consistent and usable for de-duplicating content being inserted multiple times.

eyurtsev: Yes, this will be the `BaseMedia`.

I am not sure on how to handle the hashes yet. `ids` can be meaningful (e.g., a natural key or a foreign key into another database), so I don't know if we should prevent users from using their own IDs. 

Hashes are obviously useful for de-duplication logic. It's used right now internally in the indexing API: 

https://github.com/langchain-ai/langchain/blob/91b37b2d812eea7fbd2e5c8338c0776826a54c8b/libs/core/langchain_core/indexing/api.py#L56-L56

libs/core/langchain_core/indexing/schema_02.py
eyurtsev: Yes, I think we'll try to do that. We need to do a bit of additional work to make sure that deserialization can load data that was serialized with older versions of LangChain.

eyurtsev:
Superseded by: https://github.com/langchain-ai/langchain/pull/24206

eyurtsev:
We're going to close this PR for now. The main reason is that it picks up too much complexity for functionality that while may be useful isn't super critical.

We do have the much simpler DocumentIndexer abstraction right now. (https://python.langchain.com/api_reference/core/indexing/langchain_core.indexing.base.DocumentIndex.html#langchain_core.indexing.base.DocumentIndex)

PR #23547, opened 2024-06-26
community: Replace deprecated setVectorProperty with setNodeVectorProperty

mejhana:
Description: Replace deprecated `setVectorProperty` with `setNodeVectorProperty` in Cypher query to update node embeddings. Since setNodeVectorProperty doesn't return the nodes, the cypher query returns and yields the nodes instead. 
(https://neo4j.com/docs/operations-manual/5/reference/procedures/#procedure_db_create_setNodeVectorProperty)

Dependencies: No dependencies have been introduced.

Tests: The change is covered by previous unit tests.

### DIFF
--- a/libs/community/langchain_community/vectorstores/neo4j_vector.py
+++ b/libs/community/langchain_community/vectorstores/neo4j_vector.py
@@ -1441,9 +1441,10 @@ class Neo4jVector(VectorStore):
                 "UNWIND $data AS row "
                 f"MATCH (n:`{node_label}`) "
                 "WHERE elementId(n) = row.id "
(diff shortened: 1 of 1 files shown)


jexp:
Thanks @mejhana  

It will break versions earlier than 5.13 but I think that's ok as we had 5.11 to 5.13 only a few versions and are now at 5.21.

https://neo4j.com/docs/cypher-manual/current/indexes/semantic-indexes/vector-indexes/#procedures

ccurme:
Closing as this was done separately in https://github.com/langchain-ai/langchain/pull/23725, apologies as it looks like this one was first.

PR #23548, opened 2024-06-26
docs[patch]: Update installation guide with diagram

jacoblee93:
CC @baskaryan


docs/docs/how_to/installation.mdx
```diff
@@ -21,48 +24,61 @@ import CodeBlock from "@theme/CodeBlock";
...
-A lot of the value of LangChain comes when integrating it with various model providers, datastores, etc.
+While this package acts as a sane starting point to using LangChain,
+much of the value of LangChain comes when integrating it with various model providers, datastores, etc.
 By default, the dependencies needed to do that are NOT installed. You will need to install the dependencies for specific integrations separately.
+We'll show how to do that in the next sections of this guide.
+
+## Ecosystem packages
+
+With the exception of the `langsmith` SDK, all packages in the LangChain ecosystem depend on `langchain-core`, which contains base
+classes and abstractions that other packages use.
```
baskaryan, line 35: could we add something like "The dependency graph below shows how the difference packages are related — a directed arrow indicates that the source package depends on the target package."

baskaryan (commented):
one nit, otherwise lgtm

jacoblee93:
Done thanks!

PR #23551, opened 2024-06-26
partners[openai]: Add name field to tool message to match OpenAI spec

andrewmjc:
Discovered alongside @t968914

  - **Description:**
According to OpenAI docs, tool messages (response from calling tools) must have a 'name' field.
https://cookbook.openai.com/examples/how_to_call_functions_with_chat_models

  - **Issue:** N/A (as of right now)
  - **Dependencies:** N/A
  - **Twitter handle:** N/A

Additional guidelines:


ccurme:
Hello, thanks for this.

I do see in the cookbook you linked that the completions endpoint accepts a tool message with a `"name"` field.

I have a slight hesitation because in OpenAI's [API reference](https://platform.openai.com/docs/api-reference/chat/create), this parameter is not documented (and not required), and OpenAI in the past has removed support for undocumented parameters (which is why we include those checks).

Screenshot 2024-06-27 at 11 00 34 AM

So there is arguably some risk to this, depending on how much weight you give to the documentation in the cookbooks vs. the API reference.

I'm interested in hearing what is the benefit? Is there a scenario or code snippet you could provide to help illustrate what we gain?

andrewmjc:
We ran into this issue while using google gemini's OpenAI API spec endpoint with the ChatOpenAI wrapper. https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/call-gemini-using-openai-library

This API endpoint expects a name field on tool messages, and will error out (HTTP 400) if the key does not exist. Thus, you can't use it as the base api url with the ChatOpenAI wrapper.

 I found the example in the cookbook that mentioned it as well, and so I believe this is in fact a part of the spec.  Additionally, LiteLLM (an OpenAI API spec proxy to other providers) expects this name field when translating from OpenAI to other specs. So it seems that this field has been largely accepted down stream by a lot of the community.

You are right though, this is not clearly documented by OpenAI. However, this key missing breaks many integrations with other services providing the OpenAI API spec

ccurme:
Sounds good @andrewmjc. Out of curiosity, have you considered using the `langchain-google-vertexai` or `langchain-google-genai` packages?

andrewmjc:
I am mainly working with OpenAI API proxies for governance, so I haven't really looked at the other wrappers! Might be worth considering in the future though!

PR #23576, opened 2024-06-27
experimental: handle cases in LLMGraphTransformer where head or tail is None or a list

memgonzales:
- **Description**:

   When using `LLMGraphTransformer` on some Wikipedia articles, there are cases where `rel["head"]` or `rel["tail"]` are `None` or are lists. For example, I encountered a case where `rel["tail"]` is a list of countries like so: `[Philippines, Cambodia, China]`. These cases are currently unhandled, causing `process_response()` to crash.

   This pull request bundles two changes:
   1. Skip the entry if either is `None`.
   2. If `rel["tail"]` is a list but `rel["head"]` is not, then we connect `rel["head"]` to every item in `rel["tail"]`. A similar logic is followed if `rel["head"]` is the one that is a list. If both are lists, then the items in `rel["head"]` and the items in `rel["tail"]` are connected in a pairwise fashion.

- **Issue**:
   No issue found

- **Dependencies**:
  N/A

- **Twitter handle**: 
  `@memgonzales`


memgonzales:
@baskaryan just wanted to follow up on this PR. Thanks!

ingmars1709:
I use this change also locally. When can we expect it to be merged?

memgonzales:
@eyurtsev I merged `master` into this PR. 
Could I request a review? Thanks!

ingmars1709:
I recently came across a document where the invocation `rel["tail_type"]` resulted in a `KeyError`, because the `tail_type` key itself was not present in the `rel` datastructure. 

An addition on this fix would be that you also check on the presence of keys using `get`, e.g.: `rel.get("tail_type") != None`

baskaryan:
cc @tomasonjo

efriis:

@tomasonjo a few PRs have been opened attempting to address this - would be great to get some eyes on a fix!

PR #23591, opened 2024-06-27
community: Fix LanceDB similarity search bug

v4lentin1879:
**Description:** LanceDB didn't allow querying the database using similarity score thresholds because the metrics value was missing. This PR simply fixes that bug.
**Issue:** not applicable
**Dependencies:** none
**Twitter handle:** not available


multiduplikator:
To enable similarity_search_with_score, this fix is required, since we have to set the metric to cosine, to get scores between 0 and 1, as langchain expects it. Default is l2.

This is independent of the create_index functionality. Whatever distance metric we set when creating index, does not influence the metric when doing the search. The distance metric is only used to create the IVF-PQ. Moreover, with LanceDB we should run without index (aka exhaustive search with 100% recall) until up to 100k vectors or even more.

Thanks for this fix.

libs/community/langchain_community/vectorstores/lancedb.py
```diff
@@ -371,10 +371,12 @@ def _query(
 
         prefilter = kwargs.get("prefilter", False)
         query_type = kwargs.get("query_type", "vector")
+        metrics = kwargs.get("metrics", "l2")
```
ccurme, line 374: Double-checking that:

1. This is the current default;
2. We want to maintain a separate default in Langchain.

Otherwise, may be preferable to do
```python
if metrics := kwargs.get("metrics"):
    lance_query = ...
else:
    ...
```

multiduplikator:
Checking LanceDB I can see:

`export enum MetricType {
  /**
   * Euclidean distance
   */
  L2 = "l2",

  /**
   * Cosine distance
   */
  Cosine = "cosine",

  /**
   * Dot product
   */
  Dot = "dot",
}`

libs/community/langchain_community/vectorstores/lancedb.py
v4lentin1879: 1. Thanks @multiduplikator for showing that 'l2' is the current default
2. I then suggest that we leave the default value up to lance and implement the fix as proposed by you. I've updated the PR

ccurme (approved):

PR #23593, opened 2024-06-27
docs: Fix typo in LLMChain tutorial

Doge-is-Dope:
When using `model_with_tools.invoke`, the `content` returns as an empty string. 
For more details, please refer to my [trace log](https://smith.langchain.com/public/6fd24bc4-86c4-4627-8565-9a8adaf4ad7d/r).


ccurme:
Thank you for providing the trace.

Whether content is an empty string or populated will depend on the provider, and the guide is intended to be compatible with multiple providers.

In this case I suspect that Anthropic was used when the notebook was last run, which can output text content blocks when calling tools.

However, to your point, it does say "We can see that there's now no text content" directly below that.

ccurme (approved):

PR #23631, opened 2024-06-28
standard-tests: add test for structured output

ccurme:
- add test for structured output
- fix bug with structured output for Azure
- better testing on Groq (break out Mixtral + Llama3 and add xfails where needed)


ccurme:
Standard tests pass on all packages that implement them: https://github.com/langchain-ai/langchain/actions/runs/9714768840

PR #23636, opened 2024-06-28
core: add RemoveMessage

vbarda:
This change adds a new message type `RemoveMessage`. This will enable `langgraph` users to manually modify graph state (or have the graph nodes modify the state) to remove messages by `id`

Examples:

* allow users to delete messages from state by calling

```python
graph.update_state(config, values=[RemoveMessage(id=state.values[-1].id)])
```

* allow nodes to delete messages

```python
graph.add_node("delete_messages", lambda state: [RemoveMessage(id=state[-1].id)])
```


libs/core/langchain_core/messages/modifier.py
```diff
@@ -0,0 +1,20 @@
...
+from langchain_core.messages.base import BaseMessage
+
+
+class RemoveMessage(BaseMessage):
+    """Message responsible for deleting other messages."""
+
+    type: Literal["remove"] = "remove"
+
+    def __init__(self, id: str, **kwargs: Any) -> None:
+        return super().__init__("", id=id)
```
baskaryan, line 12: probably worth plumbing through kwargs (with check that content isn't one of them)? seems a little more future proof

nfcampos (approved):

libs/core/langchain_core/messages/modifier.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any, List, Literal
+
+from langchain_core.messages.base import BaseMessage
+
+
+class RemoveMessage(BaseMessage):
+    """Message responsible for deleting other messages."""
+
+    type: Literal["remove"] = "remove"
```
eyurtsev, line 9: RemoveMessage better here for 1:1 correspondence

PR #23639, opened 2024-06-28
core[minor]: Create BaseMedia object

eyurtsev:
This PR implements a BaseContent object from which Document and Blob objects will inherit proposed here: 
https://github.com/langchain-ai/langchain/pull/23544

Alternative: Create a base object that only has an identifier and no metadata. 

For now decided against it, since that refactor can be done at a later time. It also feels a bit odd since our IDs are optional at the moment.


libs/core/langchain_core/documents/base.py
```diff
@@ -1,47 +1,269 @@
...
 
-    page_content: str
-    """String text."""
     metadata: dict = Field(default_factory=dict)
-    """Arbitrary metadata about the page content (e.g., source, relationships to other
-        documents, etc.).
+    """Arbitrary metadata associated with the content."""
+
+
+class Blob(BaseContent):
```
jacoblee93, line 39: Probably slightly too late for this but could see `Blob` being a confusing name given it shadows the actual MDN concept name

libs/core/langchain_core/documents/base.py
```diff
@@ -1,47 +1,269 @@
...
+    data: Union[bytes, str, None]
+    """Raw data associated with the blob."""
+    mimetype: Optional[str] = None
+    """MimeType not to be confused with a file extension."""
+    encoding: str = "utf-8"
+    """Encoding to use if decoding the bytes into a string.
+
+    Use utf-8 as default encoding, if decoding to string.
+    """
+    path: Optional[PathLike] = None
```
jacoblee93, line 106: Is this important/general enough to be top level or does it belong in metadata?

libs/core/langchain_core/documents/base.py
```diff
@@ -1,47 +1,269 @@
...
 
+PathLike = Union[str, PurePath]
 
-class Document(Serializable):
-    """Class for storing a piece of text and associated metadata.
-
-    Example:
 
-        .. code-block:: python
+class BaseContent(Serializable):
```
jacoblee93, line 15: Not sure about this name since we use `page_content` already one level up - how about just `BaseMedia`?

eyurtsev: brilliant name!

libs/core/langchain_core/documents/base.py
eyurtsev: It's used to allow specifying the Blob by reference rather than by value (i.e., a URL). We could've probably encoded this information using `data` + `encoding` . Right now need to keep it for backwards compatibility and we could revisit in the future.

libs/core/langchain_core/documents/base.py
eyurtsev: It already exists on the python side, but we can introduce an alias. I wasn't thinking about name collisions on the JS side

libs/core/langchain_core/documents/base.py
eyurtsev: Do you think we need an optional name?

libs/core/langchain_core/documents/base.py
baskaryan, line 39: Blob is all copy/pasted right (aside from inheritance from BaseContent)

eyurtsev: That's right

baskaryan (approved):

PR #23647, opened 2024-06-28
anthropic[patch]: tool output parser fix

baskaryan:


jasonbhart:
This seems to be a bug fix rather than an improvement. I can't currently use with_structured_output() with Anthropopic models. The output is not recognized as valid even though it is, and the same works when simply using a different model (GPT4o)

ccurme (approved):

PR #23666, opened 2024-06-29
community: Added todoist toolkit

MbBrainz:
This PR adds the the todoist toolkit to the community package. Todoist is an app for managing todos
 **Dependencies:** This pr includes an optional dependency `todoist_api_python`
 
 My twitter handle is @MbBrainz

### DIFF
--- a/libs/community/langchain_community/tools/todoist/tool.py
+++ b/libs/community/langchain_community/tools/todoist/tool.py
@@ -1,0 +1,53 @@
+from typing import Any
+
+from langchain_core.tools import BaseTool
+
+from langchain_community.utilities.todoist import TodoistAPIWrapper
+
+
+class TodoistAction(BaseTool):
+    def __init__(
+        self, name: str, description: str, mode: str, api_wrapper: TodoistAPIWrapper
+    ):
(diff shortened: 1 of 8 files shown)


libs/community/langchain_community/tools/todoist/tool.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TodoistAction(BaseTool):
+    """A tool for interacting with Todoist."""
+
+    mode: str
+    name: str = "todoist"
+    description: str = "A wrapper around Todoist API."
+    api_wrapper: TodoistAPIWrapper = Field(default_factory=TodoistAPIWrapper)
+
+    def _run(self, *args, **kwargs) -> Any:
```
baskaryan, line 17: i think we probably want to have separate Tool classes for each mode and explicitly enumerate the arguments to that API endpoint. this way those args are passed as part of the tool schema when using tool-calling apis

MbBrainz: Do you have an example that i can use? There doesnt seem to be a standard on how to define these tools

PR #23692, opened 2024-06-30
openai[patch]: add vllm to openai

Samoed:
**Description:** Add VLLMOpenAI compatible client with additional parameters


Samoed:
@ccurme

Samoed:
@efriis

Samoed:
@efriis

efriis:
Howdy! This is not the pattern we want followed with the openai partner package. Feel free to contribute new features to the VLLM connector in the community package if that's what is desired!

Samoed:
Hi! Yes, that's what I intended, but this client would require langchain_openai. I could add this dependency for the langchain community or copy-paste the OpenAI client, but I don't think either solution is ideal.

Issue #23585, opened 2024-06-27
SQL Agent extracts the table name with \n linebreaker and next line word 'Observation'


kbatsuren:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain_community.agent_toolkits import create_sql_agent
from langchain.agents.agent_toolkits import SQLDatabaseToolkit

llm = HuggingFaceEndpoint(
    endpoint_url="endpoint_url",
    max_new_tokens=512,
    top_k=10,
    top_p=0.95,
    typical_p=0.95,
    temperature=0.01,
    repetition_penalty=1.03,
)

db = SQLDatabase.from_uri("sqlite:///Chinook.db?isolation_level=IMMEDIATE")

toolkit = SQLDatabaseToolkit(db=db,llm=llm)

agent_executor = create_asql_agent(
    llm=llm,
    toolkit=toolkit,
    verbose=True,
    agent_type=AgentType.ZERO_SHOT_REACT_DESCRIPTION
)
agent_executor.invoke(
    "How many genres are there?"
)

### Error Message and Stack Trace (if applicable)

> Entering new SQL Agent Executor chain...
 I need to know the table that contains the genres.
Action: sql_db_list_tables
Action Input: 
ObservationAlbum, Artist, Customer, Employee, Genre, Invoice, InvoiceLine, MediaType, Playlist, PlaylistTrack, Track Now I know the table that contains the genres is Genre.
Action: sql_db_schema
Action Input: Genre
ObservationError: table_names {'Genre\nObservation'} not found in database I made a mistake, I should remove the Observation part.
Action: sql_db_schema
Action Input: Genre
ObservationError: table_names {'Genre\nObservation'} not found in database I made another mistake, I should remove the newline character.
Action: sql_db_schema
Action Input: Genre
ObservationError: table_names {'Genre\nObservation'} not found in database I made another mistake, I should remove the newline character and the Observation part.
Action: sql_db_schema
Action Input: Genre
ObservationError: table_names {'Genre\nObservation'} not found in database I made another mistake, I should remove the newline character and the Observation part and the curly brackets.
Action: sql_db_schema
Action Input: Genre
[...]

PR #23693, opened 2024-06-30
[Commutiy]: Prompts Fixed for ZERO_SHOT_REACT React Agent Type in `create_sql_agent` function

keenborder786:
- **Description:** The correct Prompts for ZERO_SHOT_REACT were not being used in the `create_sql_agent` function.  They were not using the specific `SQL_PREFIX` and `SQL_SUFFIX` prompts if client does not provide any prompts. This is fixed.
- **Issue:** #23585


keenborder786:
@ccurme

libs/community/langchain_community/agent_toolkits/sql/base.py
```diff
@@ -140,8 +141,9 @@ def create_sql_agent(
     toolkit = toolkit or SQLDatabaseToolkit(llm=llm, db=db)  # type: ignore[arg-type]
     agent_type = agent_type or AgentType.ZERO_SHOT_REACT_DESCRIPTION
     tools = toolkit.get_tools() + list(extra_tools)
+    if prefix is None:
+        prefix = SQL_PREFIX
```
ccurme, line 145: why move this out? we should only have a prefix if `prompt is None`, no?

libs/community/langchain_community/agent_toolkits/sql/base.py
```diff
@@ -235,5 +237,6 @@ def create_sql_agent(
         max_iterations=max_iterations,
         max_execution_time=max_execution_time,
         early_stopping_method=early_stopping_method,
+        handle_parsing_errors=True,
```
ccurme, line 240: this will change behavior for users. why add it? is it related to the fix?

libs/community/langchain_community/agent_toolkits/sql/base.py
keenborder786: This was the original  behaviour as shown in the line [here](https://github.com/keenborder786/langchain/blob/0e916d0d552fde282fa3bf51010c47536913bba3/libs/community/langchain_community/agent_toolkits/sql/base.py#L143). If you want I can change it.

libs/community/langchain_community/agent_toolkits/sql/base.py
keenborder786: reverted

libs/community/langchain_community/agent_toolkits/sql/base.py
keenborder786: Oh I see your point. It has been reverted.

keenborder786:
@ccurme

keenborder786:
@ccurme

ccurme (approved):

PR #23704, opened 2024-07-01
partner:langchain_openai Compatible with other llm (eg: deepseek-chat, glm-4) token_usage

omgbbq:
# glm-4 stream response eg:
```json
{"id":"8792764285258351830","created":1719731338,"model":"glm-4","choices":[{"index":0,"finish_reason":"stop","delta":{"role":"assistant","content":""}}],"usage":{"prompt_tokens":24,"completion_tokens":766,"total_tokens":790}}
```
# deepseek-chat stream response eg:
```json
{"choices":[{"delta":{"content":"","role":null},"finish_reason":"stop","index":0,"logprobs":null}],"created":1719831382,"id":"c579d92ecf2fb2451dea4c2d0cada5e1","model":"deepseek-chat","object":"chat.completion.chunk","system_fingerprint":"fp_3e7ab9c066","usage":{"completion_tokens":611,"prompt_tokens":24,"total_tokens":635}}
```

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -489,8 +489,9 @@ class BaseChatOpenAI(BaseChatModel):
             for chunk in response:
                 if not isinstance(chunk, dict):
                     chunk = chunk.model_dump()
-                if len(chunk["choices"]) == 0:
-                    if token_usage := chunk.get("usage"):
+                token_usage = chunk.get("usage")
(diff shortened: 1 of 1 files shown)


efriis:
don't think this logic is right in a way that could create dropped content blocks if both content and usage data come back in a single chunk (at the moment I think it just drops usage data, but not content)

if you fix in a follow-up PR, could you add some unit tests that test the different response type cases (including openai official ones)?

PR #23719, opened 2024-07-01
community: Add ChatVLLMOpenAI with implementation for `bind_tools` and `with_structured_output`

VMinB12:
- **Description:** This PR adds a chat model version for the vLLM integration. We implemented the `bind_tools` and `with_structured_output` function using vLLM's guided generation features. This makes for a very powerful setup where even small local models that were not necessarily fine-tuned for function calling can generate answers with the expected output. Under the hood, vLLM uses `outlines` or `lm-format-enforcer` to discard any tokens that do not comply with the provided schema. Thus, you are guaranteed to get an answer that complies with your specified structure.
- **Twitter handle:** @b12_consulting, @Vincent_Min


libs/community/langchain_community/chat_models/vllm.py
```diff
@@ -0,0 +1,255 @@
...
+            return "\n" + REGEX_FORMAT_INSTRUCTIONS.format(schema=schema_str)
+        elif guided_mode == "guided_choice":
+            return "\n" + CHOICE_FORMAT_INSTRUCTIONS.format(schema=schema_str)
+        elif guided_mode == "guided_grammar":
+            return "\n" + GRAMMAR_FORMAT_INSTRUCTIONS.format(schema=schema_str)
+        else:
+            raise ValueError(f"Unsupported guided_mode {guided_mode}")
+
+    @staticmethod
+    def _insert_instructions(
```
VMinB12, line 119: I would appreciate some input on how to improve this function.
I tried to handle a couple of possible `LanguageModelInput`s, but I don't expect this to be exhaustive.
Is there a better way to append instructions to an incoming `LanguageModelInput`?

VMinB12: Note that without this insertion of instructions, the `with_structured_output` function will not behave as expected. It really is crucial for the model to understand what output is expected. If the `insert_instructions` step would be removed, the output would still follow the provided schema (because vLLM will force the output to be compliant), but the quality of the generation would likely be very low.

VMinB12:
@hwchase17 @eyurtsev I would love your input on this PR. We've been using a version of this code in our company for a while and it has unlocked great potential for local models and are excited to see what the community can do with it. The open source community has long been lacking a high throughput server for local models with support for tools/structured generation. This implementation seems to be the first chat model that ticks all the boxes in [this table](https://python.langchain.com/v0.2/docs/integrations/chat/#advanced-features).

Note that this approach to structured generation works with any vLLM compatible model, even if it was not specifically tuned for tool use. Imagine creating robust agents with <3B parameter models! This approach does mean we have to insert instructions into the prompt. I left a comment in the changes where this happens, because I suspect this part may be improved a bit. Even with the current implementation, I think this is ready for review and look forward to a possible merge.

magaton:
Any update on this?

VMinB12:
@baskaryan @eyurtsev Any input on this PR?

efriis:
It looks like this is adding a chat model to the `langchain-experimental` package which is no longer hosted out of this repo (it's in https://github.com/langchain-ai/langchain-experimental if you still want it off experimental!). At a minimum to reopen this PR against this repo, the integration would ahve to be in community, but I believe there's a different VLLM integration there that would need these functions added - however based on the [status in vllm's docs, that should probably wait for some more features](https://docs.vllm.ai/en/latest/usage/tool_calling.html)

Regardless I would recommend instead get an integration package out! Future PRs against langchain would just be {docs updates, as well as registering your package in `libs/packages.yml`, deprecating this community integration in favor of your integration package}

VMinB12:
@efriis Thanks for catching that the changes now try to commit to the removed `langchain-experimental` dir. My intention was to point to `langchain-community`, but this was changed by @baskaryan in his latest commits.

I agree that a vllm integration package would be ideal. I'm not part of the vllm team, but it would be great if they would take ownership of this. @WoosukKwon, @mgoin, you are contributors to vllm if I got it right; is there any interest in this from the vLLM team?

I am fine with closing this branch for now. Now that more and more open source model with native "tool use" support are coming out, this branch (which is forcing models that are not necessarily trained for "tool use", to use tools by discarding invalid tokens) is less relevant today than it was when I created it. Let's wait for the vLLM team to add full support for tool use.

PR #23731, opened 2024-07-01
community[patch]: Update root_validators to use pre=True or pre=False

eyurtsev:
Update root_validators in preparation for pydantic 2 migration.


eyurtsev:

PR #23737, opened 2024-07-01
community[patch]: Update @root_validators to use explicit pre=True or pre=False

eyurtsev:
Issue: https://github.com/langchain-ai/langchain/issues/22819


eyurtsev:

PR #23748, opened 2024-07-02
libs: (partners/nexa-ai) integrate nexa-ai agents into langchain

zhiyuan8:
Twitter handle : @nexa4ai (https://x.com/nexa4ai)
Nexa AI : https://www.nexa4ai.com/


libs/partners/nexa-ai/tests/__init__.py
efriis, line 1: let's remove this file, revoke this api key, and store it in a secret instead!

efriis (commented):
Based on the documentation page, would it make more sense to implement this as a chat model, such that it supports tool calling with the standard interface? https://python.langchain.com/v0.2/docs/how_to/tool_calling/

This would also get structured output for free: https://python.langchain.com/v0.2/docs/how_to/structured_output/

The current implementation *looks* like a tool calling model. Is the issue that it doesn't support messages?

efriis:
converting to draft and can review when ready!

libs/partners/nexa-ai/tests/__init__.py
zhiyuan8: Hi @efriis thanks for the review, it is a fake API key just for the CICD test.

zhiyuan8:
> Based on the documentation page, would it make more sense to implement this as a chat model, such that it supports tool calling with the standard interface? https://python.langchain.com/v0.2/docs/how_to/tool_calling/
> 
> This would also get structured output for free: https://python.langchain.com/v0.2/docs/how_to/structured_output/
> 
> The current implementation _looks_ like a tool calling model. Is the issue that it doesn't support messages?

Hi @efriis , our model is an action model that only supports completion mode, it is not a conversation model. So we have not supported messages yet.

efriis:
@zhiyuan8 is there another model in langchain that has similar functionality? I don't think the current implementation properly implements our abstractions, so it's not quite mergable in its current state.

If it's nonstandard, I'd welcome a different PR that adds a generic runnable that generates actions as a dict in langchain-community! You could add documentation to a nexa provider page, but it wouldn't be documented as an llm or chat model.

efriis:
Closing as stale

As mentioned if the interface doesn't match our base models, would welcome a langchain-community integration

PR #23767, opened 2024-07-02
openai: embeddings: supported chunk_size when check_embedding_ctx_length is disabled

adubovik:
Chunking of the input array controlled by `self.chunk_size` is being ignored when `self.check_embedding_ctx_length` is disabled. Effectively, the chunk size is assumed to be equal 1 in such a case. This is suprising.

The PR takes into account `self.chunk_size` passed by the user.


adubovik:
@efriis is there something I could do to impove the PR? Thanks

efriis:
reassigning to @baskaryan who recently refactored this!

adubovik:
Hi, @baskaryan! Do you have any suggestions?

efriis (approved):

PR #23774, opened 2024-07-02
core[minor]: add upsert, streaming_upsert, aupsert, astreaming_upsert methods to the VectorStore abstraction

eyurtsev:
This PR rolls out part of the new proposed interface for vectorstores (https://github.com/langchain-ai/langchain/pull/23544) to existing store implementations.

The PR makes the following changes:

1. Adds standard upsert, streaming_upsert, aupsert, astreaming_upsert methods to the vectorstore.
2. Updates `add_texts` and `aadd_texts` to be non required with a default implementation that delegates to `upsert` and `aupsert` if those have been implemented.  The original `add_texts`  and `aadd_texts` methods are problematic as they spread object specific information across document and **kwargs. (e.g., ids are not a part of the document)
3. Adds a default implementation to `add_documents` and `aadd_documents` that delegates to `upsert` and `aupsert` respectively.
4. Adds standard unit tests to verify that a given vectorstore implements a correct read/write API.

A downside of this implementation is that it creates `upsert` with a very similar signature to `add_documents`. 
The reason for introducing `upsert` is to:
* Remove any ambiguities about what information is allowed in `kwargs`. Specifically kwargs should only be used for information common to all indexed data. (e.g., indexing timeout). 
*Allow inheriting from an anticipated generalized interface for indexing that will allow indexing `BaseMedia` (i.e., allow making a vectorstore for images/audio etc.)
 
`add_documents` can be deprecated in the future in favor of `upsert` to make sure that users have a single correct way of indexing content.

### DIFF
--- a/libs/core/langchain_core/indexing/api.py
+++ b/libs/core/langchain_core/indexing/api.py
@@ -405,7 +405,7 @@ async def _to_async_iterator(iterator: Iterable[T]) -> AsyncIterator[T]:
 
 
 async def aindex(
-    docs_source: Union[BaseLoader, Iterable[Document], AsyncIterable[Document]],
+    docs_source: Union[BaseLoader, Iterable[Document], AsyncIterator[Document]],
     record_manager: RecordManager,
     vector_store: VectorStore,
     *,
--- a/libs/core/langchain_core/indexing/api.py
+++ b/libs/core/langchain_core/indexing/api.py
@@ -405,7 +405,7 @@ async def _to_async_iterator(iterator: Iterable[T]) -> AsyncIterator[T]:
 
 
 async def aindex(
-    docs_source: Union[BaseLoader, Iterable[Document], AsyncIterator[Document]],
+    docs_source: Union[BaseLoader, Iterable[Document], AsyncIterable[Document]],
     record_manager: RecordManager,
     vector_store: VectorStore,
     *,
--- a/libs/core/langchain_core/indexing/base.py
+++ b/libs/core/langchain_core/indexing/base.py
@@ -317,3 +317,16 @@ class InMemoryRecordManager(RecordManager):
 
     async def adelete_keys(self, keys: Sequence[str]) -> None:
         self.delete_keys(keys)
+
+
+class UpsertResponse(TypedDict):
+    """A generic response for upsert operations.
+
+    The upsert response will be used by abstractions that implement an upsert
+    operation for content that can be upserted by ID.
+    """
+
+    succeeded: List[str]
+    """The IDs that were successfully indexed."""
+    failed: List[str]
+    """The IDs that failed to index."""
--- a/libs/core/langchain_core/vectorstores.py
+++ b/libs/core/langchain_core/vectorstores.py
@@ -29,6 +29,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     AsyncIterable,
+    AsyncIterator,
     Callable,
     ClassVar,
     Collection,
@@ -230,7 +231,7 @@ class VectorStore(ABC):
         /,
         batch_size: int,
         **kwargs: Any,
-    ) -> Iterator[UpsertResponse]:
+    ) -> AsyncIterator[UpsertResponse]:
         """Upsert documents in a streaming fashion. Async version of streaming_upsert.
 
         Args:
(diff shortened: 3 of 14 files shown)


libs/standard-tests/langchain_standard_tests/integration_tests/vectorstores.py
```diff
@@ -250,25 +250,8 @@ async def test_add_documents_with_ids_is_idempotent(
         await vectorstore.aadd_documents(documents, ids=["1", "2"])
         documents = await vectorstore.asimilarity_search("bar", k=2)
         assert documents == [
-            Document(page_content="bar", metadata={"id": 2}),
-            Document(page_content="foo", metadata={"id": 1}),
-        ]
-
-    async def test_add_documents_without_ids_gets_duplicated(
```
eyurtsev, line 257: Removed this unit test to allow vectorstores to implement assignment of IDs by document hash

libs/core/langchain_core/indexing/api.py
```diff
@@ -405,7 +405,7 @@ async def _to_async_iterator(iterator: Iterable[T]) -> AsyncIterator[T]:
 
 
 async def aindex(
-    docs_source: Union[BaseLoader, Iterable[Document], AsyncIterator[Document]],
+    docs_source: Union[BaseLoader, Iterable[Document], AsyncIterable[Document]],
```
eyurtsev, line 408:
```suggestion
    docs_source: Union[BaseLoader, Iterable[Document], AsyncIterator[Document]],
```

libs/core/langchain_core/indexing/base.py
```diff
@@ -317,3 +317,16 @@ def delete_keys(self, keys: Sequence[str]) -> None:
...
+class UpsertResponse(TypedDict):
+    """A generic response for upsert operations.
+
+    The upsert response will be used by abstractions that implement an upsert
+    operation for content that can be upserted by ID.
+    """
+
+    succeeded: List[str]
+    """The IDs that were successfully indexed."""
+    failed: List[str]
```
ccurme, line 331: (nit) are error messages useful to capture here

libs/core/langchain_core/vectorstores.py
```diff
@@ -73,16 +85,205 @@ def add_texts(
...
+
+        .. versionadded:: 0.2.11
+        """
+        #  Developer guidelines:
+        #
+        #  Vectorstores implementations are free to extend `upsert` implementation
+        #  to take in additional data per document.
+        #
+        #  This data **SHOULD NOT** be part of the **kwargs** parameter, instead
+        #  a sub-classes can use a Union type on `documents` to include additional
```
ccurme, line 194:
```suggestion
        #  sub-classes can use a Union type on `documents` to include additional
```

libs/core/langchain_core/vectorstores.py
```diff
@@ -186,17 +387,81 @@ async def aadd_texts(
...
             List of IDs of the added texts.
         """
+        if type(self).upsert != VectorStore.upsert:
+            from langchain_core.documents import Document
+
+            if "ids" in kwargs:
+                ids = kwargs.pop("ids")
+                if ids and len(ids) != len(documents):
+                    raise ValueError(
+                        "The number of ids must match the number of documents."
```
ccurme, line 445:
```suggestion
                        "The number of ids must match the number of documents. "
```

libs/core/langchain_core/vectorstores.py
```diff
@@ -213,6 +478,38 @@ async def aadd_documents(
...
+                    doc_with_id = Document(
+                        page_content=document.page_content,
+                        metadata=document.metadata,
+                        id=id_,
+                    )
+                    documents_.append(doc_with_id)
+            else:
+                documents_ = documents
+
+            # If upsert has been implemented, we can use it to add documents
```
ccurme, line 509: (nit) this comment is a little confusing as we only call `self.aupsert` (understood that `.aupsert` by default will use `.upsert`).

ccurme (approved):
just nits

libs/core/langchain_core/indexing/base.py
eyurtsev: No, just the IDs right now. We can expand later with a NotRequired key for error message? I'm not sure what useful just yet -- (an alternative is standardizing on exceptions)

libs/core/langchain_core/vectorstores.py
eyurtsev: Good point, I'll follow up with an update of the doc-string

PR #23787, opened 2024-07-02
core: docstrings `load`

leo-gan:

### DIFF
--- a/libs/core/langchain_core/load/dump.py
+++ b/libs/core/langchain_core/load/dump.py
@@ -12,11 +12,7 @@ def default(obj: Any) -> Any:
         obj: The object to serialize to json if it is a Serializable object.
 
     Returns:
-        A json serializable object.
-
-    Raises:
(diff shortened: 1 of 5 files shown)


libs/core/langchain_core/load/dump.py
```diff
@@ -6,7 +6,18 @@
...
+    a SerializedNotImplemented object.
+
+    Args:
+        obj: The object to serialize to json if it is a Serializable object.
+
+    Returns:
+        A json serializable object.
+
+    Raises:
+        ValueError: If the object is not a Serializable object or
```
baskaryan, line 18: i actually dont think this raises any errors on a not Serializable object

leo-gan: Thanks! Good catch.

eyurtsev (approved):

leo-gan:
@eyurtsev Thanks!

PR #23806, opened 2024-07-03
core: Accept configurable keys top-level

nfcampos:


baskaryan (approved):
do we think it'd be helpful to log/warn something in with_configurable_fields/alternatives if you try to make a configurable key with same name as one of the main keys

PR #23818, opened 2024-07-03
partners[patch]: MongoDB vectorstore to return and accept string IDs

pprados:
@eyurtsev

The mongdb have some errors.
- `add_texts() -> List` returns a list of `ObjectId`, and not a list of string
- `delete()` with `id` never remove chunks.

### DIFF
--- a/libs/core/langchain_core/output_parsers/format_instructions.py
+++ b/libs/core/langchain_core/output_parsers/format_instructions.py
@@ -8,4 +8,6 @@ the object {{"foo": ["bar", "baz"]}} is a well-formatted instance of the schema.
 Here is the output schema:
 ```
 {schema}
-```"""
+```
+Produce only the json result, without the schema.
+"""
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/output_parsers/format_instructions.py
```diff
@@ -8,4 +8,6 @@
 Here is the output schema:
 ```
 {schema}
-```"""
+```
```
eyurtsev, line 11:
```suggestion
```"""
```

libs/core/langchain_core/output_parsers/format_instructions.py
```diff
@@ -8,4 +8,6 @@
...
+Produce only the json result, without the schema.
```
eyurtsev, line 12:
```suggestion
```

libs/core/langchain_core/output_parsers/format_instructions.py
```diff
@@ -8,4 +8,6 @@
...
+"""
```
eyurtsev, line 13:
```suggestion
```

eyurtsev (approved):
Looks good, but let's separate into different PRs since the changes are very different conceptually.

FWIW I don't think we'll merge the change in format_instructions in the parser

eyurtsev (approved):

PR #23825, opened 2024-07-03
core[patch]: docstrings `output_parsers`

leo-gan:

### DIFF
--- a/libs/core/langchain_core/output_parsers/openai_tools.py
+++ b/libs/core/langchain_core/output_parsers/openai_tools.py
@@ -18,7 +18,21 @@ def parse_tool_call(
     strict: bool = False,
     return_id: bool = True,
 ) -> Optional[Dict[str, Any]]:
-    """Parse a single tool call."""
+    """Parse a single tool call.
+
+    Args:
+        raw_tool_call: The raw tool call to parse.
+        partial: Whether to parse partial JSON. Default is False.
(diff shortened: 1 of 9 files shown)


libs/core/langchain_core/output_parsers/openai_tools.py
```diff
@@ -52,7 +66,15 @@ def make_invalid_tool_call(
     raw_tool_call: Dict[str, Any],
     error_msg: Optional[str],
 ) -> InvalidToolCall:
-    """Create an InvalidToolCall from a raw tool call."""
+    """Create an InvalidToolCall from a raw tool call.
+
+    Args:
+        raw_tool_call: The raw tool call.
```
eyurtsev, line 72: Some of the doc-strings are not providing any information that's not already present in the function signature. Does this help to improve the API reference by naming arguments?

eyurtsev (approved):

libs/core/langchain_core/output_parsers/openai_tools.py
leo-gan: IMHO It is kind of the "maturity" index. If I see this, it it 100% missed content. This content can be 100% trivial and uninformative. 
We have a list of Args, dashes that tell reader "here is a place for description", and this missed description.
image

leo-gan:
@eyurtsev Thanks!

PR #23841, opened 2024-07-03
core[minor],community[minor]: Upgrade all @root_validator() to @pre_init

eyurtsev:
* Introduce a custom pre_init initializer
* Attempt migrating all community code to it


libs/core/langchain_core/utils/pydantic.py
```diff
@@ -12,3 +17,35 @@ def get_pydantic_major_version() -> int:
 
 
 PYDANTIC_MAJOR_VERSION = get_pydantic_major_version()
+
+
+# How to type hint this?
+def pre_init(func: Callable) -> Any:
+    """Decorator to run a function before model initialization."""
+
+    @root_validator(pre=True)
```
efriis, line 26: if this changes everything to pre=True, won't this create problems with values[x] not existing, where before it would?

efriis: nvm I believe you

efriis (approved):

libs/community/langchain_community/embeddings/yandex.py
```diff
@@ -76,7 +76,7 @@ class Config:
 
         allow_population_by_field_name = True
 
-    @root_validator()
```
mkhludnev, line 79: it causes #25575, which is fixed by #25617 
@eyurtsev could you please take a look?

Issue #23824, opened 2024-07-03
generate/predict fails when using sql cache


ziqizhang:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
import langchain
from langchain_openai import ChatOpenAI
from langchain.chains.conversation.base import ConversationChain
from langchain.memory import ConversationBufferMemory
from langchain_community.cache import SQLiteCache
from langchain_community.callbacks import get_openai_callback
import tiktoken,sys
from langchain_core.globals import set_llm_cache, get_llm_cache

cache = SQLiteCache(database_path=".langchain.db")
llm = ChatOpenAI(model_name='gpt-3.5-turbo',
                      openai_api_key=sys.argv[1])
memory = ConversationBufferMemory()
conversation = ConversationChain(
    llm=llm,
    memory = memory,
    verbose=True
)
with get_openai_callback() as cb:
    input="Hi, my name is Andrew"
    tokenizer=tiktoken.get_encoding("cl100k_base")
    toks=len(tokenizer.encode(text=input))
    print(toks)

    costs={"tokens used":0, "prompt tokens":0, "completion tokens":0, "successful requests":0, "cost (usd)":0}
    result = conversation.predict(input=input)
    costs['tokens used']=cb.total_tokens
    costs['prompt tokens'] = cb.prompt_tokens
    costs['completion tokens'] = cb.completion_tokens
    costs['successful requests'] = cb.successful_requests
    costs['cost (used)'] = cb.total_cost
    print(result)
    print(costs)

    res=conversation.predict(input="What is 1+1?")
    costs['tokens used'] = cb.total_tokens
    costs['prompt tokens'] = cb.prompt_tokens
    costs['completion tokens'] = cb.completion_tokens
    costs['successful requests'] = cb.successful_requests
    costs['cost (used)'] = cb.total_cost
    print(res)
    print(costs)

    res=conversation.predict(input="What is my name?")
    costs['tokens used'] = cb.total_tokens
    costs['prompt tokens'] = cb.prompt_tokens
    costs['completion tokens'] = cb.completion_tokens
    costs['successful requests'] = cb.successful_requests
    costs['cost (used)'] = cb.total_cost
[...]

PR #23842, opened 2024-07-03
core[patch]: Fix logic in BaseChatModel that processes the llm string that is used as a key for caching chat models responses

eyurtsev:
This PR should fix the following issue: https://github.com/langchain-ai/langchain/issues/23824
Introduced as part of this PR: https://github.com/langchain-ai/langchain/pull/23416

I am unable to reproduce the issue locally though it's clear that we're getting a `serialized` object which is not a dictionary somehow.

The test below passes for me prior to the PR as well

```python

def test_cache_with_sqllite() -> None:
    from langchain_community.cache import SQLiteCache

    from langchain_core.globals import set_llm_cache

    cache = SQLiteCache(database_path=".langchain.db")
    set_llm_cache(cache)
    chat_model = FakeListChatModel(responses=["hello", "goodbye"], cache=True)
    assert chat_model.invoke("How are you?").content == "hello"
    assert chat_model.invoke("How are you?").content == "hello"
```

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -1258,8 +1258,13 @@ def _cleanup_llm_representation(serialized: Any, depth: int) -> None:
     """Remove non-serializable objects from a serialized object."""
     if depth > 100:  # Don't cooperate for pathological cases
         return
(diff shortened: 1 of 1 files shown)


coolbeevip:
Look forward to being able to quickly release a version.

PR #23852, opened 2024-07-04
[HuggingFace Pipeline] add streaming support

openvino-dev-samples:


OpenVINO-dev-contest:
@baskaryan @ccurme could you help to review, thanks

Jofthomas:
Hey @OpenVINO-dev-contest, thanks for making this change.

It would be better to implement the stops with `stop_strings ` but while they work in model.generate(), this new parameter does not seem to be correctly passed to the pipeline. Hence, The StoppingCriteriaList works for me at this moment, but should be changed later on ( I'll dig transformers with the team ).

At first glance, PR looks good to me.

ccurme (commented):
Hi @OpenVINO-dev-contest, thanks for this. Could you please add tests to the `partners/huggingface/` implementation?

OpenVINO-dev-contest:
> Hi @OpenVINO-dev-contest, thanks for this. Could you please add tests to the `partners/huggingface/` implementation?

done

ccurme (approved):

Issue #23797, opened 2024-07-03
partner[chroma]: Not able to load persisted collection without calling `from_documents`  


spike-spiegel-21:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

If I create a vectorstore with:
```python
from langchain_community.embeddings import FakeEmbeddings
from langchain_chroma.vectorstores import Chroma
vectorstore = Chroma.from_documents(documents=splits, embedding = FakeEmbeddings(size=1352), collection_name="colm", persist_directory="my_dir")
```
The only way to retrieve the persisted collection from `my_dir` is:
```python
vectorstore = Chroma.from_documents(documents=splits, embedding = FakeEmbeddings(size=1352), collection_name="colm")
```
OR

```python
vectorstore = Chroma.from_text(text=..., embedding = FakeEmbeddings(size=1352), collection_name="colm")
```
Related issues:
https://github.com/langchain-ai/langchain/issues/22361
https://github.com/langchain-ai/langchain/issues/20866
https://github.com/langchain-ai/langchain/issues/19807#issuecomment-2028610882

### Description

I want dedicated `classmethod` just like this:
https://github.com/langchain-ai/langchain/blob/27aa4d38bf93f3eef7c46f65cc0d0ef3681137eb/libs/partners/qdrant/langchain_qdrant/vectorstores.py#L1351
that returns me an instance of `Chroma` without inserting any texts

### System Info

langchain-chroma                         0.1.1

RafaelXokito:
From my understanding, you are trying to reload a persisted Chroma collection without the need of having the documents.
If so, you can get a VectorStore instance from the persisted directory like this:
```python
vectorstore = Chroma(
        collection_name="colm",
        embedding_function=FakeEmbeddings(),
        persist_directory="my_dir",
    )
```

Take into account that all these attributes are optional.

spike-spiegel-21:
> From my understanding, you are trying to reload a persisted Chroma collection without the need of having the documents. If so, you can get a VectorStore instance from the persisted directory like this:
> 
> ```python
> vectorstore = Chroma(
>         collection_name="colm",
>         embedding_function=FakeEmbeddings(),
[...]

PR #23854, opened 2024-07-04
enhancement[chroma]: Added a function `from_existing_collection`.

spike-spiegel-21:
Resolves this: #23797

### DIFF
--- a/libs/partners/chroma/tests/integration_tests/test_vectorstores.py
+++ b/libs/partners/chroma/tests/integration_tests/test_vectorstores.py
@@ -528,3 +528,30 @@ def test_reset_collection(client: chromadb.ClientAPI) -> None:
     assert vectorstore._collection.count() == 0
     # Clean up
     vectorstore.delete_collection()
+
+
+def test_from_existing_collection() -> None:
+    """Tests initializing a Chroma vectorstore from an existing collection."""
+    chroma_persist_dir = "./tests/persist_dir"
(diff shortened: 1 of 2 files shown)


RafaelXokito:
I have reviewed the `from_existing_collection` method, and the method only calls the class constructor with the same parameters. This raises a concern about potential redundancy. Users can achieve the same result by directly calling the Chroma constructor with the appropriate arguments (As shown in the issue you raised).

libs/partners/chroma/tests/integration_tests/test_vectorstores.py
```diff
@@ -528,3 +528,30 @@ def test_reset_collection(client: chromadb.ClientAPI) -> None:
     assert vectorstore._collection.count() == 0
     # Clean up
     vectorstore.delete_collection()
+
+
+def test_from_existing_collection() -> None:
```
RafaelXokito, line 533: Duplicated of `test_chroma_with_persistence` test

spike-spiegel-21:
> I have reviewed the `from_existing_collection` method, and the method only calls the class constructor with the same parameters. This raises a concern about potential redundancy. Users can achieve the same result by directly calling the Chroma constructor with the appropriate arguments (As shown in the issue you raised).

Makes sense @RafaelXokito

spike-spiegel-21:
I am closing this PR. Thank you @RafaelXokito for your feedback.

Issue #23863, opened 2024-07-04
Community: Keyerror `model_name` for GPT4AllEmbeddings


Dev-Khant:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain_community.embeddings import GPT4AllEmbeddings

embeddings = GPT4AllEmbeddings()

### Error Message and Stack Trace (if applicable)

KeyError                                  Traceback (most recent call last)
[<ipython-input-6-542601a9aeec>](https://localhost:8080/#) in <cell line: 1>()
----> 1 x = GPT4AllEmbeddings()

2 frames
[/usr/local/lib/python3.10/dist-packages/langchain_community/embeddings/gpt4all.py](https://localhost:8080/#) in validate_environment(cls, values)
     37 
     38             values["client"] = Embed4All(
---> 39                 model_name=values["model_name"],
     40                 n_threads=values.get("n_threads"),
     41                 device=values.get("device"),

KeyError: 'model_name'

### Description

It can not find `model_name` from values.

### System Info

langchain==0.2.6
langchain-community==0.2.6
langchain-core==0.2.11
langchain-text-splitters==0.2.2

wenngong:
@Dev-Khant you should specific model_name param like following:

from langchain_community.embeddings import GPT4AllEmbeddings

model_name = "all-MiniLM-L6-v2.gguf2.f16.gguf"
gpt4all_kwargs = {'allow_download': 'True'}
embeddings = GPT4AllEmbeddings(
    model_name=model_name,
    gpt4all_kwargs=gpt4all_kwargs
)


And I will add model_name param validate in GPT4AllEmbeddings

wenngong:
PR for this issue：#23865

wenngong:
> PR for this issue：#23867

@Dev-Khant Please close this issue if this pr solved it.

PR #23867, opened 2024-07-04
community: add model_name param valid for GPT4AllEmbeddings

Zizo-Vi:
Description: add model_name param valid for GPT4AllEmbeddings

Issue: #23863 #22819


RafaelXokito (commented):
From my understanding, you are trying to force the `model_name` usage, which I particularly don't understand why.
Your code works. However I noticed that your changes will affect the GPT4AllEmbeddings [documentation](https://python.langchain.com/v0.2/docs/integrations/text_embedding/gpt4all/) description. You might consider update that as well if you still want to apply this changes.
Do  we really need this change?

wenngong:
> From my understanding, you are trying to force the `model_name` usage, which I particularly don't understand why.
> Your code works. However I noticed that your changes will affect the GPT4AllEmbeddings [documentation](https://python.langchain.com/v0.2/docs/integrations/text_embedding/gpt4all/) description. You might consider update that as well if you still want to apply this changes.
> Do we really need this change?

Tow ways to fix this original issue:
1. Set default value for model_name param 
2. Valid model_name param and assert not none.


I prefer the second one. As u pointed out, any way need to change the doc description.

wenngong:
> From my understanding, you are trying to force the `model_name` usage, which I particularly don't understand why.
> Your code works. However I noticed that your changes will affect the GPT4AllEmbeddings [documentation](https://python.langchain.com/v0.2/docs/integrations/text_embedding/gpt4all/) description. You might consider update that as well if you still want to apply this changes.
> Do we really need this change?

Tow ways to fix this original issue:
1. Set default value for model_name param 
2. Valid model_name param and assert not none.


I prefer the second one. As u pointed out, any way need to change the doc description. I need submit another pr for it.

RafaelXokito:
> Tow ways to fix this original issue:
> 
> 1. Set default value for model_name param
> 2. Valid model_name param and assert not none.
> 
> I prefer the second one. As u pointed out, any way need to change the doc description. I need submit another pr for it.

Sorry @wenngong I really don't believe the solution for this is making the `model_name` field mandatory at the GPT4AllEmbeddings level because this will make Embed4All `model_name` validation obsolete. In my opinion, the best option to solve this problem is the following:

```python
    model_name: Optional[str] = None
    n_threads: Optional[int] = None
    device: Optional[str] = "cpu"
    gpt4all_kwargs: Optional[dict] = {}
    client: Any  #: :meta private:

    @root_validator()
        """Validate that GPT4All library is installed."""
        try:
            from gpt4all import Embed4All

            values["client"] = Embed4All(
                model_name=values.get("model_name"),
```

This includes making the `model_name` field optional as it should, with `None` as default, because the `None` option will be ensured in Embed4All class initialization. Take into account that I replaced `values["model_name"]` with `values.get("model_name")` to make the code more coherent.

wenngong:
> > Tow ways to fix this original issue:
> > 
> > 1. Set default value for model_name param
> > 2. Valid model_name param and assert not none.
> > 
> > I prefer the second one. As u pointed out, any way need to change the doc description. I need submit another pr for it.
> 
> Sorry @wenngong I really don't believe the solution for this is making the `model_name` field mandatory at the GPT4AllEmbeddings level because this will make Embed4All `model_name` validation obsolete. In my opinion, the best option to solve this problem is the following:
> 
> ```python
>     model_name: Optional[str] = None
>     n_threads: Optional[int] = None
>     device: Optional[str] = "cpu"
>     gpt4all_kwargs: Optional[dict] = {}
>     client: Any  #: :meta private:
> 
>     @root_validator()
>     def validate_environment(cls, values: Dict) -> Dict:
>         """Validate that GPT4All library is installed."""
>         try:
>             from gpt4all import Embed4All
> 
>             values["client"] = Embed4All(
>                 model_name=values.get("model_name"),
> ```
> 
> This includes making the `model_name` field optional as it should, with `None` as default, because the `None` option will be ensured in Embed4All class initialization. Take into account that I replaced `values["model_name"]` with `values.get("model_name")` to make the code more coherent.
> 
> With this approach there's no need to change the documentation.

@RafaelXokito Okay, Agree! Your solution is better one. I will use it. U are a good reviewer. Thanks for pointing out.

libs/community/langchain_community/embeddings/gpt4all.py
```diff
@@ -22,21 +22,20 @@ class GPT4AllEmbeddings(BaseModel, Embeddings):
...
 
-    model_name: str
+    model_name: Optional[str] = None
     n_threads: Optional[int] = None
     device: Optional[str] = "cpu"
     gpt4all_kwargs: Optional[dict] = {}
     client: Any  #: :meta private:
 
-    @root_validator()
+    @root_validator(pre=True)
```
RafaelXokito, line 31: From my understanding there's no need to run the validator before the standard validators since no error will be raised from the fields.

```suggestion
    @root_validator()
```

Give me your feedback on this.

wenngong: this is for this issue : #22819 pydantic 2 migration by refactoring vanilla @root_validator() usage

RafaelXokito: Thank you for your feedback and explanation regarding the Pydantic 2 migration.
To maintain the previous behavior, wouldn't it be better to use `@root_validator(pre=False)`?

RafaelXokito (approved):

libs/community/langchain_community/embeddings/gpt4all.py
wenngong: @RafaelXokito as the following rules:
1. Pre-initialization is most useful for creating defaults for values, especially when the defaults cannot be supplied per field individually.
2. Post-initialization is most useful for doing more complex validation, especially one that involves multiple fields.

And validatee_environment func set values[“client”], so should choose pre-init.

RafaelXokito: Thank you for the explanation.

wenngong:
@RafaelXokito @ccurme please merge, thx

RafaelXokito (approved):

RafaelXokito (approved):

ccurme (approved):

PR #23873, opened 2024-07-04
community[patch]: Redis.delete should be a regular method not a static method

pprados:
The `langchain_common.vectostore.Redis.delete()` must not be a `@staticmethod`.

With the current implementation, it's not possible to have multiple instances of Redis vectorstore because all versions must share the `REDIS_URL`.

It's not conform with the base class.


eyurtsev (approved):

eyurtsev:
This is technically a breaking change, but I think we should fix this

PR #23876, opened 2024-07-04
community[minor]: add mongodb byte store

pprados:
The `MongoDBStore` can manage only documents.
It's not possible to use MongoDB for an `CacheBackedEmbeddings`.

With this new implementation, it's possible to use:
```python
CacheBackedEmbeddings.from_bytes_store(
    underlying_embeddings=embeddings,
    document_embedding_cache=MongoDBByteStore(
      connection_string=db_uri,
      db_name=db_name,
      collection_name=collection_name,
  ),
)
```
and use MongoDB to cache the embeddings !


eyurtsev:
@pprados could you add an integration test using `langchain_standard_tests`

Look at the following unit tests as examples:

https://github.com/langchain-ai/langchain/blob/d311f2218280d90bbbc265050e4029eefa288f23/libs/standard-tests/tests/unit_tests/test_in_memory_base_store.py#L16-L16

```
class TestInMemoryStore(BaseStoreSyncTests):
    @pytest.fixture
    def three_values(self) -> Tuple[bytes, bytes, bytes]:  # <-- Provide 3 
        return b"foo", b"bar", b"buzz"

    @pytest.fixture
    def kv_store(self) -> Store:
       yield implementation using mongodb that has no keys stored in it

```

pprados:
@eyurtsev 
The new version use the generic test.

eyurtsev (approved):

PR #23878, opened 2024-07-04
community[patch]: Improve ChatBaichuan init args and role

maang-h:
- **Description:** Improve ChatBaichuan init args and role
   -  ChatBaichuan adds `system` role
   - alias: `baichuan_api_base` -> `base_url`
   - `with_search_enhance` is deprecated
   - Add `max_tokens` argument

### DIFF
--- a/libs/community/langchain_community/chat_models/baichuan.py
+++ b/libs/community/langchain_community/chat_models/baichuan.py
@@ -22,6 +22,8 @@ from langchain_core.messages import (
     ChatMessageChunk,
     HumanMessage,
     HumanMessageChunk,
+    SystemMessage,
+    SystemMessageChunk,
 )
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
 from langchain_core.pydantic_v1 import Field, SecretStr, root_validator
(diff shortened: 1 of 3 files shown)


libs/community/langchain_community/chat_models/baichuan.py
```diff
@@ -113,14 +121,16 @@ def lc_secrets(self) -> Dict[str, str]:
...
-    baichuan_api_base: str = Field(default=DEFAULT_API_BASE)
+    baichuan_api_base: str = Field(default=DEFAULT_API_BASE, alias="base_url")
     """Baichuan custom endpoints"""
     baichuan_api_key: SecretStr = Field(alias="api_key")
     """Baichuan API Key"""
     baichuan_secret_key: Optional[SecretStr] = None
     """[DEPRECATED, keeping it for for backward compatibility] Baichuan Secret Key"""
     streaming: bool = False
     """Whether to stream the results or not."""
+    max_tokens: int = 2048
```
ccurme, line 132: would this introduce a new token limit by default?

maang-h: Yes, the default value for max tokens introduced in the latest documentation is 2048.

ccurme: What is the benefit of maintaining that default here? If it changes Langchain will have an incorrect default. Does it make sense for this to be `Optional[int] = None`?

maang-h: Wow! I understand, I will improve it.

ccurme (approved):

PR #23890, opened 2024-07-04
core: docstrings `prompts`

leo-gan:

### DIFF
--- a/libs/core/langchain_core/prompts/base.py
+++ b/libs/core/langchain_core/prompts/base.py
@@ -84,12 +84,14 @@ class BasePromptTemplate(
 
     @classmethod
     def get_lc_namespace(cls) -> List[str]:
-        """Get the namespace of the langchain object."""
+        """Get the namespace of the langchain object.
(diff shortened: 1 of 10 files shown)


libs/core/langchain_core/prompts/base.py
```diff
@@ -84,12 +84,14 @@ def validate_variable_names(cls, values: Dict) -> Dict:
...
-        """Get the namespace of the langchain object."""
+        """Get the namespace of the langchain object.
+        Returns ["langchain", "schema", "prompt_template"]."""
         return ["langchain", "schema", "prompt_template"]
 
     @classmethod
     def is_lc_serializable(cls) -> bool:
-        """Return whether this class is serializable."""
+        """Return whether this class is serializable.
+        Returns True."""
```
eyurtsev, line 94: Should probably be

Returns:
  True


Otherwise I don't think it'll parse correctly

eyurtsev (approved):

leo-gan:
@eyurtsev Thanks!

PR #23909, opened 2024-07-05
core: ensure `run_inline` is honored in AsyncCallback Handler

parambharat:
The `AsyncCallBancManager` that handles callbacks doesn't honor a callback handler's `run_inline` attribute.  This always executes tasks concurrently, which might not be suitable for all scenarios, especially when tasks are interdependent or a handler expects tasks to be completed in order.

This is specifically in the following methods:
[on_llm_start](https://github.com/langchain-ai/langchain/blob/0e916d0d552fde282fa3bf51010c47536913bba3/libs/core/langchain_core/callbacks/manager.py#L1676) and [on_chat_model_start](https://github.com/langchain-ai/langchain/blob/0e916d0d552fde282fa3bf51010c47536913bba3/libs/core/langchain_core/callbacks/manager.py#L1738)

**Current Behavior**

All tasks are executed concurrently using `asyncio.gather`, regardless of the handlers' `run_inline` attribute.

**Expected Behavior**

- If any handler has `run_inline` set to `True`, tasks should be executed sequentially. This ensures that tasks that may have dependencies on the completion of previous tasks are handled correctly, avoiding potential race conditions or logic errors.
- If no handlers require inline execution, tasks should be executed concurrently using `asyncio.gather`. This approach is efficient for independent tasks, reducing overall execution time compared to sequential execution.

**Code Snippet**

The following should fix the execution order and honor the `run_inline` attribute of the handler.

```python
if any([h.run_inline for h in self.handlers]):
    for task in tasks:
        await task
else:
    await asyncio.gather(*tasks)
```

**Rationale**

This change aims to introduce necessary flexibility in task execution, ensuring efficiency and correctness based on the `run_inline` attribute set in the handler involved. Without this conditional execution logic, the system would consistently execute tasks concurrently, leading to potential race conditions among tasks in an async setting.

**Impact**

Implementing this change would prevent race conditions and ordering issues, making the system more robust and reliable. 

**Expected Outcome**

With the proposed change, tasks should be executed in the correct order or concurrently, based on the `run_inline` attribute, preventing any potential race conditions or logic errors.

cc: @baskaryan, @hwchase17


libs/core/langchain_core/callbacks/manager.py
```diff
@@ -1673,7 +1672,11 @@ async def on_llm_start(
                 )
             )
 
-        await asyncio.gather(*tasks)
+        if any([h.run_inline for h in self.handlers]):
```
eyurtsev, line 1675: If a single handler has run_inline set to True, all handlers will end up running inline, which is not the correct behavior

eyurtsev (changes requested):
@parambharat thanks for the PR, i added a comment, would you mind fixing the issue so that the code still works in an optimized fashion and then we can merge?

libs/core/langchain_core/callbacks/manager.py
parambharat: good catch @eyurtsev . I've fixed it. Please let me know how the change looks.

libs/core/langchain_core/callbacks/manager.py
```diff
@@ -1645,9 +1646,9 @@ async def on_llm_start(
             else:
                 run_id_ = uuid.uuid4()
 
-            tasks.append(
-                ahandle_event(
-                    self.handlers,
+            for handler in self.handlers:
```
eyurtsev, line 1649: The implementation of `handle_event` and `ahandle_event` already takes care of inline vs. non_inline tasks. Why is this code needed at the prompt level?

eyurtsev: What immediate problem is this solving?

I'm planning on doing an overhaul of the callback system in the not too distant future, and want to avoid any unnecessary changes unless they're actually blocking an important use case

parambharat: I'm working on a WeaveTracer for Langchain [here](https://github.com/wandb/weave/pull/1661#discussion_r1665218799) and the order of the underlying OpenAI calls are not recorded correctly. This fix corrects the issue and helps record the calls in the correct order by running the inline Tracer in the same thread as the execution.

eyurtsev: Can you give an example chain that would result in things being out of order? Are you basing in the implementation on the LangChain BaseTracer?

parambharat: hi @eyurtsev . Yes, the [implementation](https://github.com/wandb/weave/blob/1865731d18fef152ff6ef39c788180a7627d8e87/weave/integrations/langchain/langchain.py#L94) sub-classes the BaseTracer. Also, here's a [simple test](https://github.com/wandb/weave/blob/1865731d18fef152ff6ef39c788180a7627d8e87/weave/integrations/langchain/langchain_test.py#L91) where we trace the prompt and the openai call. The call order in the `ainvoke` case is expected to be the same as the one in the `invoke` case. However, the assertion fails because the Openai call is traced before tracing the `ChatOpenAI` and prompt runnable.

eyurtsev: This doesn't sound like something that's possible -- as far as I'm aware langsmith traces do not show any ordering issues like that.

Could you provide a minimal reproducible example?

>> However, the assertion fails because the Openai call is traced before tracing the ChatOpenAI and prompt runnable.

There's a typo here, do you mean the openai call is traced before the prompt? (i.e., or is there a 2nd call to chat open ai?)

eyurtsev: @parambharat ping here

morganmcg1: hey @eyurtsev , Bharat is on vacation, should be on this on Thursday :)

parambharat: Hi @eyurtsev,

Here's more context, we have been building a [Langchain Tracer](https://wandb.github.io/weave/guides/integrations/langchain) for using [W&B weave](https://wandb.ai/site/weave).

The tracer tracks all runnables in langchain using the BaseTracer API that langchain's tracing mechainism exposes. In addition to tracing langchain runnables, weave also traces Openai, Litellm, Anthropic and other such vendor API calls automatically. 

For example, when the following code is run

```
import weave

# Initialize Weave with your project name
weave.init("langchain_demo")

llm = ChatOpenAI()
prompt = PromptTemplate.from_template("1 + {number} = ")

llm_chain = prompt | llm

output = llm_chain.invoke({"number": 2})
```

We get a trace that looks as follows:
image

*Notice how the `openai.chat.completions.create` call is under the `langchain.llm.ChatOpenAI` runnable. ?*

However, when we run the same with `ainvoke` we notice that the `openai.chat.completions.create` is not run in the same thread as the `langchain runnables` resulting in two different calls as show in the following screenshot.


Screenshot 2024-07-23 at 12 12 27 PM


The simplest way to recreate this is to run the following code

```
import weave

# Initialize Weave with your project name
weave.init("langchain_demo")

llm = ChatOpenAI()
prompt = PromptTemplate.from_template("1 + {number} = ")

llm_chain = prompt | llm

output = await llm_chain.ainvoke({"number": 2})
```

This issue arises from the use of `asyncio.gather` without checking for the `run_inline` argument in the `AsyncCallbackManager`. The fix ensures that the underlying Openai calls run in the same thread as the `ChatOpenai` runnable thus ensuring that the tracing is in the correct hierarchy.

eyurtsev: The issue is probably that `gather` starts a new context, and the tracer is relying on the context to be able to look up the parent run id for the openai completions call. 

```python
import asyncio
import contextvars

c = contextvars.ContextVar("c", default=None)

async def foo():
    c.set("foo")
    print("within", c.get())

# Gather swaps context
await asyncio.gather(*(foo(),))
print("after", c.get())

# Regular uses same context
await foo()
print("way after", c.get())
c.get()
```

-----

If you want to merge a fix:

1. confirm that this is the problem (gather vs. non gather)
2. Add a unit test that shows the failure (in terms of context vars). Otherwise this will break in a future refactor. If possible do it in a separate PR and mark test as expected to fail.
3. Update this PR to pass in inline handlers vs. non-linline handlers rather than passing one handler at a time (and update PR summary to explain what the issue is and what the fix is).

Something in this spirit

```python

    inline_tasks = []
    non_inline_tasks = []
    inline_handlers = [
        handler for handler in self.handlers if handler.is_inline
    ]
    non_inline_handlers = [
        handler for handler in self.handlers if not handler.is_inline
    ]
    
    for prompt in prompts:
        if run_id is not None:
            run_id_ = run_id
            run_id = None
        else:
            run_id_ = uuid.uuid4()
            
        if inline_handlers:
            inline_tasks.append(
                ahandle_event(
                    inline_handlers,
                    "on_llm_start",
                    "ignore_llm",
                    serialized,
                    [prompt],
                    run_id=run_id_,
                    parent_run_id=self.parent_run_id,
                    tags=self.tags,
                    metadata=self.metadata,
                    **kwargs,
                )
            )
        else:
            non_inline_tasks.append(
                ahandle_event(
                    non_inline_handlers,
                    "on_llm_start",
                    "ignore_llm",
                    serialized,
                    [prompt],
                    run_id=run_id_,
                    parent_run_id=self.parent_run_id,
                    tags=self.tags,
                    metadata=self.metadata,
                    **kwargs,
                )
            )

        managers.append(
            AsyncCallbackManagerForLLMRun(
                run_id=run_id_,
                handlers=self.handlers,
                inheritable_handlers=self.inheritable_handlers,
                parent_run_id=self.parent_run_id,
                tags=self.tags,
                inheritable_tags=self.inheritable_tags,
                metadata=self.metadata,
                inheritable_metadata=self.inheritable_metadata,
            )
        )
        
    for inline_task in inline_tasks:
        await inline_task
        

    await asyncio.gather(*non_inline_tasks)
```

parambharat: hey @eyurtsev just quick nudge here.

parambharat: Hi @eyurtsev:  Thank you for the comment. In my previous [commit](https://github.com/langchain-ai/langchain/pull/23909/commits/0517c66aa02caafc05db0f97418689c3fc39f433) I already separated the inline and non-inline handlers as you've show in 3. I'll update the PRsummary. 

As for 1, yes the issue is `gather` bailing out the threads in a context and thus not having the parent calls for openai being tracked correctly.

For 2, I'll work on a separate PR for the tests. Please let me know what and where the right location for such tests is.

eyurtsev: >  I already separated the inline and non-inline handlers as you've show in 3.

Could you pass the list of handlers to handle event rather than a handler at a time. I want to keep as much logic as possible inside of handle event.  If you compare the difference between the PR and the proposed change, i'm passing a list of handlers in. 

> For 2, I'll work on a separate PR for the tests. Please let me know what and where the right location for such tests is.

We place unit tests inside of tests/unit_tests and then inside a matching namespace that matches the location of the code (e.g., callbacks).

eyurtsev:
Hi @parambharat I am going to close this PR for now. Please open new PRs when you're ready to implement this change. 

What we'd be looking for:

1. A PR with a unit tests that shows the failure (it should be specifically showing the failure with context vars). Mark the test as expected to fail.
2. A follow up PR that fixes the issue and correctly describes the problem w/ solution (i.e., asyncio.gather interacting with contextvars). The PR should follow the structure of the code I specified to make sure that as much of the event handling as possible stays inside handle_event and its async variant.

Please tag me, we can merge the the changes quickly since I think we know understand what the problem is

PR #23913, opened 2024-07-05
Neo4j: Fixed similarity docs

RafaelXokito:
**Description:** There was missing some documentation regarding the `filter` and `params` attributes in similarity search methods.

### DIFF
--- a/libs/community/langchain_community/vectorstores/neo4j_vector.py
+++ b/libs/community/langchain_community/vectorstores/neo4j_vector.py
@@ -909,6 +909,9 @@ class Neo4jVector(VectorStore):
         Args:
             query (str): Query text to search for.
             k (int): Number of results to return. Defaults to 4.
+            params (Dict[str, Any]): The search params for the index type.
+                Defaults to empty dict.
+            filter (Optional[Dict[str, Any]]): Dictionary of argument(s) to filter on metadata. Defaults to None.
 
         Returns:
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/vectorstores/neo4j_vector.py
```diff
@@ -1077,13 +1092,19 @@ def similarity_search_by_vector(
         embedding: List[float],
         k: int = 4,
         filter: Optional[Dict[str, Any]] = None,
+        params: Dict[str, Any] = {},
```
eyurtsev, line 1095: This introduces a bug since param is exposed in the function signature, but isn't propagated

eyurtsev (changes requested):
@RafaelXokito thanks for the PR! When you resolve the issue, feel free to ask me to re-review the code

libs/community/langchain_community/vectorstores/neo4j_vector.py
RafaelXokito: Thank you very much for your feedback. I forgot to use the `params` attribute.
Already updated the PR.
This change was basically to keep all the methods with similar signatures, since they all use `similarity_search_with_score_by_vector` method that receives the `params` argument.

RafaelXokito:
@eyurtsev please review this PR.

eyurtsev (approved):

PR #23915, opened 2024-07-05
core, anthropic[patch]: support streaming tool calls when function has no arguments

### LINKED ISSUE #23911: bug: anthropic streaming tool calls for tools with no arguments
### Privileged issue

### Issue Content

`.tool_calls` attribute of aggregated chunks can be empty, whereas result of `.invoke` is not.

```python

def magic_function() -> int:
    """Calculates a magic function."""
    return 5

llm = ChatAnthropic(
    model="claude-3-haiku-20240307",
).bind_tools([magic_function])

query = "What is the value of magic_function()?"

full = None
for chunk in llm.stream(query):

print(full.tool_calls)
print(llm.invoke(query).tool_calls)
```
```
[]
[{'name': 'magic_function', 'args': {}, 'id': 'toolu_01HHtSuCJ4LKQfGRYncy4D5a'}]
```

ccurme:
resolves https://github.com/langchain-ai/langchain/issues/23911

When an AIMessageChunk is instantiated, we attempt to parse tool calls off of the tool_call_chunks.

Here we add a special-case to this parsing, where `""` will be parsed as `{}`.

This is a reaction to how Anthropic streams tool calls in the case where a function has no arguments:
```
{'id': 'toolu_01J8CgKcuUVrMqfTQWPYh64r', 'input': {}, 'name': 'magic_function', 'type': 'tool_use', 'index': 1}
{'partial_json': '', 'type': 'tool_use', 'index': 1}
```
The `partial_json` does not accumulate to a valid json string-- most other providers tend to emit `"{}"` in this case.


ccurme:
Integration tests passing for partner packages: https://github.com/langchain-ai/langchain/actions/runs/9811346308

libs/core/langchain_core/messages/ai.py
```diff
@@ -253,14 +253,23 @@ def init_tool_calls(cls, values: dict) -> dict:
...
-                        id=chunk["id"],
-                        error=None,
+                if chunk["args"] == "":
+                    tool_calls.append(
+                        ToolCall(
+                            name=chunk["name"] or "",
+                            args={},
+                            id=chunk["id"],
+                        )
+                    )
```
efriis, line 263: should we add this case before the exception gets thrown? maybe in the `args_ = parse_partial... if chunk["args"] != "" else {}` or something

efriis (commented):
small nit otherwise lgtm

PR #23941, opened 2024-07-07
Community: Fix Deprecated Ruff Command Usage in Makefile

happyhackerbird:
Deprecated `ruff <path> `command was used in `Makefile`, which caused errors during the execution of formatting & linting tasks. 

Changes Made:
Replaced all instances of `ruff <path> `with `ruff check <path> `in the `Makefile` to align with the new syntax required by the latest version of ruff.


eyurtsev:
Thanks @happyhackerbird i believe that we already merged these changes in!

PR #23977, opened 2024-07-08
core[minor]: Scope out rework of callback system

eyurtsev:
Scoping out how to re-work the callback system to simplify it while keeping it backwards comaptible


eyurtsev:
Closing for now, but we may come  back to this

PR #23990, opened 2024-07-08
core[minor]: IndexManager

eyurtsev:
A draft proposal for an index manager. Prior to rolling out:

- [ ] Verify with implementation for an existing vectorstore
- [ ] Add async version
- [ ] Add documentation of semantics

The IndexManager adds a control plane for indexes: 
* It'll encourage decoupling the creation of the indexes from the usage of the indexes (e.g., decouple postgres table creation from using it to search over data)
* Opens up a pathway toward implementing schema migrations for implementations that rely on schema (e.g., postgres)

### DIFF
--- a/libs/core/langchain_core/indexing/index_manager.py
+++ b/libs/core/langchain_core/indexing/index_manager.py
@@ -5,8 +5,9 @@ This index manager will provide a control plane for existing indexes.
 It also provides a path toward providing implementations the ability to
 take care of schema migrations!
 """
+
 import abc
-from typing import Any, TypedDict, List
+from typing import Any, List, TypedDict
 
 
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/indexing/index_manager.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class IndexInformation(TypedDict):
+    """Information about an index."""
+
+    id: str
+    """A string that uniquely identifies the index. Required."""
+
+
+class Index(abc.ABC):
```
eyurtsev, line 26: This is a stub for the full abstract index (not implemented here)

PR #23992, opened 2024-07-08
core[patch]: support conversion of runnables to tools

ccurme:
Open to other thoughts on UX.

string input:
```python
as_tool = retriever.as_tool()
as_tool.invoke("cat")  # [Document(...), ...]
```

typed dict input:
```python
class Args(TypedDict):
    key: int

def f(x: Args) -> str:
    return str(x["key"] * 2)

as_tool = RunnableLambda(f).as_tool(
    name="my tool",
    description="description",  # name, description are inferred if not supplied
)
as_tool.invoke({"key": 3})  # "6"
```

for untyped dict input, allow specification of parameters + types
```python
def g(x: Dict[str, Any]) -> str:
    return str(x["key"] * 2)

as_tool = RunnableLambda(g).as_tool(arg_types={"key": int})
result = as_tool.invoke({"key": 3})  # "6"
```

Passing the `arg_types` is slightly awkward but necessary to ensure tool calls populate parameters correctly:
```python
from typing import Any, Dict

def f(x: Dict[str, Any]) -> str:
    return str(x["key"] * 2)

runnable = RunnableLambda(f)
as_tool = runnable.as_tool(arg_types={"key": int})

llm = ChatOpenAI().bind_tools([as_tool])

result = llm.invoke("Use the tool on 3.")
tool_call = result.tool_calls[0]
args = tool_call["args"]
assert args == {"key": 3}

as_tool.run(args)
```

Contrived (?) example with langgraph agent as a tool:
```python
from typing import List, Literal

llm = ChatOpenAI(temperature=0)


    return input + 2


agent_1 = create_react_agent(llm, [magic_function])


class Message(TypedDict):
    role: Literal["human"]
    content: str

agent_tool = agent_1.as_tool(
    arg_types={"messages": List[Message]},
    name="Jeeves",
    description="Ask Jeeves.",
)

agent_2 = create_react_agent(llm, [agent_tool])
```


libs/core/tests/unit_tests/test_tools.py
```diff
@@ -987,3 +988,33 @@ def foo(
         },
         "required": ["bar", "baz"],
     }
+
+
+def test_convert_from_runnable() -> None:
+    def f(x: List[int]) -> int:
```
eyurtsev, line 994: Is there a version of this that uses a single `RunnableLambda`?

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -987,3 +988,33 @@ def foo(
...
+        """Calculate a maximum."""
+        return max(x)
+
+    def g(x: int) -> str:
```
eyurtsev, line 998: Most of our Runnables as defined as dicts for the first argument. Would you mind sharing an example of how that would look like?

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -987,3 +988,33 @@ def foo(
...
+
+    runnable = RunnableLambda(f) | g
+    as_tool = convert_runnable_to_tool(runnable)
+    args_schema = as_tool.args_schema
+    assert args_schema is not None
+    assert args_schema.schema() == {
+        "title": "RunnableSequenceSchema",
+        "type": "object",
+        "properties": {
+            "input": {"title": "Input", "type": "array", "items": {"type": "integer"}}
```
eyurtsev, line 1010: Questions above are b/c i'm wondering what this would look like

libs/core/langchain_core/tools.py
```diff
@@ -1218,3 +1218,32 @@ class BaseToolkit(BaseModel, ABC):
...
+def _get_description_from_runnable(runnable: Runnable) -> str:
+    """Generate a placeholder description of a runnable."""
+    input_schema = runnable.input_schema.schema()
+    return f"Takes {input_schema}."
+
+
+def _get_args_schema_for_runnable(runnable: Runnable) -> Type[BaseModel]:
+    """Get object schema from runnable input schema."""
+    schema_name = f"{runnable.get_name()}Schema"
+    input_name = "input"
```
eyurtsev, line 1232: Are we able to pick the appropriate input-name based on the runnable sequence?

libs/core/langchain_core/tools.py
```diff
@@ -1218,3 +1218,32 @@ class BaseToolkit(BaseModel, ABC):
...
+    fields = {input_name: (runnable.InputType, Field(...))}
+    return create_model(schema_name, **fields)  # type: ignore
+
+
+def convert_runnable_to_tool(
```
eyurtsev, line 1237: One odd thing is that we lose any optimizations from the runnables that might exist (e.g., batch, stream) etc.

Do we need a more direct way to promote a runnable to a tool, and have the structured tool delegate to the underlying methods?

libs/core/tests/unit_tests/test_tools.py
ccurme: added

libs/core/tests/unit_tests/test_tools.py
ccurme: added

libs/core/langchain_core/tools.py
ccurme: newest implementation does this

libs/core/langchain_core/runnables/base.py
```diff
@@ -2006,6 +2007,19 @@ async def _atransform_stream_with_config(
             if hasattr(iterator_, "aclose"):
                 await iterator_.aclose()
 
+    def as_tool(
+        self,
+        *,
+        name: Optional[str] = None,
+        description: Optional[str] = None,
+        arg_types: Optional[Dict[str, Type]] = None,
```
bracesproul, line 2015: not the biggest fan of this name. why not use `input_schema` or `paramaters` like the other major providers use?

bracesproul: Ah ignore, I see you guys use `args_schema` in py where we use `schema` in js.

libs/community/langchain_community/tools/e2b_data_analysis/tool.py
```diff
@@ -234,7 +234,7 @@ def remove_uploaded_file(self, uploaded_file: UploadedFile) -> None:
         ]
         self.description = self.description + "\n" + self.uploaded_files_description
 
-    def as_tool(self) -> Tool:
+    def as_tool(self) -> Tool:  # type: ignore[override]
```
efriis, line 237:
```suggestion
    @override
    def as_tool(self) -> Tool:
```

I think this is more standard

libs/core/langchain_core/runnables/base.py
```diff
@@ -2006,6 +2007,19 @@ async def _atransform_stream_with_config(
...
+    ) -> BaseTool:
+        from langchain_core.tools import convert_runnable_to_tool
```
efriis, line 2017: nit: comment that this is for import time optimization i assume

libs/core/langchain_core/tools.py
efriis: + on this. would it be better to create a `StructuredTool.from_runnable` and add overrides for batch/stream in the event `self._runnable` instead of `self.func` is set?

efriis (commented):
I like it. Mix of thoughts in comments

libs/core/langchain_core/runnables/base.py
ccurme: I import here to avoid a circular import

libs/community/langchain_community/tools/e2b_data_analysis/tool.py
baskaryan: unrelated to this pr but does anyone remember why this method is needed here

libs/core/langchain_core/tools.py
baskaryan: think this is probably an issue with tools that can be handled separately (eg StructuredTool.from_function should take batch/stream variants)

PR #23994, opened 2024-07-08
core[minor]: Add ToolMessage.raw_output

baskaryan:
Decisions to discuss:
1.  is a new attr needed or could additional_kwargs be used for this
2. is raw_output a good name for this attr
3. should raw_output default to {} or None
4. should raw_output be included in serialization
5. do we need to update repr/str to  exclude raw_output


libs/core/langchain_core/messages/tool.py
```diff
@@ -39,6 +55,16 @@ class ToolMessage(BaseMessage):
     type: Literal["tool"] = "tool"
     """The type of the message (used for serialization). Defaults to "tool"."""
 
+    raw_output: Dict[str, Any] = Field(default_factory=dict, exclude=True)
+    """The raw output of the tool.
+    
+    **Not part of the payload sent to the model.** Should only be specified if it is 
+    different from the message content, i.e. if only a subset of the full tool output
```
ccurme, line 62: (nit) does raw_output always contain content or could content be derived from raw output?

libs/core/langchain_core/messages/tool.py
```diff
@@ -83,6 +109,7 @@ def __add__(self, other: Any) -> BaseMessageChunk:  # type: ignore
             return self.__class__(
                 tool_call_id=self.tool_call_id,
                 content=merge_content(self.content, other.content),
+                raw_output=merge_dicts(self.raw_output, other.raw_output),
```
ccurme, line 112: I don't think this is a concern because I'm not sure how often we're summing ToolCallChunks, but this might typically error, e.g.,
```python
message_1 = ToolMessageChunk("foo", raw_output={"bar": {"baz": 123}}, tool_call_id="1")
message_2 = ToolMessageChunk("foo", raw_output={"bar": {"baz": 234}}, tool_call_id="1")
message_1 + message_2
```
```
TypeError: Additional kwargs key baz already exists in left dict and value has unsupported type <class 'int'>
```

libs/core/pyproject.toml
```diff
@@ -38,6 +38,11 @@ python = ">=3.12.4"
 [tool.ruff.lint]
 select = [ "E", "F", "I", "T201",]
 
+[tool.ruff.lint.per-file-ignores]
```
ccurme, line 41: Thank you for these

ccurme (approved):

libs/core/langchain_core/messages/tool.py
baskaryan: yea, but my thinking is that's an inavlid stream of chunks then. not clear how to combine them

libs/core/langchain_core/messages/tool.py
```diff
@@ -39,6 +54,14 @@ class ToolMessage(BaseMessage):
     type: Literal["tool"] = "tool"
     """The type of the message (used for serialization). Defaults to "tool"."""
 
+    raw_output: Any = None
```
ccurme, line 57: can this be `Any` and support serialization at the same time?

baskaryan: message will fail to deserialize if this isn't serializable (see unit test)

libs/core/langchain_core/utils/_merge.py
```diff
@@ -72,3 +72,26 @@ def merge_lists(left: Optional[List], *others: Optional[List]) -> Optional[List]
                 else:
                     merged.append(e)
     return merged
+
+
+def merge_obj(left: Any, right: Any) -> Any:
```
eyurtsev, line 77: what fun

eyurtsev (approved):

PR #24016, opened 2024-07-09
community: creation score_threshold in flashrank_rerank.py

morgandiverrez:
Description: 
 add a optional score relevance threshold for select only coherent document, it's in complement of top_n

Discussion:
add relevance score threshold in flashrank_rerank document compressors #24013

Dependencies:
 no dependencies


libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -69,11 +71,10 @@ def compress_documents(
...
 
         for r in rerank_response:
-            metadata = r["meta"]
-            metadata["relevance_score"] = r["score"]
-            doc = Document(
-                page_content=r["text"],
-                metadata=metadata,
-            )
-            final_results.append(doc)
+            if r["score"] > self.score_threshold:
```
Benvii, line 74: In the doc string it's indicated that it's the "Minimum relevance threshold to return", this mean that document with this exact score should be included in the result.

```suggestion
            if r["score"] >= self.score_threshold:
```

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -69,11 +71,10 @@ def compress_documents(
...
+                doc = Document(
+                    page_content=r["text"],
+                    metadata={"flash_rerank": {"id": r["id"], "relevance_score": r["score"]}, **r["metadata"]},
```
Benvii, line 77: This library is named `FlashrankRerank` please use this naming also.

```suggestion
                    metadata={"flashrank_rerank": {"id": r["id"], "relevance_score": r["score"]}, **r["metadata"]},
```

Benvii (changes requested):
See my 2 small comments

Benvii (approved):
Ok for me

morgandiverrez:
hello @baskaryan , is it possible to review this PR  please ?

Benvii (approved):

Benvii:
Hi @ccurme as you have already reviewed some PR about flash rank rerank, would you have a few minutes to look at this tiny PR (+9 −8) that introduce score filtering in it ?

It's really useful to filter relevant documents without having to create a custom compressor that juste perform this filtering based on metadata. It also keeps documents meta data which is crucial for any kind of production usage.

Thanks in advance 🙂

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -60,7 +62,7 @@ def compress_documents(
         callbacks: Optional[Callbacks] = None,
     ) -> Sequence[Document]:
         passages = [
-            {"id": i, "text": doc.page_content, "meta": doc.metadata}
+            {"id": i, "text": doc.page_content, "metadata": doc.metadata}
```
ccurme, line 65: is `RerankRequest` sensitive to the name of this key? what is the motivation for this change?

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -69,11 +71,10 @@ def compress_documents(
...
-            metadata["relevance_score"] = r["score"]
-            doc = Document(
-                page_content=r["text"],
-                metadata=metadata,
-            )
-            final_results.append(doc)
+            if r["score"] >= self.score_threshold:
+                doc = Document(
+                    page_content=r["text"],
+                    metadata={"flashrank_rerank": {"id": r["id"], "relevance_score": r["score"]}, **r["metadata"]},
```
ccurme, line 77: does moving putting `relevance_score` inside `flash_rerank` represent a breaking change for users?

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -69,11 +71,10 @@ def compress_documents(
...
 
         for r in rerank_response:
-            metadata = r["meta"]
-            metadata["relevance_score"] = r["score"]
-            doc = Document(
-                page_content=r["text"],
-                metadata=metadata,
-            )
-            final_results.append(doc)
+            if r["score"] >= self.score_threshold:
```
ccurme, line 74: is there any possible configuration of `client` that would return negative scores?

libs/community/langchain_community/document_compressors/flashrank_rerank.py
morgandiverrez: To prevent overloading the already defined IDs in the metadata and to identify that this information is related to flashrank_rerank, we can avoid modifications to the metadata logic by adding a parameter prefix_metadata, if defined, gets added as a prefix to the ID and relevance_score. This way, you can ensure that the data is properly associated with flashrank_rerank without altering the existing metadata structure

libs/community/langchain_community/document_compressors/flashrank_rerank.py
morgandiverrez: The scores provided by the client come from a softMax activation function and therefore can only be positive.

libs/community/langchain_community/document_compressors/flashrank_rerank.py
morgandiverrez: To prevent overloading the already defined IDs in the metadata and to identify that this information is related to flashrank_rerank, we can avoid modifications to the metadata logic by adding a parameter prefix_metadata, if defined, gets added as a prefix to the ID and relevance_score. This way, you can ensure that the data is properly associated with flashrank_rerank without altering the existing metadata structure

libs/community/langchain_community/document_compressors/flashrank_rerank.py
morgandiverrez: Just to clarify, but we can leave it as meta, it doesn't matter.

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -26,8 +26,12 @@ class FlashrankRerank(BaseDocumentCompressor):
     """Flashrank client to use for compressing documents"""
     top_n: int = 3
     """Number of documents to return."""
+    score_threshold: float = 0.0
+    """Minimum relevance threshold to return."""
     model: Optional[str] = None
     """Model to use for reranking."""
+    prefix_metadata: Optional[str] = None
```
Benvii, line 33: Set it to empty string by default  otherwise it will break when it's not defined.
```suggestion
    prefix_metadata: Optional[str] = ""
```

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -69,11 +73,11 @@ def compress_documents(
...
-            doc = Document(
-                page_content=r["text"],
-                metadata=metadata,
-            )
-            final_results.append(doc)
+            if r["score"] >= self.score_threshold:
+                doc = Document(
+                    page_content=r["text"],
+                    metadata={self.prefix_metadata + "id": r["id"],
+                              self.prefix_metadata + "relevance_score": r["score"], **r["meta"]},
```
Benvii, line 80: Apply the linter and push it again plz 🙂

Benvii (changes requested):

Benvii (approved):

libs/community/langchain_community/document_compressors/flashrank_rerank.py
```diff
@@ -26,8 +26,12 @@ class FlashrankRerank(BaseDocumentCompressor):
     """Flashrank client to use for compressing documents"""
     top_n: int = 3
     """Number of documents to return."""
+    score_threshold: float = 0.0
+    """Minimum relevance threshold to return."""
     model: Optional[str] = None
     """Model to use for reranking."""
+    prefix_metadata: str = ""
```
ccurme, line 33: Changed this from `Optional[str]` to `str` as there was no handling for the `None` case below (e.g., in `self.prefix_metadata + "id"`). Let me know if this wasn't intended.

ccurme (approved):

PR #24018, opened 2024-07-09
parsed_json is expected to be a list of dictionaries, but it seems to…

basirsedighi:
parsed_json is expected to be a list of dictionaries, but it seems to… be a single dictionary instead.
This is at libs/experimental/langchain_experimental/graph_transformers/llm.py process process_response

- [ ] **Bugfix**: "experimental: bugfix"


Undertone0809 (approved):

libs/experimental/langchain_experimental/graph_transformers/llm.py
```diff
@@ -717,9 +717,9 @@ def process_response(self, document: Document) -> GraphDocument:
         else:
             nodes_set = set()
             relationships = []
-            if not isinstance(raw_schema, str):
-                raw_schema = raw_schema.content
```
ccurme, line 721: why delete this?

ccurme (approved):

PR #24027, opened 2024-07-09
fix results description tavily_search.py

GabboM:


ccurme:
Could you please explain in the PR summary why we are moving the docstring?

Closing but will re-open if needed.

PR #24031, opened 2024-07-09
core: Move json parsing in base chat model / output parser to bg thread

nfcampos:
- add version of AIMessageChunk.__add__ that can add many chunks, instead of only 2
- In agenerate_from_stream merge and parse chunks in bg thread
- In output parse base classes do more work in bg threads where appropriate

### DIFF
--- a/libs/core/langchain_core/output_parsers/transform.py
+++ b/libs/core/langchain_core/output_parsers/transform.py
@@ -17,6 +17,7 @@ from langchain_core.outputs import (
     Generation,
     GenerationChunk,
 )
+from langchain_core.runnables.config import run_in_executor
 
 if TYPE_CHECKING:
     from langchain_core.runnables import RunnableConfig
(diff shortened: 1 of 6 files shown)


hinthornw (approved):

hinthornw (approved):

libs/core/langchain_core/output_parsers/transform.py
```diff
@@ -37,9 +38,13 @@ async def _atransform(
     ) -> AsyncIterator[T]:
         async for chunk in input:
             if isinstance(chunk, BaseMessage):
-                yield self.parse_result([ChatGeneration(message=chunk)])
+                yield await run_in_executor(
```
eyurtsev, line 41: Should we create a single thread with an event loop and delegate to it instead of re-recreating on every chunk?

eyurtsev (approved):

libs/core/langchain_core/output_parsers/transform.py
nfcampos: Hmm no the event loop manages its own thread pool, no need for us to repeat that work

PR #24033, opened 2024-07-09
Faster Stream

hinthornw:
updates on nuno's changes; incorporates the aggregated json parsing within the overloaded add operation

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -94,13 +94,9 @@ def generate_from_stream(stream: Iterator[ChatGenerationChunk]) -> ChatResult:
         ChatResult: Chat result.
     """
 
-    generation: Optional[ChatGenerationChunk] = None
-    for chunk in stream:
-        if generation is None:
-            generation = chunk
-        else:
-            generation += chunk
-    assert generation is not None
+    generation = next(stream, None)
+    if generation:
+        generation += list(stream)
     return ChatResult(
         generations=[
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/messages/base.py
```diff
@@ -196,6 +201,13 @@ def __add__(self, other: Any) -> BaseMessageChunk:  # type: ignore
                     self.response_metadata, other.response_metadata
                 ),
             )
+        elif isinstance(other, list) and all(
+            isinstance(o, BaseMessageChunk) for o in other
+        ):
+            base = self
+            for o in other:
+                base += o
```
nfcampos, line 209: can we do the more efficient version calling merge_x with the full lists?

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -86,46 +84,29 @@ class LangSmithParams(TypedDict, total=False):
...
+    generation = next(stream, None)
+    if generation:
+        generation += list(stream)
+    if generation is None:
+        raise ValueError("No generations found in stream.")
+    return ChatResult(
+        generations=[
+            ChatGeneration(
+                message=message_chunk_to_message(generation.message),
+                generation_info=generation.generation_info,
```
nfcampos, line 106: this way we lose generation info from all the other chunks, is this on purpose?

nfcampos: Ah nevermind

nfcampos (approved):

PR #24035, opened 2024-07-09
rfc: tool call id as part of runnable config

baskaryan:


baskaryan (commented):
Snippet from change we'd make to langgraph ToolNode
```python
            try:
                output = self.tools_by_name[call["name"]].invoke(
                    call["args"], {**config, **{"tool_call_id": call["id"]}}
                )
            except Exception as e:
                if not self.handle_tool_errors:
                    raise e
                output = f"Error: {repr(e)}\n Please fix your mistakes."
            if isinstance(output, ToolMessage):
                return output
            else:
                return ToolMessage(
                    content=str_output(output),
                    name=call["name"],
                    tool_call_id=call["id"],
                )
```

PR #24036, opened 2024-07-09
core: Speed up json parse for large strings

nfcampos:
for a large string:
- old 4.657918874989264
- new 0.023724667000351474


hinthornw (approved):
nice. as god intended

PR #24038, opened 2024-07-09
core[minor], integrations...[patch]: Support ToolCall as Tool input and ToolMessage as Tool output

baskaryan:
Changes:
- ToolCall, InvalidToolCall and ToolCallChunk can all accept a "type" parameter now
- LLM integration packages add "type" to all the above
- Tool supports ToolCall inputs that have "type" specified
- Tool outputs ToolMessage when a ToolCall is passed as input
- Tools can separately specify ToolMessage.content and ToolMessage.raw_output
- Tools emit events for validation errors (using on_tool_error and on_tool_end)

Example:
```python
@tool("structured_api", response_format="content_and_raw_output")
def _mock_structured_tool_with_raw_output(
    arg1: int, arg2: bool, arg3: Optional[dict] = None
) -> Tuple[str, dict]:
    """A Structured Tool"""
    return f"{arg1} {arg2}", {"arg1": arg1, "arg2": arg2, "arg3": arg3}


def test_tool_call_input_tool_message_with_raw_output() -> None:
    tool_call: Dict = {
        "name": "structured_api",
        "args": {"arg1": 1, "arg2": True, "arg3": {"img": "base64string..."}},
        "id": "123",
        "type": "tool_call",
    }
    expected = ToolMessage("1 True", raw_output=tool_call["args"], tool_call_id="123")
    tool = _mock_structured_tool_with_raw_output
    actual = tool.invoke(tool_call)
    assert actual == expected

    tool_call.pop("type")
    with pytest.raises(ValidationError):
        tool.invoke(tool_call)

    actual_content = tool.invoke(tool_call["args"])
    assert actual_content == expected.content
```


baskaryan:
corresponding ToolNode update https://github.com/langchain-ai/langgraph/pull/977

libs/core/langchain_core/tools.py
```diff
@@ -511,67 +512,37 @@ def run(
...
-            return observation
+            content = _handle_tool_error(e, flag=self.handle_tool_error)
+            run_manager.on_tool_end(content, color="red", name=self.name, **kwargs)
         except (Exception, KeyboardInterrupt) as e:
             run_manager.on_tool_error(e)
             raise e
         else:
-            run_manager.on_tool_end(observation, color=color, name=self.name, **kwargs)
-            return observation
+            run_manager.on_tool_end(content, color=color, name=self.name, **kwargs)
```
eyurtsev, line 543: This affects the astream events API. I'd argue the raw output is more important in that case than the stringified version of it.

The least surprising thing is either ToolMessage or else the 2-tuple output?

libs/core/langchain_core/tools.py
```diff
@@ -908,6 +808,8 @@ def from_function(
         return_direct: bool = False,
         args_schema: Optional[Type[BaseModel]] = None,
         infer_schema: bool = True,
+        *,
+        has_raw_output: bool = False,
```
eyurtsev, line 812: `has` -> `returns_raw_output` OR `return_format`: Literal['2-tuple']  maybe>

libs/core/langchain_core/messages/tool.py
```diff
@@ -152,6 +157,19 @@ class ToolCallChunk(TypedDict):
     """An identifier associated with the tool call."""
     index: Optional[int]
     """The index of the tool call in a sequence."""
+    type: NotRequired[Literal["tool_call_chunk"]]
+
+
+def tool_call_chunk(
```
eyurtsev, line 163: why is this needed?

eyurtsev (approved):
partial review

libs/core/langchain_core/tools.py
```diff
@@ -371,46 +374,25 @@ def get_input_schema(
...
+        tool_input, kwargs = _prep_run_args(input, config, **kwargs)
+        return await self.arun(tool_input, **kwargs)
 
     # --- Tool ---
 
-    def _parse_input(
-        self,
-        tool_input: Union[str, Dict],
-    ) -> Union[str, Dict[str, Any]]:
+    def _parse_input(self, tool_input: Union[str, Dict]) -> Union[str, Dict[str, Any]]:
```
eyurtsev, line 395: What can tool input be here? (if you know of the top of you head and have time could be good to update the doc-string)

libs/core/langchain_core/tools.py
```diff
@@ -439,23 +421,45 @@ def raise_deprecation(cls, values: Dict) -> Dict:
...
+                self._run_include_raw_output,
+                *args,
+                **kwargs,
+            )
+        else:
+            if not signature(self._arun).parameters.get("run_manager"):
+                kwargs.pop("run_manager")
+            return (await self._arun(*args, **kwargs)), None
+
+    def _run(self, *args: Any, **kwargs: Any) -> Any:
```
eyurtsev, line 451: Given that `_run` is basically untyped, what are your thoughts on adding an attribute instead to the tool that specifies the expected return format of `_run`. That way we don't need to introduce a new method

libs/core/langchain_core/messages/tool.py
baskaryan: convenience / to make sure ppl don't mess up the type

libs/core/langchain_core/tools.py
baskaryan: could log whatever actual output is (so ToolMessage if thats whats returned below)

libs/core/langchain_core/messages/base.py
```diff
@@ -11,6 +11,8 @@
 if TYPE_CHECKING:
     from langchain_core.prompts.chat import ChatPromptTemplate
 
+Content = Union[str, List[Union[str, Dict]]]
```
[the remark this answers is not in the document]
baskaryan: used for typing elsewhere

libs/core/langchain_core/tools.py
baskaryan: think this is mostly a question of minimizing contributor friction, down for whatever makes contributor life easiest

libs/core/langchain_core/messages/base.py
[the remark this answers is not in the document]
baskaryan: (but not used for typing hinting below because using vars like this obfuscates whats going on in the api ref, which is something we should fix)

libs/core/langchain_core/tools.py
eyurtsev: I think a single `_run` with an attribute would be easier to understand

libs/core/langchain_core/messages/tool.py
eyurtsev: How is that easier than using ToolCallChunk? Is it validating that there are no extra arguments or typos in the arguments?

libs/core/langchain_core/tools.py
eyurtsev: Let's do `ToolMessage`

libs/core/langchain_core/messages/tool.py
baskaryan: type is hard coded

ccurme:
Integration tests look good: https://github.com/langchain-ai/langchain/actions/runs/9893304257

libs/partners/groq/langchain_groq/chat_models.py
```diff
@@ -511,19 +512,19 @@ def _stream(
             generation = chat_result.generations[0]
             message = cast(AIMessage, generation.message)
             tool_call_chunks = [
-                {
-                    "name": rtc["function"].get("name"),
-                    "args": rtc["function"].get("arguments"),
-                    "id": rtc.get("id"),
-                    "index": rtc.get("index"),
-                }
+                create_tool_call_chunk(
```
ccurme, line 515: There are places we've duplicated logic (e.g., in `astream`) that may also need to be updated.

What will the behavior be if a tool call is missing a `type`? Possible there will be issues operating on old serialized tool calls (ok if so-- helps to know in case bugs pop up)? And do we need to update other integrations (e.g., community and other repos)?

libs/core/langchain_core/tools.py
```diff
@@ -586,99 +557,67 @@ async def arun(
...
         )
-        new_arg_supported = signature(self._arun).parameters.get("run_manager")
         run_manager = await callback_manager.on_tool_start(
             {"name": self.name, "description": self.description},
             tool_input if isinstance(tool_input, str) else str(tool_input),
             color=start_color,
             name=run_name,
-            inputs=tool_input,
             run_id=run_id,
+            # Inputs by definition should always be dicts.
```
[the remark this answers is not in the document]
baskaryan: copy/pasted from elsewhere, cc @eyurtsev

libs/partners/groq/langchain_groq/chat_models.py
baskaryan: only thing that depends on type atm is Tools when they're taking a ToolCall (which is a new feature, you couldn't pass in ToolCalls at all before)

libs/core/langchain_core/tools.py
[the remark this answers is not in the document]
eyurtsev: Comment is originally from observing this:

```python
@tool
def foo(x: str):
   return x

foo.invoke({'x': 'hello'})
```

It's possible that str is still supported on its own. I don't know what we want to do in that case -- since we have pretty odd behavior when it comes to langsmith API requiring auto upgrades to dicts

PR #24054, opened 2024-07-10
core[patch]: Add encoding options when create prompt template from a file

shenhai-ran:
- Uses default utf-8 encoding for loading prompt templates from file

### DIFF
--- a/libs/core/langchain_core/prompts/prompt.py
+++ b/libs/core/langchain_core/prompts/prompt.py
@@ -213,6 +213,7 @@ class PromptTemplate(StringPromptTemplate):
     def from_file(
         cls,
         template_file: Union[str, Path],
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/prompts/prompt.py
```diff
@@ -213,13 +213,15 @@ def from_examples(
     def from_file(
         cls,
         template_file: Union[str, Path],
+        encoding: Optional[str] = "utf-8",
```
eyurtsev, line 216: This is a breaking change for users that have been passing the encoding by position. You can move it after input_variables and use the `*` operator to force invocation by name.

Change looks good, could you add a simple unit test given that this is langchain-core?

eyurtsev (approved):

PR #24065, opened 2024-07-10
community[minor]: Add graph store extractors

cbornet:
This adds an extractor interface and an implementation for HTML pages.
Extractors are used to create GraphVectorStore Links on loaded content.

**Twitter handle:** cbornet_

### DIFF
--- a/libs/community/langchain_community/graph_vectorstores/extractors/html_link_extractor.py
+++ b/libs/community/langchain_community/graph_vectorstores/extractors/html_link_extractor.py
@@ -1,0 +1,124 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, List, Optional, Set, Union
+from urllib.parse import urldefrag, urljoin, urlparse
+
+from langchain_core.documents import Document
+from langchain_core.graph_vectorstores import Link
+
+from langchain_community.graph_vectorstores.extractors.link_extractor import (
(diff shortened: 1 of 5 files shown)


libs/community/langchain_community/graph_vectorstores/extractors/html_link_extractor.py
```diff
@@ -0,0 +1,124 @@
...
+        parse_url = _parse_url(link, page_url=url, drop_fragments=drop_fragments)
+        # Remove self links and entries for any 'a' tag that failed to parse
+        # (didn't have href, or invalid domain, etc.)
+        if parse_url and parse_url != url:
+            links.add(parse_url)
+
+    return links
+
+
+@dataclass
```
eyurtsev, line 58: If possible
```suggestion
@dataclass(frozen=True)
```

eyurtsev:
@cbornet if you're confident with the abstractions we can also commit it to `langchain` which is meant for provider agnostic implementations.

eyurtsev (approved):
Looks good to me overall -- going to give it another look likely today and then merge

libs/community/langchain_community/graph_vectorstores/extractors/html_link_extractor.py
bjchambers: Should be possible.

PR #24068, opened 2024-07-10
community[minor]: passthrough auth parameter on requests to Ollama-LLMs

rick-soptim:

**Description:**
This PR allows users of `langchain_community.llms.ollama.Ollama` to specify the `auth` parameter, which is then forwarded to all internal calls of `requests.request`. This works in the same way as the existing `headers` parameters. The auth parameter enables the usage of the given class with Ollama instances, which are secured by more complex authentication mechanisms, that do not only rely on static headers. An example are AWS API Gateways secured by the IAM authorizer, which expects signatures dynamically calculated on the specific HTTP request.

**Issue:**

Integrating a remote LLM running through Ollama using `langchain_community.llms.ollama.Ollama` only allows setting static HTTP headers with the parameter `headers`. This does not work, if the given instance of Ollama is secured with an authentication mechanism that makes use of dynamically created HTTP headers which for example may depend on the content of a given request.

**Dependencies:**

None

**Twitter handle:**

None


eyurtsev (approved):

eyurtsev (commented):
Left a suggestion in case unit tests fail

eyurtsev (approved):

eyurtsev:
There are some broken unit tests. Are you able to resolve those?

rick-SOPTIM:
I think I fixed the broken unit tests, but I'm not sure, because I'm having trouble running them locally.

rick-SOPTIM:
I fixed the remaining problem found by the linter.

Issue #24064, opened 2024-07-10
AzureSearch vectorstore does not work asyncronously


thedavgar:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

I am trying to use the Azure AI Search Vectorstore and retriever and the vectorstore and retriever (given from the vectorstore) work perfectly when doing retrieval of documents using the syncronous methods but gives an error when trying to run the async methods.

Creating the instances of embeddings and Azure Search
```python
from azure.search.documents.indexes.models import (
    SearchField,
    SearchFieldDataType,
    SimpleField,
)
from langchain_openai import AzureOpenAIEmbeddings
from langchain_community.vectorstores import AzureSearch

fields = [
  SimpleField(
      name="content",
      type=SearchFieldDataType.String,
      key=True,
      filterable=True,
  ),
  SearchField(
      name="content_vector",
      type=SearchFieldDataType.Collection(SearchFieldDataType.Single),
      searchable=True,
      vector_search_dimensions=1536,
      vector_search_profile_name="myHnswProfile",
  ),
  SimpleField(
      name="document_name",
      type=SearchFieldDataType.String,
      key=True,
      filterable=True,
  )
]

encoder = AzureOpenAIEmbeddings(
    azure_endpoint=os.getenv("EMBEDDINGS_OPENAI_ENDPOINT"),
    deployment=os.getenv("EMBEDDINGS_DEPLOYMENT_NAME"),
    openai_api_version=os.getenv("OPENAI_API_VERSION"),
    openai_api_key=os.getenv("AZURE_OPENAI_API_KEY"),
)

vectorstore = AzureSearch(
    azure_search_endpoint=os.getenv("AI_SEARCH_ENDPOINT_SECRET"),
    azure_search_key=os.getenv("AI_SEARCH_API_KEY"),
    index_name=os.getenv("AI_SEARCH_INDEX_NAME_SECRET"),
    fields=fields,
    embedding_function=encoder,
)

retriever = vectorstore.as_retriever(search_type="hybrid", k=2)
```

Syncronous methods working and returning documents
```python
vectorstore.vector_search("what is the capital of France")
retriever.invoke("what is the capital of France")
```

[...]

PR #24081, opened 2024-07-10
community: Fix Bug in Azure Search Vectorstore search asyncronously

thedavgar:

**Description**:
This PR fixes a bug described in the issue in #24064, when using the AzureSearch Vectorstore with the asyncronous methods to do search which is also the method used for the retriever. The proposed change includes just change the access of the embedding as optional because is it not used anywhere to retrieve documents. Actually, the syncronous methods of retrieval do not use the embedding neither.

With this PR the code given by the user in the issue works.

```python
vectorstore = AzureSearch(
    azure_search_endpoint=os.getenv("AI_SEARCH_ENDPOINT_SECRET"),
    azure_search_key=os.getenv("AI_SEARCH_API_KEY"),
    index_name=os.getenv("AI_SEARCH_INDEX_NAME_SECRET"),
    fields=fields,
    embedding_function=encoder,
)

retriever = vectorstore.as_retriever(search_type="hybrid", k=2)

await vectorstore.avector_search("what is the capital of France")
await retriever.ainvoke("what is the capital of France")
```

**Issue**:
The Azure Search Vectorstore is not working when searching for documents with asyncronous methods, as described in issue #24064

**Dependencies**:
There are no extra dependencies required for this change.

### DIFF
--- a/libs/community/langchain_community/vectorstores/azuresearch.py
+++ b/libs/community/langchain_community/vectorstores/azuresearch.py
@@ -1049,7 +1049,7 @@ class AzureSearch(VectorStore):
         *,
         filters: Optional[str] = None,
         **kwargs: Any,
-    ) -> Tuple[List[Document], List[float], List[List[float]]]:
+    ) -> Tuple[List[Document], List[float], Optional[List[List[float]]]]:
         """Perform vector or hybrid search in the Azure search index.
 
         Args:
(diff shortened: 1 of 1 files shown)


efriis (approved):

chrislrobert:
@thedavgar @efriis @isahers1, this PR caused the regression in issue #24740. I agree that it seems to have fixed the exception that motivated the change, but it also pulled async results handling outside the `with` block, which caused a new exception. See #24740 for details.

PR #24083, opened 2024-07-10
langchain: add document_variable_name to create_stuff_documents_chain

xdumaine:
- **Description:** `StuffDocumentsChain` uses `LLMChain` which is deprecated by langchain runnables. `create_stuff_documents_chain` is the replacement, but needs support for `document_variable_name` to allow multiple uses of the chain within a longer chain.
- **Issue:** none
- **Dependencies:** none


libs/langchain/langchain/chains/combine_documents/stuff.py
```diff
@@ -91,9 +96,16 @@ def format_docs(inputs: dict) -> str:
     ).with_config(run_name="stuff_documents_chain")
 
 
+@deprecated(
```
eyurtsev, line 99: @xdumaine we can merge the rest of this except for the deprecation. We're trying to figure out how to deprecate some of these things without causing too much churn for users that have existing code that's relying on this. 

cc @ccurme  -- has the most knowledge here in terms of which chains are safe to deprecate

eyurtsev (changes requested):
We can merge after the `@deprecation` is removed. We're mostly want the @ deprecated decorator to be used by core maintainers since we have to plan the roll out carefully for each deprecation

libs/langchain/langchain/chains/combine_documents/stuff.py
xdumaine: @eyurtsev no problem, thank you for the review/feedback!

eyurtsev (approved):

xdumaine:
Thank you!

PR #24084, opened 2024-07-10
[Core] Inherit RT Tags/Meta

hinthornw:
I think this may make local tags inheritable, however

### DIFF
--- a/libs/core/langchain_core/callbacks/manager.py
+++ b/libs/core/langchain_core/callbacks/manager.py
@@ -2079,6 +2079,12 @@ def _configure(
         )
         for handler in local_handlers_:
             callback_manager.add_handler(handler, False)
+    if run_tree is not None:
+        if run_tree.tags is not None:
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/callbacks/manager.py
```diff
@@ -2079,6 +2079,12 @@ def _configure(
         )
         for handler in local_handlers_:
             callback_manager.add_handler(handler, False)
+    if run_tree is not None:
+        if run_tree.tags is not None:
+            callback_manager.add_tags(run_tree.tags, False)
```
eyurtsev, line 2084: Booleans are easier to read if invoked by name 

```suggestion
            callback_manager.add_tags(run_tree.tags, inherit=False)
```

eyurtsev: Shouldn't inherit be True though?

PR #24088, opened 2024-07-10
community[minor]: Add ApertureDB as a vectorstore

bovlb:

- [X] *ApertureDB as vectorstore**: "community: Add ApertureDB as a vectorestore"

    - **Description:** this change provides a new community integration that uses ApertureData's ApertureDB as a vector store.
    - **Issue:** none
    - **Dependencies:** depends on ApertureDB Python SDK
    - **Twitter handle:** ApertureData

- [X] **Add tests and docs**: If you're adding a new integration, please include

Integration tests rely on a local run of a public docker image.
Example notebook additionally relies on a local Ollama server.

All lint tests pass.

Additional guidelines:


libs/community/tests/integration_tests/vectorstores/test_aperturedb.py
```diff
@@ -0,0 +1,203 @@
+"""Test ApertureDB functionality."""
+
```
eyurtsev, line 2: Could you test this with our standard test suite? if it passes the unit tests we can merge.

https://github.com/langchain-ai/langchain/blob/14a8bbc21ac81f4a0eb03db981780c86cbdb9a7f/libs/community/tests/unit_tests/vectorstores/test_inmemory.py#L6-L6

libs/community/langchain_community/vectorstores/aperturedb.py
```diff
@@ -0,0 +1,446 @@
...
+                properties=properties,
+            )
+
+            # Create indexes
+            self.utils.create_entity_index("_Descriptor", "_create_txn")
+            self.utils.create_entity_index("_DescriptorSet", "_name")
+            self.utils.create_entity_index("_Descriptor", UNIQUEID_PROPERTY)
+
+    @override
+    def add_documents(self, documents: List[Document], **kwargs: Any) -> List[str]:
```
eyurtsev, line 195: You can implement `upsert` now instead of `add_documents` or `add_texts`

libs/community/langchain_community/vectorstores/aperturedb.py
```diff
@@ -0,0 +1,446 @@
...
+        self.utils = Utils(self.connection)
+        try:
+            self.utils.status()
+        except Exception:
+            self.logger.exception("Failed to connect to ApertureDB")
+            raise
+
+        self._find_or_add_descriptor_set()
+
+    def _find_or_add_descriptor_set(self) -> None:
```
eyurtsev, line 108: FYI This makes network connections on init -- it'll complicate async support in the future.

Given that a lot of vectorstores currently do this, we'll accept the code as is

libs/community/tests/integration_tests/vectorstores/test_aperturedb.py
bovlb: Ah, I didn't see those.  I was going by existing examples.  I'll convert to that.

libs/community/langchain_community/vectorstores/aperturedb.py
bovlb: OK.  I saw that `upsert` got added, but I chose to implement it using `add_documents`.  Now that `add_documents` and `add_texts` have default implementations that use `upsert`, I could reverse that.

libs/community/langchain_community/vectorstores/aperturedb.py
bovlb: Hmm.  We do prefer to create connections on init.  What is the async use case that will break with this?  Would it be better to open the connection lazily or do we need to be using multiple connections?

libs/community/tests/integration_tests/vectorstores/test_aperturedb.py
eyurtsev: It was recently added, we're in the process of adopting this in more places -- it'll make it easier for vectorstore maintainers to verify that their implementation is correct and hopefully be able to avoid re-inventing all the test scaffolding

bovlb: In `test_deleting_documents`, I see:
```
ids = vectorstore.add_documents(documents, ids=["1", "2"])
```
Is it a new convention that `add_documents` should support an `ids` parameter?  Shouldn't we be using the `id` field in Document instead?

eyurtsev: Not a new convention -- unfortunately it's an OLD API and most implementations are expected to support it. If you implement just upsert  and drop add_documents, the correct behavior will be implemented for you including correct precedence rules for `ids` provided via add_documents vs. ids on the document object itself

bovlb: OK.  I have converted to the standard tests and they run successfully.  

I had a bit of a journey to getting the devcontainer working. 

First I had to change the Poetry version by setting `POETRY_VERSION=1.4.2` in `dev.Dockerfile`

Then I had to execute the following commands:
```bash
pip3 install pytest-vcr
pip3 install -e libs/standard-tests/
pip3 install aperturedb
adb config create host.docker.internal --host host.docker.internal --no-interactive --active
```

I also had to add the following In `.devcontainer/docker-compose.yaml` to `services/langchain`:
```yaml
extra_hosts:
    "host.docker.internal:host-gateway" 
```

libs/community/langchain_community/vectorstores/aperturedb.py
bovlb: As noted above, I have recoded to remove `add_documents`and `add_texts` in favour of `upsert`.

libs/community/tests/integration_tests/vectorstores/test_aperturedb.py
eyurtsev: most of the core developers don't use the devcontainer as far as I know -- I think it's mostly maintained by the community. I'm happy to merge the changes for the dev container

libs/community/langchain_community/vectorstores/aperturedb.py
```diff
@@ -0,0 +1,516 @@
...
+            descriptor_set: The name of the descriptor set to delete
+        """
+        from aperturedb.Utils import Utils, create_connector
+
+        db = create_connector()
+        utils = Utils(db)
+        utils.remove_descriptorset(descriptor_set)
+
+    @classmethod
+    def list_vectorstores(class_) -> None:
```
eyurtsev, line 439: We're thinking about adding an abstraction for the control plane that will allow listing, creating, deleting indices. Let me know if you would be interested in testing it out

eyurtsev (approved):

libs/community/langchain_community/vectorstores/aperturedb.py
eyurtsev: Sorry thought I responded already! 

The issues are: 
1. you can't make an async call from init without kicking off a thread with a new event loop in it
2. users may also end up with a different life cycle for the vectorstore (e.g., instantiating it on every request served by a web server) -- this is not an async specific issue

Regardless it's fine for now, most vector stores implement this pattern

eyurtsev:
@efriis / @jacoblee93 any idea?

image

libs/community/tests/integration_tests/vectorstores/test_aperturedb.py
bovlb: > I'm happy to merge the changes for the dev container

I'm planning to submit a separate PR for that.

libs/community/langchain_community/vectorstores/aperturedb.py
bovlb: Yes, please let me know.

PR #24103, opened 2024-07-11
core: mermaid: Render metadata key-value pairs when drawing mermaid graph

nfcampos:
- if node is runnable binding with metadata attached

### DIFF
--- a/libs/core/langchain_core/runnables/graph.py
+++ b/libs/core/langchain_core/runnables/graph.py
@@ -64,12 +64,32 @@ class Edge(NamedTuple):
     data: Optional[Stringifiable] = None
     conditional: bool = False
 
+    def copy(
+        self, *, source: Optional[str] = None, target: Optional[str] = None
+    ) -> Edge:
+        return Edge(
+            source=source or self.source,
+            target=target or self.target,
+            data=self.data,
+            conditional=self.conditional,
+        )
+
 
 class Node(NamedTuple):
     """Node in a graph."""
 
     id: str
+    name: str
     data: Union[Type[BaseModel], RunnableType]
+    metadata: Optional[Dict[str, Any]]
+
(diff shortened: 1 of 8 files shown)


libs/core/langchain_core/runnables/graph.py
```diff
@@ -111,7 +112,7 @@ class MermaidDrawMethod(Enum):
     API = "api"  # Uses Mermaid.INK API to render the graph
 
 
-def node_data_str(node: Node) -> str:
+def node_data_str(node: Node, *, html: bool = False) -> str:
```
vbarda, line 115: (nit) maybe a more descriptive name than html? `render_metadata`?

vbarda:
could you add a test for a node w/ metadata?

libs/core/langchain_core/runnables/graph.py
eyurtsev, line 115: update doc-string?

What's `html`?

libs/core/langchain_core/runnables/graph.py
```diff
@@ -111,35 +132,25 @@ class MermaidDrawMethod(Enum):
     API = "api"  # Uses Mermaid.INK API to render the graph
 
 
-def node_data_str(node: Node) -> str:
+def node_data_str(id: str, data: Union[Type[BaseModel], RunnableType]) -> str:
```
efriis, line 135: might be worth switching all these to be `_` prefixed to make clear they're internal

libs/core/langchain_core/runnables/graph.py
```diff
@@ -111,35 +132,25 @@ class MermaidDrawMethod(Enum):
...
     """Convert the data of a node to a string.
 
     Args:
         node: The node to convert.
+        html: Whether to format the data as HTML rich text.
```
vbarda, line 140: need to update doctring here

PR #24118, opened 2024-07-11
docs[patch]: Adds guide on passing additional params to a tool

jacoblee93:
CC @hinthornw


ccurme (approved):
Would it make sense to commit `curry` somewhere (e.g., `langchain_core.utils`), or even to add as a method to `StructuredTool`? e.g., `bind_tool_params` or something similar.

eyurtsev:
1. I think I'm opposed to merging this -- it feels far too hacky and something that should be done by the framework.
2.  It's a much more complex version of this:  https://langchain-git-jacob-toolsadditionalparams-langchain.vercel.app/v0.2/docs/how_to/tool_runtime/ 


If you're still keen on merging I think it should be at least consolidated with the existing tool_runtime how to doc

eyurtsev (changes requested):
This is too complex even if it solves an immediate problem I don't think this should be part of the how-to doc as an official recommendation

jacoblee93:
> This is too complex even if it solves an immediate problem I don't think this should be part of the how-to doc as an official recommendation

If we moved `curry` into `utils`, would that be more palatable?

I actually didn't notice the other page - it actually seems like that solves the problems we're worried about here?

PR #24123, opened 2024-07-11
core[patch]: propagate `parse_docstring` to tool decorator

ccurme:
Disabled by default.

```python

@tool(parse_docstring=True)
def foo(bar: str, baz: int) -> str:
    """The foo.

    Args:
        bar: this is the bar
        baz: this is the baz
    """
    return bar


foo.args_schema.schema()
```
```json
{
  "title": "fooSchema",
  "description": "The foo.",
  "type": "object",
  "properties": {
    "bar": {
      "title": "Bar",
      "description": "this is the bar",
      "type": "string"
    },
    "baz": {
      "title": "Baz",
      "description": "this is the baz",
      "type": "integer"
    }
  },
  "required": [
    "bar",
    "baz"
  ]
}
```


libs/core/tests/unit_tests/test_tools.py
```diff
@@ -959,6 +959,19 @@ def foo(bar: str, baz: int) -> str:
         "required": ["bar", "baz"],
     }
 
+    foo2 = tool(foo, parse_docstring=True)
```
eyurtsev, line 962: Could you add a test that shows that this throws an error it the doc-string does not follow the correct format?

eyurtsev:
1. Could you define the behavior to raise an exception on invalid doc-strings (shown below)
2. Could you add a test that verifies what happens if the tool requests the `run_manager`?


```python

@tool(parse_docstring=True)
def foo(bar: str, baz: int) -> str:
    """The foo.

    Args:
        banana: this is the bar
        monkey: this is the baz
    """
    return bar


foo.args_schema.schema()
```

----

ccurme:
> 2. Could you add a test that verifies what happens if the tool requests the `run_manager`?

Can you clarify more specifics here? There is an existing test but I'm not sure if it's missing a condition:
https://github.com/langchain-ai/langchain/blob/16e178a8c2c26c6f7d14825ea7c0d0a4a91b8a59/libs/core/tests/unit_tests/test_tools.py#L431-L446

libs/core/tests/unit_tests/test_tools.py
ccurme: done

eyurtsev (approved):

PR #24130, opened 2024-07-11
community[minor],core[major]: Move GraphVectorStore to community

eyurtsev:
* This PR moves graph vectorstore to community until there is more than one implementation based on it and the abstraction gets battle-tested.
* Renaming Node and Link to GraphStoreNode and GraphStoreLink as Node is already as a langgraph primitive.


eyurtsev:
Hi @cbornet, we're going to move the implementation to community for now. I think the strategy going forward will be to get community contributed abstractions contributed to langchain_community prior to being adopted into core to get battle-tested / proved as useful

eyurtsev:
@cbornet were you planning on adding implementations to the langchain_astradb partner package?

eyurtsev:
@cbornet touched based internally, we'll keep in core for now with `@beta` decorator. We wanted to update the names of `Node` and `Link`  to avoid collision with langgraph primitives `Node`. Let us know if `GraphStoreNode` and `GraphStoreLink` are OK or if you have a better suggestion?

cbornet:
What about `GraphNode`/`GraphLink` to keep the names short ?

eyurtsev:
Closing in favor of: https://github.com/langchain-ai/langchain/pull/24195

PR #24141, opened 2024-07-11
core[patch]: Adds curry utils function

jacoblee93:
CC @ccurme @eyurtsev @hinthornw


libs/core/langchain_core/utils/curry.py
```diff
@@ -0,0 +1,39 @@
...
+    Returns a new function whose signature omits the curried variables.
+
+    Args:
+        func: The function to curry.
+        curried_kwargs: Arguments to apply to the function.
+
+    Returns:
+        A new function with curried arguments applied.
+
+    .. versionadded:: 0.2.14"""
```
eyurtsev, line 18:
```suggestion
    .. versionadded:: 0.2.14
    """
```

eyurtsev:
@baskaryan shall we discuss tomorrow how to do this?

Do we want a curry() method on BaseTools directly so users don't have to fish for this in utils?

jacoblee93:
Per offline discussion, keeping as a util

PR #24142, opened 2024-07-11
core, standard-tests, partners: update `bind_tools` to accept Runnable

ccurme:
I don't think we should rely on passing runnables into `bind_tools`, since there's no nice way to pass in corresponding names/descriptions. Whereas it's easy to do
```python
.bind_tools(
    [
        runnable_1.as_tool(description=...),
        runnable_2.as_tool(description=...),
    ]
)
```
But `bind_tools` could accept Runnable in principle.


ccurme:
Standard test passing: https://github.com/langchain-ai/langchain/actions/runs/9896944107

ccurme:
Implementation on Vertex: https://github.com/langchain-ai/langchain-google/compare/cc/runnables_as_tools

PR #24143, opened 2024-07-11
core[patch]: Tool accept RunnableConfig

baskaryan:
Relies on #24038


libs/core/tests/unit_tests/test_tools.py
```diff
@@ -925,6 +930,81 @@ async def foo(bar: str) -> str:
...
+@tool
+async def afoo(bar: Any, bar_config: RunnableConfig) -> Any:
+    """The foo."""
+    await asyncio.sleep(0.0001)
+    assert bar_config["configurable"]["foo"] == "not-bar"
+    assert bar == "baz"
+    return bar
+
+
+@tool(infer_schema=False)
```
hinthornw, line 947: If we're gonna be doing bool toggles should we just define the tool in the test function and then parametrize the args?

libs/core/langchain_core/tools.py
```diff
@@ -1318,3 +1352,17 @@ def invoke_wrapper(callbacks: Optional[Callbacks] = None, **kwargs: Any) -> Any:
...
+def _get_runnable_config_param(func: Callable) -> Optional[str]:
+    if isinstance(func, functools.partial):
+        func = func.func
+    try:
+        type_hints = get_type_hints(func)
+    except Exception:
+        return None
+    else:
+        for name, type_ in type_hints.items():
+            if type_ is RunnableConfig:
```
eyurtsev, line 1366: awesome

libs/core/langchain_core/tools.py
```diff
@@ -1318,3 +1352,17 @@ def invoke_wrapper(callbacks: Optional[Callbacks] = None, **kwargs: Any) -> Any:
             description=description,
             args_schema=args_schema,
         )
+
+
+def _get_runnable_config_param(func: Callable) -> Optional[str]:
+    if isinstance(func, functools.partial):
```
hinthornw, line 1358: Feel like we do this in a few places. Maybe something we should just make a util for?  (the get_type_hints specifically)

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -925,6 +930,81 @@ async def foo(bar: str) -> str:
...
+    name: str = "Foo"
+    description: str = "Foo"
+
+    def _run(self, bar: Any, bar_config: RunnableConfig, **kwargs: Any) -> Any:
+        assert bar_config["configurable"]["foo"] == "not-bar"
+        assert bar == "baz"
+        return bar
+
+    async def _arun(self, bar: Any, bar_config: RunnableConfig, **kwargs: Any) -> Any:
+        await asyncio.sleep(0.0001)
```
eyurtsev, line 987: move sleep unless we really need it?

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -925,6 +930,81 @@ async def foo(bar: str) -> str:
...
+    """The foo."""
+    assert bar_config["configurable"]["foo"] == "not-bar"
+    assert bar == "baz"
+    return bar
+
+
+@tool
+async def afoo(bar: Any, bar_config: RunnableConfig) -> Any:
+    """The foo."""
+    await asyncio.sleep(0.0001)
```
eyurtsev, line 944: hy sleeping??

eyurtsev (approved):

libs/core/tests/unit_tests/test_tools.py
baskaryan: i am a simple assistant trained by openai. when i see code i copy pasta it. i'll be taking no further questions at this time. because i am incapable of answering them

libs/core/langchain_core/tools.py
```diff
@@ -784,31 +800,39 @@ def args(self) -> dict:
...
         *args: Any,
+        config: RunnableConfig,
         run_manager: Optional[CallbackManagerForToolRun] = None,
         **kwargs: Any,
     ) -> Any:
         """Use the tool."""
         if self.func:
             if run_manager and signature(self.func).parameters.get("callbacks"):
                 kwargs["callbacks"] = run_manager.get_child()
+            if config_param := _get_runnable_config_param(self.func):
```
hinthornw, line 811: should we just do this once instead of on every invocation?

baskaryan: i guess teeeechnically self.func is mutable

libs/core/langchain_core/tools.py
baskaryan: maybe for a sep pr?

PR #24145, opened 2024-07-11
core[minor],langchain[minor]: RFC test out indexing abstraction

eyurtsev:
Test out indexing abstraction on parent retriever

### DIFF
--- a/libs/core/langchain_core/indexing/base_index.py
+++ b/libs/core/langchain_core/indexing/base_index.py
@@ -16,6 +16,7 @@ from typing import (
     Optional,
     Union,
     Dict,
+    Literal,
 )
 
 from langchain_core._api import beta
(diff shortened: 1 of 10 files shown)


libs/core/langchain_core/indexing/base_index.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        # Developer guidelines
+        # 1. The filter should be a dictionary or a list of dictionaries.
+        # 2. An invalid filter should raise an exception.
+        # 3. A None filter is considered valid and should return items.
+        # 4. The **default** filter syntax should follow standard LangChain
+        #    filtering syntax.
+        #    The syntax is as follows:
+        #    - All operators and comparators should be prefixed with a "$".
+        #    - Field names are expected to be valid identifiers allowing [a-zA-Z0-9_]
```
eyurtsev, line 260: TBD nested fields

libs/core/langchain_core/indexing/base_index.py
```diff
@@ -0,0 +1,346 @@
...
+        #        "filter": "and(or(eq('field', 'value'), eq('field2', 'value2')))"A
+        #    }
+
+    @abc.abstractmethod
+    def delete_by_filter(
+        self,
+        filter: Union[Dict[str, Any], List[Dict[str, Any]]],
+        /,
+        **kwargs: Any,
+    ) -> Iterable[DeleteResponse]:
```
eyurtsev, line 289: IMO We should probably make this method just return a delete response, and stop worrying about including the ids of deleted content in the delete response as part of this interface. Those IDs aren't that useful and take extra effort to get. It'll also mean that folks don't have to iterate through delete by filter to use it.

libs/core/langchain_core/indexing/base_index.py
```diff
@@ -0,0 +1,346 @@
...
+        Users should not assume that the order of the returned items matches
+        the order of the input IDs. Instead, users should rely on the ID field of the
+        returned items.
+
+        This method should **NOT** raise exceptions if no items are found for
+        some IDs.
+        """
+
+    @abc.abstractmethod
+    def delete_by_ids(
```
eyurtsev, line 213: Potentially merge with `delete` to avoid introducing another method.

We'd need to update the response type here to just be a bool. We likely don't need counts or IDs anyway

PR #24146, opened 2024-07-11
docs: Add NVIDIA NIMs to Model Tab and Feature Table

haydeniw:
**Description:** Add NVIDIA NIMs to Model Tab and LLM Feature Table


efriis (commented):
responded in slack - let's get the feature-specific standard tests passing for chatnvidia! I don't think tool calling works quite yet.

tabs change looks good to me

docs/scripts/model_feat_table.py
```diff
@@ -87,6 +87,14 @@
         "package": "langchain-huggingface",
         "link": "/docs/integrations/chat/huggingface/",
     },
+    "ChatNVIDIA": {
+        "tool_calling": True,
+        "json_mode": True,
```
efriis, line 92:
```suggestion
        "json_mode": False,
```
If the api supports json mode, could you link me where in the docs that's shown?

efriis (approved):

Issue #24116, opened 2024-07-11
Milvus fails on `add_documents()` with new `VectorStore.upsert()` method


amiigas:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
import logging
logging.basicConfig(level=logging.DEBUG)

from langchain_core.documents import Document
from langchain_core.embeddings import FakeEmbeddings
from langchain_milvus import Milvus


milvus_vector_store = Milvus(
    index_params = {
        "metric_type": "COSINE",
        "index_type": "HNSW",
        "params": {"M": 8, "efConstruction": 64},
    },
    embedding_function=FakeEmbeddings(size=100),
    collection_name="test_langchain",
    connection_args={...}
)

docs = [
    Document(page_content='First document content.'),
    Document(page_content='Second document content.'),
]
ids = [
    "id_1",
    "id_2"
]

result = milvus_vector_store.add_documents(ids=ids, documents=docs)
print(result)
```

### Error Message and Stack Trace (if applicable)

```bash
DEBUG:langchain_milvus.vectorstores.milvus:No documents to upsert.
Traceback (most recent call last):
  File "/Users/amiigas/langchain_milvus_test.py", line 32, in <module>
    result = milvus_vector_store.add_documents(ids=ids, documents=docs)
  File "/Users/amiigas/.venv/lib/python3.10/site-packages/langchain_core/vectorstores/base.py", line 463, in add_documents
    return self.upsert(documents_, **kwargs)["succeeded"]
TypeError: 'NoneType' object is not subscriptable
```

### Description

 * Trying to add documents to Milvus vector store using `add_documents()` method.
 * Expecting to successfully add documents and see returned ids `["id_1", "id_2"]` as a method result
 * Instead method returns `None` which then throws `TypeError` after trying to access `succeeded` key
 

`VectorStore` passes `documents_` to `Milvus.upsert()` as a positional arg, which results in `documents=None` and returning `None` early

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 23.5.0: Wed May  1 20:13:18 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6030
[...]

PR #24152, opened 2024-07-11
milvus[patch]: Fix Milvus vectorstore for newer versions of langchain-core

eyurtsev:
Fix for: https://github.com/langchain-ai/langchain/issues/24116

This keeps the old behavior of add_documents and add_texts


efriis (approved):

fbossiere:
The patch introduced a bug in the aadd_documents method. The issue stems from the absence of an aadd_texts implementation in milvus.py. As a result, the method falls back to the base VectorStore class's aadd_documents, which then recursively calls the Milvus-specific aadd_documents. This creates an infinite loop, breaking the functionality.

PR #24153, opened 2024-07-11
standard-tests[patch]: add tests for runnables as tools and streaming usage metadata

ccurme:


ccurme:
Tests pass: https://github.com/langchain-ai/langchain/actions/runs/9899224780

PR #24156, opened 2024-07-11
langchain: default to langsmith sdk for pulling prompts, fallback to langchainhub

madams0013:
**Description:** Deprecating langchainhub, replacing with langsmith sdk

### DIFF
--- a/libs/langchain/langchain/hub.py
+++ b/libs/langchain/langchain/hub.py
@@ -10,11 +10,11 @@ from langchain_core.load.load import loads
 from langchain_core.prompts import BasePromptTemplate
 
 if TYPE_CHECKING:
-    from langsmith import Client
-    from langchainhub import Client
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/hub.py
```diff
@@ -33,7 +36,7 @@ def push(
     api_url: Optional[str] = None,
     api_key: Optional[str] = None,
     parent_commit_hash: Optional[str] = "latest",
-    new_repo_is_public: bool = True,
+    new_repo_is_public: bool = False,
```
efriis, line 39: possible to make these `Optional[bool] = None` and have the defaults encoded in sdk?

efriis (approved):

libs/langchain/langchain/hub.py
madams0013: yup

PR #24160, opened 2024-07-12
docs[patch]: Update tool child run docs

jacoblee93:
Documents #24143


docs/docs/how_to/index.mdx
```diff
@@ -195,7 +195,7 @@ LangChain [Tools](/docs/concepts/#tools) contain a description of the tool (to p
 - [How to: add a human in the loop to tool usage](/docs/how_to/tools_human)
 - [How to: handle errors when calling tools](/docs/how_to/tools_error)
 - [How to: disable parallel tool calling](/docs/how_to/tool_choice)
-- [How to: stream events from within a tool](/docs/how_to/tool_stream_events)
+- [How to: configure and stream steps from subruns within a tool](/docs/how_to/tool_stream_events)
```
baskaryan, line 198: nit: old title was a bit clearer to me

also think it might be worth having a separate how for just "How to: access the Runnable config within a tool" (since you might care about that independent of event streaming)

jacoblee93: Yeah agreed, I think two smaller guides would be better here rather than trying to squeeze both into one

jacoblee93: Done

baskaryan:
how to pass config to custom tool page lgtm! only nit is maybe worth very explicitly pointing out that param name doesn't matter

jacoblee93:
Done

PR #24161, opened 2024-07-12
core: docstrings `runnables` update

leo-gan:

### DIFF
--- a/libs/core/langchain_core/runnables/graph.py
+++ b/libs/core/langchain_core/runnables/graph.py
@@ -57,7 +57,14 @@ def is_uuid(value: str) -> bool:
 
 
 class Edge(NamedTuple):
-    """Edge in a graph."""
+    """Edge in a graph.
+
+    Parameters:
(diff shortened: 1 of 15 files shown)


libs/core/langchain_core/runnables/graph.py
```diff
@@ -76,14 +92,27 @@ def copy(
 
 
 class Node(NamedTuple):
-    """Node in a graph."""
+    """Node in a graph.
+
+    Parameters:
```
eyurtsev, line 97: Missing this one likely?

eyurtsev (approved):

libs/core/langchain_core/runnables/graph.py
leo-gan: Good catch, Thanks!

eyurtsev (approved):

leo-gan:
@eyurtsev Thanks! Much appreciated. It was a big review.

PR #24162, opened 2024-07-12
Del schema

hinthornw:
i really want this

### DIFF
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -290,7 +290,7 @@ def create_schema_from_function(
     # Pydantic adds placeholder virtual fields we need to strip
     valid_properties = _get_filtered_args(inferred_model, func, filter_args=filter_args)
     return _create_subset_model(
-        f"{model_name}Schema",
+        model_name,
         inferred_model,
         list(valid_properties),
         descriptions=arg_descriptions,


libs/core/langchain_core/tools.py
```diff
@@ -290,7 +290,7 @@ def create_schema_from_function(
     # Pydantic adds placeholder virtual fields we need to strip
     valid_properties = _get_filtered_args(inferred_model, func, filter_args=filter_args)
     return _create_subset_model(
-        f"{model_name}Schema",
```
RafaelXokito, line 293: You need to take into account the users that might rely on this name.

RafaelXokito (changes requested):

libs/core/langchain_core/tools.py
eyurtsev: @hinthornw we could try and make this change as part of 0.3.

@RafaelXokito are you relying on this feature?

RafaelXokito: @eyurtsev no I don't, thank you for asking.
I thought it would be good idea to  consider that scenario.

hinthornw:
Ya not going to land this in 0.2

eyurtsev:
change was made in v0.3

PR #24015, opened 2024-07-09
qdrant: New sparse embeddings provider interface - PART 1

Anush008:
## Description

This PR introduces a new sparse embedding provider interface to work with the new Qdrant implementation that will follow this PR.

Additionally, an implementation of this interface is provided with https://github.com/qdrant/fastembed.

This PR will be followed by https://github.com/Anush008/langchain/pull/3.

PR #24164, opened 2024-07-12
qdrant: new Qdrant implementation

Anush008:
## Description

Part 2 following #24015.

This PR adds a new `QdranVectorStore` class that supports dense/sparse/hybrid retrievals. Works only with Qdrant v1.10.0 or above.

Tests are to be introduced in a follow-up PR. #24165

### DIFF
--- a/libs/partners/qdrant/langchain_qdrant/__init__.py
+++ b/libs/partners/qdrant/langchain_qdrant/__init__.py
@@ -1,5 +1,13 @@
 from langchain_qdrant.fastembed_sparse import FastEmbedSparse
+from langchain_qdrant.qdrant import QdrantVectorStore, RetrievalMode
 from langchain_qdrant.sparse_embeddings import SparseEmbeddings, SparseVector
 from langchain_qdrant.vectorstores import Qdrant
 
-__all__ = ["Qdrant", "SparseEmbeddings", "SparseVector", "FastEmbedSparse"]
+__all__ = [
+    "Qdrant",
+    "QdrantVectorStore",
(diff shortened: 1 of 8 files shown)


efriis (approved):

eyurtsev:
FYI we now have standard RWAPI tests for vectorstores would be great to verify that qdrant passes all of them

Anush008:
> FYI we now have standard RWAPI tests for vectorstores would be great to verify that qdrant passes all of them

Hey @eyurtsev, could you please elaborate on this? How can I do this?

mkhludnev:
Hi @Anush008 
Thanks for this impl. Here's one novice's question. I noticed `async a*` methods and explicit `async_client` in deprecated Qdrant. And QdrantVectorStore relies on ascendant `async` methods. Does it make it less performant in certain condition?

Anush008:
I guess we didn't have the base async fallback at the time, so you has to implement using the async client.

Regardless, we're here now 😁.

mkhludnev:
Hi @Anush008 
Just FYI. I made an experiment with async implementation #26795. Takeaways:
1. I lost an idea of async generator from the old implementation https://github.com/langchain-ai/langchain/blob/2a4c5713cdabd184532454e2c06b836d23682556/libs/partners/qdrant/langchain_qdrant/vectorstores.py#L2190 and it's worth to get back to it approach. I probably check it out. 
2. async code truly runs parallel, but  wall clock gain with locally running qdrant and CPU embedding is not really significant, but anyway, specialized asyncs might be a good path toward performance.

PR #24165, opened 2024-07-12
qdrant: test new QdrantVectorStore

Anush008:
## Description

This PR adds integration tests to follow up on #24164.

By default, the tests use an in-memory instance.

To run the full suite of tests, with both in-memory and Qdrant server:

```
$ docker run -p 6333:6333 qdrant/qdrant

$ make test

$ make integration_test
```


efriis (approved):
Just confirming you ran these locally to confirm they work? They'll block the release otherwise

Anush008:
> Just confirming you ran these locally to confirm they work?
Yes.

PR #24166, opened 2024-07-12
qdrant: Documentation for the new QdrantVectorStore class

Anush008:
## Description

Follow up on #24165. Adds a page to document the latest usage of the new `QdrantVectorStore` class.

PR #24168, opened 2024-07-12
qdrant: Bump patch version

Anush008:
# Description

To release a new version of `langchain-qdrant` after #24165 and #24166.


efriis (approved):
will merge and release after #24166 is fixed

efriis:
Actually docs don't have to block. Merging and releasing

PR #24177, opened 2024-07-12
Fix neo4j type error on missing constraint information

tomasonjo:
If you use `refresh_schema=False`, then the metadata constraint doesn't exist. ATM, we used default `None` in the constraint check, but then `any` fails because it can't iterate over None value


jexp (approved):
LGTM

ccurme (approved):

PR #24185, opened 2024-07-12
core[minor]: rename ToolMessage.raw_output -> artifact

baskaryan:


eyurtsev (approved):

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -1150,33 +1150,33 @@ class _MockStructuredToolWithRawOutput(BaseTool):
...
 
     def _run(
         self, arg1: int, arg2: bool, arg3: Optional[dict] = None
     ) -> Tuple[str, dict]:
         return f"{arg1} {arg2}", {"arg1": arg1, "arg2": arg2, "arg3": arg3}
 
 
-@tool("structured_api", response_format="content_and_raw_output")
-def _mock_structured_tool_with_raw_output(
+@tool("structured_api", response_format="content_and_artifact")
```
eyurtsev, line 1161: alternative could be `with_artifact` a bit less to type, but less precise

baskaryan: "content" vs "with_artifact" feels less clear to me than "content" vs "content_and_artifact"

PR #24186, opened 2024-07-12
community[minor]: add document transformer for extracting links

bjchambers:
- **Description:** Add a DocumentTransformer for executing one or more `LinkExtractor`s and adding the extracted links to each document.
- **Issue:** n/a
- **Depedencies:** none


libs/community/langchain_community/graph_vectorstores/extractors/link_extractor_transformer.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, link_extractors: Iterable[LinkExtractor[Document]]):
+        """Create a DocumentTransformer which adds extracted links to each document."""
+        self.link_extractors = link_extractors
+
+    def transform_documents(
+        self, documents: Sequence[Document], **kwargs: Any
+    ) -> Sequence[Document]:
+        # Implement `transform_docments` directly, so that LinkExtractors which operate
+        # better in batch (`extract_many`) get a chance to do so.
+        document_links = zip(
```
eyurtsev, line 33: Do you have examples where `extract_many` is useful in terms of performance or is this done in anticipation?

nit: There should be a way to rephrase this code to be easier to read without sacrificing performance. At least I find the double usage of `zip` and `*` to be hard to follow.

libs/community/langchain_community/graph_vectorstores/extractors/link_extractor_transformer.py
```diff
@@ -0,0 +1,44 @@
...
+            documents,
+            zip(
+                *[
+                    extractor.extract_many(documents)
+                    for extractor in self.link_extractors
+                ]
+            ),
+        )
+        for document, links in document_links:
+            add_links(document, *links)
```
eyurtsev, line 43: LinkExtractorTransformer violates standard conventions in pythons of having no return type if the input is being mutated.

Users will generally not expect the input documents to be mutated as a result of LinkExtractorTransformer.

You could make this more explicit by adding an opt-in  parameter in the init to do in place mutation

libs/community/langchain_community/graph_vectorstores/extractors/link_extractor.py
```diff
@@ -14,7 +14,7 @@ class LinkExtractor(ABC, Generic[InputT]):
     """Interface for extracting links (incoming, outgoing, bidirectional)."""
 
     @abstractmethod
-    def extract_one(self, input: InputT) -> set[Link]:  # noqa: A002
+    def extract_one(self, input: InputT) -> Set[Link]:
```
eyurtsev, line 17: Why does the API use a Set vs. a Link?

1) Will all links be hashable?
2) Is order of appearance of results never valuable?

eyurtsev (approved):

eyurtsev:
looks good -- pending questions from author. It would be good to fix up the mutation of content in place since that will likely lead to bugs in users code. We can do a shallow copy of the metadata dict if we're just mutating the root of the namespace, and that shouldn't have too much of a performance consequence

libs/community/langchain_community/graph_vectorstores/extractors/link_extractor_transformer.py
bjchambers: Yeah -- I'll send some PRs in the near future, but things like Keybert use a model to extract keywords. They are able to batch the application of the model across many documents, so it is much more efficient.

https://colab.research.google.com/drive/174MxzfqwQqHeOIuGTW0_E6WFGvbT3krf#scrollTo=BLdNC65TCju6

On 100 random strings of length 8:

```
Map:   3.4769s
Batch:  0.9578s
```

libs/community/langchain_community/graph_vectorstores/extractors/link_extractor_transformer.py
bjchambers: I think it's probably better to skip the parameter and just duplicate things. Although, it's somewhat inconsistent within the codebase:

1. I based this on `doctran_text_extract` which mutates the documents in place:

https://github.com/langchain-ai/langchain/blob/cacdf96f9c2a608dfb6c69ca07a585d7cad7786d/libs/community/langchain_community/document_transformers/doctran_text_extract.py#L113

Also, beautiful_soup_transformer:

https://github.com/langchain-ai/langchain/blob/cacdf96f9c2a608dfb6c69ca07a585d7cad7786d/libs/community/langchain_community/document_transformers/beautiful_soup_transformer.py#L76

2. I see many things do what `google_translate` does, which would create a new document (and new metadata):

https://github.com/langchain-ai/langchain/blob/cacdf96f9c2a608dfb6c69ca07a585d7cad7786d/libs/community/langchain_community/document_transformers/google_translate.py#L104

3. Many of the others are somewhere in the middle -- making a new document but passing in the original metadata (so future mutations would affect both documents) or creating a shallow copy of metadata (any collections within the metadata would be mutated).

I think I'll change this to follow the google translate pattern (shallow copy of metadata) since that seems more expected (in general), although it seems like the document transformers are in a generally confusing state.

libs/community/langchain_community/graph_vectorstores/extractors/link_extractor.py
bjchambers: 1. All links are hashable
2. Order of appearance is not valuable in the graph -- a node either has a link or it doesn't.

I could see a case for `List` (preserving the order of hyperlinks in an HTML document, for instance), but the order wouldn't have any affect on traversal (eg., once you write the links into a DB and get them back, the order would be lost).

Basically -- links induce edges, and an edge exist (is present) or not.

eyurtsev (approved):

eyurtsev (approved):

libs/core/langchain_core/graph_vectorstores/links.py
```diff
@@ -66,3 +66,30 @@ def add_links(doc: Document, *links: Union[Link, Iterable[Link]]) -> None:
...
+        doc: The document to add the links to.
+        *links: The links to add to the document.
+
+    Returns:
+        A document with a shallow-copy of the metadata with the links added.
+    """
+    new_links = set(get_links(doc))
+    for link in links:
+        if isinstance(link, Iterable):
+            print(f"{link!r}")
```
eyurtsev, line 84:
```suggestion
```

libs/community/langchain_community/graph_vectorstores/extractors/__init__.py
```diff
@@ -6,20 +6,24 @@
...
-    HtmlInput,
-    HtmlLinkExtractor,
-)
 from langchain_community.graph_vectorstores.extractors.keybert_link_extractor import (
     KeybertInput,
     KeybertLinkExtractor,
 )
-from langchain_community.graph_vectorstores.extractors.link_extractor import (
+
+from .html_link_extractor import (
```
cbornet, line 14: It seems absolute imports are preferred in the codebase rather than relative ones ? (although we can find some relative imports here and there).
@eyurtsev ?

eyurtsev: Yes, generally we use absolute imports rather than explicit relative. Absolute imports can be ambiguous in some cases, but it tends to be easier to work with them w/ respect to refactors

libs/community/langchain_community/graph_vectorstores/extractors/link_extractor_transformer.py
```diff
@@ -0,0 +1,43 @@
...
+    Example:
+        .. code-block:: python
+
+            extract_links = LinkExtractorTransformer([
+                HtmlLinkExtractor().as_document_extractor(),
+            ])
+            extract_links.transform_documents(docs)
+    """
+
+    def __init__(self, link_extractors: Iterable[LinkExtractor[Document]]):
```
eyurtsev, line 24: Should probably be a Sequence in this case? i.e., length is known and one can materialize it more than once?

```suggestion
    def __init__(self, link_extractors: Sequence[LinkExtractor[Document]]):
```

eyurtsev (approved):
Feel free to make a PR for remaining updates -- they're not blocking

PR #24195, opened 2024-07-12
core[patch]: Mark GraphVectorStore as beta

eyurtsev:
* This PR marks graph vectorstore as beta.

### DIFF
--- a/libs/core/langchain_core/graph_vectorstores/base.py
+++ b/libs/core/langchain_core/graph_vectorstores/base.py
@@ -12,6 +12,7 @@ from typing import (
     Optional,
 )
 
+from langchain_core._api import beta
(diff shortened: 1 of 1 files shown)


bjchambers (approved):

eyurtsev:
@cbornet proposed `GraphNode` and `GraphLink` on the other PR. Maybe we'll keep as Node and Link for now.

FYI there's another similar abstraction from neo4j 

https://github.com/langchain-ai/langchain/blob/b3cf1ae842ed4803cfdc073e6d4fe5e0a8984884/libs/community/langchain_community/graphs/graph_document.py#L40-L40

It might be good to figure out how similar / different things are.

PR #24206, opened 2024-07-12
core[minor]: Add BaseIndex abstraction

eyurtsev:
This PR introduces the `BaseIndex` abstraction. 

## Context

We're creating a base index abstraction that defines a standard interface w/ respect to [BaseMedia](https://github.com/langchain-ai/langchain/blob/dc4797acc3a7e958d261d0b9ceeaa95e5b2cee4b/libs/core/langchain_core/documents/base.py#L15-L15), which is content that has an ID and metadata.

The interface is designed to support the following operations:

1. Storing content in the index.
2. Retrieving content by ID.
3. Querying the content based on the metadata associated with the content.

The interface does **NOT** support queries against the actual content of the BaseMedia -- this is left for more specialized interfaces like a vectorstore.

Goals:

1. Define precise semantics (e.g., upsert vs. add), no exception for get_by_ids when content is missing etc.
2. Add functionality which is useful for administration (e.g., get_by_ids, delete_by_filter etc.)
3. BaseIndex can support binary data via MediaBlob
4. Define a standard interface for filtering
5. Create a standard way for vectorstores to describe what functionality they support (`describe` classmethod)

We're creating a separate interface rather than updating VectorStore to allow other LangChain classes to also implement the same indexing API ([ParentDocumentRetriever](https://github.com/langchain-ai/langchain/blob/dc4797acc3a7e958d261d0b9ceeaa95e5b2cee4b/libs/langchain/langchain/retrievers/parent_document_retriever.py#L10-L10))

Existing vectorstores are largely unaffected will inherit from BaseIndex without any modifcations, raising NotImplementedErrors for the new methods.

A test suite will be provided to test index functionality based on information in the `describe()` classmethod (which will cover vectorstores as well)

## TODOs

- [ ] Decide name `BaseIndex` vs. `Index` (using `BaseIndex` to avoid collision with the `index` function)
- [ ] Confirm return type of `delete` API
- [ ] Confirm location of abstraction and names of methods / variables
- [ ] Add missing async methods


libs/core/langchain_core/indexing/base.py
```diff
@@ -421,29 +827,3 @@ async def adelete_keys(self, keys: Sequence[str]) -> None:
             keys: A list of keys to delete.
         """
         self.delete_keys(keys)
-
-
-class UpsertResponse(TypedDict):
```
eyurtsev, line 426: moved to top of file

libs/core/langchain_core/vectorstores/base.py
```diff
@@ -565,6 +516,54 @@ async def asearch(
...
+            successfully deleted from the vectorstore and the list of IDs that failed
+            to be deleted.
+
+        .. versionadded:: 0.2.15
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not yet support delete_by_filter."
+        )
+
+    def get_by_filter(
```
bjchambers, line 542: Is there a way to filter and order by similarity? Would that be a `get_by_filter` with a `sort` based on the similarity? Or is that a similarity search with a filter? Something else?

The above makes me wonder if similarity search could be expressed as a sort, in which case basic similarity search is `filter = None, sort = Similarity`?

eyurtsev: I was thinking about introducing another method to capture search semantics. 


```python
def get_by_query(
        self,
        *,
        query: Q, # <-- e.g., text query or vector or BaseMedia (up to the implementation)
        filter: Optional[Union[Dict[str, Any], List[Dict[str, Any]]]] = None,
        limit: Optional[int] = None,
        page: int = 0, # <-- Pagination added
        sort: Optional[Sort] = None,
        **kwargs,
) -> List[Hit]:  # <-- A hit would contain information like score, optional content etc.
```

----

i agree that sort and similarity can be viewed as the same thing. I need to think a bit more to see if we can merge them.

bjchambers:
I could see a case for filtering on mime type too. Maybe possible via the
filter here, but thought I’d toss that out.

libs/core/langchain_core/vectorstores/base.py
eyurtsev: @bjchambers Would standardization of filters, the ability to query by vector and the ability to retrieve vectors be sufficient to implement the search traversal patterns needed for GraphVectorStore?

At indexing time, the implementation can standardized fields into metadata; e.g., `namespace` (e.g., edge, node),  `edge_type` (just for edges), `edge_data` (e.g., the value of an href).

At query time, the implementation can rely on the ability to query by vector and filter on the standardized fields. Is that sufficient to implement GraphVectorStore? Is there a way to limit traversal by depth correctly (to avoid re-exploring nodes?)

bjchambers: Also "query-by-ID" to get all the nodes with the selected IDs, but that could be part of the above. Currently, it does the traversal with just the IDs and then fetches those IDs. So maybe also:

1. `query-ids-by-vector`
2. `query-ids-by-filter`
3. `query-ids-by-vector-and-filter`
4. `query-by-ids`

The logic being that each node may be reached many times during the traversal -- rather than pulling back all the data each time, it is likely better to just pull back the data necessary for traversing. That is currently (i) the ID used for identifying the reachable nodes, (ii) the embedding (used for the MMR score calculation), and (iii) the outgoing tags (used for finding the next set of edges). So maybe a version that accepts a list of which fields to actually fetch (eg., we would only need the ID, the metadata["outgoing"], and the embedding). But that could perhaps be a future optimization...

libs/core/langchain_core/indexing/__init__.py
```diff
@@ -7,13 +7,15 @@
...
 from langchain_core.indexing.base import (
+    BaseIndex,
     InMemoryRecordManager,
     RecordManager,
     UpsertResponse,
 )
 
 __all__ = [
     "aindex",
+    "BaseIndex",
```
kurtisvg, line 18: note: I find the "Index" part of this name confusing -- this might be because Index's in databases/programming are a different thing. Something like "Store" or "Collection" would be more intuitive to me.

libs/core/langchain_core/indexing/api.py
```diff
@@ -265,6 +268,9 @@ def index(
             "delete" and "add_documents" required methods.
         ValueError: If source_id_key is not None, but is not a string or callable.
     """
+    # Local scope to avoid circular imports
+    from langchain_core.vectorstores import VectorStore
```
kurtisvg, line 272: As a rule of thumb, circular imports tend to cause subtle problems are are usually best avoided all together.

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,414 @@
...
+
+    If missing, the default sort order is ascending.
+    """
+
+
+# The list of supported filter operations will need to be updated as new standard
+# operators are added to the filtering syntax.
+FilterOp = Literal[
+    "$eq", "$ne", "$lt", "$lte", "$gt", "$gte", "$in", "$nin", "$and", "$not", "$or"
+]
```
kurtisvg, line 46: Any reason not to use an Enum here?

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,414 @@
...
+
+from typing_extensions import NotRequired, TypedDict
+
+from langchain_core._api import beta
+from langchain_core.documents.base import BaseMedia
+from langchain_core.runnables import run_in_executor
+from langchain_core.utils import abatch_iterate, batch_iterate
+
+
+class Sort(TypedDict):
```
kurtisvg, line 30: nit: I think in general, [Dataclasses](https://docs.python.org/3/library/dataclasses.html) are recommended over TypedDict. Any reason to prefer a TypedDict here?

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,414 @@
...
+            yield self.upsert(item_batch, **kwargs)
+
+    @abc.abstractmethod
+    @beta(message="Added in 0.2.15. The API is subject to change.")
+    def upsert(self, items: Sequence[T], /, **kwargs: Any) -> UpsertResponse:
+        """Upsert items into the index.
+
+        The upsert functionality should utilize the ID field of the content object
+        if it is provided. If the ID is not provided, the upsert method is free
+        to generate an ID for the content.
```
kurtisvg, line 173: Can you provide more strict guidance here? It's common to stick to [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119) terms.

Specifically -- is this "SHOULD" (recommended) or "MAY" (Optional) behavior?

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,414 @@
...
+
+        When an ID is specified and the content already exists in the vectorstore,
+        the upsert method should update the content with the new data. If the content
```
kurtisvg, line 176: Is there a way to insert without upsert?

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,414 @@
...
+        Returns:
+            UpsertResponse: A response object that contains the list of IDs that were
+            successfully added or updated in the vectorstore and the list of IDs that
+            failed to be added or updated.
+
+        .. versionadded:: 0.2.15
+        """
+
+    @beta(message="Added in 0.2.11. The API is subject to change.")
+    async def astreaming_upsert(
```
kurtisvg, line 192: should this have the same warning as `streaming_upsert`?

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,414 @@
...
+            ids: List of ids to delete.
+            kwargs: Additional keyword arguments. This is up to the implementation.
+
+        Returns:
+            DeleteResponse: A response object that contains the list of IDs that were
+            successfully deleted and the list of IDs that failed to be deleted.
+
+            OR
+
+            bool: A boolean indicating whether the delete operation was successful.
```
kurtisvg, line 288: What about partial successes? Should that be "true" or "false" is some items were removed?

libs/core/langchain_core/vectorstores/base.py
```diff
@@ -288,8 +235,10 @@ def embeddings(self) -> Optional[Embeddings]:
         )
         return None
 
-    def delete(self, ids: Optional[List[str]] = None, **kwargs: Any) -> Optional[bool]:
-        """Delete by vector ID or other criteria.
+    def delete(
+        self, ids: Optional[List[str]] = None, **kwargs: Any
+    ) -> Union[DeleteResponse, None, bool]:
```
kurtisvg, line 240: In general, I would prefer introducing a new method with a different return as opposed to trying to overload this one. Otherwise don't get what I want from type checking here -- I have to do extra work to sus out the type. This also makes it easier to deprecate the function in the future.

libs/core/langchain_core/indexing/base.py
eyurtsev: Do you think it's necessary to be able to add without specifically having upsert semantics? At the moment, docs have an optional ID associated with them, so the upsert works like an add in that case.

kurtisvg: From a relational database perspective, the conflict depends on what the primary key is. Is the assumption always the the ID is a primary key, or can you have primary keys on different fields? I might want to add a new entry, but not overwrite if their is an existing one.

PR #24209, opened 2024-07-12
standard-tests[patch]: check for ToolCall["type"]

baskaryan:


eyurtsev (approved):

mdrxy:
@eyurtsev @baskaryan are there generally any reasons beyond deserialization for including an explicit `type` literal in messages (e.g. `ToolCall`)? Poking around the code base and this piqued my curiosity.

PR #24254, opened 2024-07-15
community: add bm25s integration

mhattingpete:
**Description:** Added BM25S integration to langchain community to provide a much faster version of the BM25 package
**Issue:** NA
**Dependencies:** NA
**Twitter handle:** NA (Use GitHub instead)


libs/community/langchain_community/retrievers/bm25s.py
```diff
@@ -0,0 +1,137 @@
...
+from langchain_core.pydantic_v1 import Field
+from langchain_core.retrievers import BaseRetriever
+
+
+class BM25SRetriever(BaseRetriever):
+    """`BM25S` retriever."""
+
+    retriever: Any
+    """ BM25 retriever."""
+    docs: List[Document] = Field(repr=False)
```
eyurtsev, line 16: Why is docs an attribute on the retriever? Why is it needed after the index has been built?

eyurtsev, line 16: A copy of docs should not be retained after the index is created

libs/community/langchain_community/retrievers/bm25s.py
```diff
@@ -0,0 +1,137 @@
...
+    ) -> List[Document]:
+        try:
+            import bm25s
+        except ImportError:
+            raise ImportError(
+                "Could not import bm25s, please install with `pip install "
+                "bm25s`."
+            )
+        queries_tokenized = bm25s.tokenize(query, stopwords=self.stopwords, stemmer=self.stemmer)
+        return_docs = self.retriever.retrieve(queries_tokenized, k=self.k, n_threads=self.n_threads, chunksize=self.chunksize)
```
eyurtsev, line 135: The retriever is being used like a vectorstore. Why is it not implemented as a vectorstore?

eyurtsev (changes requested):
Hi @mhattingpete  I am not sure that exposing this as a retriever makes sense -- it looks more like a vectorstore to me that supports indexing content and then retrieving content from it.

Consider re-implementing as a vectorstore that has access to a BaseStore where it can store the underlying documents.

libs/community/langchain_community/retrievers/bm25s.py
mhattingpete: I can change it such that it is not required by building the return_docs instead. But this requires having the documents stored in BM25S as a corpus instead. Currently the ID's is returned by BM25S but it can be changed to return the documents instead.

mhattingpete: Same comment as above

libs/community/langchain_community/retrievers/bm25s.py
mhattingpete: It is implemented in this way as that is how the original BM25 is implemented. I have tried keeping the alike as much as possible. However if you prefer I can definitely change this to a vectorstore

mhattingpete:
Hi @eyurtsev. Thanks for the feedback. It is implemented in this way as that is how the original BM25 is implemented. I have tried keeping the alike as much as possible. However if you prefer I can definitely change this to a vectorstore

eyurtsev:
@mhattingpete maybe a vectorstore isn't needed here. I assume you're looking for an interface that allows you to add documents into an in memory store and then search through it using BM25?

eyurtsev:
Does the bm25s implementation support updating an index with additional documents after its creation?

mhattingpete:
> Does the bm25s implementation support updating an index with additional documents after its creation?

I will investigate this. If so I think it would be better suited for a vectorstore as you mention :)

eyurtsev:
@mhattingpete Could you check whether it supports addition of content but don't change this into a vectorstore yet -- we're considering adding an indexing interface that would allow implementing this without having to implement the rest of the vectorstore methods

mhattingpete:
> @mhattingpete Could you check whether it supports addition of content but don't change this into a vectorstore yet -- we're considering adding an indexing interface that would allow implementing this without having to implement the rest of the vectorstore methods

I have now done the investigation. It doesn't seem to support adding more content after the initial build without a complete rebuild.

libs/community/langchain_community/retrievers/bm25s.py
mhattingpete: Closed according to latests comments

mhattingpete:
Hi @eyurtsev. Did you have time to take a second look at this?

eyurtsev:
> I have now done the investigation. It doesn't seem to support adding more content after the initial build without a complete rebuild.

We're in the process of adding a simple indexer abstraction, but it assumes that the index can be updated after the fact (since that's what's needed for most production applications). Give me a bit of time to think about how to handle this use case.

mhattingpete:
Hi @eyurtsev, any news on this?

eyurtsev:
@mhattingpete Apologies -- not yet. We're still discussing internally on how to handle the indexing interface, and on whether we want to be accommodating one shot indexers.

There's also an existing BM25 retriever (https://python.langchain.com/v0.2/docs/integrations/retrievers/bm25/). It's unclear to me the benefit of merging another BM25 retriever if it's a one shot indexer (even if it's faster). One shot indexers aren't really usable in a production setting, and are mostly used for prototyping / notebook environments.

What's your use case?

xhluca:
Wanted to quickly chime in regarding the discussion, hopefully I can clarify a few things:
* BM25S is not designed to be a vector store/DB and does not support incremental updates (since we are computing the relevance score during document indexing, and adding new documents would change the relevance score of previous document tokens).
* BM25S works the same way as rank-bm25, bm25-pt, etc. It is up to 500x faster than rank-bm25 and [comparable to PISA for smaller corpora](https://github.com/xhluca/bm25-benchmarks) (although PISA is really good for larger ones). The only required dependencies are Numpy (same as rank-bm25) and Scipy.
* Rank-BM25 supports 3 variants of BM25 (although [BM25L's algorithm received a patch](https://github.com/dorianbrown/rank_bm25/pull/20) that is not in the pypi version yet). BM25S supports 5 (although more variants could be created by changing idf_method).
* Both are bring-your-own tokenizer, though BM25S is packaged with a Python tokenizer with regex splitting, stopwords and allow stemming (if PyStemmer is installed).

No need to replace rank-bm25 if it works well for the users! BM25S was mainly designed for a lightweight but fast lexical retrieval, but users might not see benefits if there are <5000 documents in a corpus. Rank-bm25 might make sense if users want to change k1, b and delta at retrieval time.

mhattingpete:
Hi @eyurtsev and @xhluca.

Thanks for your responses. Currently I'm using the BM25retriever in a hybrid retrieval for a RAG application. I have close to 10K+ documents and therefore the current BM25 implementation causes a headache.

I don't agree though that indexers is not relevant for production setting as often many companies don't add new documents that is relevant for the application on a daily basis. Thereby not requiring a reindexing all the time. Additionally combining sparse retrievers with dense retrievers heavily improves the relevance of the documents for many use cases.

PR #24258, opened 2024-07-15
community[minor]: Add ability to specify file encoding and json encoding for FileChatMessageHistory

yabooung:
Description:
Add UTF-8 encoding support

Issue:
Inability to properly handle characters from certain languages (e.g., Korean)

Fix:
Implement UTF-8 encoding in FileChatMessageHistory


libs/community/langchain_community/chat_message_histories/file.py
```diff
@@ -20,21 +20,25 @@ def __init__(self, file_path: str) -> None:
         self.file_path = Path(file_path)
         if not self.file_path.exists():
             self.file_path.touch()
-            self.file_path.write_text(json.dumps([]))
```
eyurtsev, line 23: write_text accepts an encoding parameter

write_text(json.dumps([]), encoding='utf-8')

libs/community/langchain_community/chat_message_histories/file.py
```diff
@@ -20,21 +20,25 @@ def __init__(self, file_path: str) -> None:
...
+            with self.file_path.open("w", encoding="utf-8") as f:
```
eyurtsev, line 23: To add the options in a non breaking way, you'll need to add encoding and ensure ascii as part of the initializer.

If we set the default encoding to utf-8, we could break existing user message history (e.g., if their default encoding is a windows encoding)

eyurtsev (changes requested):
Thanks for the contribution! To merge, we need to expose the options via the initializer with defaults matching the previous defaults (encoding=None and ensure_ascii=True).

libs/community/langchain_community/chat_message_histories/file.py
yabooung: Thanks for the advice

Undertone0809 (approved):

libs/community/langchain_community/chat_message_histories/file.py
```diff
@@ -11,30 +11,48 @@
 class FileChatMessageHistory(BaseChatMessageHistory):
     """Chat message history that stores history in a local file."""
 
-    def __init__(self, file_path: str) -> None:
+    def __init__(
+        self,
+        file_path: str,
```
eyurtsev, line 16:
```suggestion
        file_path: str,
        *,
```

eyurtsev (approved):

eyurtsev (approved):

PR #24270, opened 2024-07-15
community[patch]: Update TavilySearch to use TavilyClient instead of the deprecated Client

2020uce0047:
On using TavilySearchAPIRetriever with any conversation chain getting error : 

`TypeError: Client.__init__() got an unexpected keyword argument 'api_key'`

It is because the retreiver class is using the depreciated `Client` class, `TavilyClient` need to be used instead.

### DIFF
--- a/libs/community/langchain_community/retrievers/tavily_search_api.py
+++ b/libs/community/langchain_community/retrievers/tavily_search_api.py
@@ -31,7 +31,11 @@ class TavilySearchAPIRetriever(BaseRetriever):
         self, query: str, *, run_manager: CallbackManagerForRetrieverRun
     ) -> List[Document]:
         try:
-            from tavily import TavilyClient
+            try:
+                from tavily import TavilyClient
+            except TypeError:
+                # Older of tavily used Client
+                from tavily import Client as TavilyClient
         except ImportError:
             raise ImportError(
                 "Tavily python package not found. "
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/retrievers/tavily_search_api.py
```diff
@@ -31,14 +31,18 @@ def _get_relevant_documents(
         self, query: str, *, run_manager: CallbackManagerForRetrieverRun
     ) -> List[Document]:
         try:
-            from tavily import Client
+            try:
+                from tavily import TavilyClient
+            except TypeError:
+                # Older of tavily used Client
+                from tavily import Client as TavilyClient
```
eyurtsev, line 38: @2020uce0047 added path for compatibility with older clients. Could you confirm that this makes sense?

eyurtsev (approved):

libs/community/langchain_community/retrievers/tavily_search_api.py
```diff
@@ -31,14 +31,18 @@ def _get_relevant_documents(
         self, query: str, *, run_manager: CallbackManagerForRetrieverRun
     ) -> List[Document]:
         try:
-            from tavily import Client
+            try:
+                from tavily import TavilyClient
+            except TypeError:
```
eyurtsev, line 36:
```suggestion
            except ImportError:
```

PR #24280, opened 2024-07-15
openai: remove some params from default serialization

efriis:


baskaryan (approved):

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -455,26 +455,27 @@ def validate_environment(cls, values: Dict) -> Dict:
...
+            "model": self.model_name,
+            "stream": self.streaming,
+            "n": self.n,
+            "temperature": self.temperature,
+            **{k: v for k, v in exclude_if_none.items() if v is not None},
             **self.model_kwargs,
         }
-        if self.max_tokens is not None:
-            params["max_tokens"] = self.max_tokens
-        if self.stop:
```
baskaryan, line 474: nit: stop is excluded if its falsey. is default value for stop None or empty list?

efriis: I think this was handled by the `self.stop or None` above

PR #24306, opened 2024-07-16
community: added stream and astream to chatyandexgpt

olgamurraft:
**Description**
added stream and astream support to ChatYandexGPT


libs/community/langchain_community/chat_models/yandex.py
```diff
@@ -115,16 +116,43 @@ async def _agenerate(
...
             ValueError: if the last message in the list is not from human.
         """
-        text = await acompletion_with_retry(self, messages=messages)
-        text = text if stop is None else enforce_stop_tokens(text, stop)
-        message = AIMessage(content=text)
-        return ChatResult(generations=[ChatGeneration(message=message)])
+        async for text in acompletion_with_retry(self, messages=messages):
+            text = text if stop is None else enforce_stop_tokens(text, stop)
+            message = AIMessage(content=text)
+            return ChatResult(generations=[ChatGeneration(message=message)])
```
ccurme, line 122: This appears to make things slightly more confusing (iterating through something that is implicitly length 1), maybe in the name of slightly reducing duplicated code. I'd be in favor of having separate completion with retry functions for stream / invoke.

ccurme (commented):
Missing type hints.

PR #24342, opened 2024-07-17
experimental: add max_chunk_size to SemanticChunker

RafaelXokito:
**Description:** 
This PR adds a `max_chunk_size` parameter to the `SemanticChunker` class. The `max_chunk_size` ensures that no chunk exceeds the specified size, splitting larger chunks accordingly. This feature enhances the chunking process by maintaining manageable chunk sizes.

**Issue:** 
Fixes #18014

### DIFF
--- a/libs/experimental/langchain_experimental/cpal/base.py
+++ b/libs/experimental/langchain_experimental/cpal/base.py
@@ -41,9 +41,9 @@ class _BaseStoryElementChain(Chain):
     chain: LLMChain
     input_key: str = Constant.narrative_input.value  #: :meta private:
     output_key: str = Constant.chain_answer.value  #: :meta private:
-    pydantic_model: ClassVar[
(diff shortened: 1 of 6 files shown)


SatyamMattoo:
It will be really helpful if this feature gets added. For now, the Semantic Chunker returns very large chunks in some cases and the only way to limit that is creating a custom class I guess?

gustavz:
Can we please get this in @hwchase17 ? thanks!

repoofideas:
Waiting for this feature to be implemented.

efriis:

regarding max_chunk_size, wouldn't it be more effective to just pass the output of semanticchunker to one of the other text splitters that follows a strict chunk-size strategy? that way the user can decide which strategy to use to keep the chunks below a given size! this pr would be prescriptive if I understand correctly

PR #24356, opened 2024-07-17
couchbase: Add chat message history

nithishr:
**Description:** : Add support for chat message history using Couchbase


libs/partners/couchbase/pyproject.toml
```diff
@@ -54,6 +54,7 @@ pytest = "^7.4.3"
 pytest-asyncio = "^0.23.2"
 pytest-socket = "^0.7.0"
 syrupy = "^4.0.2"
+langchain = "^0.2.9"
```
efriis, line 57: why is `langchain` and not just `langchain-core` needed as a dependency of langchain-couchbase?

nithishr: `langchain` is a dependency only for testing.  This was required for the `ConversationBufferMemory` which is part of langchain.
Do you think I should handle it differently? Or test differently?

efriis: my mistake - thanks for clarifying!

efriis (approved):

PR #24360, opened 2024-07-17
core[patch]: Fix regression requiring input_variables in few chat prompt templates

eyurtsev:
* Fix regression that requires users passing input_variables=[].

* Regression introduced by my own changes to this PR: https://github.com/langchain-ai/langchain/pull/22851

### DIFF
--- a/libs/core/langchain_core/prompts/few_shot.py
+++ b/libs/core/langchain_core/prompts/few_shot.py
@@ -18,7 +18,7 @@ from langchain_core.prompts.string import (
     check_valid_template,
     get_template_variables,
 )
-from langchain_core.pydantic_v1 import BaseModel, Extra, root_validator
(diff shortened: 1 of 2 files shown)


ccurme (approved):

AlSaltAlBahri:
it looks like this fix has caused introduced an issue. I was running langchain-core 2.20.0 and just upgraded to 2.21.0 today and while running an agent it is returning 

```   
   agent = create_tool_calling_agent(LLM, tools, prompt)  
    agent_executor = AgentExecutor(
    agent=agent,
    tools=tools,
    verbose=True,
    handle_parsing_errors=True,
    return_intermediate_steps=False,
    include_run_info=True,
    memory=memory)
```

is not returning `"'FieldInfo' object has no attribute 'required'"`

PR #24364, opened 2024-07-17
core[minor]: Add minimal document indexer

eyurtsev:
* This PR adds a minimal document indexer abstraction.
* The abstraction comes with a test suite that can verify that the indexer implements the correct semantics.
* This abstraction separates the sync and async interfaces.

* At a later date, we can extend this abstraction with `class Indexer(Generic[T])` where T is bound to BaseMedia, and redefine document indexer as an instance of Index `class DocumentIndexer(Indexer[Document])`


libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,249 @@
...
+
+
+class DeleteResponse(TypedDict, total=False):
+    """A generic response for delete operation.
+
+    The fields in this response are optional and whether the vectorstore
+    returns them or not is up to the implementation.
+    """
+
+    num_deleted: int
```
nfcampos, line 54: why do we need counts in delete response but not upsert ?

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,249 @@
...
+    operation for content that can be upserted by ID.
+
+    Upsert APIs that accept inputs with IDs and generate IDs internally
+    will return a response that includes the IDs that succeeded and the IDs
+    that failed.
+
+    If there are no failures, the failed list will be empty, and the order
+    of the IDs in the succeeded list will match the order of the input documents.
+
+    If there are failures, the response becomes ill defined, and a user of the API
```
eyurtsev, line 34: cc @nfcampos -- ill defined semantics on failure if IDs are not provided. I think this is OK, but we can change failed to be `failed_index` and match the index of the input documents (or do something else)

libs/core/langchain_core/indexing/base.py
eyurtsev: 1) It's not required a required field for the delete response (total=False)
2) on the upsert, we can add `num_upserted` as non required, but it's easy to derive from "succeeded" field which is a required field

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,264 @@
...
+    succeeded: Sequence[str]
+    """The IDs that were successfully deleted.
+    
+    If returned, this should only include *actual* deletions.
+    
+    If the ID did not exist to begin with,
+    it should not be included in this list.
+    """
+
+    failed: Sequence[str]
```
eyurtsev, line 72: We could delete the failed fields. They're not required as it is, but not super clear whether they'll be used or not.

The only usage I could forsee and would make sense is on a streaming upsert if an entire batch fails even after retries (thought maybe there will be a better parameterization).

An alternative is to introduce a `not_found` equivalent, but I expect that it just doesn't make sense to go through the trouble of computing most of the time (will likely encourage implementations to add latency)

eyurtsev:
My ability to lint is beyond ridiculous

libs/core/langchain_core/indexing/base.py
```diff
@@ -1,8 +1,263 @@
...
+    
+    Please note that deleting an ID that 
+    does not exist is **NOT** considered a failure.
+    """
+
+    num_failed: int
+    """The number of items that failed to be deleted."""
+
+
+@beta(message="Added in ___version___. The API is subject to change.")
```
efriis, line 82: fill this out

libs/core/langchain_core/indexing/in_memory.py
```diff
@@ -0,0 +1,86 @@
...
+        found_documents = []
+
+        for id_ in ids:
+            if id_ in self.store:
+                found_documents.append(self.store[id_])
+
+        return found_documents
+
+
+class AsyncInMemoryDocumentIndexer(AsyncDocumentIndexer):
```
efriis, line 65: just confirming - doing these separately or modifying to combine as a single?

I'm good with either. Moderately in favor of keeping single implementations to match the rest of them for now, and splitting everything altogether if we decide that's the path forward.

efriis (approved):
comfortable merging as beta. Can work with some partners to implement while in beta to get feedback?

I think last we chatted we were planning on merging sync/async base class into one to match others. Comfortable with what you think is best.

libs/core/langchain_core/indexing/base.py
eyurtsev: It was left as a template to be updated right prior to merging

libs/core/langchain_core/indexing/in_memory.py
eyurtsev: Combining into one because we're subclassing from retriever now which has the interfaces merged

eyurtsev:
Closed in favor of: https://github.com/langchain-ai/langchain/pull/25062/files

PR #24379, opened 2024-07-18
experimental[patch]: block a few more things from PALValidator

eyurtsev:
* Please see security warning already in existing class.
* The approach here is fundamentally insecure as it's relying on a block
  approach rather than an approach based on only running allowed nodes.
  So users should only use this code if its running from a properly sandboxed
  environment.

### DIFF
--- a/libs/experimental/langchain_experimental/pal_chain/base.py
+++ b/libs/experimental/langchain_experimental/pal_chain/base.py
@@ -20,13 +20,21 @@ from langchain_experimental.pal_chain.math_prompt import MATH_PROMPT
 from langchain_experimental.pydantic_v1 import Extra, Field, root_validator
 from langchain_experimental.utilities import PythonREPL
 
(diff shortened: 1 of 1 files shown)


bengabay1994:
Looks, Good. I understand as you said that this approach is flawed "by design" so I guess as it is now it will always be potentially vulnerable to code execution. Thanks for the quick response :).

eyurtsev:
> Looks, Good. I understand as you said that this approach is flawed "by design" so I guess as it is now it will always be potentially vulnerable to code execution. Thanks for the quick response :).

Yes that's correct! The code has warnings to that effect and a forced opt-in https://github.com/langchain-ai/langchain/pull/24379/files#diff-6f31cb1c32471bc31a3a1d3cd14f201df827a30f72a5876a04bc7d46f1e59634R140

The only way to use it safely is to run it from a properly sandboxed environment.

PR #24391, opened 2024-07-18
community: Add stop sequence support for ExllamaV2-loaded models

ant-8:
### Description

This PR introduces stop sequence support for ExllamaV2-loaded LLMs through the following changes:

1. **Replace Generator**:
   - Replaced `ExLlamaV2BaseGenerator` with `ExLlamaV2DynamicGenerator`.
   - This change allows access to the `generate(...)` method, enabling the specification of `stop_conditions` that function equivalently to `stop_sequences`.

2. **Streaming Enhancements**:
   - Makes use of `generator.set_stop_conditions(...)`.

### Issue

N/A

### Dependencies

No additional dependencies required for this change.

### DIFF
--- a/libs/community/langchain_community/llms/exllamav2.py
+++ b/libs/community/langchain_community/llms/exllamav2.py
@@ -27,7 +27,6 @@ class ExLlamaV2(LLM):
     #TODO:
     - Add loras support
     - Add support for custom settings
-    - Add support for custom stop sequences
     """
 
     client: Any
@@ -78,7 +77,7 @@ class ExLlamaV2(LLM):
                 ExLlamaV2Tokenizer,
             )
             from exllamav2.generator import (
-                ExLlamaV2BaseGenerator,
+                ExLlamaV2DynamicGenerator,
                 ExLlamaV2StreamingGenerator,
             )
         except ImportError:
@@ -116,7 +115,7 @@ class ExLlamaV2(LLM):
         if values["streaming"]:
             generator = ExLlamaV2StreamingGenerator(model, exllama_cache, tokenizer)
         else:
-            generator = ExLlamaV2BaseGenerator(model, exllama_cache, tokenizer)
+            generator = ExLlamaV2DynamicGenerator(model, exllama_cache, tokenizer)
 
         # Configure the model and generator
         values["stop_sequences"] = [x.strip().lower() for x in values["stop_sequences"]]
@@ -153,6 +152,10 @@ class ExLlamaV2(LLM):
     ) -> str:
         generator = self.generator
 
+        if stop is None:
+            stop = []
+        stop_conditions = self.stop_sequences + stop
+
         if self.streaming:
             combined_text_output = ""
             for chunk in self._stream(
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/llms/exllamav2.py
```diff
@@ -116,7 +115,7 @@ def validate_environment(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         if values["streaming"]:
             generator = ExLlamaV2StreamingGenerator(model, exllama_cache, tokenizer)
         else:
-            generator = ExLlamaV2BaseGenerator(model, exllama_cache, tokenizer)
+            generator = ExLlamaV2DynamicGenerator(model, exllama_cache, tokenizer)
```
ccurme, line 118: `generator` is a public attribute. Could this result in breaking changes for users who are accessing `llm.generator`? I'm not familiar with ExLlama but t looks like `ExLlamaV2DynamicGenerator` does not inherit from `ExLlamaV2BaseGenerator`.

ant-8: You're right. `ExLlamaV2DynamicGenerator` is missing attributes (`sequence_ids` and `abort_event`) that are present in `ExLlamaV2BaseGenerator`.

The exllama author's [writeup](https://github.com/turboderp/exllamav2/blob/69cd2b1df8fb1399e9745b27852ebb4287e6b63d/doc/dynamic.md) indicates that `ExLlamaV2DynamicGenerator` is a more efficient implementation. Additionally, [examples](https://github.com/turboderp/exllamav2/tree/master/examples) in their repository that previously used `ExLlamaV2BaseGenerator` have been moved to a deprecated folder, and replaced with ones using `ExLlamaV2DynamicGenerator` for non-streaming generation.

Maybe we could introduce an optional flag to use the dynamic generator? Something like `use_dynamic_generator`

If so, we should issue warnings in the following cases:
- When `use_dynamic_generator=False` and `stop_sequences` is specified, as `ExLlamaV2BaseGenerator` does not support this functionality out of the box.
- When `use_dynamic_generator=True` and `streaming=True`, since streaming uses `ExLlamaV2StreamingGenerator`.

ant-8: > `generator` is a public attribute. Could this result in breaking changes for users who are accessing `llm.generator`? I'm not familiar with ExLlama but t looks like `ExLlamaV2DynamicGenerator` does not inherit from `ExLlamaV2BaseGenerator`.

You're right, `ExLlamaV2DynamicGenerator` is missing attributes (`sequence_ids` and `abort_event`) that are present in `ExLlamaV2BaseGenerator`.

The exllama author's [writeup](https://github.com/turboderp/exllamav2/blob/69cd2b1df8fb1399e9745b27852ebb4287e6b63d/doc/dynamic.md) indicates that `ExLlamaV2DynamicGenerator` is a more efficient implementation. Additionally, [examples](https://github.com/turboderp/exllamav2/tree/master/examples) in their repository that previously used `ExLlamaV2BaseGenerator` have been moved to a deprecated folder, and replaced with ones using `ExLlamaV2DynamicGenerator` for non-streaming generation.

Maybe we could introduce an optional flag to use the dynamic generator? Something like `use_dynamic_generator`

If so, we should issue warnings in the following cases:
- When `use_dynamic_generator=False` and `stop_sequences` is specified, as `ExLlamaV2BaseGenerator` does not support this functionality out of the box.
- When `use_dynamic_generator=True` and `streaming=True`, since streaming uses `ExLlamaV2StreamingGenerator`.

ccurme:
@ant-8 do you want to continue iterating on this?

ant-8:
@ccurme Couldn't find the time to come up with a better implementation. Closing

PR #24408, opened 2024-07-18
standard-tests[patch]: Add pytest assert rewrites

eyurtsev:
This will surface nice error messages in subclasses that fail assertions.

### DIFF
--- a/libs/core/tests/unit_tests/stores/test_in_memory.py
+++ b/libs/core/tests/unit_tests/stores/test_in_memory.py
@@ -1,11 +1,10 @@
 from typing import Tuple
 
 import pytest
-from langchain_standard_tests.integration_tests import (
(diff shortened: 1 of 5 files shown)


libs/core/tests/unit_tests/stores/test_in_memory.py
```diff
@@ -1,6 +1,36 @@
...
+    BaseStoreAsyncTests,
+    BaseStoreSyncTests,
+)
+from langchain_standard_tests.integration_tests.base_store import V
+
 from langchain_core.stores import InMemoryStore
 
 
+# Check against standard tests
+class TestSyncInMemoryStore(BaseStoreSyncTests):
```
eyurtsev, line 14: Added unit tests for in memory store and confirmed that the assertions work well

PR #24414, opened 2024-07-18
community: add ParquetLoader DocumentLoader

knustx:
- **Description:** Adds a new document loader to Community to load individual parquet files commonly used in "big data" systems
- **Issue:** N/A
- **Dependencies:** Pyarrow is required to load the parquet file and is only added as an integration test dependency
- **Twitter handle:** N/A

- [x] **Lint and test**: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified.


libs/community/pyproject.toml
```diff
@@ -100,6 +101,7 @@ anthropic = "^0.3.11"
 fireworks-ai = "^0.9.0"
 vdms = "^0.0.20"
 exllamav2 = "^0.0.18"
+pyarrow = "^17.0.0"
```
eyurtsev, line 104: Could you remove changes from pyproject toml and revert the poetry lock file? Instead you can the extended testing requirements file.

PR #24415, opened 2024-07-18
core: pydantic output parser streaming fix

efriis:


libs/langchain/tests/unit_tests/output_parsers/test_pydantic_parser.py
efriis, line 1: moved to core tests

libs/core/tests/unit_tests/output_parsers/test_pydantic_parser.py
```diff
@@ -96,3 +101,125 @@ def test_json_parser_chaining(
...
+        ]
+
+    parser = PydanticOutputParser(pydantic_object=Person)
+    chain = llm | parser
+    assert chain.invoke({}) == Person(name="King Louis XVI", age=38)
+    stream_result = list(chain.stream({}))
+    assert stream_result == [
+        Person(name="King Louis XVI", age=3),
+        Person(name="King Louis XVI", age=38),
+    ]
```
efriis, line 186: this is the new case

baskaryan, line 186: why isn't name streamed

mattf: lgtm

efriis: because before there's any age, it doesn't comply to pydantic schema (requires int age)

Issue #22782, opened 2024-06-11
with_structured_output format depends on whether we're using pydantic proper or pydantic.v1 namespace


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

```python

from pydantic.v1 import BaseModel
from pydantic import BaseModel as BaseModelV2

class Answer(BaseModel):
    answer: str


class Answer2(BaseModelV2):
    """"The answer."""
    answer: str


from langchain_openai import ChatOpenAI

model = ChatOpenAI()

model.with_structured_output(Answer).invoke('the answer is foo') # <-- Returns pydantic object

model.with_structured_output(Answer2).invoke('the answer is foo') # <--- Returns dict
```

eyurtsev:
There's a bug in the chat model when checking whether a pydantic model  (`from pydantic import BaseModel`) ends up being classified as not a pydantic model. 

But if we fix that logic, we get stuck here: 

https://github.com/langchain-ai/langchain/blob/232908a46d6371ff3d65381966ffe4831545cb25/libs/core/langchain_core/output_parsers/openai_tools.py#L177-L177

The code 

```
class PydanticToolsParser(JsonOutputToolsParser):
    """Parse tools from OpenAI response."""

    tools: List[Type[BaseModel]]
```


Which will not work since JsonOutputToolsParser inherits from BaseModel as well

hinthornw:
It also emits the wrong schema  in general for nested objects

PR #24417, opened 2024-07-18
core,groq,openai,mistralai,robocorp,fireworks,anthropic[patch]: Update BaseModel subclass and instance checks to handle both v1 and proper namespaces

baskaryan:
After this PR chat models will correctly handle pydantic 2 with bind_tools and with_structured_output.

```python
import pydantic
print(pydantic.__version__)
```
2.8.2

```python

class Add(BaseModel):
    x: int
    y: int

model = ChatOpenAI().bind_tools([Add])
print(model.invoke('2 + 5').tool_calls)

model = ChatOpenAI().with_structured_output(Add)
print(type(model.invoke('2 + 5')))
```

```
[{'name': 'Add', 'args': {'x': 2, 'y': 5}, 'id': 'call_PNUFa4pdfNOYXxIMHc6ps2Do', 'type': 'tool_call'}]
<class '__main__.Add'>
```

```python
from pydantic.v1 import BaseModel, Field

class Add(BaseModel):
    x: int
    y: int

model = ChatOpenAI().bind_tools([Add])
print(model.invoke('2 + 5').tool_calls)

model = ChatOpenAI().with_structured_output(Add)
print(type(model.invoke('2 + 5')))
```

```python
[{'name': 'Add', 'args': {'x': 2, 'y': 5}, 'id': 'call_hhiHYP441cp14TtrHKx3Upg0', 'type': 'tool_call'}]
<class '__main__.Add'>
```

Addresses issues: https://github.com/langchain-ai/langchain/issues/22782


libs/core/langchain_core/pydantic_v1/__init__.py
```diff
@@ -21,3 +22,25 @@
     _PYDANTIC_MAJOR_VERSION: int = int(metadata.version("pydantic").split(".")[0])
 except metadata.PackageNotFoundError:
     _PYDANTIC_MAJOR_VERSION = 0
+
+
+def _issubclass_base_model(type_: Type) -> bool:
```
eyurtsev, line 27: This shouldn't be in `pydantic_v1` -> can we place in langchain_core.utils.pydantic? 

pydantic_v1 is only the layer that checks that we're using the v1 namespace -- it's for the most part unaware of BaseModel from v2

eyurtsev: https://github.com/langchain-ai/langchain/pull/24418/files#diff-ff28020c5f1073a8b63bcd9d8b756a187fd682cb81935295120c63b207071896R23

eyurtsev:
@baskaryan thank you! I'll update this tomorrow. I think we can roll this out after we also get the standard tests for chat models updated and passing. Note that this **could** be a breaking change for some users if they've been relying on the functinality given the bug with with_structured_output and pydantic 2

eyurtsev:
Taking over

eyurtsev:
https://github.com/langchain-ai/langchain/issues/22782

eyurtsev (approved):

PR #24418, opened 2024-07-18
core[minor]: Support all versions of pydantic base model in argsschema

eyurtsev:
This adds support to any pydantic base model for tools.

The only potential issue is that `get_input_schema()` will not always return a v1 base model.

### DIFF
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -360,11 +360,6 @@ def create_schema_from_function(
         A pydantic model with the same arguments as the function.
     """
     # https://docs.pydantic.dev/latest/usage/validation_decorator/
-
-    from pydantic import (
-        validate_arguments,
-    )
-
     validated = validate_arguments(func, config=_SchemaConfig)  # type: ignore
     inferred_model = validated.model  # type: ignore
     filter_args = filter_args if filter_args is not None else FILTERED_ARGS
@@ -482,7 +477,7 @@ class ChildTool(BaseTool):
 
     def __init__(self, **kwargs: Any) -> None:
         """Initialize the tool."""
-        if "args_schema" in kwargs:
+        if "args_schema" in kwargs and kwargs["args_schema"] is not None:
             if not is_basemodel_subclass(kwargs["args_schema"]):
                 raise SchemaAnnotationError(
                     f"args_schema must be a subclass of pydantic BaseModel. "
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -42,20 +42,10 @@ from typing import (
     Tuple,
     Type,
     Union,
-    cast,
     get_type_hints,
 )
 
-from typing_extensions import Annotated, get_args, get_origin
-
 from pydantic import (
-    BaseModel,
-    ConfigDict,
-    Extra,
-    Field,
-    ValidationError,
-    create_model,
-    root_validator,
     validate_arguments,
 )
 from pydantic.fields import FieldInfo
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/tools.py
```diff
@@ -119,6 +125,12 @@ def _create_subset_model(
     fn_description: Optional[str] = None,
 ) -> Type[BaseModel]:
     """Create a pydantic model with only a subset of model's fields."""
+    if PYDANTIC_MAJOR_VERSION == 2:
+        from pydantic.v1 import create_model  # pydantic: ignore
+    else:
+        from pydantic import (  # type: ignore[no-redef] # pydantic: ignore
+            create_model,
+        )
```
baskaryan, line 133: why not `from langchain_core.pydantic_v1 import create_tool`?

baskaryan: ie use the create_model_v1 that's already imported

libs/core/langchain_core/tools.py
```diff
@@ -138,6 +150,75 @@ def _create_subset_model(
     return rtn
 
 
+def _create_subset_model_2(
```
baskaryan, line 153: nit: v2

libs/core/langchain_core/tools.py
```diff
@@ -138,6 +150,75 @@ def _create_subset_model(
...
+        from pydantic.v1 import BaseModel as BaseModelV1  # pydantic: ignore
+
+        if issubclass(model, BaseModelV1):
+            return _create_subset_model_v1(
+                name,
+                model,
+                field_names,
+                descriptions=descriptions,
+                fn_description=fn_description,
+            )
```
baskaryan:
```suggestion
from langchain_core.pydantic_v1 import BaseModelV1Like

    if isinstance(model, BaseModelV1Like):
        return _create_subset_model_v1(
            name,
            model,
            field_names,
            descriptions=descriptions,
            fn_description=fn_description,
        )
    else:
      ...
```

libs/core/langchain_core/tools.py
```diff
@@ -403,6 +380,16 @@ class ChildTool(BaseTool):
     two-tuple corresponding to the (content, artifact) of a ToolMessage.
     """
 
+    def __init__(self, **kwargs: Any) -> None:
+        """Initialize the tool."""
+        if "args_schema" in kwargs and kwargs["args_schema"] is not None:
+            if not is_basemodel_subclass(kwargs["args_schema"]):
+                raise SchemaAnnotationError(
```
baskaryan, line 387: is this a SchemaAnnotationError, sounds like a type error or value error?

baskaryan: do we need to update line 311 as well?

eyurtsev: Probably not in this PR for line 311 I don't know what its role was. Users could be relying on the identity of the exception

libs/core/langchain_core/utils/pydantic.py
```diff
@@ -64,3 +128,106 @@ def wrapper(cls: Type[BaseModel], values: Dict[str, Any]) -> Dict[str, Any]:
         return func(cls, values)
 
     return wrapper
+
+
+def _create_subset_model_v1(
```
tdiggelm, line 133: Unfortunately this function fails for pydantic v1 models that use `Annotated` with `Field`, e.g.

```python
class InputModel(BaseModel):
    query: Annotated[str, pydantic_v1.Field(description="Hello World")]

_create_subset_model_v1("test", InputModel, InputModel.__annotations__.keys())
```

This produces the following error:

```plain
ValueError: cannot specify `Annotated` and value `Field`s together for 'query'
```

PR #24422, opened 2024-07-18
fix RAG with quantized embeddings notebook

rbrugaro:
1. Fix HuggingfacePipeline import error to newer partner package
 2. Switch to IPEXModelForCausalLM for performance

There are no dependency changes since optimum intel is also needed for QuantizedBiEncoderEmbeddings


ccurme (approved):

rbrugaro:
thanks @ccurme I fixed lint error

PR #24431, opened 2024-07-19
[Vector Store] Fix `_euclidean_relevance_score_fn` in base vectorstore

ZhangShenao:
According to the function `_euclidean_relevance_score_fn` in base vectorstore, the euclidean_relevance_score may be negative if the dimension of vector is more than 2. I encountered this warning when using Faiss. Would it be better to change the calculation method of relevant_score?

### DIFF
--- a/libs/core/langchain_core/vectorstores/base.py
+++ b/libs/core/langchain_core/vectorstores/base.py
@@ -635,7 +635,7 @@ class VectorStore(ABC):
         # This function converts the Euclidean norm of normalized embeddings
         # (0 is most similar, sqrt(2) most dissimilar)
         # to a similarity function (0 to 1)
-        return 1.0 - distance / math.sqrt(2)
+        return 1.0 / (1.0 + distance)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


ccurme:
Related to https://github.com/langchain-ai/langchain/issues/18709

This issue requires broader consideration than is provided by this update to core. The method has a comment that reads:
```
# The 'correct' relevance function
# may differ depending on a few things, including:
# - the distance / similarity metric used by the VectorStore
# - the scale of your embeddings (OpenAI's are unit normed. Many
#  others are not!)
# - embedding dimensionality
# - etc.
# This function converts the Euclidean norm of normalized embeddings
# (0 is most similar, sqrt(2) most dissimilar)
# to a similarity function (0 to 1)
```
In this case the issue is with how this is applied to FAISS. Changing the function in core will have unwarranted changes for users of all vector stores.

chrispy-snps:
@ccurme - per #18709, the correct fix is not to change the global default relevance function (which would indeed have undesirable side effects), but to implement a new `distance_squared` function that is correct for FAISS and ChromaDB.

PR #24436, opened 2024-07-19
docs: Update langchain-openai package version in chat_token_usage_tracking

v0idzz:
This PR updates docs to mention correct version of the `langchain-openai` package required to use the `stream_usage` parameter.

As it can be noticed in the details of this [merge commit](https://github.com/langchain-ai/langchain/commit/722c8f50ea0a6ff2e1a35b2380408eca02dbe06a), that functionality is available only in `langchain-openai >= 0.1.9` while docs state it's available in `langchain-openai >= 0.1.8`.


ccurme (approved):
Thank you!

PR #24458, opened 2024-07-19
community: Update aerospike-vector-search client to 1.0.0 to support aerospike-vector-search server 0.9.0

jdogmcsteezy:

 **Description:** 
Updated the docs notebook to use the latest client and updated the docker compose config files to support the latest server release.

I also remove --snapshot-warn-unused and asyncio_mode from the pyproject.toml ini_options. These options were causing some "unknown" flag issues while using a supported version of pytest. 

```
 pytest --version
pytest 7.4.4
```


jdogmcsteezy:
@baskaryan, please have a look when you have a moment

jdogmcsteezy:
@baskaryan, I know this is a busy repo with lots of PRs. Do you have any idea when this kind of PR will get merged?

jdogmcsteezy:
@efriis

libs/community/pyproject.toml
efriis, line 1: revert

libs/community/langchain_community/vectorstores/aerospike.py
```diff
@@ -545,7 +545,7 @@ def from_texts(
         cls,
         texts: List[str],
         embedding: Embeddings,
-        metadatas: Optional[List[dict]] = None,
+        labels: Optional[List[dict]] = None,
```
efriis, line 548: breaking change

libs/community/langchain_community/vectorstores/aerospike.py
```diff
@@ -155,7 +155,7 @@ def convert_distance_strategy(
     def add_texts(
         self,
         texts: Iterable[str],
-        metadatas: Optional[List[dict]] = None,
+        labels: Optional[List[dict[str, str]]] = None,
```
efriis, line 158: As a list of dictionaries I think `metadatas` should probably stand

jdogmcsteezy: The labels can not be nested and must have keys and values of type string. Like so
```
{"model": "llama3", "date": "1/1/25"}
```

Are you suggesting that I revert it?

libs/community/langchain_community/vectorstores/aerospike.py
jdogmcsteezy: Our client was still in beta when this was initially created. This breaking change is OK since it better matches our documentation. If you still would rather we not commit this breaking change, I will revert it.

efriis:
This isn't mergable in its current state because

- failing ci
- naming changes in library are breaking changes (for anyone passing `metadatas=` currently)

want to reopen with just the docs changes telling people to use the new client? From my skimming, it looks like the existing integration will work with the client.

PR #24459, opened 2024-07-19
core[minor]: Relax constraints on type checking for tools and parsers

eyurtsev:
This will allow tools and parsers to accept pydantic models from any of the
following namespaces:

* pydantic.BaseModel with pydantic 1
* pydantic.BaseModel with pydantic 2
* pydantic.v1.BaseModel with pydantic 2

### DIFF
--- a/libs/core/langchain_core/utils/pydantic.py
+++ b/libs/core/langchain_core/utils/pydantic.py
@@ -7,23 +7,15 @@ import textwrap
 from functools import wraps
 from typing import Any, Callable, Dict, List, Optional, Type, Union
 
-import pydantic
+import pydantic  # pydantic: ignore
 
-from langchain_core.pydantic_v1 import BaseModel, root_validator
-
-
-def get_pydantic_major_version() -> int:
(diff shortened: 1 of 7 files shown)


libs/core/langchain_core/utils/pydantic.py
```diff
@@ -37,13 +58,13 @@ def is_basemodel_subclass(cls: Type) -> bool:
         return False
 
     if PYDANTIC_MAJOR_VERSION == 1:
-        from pydantic import BaseModel as BaseModelV1Proper
+        from pydantic import BaseModel as BaseModelV1Proper  # pydantic: ignore
```
baskaryan, line 61: woah what's pydantic: ignore

baskaryan (approved):

libs/core/langchain_core/utils/pydantic.py
eyurtsev: It's our own linter to make sure that we don't import pydantic directly

PR #24460, opened 2024-07-19
community: remove integration test deps

efriis:
they arent used


libs/community/tests/integration_tests/agent/test_ainetwork_agent.py
efriis, line 1: ruff yellin

PR #24466, opened 2024-07-20
community: Improve QianfanChatEndpoint tool result to model

maang-h:
- **Description:** `QianfanChatEndpoint` When using tool result to answer questions, the content of the tool is required to be in Dict format. Of course, this can require users to return Dict format when calling the tool, but in order to be consistent with other Chat Models, I think such modifications are necessary.

### DIFF
--- a/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
+++ b/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import uuid
 from operator import itemgetter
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
```diff
@@ -74,6 +75,20 @@ def convert_message_to_dict(message: BaseMessage) -> dict:
...
 
 
+def _create_tool_content(content: Union[str, List[Union[str, Dict[Any, Any]]]]) -> str:
+    """Convert tool content to dict scheme."""
+    if isinstance(content, str):
+        try:
+            if isinstance(json.loads(content), dict):
+                return content
+            else:
+                return json.dumps({"original_content": content})
```
ccurme, line 85: out of curiosity, are there Baidu docs that give examples of tool calling, to see what is generally expected here? are all return values from tools expected to have names?

if we are forced to pick a magic name I'd have a slight preference for `"result"` or `"tool_result"` or similar. but don't feel strongly.

maang-h: The docs url: https://cloud.baidu.com/doc/WENXINWORKSHOP/s/jlil56u11#function-call%E7%A4%BA%E4%BE%8B%EF%BC%88%E5%A4%9A%E8%BD%AE%EF%BC%89
And `tool_result` is more descriptive than `original_content`, I will modify it.

ccurme (approved):

PR #24467, opened 2024-07-20
Rename ‘embedding’ to ‘embeddings’ and use np.max for better performance for maximal_marginal_relevance in qdrant client package

Ray0907:
- **Naming Consistency**: I found that the README example uses 'embeddings', but the code uses 'embedding'. I believe 'embeddings' makes more sense for consistency.
- **Performance Improvement**: I replaced `np.max` with `max` for better performance in maximal_marginal_relevance.

### DIFF
--- a/libs/partners/qdrant/langchain_qdrant/_utils.py
+++ b/libs/partners/qdrant/langchain_qdrant/_utils.py
@@ -27,7 +27,7 @@ def maximal_marginal_relevance(
         for i, query_score in enumerate(similarity_to_query):
             if i in idxs:
                 continue
-            redundant_score = max(similarity_to_selected[i])
+            redundant_score = np.max(similarity_to_selected[i])
             equation_score = (
                 lambda_mult * query_score - (1 - lambda_mult) * redundant_score
             )
--- a/libs/partners/qdrant/langchain_qdrant/qdrant.py
+++ b/libs/partners/qdrant/langchain_qdrant/qdrant.py
@@ -45,7 +45,7 @@ class QdrantVectorStore(VectorStore):
         .. code-block:: python
         from langchain_qdrant import QdrantVectorStore
 
-        store = QdrantVectorStore.from_existing_collection("my-collection", embedding, url="http://localhost:6333")
+        store = QdrantVectorStore.from_existing_collection("my-collection", embeddings, url="http://localhost:6333")
     """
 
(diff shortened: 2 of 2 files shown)


ccurme:
Changing the kwarg names like this will break any user code using the existing name. I think it's preferable to update the documentation to resolve any inconsistencies.

Regarding use of `np.max`: we'd need to add this as an explicit dependency to the package (although I believe it is implicitly already installed). Can you quantify or profile what sort of performance increase we would gain?

Anush008:
Hey. We can't change the kwarg name since it'll break usage.
It was named `embedding` and not `embeddings` to maintain parity with the base interface's naming, i.e. https://github.com/langchain-ai/langchain/blob/d98b830e4b965997299d04e2aea769b01b856cee/libs/core/langchain_core/vectorstores/base.py#L1040-L1045

Ray0907:
> Changing the kwarg names like this will break any user code using the existing name. I think it's preferable to update the documentation to resolve any inconsistencies.
> 
> Regarding use of `np.max`: we'd need to add this as an explicit dependency to the package (although I believe it is implicitly already installed). Can you quantify or profile what sort of performance increase we would gain?

@ccurme I understand. It’s important to ensure consistency between the code and the example. Changing the example might be the best option. Additionally, as far as I know, using np.max is more efficient for numpy arrays, right? How can we determine the performance of these two methods? Should we test them using the same array to compare their performance?

Ray0907:
> Hey. We can't change the kwarg name since it'll break usage. It was named `embedding` and not `embeddings` to maintain parity with the base interface's naming, i.e.
> 
> https://github.com/langchain-ai/langchain/blob/d98b830e4b965997299d04e2aea769b01b856cee/libs/core/langchain_core/vectorstores/base.py#L1040-L1045

@Anush008  Yeah, maybe I didn’t consider the core module from Langchain. Thanks for your advice.

ccurme:
> @ccurme I understand. It’s important to ensure consistency between the code and the example. Changing the example might be the best option. Additionally, as far as I know, using np.max is more efficient for numpy arrays, right? How can we determine the performance of these two methods? Should we test them using the same array to compare their performance?

I took a closer look at this and think it's fine to use `np` since the output of cosine_similarity is a numpy array.

efriis:
closing as stale. Feel free to reopen with just the `np` changes!

PR #24469, opened 2024-07-20
[Vector Store] Fix function `add_texts` in `TencentVectorDB`

ZhangShenao:
Regardless of whether `embedding_func` is set or not, the 'text' attribute of document should be assigned, otherwise the `page_content` in the document of the final search result will be lost

### DIFF
--- a/libs/community/langchain_community/vectorstores/tencentvectordb.py
+++ b/libs/community/langchain_community/vectorstores/tencentvectordb.py
@@ -375,8 +375,7 @@ class TencentVectorDB(VectorStore):
                 }
                 if embeddings:
                     doc_attrs["vector"] = embeddings[id]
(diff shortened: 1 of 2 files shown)


ccurme:
Would you mind adding a test and confirming that existing integration tests pass? https://github.com/langchain-ai/langchain/blob/master/libs/community/tests/integration_tests/vectorstores/test_tencentvectordb.py

ZhangShenao:
> Would you mind adding a test and confirming that existing integration tests pass? https://github.com/langchain-ai/langchain/blob/master/libs/community/tests/integration_tests/vectorstores/test_tencentvectordb.py

I've added a test case, please review

ccurme (approved):

PR #24376, opened 2024-07-18
[community]: adding artifact to Tavily search

isahers1:
This allows you to get raw content as well as the answer, instead of just getting the results.

PR #24480, opened 2024-07-21
community: Avoid return str instead of json for results tool

dopc:
**PR message**: ***Delete this entire checklist*** and replace with
    - **Description:** GoogleSerperResults was returning `string` instead of JSON object. Fixed this.
    - **Issue:** 
    - **Dependencies:** 
    - **Twitter handle:** 0dpc0

### DIFF
--- a/libs/community/langchain_community/tools/google_serper/tool.py
+++ b/libs/community/langchain_community/tools/google_serper/tool.py
@@ -58,7 +58,7 @@ class GoogleSerperResults(BaseTool):
         run_manager: Optional[CallbackManagerForToolRun] = None,
     ) -> str:
         """Use the tool."""
-        return str(self.api_wrapper.results(query))
+        return self.api_wrapper.results(query)
 
     async def _arun(
         self,
@@ -67,4 +67,4 @@ class GoogleSerperResults(BaseTool):
     ) -> str:
         """Use the tool asynchronously."""
 
-        return (await self.api_wrapper.aresults(query)).__str__()
+        return (await self.api_wrapper.aresults(query))


libs/community/langchain_community/tools/google_serper/tool.py
```diff
@@ -58,7 +58,7 @@ def _run(
         run_manager: Optional[CallbackManagerForToolRun] = None,
     ) -> str:
         """Use the tool."""
-        return str(self.api_wrapper.results(query))
+        return self.api_wrapper.results(query)
```
eyurtsev, line 61: @dopc this is a breaking change. The change itself is also not correct given that the tool claims to return `str` (see function signature). If you want to use the tool to return things that are not strings, you can use the underlying API wrapper.

@baskaryan tagging you in case there's any general refactor that can be done w/ the new `artifact` attribute in tool message

eyurtsev (changes requested):
We don't want to make this change. You can instead access the underlying APIWrapper, tools are currently consumed by langchain models which expect a `str` input as part of the invocation.

libs/community/langchain_community/tools/google_serper/tool.py
baskaryan: @eyurtsev we can start returning both the raw results and the string as a tuple, with response_format=content_and_artifact. here's an example https://github.com/langchain-ai/langchain/pull/24376

hwchase17:
we now support this more properly - see https://github.com/langchain-ai/langchain/pull/24376/files for an example

closing, but if you want to reopen that would be the way

PR #24485, opened 2024-07-21
core: New OutputFunctionsFixingParser to support OpenAI functions parsers

MartinGotelli:
Hey guys, thank you for all your work! Before doing anything, let me know if you think this should be part of the community project. I can move it.

 - **Description:** New parser to fix possible problems when using OpenAI functions. This works great wrapping PydanticOutputFunctionsParser, we are using this in production at my company.
```python
def create_functions_runnable(output_models: List[Type[BaseModel]], model, prompt):
    parser = get_openai_output_parser(output_models)
    output_parser = OutputFunctionsFixingParser.from_llm(
        parser=parser,
        llm=model,
        functions=output_models,
    )
    return create_openai_fn_runnable(output_models, model, prompt, output_parser=output_parser)
```
When the ChatOpenAI model decides to return a JSON instead of a message with functions_call this tool comes in handy because you have a specific prompt to return a function call.
image
image
 - **Issue:** N/A
 - **Dependencies:** N/A
 - **Twitter handle:** @martinferenaz


efriis:
Great that you're using this well! Would recommend keeping this in your codebase, and this isn't the type of thing we'd add to the library because it depends on specific prompts. In fact this is more of a chain than an output parser!

PR #24486, opened 2024-07-21
core[patch]: ChatPromptTemplate.init same as ChatPromptTemplate.from_…

baskaryan:


libs/core/langchain_core/prompts/chat.py
```diff
@@ -921,6 +921,86 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
...
+
+        kwargs = {
+            **dict(
+                input_variables=sorted(input_vars),
+                optional_variables=sorted(optional_variables),
+                partial_variables=partial_vars,
+            ),
+            **kwargs,
+        }
+        cast(Type[ChatPromptTemplate], super()).__init__(messages=_messages, **kwargs)
```
nfcampos, line 1002: Why is cast needed

libs/core/langchain_core/prompts/chat.py
```diff
@@ -921,6 +921,86 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
...
+
```
nfcampos, line 1003: Shouldn't from_messages now delegate to init

libs/core/langchain_core/prompts/chat.py
```diff
@@ -921,6 +921,86 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
...
+            Instantiation from mixed message formats:
+
+            .. code-block:: python
+
+                template = ChatPromptTemplate([
+                    SystemMessage(content="hello"),
+                    ("human", "Hello, how are you?"),
+                ])
+
```
eyurtsev, line 975: perhaps we add a ..versionadded and say that prior to that version initialization is via the classmethod or something of that sort?

libs/core/langchain_core/prompts/chat.py
baskaryan: mypy thinks its ``super().__init__()`` output BaseChatPromptTemplate

PR #24502, opened 2024-07-22
community: Remove targets_table from C* GraphVectorStore

bjchambers:
- **Description:** Remove the unnecessary `targets_table` parameter


bjchambers:
@cbornet

cbornet (approved):
LGTM!

eyurtsev (approved):

eyurtsev:
This looks like a breaking change. I realize that you're managing this integration, so merging in. If dangerous you could also raise a deprecation error next time and/or a warning that the parameter has been removed.

cbornet:
We ensured a deprecation path for the code. In new versions of ragstack-ai-knowledge-store, targets_table is allowed at init and ignored and will output a deprecation warning.
It's a breaking change on the implementation schema though. We think it's still possible to do it at this stage.

PR #24521, opened 2024-07-22
[partner]: minor change to embeddings for Ollama

isahers1:


jacoblee93 (approved):
LGTM if you've tested but fix lint

PR #24544, opened 2024-07-23
partners[milvus]: add dynamic field

zc277584121:
add dynamic field feature to langchain_milvus
more unittest, more robustic

plan to deprecate the `metadata_field` in the future, because it's function is the same as `enable_dynamic_field`, but the latter one is a more advanced concept in milvus


efriis (approved):

zc277584121:
thank you very much

PR #24551, opened 2024-07-23
CI: Temporarily disable min version checking on pull request

eyurtsev:
Short term to fix CI


eyurtsev:
@efriis Not obvious to me what's going on, but it seems an older version of langchain-core was still being installed when running tests against langchain-community from master

PR #24552, opened 2024-07-23
openai[patch]: move test

ccurme:
No-override tests (https://github.com/langchain-ai/langchain/pull/24407) include a condition that integrations not implement additional tests.

### DIFF
--- a/libs/partners/openai/tests/integration_tests/chat_models/test_base.py
+++ b/libs/partners/openai/tests/integration_tests/chat_models/test_base.py
@@ -1,7 +1,9 @@
 """Test ChatOpenAI chat model."""
 
+import base64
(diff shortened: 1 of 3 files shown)


eyurtsev:
@ccurme Why should we allow integrations to provide additional tests? They can create an extra test suite for that purpose?

ccurme:
> @ccurme Why should we allow integrations to provide additional tests? They can create an extra test suite for that purpose?

I guess it's a small convenience so we keep track of what model / parameters are tested in one place. But will update to move it to the openai-specific tests.

eyurtsev:
I suspect it'll be good to separate them -- a bit extra work, but then it becomes easier to identify the additional tests

PR #24566, opened 2024-07-23
experimental: safe parse NERs from list of dictionaries

Otnielush:
**Issue:** #24493, by default some LLM generations return output + input promt. In this case after json extraction structure can be in format: [[],[], {_usefull_data_}, ...]. 
I created new fuction to safe parse unpredictable outputs


baskaryan:
cc @tomasonjo 

also pr is stil failing lint. to format and lint locally run
```bash
cd libs/experimental
poetry install --with lint,typing
make format
make lint
```

tomasonjo:
Do I  undestand correctly you just ignore results if they are incorrect? Nothing else is done in this PR?

Otnielush:
> Do I undestand correctly you just ignore results if they are incorrect? Nothing else is done in this PR?
Avoid errors if data not parsed correctly

tomasonjo:
I want to propagate errors, maybe we need special parsing for hf models
where we remove input prompt first

V ned., 28. jul. 2024, 12:25 je oseba Otniel ***@***.***>
napisala:

Otnielush:
> cc @tomasonjo
> 
> also pr is stil failing lint. to format and lint locally run
> 
> ```shell
> cd libs/experimental
> poetry install --with lint,typing
> make format
> make lint
> ```

is there easy way to do it on windows? ;)

PR #24570, opened 2024-07-23
core: Mermaid graph image rendering with `playwright`

fedeoliv:
**Description**: Given that [pyppeteer is no longer maintained](https://github.com/pyppeteer/pyppeteer), this PR introduces the use of [playwright-python](https://github.com/microsoft/playwright-python) as a new alternative to generate Mermaid syntax and render flowcharts from graph data. This includes support for custom node colors and edge curve styles, as well as the ability to export the generated graphs to PNG.

**Dependencies**: `playwright` dependency and browser installation via `playwright install`.


fedeoliv:
cc @baskaryan

libs/core/langchain_core/runnables/graph_mermaid.py
```diff
@@ -164,7 +164,15 @@ def draw_mermaid_png(
     Raises:
         ValueError: If an invalid draw method is provided.
     """
-    if draw_method == MermaidDrawMethod.PYPPETEER:
+    if draw_method == MermaidDrawMethod.PLAYWRIGHT:
+        import asyncio
+
+        img_bytes = asyncio.run(
```
eyurtsev, line 170: This does not handle the event loop correctly, so we'll need to fix this prior to merging

libs/core/langchain_core/runnables/graph_mermaid.py
```diff
@@ -164,7 +164,15 @@ def draw_mermaid_png(
     Raises:
         ValueError: If an invalid draw method is provided.
     """
-    if draw_method == MermaidDrawMethod.PYPPETEER:
+    if draw_method == MermaidDrawMethod.PLAYWRIGHT:
```
eyurtsev, line 167: The draw_method should support inputs as strings and not require users to input enums

libs/core/langchain_core/runnables/graph.py
```diff
@@ -173,6 +173,7 @@ class NodeStyles:
 class MermaidDrawMethod(Enum):
     """Enum for different draw methods supported by Mermaid"""
 
+    PLAYWRIGHT = "playwright"  # Uses Playwright to render the graph
```
eyurtsev, line 176: We need to properly document this enum so it surfaces in API reference docs.

eyurtsev (commented):
@fedeoliv thank you for submitting this PR. I'm confirming whether we want to keep supporting rendering of the graphs via the browser. If that's the case, I'll take over this PR to fix handling of the event loop (the current implementation will break in a notebook environment).

efriis:
closing for now - this isn't mergable in the current state for reasons outlined above!

efriis:
actually before closing - @fedeoliv any interest in fixing the event loop and CI issues?

efriis:
@eyurtsev is handling in #25399 !

PR #24573, opened 2024-07-23
docs: add how-to guide on handling large numbers of tools

ccurme:


bracesproul (commented):
Looks reasonable, my one implementation detail comment is I'd probably generate a summary for each tool and embed that alongside the schema. Not too big a deal if not included though

ccurme:
Added a guide to langgraph docs instead: https://langchain-ai.github.io/langgraph/how-tos/many-tools/#define-the-tools

PR #24579, opened 2024-07-23
community: fixing azure_openai.py

antunsz:
fixing the min dict value for chunk_size. The key could not exists at the moment of the access.

- [ x] **Lint and test**: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified.


baskaryan (commented):
this is a deprecated implementation (see L17), could we port any fixes to the maintained version in`libs/partners/openai/embeddings/azure.py`?

antunsz:
> this is a deprecated implementation (see L17), could we port any fixes to the maintained version in`libs/partners/openai/embeddings/azure.py`?

Absolutely! Thanks for the directions 🤜🤛 ! I'm closing this PR and moving the changes to the properly version.

PR #24585, opened 2024-07-23
docs[patch]: Adds components for prereqs, compatibility, fix chat model tab issue

jacoblee93:
Added to `docs/how_to/tools_runtime` as a proof of concept, will apply everywhere if we like.

A bit more compact than the default callouts, will help standardize the layout of our pages since we frequently use these boxes.

Screenshot 2024-07-23 at 4 49 02 PM


eyurtsev:
@jacoblee93 is there any way to reduce the API we need to know about inside the notebooks? And add a directive that'll get post-processed into the appropriate format?

docs/src/theme/Prerequisites.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import Admonition from '@theme/Admonition';
+
+export default function Prerequisites({ titlesAndLinks }) {
+    return (
+        <Admonition type="info" title="Recommended Prereading" icon="📚">
```
baskaryan, line 6: i think i like "Conceptual Prereqs" more than recommended reading — they really are prereqs for understanding what's going on, not just suggestions

jacoblee93: Changed back to `Prerequisites`

jacoblee93:
> @jacoblee93 is there any way to reduce the API we need to know about inside the notebooks? And add a directive that'll get post-processed into the appropriate format?

We *could* swizzle the admonitions component and use a custom one if you pass a specific type in. However, I think part of the idea here is actually to reduce the amount of customization a writer can add here, and if we go with that approach you'd be able to enter whatever you want.

baskaryan (approved):

efriis (approved):
I like it! Should we do the same format for related section in #24829 ?

jacoblee93:
We could! I think fine as a followup though

daniel-jscraft:
ha! this is nice 👍

PR #24601, opened 2024-07-24
core: [Vector Store] Optimize in memory vector store

ZhangShenao:
- Change memory store to Dict[str, Document]
- Save embedded vector in metadata.
- Add unit test for getting vector


baskaryan (commented):
what's the motivation for this pr?

eyurtsev (changes requested):
Hi thanks for the contribution!

This change was not done to avoid making a breaking change in `self.store` since it's possible that some users rely on the existing format.

The vector embedding should not be a part of the metadata.

ZhangShenao:
> Hi thanks for the contribution!
> 
> This change was not done to avoid making a breaking change in `self.store` since it's possible that some users rely on the existing format.
> 
> The vector embedding should not be a part of the metadata.

Thank you for your guidance. Could you please help modify this PR?

hwchase17:
> * Change memory store to Dict[str, Document]

I dont think we want to make this change at the moment - it would be a break change
> * Save embedded vector in metadata.

Currently not where we put vectors

> * Add unit test for getting vector

This may be useful, but should be separate pr

Issue #22296, opened 2024-05-29
Standardize ChatModel docstrings and integration docs


baskaryan:
## Issue

To make our chat model integrations as easy to use as possible we need to make sure the docs for them are thorough and standardized. There are two parts to this: updating the chat model docstrings and updating the actual integration docs.

This needs to be done for each ChatModel integration, ideally with one PR per ChatModel.

Related to broader issues #21983 and #22005.

## Docstrings

Each ChatModel class docstring should have the sections shown in the [Appendix](#appendix) below. The sections should have input and output code blocks when relevant. See ChatOpenAI [docstrings](https://github.com/langchain-ai/langchain/blob/f39e1a22884005390a3e5aa2beffaadfdc7028dc/libs/partners/openai/langchain_openai/chat_models/base.py#L1120) and [corresponding API reference](https://api.python.langchain.com/en/latest/chat_models/langchain_openai.chat_models.base.ChatOpenAI.html) for an example.

To build a preview of the API docs for the package you're working on run (from root of repo):

```bash
make api_docs_clean; make api_docs_quick_preview API_PKG=openai
```

where `API_PKG=` should be the parent directory that houses the edited package (e.g. community, openai, anthropic, huggingface, together, mistralai, groq, fireworks, etc.). This should be quite fast for all the partner packages.

## Doc pages

Each ChatModel [docs page](https://python.langchain.com/v0.2/docs/integrations/chat/) should follow [this template](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/chat.ipynb). See [ChatOpenAI](https://python.langchain.com/v0.2/docs/integrations/chat/openai/) for an example.

You can use the `langchain-cli` to quickly get started with a new chat model integration docs page (run from root of repo):

```bash
poetry run pip install -e libs/cli
poetry run langchain-cli integration create-doc --name "foo-bar" --name-class FooBar --destination-dir ./docs/docs/integrations/chat/
```

where `--name` is the integration package name without the "langchain-" prefix and `--name-class` is the class name without the "Chat" prefix. This will create a template doc with some autopopulated fields at docs/docs/integrations/chat/foo_bar.ipynb.

To build a preview of the docs you can run (from root):

```bash
make docs_clean
make docs_build
cd docs/build/output-new
yarn
yarn start
```

## Appendix

Expected sections for the ChatModel class docstring.

```python
    """__ModuleName__ chat model integration.

    Setup:
      ...

[...]

PR #24616, opened 2024-07-24
docs: standardize groq chat model docs 

baskaryan:
part of #22296


baskaryan:
@isahers1 could we add a link here in intro https://console.groq.com/docs/models

PR #24617, opened 2024-07-24
docs: standardize Together docs

baskaryan:
Part of #22296


baskaryan:
@isahers1 could we update the link to their docs to https://docs.together.ai/docs/chat-models

PR #24619, opened 2024-07-24
make image inputs compatible with langchain_ollama

isahers1:


efriis (commented):
why are we removing tool call ids? I believe that's a feature of e.g. mistral v3. Is it not compatible with some other models?

isahers1:
> why are we removing tool call ids? I believe that's a feature of e.g. mistral v3. Is it not compatible with some other models?

This is the definition of Message in ollama, it does not include tool_call_id so I removed. Lmk if that is the wrong approach:

```
class Message(TypedDict):
  """
  Chat message.
  """

  role: Literal['user', 'assistant', 'system']
  "Assumed role of the message. Response messages always has role 'assistant'."

  content: str
  'Content of the message. Response messages contains message fragments when streaming.'

  images: NotRequired[Sequence[Any]]
  """
  Optional list of image data for multimodal models.

  Valid input types are:

  - `str` or path-like object: path to image file
  - `bytes` or bytes-like object: raw image data

  Valid image formats depend on the model. See the model card for more information.
  """

  tool_calls: NotRequired[Sequence[ToolCall]]
  """
  Tools calls to be made by the model.
  """
```

libs/partners/ollama/tests/integration_tests/test_chat_models.py
```diff
@@ -15,3 +17,19 @@ def chat_model_class(self) -> Type[ChatOllama]:
     @property
     def chat_model_params(self) -> dict:
         return {"model": "llama3-groq-tool-use"}
+
```
baskaryan, line 20: we should turn on image integration tests
```python
    @property
        return True
```

baskaryan (approved):

PR #24625, opened 2024-07-24
docs: Update pydantic compatibility

eyurtsev:
Update pydantic compatibility. This will only be true after we release the partner packages.


docs/docs/how_to/pydantic_compatibility.md
```diff
@@ -1,27 +1,102 @@
...
+Internally, LangChain continues to use the [Pydantic V1](https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features) via
+the v1 namespace of Pydantic 2.
 
-User can either pin to pydantic v1, and upgrade their code in one go once LangChain has migrated to v2 internally, or they can start a partial migration to v2, but must avoid mixing v1 and v2 code for LangChain.
+Because Pydantic does not support mixing .v1 and .v2 objects, users should be aware of a number of issues
+when using LangChain with Pydantic.
+
+:::note
+
+Pydantic 1 End of Life was in June 2024. LangChain will be dropping support for Pydantic 1 in the near future,
```
baskaryan, line 16: shoudl this be one of the bullets at the top?

also would explicitly say something to effect of "This change will be accompanied with a minor version bump across all relevant LangChain libraries"

docs/docs/how_to/pydantic_compatibility.md
```diff
@@ -1,27 +1,102 @@
...
+| `BaseChatModel.bind_tools`             | Yes        | langchain-core>=0.2.23, appropriate version of partner package |
+| `BaseChatModel.with_structured_output` | Yes        | langchain-core>=0.2.23, appropriate version of partner package |
+| `Tool.from_function`                   | Yes        | langchain-core>=0.2.23                                         |
+| `StructuredTool.from_function`         | Yes        | langchain-core>=0.2.23                                         |
+
+
+Partner packages that accept pydantic v2 objects via `bind_tools` or `with_structured_output` APIs:
+
+| Package Name        | pydantic v1 | pydantic v2 |
+|---------------------|-------------|-------------|
```
baskaryan, line 40: what if we put langchain-core in here, and instead of the above table just have a list of the APIs and say: these all support pydantic 2 as of the following package versions:

eyurtsev: I was trying to guard against force install of incompatible package versions

baskaryan: watcha mean? which versions would be incompatible? these are all min bounds

eyurtsev: Pip force install unless it's no longer possible

PR #24628, opened 2024-07-24
core[patch]: introduce ToolMessage.status

baskaryan:
Anthropic models (including via Bedrock and other cloud platforms) accept a status/is_error attribute on tool messages/results (specifically in `tool_result` content blocks for Anthropic API). Adding a ToolMessage.status attribute so that users can set this attribute when using those models


ccurme (approved):
Could be worth some description in the PR summary of the motivation for the new attribute. e.g.,

- Supported by certain providers (Anthropic, others?)
- Enables more precise control in orchestration (e.g., can force certain behavior or constrain agent actions if a tool errors)
- ???

baskaryan:
> Could be worth some description in the PR summary of the motivation for the new attribute. e.g.,
> 
> * Supported by certain providers (Anthropic, others?)
> * Enables more precise control in orchestration (e.g., can force certain behavior or constrain agent actions if a tool errors)
> * ???

main motivation is first point. its part of the api of some providers (bedrock and anthropic) and currently our messages aren't expressive enough to match those apis

eyurtsev:
@baskaryan could you update the PR summary to explain the motivation. Are those APIs taking the tool status as an input?

eyurtsev (approved):

libs/core/langchain_core/messages/tool.py
```diff
@@ -67,6 +64,12 @@ class ToolMessage(BaseMessage):
     .. versionadded:: 0.2.17
     """
 
+    status: Literal["success", "error"] = "success"
+    """Status of the tool invocation.
```
eyurtsev, line 68: Perhaps expand the doc-string here?

```
"error" means that the tool invocation failed. This can occur due to ...
```

PR #24636, opened 2024-07-24
[docs]: add doctoring to ChatTogether

isahers1:


libs/partners/together/langchain_together/chat_models.py
```diff
@@ -19,19 +19,258 @@
...
+    Key init args — completion params:
+        model: str
+            Name of model to use.
+        temperature: float
+            Sampling temperature.
+        max_tokens: Optional[int]
+            Max number of tokens to generate.
+        logprobs: Optional[bool]
+            Whether to return logprobs.
+        stream_options: Dict
```
baskaryan, line 42: does together support this?

isahers1: nope, I was being silly

baskaryan (approved):

PR #24637, opened 2024-07-24
docs[patch]: Simplify tool calling guide, improve tool calling conceptual guide

jacoblee93:
Lots of duplicated content from concepts, missing pointers to the second half of the tool calling loop

Simpler + more focused + a more prominent link to the second half of the loop was what I was aiming for, but down to be more conservative and just more prominently link the "passing tools back to the model" guide.

I have also moved the tool calling conceptual guide out from under `Structured Output` (while leaving a small section for structured output-specific information) and added more content. The existing `#functiontool-calling` link will go to this new section.


docs/docs/how_to/tool_calling.ipynb
baskaryan, line 1: "If you are looking for an example of **using** tools"

maybe to be super duper clear: 
"If you want to see how to use the model-generated tool call to actually invoke a tool, ..."

jacoblee93: `If you want to see how to use the model-generated tool call to actually invoke a tool function` since `invoke a tool` sounds too much like `call a tool`

PR #24641, opened 2024-07-25
core[patch]: convert TypedDict to tool schema support

baskaryan:
supports following UX

```python
    class SubTool(TypedDict):
        """Subtool docstring"""

        args: Annotated[Dict[str, Any], {}, "this does bar"]

    class Tool(TypedDict):
        """Docstring
        Args:
            arg1: foo
        """

        arg1: str
        arg2: Union[int, str]
        arg3: Optional[List[SubTool]]
        arg4: Annotated[Literal["bar", "baz"], ..., "this does foo"]
        arg5: Annotated[Optional[float], None]
```

- can parse google style docstring
- can use Annotated to specify default value (second arg)
- can use Annotated to specify arg description (third arg)
- can have nested complex types


libs/core/langchain_core/utils/function_calling.py
```diff
@@ -187,6 +187,67 @@ def convert_python_function_to_openai_function(
...
 
+def _convert_typed_dict_to_openai_function(typed_dict: Type) -> FunctionDescription:
+    model = cast(Type[BaseModel], _convert_typed_dict_to_pydantic(typed_dict))
+    return convert_pydantic_to_openai_function(model)
+
+
+_MAX_TYPED_DICT_RECURSION = 25
+
+
+def _convert_typed_dict_to_pydantic(type_: Type, *, depth: int = 0) -> Type:
```
eyurtsev, line 198: Is there anyway to leverage pydantic here?

Main issue seems to be doc-strings?

```python
from typing_extensions import TypedDict, Annotated

from pydantic import RootModel

class Meow(TypedDict):
"""Meow"""
c: str

class Foo(TypedDict):
a: Annotated[int, 'hello']
b: float
meow: Meow

Q = RootModel[Foo]

Q.schema()
```

Do you think we'll accept dataclasses or nested dataclasses / typed-dicts?

eyurtsev: from dataclasses import dataclass, field

@dataclass
class Foo:
a: int = field(metadata={'description': 'hello'})

    def add_one_to_a(self):
        return self.a + 1

RootModel[Foo].schema()

foo = RootModel[Foo].parse_obj({'a': 5}).root

foo.add_one_to_a()

baskaryan: does pydantic v2 have RootModel?

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -187,6 +189,81 @@ def convert_python_function_to_openai_function(
...
+    type_: Type,
+    *,
+    visited: Dict,
+    depth: int = 0,
+) -> Type:
+    if type_ in visited:
+        return visited[type_]
+    elif depth >= _MAX_TYPED_DICT_RECURSION:
+        return type_
+    elif is_typeddict(type_):
```
eyurtsev, line 214: Does this work with typeddict from typing and from typing_extensions?

baskaryan: yep, in unit tests

PR #24680, opened 2024-07-25
partners/unstructured: fix client api_url

Coniferish:
**Description:** Add empty string default for api_key  and change `server_url` to `url` to match existing loaders.

Additional guidelines:


libs/partners/unstructured/langchain_unstructured/document_loaders.py
```diff
@@ -106,17 +106,17 @@ def __init__(
...
                 )
 
-        unstructured_api_key = api_key or os.getenv("UNSTRUCTURED_API_KEY")
-        unstructured_url = server_url or os.getenv("UNSTRUCTURED_URL") or _DEFAULT_URL
+        self.file_path = file_path
+        self.file = file
+        self.partition_via_api = partition_via_api
+        self.post_processors = post_processors
 
+        unstructured_api_key = api_key or os.getenv("UNSTRUCTURED_API_KEY") or ""
```
efriis, line 114: yes on this line!

efriis (commented):
can we just do line 114 and revert the rest?

Coniferish:
> can we just do line 114 and revert the rest?

I'd like to keep the switch to `url` so users are referencing the same param as the other, preexisting loaders, but I'll revert rearranging the variable orders

libs/partners/unstructured/langchain_unstructured/document_loaders.py
```diff
@@ -88,14 +88,14 @@ def __init__(
         # SDK parameters
         api_key: Optional[str] = None,
         client: Optional[UnstructuredClient] = None,
-        server_url: Optional[str] = None,
+        url: Optional[str] = None,
```
baskaryan, line 91: breaking change

efriis: I'm ok with this as a 0.1.1 release, and a yank of 0.1.0 - this is currently an unannounced package.

Coniferish: I dropped the commit that changed it back to `server_url` and updated the version to 0.1.1, so the PR should be good to go now @efriis

efriis (approved):

PR #24682, opened 2024-07-25
[experimental]: minor fix to open assistants code

isahers1:


libs/langchain/langchain/agents/openai_assistant/base.py
```diff
@@ -621,9 +621,10 @@ async def _aparse_intermediate_steps(
     ) -> dict:
         last_action, last_output = intermediate_steps[-1]
         run = await self._wait_for_run(last_action.run_id, last_action.thread_id)
-        required_tool_call_ids = {
-            tc.id for tc in run.required_action.submit_tool_outputs.tool_calls
-        }
+        if run.required_action:
+            required_tool_call_ids = {
+                tc.id for tc in run.required_action.submit_tool_outputs.tool_calls
+            }
```
bengladwell, line 627: In the case when `required_action` is `None`, doesn't that leave `required_tool_call_ids` unbound, which will result in an error on line 631:

```python
if action.tool_call_id in required_tool_call_ids
```
in the list comprehension?

bengladwell (commented):
I think this code might lead to a different error.

A test that covers this case might be wise.

libs/langchain/langchain/agents/openai_assistant/base.py
isahers1: You're right. Just added code to fix that error - do you mind adding a test case and I can review and then merge?

libs/experimental/langchain_experimental/llms/ollama_functions.py
```diff
@@ -132,6 +133,9 @@ def parse_response(message: BaseMessage) -> str:
     raise ValueError(f"`message` is not an instance of `AIMessage`: {message}")
 
 
+@deprecated(  # type: ignore[arg-type]
```
baskaryan, line 136: unrelated to this pr

isahers1:
@bengladwell Do you have time to add the test? I can do later if not.

bengladwell:
Thanks @isahers1 - I'm in the weeds on some things at the moment. I was hoping to get to this late in the week. If it would be better to get to this sooner, then yeah, that would be great if you would take it.

isahers1:
@bengladwell Later in the week is fine, am busy with some other stuff rn so will take a look again start of next week.

PR #24685, opened 2024-07-25
mongodb: bson optional import

efriis:


eyurtsev (approved):

libs/partners/mongodb/langchain_mongodb/utils.py
```diff
@@ -15,8 +15,6 @@
 from typing import Any, Dict, List, Union
 
 import numpy as np
-from bson import ObjectId
```
baskaryan, line 18: could only import if type checking and then not have to update func signatures

efriis: sphinx still runs type checking blocks

efriis: (at least this is what I was told. If they don't run in sphinx, agreed this would be better)

Issue #24600, opened 2024-07-24
DOC: Running Output-fixing parser example code results in an error


skydowx:
### URL

https://python.langchain.com/v0.2/docs/how_to/output_parser_fixing/

### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

I was running the code in the "How to use the output-fixing parser" page. After running the last line of code `new_parser.parse(misformatted)` instead of fixing it and returning the correct output, it gives an error:

```
ValidationError: 1 validation error for Generation
text
  str type expected (type=type_error.str)
```

scribu:
Just created a quick PR that should fix the issue: https://github.com/langchain-ai/langchain/pull/24687

PR #24687, opened 2024-07-25
langchain: Make OutputFixingParser.from_llm() create a useable retry chain

scribu:
Description: OutputFixingParser.from_llm() creates a retry chain that returns a Generation instance, when it should actually just return a string.
Issue: https://github.com/langchain-ai/langchain/issues/24600
Twitter handle: scribu


scribu:
There's a remaining lint error:

> tests/unit_tests/output_parsers/test_fix.py:75: error: Argument "llm" to "from_llm" of "OutputFixingParser" has incompatible type "RunnableLambda[str, AIMessage]"; expected "BaseLanguageModel[Any]"

Not sure what the best way to fix it would be.

PR #24691, opened 2024-07-25
rfc: community[patch]: add LCEL API chain

ccurme:
Option 1: pass system message, user query, AI message with tool calls, tool message to chat model to generate final response.

Option 2: format user query and tool message content into a second prompt template with system + human message, and pass to chat model to generate final response.

Option 1 is implemented here. We can also go with Option 2, which looks more like
```python
RunnablePassthrough.assign(
    tool_result=request_prompt | llm_with_tools | execute_tools
)
| response_prompt
| llm
```

### DIFF
--- a/libs/community/langchain_community/chains/openapi/chain.py
+++ b/libs/community/langchain_community/chains/openapi/chain.py
@@ -3,19 +3,29 @@
 from __future__ import annotations
 
 import json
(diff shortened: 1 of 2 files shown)


ccurme:
Closing as we will look into implementing with langgraph.

PR #24702, opened 2024-07-26
remove nvidiaembeddings from community

raspawar:
**community: remove NVIDIAEmbeddings**

cc: @mattf @dglogo @sumitkbh


mattf:
@efriis having previously marked the NeMoEmbeddings as deprecated, we would now like to remove it. users should go to NVIDIAEmbeddings from the langchain-nvidia-ai-endpoints package instead. what else should we do to complete the deprecation -> removed process?

baskaryan:
strong preference to not remove it, as this is a breaking change. we shoudl only do this when we do a minor version bump for langchain-community. the deprecation warning lists a specific version until which this import will be supported and we shoudl respect that

docs/docs/integrations/text_embedding/nemo.ipynb
baskaryan, line 1: don't think we want to remove this doc regardless of the deprecation, we should instead update it to import from the new package

dglogo:
> strong preference to not remove it, as this is a breaking change. we shoudl only do this when we do a minor version bump for langchain-community. the deprecation warning lists a specific version until which this import will be supported and we shoudl respect that

That's fair. Do you know when we can expect a minor version bump and what's the best way to make sure we make this change with the bumb? Thanks
cc: @raspawar @mattf

mattf:
> > strong preference to not remove it, as this is a breaking change. we shoudl only do this when we do a minor version bump for langchain-community. the deprecation warning lists a specific version until which this import will be supported and we shoudl respect that
> 
> That's fair. Do you know when we can expect a minor version bump and what's the best way to make sure we make this change with the bumb? Thanks cc: @raspawar @mattf

it looks like we marked this for removal in 0.2.0 (6 may, https://github.com/langchain-ai/langchain/pull/21239) a few weeks before 0.2.0 was released (17 may, https://pypi.org/project/langchain-community/0.2.0/).

@baskaryan i don't want to break any users who are relying on this. how can we get info on usage?

efriis (commented):
Howdy! Bagatur's comments still stand regarding this breaking change.

I'm good with the docs removal. Let's re-tag as 0.3.0 or 0.4.0 (slight preference for 0.4.0 - that's when we're planning more of these removals) for actual removal

Can also update the @deprecated tag to have alternative_import="langchain_nvidia_ai_endpoints.NVIDIAEmbeddings" instead of this message to make it hook into auto deprecation messages better

but let's keep the file stub as to not break anyone currently importing it

dglogo:
Thanks @efriis and @baskaryan. I think it's a good sweet spot.

efriis:
done in #25720

PR #24732, opened 2024-07-27
docs: component list generation

leo-gan:
**Issue:**
The current API Reference is built on the "package" perspective navigation. Each package is represented by a separate API Reference page.
The API Ref does not have the "component" perspective navigation. For example, how do you get a **list of all** embeddings, LLMs, tools, etc.?

With more and more components extracted in separate packages, the API Docs is also splitting into package-related pages. As a result, it is harder and harder to overview all components from all packages in one go.

Idea: Collect classes from all existing LC and partner packages; create the list of classes by "component" (by the component base class inheritance. Say, by the langchain_core.embeddings.Embeddings class for the embedding_model components.). 

**PR:** created a script to gather the component classes in the `docs/data/_component_classes.yaml` file. This file can be used to create the "component_perspective" navigation in API Reference which adds to the current "package_perspective" navigation.


efriis:
hey @leo-gan just confirming this is a prototype that doesn't actually do anything, and the expectation is we'd integrate it?

if so I'll probably close, and we can reopen if it's useful in the future. I don't think this makes sense as-is, and ideally it would happen as part of the api docs build process instead

PR #24738, opened 2024-07-27
core[patch]: add necessary @classmethod decorator for methods in `_FewShotPromptTemplateMixin`

ZhangShenao:
add necessary @classmethod decorator for methods in `_FewShotPromptTemplateMixin`


baskaryan:
i dont think these need classmethod decorators, the root_validator should already take care of that. are you seeing some issues with these methods?

PR #24741, opened 2024-07-27
core[patch]: support tool calls with non-pickleable args in tools

ccurme:
Deepcopy raises with non-pickleable args.

### DIFF
--- a/libs/core/langchain_core/tools.py
+++ b/libs/core/langchain_core/tools.py
@@ -20,7 +20,6 @@ tool for the job.
 from __future__ import annotations
 
 import asyncio
-import copy
 import functools
 import inspect
 import json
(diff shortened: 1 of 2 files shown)


hinthornw (approved):

libs/core/langchain_core/tools.py
```diff
@@ -1482,9 +1481,8 @@ def _prep_run_args(
 ) -> Tuple[Union[str, Dict], Dict]:
     config = ensure_config(config)
     if _is_tool_call(input):
-        input_copy = copy.deepcopy(input)
-        tool_call_id: Optional[str] = cast(ToolCall, input_copy)["id"]
-        tool_input: Union[str, dict] = cast(ToolCall, input_copy)["args"]
+        tool_call_id: Optional[str] = cast(ToolCall, input)["id"]
+        tool_input: Union[str, dict] = cast(ToolCall, input)["args"].copy()
```
baskaryan, line 1485: should we try/except deepcopy instead?

baskaryan: maybe i take it back, are we sure we need to copy at all?

ccurme: If I remove `.copy` then this test fails because the config gets added to the tool call args: https://github.com/langchain-ai/langchain/blob/df786087412e40d3c473d61af4e133373e2eaabe/libs/core/tests/unit_tests/test_tools.py#L981-L988

This will generally mutate the AIMessage holding the tool calls as well.

lmk if you see a better solution and can revisit.

baskaryan: should we update how we inject args instead of copyign here?

Issue #24742, opened 2024-07-27
HuggingFaceEndpoint `Endpoint URL` validation Error


keenborder786:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_huggingface import ChatHuggingFace,HuggingFaceEndpoint

os.environ['HUGGINGFACEHUB_API_TOKEN'] = 'xxxxxxxxx'
llm = HuggingFaceEndpoint(
                repo_id="microsoft/Phi-3-mini-4k-instruct",
                task="text-generation",
                max_new_tokens=512,
                do_sample=False,
                repetition_penalty=1.03,
            )

```

### Error Message and Stack Trace (if applicable)

Traceback (most recent call last):
  File "/Users/mac/langchain/test.py", line 18, in <module>
    llm = HuggingFaceEndpoint(
  File "/opt/anaconda3/envs/langchain/lib/python3.9/site-packages/pydantic/v1/main.py", line 341, in __init__
    raise validation_error
pydantic.v1.error_wrappers.ValidationError: 1 validation error for HuggingFaceEndpoint
__root__
  Did not find endpoint_url, please add an environment variable `HF_INFERENCE_ENDPOINT` which contains it, or pass `endpoint_url` as a named parameter. (type=value_error)

### Description

I am trying to initialize the `HuggingFaceEndpoint`, but despite passing the correct `repo_id`, I am encountering an error. I have identified the bug: even though I provide the `repo_id`, the `HuggingFaceEndpoint` validation always checks for the `endpoint_url`, which is incorrect. If the `repo_id` is passed, it should not be checking for the `endpoint_url`. I will create a PR to fix this issue.

### System Info

Package Information
-------------------
> langchain_core: 0.2.24
> langchain: 0.2.11
> langchain_community: 0.2.10
> langsmith: 0.1.93
> langchain_text_splitters: 0.2.0
> langchainhub: 0.1.17

baskaryan:
does the issue only occur if you have the HF_INFERENCE_ENDPOINT env var set?

CrasCris:
Why dont use the Huggingface Hub ? i dont see that error using the hub

[...]

PR #24743, opened 2024-07-27
huggingface: fix `model` param population

keenborder786:
- **Description:** Fix the validation error for `endpoint_url` for HuggingFaceEndpoint. I have given a descriptive detail of the isse in the issue that I have created.
- **Issue:** #24742


keenborder786:
@eyurtsev, @ccurme please review

libs/partners/huggingface/langchain_huggingface/llms/huggingface_endpoint.py
```diff
@@ -145,19 +145,18 @@ def build_extra(cls, values: Dict[str, Any]) -> Dict[str, Any]:
             )
 
         values["model_kwargs"] = extra
-
-        values["endpoint_url"] = get_from_dict_or_env(
-            values, "endpoint_url", "HF_INFERENCE_ENDPOINT", None
-        )
-
-        if values["endpoint_url"] is None and "repo_id" not in values:
+        if "endpoint_url" not in values and "repo_id" not in values:
```
eyurtsev, line 148: Could you update the doc-string for the wrapper to explain which variables can be specified?

----

I haven't used this wrapper, but the new code does not behave correctly as far as I can tell. If repo_id is not strictly necessary for initializaiton, then a user may be relying on endpoint_url being specified via an env variable. This scenario will cause a validation error right now.


> I am trying to initialize the HuggingFaceEndpoint, but despite passing the correct repo_id, I am encountering an error. I have identified the bug: even though I provide the repo_id, the HuggingFaceEndpoint validation always checks for the endpoint_url, which is incorrect. If the repo_id is passed, it should not be checking for the endpoint_url. I will create a PR to fix this issue.

Old validation does not seem to be checking endpoint_url, it seems to be *loading* it from the environment by default.

Possible solutions:
1) Throw an error if both values can be loaded
2) Only try to load_endpoint url after repo_id is missing and only then do the check that at least one of them is specified (this is still too implicit -- so bad solution)
3) Change precedence order if it's obvious that repo_id should be favored over endpoint_url (too implicit -- so bad solution)

eyurtsev (dismissed):
Left a PR explaining the fix. The current fix introduces another bug in a situation where a user is relying for the endpoint URL to be loaded from the env

keenborder786:
@eyurtsev will have a look

libs/partners/huggingface/langchain_huggingface/llms/huggingface_endpoint.py
keenborder786: - I have updated the wrapper's docstring.

- Previously, the `endpoint_url` was first checked from environment variables. If the `endpoint_url` was not set as an environment variable and only `repo_id` was provided, an error would occur despite the `repo_id` being passed. I have now revised this behavior so that the code first checks whether the user has provided a `repo_id` or `endpoint_url` directly. Only after that, if `repo_id` is not provided will it then retrieve the values from the environment variables.

-  See below please.
```python
llm = HuggingFaceEndpoint(
                repo_id="microsoft/Phi-3-mini-4k-instruct", # user intend to use repo_id but validation error will be raised
for `endpoint_url` env variable missing which is wrong since user has passed repo_id
                huggingfacehub_api_token = 'xxxxxx',
                task="text-generation",
                max_new_tokens=512,
                do_sample=False,
                repetition_penalty=1.03,
            )
```

libs/partners/huggingface/langchain_huggingface/llms/huggingface_endpoint.py
```diff
@@ -145,19 +146,18 @@ def build_extra(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
             raise ValueError(
                 "Please specify an `endpoint_url` or `repo_id` for the model."
             )
-        if values["endpoint_url"] is not None and "repo_id" in values:
+        if "endpoint_url" in values and "repo_id" in values:
             raise ValueError(
                 "Please specify either an `endpoint_url` OR a `repo_id`, not both."
             )
+        if "repo_id" not in values:
+            values["endpoint_url"] = get_from_dict_or_env(
```
eyurtsev, line 158: We lost validation on repo_id being unspecified and endpoint_url being unspecified

keenborder786: Nope, the validation is happening above. See Line 149. @eyurtsev

keenborder786:
@eyurtsev please see

keenborder786:
@ccurme

keenborder786:
@ccurme @hwchase17

efriis:
ended up rewriting the implementation to also include the `model` param. This is a lightly breaking change but only for people that were incorrectly populating the `model` param as well as others, which was likely done by mistake.

PR #24754, opened 2024-07-28
Enhance RecursiveCharacterTextSplitter.from_language method

minki-j:
**PR title**: "core: Enhance RecursiveCharacterTextSplitter.from_language method"

**PR message**:
    - **Description:** This PR updates the `from_language` class method of `RecursiveCharacterTextSplitter`. It adds a condition to check if any separator is a newline or space character, and if `keep_separator` is set to True. In such cases, it sets `strip_whitespace=False` in the constructor call. This change prevents it from striping separator later in the logic.


baskaryan:
could we add a unit test to illustrate the change?

minki-j:
@baskaryan I've added a test for this update. Let me know if there is anything I need to add. 

## TEST INSTRUCTION

**Before the PR**
Run the script and confirm that separator, in this case '\n\n', is stripped even though I've set keep_separator as True.

**After the PR**
Run the script and check if the separator (\n\n) persists after splitting. 

**SCRIPT FOR THE TEST**
```
from langchain_core.documents.base import Document
from langchain_text_splitters import (
    Language,
    RecursiveCharacterTextSplitter,
)

documents = [
    Document(
        """g = StateGraph(State)
g.add_node("entry", RunnablePassthrough())
g.set_entry_point("entry")

g.add_edge("entry", n(hypothesis))

g.add_node(n(hypothesis), hypothesis)
g.add_edge(n(hypothesis), n(retrieve_code_snippets))

g.add_node(n(retrieve_code_snippets), retrieve_code_snippets)
g.add_edge(n(retrieve_code_snippets), n(confirmation))

g.add_node(n(confirmation), confirmation)
g.add_edge(n(confirmation), n(update_hypothesis_with_confirmation))
"""
    )
]


python_splitter = RecursiveCharacterTextSplitter.from_language(
    language=Language.PYTHON,
    chunk_size=200,
    chunk_overlap=0,
    keep_separator=True, # python separators: '\nclass ', '\ndef ', '\n\tdef ', '\n\n', '\n', ' ', ''
)

documents_split = python_splitter.split_documents(documents)

print(documents_split[1].page_content) # print the second chunk 
```

minki-j:
@baskaryan Could you check my new comments please?

efriis:
ci is still failing - could you take a look?

minki-j:
@efriis  I've resolved the CI issues. Please rerun the test.

ccurme (commented):
@minki-j Thanks for opening this.

Could you say more about the motivation for retaining whitespace in this case? I'm concerned users might be setting `keep_separator` with the intent of keeping things like `class` and `def` and be surprised by the addition of whitespace to their splits.

Separately, there are some issues with the current implementation. One is that this initialization now breaks:
```python
from langchain_text_splitters import (
    Language,
    RecursiveCharacterTextSplitter,
)

splitter = RecursiveCharacterTextSplitter.from_language(
    language=Language.PYTHON,
    strip_whitespace=True,
    keep_separator=True,
)
```

Another is that there is an inconsistency between initializing via `__init__` and `from_language`.

`from_language`:
```python
from langchain_text_splitters import (
    Language,
    RecursiveCharacterTextSplitter,
)

params = {
    "chunk_size": 10,
    "chunk_overlap": 0,
    "keep_separator": True,
}

splitter = RecursiveCharacterTextSplitter.from_language(
    language=Language.PYTHON,
    **params,
)
    
example = "\n\nclass MyClass"

splitter.split_text(example)
```
```
['\n', '\nclass', ' MyClass']
```

`__init__`:
```python
separators = RecursiveCharacterTextSplitter.get_separators_for_language(
    Language.PYTHON
)

splitter = RecursiveCharacterTextSplitter(
    separators=separators,
    **params
)

splitter.split_text(example)
```
```
['class', 'MyClass']
```
(On master branch these are both just `['class', 'MyClass']`.)

We can fix these things, but ultimately this is a mildly breaking change for users and I think the choice is defensible either way. So would be good to hear the motivation.

minki-j:
Thanks for reviewing this PR, @ccurme

The original method strips whitespaces from each chunk when the code is chunked. When combining adjacent retrieved chunks, these stripped whitespaces cause problems. I just added an option to prevent that from happening. 

Hope this would help you understand the PR. Let me know if it was not clear.

ccurme:
@minki-j could we just pass in `strip_whitespace=False` when calling `.from_language`, if that is desired?

I think this comes down to "what do we do when both `strip_whitespace` and `keep_separators` are True, and separators contain whitespace". Instead of altering the current behavior IMO it's preferable to set `strip_whitespace=False` and strip any unintended whitespace after the fact.

PR #24755, opened 2024-07-28
[RFC] accept with_config(name=...)

hinthornw:
I've made this mistake a couple times and have seen others make this mistake.

Also considered a `warnings.warn()` saying "did you mean 'run_name'. Would be open to that to keep the API more pure.


efriis (commented):
indifferent - doesn't feel like a huge pain point to me. fixable by wrapping args in `dict(...)`, and having 2 diff ways feels like could be confusing

PR #24792, opened 2024-07-29
langchain[patch]: add deprecations

ccurme:


libs/langchain/langchain/agents/agent_toolkits/vectorstore/base.py
```diff
@@ -26,6 +37,40 @@ def create_vectorstore_agent(
...
+    calling methods.
+
+        .. code-block:: python
+
+            from langchain_core.tools import create_retriever_tool
+            from langchain_core.vectorstores import InMemoryVectorStore
+            from langchain_openai import ChatOpenAI, OpenAIEmbeddings
+            from langgraph.prebuilt import create_react_agent
+
+            llm = ChatOpenAI(model="gpt-3.5-turbo-0125", temperature=0)
```
baskaryan, line 50: nit: would use 4o at this pt

libs/langchain/langchain/agents/agent_toolkits/vectorstore/base.py
```diff
@@ -26,6 +37,40 @@ def create_vectorstore_agent(
...
+
+            vector_store = InMemoryVectorStore.from_texts(
+                [
+                    "Dogs are great companions, known for their loyalty and friendliness.",
+                    "Cats are independent pets that often enjoy their own space.",
+                ],
+                OpenAIEmbeddings(),
+            )
+
+            tool = create_retriever_tool(
```
baskaryan, line 60: should we show how to make a qa-with-sources type tool?

libs/langchain/langchain/agents/agent_toolkits/vectorstore/base.py
```diff
@@ -67,6 +122,55 @@ def create_vectorstore_router_agent(
 ) -> AgentExecutor:
     """Construct a VectorStore router agent from an LLM and tools.
 
+    Note: this class is deprecated. See below for a replacement that uses tool
```
baskaryan, line 125: would explicitly callout that its using langgraph, you need to have langgraph installed

libs/langchain/langchain/chains/api/base.py
```diff
@@ -69,7 +79,111 @@ class APIChain(Chain):
...
+                # NOTE: There are inherent risks in giving models discretion
+                # to execute real-world actions. We must "opt-in" to these
+                # risks by setting allow_dangerous_request=True to use these tools.
+                # This can be dangerous for calling unwanted requests. Please make
+                # sure your custom OpenAPI spec (yaml) is safe and that permissions
+                # associated with the tools are narrowly-scoped.
+                ALLOW_DANGEROUS_REQUESTS = True
+
+                # Subset of spec for https://jsonplaceholder.typicode.com
+                api_spec = \"\"\"
```
baskaryan, line 116: does it matter for yaml that these are tabbed

libs/langchain/langchain/chains/combine_documents/map_rerank.py
baskaryan, line 1: are we planning on deprecating?

libs/langchain/langchain/chains/combine_documents/stuff.py
```diff
@@ -95,6 +96,15 @@ def format_docs(inputs: dict) -> str:
     ).with_config(run_name="stuff_documents_chain")
 
 
+@deprecated(
+    since="0.2.13",
+    removal="1.0",
```
baskaryan, line 101: 1.0 or 0.3?

libs/langchain/langchain/chains/mapreduce.py
```diff
@@ -23,6 +24,16 @@
 from langchain.chains.llm import LLMChain
 
 
+@deprecated(
+    since="0.2.13",
+    removal="0.4.0",
```
baskaryan, line 29: seeing 1.0 for some stuff and 0.4 for other

libs/langchain/langchain/chains/combine_documents/stuff.py
ccurme: It was decided to save 0.3 for pydantic changes and not remove anything in 0.3. After that decision, some deprecations were set for a 0.4+ removal.

In June it was then decided to remove chains in 1.0 (right after that discussion I updated it for LLMChain in https://github.com/langchain-ai/langchain/pull/23373).

I have a separate todo to update all removal targets to 1.0. OK to revisit this decision and keep some in 0.4 if we like.

libs/langchain/langchain/agents/agent_toolkits/vectorstore/base.py
ccurme: thanks, changed this everywhere

libs/langchain/langchain/agents/agent_toolkits/vectorstore/base.py
ccurme: thanks, done here and elsewhere

libs/langchain/langchain/chains/api/base.py
ccurme: fixed, ty

libs/langchain/langchain/chains/combine_documents/map_rerank.py
ccurme: Acting alone I would not deprecate until we have a constructor. Same is true for map reduce, refine, etc.

Related, we have a [summarization tutorial](https://python.langchain.com/v0.2/docs/tutorials/summarization/) that uses these legacy chains. We need to update this doc. I think if we stuff the different langgraph implementations into the doc, it will be too complex and readers may lose the point.

Some options:
- Have separate guides and link out to them (e.g., a section of how-to guides) instead.
- Add langgraph to langchain deps, add constructors, and update the guide.
- ???

libs/langchain/langchain/chains/mapreduce.py
ccurme: thanks

PR #24795, opened 2024-07-29
cli/docs: llm integration template standardization

efriis:


libs/cli/langchain_cli/integration_template/docs/llms.ipynb
baskaryan, line 1: maybe should wait to use LANGSMITH_API_KEY and LANGSMITH_TRACING till langsmith docs do so (which requires js sdk supporting them)

efriis: will switch to `LANGCHAIN_API_KEY` and `LANGCHAIN_TRACING_V2`

PR #24798, opened 2024-07-29
docs[patch]: Adds key-value store to conceptual guide

jacoblee93:


baskaryan (commented):
can we rename the section in integrations from "Stores" to "Key-value stores" as part of this

jacoblee93:
> can we rename the section in integrations from "Stores" to "Key-value stores" as part of this

Was going to do in a followup but can do here instead as well

PR #24801, opened 2024-07-30
add langchain_ollama to universal chat model

isahers1:


libs/langchain/langchain/chat_models/base.py
```diff
@@ -336,8 +336,8 @@ def _init_chat_model_helper(
 
         return ChatFireworks(model=model, **kwargs)
     elif model_provider == "ollama":
-        _check_pkg("langchain_community")
-        from langchain_community.chat_models import ChatOllama
+        _check_pkg("langchain_ollama")
+        from langchain_ollama import ChatOllama
```
baskaryan, line 340: for backwards compatibility we should probably try/except this and fallback to community version

jacoblee93:
Closing in favor of #24821

PR #24804, opened 2024-07-30
docs[patch]: Adds kv store integration docs template

jacoblee93:


eyurtsev:
1) In the overview section, could we ask for a reference to the underlying technology or API somewhere?
2) For feature table why do we want Serializable?
3) For credentials -> what do we do if service may not require any and can be launched via docker container? (e.g., redis)
4) For usage, let's add mdelete as well?

eyurtsev (approved):
Looks great!

PR #24826, opened 2024-07-30
docs: Add pydantic compatibility to side bar

eyurtsev:
Add pydantic compatibility to side bar

### DIFF
--- a/docs/sidebars.js
+++ b/docs/sidebars.js
@@ -78,7 +78,7 @@ module.exports = {
         {
           type: 'doc',
           id: "how_to/pydantic_compatibility",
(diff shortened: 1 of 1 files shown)


docs/sidebars.js
```diff
@@ -75,6 +75,11 @@ module.exports = {
         "versions/overview",
         "versions/release_policy",
         "versions/packages",
+        {
+          type: 'doc',
+          id: "docs/how_to/pydantic_compatibility",
```
eyurtsev, line 80:
```suggestion
          id: "how_to/pydantic_compatibility",
```

PR #24827, opened 2024-07-30
pydantic migration v3 -- w/ individual commits

eyurtsev:
This PR is just a reference PR that attempts to discover all the possible points of failure during this migration.

We'll be cherry-picking some commits from this branch to master.

In September, this PR will be used as reference to kick off a 0.3 branch.


eyurtsev:
reference:

    @classmethod
    def model_parametrized_name(cls, params: Tuple[Type[Any], ...]) -> str:
        return f'{params[0].__name__.title()}Response'

PR #24858, opened 2024-07-31
[docs]: vector store integration pages

isahers1:


docs/docs/integrations/vectorstores/faiss_index/index.faiss
efriis, line 1: do we want to commit this or remove?

efriis (approved):
small nit otherwise lgtm

docs/docs/integrations/vectorstores/faiss_index/index.faiss
isahers1: remove I think, it was leftover from a previous PR. unless I'm missing something it's just a file generated when you run the notebook

efriis: nice added a .gitignore so it doesn't get committed back

Issue #24660, opened 2024-07-25
Model with history work well on `invoke`, but not well in `stream` (many parts exactly same to official tutorial `Build a Chatbot`)


SuperKevinZhou:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

I've tried this code on 2 platforms(JetBrains Datalore Online and Replit), and they both give me the same error.
```py
# -*- coding: utf-8 -*-

# Some API KEY and model name
GROQ_API_KEY = "MY_GROQ_KEY"# I have filled this, no problem in this
llm_name = "llama3-groq-70b-8192-tool-use-preview"

# Import
from langchain_groq import ChatGroq
from langchain_core.messages import AIMessage, SystemMessage, HumanMessage

from langchain_core.chat_history import (
    BaseChatMessageHistory,
    InMemoryChatMessageHistory,
)
from langchain_core.runnables.history import RunnableWithMessageHistory

# Chat History Module
store = {}

# The exactly same code in the tutorial
def get_session_history(session_id: str) -> BaseChatMessageHistory:
    if session_id not in store:
        store[session_id] = InMemoryChatMessageHistory()
    return store[session_id]

model = ChatGroq(
    model = llm_name,
    temperature = 0.5,
    max_tokens = 1024,
    stop_sequences = None,
    api_key = GROQ_API_KEY
)

with_message_history = RunnableWithMessageHistory(model, get_session_history)

# Session ID
config = {"configurable": {"session_id": "abc"}}

model.invoke([HumanMessage(content = "Hi! My name's Kevin.")])

# Stream: I fail in this

for chunk in with_message_history.stream(
    [HumanMessage(content = "What's my name?")],
    config = config,
):
    print(chunk.content, end = '')

print()

print("Done!")

# Invoke: This works well just as I want
response = with_message_history.invoke(
    [HumanMessage(content="Hi! I'm Bob")],
    config=config,
)

print(response.content)# This works
```

### Error Message and Stack Trace (if applicable)

Your name is Kevin.
Done!
Error in RootListenersTracer.on_chain_end callback: ValueError()
Error in callback coroutine: ValueError()

### Description

[...]

PR #24862, opened 2024-07-31
core: runnable config ensure_config deep copy from var_child_runnable…

wulifu2hao:
**issue**: #24660 
  RunnableWithMessageHistory.stream result in error because the [evaluation](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/runnables/branch.py#L220) of the branch [condition](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/history.py#L328C1-L329C1) unexpectedly trigger the "[on_end](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/history.py#L332)" (exit_history) callback of the default branch 

**descriptions**
  After a lot of investigation I'm convinced that the root cause is that 
  1. during the execution of the runnable, the [var_child_runnable_config](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/config.py#L122) is shared between the branch [condition](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/history.py#L328C1-L329C1) runnable  and the [default branch runnable](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/history.py#L332) within the same context 
  2.  when the default branch runnable runs, it gets the [var_child_runnable_config](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/config.py#L163) and may unintentionally [add more handlers ](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/config.py#L325)to the callback manager of this config
   3. when it is again the turn for the [condition](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/history.py#L328C1-L329C1) to run, it gets the  `var_child_runnable_config` whose callback manager has the handlers added by the default branch. When it runs that handler (`exit_history`) it leads to the error 
   
   with the assumption that, the `ensure_config` function actually does want to create a immutable copy from `var_child_runnable_config` because it starts with an [`empty` variable ](https://github.com/langchain-ai/langchain/blob/99eb31ec4121e146e007ea584b2d5f4aff2a4337/libs/core/langchain_core/runnables/config.py#L156), i go ahead to do a deepcopy to ensure that future modification to the returned value won't affect the `var_child_runnable_config` variable
   
   Having said that I actually 
1. don't know if this is a proper fix
2. don't know whether it will lead to other unintended consequence 
3. don't know why only "stream" runs into this issue while "invoke" runs without problem

so @nfcampos @hwchase17 please help review, thanks!


libs/core/langchain_core/runnables/config.py
```diff
@@ -161,9 +162,10 @@ def ensure_config(config: Optional[RunnableConfig] = None) -> RunnableConfig:
         configurable={},
     )
     if var_config := var_child_runnable_config.get():
-        empty.update(
-            cast(RunnableConfig, {k: v for k, v in var_config.items() if v is not None})
+        deep_copied_config = copy.deepcopy(
+            {k: v for k, v in var_config.items() if v is not None}
         )
+        empty.update(cast(RunnableConfig, deep_copied_config))
```
nfcampos, line 168:
```suggestion
        empty = merge_configs(empty, {k: v for k, v in var_config.items() if v is not None})
```

wulifu2hao: Hi @nfcampos thanks for the suggestion

If I follow your suggestion, I wonder
1. if I can still achieve the purpose of getting a "safe to mutate" copy of the config ? For example in `"k, v" of var_config.items()` could be the callbacks, and the reader of it can unintended mutation of the fields in it like the "handlers" 
2. if calling the "merge_configs" function in "ensure_config" will lead to infinite recursion? since merge_configs also calls ensure_config

nfcampos: hi, on the first point, notice that merge_configs doesn't mutate any of the parts of config it modifies, it create new copies. on the 2nd point, good point, I'll make a version of merge_configs that doesn't call ensure_config

libs/core/langchain_core/runnables/config.py
```diff
@@ -291,7 +307,7 @@ def merge_configs(*configs: Optional[RunnableConfig]) -> RunnableConfig:
                     **(config.get(key) or {}),  # type: ignore
                 }
             elif key == "tags":
-                base[key] = list(  # type: ignore
+                base[key] = sorted(  # type: ignore
```
baskaryan, line 310: ooc is this related to fix

baskaryan (commented):
for mutable vals (callbacks, whatevers in configurable), they can be mutated right? should we document this?

libs/core/langchain_core/runnables/config.py
nfcampos: set has undefined order, so it was a mistake to pass things through set without sorting them after

baskaryan (approved):

wulifu2hao:
Hi @nfcampos @baskaryan thanks for the effort to make changes and merge this PR.

However the original issue is not resolved... As I've explained in the original description of the PR, the issue is that without doing a real deepcopy of [var_config](https://github.com/langchain-ai/langchain/blob/245cb5a252c985986d52a7a5c161991c62d5ee56/libs/core/langchain_core/runnables/config.py#L170) , any subsequent mutation of the returned value of the [ensure_config](https://github.com/langchain-ai/langchain/blob/245cb5a252c985986d52a7a5c161991c62d5ee56/libs/core/langchain_core/runnables/config.py#L153) function like [that](https://github.com/langchain-ai/langchain/blob/245cb5a252c985986d52a7a5c161991c62d5ee56/libs/core/langchain_core/runnables/config.py#L338) in the merge_configs function will end up unintentionally mutate the underlying value of [var_child_runnable_config](https://github.com/langchain-ai/langchain/blob/245cb5a252c985986d52a7a5c161991c62d5ee56/libs/core/langchain_core/runnables/config.py#L129) within the same context

The below example illustrate the idea 
```
from langchain_core.callbacks import BaseCallbackManager, BaseCallbackHandler

original = BaseCallbackManager(
    handlers=[],
    inheritable_handlers=[],
)

print(original.handlers)
# output []


copied = original.copy()
copied.handlers.append(
    BaseCallbackHandler()
)

print(original.handlers)
# output [<langchain_core.callbacks.base.BaseCallbackHandler object at 0x100b39fd0>] , the original.handlers are modified

```

I hope that clarifies. If the right solution is not to create a deep copy to make it "mutation proof", then some discipline needs to be exercised not to mutate some underlying fields by the reader of the var_child_runnable_config variable

PR #24863, opened 2024-07-31
community: fix issue of the existence of numeric object in `additional_kwargs` a…

Dobiichi-Origami:
- **Description:** A previous PR breaks the code from `baidu_qianfan_endpoint.py` which causes the malfunction of streaming

### DIFF
--- a/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
+++ b/libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
@@ -654,6 +654,15 @@ class QianfanChatEndpoint(BaseChatModel):
                         role="assistant",
                         additional_kwargs=additional_kwargs,
                         usage_metadata=msg.usage_metadata,
+                        tool_call_chunks=[
+                            tool_call_chunk(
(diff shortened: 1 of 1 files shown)


Dobiichi-Origami:
PTAL @efriis

Dobiichi-Origami:
@ccurme

libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
```diff
@@ -613,6 +617,15 @@ def _stream(
                         role="assistant",
                         additional_kwargs=additional_kwargs,
                         usage_metadata=msg.usage_metadata,
+                        tool_call_chunks=[
```
ccurme, line 620: Should we add this to `_astream` below, as well?

ccurme (approved):
Thank you

libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
Dobiichi-Origami: oh, I forgot that

PR #24866, opened 2024-07-31
fix create_xml_agent with parameterless function as a tool

QunBB:
When using `create_xml_agent` or `create_json_chat_agent` to create a agent, and the function corresponding to the tool is a parameterless function, the `XMLAgentOutputParser` or `JSONAgentOutputParser` will parse the tool input into an empty string, `BaseTool` will parse it into a positional argument.
So, the program will crash finally because we invoke a parameterless function but with a positional argument.
```python
from langchain import hub
from langchain.agents import AgentExecutor, create_json_chat_agent, create_xml_agent

prompt = hub.pull("hwchase17/react-chat-json")

llm = ChatOpenAI()

# agent = create_xml_agent(llm, tools, prompt)
agent = create_json_chat_agent(llm, tools, prompt)

agent_executor.invoke(......)
```


baskaryan (commented):
could we add unit tests for this change?

QunBB:
@baskaryan 
I will re-push the unit tests with new pull request, the file of this change had removed and transfered. 
I think i should close this pull request because of my misoperation.

PR #24897, opened 2024-07-31
community[patch]: Make the pydantic linter stricter

eyurtsev:
Stricter linting of deprecated pydantic features.


eyurtsev:
Used to catch stuff like this: https://github.com/langchain-ai/langchain/pull/24896

PR #24918, opened 2024-08-01
patch[cli] Fix bug in `check_imports.py`

ZhangShenao:
The variable `has_failure` in check_imports.py is wrong-declared. It's actually an another variable.


ccurme (approved):
Thanks!

PR #24936, opened 2024-08-01
core[patch]: Fix tool args schema inherited field parsing

### LINKED ISSUE #24925: BaseTool's `tool_call_schema` ignores inherited fields of  an `args_schema`, causing incomplete tool inputs
### Checked other resources

### Example Code

```python
from langchain.pydantic_v1 import BaseModel, Field

class ModelA(BaseModel):
    field_a: str = Field(description='Base class field')

class ModelB(ModelA):
    field_b: str = Field(description='Subclass class field')
    
mytool = tool(func, args_schema=ModelB)
```

_No response_

### Description

Hi, 

I noticed that in the current version of langchain_core, tools using an args_schema have incomplete inputs if the schema is derived from a superclass. That is because as of recently there is a property `tool_call_schema`, which creates a schema with only "non-injected" fields: https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/tools.py#L387
However, it derives the field names to be retained from the `__annotations__` property of the schema, which does not inherit fields of the base class. Hence, all fields from the base class (ModelA in the example above) are deleted. This causes incomplete tool inputs when using schemas that use inheritance.

Is this a regression or should the schemas be used differently?

Thanks!

Valentin

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  #1 SMP PREEMPT_DYNAMIC Mon Jul 15 21:39:34 UTC 2024
> Python Version:  3.11.4 (main, Jul 30 2024, 10:36:58) [GCC 14.1.1 20240522]

Package Information
-------------------
> langchain_core: 0.2.26
> langchain: 0.2.11
> langchain_community: 0.2.10
> langsmith: 0.1.94
> langchain_openai: 0.1.19
> langchain_text_splitters: 0.2.2
> langserve: 0.2.2

baskaryan:
Fix #24925


eyurtsev (approved):

libs/core/langchain_core/tools.py
```diff
@@ -1650,3 +1650,11 @@ def _filter_schema_args(func: Callable) -> List[str]:
         filter_args.append(config_param)
     # filter_args.extend(_get_non_model_params(type_hints))
     return filter_args
+
+
+def _get_all_basemodel_annotations(cls: TypeBaseModel) -> Dict[str, Type]:
+    annotations: Dict[str, Type] = {}
+    for parent in inspect.getmro(cls):
+        if is_basemodel_subclass(parent):
+            annotations = {**getattr(parent, "__annotations__", {}), **annotations}
```
eyurtsev, line 1659: This could fail with forward refs

PR #24941, opened 2024-08-01
docs: fix kv store column headers

ccurme:
Screenshot 2024-08-01 at 12 32 19 PM


ccurme:
After:
Screenshot 2024-08-01 at 12 48 32 PM

jacoblee93:
Thank you!

PR #24946, opened 2024-08-01
core[minor]: Add get_input_jsonschema, get_output_jsonschema, get_config_jsonschema

eyurtsev:
This PR adds methods to directly get the json schema for inputs, outputs, and config.
Currently, it's delegating to the underlying pydantic implementation, but this may be changed in the future to be independent of pydantic.


libs/core/langchain_core/runnables/base.py
```diff
@@ -341,6 +378,43 @@ def get_output_schema(
...
+            .. code-block:: python
+
+                from langchain_core.runnables import RunnableLambda
+
+                def add_one(x: int) -> int:
+                    return x + 1
+
+                runnable = RunnableLambda(add_one)
+
+                print(runnable.get_input_jsonschema())
```
eyurtsev, line 403:
```suggestion
                print(runnable.get_output_jsonschema())
```

eyurtsev:
Merged here: https://github.com/langchain-ai/langchain/pull/26034

Issue #24957, opened 2024-08-02
AthenaLoader profile type hinting is incorrect


jimwbaldwin:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
        athena_loader = AthenaLoader(
            query=f"SELECT 1",
            database="default",
            s3_output_uri="s3://fake-bucket/fake-prefix",
            profile_name=None,
        )
```

This code works but the type hinting is incorrect which results in error warnings from type checkers.

### Error Message and Stack Trace (if applicable)

![image](https://github.com/user-attachments/assets/d23beaa3-3ba8-4701-8b50-5c79d12cd61f)

### Description

The Athena loader has code to handle a non-profile. I think should be an optional kwarg like this:
```python
profile_name: Optional[str] = None,
```

The code here shows that `None` is actually handle and is a valid input.
https://github.com/langchain-ai/langchain/blob/d7688a4328f5d66f3b274db6e7b024a24b15cc8e/libs/community/langchain_community/document_loaders/athena.py#L62-L67

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000
> Python Version:  3.11.4 (main, Mar 26 2024, 16:28:52) [Clang 15.0.0 (clang-1500.3.9.4)]

Package Information
-------------------
> langchain_core: 0.2.25
> langchain: 0.2.11
> langchain_community: 0.2.10
> langsmith: 0.1.94
> langchain_openai: 0.1.19
> langchain_postgres: 0.0.9
> langchain_text_splitters: 0.2.2
> langgraph: 0.1.17

Packages not installed (Not Necessarily a Problem)
--------------------------------------------------
The following packages were not found:

> langserve

jimwbaldwin:
I'm happy to raise a PR for this. Just raising the issue before I do that.

PR #24958, opened 2024-08-02
community: make AthenaLoader profile_name optional and fix type hint

jimwbaldwin:
- **Description:** This PR makes the AthenaLoader profile_name optional and fixes the type hint which says the type is `str` but it should be `str` or `None` as None is handled in the loader init. This is a minor problem but it just confused me when I was using the Athena Loader to why we had to use a Profile, as I want that for local but not production.
- **Issue:** #24957 
- **Dependencies:** None.


ccurme (approved):

jimwbaldwin:
I am running make tests locally and its fine. Not sure why it failed earlier but its passing now.

image

Not sure how to make it build now.

PR #24979, opened 2024-08-02
standard-tests: set integration test parameters independent of unit test

ccurme:
This ends up getting set in integration tests.


efriis (approved):
nice. Might be worth adding `"model": "mistral-small"` to demonstrate desired behavior of passing explicit model

efriis:
Read it wrong. This is general not mistral

Issue #24800, opened 2024-07-30
Standardize vector store docs


isahers1:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

## Issue

To make our vector store integrations as easy to use as possible we need to make sure the docs for them are thorough and standardized. There are two parts to this: updating the vector store docstrings and updating the actual integration docs.

This needs to be done for each VectorStore integration, ideally with one PR per VectorStore.

Related to broader issues #21983 and #22005.

## Docstrings

Each VectorStore class docstring should have the sections shown in the [Appendix](#appendix) below. The sections should have input and output code blocks when relevant.

To build a preview of the API docs for the package you're working on run (from root of repo):

```bash
make api_docs_clean; make api_docs_quick_preview API_PKG=openai
```

where `API_PKG=` should be the parent directory that houses the edited package (e.g. community, openai, anthropic, huggingface, together, mistralai, groq, fireworks, etc.). This should be quite fast for all the partner packages.

## Doc pages

Each vector store [docs page](https://python.langchain.com/v0.2/docs/integrations/vectorstores/) should follow [this template](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/vectorstores.ipynb). 

You can use the `langchain-cli` to quickly get started with a new vectorstore integration docs page (run from root of repo):

```bash
poetry run pip install -e libs/cli
poetry run langchain-cli integration create-doc --name "foo-bar" --name-class FooBar --component-type VectorStore --destination-dir ./docs/docs/integrations/vectorstores/
```

where `--name` is the integration package name without the "langchain-" prefix and `--name-class` is the class name without the "VectorStore" prefix. This will create a template doc with some autopopulated fields at docs/docs/integrations/vectorstores/foo_bar.ipynb.

To build a preview of the docs you can run (from root):

```bash
make docs_clean
make docs_build
cd docs/build/output-new
yarn
yarn start
```

## Appendix

Expected sections for the VectorStore class docstring.

```python
    """__ModuleName__ vector store integration.

    Setup:
      ...

    Key init args - indexing params:
      ...

    Key init args - client params:
      ...

    See full list of supported init args and their descriptions in the params section.

    Instantiate:
      ...

    Add Documents:
      ...

    Update Documents:
      ...

    Delete Documents:
      ...

    Search:
      ...

    Search with score:
      ...

    Use as Retriever:
      ...

[...]

PR #25003, opened 2024-08-02
community: add sqlite-vec vectorstore

abhiaagarwal:
**Description**:

Adds a vector store integration with [sqlite-vec](https://alexgarcia.xyz/sqlite-vec/), the successor to sqlite-vss that is a single C file with no external dependencies.

Pretty straightforward, just copy-pasted the sqlite-vss integration and made a few tweaks and added integration tests. Only question is whether all documentation should be directed away from sqlite-vss if it is defacto deprecated (cc @asg017).


libs/community/langchain_community/vectorstores/sqlitevec.py
```diff
@@ -0,0 +1,242 @@
...
+        metadatas: Optional[List[dict]] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        """Add more texts to the vectorstore index.
+        Args:
+            texts: Iterable of strings to add to the vectorstore.
+            metadatas: Optional list of metadatas associated with the texts.
+            kwargs: vectorstore specific parameters
+        """
+        max_id = self._connection.execute(
```
asg017, line 121: would a `RETURNING` clause make more sense here, instead of `max_id`? Just to avoid multiple SQL statements, and would avoid subtle bugs with SQLite autoincrement values

asg017:
Hey thanks for putting this up!

I will be deprecating `sqlite-vss` very soon, `sqlite-vec` is an improvement in nearly every single way and having two similar extensions doesn't make much sense. I'll just need to put together a migration guide before archiving the repo. 

Also to note: I'll be adding proper metadata columns + filtering in the near future, which should simplify a lot of code here, but that would be for a little longer

Happy to take a closer look if needed

libs/community/langchain_community/vectorstores/sqlitevec.py
abhiaagarwal: I wasn't even aware of RETURNING (I copied this impl directly from the sqlite-vss), I'll try that! That also means I can remove the trigger and avoid having two copies of the embedding in the database.

philippe2803:
@abhiaagarwal I actually submitted a PR doing the actual same thing here: https://github.com/langchain-ai/langchain/pull/25024 and did not pay attention someone else had already done it less than a day before me! I will close my PR then. Congratulation for beating me to it!

liunux4odoo:
Excellent work.
An advice: it could be problematic if  use raw sqlite connection directly, maybe it's better to use sqlalchemy as langchain-postgres does? 
Also langchain-postgres would be a good reference that supports async methods and metadata filtering,

abhiaagarwal:
@liunux4odoo I'm totally open to that — I've been trying to figure out an async client and it's been not fun. Let me try the sqlalchemy approach!

philippe2803:
@abhiaagarwal I feel your PR should go forward as it is for now, it's perfectly reasonable as it is. Sqlite-vec is a major vector store alternative that needs to be reflected in Langchain. I don't think we should wait.

@liunux4odoo I like the idea of integrating it with SQLAlchemy but this feels like a major disruption for the purpose of this PR. I don't think the need was ever expressed with sqlite-vss to have SQLAlchemy, so how about we just go ahead without it for now, and if the need comes along we do a refacto for it?

@asg017 I was completely unaware of the RETURNING feature (thanks for bringing that up). So I did try to implement it, but unfortunately got some error. It seems that [RETURNING is not available (yet?) when using `cursor.executemany` method](https://github.com/python/cpython/issues/100021).

libs/community/langchain_community/vectorstores/sqlitevec.py
philippe2803: I was completely unaware of the RETURNING clause as well. Unfortunately, this does not seem to be working with the sqlite3 `executemany` method as pointed here: https://github.com/python/cpython/issues/100021

philippe2803 (approved):

philippe2803:
@baskaryan could this PR be looked at and merged soon?

abhiaagarwal:
@philippe2803 yep, I discovered the RETURNING thing as well with executemany, I can actually solve it I believe by just doing multiple executes with RETURNING. I haven't benchmarked the performance characteristics yet, but it's probably not worth it.

efriis:
Howdy! Looks good other than the doc - could you rewrite to match the specified format? There's a cli command that bootstraps it here: https://github.com/langchain-ai/langchain/issues/24800

philippe2803:
@abhiaagarwal would you be OK to give me permission to the sqlite-vec branch on your langchain fork? I have made the changes (I believe) necessary for the docs build to pass. This would get this PR merged shortly.

abhiaagarwal:
@philippe2803 absolutely, I just did it, let me know if it works for you!

philippe2803:
@efriis looks like this is all passing now. I believe this PR can now be merged.

philippe2803:
@efriis sorry for being pushy, but could this PR be merged? It's been pending for quite some time now, and every GHA are currently passing. At your disposal if I can help in any ways.

philippe2803:
@baskaryan @efriis could we please move forward with this PR please? it has been hanging for over a week now, although everything has passed. Happy to help in any ways

DecisionNerd:
@efriis anything the community can do to push this forward?

efriis:
Hey team! Linking the [review process](https://python.langchain.com/docs/contributing/review_process/) - the status is currently "needs support" so can give it an upvote to get it through!

efriis (approved):

ajram23:
@efriis Thank you! Appreciate the approval!
 one quick question: Will this package be available for langchain 2.X or will it require to migrate to 3.X.

efriis:
will require 0.3 because the current langchain-community library depends on pydantic v2 in other areas!

efriis:
https://python.langchain.com/docs/versions/v0_3/

PR #25009, opened 2024-08-03
core[patch]: Remove default BaseModel init docstring

baskaryan:
Currently a default init docstring gets appended to the class docstring of every BaseModel inherited object. This removes the default init docstring.

Screenshot 2024-08-02 at 5 09 55 PM


libs/core/langchain_core/load/serializable.py
```diff
@@ -107,6 +107,11 @@ class Serializable(BaseModel, ABC):
         as part of the serialized representation.
     """
 
+    # Remove default BaseModel init docstring.
+    def __init__(self, **kwargs: Any) -> None:
+        """"""
+        super().__init__(**kwargs)
```
efriis, line 113:
```suggestion
    def __init__(self, *args: Any, **kwargs: Any) -> None:
        """"""
        super().__init__(*args, **kwargs)
```
Think ya might need this

efriis (commented):
love it!

efriis (approved):
ok fine u win

PR #25028, opened 2024-08-03
Partner: Milvus - fixed bug when using partition key and dynamic fields together

stuartkmarsh:
**Description:**

This PR fixes a bug where if `enable_dynamic_field` and `partition_key_field` are enabled at the same time, a pymilvus error occurs. 

 Milvus requires the partition key field to be a full schema defined field, and not a dynamic one, so it will throw the error "the specified partition key field {field} not exist" when creating the collection.

When `enabled_dynamic_field` is set to `True`, all schema field creation based on `metadatas` is skipped.  This code now checks if `partition_key_field` is set, and creates the field.

Integration test added.

**Twitter handle:** StuartMarshUK


zc277584121:
LGTM, thanks you @stuartkmarsh  for your contribution , it is a perfect fix

Issue #25031, opened 2024-08-04
Pydantic field metadata not being serialized in tool calls


gramliu:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from pydantic import BaseModel, Field
from typing import List
from langchain_core.tools import tool
from langchain_anthropic import ChatAnthropic
from langchain_core.messages import AIMessage
from dotenv import load_dotenv

load_dotenv()

class SampleModel(BaseModel):
    numbers: List[int] = Field(description="Favorite numbers", min_length=10, max_length=15)

@tool(args_schema=SampleModel)
def choose_numbers():
    """Choose your favorite numbers"""
    pass

model = ChatAnthropic(model_name="claude-3-haiku-20240307", temperature=0)
model = model.bind_tools([choose_numbers], tool_choice="choose_numbers")

result: AIMessage = model.invoke("Hello world!")
print(result.tool_calls[0]["args"])
# Output: {'numbers': [7, 13, 42]}
```

### Description

When additional metadata is specified on Pydantic fields, e.g. `min_length` or `max_length`, these aren't serialized into the final output that's sent out to the model. In the attached screenshot from Langsmith, although it carries over the description, the `length` params are missing.

https://smith.langchain.com/public/62060327-e5be-4156-93cb-6960078ec7fb/r

<img width="530" alt="image" src="https://github.com/user-attachments/assets/4fa2a6b6-105b-43c2-957f-f56b84fef10b">

### System Info

```
langchain==0.2.11
langchain-anthropic==0.1.21
langchain-community==0.2.10
langchain-core==0.2.24
langchain-google-genai==1.0.8
langchain-google-vertexai==1.0.7
langchain-openai==0.1.19
langchain-text-splitters==0.2.2
````

**Comment by keenborder786** [comment]
No, it is being serialized, you just need to specify the constraint as following and it should work:

```python

class SampleModel(BaseModel):
    numbers: conlist(int,min_items=2,max_items=4)  = Field(description="Favorite numbers")

@tool(args_schema=SampleModel)
def choose_numbers():
    """Choose your favorite numbers"""
    pass


```

gramliu:
[...]

PR #25032, opened 2024-08-04
core: Add pydantic metadata to subset model

gramliu:
- **Description:** This includes Pydantic field metadata in `_create_subset_model_v2` so that it gets included in the final serialized form that get sent out.
- **Issue:** #25031 
- **Dependencies:** n/a
- **Twitter handle:** @gramliu


eyurtsev:
@baskaryan re-assigning to you since youv'e started tackling the issue.

@gramliu what problem does this solve? Is the goal to render the metadata in langsmith specifically? or is there a different purpose for this change.

gramliu:
@eyurtsev the pydantic metadata include additional field constraints like min/max length on arrays. This is the main use case I'm trying to address

baskaryan:
thanks for flagging and fixing @gramliu!

PR #25040, opened 2024-08-04
partners: MongoDB Add Hybrid and Full-Text Search Retrievers (Deprecated)

caseyclements:
## Description

This pull-request extends the existing vector search strategies of MongoDBAtlasVectorSearch to include Hybrid (Reciprocal Rank Fusion) and Full-text via new Retrievers.

There is a small breaking change in the form of the `prefilter` kwarg to search. For this, and because we have now added a great deal of features since 0.1.0, we plan to bump the version to 0.2.0.

### Checklist
* Unit tests have been extended
* formatting has been applied
* Two mypy errors remain which will either go away in CI or be simplified.


efriis (commented):
If you could address merge conflicts that would be great - I think some of the types might have changed regarding filters - just want to make sure that's intentional!

caseyclements:
Of course. Will do tomorrow morning.

PR #25044, opened 2024-08-05
core: autodetect more ls params

efriis:


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -473,9 +473,37 @@ def _get_ls_params(
...
+            ls_params["ls_model_name"] = self.model_name
+
+        # temperature
+        if hasattr(self, "temperature") and isinstance(self.temperature, float):
+            ls_params["ls_temperature"] = self.temperature
+        elif "temperature" in kwargs and isinstance(kwargs["temperature"], float):
+            ls_params["ls_temperature"] = kwargs["temperature"]
+
+        # max_tokens
+        if hasattr(self, "max_tokens") and isinstance(self.max_tokens, int):
```
efriis, line 502: could also look for `max_tokens_to_sample` (community anthropic) but might be better to just override in that class

one other place to look for all of these is in `self.model_kwargs` if it exists

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -473,9 +473,37 @@ def _get_ls_params(
         **kwargs: Any,
     ) -> LangSmithParams:
         """Get standard params for tracing."""
-        ls_params = LangSmithParams(ls_model_type="chat")
+
+        # get default provider from class name
+        default_provider = self.__class__.__name__
+        if default_provider.startswith("Chat"):
+            default_provider = default_provider[4:]
```
ccurme, line 480:
```suggestion
            default_provider = default_provider[4:].lower()
```

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -473,9 +473,37 @@ def _get_ls_params(
...
+        # model
+        if hasattr(self, "model") and isinstance(self.model, str):
+            ls_params["ls_model_name"] = self.model
+        elif hasattr(self, "model_name") and isinstance(self.model_name, str):
+            ls_params["ls_model_name"] = self.model_name
+
+        # temperature
+        if hasattr(self, "temperature") and isinstance(self.temperature, float):
+            ls_params["ls_temperature"] = self.temperature
+        elif "temperature" in kwargs and isinstance(kwargs["temperature"], float):
```
ccurme, line 498: here and below: should we prefer parameter in kwargs over the (default) attribute?

efriis: smart

ccurme (approved):

PR #25057, opened 2024-08-05
mongodb: Add Hybrid and Full-Text Search Retrievers, release 0.1.8

caseyclements:
## Description

This pull-request extends the existing vector search strategies of MongoDBAtlasVectorSearch to include Hybrid (Reciprocal Rank Fusion) and Full-text via new Retrievers.

There is a small breaking change in the form of the `prefilter` kwarg to search. For this, and because we have now added a great deal of features, including programmatic Index creation/deletion since 0.1.0, we plan to bump the version to 0.2.0.

### Checklist
* Unit tests have been extended
* formatting has been applied
* One mypy error remains which will either go away in CI or be simplified.


libs/partners/mongodb/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "langchain-mongodb"
-version = "0.1.7"
+version = "0.2.0"
```
efriis, line 7:
```suggestion
version = "0.1.8"
```
This is not a breaking change right?

efriis (approved):

libs/partners/mongodb/pyproject.toml
caseyclements: Prakul's convinced me that making a breaking change for a small improvement in UX isn't worth it, so I'll only bump to 0.1.8 and roll back type of pre_filters. Doing this afternoon or tomorrow morning.

caseyclements:
Allow me to resolve these conflicts. One file, in the docstring.

prakul:
@efriis, this looks good to be merged now?

efriis (approved):

PR #25061, opened 2024-08-05
docs: versions sidebar

baskaryan:


docs/docs/versions/overview.mdx
```diff
@@ -1,9 +1,9 @@
 ---
 sidebar_position: 0
-sidebar_label: Overview
+sidebar_label: Overview of v0.2
```
eyurtsev, line 3: What about 0.2 Overview?

docs/docs/versions/v0_2/index.mdx
```diff
@@ -2,7 +2,7 @@
 sidebar_position: 1
 ---
 
-# LangChain v0.2
+# Migrating to LangChain v0.2
```
eyurtsev, line 5: The titles become long in the side bar? Is `Migrating to 0.2`  reasonable?

baskaryan: that is what it is, the sidebar label for this one is set in sidebar.js

PR #24565, opened 2024-07-23
add retriever to indexer

eyurtsev:
* Combine indexer and retriever; i.e., make the index searchable
* Create a one shot variant of an indexer
* Create abstraction for generalized vectore store (this would also need the
  delete by query APIs etc.)

PR #25062, opened 2024-08-05
core[minor]: Introduce DocumentIndex abstraction

eyurtsev:
This PR adds a minimal document indexer abstraction.

The goal of this abstraction is to allow developers to create custom retrievers that also have a standard indexing API and allow updating the document content in them.

The abstraction comes with a test suite that can verify that the indexer implements the correct semantics.

This is an iteration over a previous PRs (https://github.com/langchain-ai/langchain/pull/24364). The main difference is that we're sub-classing from BaseRetriever in this iteration and as so have consolidated the sync and async interfaces.

The main problem with the current design is that runt time search configuration has to be specified at init rather than provided at run time.

We will likely resolve this issue in one of the two ways:

(1) Define a method (`get_retriever`) that will allow creating a retriever at run time with a specific configuration.. If we do this, we will likely break the subclass on BaseRetriever
(2) Generalize base retriever so it can support structured queries

---- 

Related PRs that have some variations

* https://github.com/langchain-ai/langchain/pull/24565
* https://github.com/langchain-ai/langchain/pull/24364

### DIFF
--- a/libs/core/langchain_core/indexing/base.py
+++ b/libs/core/langchain_core/indexing/base.py
@@ -491,7 +491,7 @@ class DeleteResponse(TypedDict, total=False):
     """The number of items that failed to be deleted."""
 
 
-@beta(message="Added in 0.2.9. The abstraction is subject to change.")
+@beta(message="Added in 0.2.29. The abstraction is subject to change.")
 class DocumentIndex(BaseRetriever):
     """A document retriever that supports indexing operations.
 
@@ -507,7 +507,7 @@ class DocumentIndex(BaseRetriever):
     2. Fetching document by ID.
     3. Searching for document using a query.
 
-    .. versionadded:: 0.2.9
+    .. versionadded:: 0.2.29
     """
 
     @abc.abstractmethod
@@ -531,7 +531,7 @@ class DocumentIndex(BaseRetriever):
             successfully added or updated in the vectorstore and the list of IDs that
             failed to be added or updated.
 
-        .. versionadded:: 0.2.9
+        .. versionadded:: 0.2.29
         """
 
     async def aupsert(
@@ -556,7 +556,7 @@ class DocumentIndex(BaseRetriever):
             successfully added or updated in the vectorstore and the list of IDs that
             failed to be added or updated.
 
-        .. versionadded:: 0.2.9
+        .. versionadded:: 0.2.29
         """
         return await run_in_executor(
             None,
@@ -631,7 +631,7 @@ class DocumentIndex(BaseRetriever):
         Returns:
             List[Document]: List of documents that were found.
 
-        .. versionadded:: 0.2.9
+        .. versionadded:: 0.2.29
         """
 
     async def aget(
@@ -659,7 +659,7 @@ class DocumentIndex(BaseRetriever):
         Returns:
             List[Document]: List of documents that were found.
 
-        .. versionadded:: 0.2.9
+        .. versionadded:: 0.2.29
         """
         return await run_in_executor(
             None,
--- a/libs/core/langchain_core/indexing/base.py
+++ b/libs/core/langchain_core/indexing/base.py
@@ -530,8 +530,6 @@ class DocumentIndex(BaseRetriever):
             UpsertResponse: A response object that contains the list of IDs that were
             successfully added or updated in the vectorstore and the list of IDs that
             failed to be added or updated.
-
-        .. versionadded:: 0.2.29
         """
 
     async def aupsert(
@@ -555,8 +553,6 @@ class DocumentIndex(BaseRetriever):
             UpsertResponse: A response object that contains the list of IDs that were
             successfully added or updated in the vectorstore and the list of IDs that
             failed to be added or updated.
-
-        .. versionadded:: 0.2.29
         """
         return await run_in_executor(
             None,
@@ -630,8 +626,6 @@ class DocumentIndex(BaseRetriever):
 
         Returns:
             List[Document]: List of documents that were found.
-
-        .. versionadded:: 0.2.29
         """
 
     async def aget(
@@ -658,8 +652,6 @@ class DocumentIndex(BaseRetriever):
 
         Returns:
             List[Document]: List of documents that were found.
-
-        .. versionadded:: 0.2.29
         """
         return await run_in_executor(
             None,
--- a/libs/core/langchain_core/indexing/base.py
+++ b/libs/core/langchain_core/indexing/base.py
@@ -1,8 +1,14 @@
 from __future__ import annotations
 
+import abc
 import time
 from abc import ABC, abstractmethod
-from typing import Dict, List, Optional, Sequence, TypedDict
+from typing import Any, Dict, List, Optional, Sequence, TypedDict
+
+from langchain_core._api import beta
+from langchain_core.documents import Document
+from langchain_core.retrievers import BaseRetriever
+from langchain_core.runnables import run_in_executor
 
 
 class RecordManager(ABC):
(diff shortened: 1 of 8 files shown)


libs/core/langchain_core/indexing/base.py
```diff
@@ -447,3 +453,217 @@ class UpsertResponse(TypedDict):
...
+
+    1. Storing document in the index.
+    2. Fetching document by ID.
+    3. Searching for document using a query.
+
+    .. versionadded:: 0.2.29
+    """
+
+    @abc.abstractmethod
+    def upsert(self, items: Sequence[Document], /, **kwargs: Any) -> UpsertResponse:
```
efriis, line 514:
```suggestion
    def upsert(self, items: Sequence[Document], /, *, **kwargs: Any) -> UpsertResponse:
```
not super relevant for this one, but I think the breaking change we want to enforce is if folks add kwargs in future, they should be kwarg-only. `/` only enforces left side as positional

libs/core/langchain_core/indexing/base.py
```diff
@@ -447,3 +453,217 @@ class UpsertResponse(TypedDict):
...
+        Returns:
+            UpsertResponse: A response object that contains the list of IDs that were
+            successfully added or updated in the vectorstore and the list of IDs that
+            failed to be added or updated.
+
+        .. versionadded:: 0.2.29
+        """
+
+    async def aupsert(
+        self, items: Sequence[Document], /, **kwargs: Any
```
efriis, line 536:
```suggestion
        self, items: Sequence[Document], /, *, **kwargs: Any
```

libs/core/langchain_core/indexing/base.py
```diff
@@ -447,3 +453,217 @@ class UpsertResponse(TypedDict):
...
+        """
+        return await run_in_executor(
+            None,
+            self.upsert,
+            items,
+            **kwargs,
+        )
+
+    @abc.abstractmethod
+    def delete(self, ids: Optional[List[str]] = None, **kwargs: Any) -> DeleteResponse:
```
efriis, line 565:
```suggestion
    def delete(self, ids: Optional[List[str]], /, *, **kwargs: Any) -> DeleteResponse:
```
if `delete(None)` deletes everything, we should probably require passing in None or something rather than that being default behavior

libs/core/langchain_core/indexing/base.py
```diff
@@ -447,3 +453,217 @@ class UpsertResponse(TypedDict):
...
+                For example, can include an option to delete the entire index,
+                or else issue a non-blocking delete etc.
+
+        Returns:
+            DeleteResponse: A response object that contains the list of IDs that were
+            successfully deleted and the list of IDs that failed to be deleted.
+        """
+
+    async def adelete(
+        self, ids: Optional[List[str]] = None, **kwargs: Any
```
efriis, line 582:
```suggestion
        self, ids: Optional[List[str]], /, *, **kwargs: Any
```

libs/core/langchain_core/indexing/base.py
```diff
@@ -447,3 +453,217 @@ class UpsertResponse(TypedDict):
...
+            self.delete,
+            ids,
+            **kwargs,
+        )
+
+    @abc.abstractmethod
+    def get(
+        self,
+        ids: Sequence[str],
+        /,
```
efriis, line 608:
```suggestion
        /,
        *,
```

libs/core/langchain_core/indexing/base.py
```diff
@@ -447,3 +453,217 @@ class UpsertResponse(TypedDict):
...
+        Returns:
+            List[Document]: List of documents that were found.
+
+        .. versionadded:: 0.2.29
+        """
+
+    async def aget(
+        self,
+        ids: Sequence[str],
+        /,
```
efriis, line 634:
```suggestion
        /,
        *,
```

libs/core/langchain_core/indexing/in_memory.py
```diff
@@ -0,0 +1,97 @@
...
+
+from langchain_core._api import beta
+from langchain_core.callbacks import CallbackManagerForRetrieverRun
+from langchain_core.documents import Document
+from langchain_core.indexing import UpsertResponse
+from langchain_core.indexing.base import DeleteResponse, DocumentIndex
+from langchain_core.pydantic_v1 import Field
+
+
+@beta(message="Introduced in version 0.2.29")
```
efriis, line 12:
```suggestion
@beta(message="Introduced in version 0.2.29. Underlying abstraction subject to change.")
```

efriis: and make /* stuff match whatever you decide in underlying abstraction below

libs/core/langchain_core/indexing/in_memory.py
```diff
@@ -0,0 +1,97 @@
...
+            if id_ in self.store:
+                found_documents.append(self.store[id_])
+
+        return found_documents
+
+    async def aupsert(
+        self, items: Sequence[Document], /, **kwargs: Any
+    ) -> UpsertResponse:
+        """Upsert items into the index."""
+        return self.upsert(items, **kwargs)
```
efriis, line 75: shouldn't we just rely on underlying thing here instead? This will block eventloop, so should at least run in executor right?

libs/core/langchain_core/indexing/in_memory.py
```diff
@@ -0,0 +1,97 @@
...
+
+    async def adelete(
+        self, ids: Optional[List[str]] = None, **kwargs: Any
+    ) -> DeleteResponse:
+        """Delete by ID."""
+        return self.delete(ids, **kwargs)
+
+    async def aget(self, ids: Sequence[str], /, **kwargs: Any) -> List[Document]:
+        """Get by ids."""
+        return self.get(ids, **kwargs)
```
efriis, line 85: same for all of these

libs/core/langchain_core/indexing/in_memory.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _get_relevant_documents(
```
efriis, line 87: Shouldn't this be implemented on the core abstraction?

libs/core/tests/unit_tests/indexing/test_in_memory_indexer.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+async def test_async_retriever() -> None:
+    index = InMemoryDocumentIndex()
+    documents = [
+        Document(id="1", page_content="hello world"),
+        Document(id="2", page_content="goodbye cat"),
+    ]
+    await index.aupsert(documents)
+    assert (await index.ainvoke("hello")) == [documents[0], documents[1]]
```
efriis, line 49: if these are handled by standard tests anyways let's just delete these here

efriis (commented):
looks good! Small nits

libs/core/langchain_core/indexing/in_memory.py
eyurtsev: Toy implementation, so don't think anyone will use with much data in which case should execute without a run executor. But if folks will use as reference maybe we should delegate to underlying implementation

libs/core/langchain_core/indexing/in_memory.py
eyurtsev: I ended up removing the search method. I think the generic search method makes sense if we're doing a `create_retriever()` method, but it doesn't make that much sense if we're sub-classing from retriever.

libs/core/langchain_core/indexing/base.py
eyurtsev: I think this will result in a syntax error since it's followed by **kwargs

libs/core/langchain_core/indexing/base.py
eyurtsev: There's a unit test that guarantees that delete(None) raises an exception. I think the semantics should be such that `ids`. With those semantics the current function signature is fine

libs/core/tests/unit_tests/indexing/test_in_memory_indexer.py
eyurtsev: Not in standard tests since we can't standardize on the search methodology

libs/standard-tests/langchain_standard_tests/integration_tests/indexer.py
```diff
@@ -0,0 +1,392 @@
...
+
+        index.upsert(documents)
+        index.delete(["1", "2"])
+        assert index.get(["1", "2", "3"]) == [
+            Document(page_content="baz", metadata={"id": 3}, id="3")
+        ]
+
+    def test_delete_no_args(self, index: DocumentIndex) -> None:
+        """Test delete with no args raises ValueError."""
+        with pytest.raises(ValueError):
```
eyurtsev, line 170: cc @efriis -- parameterless delete will raise

efriis (approved):
discussed

PR #25069, opened 2024-08-05
core[patch]: Deprecating beta upsert APIs in vectorstore

eyurtsev:
This PR deprecates the beta upsert APIs in vectorstore.

We'll introduce them in a V2 abstraction instead to keep the existing vectorstore implementations lighter weight.

The main problem with the existing APIs is that it's a bit more challenging to
implement the correct behavior w/ respect to IDs since ID can be present in
both the function signature and as an optional attribute on the document
object.

But VectorStores that pass the standard tests should have implemented the semantics properly!


libs/core/langchain_core/vectorstores/base.py
```diff
@@ -458,37 +268,21 @@ def add_documents(self, documents: List[Document], **kwargs: Any) -> List[str]:
...
+                # If there's at least one valid ID, we'll assume that IDs
+                # should be used.
+                if any(ids):
+                    kwargs["ids"] = ids
+
+            texts = [doc.page_content for doc in documents]
+            metadatas = [doc.metadata for doc in documents]
+            return self.add_texts(texts, metadatas, **kwargs)
+        raise NotImplementedError(
+            f"`add_documents` has not been implemented for {self.__class__.__name__} "
```
efriis, line 284:
```suggestion
            f"`add_documents` and `add_texts` has not been implemented for {self.__class__.__name__} "
```

efriis (approved):
one error messaging suggestion

libs/community/langchain_community/vectorstores/aperturedb.py
```diff
@@ -460,7 +460,45 @@ def list_vectorstores(class_) -> None:
...
+                    id=id_,
+                )
+                documents_.append(doc_with_id)
+        else:
+            documents_ = documents
+
+        # If upsert has been implemented, we can use it to add documents
+        return self.upsert(documents_, **kwargs)["succeeded"]
+
     def upsert(self, items: Sequence[Document], /, **kwargs: Any) -> UpsertResponse:
```
baskaryan, line 502: should we deprecate?

eyurtsev: No, don't think so, it's not the base abstraction and there are other vectorstores that have random methods

PR #25070, opened 2024-08-05
core: base language model disable_streaming

efriis:


libs/core/langchain_core/language_models/base.py
```diff
@@ -107,6 +108,8 @@ class BaseLanguageModel(
         default=None, exclude=True
     )
     """Optional encoder to use for counting tokens."""
+    disable_streaming: Union[bool, Literal["tool_calling"]] = False
+    """Whether to disable streaming for this model."""
```
eyurtsev, line 112: Could you document what the literal tool_calling does

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -762,20 +776,29 @@ def _generate_with_cache(
...
-                        True
-                        for h in run_manager.handlers
-                        if isinstance(h, _StreamingCallbackHandler)
-                    ),
-                    False,
-                )
-                if run_manager
-                else False
-            ),
+        if (
```
baskaryan, line 779: could we factor this out into a method (with an async flag)

libs/community/langchain_community/llms/llamafile.py
efriis, line 1: switching this to a whitelist because old logic breaks for any parameter add.

libs/core/langchain_core/language_models/chat_models.py
efriis: ya

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -304,6 +316,39 @@ async def ainvoke(
...
+        ] = None,
+        **kwargs: Any,
+    ) -> bool:
+        sync_not_implemented = type(self)._stream == BaseChatModel._stream
+        async_not_implemented = type(self)._astream == BaseChatModel._astream
+
+        # Check if streaming is implemented.
+        if (not async_api) and sync_not_implemented:
+            return False
+        if async_api and async_not_implemented and sync_not_implemented:
```
eyurtsev, line 334: Could you add a comment here to make sure we don't lose the `sync_not_implemented`?

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -304,6 +316,39 @@ async def ainvoke(
...
 
+    def _should_stream(
+        self,
+        *,
+        async_api: bool,
+        run_manager: Optional[
+            Union[CallbackManagerForLLMRun, AsyncCallbackManagerForLLMRun]
+        ] = None,
+        **kwargs: Any,
+    ) -> bool:
```
eyurtsev, line 327: Would a bit of documentation here be helpful to explain WHY one might need to disable streaming?

eyurtsev (approved):

PR #25073, opened 2024-08-05
Set Context in RunnableSequence & RunnableParallel

hinthornw:


libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -172,9 +174,12 @@ def invoke(
             try:
                 if self.exception_key and last_error is not None:
                     input[self.exception_key] = last_error
-                output = runnable.invoke(
+                child_config = patch_config(config, callbacks=run_manager.get_child())
```
eyurtsev, line 177: Do we need similar changes in RunnableParallel? 

Is there any utility code that can be pulled out?

hinthornw: Likely ya

hinthornw: Maybe `patch_and_set_context()` or something

hinthornw: Decided against for now

eyurtsev (approved):

PR #25078, opened 2024-08-05
ollama: allow base_url, headers, and auth to be passed

isahers1:


efriis (commented):
creating clients on each request seems a bit heavy - should we instantiate on init instead, or just pass through to e.g. `ollama.x` functions?

efriis (approved):

Issue #25072, opened 2024-08-05
extra="allow" not working after langchain-core==0.2.27


wgrcunha:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code:
```python
from typing import Any, Type

from langchain.pydantic_v1 import BaseModel, Field
from langchain_core.tools import BaseTool
from langchain_core.utils.function_calling import convert_to_openai_tool

name = "testing"
description = "testing"


def run(some_param, **kwargs):
    pass


class ToolSchema(BaseModel, extra="allow"):
    some_param: str = Field(default="", description="some_param")


class RunTool(BaseTool):
    name = name
    description = description
    args_schema: Type[BaseModel] = ToolSchema

    def _run(
        self,
        some_param: str = "",
    ) -> Any:
        return run(
            some_param=some_param,
            **self.metadata,
        )


convert_to_openai_tool(RunTool())

```

### Error Message and Stack Trace (if applicable)

```
Traceback (most recent call last):
  File "/home/freire/dev/navi-ai-api/app/playground.py", line 34, in <module>
    convert_to_openai_tool(RunTool())
  File "/home/freire/dev/navi-ai-api/venv/lib/python3.10/site-packages/langchain_core/utils/function_calling.py", line 392, in convert_to_openai_tool
    function = convert_to_openai_function(tool)
  File "/home/freire/dev/navi-ai-api/venv/lib/python3.10/site-packages/langchain_core/utils/function_calling.py", line 363, in convert_to_openai_function
    return cast(Dict, format_tool_to_openai_function(function))
  File "/home/freire/dev/navi-ai-api/venv/lib/python3.10/site-packages/langchain_core/_api/deprecation.py", line 168, in warning_emitting_wrapper
    return wrapped(*args, **kwargs)
  File "/home/freire/dev/navi-ai-api/venv/lib/python3.10/site-packages/langchain_core/utils/function_calling.py", line 282, in format_tool_to_openai_function
    if tool.tool_call_schema:
  File "/home/freire/dev/navi-ai-api/venv/lib/python3.10/site-packages/langchain_core/tools.py", line 398, in tool_call_schema
    return _create_subset_model(
[...]

PR #25080, opened 2024-08-05
core[patch]: Fix annotation parsing in schema

eyurtsev:


eyurtsev:
https://github.com/langchain-ai/langchain/issues/25072

baskaryan:
should be fixed and tested in #25084

PR #25083, opened 2024-08-05
community: add GitHub LLM integration

Josephrp:
community: add GitHub LLM integration

**Description:** This PR adds a new LLM integration for GitHub's inference endpoint. The `GithubLLM` class allows users to interact with various models hosted on GitHub's platform, including models from AI21, Cohere, Meta, Mistral, OpenAI, and Microsoft. The integration supports single-turn conversations, multi-turn conversations, and streaming responses.

**Issue:** https://github.com/langchain-ai/langchain/discussions/25082

**Dependencies:** This integration requires the `requests` library, which is already a dependency of the `langchain-community` package.

**Twitter handle:** @Josephpollack 

Key features of the `GithubLLM` integration:
1. Supports a wide range of models (22 models listed)
2. Allows setting custom system prompts
3. Supports multi-turn conversations
4. Provides streaming capability
5. Includes model validation to ensure only supported models are used

- [x] I've added unit tests for the `GithubLLM` class and created an example notebook demonstrating its usage. 
- [x] The integration follows LangChain's best practices for optional dependency imports and backwards compatibility.

**OUTSTANDING TASKS:**

- [x] implement async 
- [x] handle empty system prompt 
- [x] map max tokens per model


libs/community/langchain_community/llms/github.py
```diff
@@ -0,0 +1,286 @@
...
+from typing import Any, Dict, List, Mapping, Optional, Generator
+
+import requests
+from langchain_core.callbacks import CallbackManagerForLLMRun
+from langchain_core.language_models.llms import LLM
+from langchain_core.pydantic_v1 import Extra, Field, root_validator
+
+logger = logging.getLogger(__name__)
+
+class GithubLLM(LLM):
```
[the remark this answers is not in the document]
Josephrp: thanks for the quick review , i will better inform myself and come back the improvements suggested 👍🏻

libs/community/langchain_community/chat_models/githubllm.py
```diff
@@ -0,0 +1,314 @@
...
+    else:
+        raise ValueError(f"Got unknown role {role}")
+
+class GithubLLM(BaseChatModel):
+    """GitHub LLM with Azure Fallback"""
+
+    github_endpoint_url: str = "https://models.inference.ai.azure.com/chat/completions"
+    model: str
+    github_api_key: SecretStr = Field(default_factory=lambda: SecretStr(os.environ.get("GITHUB_TOKEN", "")))
+    azure_api_key: SecretStr = Field(default_factory=lambda: SecretStr(os.environ.get("AZURE_API_KEY", "")))
```
ccurme, line 49: does this just change the auth header? would anyone ever want to instantiate with both of these keys? should we just accept `api_key`, let the user decide what key to provide, and set based on env vars in `validate_environment`?

libs/community/langchain_community/chat_models/githubllm.py
```diff
@@ -0,0 +1,314 @@
...
+    if role == "user":
+        return HumanMessage(content=content)
+    elif role == "assistant":
+        return AIMessage(content=content)
+    elif role == "system":
+        return SystemMessage(content=content)
+    else:
+        raise ValueError(f"Got unknown role {role}")
+
+class GithubLLM(BaseChatModel):
```
ccurme, line 43: We should rename this to `Chat____` to match convention for other models (e.g., `ChatGithub`).

I'm wondering if `Github` is the right name for this integration, though, given that the endpoint is https://models.inference.ai.azure.com/. My first thought would be `azure_ai_inference` / `ChatAzureAIInference` or similar, given the naming in the docs [here](https://learn.microsoft.com/en-us/azure/ai-studio/reference/reference-model-inference-api). Let me know if I'm looking at the wrong product.

libs/community/langchain_community/chat_models/githubllm.py
Josephrp: well, my thinking is that using the github PAT is completely free for a while , and if they are both instantiated then there is a "graceful fallback" to the azure account

libs/community/langchain_community/chat_models/githubllm.py
Josephrp: so i'm calling it Github because the product page is here : https://github.com/marketplace/models and this implementation looks for a github PAT first then falls back on azure ...

that said, i'm totally agnostic about this , we could even entirely drop github PAT support to favor azure , although using github PAT makes using the object completely free , which is quite nice :-)

Josephrp:
>### Issue 

> GithubLLM is similar or analogous to [Azure Endpoint Chat Model](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/chat_models/azureml_endpoint.py) and might benefit from a merge or as an improvement instead 

https://github.com/langchain-ai/langchain/discussions/25082#discussioncomment-10273955

efriis:
draft has been open a while - closing and feel free to reopen as a full pr for a review!

PR #25108, opened 2024-08-06
core[patch]: check for model_fields attribute

ccurme:
`__fields__` raises a warning in pydantic v2


baskaryan (commented):
nice! should we add a test? will attempting to get the attr on a v1 model emit a warning?

ccurme:
> nice! should we add a test? will attempting to get the attr on a v1 model emit a warning?

added `@pytest.mark.filterwarnings("error")` to that particular test. The test fails with the original code and passes now. I think this is sufficient as the test runs with both v1 and v2 namespace.

baskaryan (approved):

PR #25111, opened 2024-08-06
openai[patch]: enable strict tool calling

baskaryan:
Introduced https://openai.com/index/introducing-structured-outputs-in-the-api/


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -970,11 +984,16 @@ def bind_tools(
                     - ``"any"`` or ``"required"`` or ``True``: force at least one tool to be called.
                     - dict of the form ``{"type": "function", "function": {"name": <<tool_name>>}}``: calls <<tool_name>> tool.
                     - ``False`` or ``None``: no effect, default OpenAI behavior.
+            strict: If True, model output is guaranteed to exactly match the JSON Schema
+                provided in the tool definition.
```
ccurme, line 988:
```suggestion
                provided in the tool definition. Enabled by default for `"gpt-"` models.
```
?

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -374,6 +374,8 @@ def convert_to_openai_function(
 
 def convert_to_openai_tool(
     tool: Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool],
+    *,
+    strict: Optional[bool] = None,
```
ccurme, line 378: Do we want to add this in core, or just glom it onto the dict in `ChatOpenAI`?

ccurme (approved):
LGTM assuming:

- we've checked that integration tests are passing `strict=True` for openai
- tests for unstructured / together (whatever inherits BaseChatOpenAI) pass. Azure overwrites bind_tools + with_structured_output so should be fine.

---

In their [post](https://openai.com/index/introducing-structured-outputs-in-the-api/), OpenAI calls out another way of generating structured outputs beyond tool calling via `response_format`:

> 2. A new option for the response_format parameter: developers can now supply a JSON Schema via json_schema, a new option for the response_format parameter. This is useful when the model is not calling a tool, but rather, responding to the user in a structured way. This feature works with our newest gpt-4o models: gpt-4o-2024-08-06, released today, and gpt-4o-mini-2024-07-18. When a response_format is supplied with strict: true, model outputs will match the supplied schema.

This seems more consistent with the intent of `.with_structured_output`. Maybe worth exploring for the models that support it.

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -382,15 +384,20 @@ def convert_to_openai_tool(
...
     """
     if isinstance(tool, dict) and tool.get("type") == "function" and "function" in tool:
         return tool
-    function = convert_to_openai_function(tool)
-    return {"type": "function", "function": function}
+    oai_function = convert_to_openai_function(tool)
+    oai_tool: Dict[str, Any] = {"type": "function", "function": oai_function}
+    if strict is not None:
+        oai_tool["strict"] = strict
+    return oai_tool
```
bracesproul, line 400: At least in the JS SDK the `strict` field is inside the `function` field of a tool. Would double check in py

baskaryan: great catch

baskaryan: hm their api and py docs seem to suggest different things — maybe both are supported?
Screenshot 2024-08-06 at 12 25 16 PM
Screenshot 2024-08-06 at 12 25 25 PM

bracesproul: weird, I'm going to pass it via the `function` field in JS just to make types happy, and if they remove it'll cause issues we'll notice down the line.

libs/core/langchain_core/utils/function_calling.py
baskaryan: imo core, suspect other apis will follow

baskaryan: also this function is defined as returning oai format so seems fair game

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -371,9 +377,18 @@ def convert_to_openai_function(
...
+        # As of 08/06/24, OpenAI requires that additionalProperties be supplied and set
+        # to False if strict is True.
+        oai_function["parameters"]["additionalProperties"] = False
+    return oai_function
+
 
 def convert_to_openai_tool(
     tool: Union[Dict[str, Any], Type[BaseModel], Callable, BaseTool],
+    *,
+    strict: Optional[bool] = None,
```
eyurtsev, line 391: What's the difference between None and False?

eyurtsev: ah i see below -- maybe we can expand doc-string here as well since some users use this function directly?

efriis (approved):
works for me! If we wanted to be more prescriptive / assuming this has better performance across the board, we could opt people into strict, and only opt out for the list of models that don't support. But that could also be tricky

ccurme (approved):

PR #25118, opened 2024-08-06
docs: `integrations` reference updates 4

leo-gan:
Added missed references; missed provider pages.

### DIFF
--- a/docs/docs/integrations/providers/asknews.mdx
+++ b/docs/docs/integrations/providers/asknews.mdx
@@ -1,18 +1,8 @@
 # AskNews
 
->[AskNews](https://asknews.app/) infuses any LLM with the latest global news 
-> (or historical news), using a single natural language query. Specifically, 
-> `AskNews` is enriching over 300k articles per day by translating, summarizing, 
-> extracting entities, and indexing them into hot and cold vector databases. 
-> `AskNews` puts these vector databases on a low-latency endpoint for you. 
-> When you query `AskNews`, you get back a prompt-optimized string that contains all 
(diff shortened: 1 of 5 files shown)


leo-gan:
@eyurtsev Could you, please, check the Vercel error message? I don't have an access for it.

leo-gan:
@ccurme Can you review, please?

docs/docs/integrations/providers/asknews.mdx
```diff
@@ -0,0 +1,43 @@
+# AskNews
+
+>[AskNews](https://asknews.app/) infuses any LLM with the latest global news
```
ccurme, line 3: I don't think we should be copy/pasting material from other websites into the LangChain docs, especially as the attribution here is a little unclear-- we link to https://asknews.app but this appears lifted from https://docs.asknews.app/en. I don't think formatting with a comment block is sufficient. IMO this should be a high-level summary and a link to their page for more information.

leo-gan: Yes, I agree. Ideally, the company itself should create a page in the `providers/`. Maybe, we can send a request to all providers, asking them to provide us the "short company description"?
Now, it is mostly copy-past from the provider materials inside a comment block. It is better than no provider info at all. 

Sometimes, it takes a lot of time to find the company information. 
Try to find information about the provider of the EmbaasEmbeddings class, for example.

ccurme (approved):

ccurme:
Thank you for your help with this!

Issue #19933, opened 2024-04-02
HuggingFacePipeline does not use chat template


bibhas2:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Hugging Face pipeline now has [support for chat templates](https://huggingface.co/docs/transformers/main/en/chat_templating#is-there-an-automated-pipeline-for-chat). This calls the ``apply_chat_template()`` of the tokenizer. This is a super useful feature which formats the input correctly according to the model. To apply the template one needs to pass a messages list to the pipeline as input (and not a prompt text).

Langchain's ``HuggingFacePipeline`` class is written in a way that only prompt text is passed to the pipeline. We can see this in the ``‎HuggingFacePipeline._generate`` method. As a result the prompt is constructed using Langchain's default template which is not the same as what the model works best with.

Let's build an example.

```python
import torch
from langchain_core.prompts import ChatPromptTemplate
from langchain_core.output_parsers import StrOutputParser
from langchain_community.llms.huggingface_pipeline import HuggingFacePipeline

def test_agnostic_prompt(llm):
    prompt = ChatPromptTemplate.from_messages(
        [
            ("human", "When was Abraham Lincoln born?"),
            ("ai", "Abraham Lincoln was born on February 12, 1809."),
            ("human", "How old was he when he died?"),
            ("ai", "Abraham Lincoln died on April 15, 1865, at the age of 56."),
            ("human", "{question}"),
        ]
    )

    output_parser = StrOutputParser()

    chain = prompt | llm | output_parser

    reply = chain.invoke({"question": "Where did he die?"})

    print(reply)

hf_llm = HuggingFacePipeline.from_model_id(
        model_id="TinyLlama/TinyLlama-1.1B-Chat-v1.0",
        task="text-generation",
        pipeline_kwargs={"max_new_tokens": 128})

test_agnostic_prompt(hf_llm)
```

This sends the following prompt.

```
Human: When was Abraham Lincoln born?
AI: Abraham Lincoln was born on February 12, 1809.
Human: How old was he when he died?
AI: Abraham Lincoln died on April 15, 1865, at the age of 56.
Human: Where did he die?
```

[...]

Issue #24437, opened 2024-07-19
complete prompt is appended at the start of my response generated by llama3


ibtsamraza:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

    prompt = PromptTemplate(
    template="""<|begin_of_text|><|start_header_id|>system<|end_header_id|> You are an assistant for question-answering tasks.
    Use the following pieces of retrieved context to answer the question and give response from the context given to you as truthfully as you can.
    Do not add anything  from you and If you don't know the answer, just say that you don't know.
    <|eot_id|>
    <|start_header_id|>user<|end_header_id|>
    Question: {question}
    Context: {context}
    Chat History: {chat_history}
    Answer: <|eot_id|><|start_header_id|>assistant<|end_header_id|>""",
    input_variables=["question", "context", "chat_history"],
)

    global memory
    memory = ConversationBufferWindowMemory(k=4,
        memory_key='chat_history', return_messages=True, output_key='answer')

    # LLMs Using API
    
    llm = HuggingFaceHub(repo_id='meta-llama/Meta-Llama-3-8B-Instruct', huggingfacehub_api_token=api_key", model_kwargs={
                      "temperature": 0.1,"max_length": 300, "max_new_tokens": 300})
  

    compressor = CohereRerank()
    compression_retriever = ContextualCompressionRetriever(
        base_compressor=compressor, base_retriever=retriever3
    )
    

    global chain_with_memory

    # Create the custom chain
    chain_with_memory = ConversationalRetrievalChain.from_llm(
        llm=llm,
        memory=memory,
        retriever=compression_retriever,
        combine_docs_chain_kwargs={"prompt": prompt},
        return_source_documents=True,
    )
    

### Error Message and Stack Trace (if applicable)

[...]

PR #25136, opened 2024-08-07
Fix for issue #19933: langchain_huggingface

Soumil32:
This partially fixes issue [#19933](https://github.com/langchain-ai/langchain/issues/19933) in which Langchain was not taking the taking steps to apply the correct formatting for different chat models.

The change involves passing the prompt in the ChatML format of the pipeline. This way the pipeline can automatically format accordingly.  

In addition, the argument `return_full_text=False` is being passed into the pipeline in `huggingace_pipeline.py`. This causes the pipeline to only return the newly generated text whereas the default behaviour is to return the entire chat history which we do not need. The output can then be parsed back into an `AIMessage`.

This still only works when `ChatHuggingFace` is used. If either `model_id` or `tokenizer` are not set when initialising `ChatHuggingFace` , it defaults to `gpt2` tokenizer which is what caused the issue in the first place. The changes I have made correctly handle parsing the model response back into `AIMessage`. The initial issue still might need to be investigated but doing this should work as the tokenizer should be picked up from the pipeline passed into `HuggingFacePipeline`.
```python
from langchain_huggingface import ChatHuggingFace, HuggingFacePipeline
from langchain_core.messages import HumanMessage, SystemMessage
from transformers import pipeline, AutoTokenizer, AutoModelForCausalLM

model_id = "hugging-quants/Meta-Llama-3.1-8B-Instruct-BNB-NF4" #replace with your model
tokenizer = AutoTokenizer.from_pretrained(model_id)
terminators = [
    tokenizer.eos_token_id,
    tokenizer.convert_tokens_to_ids("<|eot_id|>")
]
model = AutoModelForCausalLM.from_pretrained(
  model_id,
  torch_dtype=torch.bfloat16,
  low_cpu_mem_usage=True,
  device_map="auto",
)

pipe = pipeline(
    "text-generation", model=model, tokenizer=tokenizer, max_new_tokens=40, eos_token_id=terminators, pad_token_id=tokenizer.eos_token_id
)

messages = [
    SystemMessage(content="You are a translator which converts the users input from English to French. Never refer to yourself. Just provide the translated text."),
    HumanMessage(content="Hi, how are you?")
]


hf = HuggingFacePipeline(pipeline=pipe)
chat_model = ChatHuggingFace(llm=hf, model_id=model_id)

response = chat_model.invoke(messages)

print(response)
```
Sorry if my language isn't clear. It is my first time contributing to any OSS project. If you have any questions or clarification is need, please say so! 😊

Edit: Did some more testing and my changes also fix the iisue the issue all together instead of partially! `model_id` or `tokenizer` do not need to be set anymore.


Soumil32:
@efriis whats the status on this?

Soumil32:
This also seems to fix [issue #24437](https://github.com/langchain-ai/langchain/issues/24437l).

Soumil32:
@ccurme Is there anything i still need to do?

libs/partners/huggingface/langchain_huggingface/llms/huggingface_pipeline.py
```diff
@@ -269,6 +270,7 @@ def _generate(
             # Process batch of prompts
             responses = self.pipeline(
                 batch_prompts,
+                return_full_text=False,
```
ccurme, line 273: Can we do this in a separate issue / PR?

ccurme (commented):
Could we add some tests to demonstrate the updated behavior?

DanielViglione:
@Soumil32 
The issue still persists in the HuggingFaceEndpoint class which is in the same directory as HuggingFacePipeline. The two classes should produce similar output, so by fixing one, the other one should be fixed as well.

vbarda:
@Soumil32 would you be able to add a test for this change? you might need to also resolve some merge conflicts

vbarda:
Closing due to inactivity

PR #25157, opened 2024-08-08
docs: Fix reference to SQL QA migration

JuanFKurucz:
**Description:** I found that the link to the notebook in the Migration notes is broken, i found that it was linked to this file https://github.com/langchain-ai/langchain/blob/v0.0.250/docs/extras/use_cases/tabular/sql_query.ipynb and i think now this tutorial https://github.com/JuanFKurucz/langchain/blob/master/docs/docs/tutorials/sql_qa.ipynb is the best fit for this reference

**Twitter handle:** @juanfkurucz

### DIFF
--- a/MIGRATE.md
+++ b/MIGRATE.md
@@ -52,7 +52,7 @@ Now:
 
 `from langchain_experimental.sql import SQLDatabaseChain`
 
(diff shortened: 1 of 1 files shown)


ccurme (approved):
Thanks for the catch! I updated the link to point to the official docs.

PR #25189, opened 2024-08-08
core[minor]: Add from_env utility

eyurtsev:
Add a utility that can be used as a default factory

The goal will be to start migrating from of the pydantic models to use `from_env` as a default factory if possible.

```python

from pydantic import Field, BaseModel
from langchain_core.utils import from_env

class Foo(BaseModel):
   name: str = Field(default_factory=from_env('HELLO'))
```


eyurtsev:
Force merged -- all unit tests and lints passed but CI was hanging

PR #25198, opened 2024-08-08
core[minor]: Add factory for looking up secrets from the env

eyurtsev:
Add factory method for looking secrets from the env.


eyurtsev:
cc @ccurme , @baskaryan , @efriis another factory method for secrets

PR #25204, opened 2024-08-08
core[patch]: fix deprecation pydantic bug

baskaryan:
#25004 is incompatible with pydantic < 1.10.17. Introduces fix for this.

### DIFF
--- a/libs/core/langchain_core/_api/deprecation.py
+++ b/libs/core/langchain_core/_api/deprecation.py
@@ -138,8 +138,9 @@ def deprecated(
             from pydantic.v1.fields import FieldInfo  # pydantic: ignore
         except ImportError:
             # pydantic < 1.10.17
-            from pydantic.fields import FieldInfo  # pydantic: ignore
-
+            from pydantic.fields import (  # type: ignore[assignment]  # pydantic: ignore
+                FieldInfo,
+            )
 
         def emit_warning() -> None:
             """Emit the warning."""
--- a/libs/core/langchain_core/_api/deprecation.py
+++ b/libs/core/langchain_core/_api/deprecation.py
@@ -133,7 +133,7 @@ def deprecated(
         _package: str = package,
     ) -> T:
         """Implementation of the decorator returned by `deprecated`."""
-        from langchain_core.utils.pydantic import FieldInfoV1_
+        from langchain_core.utils.pydantic import FieldInfoV1
 
         def emit_warning() -> None:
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/_api/deprecation.py
```diff
@@ -133,7 +133,13 @@ def deprecate(
         _package: str = package,
     ) -> T:
         """Implementation of the decorator returned by `deprecated`."""
-        from pydantic.v1.fields import FieldInfo  # pydantic: ignore
+        try:
+            # pydantic >= 1.10.17
```
eyurtsev, line 137: `pydantic 2 `

libs/core/langchain_core/_api/deprecation.py
```diff
@@ -133,7 +133,13 @@ def deprecate(
...
+            from pydantic.v1.fields import FieldInfo  # pydantic: ignore
```
eyurtsev, line 138: Could we move to compat layer in langchain_core.utils.pydantic or make sure that we can accept FieldInfo that comes from either namespace and either version of pydantic? Otherwise this is just another thing that could break during 0.3

eyurtsev (approved):

libs/core/langchain_core/_api/deprecation.py
```diff
@@ -208,7 +208,7 @@ def warn_if_direct_instance(
                 )
                 return cast(T, obj)
 
-        elif isinstance(obj, FieldInfo):
+        elif isinstance(obj, FieldInfoV1):
             from langchain_core.pydantic_v1 import Field
```
eyurtsev, line 212: this will cause issues -- we're re-writing all langchain_core.pydantic_v1 import Field into from pydantic import Field

eyurtsev, line 212:
```suggestion
```

eyurtsev: and below we can use FieldInfoV1?

and make a version for FieldInfoV2 so it works after 0.3 migration?

hinthornw:
Getting this running locally:
```
ImportError: cannot import name 'deprecated' from partially initialized module 'langchain_core._api.deprecation' (most likely due to a circular import) (/Users/wfh/code/lc/langchain/libs/core/langchain_core/_api/deprecation.py)
```

hinthornw (approved):

PR #25217, opened 2024-08-09
docs: `integrations` references update 7

leo-gan:
Added missed provider pages. Fixed formats and added descriptions and links.

### DIFF
--- a/docs/docs/integrations/chat/octoai.ipynb
+++ b/docs/docs/integrations/chat/octoai.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# ChatOctoAI\n",
+    "# Octo AI\n",
     "\n",
     "[OctoAI](https://docs.octoai.cloud/docs) offers easy access to efficient compute and enables users to integrate their choice of AI models into applications. The `OctoAI` compute service helps you run, tune, and scale AI applications easily.\n",
     "\n",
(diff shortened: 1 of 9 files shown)


docs/docs/integrations/chat/octoai.ipynb
```diff
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# ChatOctoAI\n",
+    "# Octo AI\n",
```
ccurme, line 7: admittedly this is inconsistent in the docs, but the [template](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/chat.ipynb) for chat models specifies a "Chat" prefix (see issue here: https://github.com/langchain-ai/langchain/issues/22296)

ccurme (approved):

docs/docs/integrations/chat/octoai.ipynb
leo-gan: Change the template, please. The integration example page is a document for **humans**. The page title should not be the names of classes, especially because the examples are not "example per class" pages. Integration examples are about the **integrated System/Service/Package** not about the integration implementation (hence not about classes/functions). We have API Ref that is exactly about the implementation. 
The `docs/integrations/` pages for the first-time readers who chose the best integration for the use case or learn how this integration works [in the most base use case]. 
The API Reference is for developers who work with real projects and need all internal details.

PR #25226, opened 2024-08-09
community[patch]: change default Neo4j username/password

danodonovan:
**Description:**

Change the default Neo4j username/password (when not supplied as environment variable or in code) from `None` to `""`. 

Neo4j has an option to [disable auth](https://neo4j.com/docs/operations-manual/current/configuration/configuration-settings/#config_dbms.security.auth_enabled) which is helpful when developing. When auth is disabled, the username / password through the `neo4j` module should be `""` (ie an empty string). 

Empty strings get marked as false in `langchain_core.utils.env.get_from_dict_or_env` -- changing this code / behaviour would have a wide impact and is undesirable. 

In order to both _allow_ access to Neo4j with auth disabled and _not_ impact `langchain_core` this patch is presented. The downside would be that if a user forgets to set NEO4J_USERNAME or NEO4J_PASSWORD they would see an invalid credentials error rather than missing credentials error. This could be mitigated but would result in a less elegant patch! 

**Issue:**
Fix issue where langchain cannot communicate with Neo4j if Neo4j auth is disabled.

**Twitter handle:** @danodonovan


ccurme (approved):

jexp:
Hmm when auth is disabled, wouldn't it rather leave off the auth parameter in the driver?

Missing credentials is actually more helpful than "wrong credentials" here ... I can imagine searching for the bug e.g. a wrong path or a misspelled env variable for ages when it says wrong credentials instead of missing.

eyurtsev (dismissed):
"" credentials can be specified manually by the user, handled here with an if-else

danodonovan:
@jexp @eyurtsev thank you for the feedback, I agree. 6a60dad follows jexp suggestion to remove auth from the driver if `""` is intentionally provided

danodonovan:
@eyurtsev @jexp Sorry for pushing this PR past you again, the changes would be helpful for me if accepted.

baskaryan:
cc @tomasonjo

tomasonjo:
LGTM

V čet., 29. avg. 2024, 11:27 je oseba Bagatur ***@***.***>
napisala:

hwchase17:
needs to be formatted, otherwise lgtm

danodonovan:
> needs to be formatted, otherwise lgtm

Thanks @hwchase17 -- sorry to have missed the linting issue, it should be green-lit now

PR #25231, opened 2024-08-09
community: Add llm-extraction option to FireCrawl Document Loader

shivendrasoni:
**Description:** This minor PR aims to add `llm_extraction` to Firecrawl loader.  This feature is supported on API and PythonSDK, but the langchain loader omits adding this to the response.
**Twitter handle:** [scalable_pizza](https://x.com/scalablepizza)

### DIFF
--- a/libs/community/langchain_community/document_loaders/firecrawl.py
+++ b/libs/community/langchain_community/document_loaders/firecrawl.py
@@ -66,5 +66,5 @@ class FireCrawlLoader(BaseLoader):
             yield Document(
                 page_content=doc.get("markdown", ""),
                 metadata=doc.get("metadata", {}),
-                llm_extraction: doc.get("llm_extraction") if self.params.get('extractorOptions', {}).get('mode') == 'llm-extraction' else None
+                llm_extraction= doc.get("llm_extraction") if self.params.get('extractorOptions', {}).get('mode') == 'llm-extraction' else None
             )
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this. `llm_extraction` is not an attribute of [Document](https://api.python.langchain.com/en/latest/documents/langchain_core.documents.base.Document.html). Was the intent here to store `doc.get("llm_extraction")` in the document metadata?

shivendrasoni:
> Thanks for this. `llm_extraction` is not an attribute of [Document](https://api.python.langchain.com/en/latest/documents/langchain_core.documents.base.Document.html). Was the intent here to store `doc.get("llm_extraction")` in the document metadata?

Hi, yes and I realized it just now. I am just thinking going by the API response schema `llm_extraction`, should be parallel to the page_content, but I realized that even thought Document allows kwargs, it isn't getting set for this arbitrary param.

Do you think saving it in meta is semantically correct?

ccurme:
Is `llm_extraction` serializable? Do you have an example? Document metadata isn't very opinionated and there aren't many constraints on it. If it's useful metadata about the document and is serializable then makes sense to me.

shivendrasoni:
> Is `llm_extraction` serializable? Do you have an example? Document metadata isn't very opinionated and there aren't many constraints on it. If it's useful metadata about the document and is serializable then makes sense to me.

Hi, yes, llm_extraction is serializable since it is always guaranteed to be a valid json object.

ccurme (approved):

PR #25240, opened 2024-08-09
core[patch]: add standard tracing params for retrievers

ccurme:


libs/core/langchain_core/retrievers.py
```diff
@@ -167,6 +182,19 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
             len(set(parameters.keys()) - {"self", "query", "run_manager"}) > 0
         )
 
+    def _get_ls_params(self, **kwargs: Any) -> LangSmithRetrieverParams:
+        """Get standard params for tracing."""
+
+        default_retriever_name = self.__class__.__name__
```
eyurtsev, line 188: use `self.get_name()` (or double check the correct name in base runnable)

eyurtsev (approved):
Looks good except for the Runnable name which is more nuanced

efriis (approved):

PR #25246, opened 2024-08-09
[docs]: add every tool to a table

isahers1:


docs/docs/integrations/tools/pandas.ipynb
baskaryan, line 1: i think this is a pretty popular one

isahers1: I just don't see how it's a tool, there is no tool at all just agents. I am fine with leaving it somewhere but to me it feels wrong to leave it in tools

Issue #25218, opened 2024-08-09
AttributeError: 'str' object has no attribute 'text' in OpenAI document_loaders/audio


michmzr:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Top

``` python
import logging
import os
import random
from concurrent.futures import ThreadPoolExecutor, as_completed
from enum import Enum
from typing import BinaryIO
from typing import cast, Literal, Union

from langchain_community.document_loaders.generic import GenericLoader
from langchain_community.document_loaders.parsers.audio import OpenAIWhisperParser
from pydub import AudioSegment

from cache import conditional_lru_cache
from youtube.loader import YoutubeAudioLoader
....

    loader = GenericLoader(YoutubeAudioLoader([url], save_dir, proxy_servers),
                           OpenAIWhisperParser(api_key=get_settings().openai_api_key,
                                               language=lang.value,
                                               response_format="srt",
                                               temperature=0
                                               ))

```

YoutubeAudioLoader is my customization of Langchain YoutubeAudioLoader which allows the use of a proxy for Youtube.

```python
import random
from typing import Iterable, List

from langchain_community.document_loaders.blob_loaders import FileSystemBlobLoader
from langchain_community.document_loaders.blob_loaders.schema import Blob, BlobLoader


class YoutubeAudioLoader(BlobLoader):
    """Load YouTube urls as audio file(s)."""

    def __init__(self, urls: List[str], save_dir: str, proxy_servers: List[str] = None):

        if not isinstance(urls, list):
            raise TypeError("urls must be a list")

        self.urls = urls
        self.save_dir = save_dir
        self.proxy_servers = proxy_servers

    def yield_blobs(self) -> Iterable[Blob]:
        """Yield audio blobs for each url."""

        try:
            import yt_dlp
        except ImportError:
            raise ImportError(
                "yt_dlp package not found, please install it with "
                "`pip install yt_dlp`"
            )

        # Use yt_dlp to download audio given a YouTube url
        ydl_opts = {
            "format": "m4a/bestaudio/best",
            "noplaylist": True,
            "outtmpl": self.save_dir + "/%(title)s.%(ext)s",
            "postprocessors": [
[...]

PR #25271, opened 2024-08-11
[Community]: Fix - Open AI Whisper `client.audio.transcriptions` returning Text Object which raises error

keenborder786:
- **Description:** The following [line](https://github.com/langchain-ai/langchain/blob/fd546196ef0fafa4a4cd7bb7ebb1771ef599f372/libs/community/langchain_community/document_loaders/parsers/audio.py#L117) in `OpenAIWhisperParser` returns a text object for some odd reason despite the official documentation saying it should return `Transcript` Instance which should have the text attribute. But for the example given in the issue and even when I tried running on my own, I was directly getting the text. The small PR accounts for that.
 - **Issue:** : #25218
 

I was able to replicate the error even without the GenericLoader as shown below and the issue was with `OpenAIWhisperParser`

```python
parser = OpenAIWhisperParser(api_key="sk-fxxxxxxxxx",
                                            response_format="srt",
                                            temperature=0)

list(parser.lazy_parse(Blob.from_path('path_to_file.m4a')))
```


keenborder786:
@ccurme

ccurme (approved):

Issue #25278, opened 2024-08-11
Syntax Error is getting generated in all queries  made using "SQLDatabaseChain.from_llm"


akshayg147:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

db = SQLDatabase.from_uri(f"mysql+pymysql://{db_user}:{db_password}@{db_host}/{db_name}",sample_rows_in_table_info=3)
db_chain = SQLDatabaseChain.from_llm(llm,db,verbose=True)
db_chain("quantity of nike shirts of white colour?")

### Error Message and Stack Trace (if applicable)

sqlalchemy.exc.ProgrammingError: (pymysql.err.ProgrammingError) (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '```sql\nSELECT \n    `stock_quantity`\nFROM \n    `t_shirts`\nWHERE \n    `brand` = 'N' at line 1")

### Description

I encountered a problem when using the langchain-google-genai package to generate and execute SQL queries using SQLDatabasechain package against a MySQL database. The generated SQL query contains Markdown-style code block delimiters , which are causing a syntax error when executed.
Output query generated:
\```sql
SELECT 
    `stock_quantity`
FROM 
    `t_shirts`
WHERE 
    `brand` = 'Nike'
    AND `color` = 'White'
LIMIT 
    5;
 \```
 
 #### **Expected Behavior:**
The generated SQL query should not include Markdown-style code block delimiters or any other non-SQL syntax elements. It should produce a clean SQL query that can be executed without errors.

Please advise on how to handle this or if there is a fix that can be applied to prevent this issue.

### System Info

langchain-experimental==0.0.64
google-generativeai==0.7.2
langchain==0.2.12

keenborder786:
I would recommend to use the LCEL based SQL chain rather the experimental SQL Chain:

```sql

from langchain.chains import create_sql_query_chain
from langchain_community.utilities import SQLDatabase

db = SQLDatabase.from_uri(f"mysql+pymysql://{db_user}:{db_password}@{db_host}/{db_name}",sample_rows_in_table_info=3)
llm = <Your LLM Model>
[...]

PR #25282, opened 2024-08-11
modified MYSQL prompt in libs/langchain/langchain/chains/sql_database…

akshayg147:
…/prompt.py

- [x] experimental: Ensure MYSQL queries are generated perfectly so can be used by the next sequence in SQLDatabaseChain to execute them directly on database"
  - infra: langchain/libs/langchain/langchain/chains/sql_database/prompt.py --> modified _mysql_prompt

    - **Description:** This PR addresses an issue where the SQL queries generated by the GoogleGenerativeAI model in the langchain-experimental package included unwanted syntax. This query when used through SQLDatabaseChain caused the error. The generated queries now start directly with SELECT and do not contain any additional formatting or text. The fix involved modifying the prompt in the prompt.py file to ensure clean SQL query generation.
    - **Issue:** #25278
    - **Dependencies:** No new dependencies are required for this change.
    - **Twitter handle:** @CodeFrek

- [x] **Add tests and docs**: This the image shows the that same query runs perfectly, with the modified prompt.
image

- [x] **Lint and test**: Not required.

### DIFF
--- a/libs/langchain/langchain/chains/sql_database/prompt.py
+++ b/libs/langchain/langchain/chains/sql_database/prompt.py
@@ -135,7 +135,7 @@ Use the following format:
 Question: Question here
 SQLQuery: SQL Query to run
 SQLResult: Result of the SQLQuery
-Answer: Final answer here
+Answer: Final answer here (make sure the queiry do not contain 'sql' at the start, the query should start with SELECT)
 
 """
 


ccurme:
Thanks for this. We avoid optimizing prompts because it will change behavior for existing users. My recommendation is to pass in a custom prompt.

akshayg147:
Hey @ccurme ,

I was working on a project with an LLM and encountered some issues with the SQL query generation. After trying various prompts, I found that the correct sequence, which provided accurate responses, involved optimizing the BaseTemplate prompt. This improvement was necessary, especially since the existing prompt was producing suboptimal results with Google PaLM, which in turn was breaking the SQLDatabaseChain method.

Given that this is still in the experimental phase, I believe the prompt needs an update. Even if this particular change isn’t the final solution, it's clear that the current prompt isn't performing well and should be revised. The current prompt is generating poor results, and it's affecting the overall functionality of the SQLDatabaseChain method.

Looking forward to your feedback on this!

Thanks.

Best regards,
Akshay

ccurme:
> Hey @ccurme ,
> 
> I was working on a project with an LLM and encountered some issues with the SQL query generation. After trying various prompts, I found that the correct sequence, which provided accurate responses, involved optimizing the BaseTemplate prompt. This improvement was necessary, especially since the existing prompt was producing suboptimal results with Google PaLM, which in turn was breaking the SQLDatabaseChain method.
> 
> Given that this is still in the experimental phase, I believe the prompt needs an update. Even if this particular change isn’t the final solution, it's clear that the current prompt isn't performing well and should be revised. The current prompt is generating poor results, and it's affecting the overall functionality of the SQLDatabaseChain method.
> 
> Looking forward to your feedback on this!
> 
> Thanks.
> 
> Best regards, Akshay

Hi @akshayg147,

The scenario I am concerned about is that someone is happily using the chain, and after this update, behavior changes unexpectedly-- as an aside, it's also not clear to me that all queries should start with `SELECT` (let me know if I'm mistaken there).

The default is not intended to support everyone's use-case, because it is likely not possible for one prompt to do that. Starting with the default prompt, making modifications and then passing that in is encouraged, and should satisfy your needs here.

akshayg147:
Hey @ccurme,

I completely understand and respect your decision to avoid breaking the code of developers who are already using it successfully. However, based on my extensive testing, I have strong concerns that users relying on Google PaLM and SQLDatabaseChain may not be getting accurate queries. This is something I feel confident about and wanted to bring to your attention.

You’re absolutely right that not every SQL query starts with SELECT, so I refined the prompt to:
```"(make sure the query does not contain 'sql' at the start, the query should start with valid SQL syntax)"```

This modification has consistently produced correct queries across all types of SQL statements.

I believe many users tend to use ```from langchain.chains import create_sql_query_chain``` (as suggested by Mohammad), which works exceptionally well.

Thank you for your time—I’ve learned a lot about open source and the LLM space through this experience. I'm excited to continue contributing to making Langchain even better!

Looking forward to it!

Cheers.

PR #25300, opened 2024-08-12
community: kwargs for CassandraGraphVectorStore

bjchambers:
- **Description:** pass kwargs from CassandraGraphVectorStore to underlying store


bjchambers:
@cbornet

cbornet (approved):

libs/community/langchain_community/graph_vectorstores/cassandra.py
```diff
@@ -32,6 +32,7 @@ def __init__(
         session: Optional[Session] = None,
         keyspace: Optional[str] = None,
         setup_mode: SetupMode = SetupMode.SYNC,
+        **kwargs: Any,
```
bjchambers, line 35: @cbornet with this change, we could drop most (all?) of the optional arguments (eg., node_table, session, keyspace, etc.) and just let the `kwargs` pass them through. That would let the underlying `GraphStore` drive more of the options. Thoughts?

cbornet: I think keeping the arguments explicit is better for documentation, code completion, ...?

ccurme (approved):

PR #25310, opened 2024-08-12
langchain[patch]: deprecate various chains

ccurme:
- [x] NatbotChain: move to community, deprecate langchain version. Update to use `prompt | llm | output_parser` instead of LLMChain.
- [x] LLMMathChain: deprecate + add langgraph replacement example to API ref
- [x] HypotheticalDocumentEmbedder (retriever): update to use `prompt | llm | output_parser` instead of LLMChain
- [x] FlareChain: update to use `prompt | llm | output_parser` instead of LLMChain
- [x] ConstitutionalChain: deprecate + add langgraph replacement example to API ref
- [x] LLMChainExtractor (document compressor): update to use `prompt | llm | output_parser` instead of LLMChain
- [x] LLMChainFilter (document compressor): update to use `prompt | llm | output_parser` instead of LLMChain
- [x] RePhraseQueryRetriever (retriever): update to use `prompt | llm | output_parser` instead of LLMChain


libs/langchain/langchain/chains/natbot/base.py
```diff
@@ -5,15 +5,27 @@
...
+@deprecated(
+    since="0.2.13",
+    message=(
+        "Importing NatBotChain from langchain is deprecated and will be removed in "
+        "langchain 1.0. Please import from langchain_community instead: "
+        "from langchain_community.chains.natbot import NatBotChain. "
+        "You may need to pip install -U langchain-community."
+    ),
+    removal="1.0",
+)
```
hwchase17, line 28: should we do the thing where we just re-import this?

hwchase17: i also dont think this needs to be a chain at all?

ccurme: I considered re-importing but because langchain no longer depends on langchain-community, this would be a breaking change. Can do if preferred.

IMO this should be a community integration, which is why I set up imports there. The [natbot](https://github.com/nat/natbot) repo has had no activity for 2 years and we originally specified this as "deprecate without replacement." So didn't think refactoring it is high priority.

PR #25330, opened 2024-08-13
Add kwargs for json.dumps inside `JsonOutputParser.get_format_instructions`

LucienShui:
I have a pydantic object:

```python

class Foo(BaseModel):
    bar: Optional[str] = Field(default=None, description="这是一个说明")
```

Then I want to prompt llm with it:

```python


baskaryan:
@LucienShui could you add a unit test to demonstrate and test the functionality?

hwchase17:
needs more info an description of why

PR #25332, opened 2024-08-13
qianfan generate/agenerate with usage_metadata

bimslab:


libs/community/langchain_community/chat_models/baidu_qianfan_endpoint.py
```diff
@@ -512,22 +512,30 @@ def _generate(
...
                 )
                 completion += chunk.text
-            lc_msg = AIMessage(content=completion, additional_kwargs={})
+                if isinstance(chunk.message, AIMessageChunk):
+                    usage_metadata = chunk.message.usage_metadata
+
+            lc_msg = AIMessage(
+                content=completion,
+                additional_kwargs={},
+                usage_metadata=usage_metadata,
```
ccurme, line 529: it looks like `usage_metadata` will be the last value from the loop. is it typically present in the last value? (e.g., so it won't get overwritten with None during the loop)

ccurme (approved):

PR #25338, opened 2024-08-13
openai: Add support for tools in with_structured_output method

tim-watcha:
## Description
This PR introduces support for state tools that the LLM can utilize. Providing tools to the LLM, even when not directly used, can influence and improve the LLM's output.

## Changes
- Implemented functionality to pass tool lists to the LLM
- Added support for the LLM to recognize available tools
- Ensured compatibility with 'function_calling' 'json_mode' and 'json_schema' methods

## Examples
1. Without tool provided:
스크린샷 2024-08-13 오후 7 42 20

2. With tool provided:
스크린샷 2024-08-13 오후 7 43 32

## Benefits
- Enhances LLM awareness of available tools
- Improves output quality and relevance
- Provides more flexibility in LLM interactions


efriis (commented):
I don't think this contribution makes sense on `with_structured_output` - why wouldn't you just bind tools directly with `llm.bind_tools()` and skip with_structured_output in this case?

hwchase17:
> I don't think this contribution makes sense on `with_structured_output` - why wouldn't you just bind tools directly with `llm.bind_tools()` and skip with_structured_output in this case?

i think the use case is he wants to returned structured output, but he wants the LLM to know about other tools (since that impacts the results). Currently this is only achievable with 
```

class Plan(BaseModel):
    steps: list[str]


class YahooSearch(BaseModel):
    """queries should be 5 words"""
    query: str

model = ChatOpenAI().with_structured_output(Plan).bind(tools=[convert_to_openai_tool(Plan), convert_to_openai_tool(YahooSearch)])
```
which isnt super nice

efriis:
ah makes sense

leaving as in review to check after 0.3 release. If we do this we should probably do it in core to enforce across different providers, not just openai

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1401,7 +1412,7 @@ class AnswerWithJustification(BaseModel):
                 )
             tool_name = convert_to_openai_tool(schema)["function"]["name"]
             llm = self.bind_tools(
-                [schema],
+                [schema, *tools] if tools else [schema],
```
efriis, line 1411: open question whether we want to support custom ordering here too because that can affect model output

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1416,7 +1427,15 @@ class AnswerWithJustification(BaseModel):
                     key_name=tool_name, first_tool_only=True
                 )
         elif method == "json_mode":
-            llm = self.bind(response_format={"type": "json_object"})
+            if tools:
```
efriis, line 1426: seems sensible to me to enforce strict when passing tools here

libs/partners/openai/langchain_openai/chat_models/base.py
efriis: allowing passing in schema included

libs/partners/openai/langchain_openai/chat_models/base.py
tim-watcha: 스크린샷 2024-10-09 오후 2 45 00
As you can see here, when use `json_mode`, tools must be strict by langchain rule

libs/partners/openai/langchain_openai/chat_models/base.py
tim-watcha: It might be needed for experimental reasons, but I'm wondering how we can make the parameters neater.

Option 1:
Check tool names to see whether the schema is included in the tools.

Option 2:
Add a schema_name option to use as the tool_name.

Option 3:
Add the feature when someone raises an issue.

tim-watcha:
Is there anything that I need to answer? @efriis @hwchase17

efriis:
nothing at the moment - sorry for the delay! anything we do here we'll want to do against core as well to make sure it's a standard interface instead of an openai-specific feature

haven't had time as of yet

efriis:
going to close this in favor of AIMessage.parsed work! It's still in progress and follow along in the linked PRs for updates instead of here

PR #25360, opened 2024-08-13
Merge support for group manager

hinthornw:


libs/core/langchain_core/callbacks/manager.py
```diff
@@ -1622,6 +1622,32 @@ def copy(self) -> CallbackManagerForChainGroup:
...
+            parent_run_id=self.parent_run_id or other.parent_run_id,
+            handlers=[],
+            inheritable_handlers=[],
+            tags=list(set(self.tags + other.tags)),
+            inheritable_tags=list(set(self.inheritable_tags + other.inheritable_tags)),
+            metadata={
+                **self.metadata,
+                **other.metadata,
+            },
+            parent_run_manager=self.parent_run_manager,
```
hinthornw, line 1638: Would be missing otherwise

libs/core/langchain_core/callbacks/base.py
```diff
@@ -911,15 +911,38 @@ def __init__(
...
-            inheritable_metadata=self.inheritable_metadata,
+            tags=self.tags.copy(),
+            inheritable_tags=self.inheritable_tags.copy(),
+            metadata=self.metadata.copy(),
+            inheritable_metadata=self.inheritable_metadata.copy(),
         )
 
+    def merge(self: T, other: BaseCallbackManager) -> T:
+        manager = self.__class__(
+            parent_run_id=self.parent_run_id or other.parent_run_id,
```
eyurtsev, line 925: Are we missing an assert here to verify that the run ids are the same?

libs/core/langchain_core/callbacks/base.py
```diff
@@ -911,15 +911,38 @@ def __init__(
     def copy(self: T) -> T:
         """Copy the callback manager."""
         return self.__class__(
-            handlers=self.handlers,
-            inheritable_handlers=self.inheritable_handlers,
+            handlers=self.handlers.copy(),
```
eyurtsev, line 914: why are copies required on self.handlers? Do we ever mutate the original list?

eyurtsev (approved):

PR #25368, opened 2024-08-14
Add Pubmed Central Integration

XinlingWang0628:
This Pull Request introduces an integration for Pubmed Central into the Langchain_community. The integration enables the retrieval of full-text articles from Pubmed Central using PMC_ID (Pubmed Central ID). Currently, the existing Pubmed integration in Langchain only provides paper summaries, making it limited in scope. 

By incorporating this feature, users will be able to access the complete content of Pubmed Central papers. Access to full-text articles supports more comprehensive analysis and better integration with various research workflows.

In this PR, I introduce the PmcIDSearchWrapper class, which provides a unified interface for interacting with the PubMed Central (PMC) API. It enables searching for full-text articles by PMC ID, downloading files from AWS S3, and extracting text from XML files.

I also added the PmcIDyRun tool, which integrates PmcIDSearchWrapper with LangChain, allowing for efficient PMC ID-based queries.

To ensure everything works as expected, I have included integration tests that verify the functionality of PmcIDSearchWrapper and its integration with PmcIDyRun.

Please let me if anything need to be modify. Thanks!!


libs/community/langchain_community/utilities/pubmed_extract_fulltext.py
```diff
@@ -0,0 +1,603 @@
...
+    # model_config = ConfigDict(ignored_types=(IgnoredType,))
+    parse: Any  #: :meta private:
+
+    # Default values for the parameters
+
+    pmc_url = "https://www.ncbi.nlm.nih.gov/pmc/oai/oai.cgi"
+    pmid_convert_url = "https://www.ncbi.nlm.nih.gov/pmc/utils/idconv/v1.0/"
+
+    # Paths to resource files
+    pmids_fulltext_path = os.path.join(os.path.dirname(__file__), "pmids_fulltext.txt")
```
eyurtsev, line 49: What are these resource files? Are they ever used by the integration?

eyurtsev (changes requested):
Hi! Thank you for trying spending the time to write the client.

The code looks useful, but it'll require quite a bit more work before we can merge it.

It looks like there are a lot of different parsing libraries involved:

bs4, xml, lxml, xmltodict

High level:

* The parsing logic should be isolated and pulled out of the client, the logic will need to be cleaned up and unit tested.
* The utility itself needs more documentation to explain what it does: e.g., why is s3 necessary? Why are files paths declared in the init etc. Which APIs it is using etc.

libs/community/langchain_community/utilities/pubmed_extract_fulltext.py
XinlingWang0628: Hi Eugene, Thanks for quick review and suggestions. I will update the PR. 
These resource can useful to retrieve full-text by pmid. Convert pmid to pmcid, then retrieve full-text. I will add a function to achieve it. Thanks.

XinlingWang0628:
Hi @eyurtsev, I made parsing logic in a separate file, created unit test for it and added explantation. Unit tests all passed, but there are some independency errors which I might need some guidance. For example, I install boto3 followed instruction [here,](https://python.langchain.com/v0.1/docs/contributing/code/#working-with-optional-dependencies) but error message said " ModuleNotFoundError: No module named 'boto3'".

eyurtsev:
hi @XinlingWang0628  -- we won't be able to merge this code into langchain-community in its current state. There's a lot of specialized parsing and too many different dependencies being used. This is not code that will be easy for other developers to maintain.

What I would suggest is starting a new repository and publishing a client from that repository that's a standalone client. Then hopefully things are modular enough that you could be able have a light weight integration using the library client wrapper via an appropriate langchain abstraction

XinlingWang0628:
Hi @eyurtsev ,

I've updated the PMC parsing logic to make it more lightweight by removing the S3 download functionality, pmid_convert_to_pmcid and its dependencies. However, I've encountered a circular import issue while testing: `ImportError: cannot import name 'Response' from partially initialized module 'requests' (most likely due to a circular import) (/Users/xinling/langchain/libs/community/langchain_community/utilities/requests.py)` I haven't been able to pass the test cases yet due to this issue. Happy to modify the test cases if needed once this error is fixed.

efriis:
going to close again - still too many issues unfortunately. Would recommend checking CI output, and also implementing one of the langchain abstractions! here's our integration guide if you want to take another go in the future:

PR #25375, opened 2024-08-14
community[minor]: Opensearch hybridsearch implementation

karthikbharadhwajKB:
community: add hybrid search in opensearch

# Langchain OpenSearch Hybrid Search Implementation

## Implementation of Hybrid Search: 

I have taken LangChain's OpenSearch integration to the next level by adding hybrid search capabilities. Building on the existing OpenSearchVectorSearch class, I have implemented Hybrid Search functionality (which combines the best of both keyword and semantic search). This new functionality allows users to harness the power of OpenSearch's advanced hybrid search features without leaving the familiar LangChain ecosystem. By blending traditional text matching with vector-based similarity, the enhanced class delivers more accurate and contextually relevant results. It's designed to seamlessly fit into existing LangChain workflows, making it easy for developers to upgrade their search capabilities.

In implementing the hybrid search for OpenSearch within the LangChain framework, I also incorporated filtering capabilities. It's important to note that according to the OpenSearch hybrid search documentation, only post-filtering is supported for hybrid queries. This means that the filtering is applied after the hybrid search results are obtained, rather than during the initial search process.

**Note:** For the implementation of hybrid search, I strictly followed the official OpenSearch Hybrid search documentation and I took inspiration from https://github.com/AndreasThinks/langchain/tree/feature/opensearch_hybrid_search
Thanks Mate!  

### Experiments

I conducted few experiments to verify that the hybrid search implementation is accurate and capable of reproducing the results of both plain keyword search and vector search.

Experiment - 1
Hybrid Search
Keyword_weight: 1, vector_weight: 0

I conducted an experiment to verify the accuracy of my hybrid search implementation by comparing it to a plain keyword search. For this test, I set the keyword_weight to 1 and the vector_weight to 0 in the hybrid search, effectively giving full weightage to the keyword component. The results from this hybrid search configuration matched those of a plain keyword search, confirming that my implementation can accurately reproduce keyword-only search results when needed. It's important to note that while the results were the same, the scores differed between the two methods. This difference is expected because the plain keyword search in OpenSearch uses the BM25 algorithm for scoring, whereas the hybrid search still performs both keyword and vector searches before normalizing the scores, even when the vector component is given zero weight. This experiment validates that my hybrid search solution correctly handles the keyword search component and properly applies the weighting system, demonstrating its accuracy and flexibility in emulating different search scenarios.

Experiment - 2
Hybrid Search
keyword_weight = 0.0, vector_weight = 1.0

For experiment-2, I took the inverse approach to further validate my hybrid search implementation. I set the keyword_weight to 0 and the vector_weight to 1, effectively giving full weightage to the vector search component (KNN search). I then compared these results with a pure vector search. The outcome was consistent with my expectations: the results from the hybrid search with these settings exactly matched those from a standalone vector search. This confirms that my implementation accurately reproduces vector search results when configured to do so. As with the first experiment, I observed that while the results were identical, the scores differed between the two methods. This difference in scoring is expected and can be attributed to the normalization process in hybrid search, which still considers both components even when one is given zero weight. This experiment further validates the accuracy and flexibility of my hybrid search solution, demonstrating its ability to effectively emulate pure vector search when needed while maintaining the underlying hybrid search structure.

Experiment - 3
Hybrid Search - balanced

keyword_weight = 0.5, vector_weight = 0.5

For experiment-3, I adopted a balanced approach to further evaluate the effectiveness of my hybrid search implementation. In this test, I set both the keyword_weight and vector_weight to 0.5, giving equal importance to keyword-based and vector-based search components. This configuration aims to leverage the strengths of both search methods simultaneously. By setting both weights to 0.5, I intended to create a scenario where the hybrid search would consider lexical matches and semantic similarity equally. This balanced approach is often ideal for many real-world applications, as it can capture both exact keyword matches and contextually relevant results that might not contain the exact search terms.

Kindly verify the notebook for the experiments conducted!  

**Notebook:** https://github.com/karthikbharadhwajKB/Langchain_OpenSearch_Hybrid_search/blob/main/Opensearch_Hybridsearch.ipynb

### Instructions to follow for Performing Hybrid Search:

**Step-1: Instantiating OpenSearchVectorSearch Class:**
```python
opensearch_vectorstore = OpenSearchVectorSearch(
    index_name=os.getenv("INDEX_NAME"),
    embedding_function=embedding_model,
    opensearch_url=os.getenv("OPENSEARCH_URL"),
    http_auth=(os.getenv("OPENSEARCH_USERNAME"),os.getenv("OPENSEARCH_PASSWORD")),
    use_ssl=False,
    verify_certs=False,
    ssl_assert_hostname=False,
    ssl_show_warn=False
)
```

**Parameters:**
1. **index_name:** The name of the OpenSearch index to use.
2. **embedding_function:** The function or model used to generate embeddings for the documents. It's assumed that embedding_model is defined elsewhere in the code.
3. **opensearch_url:** The URL of the OpenSearch instance.
4. **http_auth:** A tuple containing the username and password for authentication.
5. **use_ssl:** Set to False, indicating that the connection to OpenSearch is not using SSL/TLS encryption.
6. **verify_certs:** Set to False, which means the SSL certificates are not being verified. This is often used in development environments but is not recommended for production.
7. **ssl_assert_hostname:** Set to False, disabling hostname verification in SSL certificates.
8. **ssl_show_warn:** Set to False, suppressing SSL-related warnings.

**Step-2: Configure Search Pipeline:**

To initiate hybrid search functionality, you need to configures a search pipeline first. 

**Implementation Details:**

This method configures a search pipeline in OpenSearch that:
1. Normalizes the scores from both keyword and vector searches using the min-max technique.
2. Applies the specified weights to the normalized scores.
3. Calculates the final score using an arithmetic mean of the weighted, normalized scores.

**Parameters:**

* **pipeline_name (str):** A unique identifier for the search pipeline. It's recommended to use a descriptive name that indicates the weights used for keyword and vector searches.
* **keyword_weight (float):** The weight assigned to the keyword search component. This should be a float value between 0 and 1. In this example, 0.3 gives 30% importance to traditional text matching.
* **vector_weight (float):** The weight assigned to the vector search component. This should be a float value between 0 and 1. In this example, 0.7 gives 70% importance to semantic similarity.

```python
opensearch_vectorstore.configure_search_pipelines(
    pipeline_name="search_pipeline_keyword_0.3_vector_0.7",
    keyword_weight=0.3,
    vector_weight=0.7,
)
```

**Step-3: Performing Hybrid Search:**

After creating the search pipeline, you can perform a hybrid search using the `similarity_search()` method (or) any methods that are supported by `langchain`. This method combines both `keyword-based and semantic similarity` searches on your OpenSearch index, leveraging the strengths of both traditional information retrieval and vector embedding techniques.

**parameters:**
* **query:** The search query string.
* **k:** The number of top results to return (in this case, 3).
* **search_type:** Set to `hybrid_search` to use both keyword and vector search capabilities.
* **search_pipeline:** The name of the previously created search pipeline.

```python
query = "what are the country named in our database?"

top_k = 3

pipeline_name = "search_pipeline_keyword_0.3_vector_0.7"

matched_docs = opensearch_vectorstore.similarity_search_with_score(
                query=query,
                k=top_k,
                search_type="hybrid_search",
                search_pipeline = pipeline_name
            )

matched_docs
```

twitter handle: @iamkarthik98


libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -732,7 +740,47 @@ async def adelete(
...
+                    vector_weight
+                    ]
+                }
+                }
+            }
+            }
+        ]
+        }
+        try:
+            response = requests.put(url, auth=auth, json=payload, verify=False)
```
eyurtsev, line 776: why is this not using the existing opensearch client?

eyurtsev, line 776: verify=False is a security risk

eyurtsev (changes requested):

eyurtsev:
Hi @karthikbharadhwajKB, thank you for the contribution!

We need the following changes to get this code merged:

1. Add integration tests that show the code is working as expected
2. Use the opensearch client instead of issuing separate network requests via the requests library

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
karthikbharadhwajKB: I'm actually using OpenSearch locally. that's why I set it to "False".

karthikbharadhwajKB: I couldn't find any OpenSearch client method to configure/create a search pipeline. I found this https://opensearch-project.github.io/opensearch-py/api-ref/clients/ingest_client.html#opensearchpy.client.ingest.IngestClient.put_pipeline  
In their documentation, they stated that put_pipeline() is used to create or update pipelines. 
I tried but I'm getting some exceptions. 

I will post the exception I'm getting. let me know if you know how to do it!

karthikbharadhwajKB: I understood, that for AOSS we need to be careful.

karthikbharadhwajKB: Hey @eyurtsev I found a way to do it! I will update the code and let you know.

karthikbharadhwajKB:
@eyurtsev 

- [✅ ] Use the OpenSearch client instead of issuing separate network requests via the requests library - I have removed using requests library and Now update it with the OpenSearch client API for configuring_search_pipeline & performing Hybrid Search.

- [✅]  Add integration tests that show the code is working as expected - I have performed some tests locally on hybrid search & hybrid search with post filtering and It is working as expected and also it's not breaking other search_types. 

Here you can see working Hybrid Search feature and some experiments (where I tried to reproduce the exact keyword search results and approximate_search with tweaking keyword_weight & vector_weight).  
Notebook: https://github.com/karthikbharadhwajKB/Langchain_OpenSearch_Hybrid_search/blob/main/Opensearch_Hybridsearch.ipynb
I also tested Hybrid_Search, Hybrid_Search with Post filter & default Approximate_Search functionalities. 
Notebook: https://github.com/karthikbharadhwajKB/Langchain_OpenSearch_Hybrid_search/blob/main/Tests.ipynb

karthikbharadhwajKB:
Hey buddy @eyurtsev 

- [✅] Add integration tests that show the code is working as expected - I have added integration tests for configuring_search_pipeline, hybrid_search, hybrid_search_with_score & hybrid_search_with_post_filter. 

Please have a look. if everything seems fine then merge it! 🚀

karthikbharadhwajKB:
Hey @eyurtsev, I have made the changes you requested for the merge. Please let me know if I'm missing anything.

karthikbharadhwajKB:
Hey @eyurtsev, I have made the changes you requested for the merge. Please let me know if I'm missing anything.

eyurtsev:
Apologies for the delay we had a code freeze for the 0.3 release. Taking a look

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -733,6 +801,120 @@ async def adelete(
...
+                    }
+                }
+            ],
+        }
+        try:
+            response = self.client.transport.perform_request(
+                method="PUT", url=path, body=payload
+            )
+            return response
+        except Exception as e:
```
eyurtsev, line 843: Could you remove the except Exception as e -- it's a no-op. Instead let's let the exception bubble up without the surrounding try / except + rethrow

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -733,6 +801,120 @@ async def adelete(
...
+            >>> search_pipeline_exists("my_pipeline_1")
+            True
+            >>> search_pipeline_exists("my_pipeline_2")
+            False
+        """
+        try:
+            existed_pipelines = self.client.transport.perform_request(
+                method="GET", url=f"/_search/pipeline/"
+            )
+        except Exception as e:
```
eyurtsev, line 869: same as above

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -401,6 +402,73 @@ def _default_painless_scripting_query(
     }
 
 
+def _default_hybrid_search(
+    query_text: str, query_vector: List[float], k: int = 4
+) -> Dict:
+    """Returns payload for performing hybrid search for given options.
+    Args: options: dict containing the following
```
eyurtsev, line 409: nit:https://google.github.io/styleguide/pyguide.html#244-decision  

take a look at the URL above that shows google style doc-strings

the spec is whitespace sensitive as far as I know, so it should be a new line after Args:, a new line before Args:, and a new line after Returns:

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -401,6 +402,73 @@ def _default_painless_scripting_query(
...
+        "size": k,
+    }
+
+    return payload
+
+
+def _hybrid_search_with_post_filter(
+    query_text: str, query_vector: List[float], k: int, post_filter: Optional[Dict] = {}
+) -> Dict:
+    """Returns payload for performing hybrid search with post filter for given options.
```
eyurtsev, line 441: same nit as above

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -401,6 +402,73 @@ def _default_painless_scripting_query(
...
+            }
+        },
+        "size": k,
+    }
+
+    return payload
+
+
+def _hybrid_search_with_post_filter(
+    query_text: str, query_vector: List[float], k: int, post_filter: Optional[Dict] = {}
```
eyurtsev, line 439: let's not assign a mutable default, 
```python
post_filter: Optional[Dict] = None,
```

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -1043,6 +1227,34 @@ def _raw_similarity_search_with_score_by_vector(
                 vector_field,
                 score_threshold=score_threshold,
             )
+
+        elif search_type == HYBRID_SEARCH:
+            search_pipeline = kwargs.get("search_pipeline", "")
+            post_filter = kwargs.get("post_filter", {})
+            query_text = kwargs.get("query_text", "")
```
eyurtsev, line 1234: is an empty `query_text` an actual possible path? if not, assert that query text exists in kwargs, and if it doesn't raise a helpful error for the user

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -1043,6 +1227,34 @@ def _raw_similarity_search_with_score_by_vector(
...
+            else:
+                # hybrid search without post filter
+                payload = _default_hybrid_search(query_text, embeded_query, k)
+            try:
+                response = self.client.transport.perform_request(
+                    method="GET", url=path, body=payload
+                )
+                return [hit for hit in response["hits"]["hits"]]
+
+            except Exception as e:
```
eyurtsev, line 1255: the try except doesn't do anything

libs/community/tests/integration_tests/vectorstores/test_opensearch.py
```diff
@@ -66,6 +66,112 @@ def test_opensearch_with_custom_field_name() -> None:
...
+        texts,
+        ConsistentFakeEmbeddings(),
+        metadatas=metadatas,
+        opensearch_url=DEFAULT_OPENSEARCH_URL,
+    )
+    output = docsearch.similarity_search(
+        query="foo",
+        k=2,
+        search_type=HYBRID_SEARCH,
+        search_pipeline="test_search_pipeline",
```
eyurtsev, line 125: this test should be independent of other tests. It shouldn't assume that a test search pipeline exists. Could you implement a clean up step in the tests above, so they don't live a configured pipeline?

eyurtsev:
Hi @karthikbharadhwajKB -- this is looking pretty close to done. Apologies for the delay in reviewing. I left a few minor comments.

libs/community/tests/integration_tests/vectorstores/test_opensearch.py
karthikbharadhwajKB: But to perform the hybrid search. it needs a search_pipeline. if we don't pass the search_pipeline name it will raise an error.

karthikbharadhwajKB: I also have another idea to merge both configure_search_pipeline  & hybrid_Search into a single functionality 

where the user directly passes search_type = hybrid_search and weightage for (keyword_search & vector_search) in tuple (0.7, 0.3) -> internally we can check if the search_pipeline already exists. if it exists, it will use it or else it will configure a new search pipeline.

karthikbharadhwajKB:
Hey @eyurtsev I have made the changes that you have requested. please check and let me know.

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -733,6 +793,115 @@ async def adelete(
...
+            False
+        """
+        
+        existed_pipelines = self.client.transport.perform_request(
+            method="GET", url=f"/_search/pipeline/"
+        )
+
+        return pipeline_name in existed_pipelines
+
+    def get_search_pipeline_info(self, pipeline_name: str) -> Optional[Dict]:
```
eyurtsev, line 863: sorry just caught this: `pipeline_name` needs to be sanitized here potentially -- i don't know what the underlying client being used, but there's a chance that escaping will be possible.
You could verify that pipeline_name is a proper identifier (single if statement)

```
In [6]: "..qwe".isidentifier()   
```

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -733,6 +793,115 @@ async def adelete(
...
+            pipeline_name: Name of the pipeline
+            keyword_weight: Weight for keyword search
+            vector_weight: Weight for vector search
+        Returns:
+            response: Acknowledgement of the pipeline creation.
+            (if there is any error while configuring the pipeline, it will return None)
+        Raises:
+            Exception: If an error occurs
+        """
+        path = f"/_search/pipeline/{pipeline_name}"
```
eyurtsev, line 814: see comment below about pipeline_name

karthikbharadhwajKB:
Hey @eyurtsev I have made the changes that you requested to do. 

Please merge if everything looks fine 🚀

eyurtsev (approved):

karthikbharadhwajKB:
Hey @eyurtsev, I can see that there are 3 checks were failed. Can you please let me know what should I do?

karthikbharadhwajKB:
Hey @eyurtsev, I can see that there are 3 checks were failed. Can you please let me know what should I do?

karthikbharadhwajKB:
Hey @eyurtsev, I can see that there are 3 checks were failed. Can you please let me know what should I do?

karthikbharadhwajKB:
Thank you @eyurtsev, Now everything seems fine from my side. Please Merge this feature branch if everything is fine.

eyurtsev (approved):

Issue #24621, opened 2024-07-24
TypeError('Object of type CallbackManagerForToolRun is not JSON serializable') on Coder agent


arthur-lachini-advisia:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the [LangGraph](https://langchain-ai.github.io/langgraph/)/LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangGraph/LangChain rather than my code.
- [X] I am sure this is better as an issue [rather than a GitHub discussion](https://github.com/langchain-ai/langgraph/discussions/new/choose), since this is a LangGraph bug and not a design question.

### Example Code

```python
for s in graph.stream(
    {
        "messages": [
            HumanMessage(content="Code hello world and print it to the terminal")
        ]
    }
):
    if "__end__" not in s:
        print(s)
        print("----")
```

### Error Message and Stack Trace (if applicable)

```shell
TypeError('Object of type CallbackManagerForToolRun is not JSON serializable')Traceback (most recent call last):


  File "c:\Users\arthur.lachini\AppData\Local\Programs\Python\Python312\Lib\site-packages\langgraph\pregel\__init__.py", line 946, in stream
    _panic_or_proceed(done, inflight, loop.step)


  File "c:\Users\arthur.lachini\AppData\Local\Programs\Python\Python312\Lib\site-packages\langgraph\pregel\__init__.py", line 1347, in _panic_or_proceed
    raise exc


  File "c:\Users\arthur.lachini\AppData\Local\Programs\Python\Python312\Lib\site-packages\langgraph\pregel\executor.py", line 60, in done
    task.result()


  File "c:\Users\arthur.lachini\AppData\Local\Programs\Python\Python312\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^


  File "c:\Users\arthur.lachini\AppData\Local\Programs\Python\Python312\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception


  File "c:\Users\arthur.lachini\AppData\Local\Programs\Python\Python312\Lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


[...]

PR #25377, opened 2024-08-14
Update tools.py

karthikcs:
Explicitly return the dict, if not we get JSON parsing error

- [ ] **PR title**: "core: tools.py should return dict, but it was not explicitly converted, hence was throwing the JSON parsing"

- [ ] **PR message**: 
    - **Description:** Explicitly return the dict, if not we get JSON parsing error
    - **Issue:** TypeError('Object of type CallbackManagerForToolRun is not JSON serializable') on Coder agent
    - **Dependencies:** No dependencies
    - **Twitter handle:** @karthikseeyes

  Unit test doc: https://github.com/karthikcs/langchain/blob/master/Unit%20test%20tools.py.pdf
 

### DIFF
--- a/Unit Test/code_test.py
+++ b/Unit Test/code_test.py
@@ -1,0 +1,61 @@
+from langchain_openai import AzureChatOpenAI
+from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
+from langchain.agents import AgentExecutor, create_openai_tools_agent
+from langchain_experimental.tools import PythonREPLTool
+from llm_factory import llm
+from langchain_core.messages import BaseMessage, HumanMessage
+
+
+def create_agent(llm: AzureChatOpenAI, tools: list, system_prompt: str):
+    """
+    Creates an agent with the given LLM, tools, and system prompt.
+
+    Args:
+        llm (AzureChatOpenAI): The LLM used by the agent.
+        tools (list): The list of tools used by the agent.
+        system_prompt (str): The system prompt for the agent.
+
+    Returns:
+        AgentExecutor: The created agent executor.
+
+    """
+    prompt = ChatPromptTemplate.from_messages(
+        [
+            (
+                "system",
+                system_prompt,
+            ),
+            MessagesPlaceholder(variable_name="messages"),
+            MessagesPlaceholder(variable_name="agent_scratchpad"),
+        ]
+    )
+    agent = create_openai_tools_agent(llm, tools, prompt)
+    executor = AgentExecutor(agent=agent, tools=tools, verbose=True, handle_parsing_errors=True)
+    return executor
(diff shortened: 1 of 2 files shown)


eyurtsev (changes requested):
Hi @karthikcs,

Thanks for the change!

Could you add a unit test that shows what test case is fixed by this?

eyurtsev:
@baskaryan tagging you in case its obvious why this is needed

karthikcs:
@eyurtsev I was facing an issue mentioned in the https://github.com/langchain-ai/langchain/issues/24621 and the solution worked out for me was https://github.com/langchain-ai/langchain/issues/24621#issuecomment-2249409492. I think it should be merged so that root cause is resolved.

karthikcs:
> Hi @karthikcs,
> 
> Thanks for the change!
> 
> Could you add a unit test that shows what test case is fixed by this?

OK I will add a UT

karthikcs:
> Hi @karthikcs,
> 
> Thanks for the change!
> 
> Could you add a unit test that shows what test case is fixed by this?

@eyurtsev I have added a UT file. Please review and approve

baskaryan:
hi @karthikcs, i don't follow what the issue here is. what type other than dict or string are we trying to pass to tools, and why?

hwchase17:
the unit test is not in the right format and I also dont realy understand the motivation. I would suggest opening an issue and we can proceed from there

karthikcs:
> hi @karthikcs, i don't follow what the issue here is. what type other than dict or string are we trying to pass to tools, and why?

@baskaryan Sorry for delayed response. I am not explicitly passing anything to this function. I am instantiating PythonREPL tool and it is failing consistently. When I debug I found following 

FILE: langchain_openai/chat_models/base.py 
function `_lc_tool_call_to_openai_tool_call`
The parameter `tool_call["args"]` is causing the issue while converting the python code to JSON in the line 
`json.dumps(tool_call["args"]),`

When I debugged I found the value in `tool_call["args"]` was 

```
{'query': 'product = 889384 * 77734\nproduct', 'run_manager': <langchain_core.callbacks.manager.CallbackManagerForToolRun object at 0x000001CD53DBAA80>}
```

Expected value is only the code, without the `run_manager` element.

```
{'query': 'product = 889384 * 77734\nproduct'}
```

When I add `dict(tool_input)` in the function `_to_args_and_kwargs` in (langchain_core/tools/base.py) the unnecessary element `run_manager` is not added to `tool_calls`. So, please consider this suggestion.

karthikcs:
> hi @karthikcs, i don't follow what the issue here is. what type other than dict or string are we trying to pass to tools, and why?

Dear Harrison, is there any guide I can follow?

PR #25399, opened 2024-08-14
feat(core): add async api to draw graph png

eyurtsev:
- feature: graph image rendering with playwright
- fix: playwright imports inside run_mermaid_using_playwright method
- chore: unused method removed

- Vendored in mermaid.min.js (would need to rebuild from source) => increases dist size by ~ 1MB (from 300kb to 1.3 MB)
- adds async APIs to avoid funny business with event loop


libs/core/langchain_core/runnables/graph_mermaid.py
```diff
@@ -164,28 +177,276 @@ def draw_mermaid_png(
...
+    if draw_method_ == MermaidDrawMethod.PLAYWRIGHT:
+        img_bytes = await _render_mermaid_using_playwright(
+            mermaid_syntax, output_file_path, background_color, padding
+        )
+    elif draw_method_ == MermaidDrawMethod.PYPPETEER:
+        img_bytes = await _render_mermaid_using_pyppeteer(
+            mermaid_syntax, output_file_path, background_color, padding
+        )
+    elif draw_method_ == MermaidDrawMethod.API:
+        img_bytes = _render_mermaid_using_api(
```
nfcampos, line 260: this must be blocking event loop?

libs/core/langchain_core/runnables/graph_mermaid.py
```diff
@@ -164,28 +177,276 @@ def draw_mermaid_png(
     Raises:
         ValueError: If an invalid draw method is provided.
     """
-    if draw_method == MermaidDrawMethod.PYPPETEER:
+    if draw_method == "api":
+        draw_method_ = MermaidDrawMethod.API
+    elif draw_method == "playwright":
+        draw_method_ = MermaidDrawMethod.PLAYWRIGHT
+    elif draw_method == "pyppeteer":
+        draw_method_ = MermaidDrawMethod.PYPPETEER
```
nfcampos, line 185: how does this one work in sync?

eyurtsev: Currently, kicks off an event loop even if one exists already (so requires usage of nest_asyncio). We can kick it off on a thread to fix this up. (Don't think it offers a sync API.)

marctorsoc:
@eyurtsev Any chance you'll finish this? I'd really like to use playwright to render mermaid :)

PR #25432, opened 2024-08-15
convert _DocumentWithState to Document in ContextualCompressionRetriever,

hpx502766238:
To solve this problem:https://github.com/langchain-ai/langchain/discussions/17876#discussion-6249465
convert _DocumentWithState to Document,because type (numpy.float64) in  _DocumentWithState is not serializable when expose a retriever by langserve.


eyurtsev (changes requested):
Hi thanks for the contribution. This does not appear the correct place to fix the issue.

`contextual_compression` should not be doing this kind of a change.

This should either be fixed in:
* The entity is populating the documents with numpy arrays
* langserve by allowing users to specify a custom serializer 
* User's code in langserve -- the runnable may need another step to handle non serializable values

eyurtsev:
I'm going to close the PR since this is not the place where the change should be made. Feel free to re-open / comment if you do not agree

PR #23971, opened 2024-07-08
community: chat_message_history.dynamodb Update DynamoDB to optionally use an item per message

LeeroyHannigan:
Thank you for contributing to LangChain!
- **Description:**
     This add an additional parameter to `DynamoDBChatMessageHistory` that allows each message to be contained in its own item. This is a best practice for DynamoDB, and avoids hitting the 400KB item size limit. 
- **Issue:**
    The issue is the chance of hitting the 400KB items size limit in DynamoDB. `history_size` works to some degree, but at the risk of losing relevant parts of the conversation.
- **Dependencies:** any dependencies required for this change
    Imported `boto3.dynamodb.conditions import Key` which is already a dependency with `boto3`
- **Twitter handle:** if your PR gets announced, and you'd like a mention, we'll gladly shout you out!
    `leeroy_hannigan`

- [ ] **Add tests and docs**: If you're adding a new integration, please include
  1. a test for the integration, preferably unit tests that do not rely on network access,
  2. an example notebook showing its use. It lives in `docs/docs/integrations` directory.

- [ ] **Lint and test**: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified. See contribution guidelines for more: https://python.langchain.com/docs/contributing/

Additional guidelines:
- Make sure optional dependencies are imported within a function.
- Please do not add dependencies to pyproject.toml files (even optional ones) unless they are required for unit tests.
- Most PRs should not touch more than one package.
- Changes should be backwards compatible.
- If you are adding something to community, do not re-import it in langchain.

If no one reviews your PR within a few days, please @-mention one of baskaryan, efriis, eyurtsev, ccurme, vbarda, hwchase17.

PR #20275, opened 2024-04-10
langchain-mongodb: add history_size attribute to MongoDBChatMessageHistory

smalyu:
Description: This PR introduces a new attribute, history_size, to the MongoDBChatMessageHistory class, allowing users to set a maximum number of messages to store in the MongoDB database. This feature helps manage memory more effectively by limiting the history size. The implementation includes logic to automatically trim the oldest messages when the history limit is exceeded.

Issue: None

Dependencies: None

PR #20267, opened 2024-04-10
community: add history_size attribute to RedisChatMessageHistory

smalyu:
**Description**: This PR introduces a new attribute, `history_size`, to the `RedisChatMessageHistory` class, allowing users to set a maximum number of messages to store in the Redis database. This feature helps manage memory more effectively by limiting the history size. The implementation includes logic to automatically trim the oldest messages when the history limit is exceeded.

**Issue**: None

**Dependencies**: Requires the `redis` package, which is already included in the existing dependencies.

PR #25435, opened 2024-08-15
limit the most recent documents to fetch from MongoDB database.

kk9393:
limit the most recent documents to fetch from MongoDB database.

- [ ] **limit the most recent documents to fetch from MongoDB database.**: "langchain_mongodb: limit the most recent documents to fetch from MongoDB database."

    - **Description:** Added a doc_limit parameter which enables the limit for the documents to fetch from MongoDB database
    - **Issue:** 
    - **Dependencies:** None


libs/partners/mongodb/langchain_mongodb/chat_message_histories.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         session_id_key: str = DEFAULT_SESSION_ID_KEY,
         history_key: str = DEFAULT_HISTORY_KEY,
         create_index: bool = True,
+        doc_limit: int = None,
```
eyurtsev, line 71: Could you rename the parameter and fix the type signature? (it's an Optional[Int])


`doc_limit` is not a meaningful name in this context. You can do something like `max_messages` and document that this is a limit on the max number of messages that can be fetched.

---

How are you using this capability with chat history? Specifically, how are you dealing with the system message (the first message in a chat history conversation).

ccurme (commented):
Thanks for this. Would you mind adding a test to verify the implementation?

libs/partners/mongodb/langchain_mongodb/chat_message_histories.py
kk9393: Hi @eyurtsev Thanks for quick response. Your suggestions are helpful. I have changed the file according to your suggestions.

I haven't handled such scenario of dealing with initial messages. Basically I am building this code as I get in trouble while developing application. I am sure I will have to handle such cases in the future and will contribute to it soon.

kk9393:
> Thanks for this. Would you mind adding a test to verify the implementation?

@ccurme Thanks for the reply.

I have tested few scenarios on my local machine and it is working fine. Can you please guide me on how do I add a test to verify the implementation? Thanks! :-)

libs/partners/mongodb/langchain_mongodb/chat_message_histories.py
```diff
@@ -97,6 +100,7 @@ def __init__(
         self.collection_name = collection_name
         self.session_id_key = session_id_key
         self.history_key = history_key
+        self.doc_limit = doc_limit
```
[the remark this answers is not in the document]
kk9393: because developer or the app will not have the access to number of messages in chat session at any point of time. But developer can fix the max messages he/she wants in their application. I have changed the parameter name to max_messages as per @eyurtsev's suggestion.

I hope it explains. Please let me know if you have any more suggestions Thanks!

eyurtsev:
@ccurme assigning to you.

For the record, see related PRs:

https://github.com/langchain-ai/langchain/pull/20267
https://github.com/langchain-ai/langchain/pull/23971/files#diff-45be1b144f8385979abf22e583484ac0b441ef3367c1af334885375197bb98b5R61

Looks like `history-size ` is the current name in other implementations

And rejected PR from mondo DB: https://github.com/langchain-ai/langchain/pull/20275

kk9393:
> @ccurme assigning to you.
> 
> For the record, see related PRs:
> 
> #20267 https://github.com/langchain-ai/langchain/pull/23971/files#diff-45be1b144f8385979abf22e583484ac0b441ef3367c1af334885375197bb98b5R61
> 
> Looks like `history-size ` is the current name in other implementations
> 
> And rejected PR from mondo DB: #20275
> 
> we should figure out whether we want to support this or not on the integration
> 
> One one hand I don't mind - on the other hand it doesn't actually solve the problem that users have even though it appears to.

@eyurtsev although it doesn't summerise the conversation, it trims the conversation. Which is one of the two methods listed in the docs for handling memory.

ccurme:
> > Thanks for this. Would you mind adding a test to verify the implementation?
> 
> @ccurme Thanks for the reply.
> 
> I have tested few scenarios on my local machine and it is working fine. Can you please guide me on how do I add a test to verify the implementation? Thanks! :-)

My suggestion is to add a test here: https://github.com/langchain-ai/langchain/blob/master/libs/partners/mongodb/tests/integration_tests/test_chat_message_histories.py

libs/partners/mongodb/langchain_mongodb/chat_message_histories.py
```diff
@@ -114,7 +118,8 @@ def __init__(
     def messages(self) -> List[BaseMessage]:  # type: ignore
         """Retrieve the messages from MongoDB"""
         try:
-            cursor = self.collection.find({self.session_id_key: self.session_id})
+            skip_count = max(0, self.collection.count_documents({}) - self.doc_limit) if self.doc_limit != None else 0
+            cursor = self.collection.find({self.session_id_key: self.session_id}, skip=skip_count)
```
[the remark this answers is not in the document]
ccurme: I may be misunderstanding what you are saying, but given that

> 0 is giving an error

Can you clarify why we are setting the default to be zero?

libs/partners/mongodb/langchain_mongodb/chat_message_histories.py
```diff
@@ -114,7 +118,8 @@ def __init__(
     def messages(self) -> List[BaseMessage]:  # type: ignore
         """Retrieve the messages from MongoDB"""
         try:
-            cursor = self.collection.find({self.session_id_key: self.session_id})
+            skip_count = max(0, self.collection.count_documents({}) - self.max_messages) if self.max_messages != None else 0
```
ccurme, line 121: Instead of calling `collection.count_documents` each time, is there a more appropriate method to use? e.g., `limit`? `skip` is skipping the first N messages instead of limiting the results to N messages, right?

I think ideally we expose an argument that limits the number of messages returned and name it `history_size` (since as @eyurtsev points out this name has been chosen for similar purposes elsewhere).

Failing that, my preference would be to expose `skip` directly and just let users access it.

libs/partners/mongodb/langchain_mongodb/chat_message_histories.py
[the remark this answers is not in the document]
kk9393: @ccurme , I have given thought over that and I think ideal default value would be something like 4 or 6 where application has access to last 2-3 conversions by default. Please let me know if you have suggestions. Should I change it to let's say 6?

libs/partners/mongodb/langchain_mongodb/chat_message_histories.py
kk9393: The query is giving top few results from the starting of conversation If we use `limit`. Instead we would like to have latest messages to append into history.

If we expose `skip`, then user will have to query mongodb database (`collection.count_documents`) every time to count the messages in collection, which is suboptimal solution from point of end user.

I think `history_size` as exposed parameter with default value of 6 or 8 is an ideal condition.

Please let me know your thoughts. I will change the code accordingly and commit.

ccurme (commented):
Hi @kk9393,

I pushed up a minor update. The implementation is largely as you had it. Would appreciate your thoughts on:

- Is using `.count_documents` and `skip` performant and reliable? Is it susceptible to race conditions-- e.g., updates to the table in between `.count_documents` and `skip`?
- If there are potential issues, are there any alternative implementations that are preferable? e.g., something like `collection.find().sort("_id", -1).limit(history_size)`.

Regarding the default: I think it's important to keep at None to avoid breaking changes for current users.

A test in https://github.com/langchain-ai/langchain/blob/master/libs/partners/mongodb/tests/integration_tests/test_chat_message_histories.py would also be welcome.

kk9393:
> Hi @kk9393,
> 
> I pushed up a minor update. The implementation is largely as you had it. Would appreciate your thoughts on:
> 
> * Is using `.count_documents` and `skip` performant and reliable? Is it susceptible to race conditions-- e.g., updates to the table in between `.count_documents` and `skip`?
> * If there are potential issues, are there any alternative implementations that are preferable? e.g., something like `collection.find().sort("_id", -1).limit(history_size)`.
> 
> Regarding the default: I think it's important to keep at None to avoid breaking changes for current users.
> 
> A test in https://github.com/langchain-ai/langchain/blob/master/libs/partners/mongodb/tests/integration_tests/test_chat_message_histories.py would also be welcome.

Hi @ccurme ,

Thanks for the update.

I don't think there will be a race problem at least in chat application cases since every conversation has it's own sessions. New docs can not be pushed until LLM reply is received.

The problem with ```collection.find().sort("_id", -1).limit(history_size)``` is that it will retrieve the messages in reverse order.

In any case, we are currently working on MongoDB implementation and will check whether we get any problems with the implementation. Thanks for your inputs.

Also, our team members will keep contributing to the Langchain MongoDB implementation in coming days. I think the version you have pushed right now is the best possible solution we have at the moment.

Thanks!

ccurme (approved):

kk9393:
Hi @ccurme 

Thank you for merging it! Could you please let me know when the next version of the `langchain-mongodb` pip package will be released? I currently see v0.1.8, which I believe was released last month. I’m unable to access the `history_size` feature in that version.

Thanks.

BololZ:
Hi,

Thanks for the feature, l’m looking for it, perfect.
But it's not available in production. Can I help ?

Thanks

efriis:
will release in #26193

PR #25449, opened 2024-08-15
RFC: Add ExampleSelector.add_examples

baskaryan:


libs/core/langchain_core/example_selectors/base.py
```diff
@@ -26,6 +27,37 @@ async def aadd_example(self, example: Dict[str, str]) -> Any:
...
+
+        Args:
+            examples: A list of dictionaries mapping input variable names to their
+                values.
+        """
+
+        if self.__class__.add_examples != BaseExampleSelector.add_examples:
+            return await run_in_executor(None, self.add_examples, examples)
+        else:
+            coros = map(self.aadd_example, examples)
```
baskaryan, line 57: could check if aadd_example has non-default implementation and just call add_example directly if not

libs/core/langchain_core/example_selectors/base.py
```diff
@@ -26,6 +27,37 @@ async def aadd_example(self, example: Dict[str, str]) -> Any:
 
         return await run_in_executor(None, self.add_example, example)
 
+    def add_examples(self, examples: List[Dict[str, str]]) -> Any:
```
eyurtsev, line 30: Could we swap to Sequence types instead of List t types?

eyurtsev (approved):

PR #25483, opened 2024-08-16
community: added stream and astream to chatyandexgpt

olgamurraft:
**Description**
added stream and astream support to ChatYandexGPT


libs/community/langchain_community/chat_models/yandex.py
```diff
@@ -120,11 +121,45 @@ async def _agenerate(
...
+        self,
+        messages: List[BaseMessage],
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> Iterator[ChatGenerationChunk]:
+        stream_resp = completion_with_retry(self, messages=messages, stream=True)
+        current_text = ""
+        for data in stream_resp:
+            delta = data[len(current_text) :]
```
vhaldemar, line 134: You can't do this, because the common prefix of the generations may change during streaming from yandex gpt :(

PR #25501, opened 2024-08-16
standard-tests[patch]: async variations of all tests

baskaryan:


libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
```diff
@@ -133,7 +133,7 @@ def test_conversation(self, model: BaseChatModel) -> None:
         assert isinstance(result.content, str)
         assert len(result.content) > 0
 
-    def test_usage_metadata(self, model: BaseChatModel) -> None:
+    async def test_usage_metadata(self, model: BaseChatModel) -> None:
```
eyurtsev, line 136: Could you duplicate the sync test and create an async version of it? Instead of doing both sync and async in the same test?

eyurtsev (commented):
We should try to separate sync and async tests.

The tests should ideally be completely separated at the class level

PR #25514, opened 2024-08-17
core[patch]: content block prompt template

baskaryan:
Screenshot 2024-08-16 at 5 37 23 PM


ccurme:
Looks good.

How should `ImagePromptTemplate` ideally be related to `ContentBlockPromptTemplate`?

The example in the screenshot is helpful to understand a use-case we're enabling (e.g., a test with both text + image blocks in one message or any other illustrative examples).

ccurme (approved):

libs/core/langchain_core/prompts/chat.py
```diff
@@ -459,12 +460,12 @@ class _ImageTemplateParam(TypedDict, total=False):
...
 
-    prompt: Union[
-        StringPromptTemplate, List[Union[StringPromptTemplate, ImagePromptTemplate]]
-    ]
+    prompt: Union[StringPromptTemplate, List[BasePromptTemplate]]
     """Prompt template."""
     additional_kwargs: dict = Field(default_factory=dict)
     """Additional keyword arguments to pass to the prompt template."""
+    name: Optional[str] = None
+    """Name of message"""
```
efriis, line 468: name of the actor right?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -626,20 +625,31 @@ def format(self, **kwargs: Any) -> BaseMessage:
         if isinstance(self.prompt, StringPromptTemplate):
             text = self.prompt.format(**kwargs)
             return self._msg_class(
-                content=text, additional_kwargs=self.additional_kwargs
+                content=text, additional_kwargs=self.additional_kwargs, name=self.name
```
efriis, line 628: might be worth removing the default on the name field to discover if anywhere missed in plumbing (and then re-add before merging). Feel like there are more places this should be passed through but maybe not?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -626,20 +625,31 @@ def format(self, **kwargs: Any) -> BaseMessage:
...
             content: List = []
             for prompt in self.prompt:
                 inputs = {var: kwargs[var] for var in prompt.input_variables}
                 if isinstance(prompt, StringPromptTemplate):
-                    formatted: Union[str, ImageURL] = prompt.format(**inputs)
+                    formatted: Union[str, dict] = prompt.format(**inputs)
                     content.append({"type": "text", "text": formatted})
                 elif isinstance(prompt, ImagePromptTemplate):
-                    formatted = prompt.format(**inputs)
+                    formatted = cast(dict, prompt.format(**inputs))
```
efriis, line 638: why is this return type not a dict?

libs/core/tests/unit_tests/prompts/test_chat.py
```diff
@@ -863,3 +875,32 @@ async def test_chat_tmpl_serdes(snapshot: SnapshotAssertion) -> None:
     )
     assert dumpd(template) == snapshot()
     assert load(dumpd(template)) == template
+
+
+def test_chat_content_block() -> None:
```
efriis, line 880: should we write some serialization snapshot tests and run them before and after this pr?

efriis (commented):
Summary of how this remaps inheritance would be great - I think only adds contentblock to BasePromptTemplate and then swaps union for BasePromptTemplate

but this would also allow passing in ChatPromptTemplate in some areas which might not be allowed...

libs/core/langchain_core/prompts/chat.py
baskaryan: its a typed dict that mypy is annoying about

libs/core/tests/unit_tests/prompts/test_chat.py
baskaryan: already done (see test above)

libs/core/langchain_core/prompts/chat.py
baskaryan: mm where else would it be outside of format/aformat?

libs/core/langchain_core/prompts/chat.py
efriis: should update docstring

libs/core/langchain_core/prompts/chat.py
efriis: maybe ChatPromptTemplate.from_template?

efriis: Thoughts? https://github.com/langchain-ai/langchain/pull/25514/commits/a11f6118a5c2b6dbfe50206190703072ff8aab7e

PR #25520, opened 2024-08-17
core: Add B(bugbear) ruff rules

cbornet:


libs/core/langchain_core/prompts/string.py
```diff
@@ -81,17 +82,17 @@ def validate_jinja2(template: str, input_variables: List[str]) -> None:
         warning_message += f"Extra variables: {extra_variables}"
 
     if warning_message:
-        warnings.warn(warning_message.strip())
+        logging.warning(warning_message.strip())
```
cbornet, line 85: From https://docs.python.org/3/howto/logging.html#when-to-use-logging :
> [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning
A logger’s [warning()](https://docs.python.org/3/library/logging.html#logging.Logger.warning) method if there is nothing the client application can do about the situation, but the event should still be noted

libs/core/langchain_core/tracers/evaluation.py
```diff
@@ -144,11 +144,7 @@ def _evaluate_in_project(self, run: Run, evaluator: langsmith.RunEvaluator) -> N
         example_id = str(run.reference_example_id)
         with self.lock:
             for res in eval_results:
-                run_id = (
-                    str(getattr(res, "target_run_id"))
-                    if hasattr(res, "target_run_id")
-                    else str(run.id)
-                )
+                run_id = str(getattr(res, "target_run_id", run.id))
```
cbornet, line 147: What if `res. target_run_id` is `None` ?

libs/core/langchain_core/tracers/core.py
```diff
@@ -516,15 +516,19 @@ def __copy__(self) -> _TracerCore:
 
     def _end_trace(self, run: Run) -> Union[None, Coroutine[Any, Any, None]]:
         """End a trace for a run."""
+        return None
```
cbornet, line 519: due to https://github.com/PyCQA/flake8-bugbear/issues/301
Do we prefer to ignore the rule with `noqa` ?

libs/core/langchain_core/vectorstores/base.py
```diff
@@ -568,7 +569,7 @@ def similarity_search_with_relevance_scores(
                 if similarity >= score_threshold
             ]
             if len(docs_and_similarities) == 0:
-                warnings.warn(
+                logger.warning(
```
cbornet, line 572: From https://docs.python.org/3/howto/logging.html#when-to-use-logging :
> [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning
A logger’s [warning()](https://docs.python.org/3/library/logging.html#logging.Logger.warning) method if there is nothing the client application can do about the situation, but the event should still be noted

libs/core/langchain_core/_api/beta_decorator.py
```diff
@@ -270,7 +270,7 @@ def warn_beta(
             message += f" {addendum}"
 
     warning = LangChainBetaWarning(message)
-    warnings.warn(warning, category=LangChainBetaWarning, stacklevel=2)
```
baskaryan, line 273: whats this do?

baskaryan (commented):
mostly curious to understand what impact changing warning stacklevel will have

libs/core/langchain_core/_api/beta_decorator.py
cbornet: It gives better information about which user-side call triggered the warning.
With stacklevel = 2
```
tests/unit_tests/language_models/chat_models/test_rate_limiting.py::test_rate_limit_invoke
  /Users/christophebornet/workspace/langchain/libs/core/langchain_core/_api/beta_decorator.py:87: LangChainBetaWarning: Introduced in 0.2.24. API subject to change.
    warn_beta(
```

With stacklevel = 4
```
tests/unit_tests/language_models/chat_models/test_rate_limiting.py::test_rate_limit_invoke
  /Users/christophebornet/workspace/langchain/libs/core/tests/unit_tests/language_models/chat_models/test_rate_limiting.py:13: LangChainBetaWarning: Introduced in 0.2.24. API subject to change.
    rate_limiter=InMemoryRateLimiter(
```
I believe warn_beta is always called through `@beta` so it makes more sense to warn on the caller of the `@beta` annotated class.
I tried to adjust all stacklevels with this in mind.

libs/core/langchain_core/runnables/base.py
```diff
@@ -4153,7 +4153,7 @@ def __init__(
     @property
     def InputType(self) -> Any:
         """The type of the input to this Runnable."""
-        func = getattr(self, "func", None) or getattr(self, "afunc")
+        func = getattr(self, "func", None) or self.afunc
```
baskaryan, line 4156: self.afunc isn't guaranteed to be set

libs/core/langchain_core/_api/beta_decorator.py
baskaryan: ooh thats great! yea the beta warnings aren't great atm

libs/core/langchain_core/runnables/base.py
cbornet: if it is not set then `getattr` will raise an Exception. Just like referencing `self.afunc`.
It's ruff rule B009: https://docs.astral.sh/ruff/rules/get-attr-with-constant/

baskaryan (approved):

libs/core/langchain_core/runnables/retry.py
baskaryan, line 1: cc @nfcampos

libs/core/langchain_core/utils/mustache.py
```diff
@@ -425,11 +429,13 @@ def _get_partial(name: str, partials_dict: Dict[str, str]) -> str:
...
 g_token_cache: Dict[str, List[Tuple[str, str]]] = {}
 
+EMPTY_DICT: MappingProxyType[str, str] = MappingProxyType({})
+
 
 def render(
     template: Union[str, List[Tuple[str, str]]] = "",
-    data: Dict[str, Any] = {},
-    partials_dict: Dict[str, str] = {},
+    data: Mapping[str, Any] = EMPTY_DICT,
```
baskaryan, line 437: should we just make this Optional and default to None?

libs/core/langchain_core/runnables/utils.py
```diff
@@ -748,7 +748,7 @@ def is_async_generator(
     """
     return (
         inspect.isasyncgenfunction(func)
-        or hasattr(func, "__call__")
```
nfcampos, line 751: I think this was on purpose cc @hinthornw I think

libs/core/langchain_core/tracers/core.py
```diff
@@ -533,39 +537,52 @@ def _on_llm_new_token(
...
         """Process the Retriever Run upon start."""
+        return None
 
     def _on_retriever_end(self, run: Run) -> Union[None, Coroutine[Any, Any, None]]:
         """Process the Retriever Run."""
+        return None
 
     def _on_retriever_error(self, run: Run) -> Union[None, Coroutine[Any, Any, None]]:
         """Process the Retriever Run upon error."""
+        return None
```
nfcampos, line 588: this doesn't make a ton of sense, its either an abstract method or it isn't

baskaryan: seems like separate issue from this pr (having nothing is equivalent to returning None)

libs/core/langchain_core/utils/mustache.py
cbornet: That's the other solution which was used before MappingProxyType was introduced.
I think that it's not as good to use Optional as from a type perspective, it allows to pass None for no good reason.
But if you prefer Optional, I don't mind changing this.

libs/core/langchain_core/runnables/utils.py
cbornet: It comes from https://docs.astral.sh/ruff/rules/unreliable-callable-check/
Is it a false positive ? Should the rule be ignored here ?

libs/core/langchain_core/tracers/core.py
cbornet: See previous comment: https://github.com/langchain-ai/langchain/pull/25520/files/dd1df352ffb2731807f10224b0919039782939d7#r1720771953
It's not an abstract method, but ruff thinks it should be if we don't add `return None`.
I can add `noqa` instead if you prefer.

efriis (approved):
thank you !!

libs/core/langchain_core/runnables/utils.py
hinthornw: Here this is explicitly just a guard so that we don't err when checking `inspect.isasyncgenfunction(func.__call__)` in the next line, so i think it's a false positive.

hinthornw: could alternatiely do `inspect.isasyncgenfunction(getattr(func, '__call__', None))` if we want the linter to be happy

Issue #10446, opened 2023-09-11
Summarize agent_scratchpad when it exceeds n tokens


mdziezyc:
### Feature request

Similarly to `memory=ConversationSummaryBufferMemory(llm=llm, max_token_limit=n)` passed in `initialize_agent`, there should be a possibility to pass `ConversationSummaryBufferMemory` like-object which would summarize the `intermediate_steps` in the agent if the `agent_scratchpad` created from the `intermediate_steps`  exceeds `n` tokens

### Motivation

Agents can run out of the context window when solving a complex problem with tools.

### Your contribution

I can't commit to anything for now.

rhamnett:
Did you solve this? I am looking to solve the same issue when the agent loops and continues to pull extra documents from the vector store. I think what needs to happen is the scratchpad almost needs to be run through a document compressor? Thoughts?

mdziezyc:
Hi! I have not developed a solution. For now, I just moved on to other things.

I saw that there is a `intermediate_steps` variable in `/langchain/agents/agent.py:1029` which is than changed to the text `agent_scratchpad`. IMHO the solution is either to compress `intermediate_steps` or the resulting `agent_scratchpad`. TBH I hoped for any input from dosu bot.

I'd suggest that anyone who is interested in this feature should "up vote" the issue - maybe we will be able to gain some momentum. :)

rhamnett:
I had the same thought, I will try and work on compressing the agent_scratchpad inside the loop and let you know if I have any success.

obujacz:
@dosu-bot thank you for answer. However, my understanding of the problem is a little bit different. I would like to summarize `agent_scratchpad` created from `intermediate_steps` within one prompt from the user, so that complex problem can be solved.
Your solution adds content of the `agent_scratchpad` to the memory of the entire conversation, this not a desired behavior. Could you adjust your reponse?

mdziezyc:
I have to be the one to ask dosu bot :) 

@dosu-bot thank you for answer. However, my understanding of the problem is a little bit different. I would like to summarize `agent_scratchpad` created from `intermediate_steps` within one prompt from the user, so that complex problem can be solved.
Your solution adds content of the `agent_scratchpad` to the memory of the entire conversation, this not a desired behavior. Could you adjust your reponse?

rhamnett:
At least we know where to start looking now, I wouldn't be using 'sumy' but an extra call to the existing given LLM. I haven't had time to progress this personally.

[...]

PR #25527, opened 2024-08-18
community: [improvement] An approach to (partially) solve the "request too large" issue on browser automation 

zysilm:
- **Description:** 

In some cases, the LLM agent tools respond with large texts (such as the "extract_text" tool of the PlayWright toolkit). This will often cause the OpenAI API to return error 429, for instance: "Error code: 429 Request too large for gpt-4 in organization org-... on tokens per min (TPM): Limit 10000, Requested 16670". Also, large requests will increase the generation cost as the number of tokens increases. 

A common approach to solve this is to make a summarization to reduce the tool response. However, in the tasks like browser automation, we often want to keep the original text, so that the LLM would have a chance to reference back to the appropriate element. 

Another approach is to split the tool response into chunks and use RAG to retrieve the relevant text chunks. However, this approach will require to access the original user input in the prompt. This will be tricky with the current implementation of langchain, as the prompt is consumed in the chain and can't be forwarded into the tools. 

In this pull request, I made some prototypical implementations to "inject" the prompt into the tools. The "injector" implemented in the libs/core/langchain_core/prompts/injector.py accepts a list of objects (where they need to be injected with additional information. In this case it would be the tools), takes Input from RunnableSerializable, passes this Input attributes of the objects in the list, and then forward the original Input as Output without any modification. 
Then, the tool can use the prompt, extract user input, and retrieve the relevant text chuck, instead of forwarding the entire text to the prompt. 

Unfortunately, I was not able to get PlayWright running with Jupyter Notebook. But I  did some example scripts in my personal repository: https://github.com/zysoong/langchain-example

Where the script "browser_without_retrieval_example.py" using the current implementation will often fail due to "Request too large" error from OpenAI. In comparison, the script "browser_with_retrieval_example.py" will often get better results. 
This is a prototypical implementation and since it added new classes and methods to the core package, tools, and builtin chains. it's likely to be rejected. But I would be glad to hear the idea of improvements from your side (Maybe the RAG approach can be done without adding anything in the framework, which I didn't know)
- **Issue:** #10446
- **Dependencies:** None


efriis:
Hey there! Introducing a core abstraction for this is not desired, and because this implements the feature as a core abstraction, I believe it should be closed per the [review process](https://python.langchain.com/v0.2/docs/contributing/review_process/)

Feel free to either
- resubmit an edit to the community integration that only adds this feature to a specific integration, or
- file an issue concisely describing why this abstraction is desired as a general feature instead of just a feature (without an abstraction) for an integration. Feel free to tag me!

PR #25542, opened 2024-08-19
groq[patch]: update model used for llama tests

ccurme:
`llama-3.1-8b-instant` often fails some of the tool calling standard tests. Here we update to `llama-3.1-70b-versatile`.


ccurme:
Tests pass: https://github.com/langchain-ai/langchain/actions/runs/10454696265/job/28947918329

PR #25548, opened 2024-08-19
together, standard-tests: specify tool_choice in standard tests

ccurme:
Here we allow standard tests to specify a value for `tool_choice` via a `tool_choice_value` property, which defaults to None.

Chat models [available in Together](https://docs.together.ai/docs/chat-models) have issues passing standard tool calling tests:
- llama 3.1 models currently [appear to rely on user-side parsing](https://docs.together.ai/docs/llama-3-function-calling) in Together;
- Mixtral-8x7B and Mistral-7B (currently tested) consistently do not call tools in some tests.

Specifying tool_choice also lets us remove an existing `xfail` and use a smaller model in Groq tests.

### DIFF
--- a/libs/partners/groq/tests/integration_tests/test_standard.py
+++ b/libs/partners/groq/tests/integration_tests/test_standard.py
@@ -1,6 +1,6 @@
 """Standard LangChain interface tests"""
 
-from typing import Type
+from typing import Optional, Type
 
 import pytest
 from langchain_core.language_models import BaseChatModel
@@ -32,6 +32,11 @@ class TestGroqLlama(BaseTestGroq):
             "temperature": 0,
             "rate_limiter": rate_limiter,
         }
+
+    @property
+    def tool_choice_value(self) -> Optional[str]:
+        """Value to use for tool choice when used in tests."""
+        return "any"
 
     @pytest.mark.xfail(
         reason=("Fails with 'Failed to call a function. Please adjust your prompt.'")
--- a/libs/partners/groq/tests/integration_tests/test_standard.py
+++ b/libs/partners/groq/tests/integration_tests/test_standard.py
@@ -28,10 +28,16 @@ class TestGroqLlama(BaseTestGroq):
     @property
     def chat_model_params(self) -> dict:
         return {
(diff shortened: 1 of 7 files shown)


libs/partners/groq/tests/integration_tests/test_standard.py
```diff
@@ -28,11 +28,22 @@ class TestGroqLlama(BaseTestGroq):
     @property
     def chat_model_params(self) -> dict:
         return {
-            "model": "llama-3.1-70b-versatile",
+            "model": "llama-3.1-8b-instant",
```
baskaryan, line 31: ooc why go smaller instead of bigger

libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
```diff
@@ -170,7 +170,14 @@ def test_stop_sequence(self, model: BaseChatModel) -> None:
     def test_tool_calling(self, model: BaseChatModel) -> None:
         if not self.has_tool_calling:
             pytest.skip("Test requires tool calling.")
-        model_with_tools = model.bind_tools([magic_function])
+        if self.tool_choice_value == "dict":
```
baskaryan, line 173: this format is openai-specific (eg anthropic doesn't support i dont think). i actually tihnk more models support just passing in tool name as a string to bind_tools, and think thats what we'd want to standardize on from a devx perspective anyways

libs/partners/groq/tests/integration_tests/test_standard.py
ccurme: We can do 70b (I changed it from 8b to 70b this morning). My thought is if they support the same features we should prefer smaller models for tests in spirit of testing functionality vs. benchmarking.

libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py
ccurme: Thanks, updated this to tool name.

baskaryan (approved):

PR #25584, opened 2024-08-20
chroma: fix typing

efriis:


efriis:
doing in #25586 instead

PR #25587, opened 2024-08-20
docs[patch]: Hide and deindex template pages from in-docs search

jacoblee93:
Adds a new element to the bottom of these pages that I'll configure our docs search crawler to skip:

```
<span data-lc-docs-search-deindexed="true"></span>
```


efriis:
doing #25717 instead

PR #25590, opened 2024-08-20
community: Fix Hybrid Search for non-Databricks managed embeddings

erikml-db:
Description: Send both the query and query_embedding to the Databricks index for hybrid search.

Issue: When using hybrid search with non-Databricks managed embedding we currently don't pass both the embedding and query_text to the index. Hybrid search requires both of these. This change fixes this issue for both `similarity_search` and `similarity_search_by_vector`.


efriis (approved):
any chance we could add a unit test for this?

erikml-db:
Added tests and fixed the ones that are broken

PR #25591, opened 2024-08-20
openai[patch]: add disabled_params

baskaryan:
- enable json schema mode in azure openai
- add ability to disable different params


libs/partners/openai/langchain_openai/chat_models/azure.py
```diff
@@ -529,6 +521,7 @@ class Joke(BaseModel):
     Distinct from the Azure deployment name, which is set by the Azure user.
     Used for tracing and token counting. Does NOT affect completion.
     """
+    disabled_params: List[str] = Field(default_factory=lambda: ["parallel_tool_calls"])
```
ccurme, line 524: As a class attribute this seems somewhat mysterious, as it only applies to `with_structured_output`. My first thought is to add a `bind_kwargs: Optional[dict]` kwarg to with_structured_output, e.g.,
```python
if bind_kwargs is None:
    bind_kwargs = {
        "tool_choice": tool_name,
        "parallel_tool_calls": False,
        "strict": strict,
    }
```
then Azure can override `with_structured_output` and pass this in to the parent class method.

Alternatively, can update the naming here / clarify in documentation.

ccurme (approved):

libs/partners/openai/langchain_openai/chat_models/azure.py
baskaryan: i think disabled_params is something we're gonna find useful beyond structured output if/when azure and openai apis diverge again

if it works well could generalize it to others integrations like bedrock (where apis are model-specific) and vertexai

efriis (approved):

baskaryan:
outdated given latest azure oai api versions, will open new pr

PR #25601, opened 2024-08-21
docs: how to use langsmith few shot

baskaryan:
Requires langsmith 0.1.101 release


hinthornw (approved):
> This enables searching over the dataset, and will make sure that anytime we update/add examples they are also indexed.

Nit: no comma before “ and”


Schemas tripped me up. I know it’s a how-to, but personally think still worth adding a line like:

“Our query `{“input”: string}` matches the dataset schema”

and maybe show read_dataset() and then show the input schema?

and/or a comment about "We can also index any langsmith dataset where we have provided an input schema"

jakerachleff:
@baskaryan left some comments on the vercel preview but otherwise looks good

baskaryan:
will land as is and fast follow

PR #25648, opened 2024-08-22
databricks: add vector search and embeddings

B-Step62:
### Summary

Add `DatabricksVectorSearch` and `DatabricksEmbeddings` classes to the `langchain-databricks` partner packages. Core functionality is unchanged, but the vector search class is largely refactored for readability and maintainability.

This PR does not add integration tests yet. This will be added once the Databricks test workspace is ready.

Tagging @efriis as POC

### Tracker
[✅] Create a package and imgrate ChatDatabricks
[✍️] Migrate DatabricksVectorSearch, DatabricksEmbeddings, and their docs
~[ ] Migrate UCFunctionToolkit and its doc~
[ ] Add provider document and update README.md
[ ] Add integration tests and set up secrets (after moved to an external package)
[ ] Add deprecation note to the community implementations.


libs/partners/databricks/langchain_databricks/utils.py
```diff
@@ -0,0 +1,101 @@
...
+            "required dependencies."
+        ) from e
+
+
+# Utility function for Maximal Marginal Relevance (MMR) reranking.
+# Copied from langchain_community/vectorstores/utils.py to avoid cross-dependency
+Matrix = Union[List[List[float]], List[np.ndarray], np.ndarray]
+
+
+def maximal_marginal_relevance(
```
B-Step62, line 30: @efriis This MMR utility is used in most of partner vector store implementations, but the code is duplicated to each package for avoiding dependency to the community package. What about moving this into the core package?
https://github.com/search?q=repo%3Alangchain-ai%2Flangchain%20maximal_marginal_relevance&type=code

libs/partners/databricks/langchain_databricks/vectorstores.py
```diff
@@ -0,0 +1,831 @@
...
+            )
+            retriever.invoke("thud")
+        .. code-block:: python
+            [Document(metadata={'id': '2'}, page_content='thud')]
+    """  # noqa: E501
+
+    def __init__(
+        self,
+        endpoint: str,
+        index_name: str,
```
B-Step62, line 201: Updated the constructor to take `endpoint` and `index_name` strings, instead of a `VectorSearchIndex` object ([previous code](https://github.com/langchain-ai/langchain/blob/e958f76160c6542072c7bd334fbb3c21d327e8bc/libs/community/langchain_community/vectorstores/databricks_vector_search.py#L119)). Two motivations for this change:
1. Convenience, users don't need to create client instance and query for the index.
2. Migration to Databricks SDK. The index object returned from Databricks SDK is different from `VectorSearchIndex` object. In the new signature, we can do the migration without asking any change in users' code.

harupy (approved):
LGTM!

libs/partners/databricks/langchain_databricks/embeddings.py
```diff
@@ -0,0 +1,91 @@
...
+                ...
+            ]
+
+    """
+
+    endpoint: str
+    """The endpoint to use."""
+    target_uri: str = "databricks"
+    """The parameters to use for queries."""
+    query_params: Dict[str, str] = {}
```
harupy, line 62: are there any non-string params?

B-Step62: Great catch, yes, there seems to be some non-string params. 
https://platform.openai.com/docs/api-reference/embeddings/create

baskaryan (commented):
we should deprecate the versions of embeddings and vectorstores in langchain-community

B-Step62:
@baskaryan Yes! We are plan to do that after releasing the first version of `langchain-databricks`, or should we do that here?

baskaryan:
> @baskaryan Yes! We are plan to do that after releasing the first version of `langchain-databricks`, or should we do that here?

we can do it here, will just need to make sure we release langchain-databricks next before we release langchain-community

B-Step62:
@baskaryan I think we should do that after the `langchain-databricks` release, then. There are a few todos before releasing the package other than this PR - e.g. move to a dedicated package. We're working with @efriis to set these up.

baskaryan:
also should make sure this is included in new vecstore if not already #25590

B-Step62:
> also should make sure this is included in new vecstore if not already https://github.com/langchain-ai/langchain/pull/25590

Great catch! Will update the new class to include this fix.

libs/partners/databricks/langchain_databricks/utils.py
efriis: #25689 merged!

efriis (approved):

efriis (approved):

PR #25652, opened 2024-08-22
[community] Support passing extra params for executing functions in UCFunctionToolkit

serena-ruan:

Support passing extra params when executing UC functions:
The params should be a dictionary with key EXECUTE_FUNCTION_ARG_NAME, the assumption is that the function itself doesn't use such variable name (starting and ending with double underscores), and if it does we raise Exception.
If invalid params passing to the execute_statement, we raise Exception as well.

Additional guidelines:

### DIFF
--- a/libs/community/langchain_community/tools/databricks/_execution.py
+++ b/libs/community/langchain_community/tools/databricks/_execution.py
@@ -150,7 +150,7 @@ def execute_function(
         ws.statement_execution.execute_statement
     ).parameters
     if not any(
-        p.kind == p.VAR_POSITIONAL or p.kind == p.VAR_KEYWORD
+        p.kind in (p.VAR_POSITIONAL, p.VAR_KEYWORD)
         for p in allowed_execute_statement_args.values()
     ):
         invalid_params = set()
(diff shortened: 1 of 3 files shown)


libs/community/langchain_community/tools/databricks/_execution.py
```diff
@@ -122,16 +130,49 @@ def execute_function(
...
+        and function.input_params.parameters
+        and any(
+            p.name == EXECUTE_FUNCTION_ARG_NAME
+            for p in function.input_params.parameters
+        )
+    ):
+        raise ValueError(
+            "Parameter name conflicts with the reserved argument name for executing "
+            f"functions: {EXECUTE_FUNCTION_ARG_NAME}. "
+            f"Please rename the parameter {EXECUTE_FUNCTION_ARG_NAME}."
```
harupy, line 144: Is renaming the parameter the only solution...?

serena-ruan: I think it's really rare for such situation, do we need to support other solutions?

harupy: No, but I just thought renaming a parameter can be challenging sometimes.

harupy: I see, the value `EXECUTE_FUNCTION_ARG_NAME` is `__executor_args__`, which is unlikely to confict.

serena-ruan:
Hi @baskaryan Do you know why the CI fails to install the dependency? I've added it in the extended_testing_deps.txt file.

baskaryan (approved):

harupy (approved):
LGTM!

PR #25660, opened 2024-08-22
community: add retry for session expired exception in neo4j

noahmay:
Description: The neo4j driver can raise a SessionExpired error, which is considered a retriable error. If a query fails with a SessionExpired error, this change retries every query once. This change will make the neo4j integration less flaky.
Twitter handle: noahmay_


jexp (approved):
LGTM

ccurme (approved):

PR #25663, opened 2024-08-22
community: add metadata filter to CassandraGraphVectorStore

epinzur:
- **Description:** 
  - Added metadata filtering support to `langchain_community.graph_vectorstores.cassandra.CassandraGraphVectorStore`
  - Also fixed type conversion issues highlighted by mypy.
- **Dependencies:** 
  - `ragstack-ai-knowledge-store 0.2.0` (released July 23, 2024)

### DIFF
--- a/libs/community/langchain_community/graph_vectorstores/cassandra.py
+++ b/libs/community/langchain_community/graph_vectorstores/cassandra.py
@@ -120,18 +120,31 @@ class CassandraGraphVectorStore(GraphVectorStore):
         return store
 
     def similarity_search(
-        self, query: str, k: int = 4, **kwargs: Any
+        self,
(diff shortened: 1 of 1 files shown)


ccurme (approved):
Thanks for this. Verified that `{}` is the correct default in `ragstack_knowledge_store` (see for example [here](https://github.com/datastax/ragstack-ai/blob/b66bef9c72f74e87f88cb2879bc579b07a3c5690/libs/knowledge-store/ragstack_knowledge_store/graph_store.py#L685)).

Lint appears to pass without the type casting so I removed them: https://github.com/langchain-ai/langchain/actions/runs/10513516808/job/29129296956?pr=25663

PR #25674, opened 2024-08-22
core[patch]: dict chat prompt template support

baskaryan:
- Support passing dicts as templates to chat prompt template
- Support making *any* attribute on a message a runtime variable
- Significantly simpler than trying to update our existing prompt template classes

```python
    template = ChatPromptTemplate(
        [
            {
                "role": "assistant",
                "content": [
                    {
                        "type": "text",
                        "text": "{text1}",
                    },
                    {"type": "image_url", "image_url": {"path": "{local_image_path}"}},
                ],
                "name": "{name1}",
                "tool_calls": [
                    {
                        "name": "{tool_name1}",
                        "args": {"arg1": "{tool_arg1}"},
                        "id": "1",
                        "type": "tool_call",
                    }
                ],
            },
            {
                "role": "tool",
                "content": "{tool_content2}",
                "tool_call_id": "1",
                "name": "{tool_name1}",
            },
        ]
    )

```

will likely close #25514 if we like this idea and update to use this logic


libs/core/langchain_core/prompts/message.py
```diff
@@ -0,0 +1,167 @@
...
+    get_template_variables,
+)
+from langchain_core.utils.image import image_to_data_url
+from langchain_core.utils.interactive_env import is_interactive_env
+
+if TYPE_CHECKING:
+    from langchain_core.prompts.chat import ChatPromptTemplate
+
+
+class BaseMessagePromptTemplate(Serializable, ABC):
```
baskaryan, line 21: no change, just moved

libs/core/langchain_core/prompts/message.py
```diff
@@ -0,0 +1,167 @@
...
+        Returns:
+            Combined prompt template.
+        """
+        from langchain_core.prompts.chat import ChatPromptTemplate
+
+        prompt = ChatPromptTemplate(messages=[self])  # type: ignore[call-arg]
+        return prompt + other
+
+
+class _DictMessagePromptTemplate(BaseMessagePromptTemplate):
```
baskaryan, line 100: new type

baskaryan:
TODO: 

- [ ]   make sure new prompt type can be opened in langsmith playground
- [ ]   make sure new prompt type can be saved in prompt hub

libs/core/langchain_core/prompts/chat.py
```diff
@@ -51,81 +54,6 @@
     from collections.abc import Sequence
 
 
-class BaseMessagePromptTemplate(Serializable, ABC):
```
baskaryan, line 54: no functional change just moved

ccurme:
Reducing the scope of this to avoid breaking changes:

```python
# Branch

template = ChatPromptTemplate(
    [
        {
            "role": "assistant",
            "content": [
                {
                    "type": "text",
                    "text": "{text1}",
                },
                {"type": "image_url", "image_url": {"url": "{url1}"}},
                
                # Adding the block below raises ValueError on master
                # Here it will get appended to `.prompt` as a dict template
                {
                    "type": "file",
                    "source_type": "base64",
                    "source": "{data1}",
                    "mime_type": "application/pdf"
                },
            ],
        }
    ]
)
```

Will now only apply inside content for a message, so templates specified via the usual `{"role": ..., "content": ...}` format will remain `(Role)MessagePromptTemplate`

ccurme (approved):

Issue #25555, opened 2024-08-19
Ollama Model support For tool calling


jivaniyash:
### URL

https://python.langchain.com/v0.2/docs/integrations/chat/ollama/

### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

DOC: <Issue related to /v0.2/docs/integrations/chat/ollama/>

Is it just one model [https://ollama.com/library/llama3-groq-tool-use] available for tool-calling for Ollama?

https://python.langchain.com/v0.2/docs/integrations/chat/ollama/#tool-calling

gbaian10:
This mainly depends on whether the model itself supports it. You can check the models card on Ollama to find which models support tools.

Related blog: [Tool Support](https://ollama.com/blog/tool-support)
Card: [Model with Tools](https://ollama.com/search?c=tools)

PR #25681, opened 2024-08-22
community, langchain: Ollama Model support for tool calling

zysilm:
**Description:** Some fixes for tool calls with Ollama: 
 - Fix Rest API call: ollama needs "tools" being given in request.tools. Previous was request.options.tools: https://github.com/ollama/ollama/blob/main/docs/api.md#chat-request-with-tools
 - Replace tool_call_id with empty string for local Ollama as it's not required
 - Set stream to be false in request body. Otherwise the tool_calls can't be responded in the response.message.tool_calls key (it will respond some natural language indicating that tool was called in this case)
 - Message handling: Ollama is not returning "AgentFinish" if job is done
 - Example script can be found in the comment of this ticket

**Issue:** the issue #25555 it fixes, if applicable
**Dependencies:** None


zysoong:
Example script
`
from typing import Sequence

from langchain.agents.output_parsers.ollama_tools import OllamaToolsAgentOutputParser
from langchain_community.agent_toolkits import PlayWrightBrowserToolkit
from langchain_community.tools.playwright.utils import (
    create_sync_playwright_browser,
)
from langchain.agents.format_scratchpad.openai_tools import (
    format_to_openai_tool_messages,
)
from langchain.agents import AgentExecutor
from langchain_core.language_models import BaseLanguageModel
from langchain_core.runnables import Runnable, RunnablePassthrough
from langchain_experimental.llms.ollama_functions import ChatOllama


def create_ollama_tools_agent(
        llm: BaseLanguageModel, tools: Sequence[BaseTool], prompt: ChatPromptTemplate
) -> Runnable:
    """Create an agent that uses Ollama tools.
    The prompts will be injected to the tools
    automatically.
    Check the documentation of 'create_openai_tools_agent'
    for detailed instructions.

    Args:
        llm: LLM to use as the agent.
        tools: Tools this agent has access to. Prompts will be injected as 'prompt'
            attribute automatically.
        prompt: The prompt to use. See Prompt section below for more on the expected
            input variables.
    """
    missing_vars = {"agent_scratchpad"}.difference(
        prompt.input_variables + list(prompt.partial_variables)
    )
    if missing_vars:
        raise ValueError(f"Prompt missing required variables: {missing_vars}")

    llm_with_tools = llm.bind(tools=[convert_to_openai_tool(tool) for tool in tools])

    agent = (
            RunnablePassthrough.assign(
                agent_scratchpad=lambda x: format_to_openai_tool_messages(
                    x["intermediate_steps"]
                )
            )
            | prompt
            | llm_with_tools
            | OllamaToolsAgentOutputParser()
    )
    return agent


if __name__ == '__main__':
        MessagesPlaceholder(variable_name="chat_history", optional=True),
        ("user",
         "Go to https://ffxiv.consolegameswiki.com/wiki/{quest} "
         "and find the previous quests recursively. "),
        MessagesPlaceholder(variable_name="agent_scratchpad"),
    ])

    sync_browser = create_sync_playwright_browser()
    toolkit = PlayWrightBrowserToolkit.from_browser(sync_browser=sync_browser)
    tools = toolkit.get_tools()
    llm = ChatOllama(model="llama3.1")

    agent = create_ollama_tools_agent(llm, tools, prompt)
    command = {
        "quest": "On_Rough_Seas"
    }
    agent_executor.invoke(command)
`

libs/community/langchain_community/chat_models/ollama.py
baskaryan, line 1: this implementation is deprecated in favor of the one in libs/partners/ollama (which contains the langchain-ollama package), can we make all changes there?

zysoong: My changes were already implemented by someone else in the libs/partners/ollama: 
https://github.com/langchain-ai/langchain/blob/master/libs/partners/ollama/langchain_ollama/chat_models.py#L398
https://github.com/langchain-ai/langchain/blob/master/libs/partners/ollama/langchain_ollama/chat_models.py#L514
https://github.com/langchain-ai/langchain/blob/master/libs/partners/ollama/langchain_ollama/chat_models.py#L518

The one in /partners/ollama is not using the Ollama rest API but the function call of the ollama client. But the results were similar. (similarly bad with the 8B model XD)

I didn't reverse my changes in the langchain_community/chat_models/ollama, as one could still get some results if he occasionally used this class. Maybe adding some deprecation decorator will be nice idea.

baskaryan: i believe there already is a deprecation decorator (line 89)

zysoong: @baskaryan Yeah but that one is decorated on the function level ("_format_message_as_text"), which is called by another deprecated function "_legacy_stream". The changes I made like in the function "_convert_messages_to_ollama_messages" are not marked as deprecated and will be called by "_stream" (not deprecated). But the function "_convert_messages_to_ollama_messages" didn't work as well as it couldn't sort out ToolMessage. (should also be kind of deprecated)

Can we make the whole class decorated as "deprecated" as it is no longer intended to be used?

baskaryan: ah i missed that, yes we should definitely deprecate the whole class! thanks for flagging

zysoong:
Added deprecation decorator on the community ollama and moved the ollama unit test to the partner package.

efriis:
don't think this is desired. the ollama partner package supports tool calling already, and we're moving towards langgraph agents as opposed to a provider specific agent output parser like shown here. Will close!

Will deprecate community ollama integrations in a separate pr - good catch

efriis:
#26733

PR #25644, opened 2024-08-21
core+partners/anthropic: Anthropic prompt caching

mrdrprofuroboros:
**Description:** Added support for Anthropic prompt caching, see https://github.com/langchain-ai/langchain/discussions/25625
**Issue:** the issue # it fixes, if applicable
**Dependencies:** bump anthropic>=0.34.0

* just found that it fails the test
will fix it and add usage example to the notebook, so far here:

```python
from langchain_anthropic import ChatAnthropic
from langchain_core.messages import HumanMessage, SystemMessage

model = ChatAnthropic(model="claude-3-5-sonnet-20240620", beta=True)
model.invoke([
    SystemMessage([{
        "type": "text",
        "text": "foo"*1000,
        "cache_control": {"type": "ephemeral"}
    }]),
    HumanMessage("hi!"),
])
```

```
AIMessage(content='Hello! How can I assist you today?', response_metadata={'id': 'msg_01XnYziv7oZtaRw23d45ivSi', 'model': 'claude-3-5-sonnet-20240620', 'stop_reason': 'end_turn', 'stop_sequence': None, 'usage': {'cache_creation_input_tokens': 1500, 'cache_read_input_tokens': 0, 'input_tokens': 9, 'output_tokens': 12}}, id='run-31a51a4f-bb0c-4b24-9c75-d7f7e5f99a89-0', usage_metadata={'input_tokens': 9, 'output_tokens': 12, 'total_tokens': 21, 'cache_creation_input_tokens': 1500, 'cache_read_input_tokens': 0})
```

PR #25684, opened 2024-08-22
rfc: anthropic cache usage

baskaryan:
Alternative to #25644

What do we want to do with anthropic cache token counts and UsageMetadata?
1. do nothing for now: UsageMetadata is meant to be standard across models and we dont know what other providers will do with caching. plus the info is there is AIMessage.response_metadata. main issue here is rn cached input tokens dont show up anywhere (neither in input_tokens nor total_tokens)
2. just add cached input tokens to total_tokens, so at least there's some way to figure out that those tokens exist
3. store cache token counts as their own keys on AIMessage.usage_metadata for anthropic outputs (current implementation in this pr). easy to implement and makes this info easily accessible to user, but will this lock us into to something we can't break in the future?
4. add cache token counts to existing "input_tokens" key (anthropic's data.usage.input_tokens does *not* take into account cached tokens so this isn't double counting). not useful for estimating pricing but useful for general token counting
5. add something like a "cost_adjusted_input_tokens" key that counts tokens in terms of costs (1.5 * cache_creation + 1 * input + 0.1 * cache_read). useful for estimating prices, but makes it a little harder to figure out when you've got cache hits/misses. is that important?

### DIFF
--- a/libs/partners/anthropic/langchain_anthropic/chat_models.py
+++ b/libs/partners/anthropic/langchain_anthropic/chat_models.py
@@ -757,9 +757,18 @@ class ChatAnthropic(BaseChatModel):
             "output_tokens": data.usage.output_tokens,
             "total_tokens": data.usage.input_tokens + data.usage.output_tokens,
         }
+        if hasattr(data.usage, "cache_creation_input_tokens"):
+            msg.usage_metadata["cache_creation_input_tokens"] = (
+                data.usage.cache_creation_input_tokens
+            )
+            msg.usage_metadata["total_tokens"] += data.usage.cache_creation_input_tokens
+        if hasattr(data.usage, "cache_read_input_tokens"):
+            msg.usage_metadata["cache_read_input_tokens"] = (
+                data.usage.cache_read_input_tokens
+            )
+            msg.usage_metadata["total_tokens"] += data.usage.cache_read_input_tokens
         return ChatResult(
-            generations=[ChatGeneration(message=msg)],
-            llm_output=llm_output,
+            generations=[ChatGeneration(message=msg)], llm_output=llm_output
         )
 
     def _generate(


efriis (approved):

mrdrprofuroboros:
One more thing I'm facing right now is PropmtTemplate - I can't find a way to use it with Anthropic cache 'cause it wipes out the 'cache_control' attribute from the text block. 
https://github.com/langchain-ai/langchain/blob/92abf62292e10a837f1e83cd1d9b4b4ab66dc94f/libs/core/langchain_core/prompts/chat.py#L521
Here it just extracts text and throws all the rest away

Are there any ideas / workarounds?

baskaryan:
> One more thing I'm facing right now is PropmtTemplate - I can't find a way to use it with Anthropic cache 'cause it wipes out the 'cache_control' attribute from the text block.
> 
> https://github.com/langchain-ai/langchain/blob/92abf62292e10a837f1e83cd1d9b4b4ab66dc94f/libs/core/langchain_core/prompts/chat.py#L521
> 
> 
> Here it just extracts text and throws all the rest away
> Are there any ideas / workarounds?

thinking of adding support for something like this, any thoughts? https://github.com/langchain-ai/langchain/pull/25674

mrdrprofuroboros:
lgtm
so far I just made a runnable lambda that injects cache_control after template compilation. this might be a bit cleaner for me since I substitute models and e.g. groq doesn't like that sort of dict messages and I have to strip them anyway
Does that make sense to control in templates? some param to either keep dicts or flat everything down

huynhphuchuy:
> lgtm so far I just made a runnable lambda that injects cache_control after template compilation. this might be a bit cleaner for me since I substitute models and e.g. groq doesn't like that sort of dict messages and I have to strip them anyway Does that make sense to control in templates? some param to either keep dicts or flat everything down

Hi @mrdrprofuroboros , could you share your approach with `PromptTemplate`? thanks :((

mrdrprofuroboros:
Oh, sorry, I missed your message
sure, here's what I came up with, rather hacky, but works for me
```python


def to_cached(content: str | List[Dict[str, Any]]) -> List[Dict[str, Any]]:
    if isinstance(content, str):
        return [{
            "type": "text",
            "text": content,
        }]
    if isinstance(content[-1], str):
        content[-1] = {
            "type": "text",
            "text": content[-1],
        }
    else:
        content[-1]["cache_control"] = {"type": "ephemeral"}
    return content


def add_cache_control(compiled_chat: ChatPromptValue) -> ChatPromptValue:
    """
    Anthropic supports maximum 4 blocks with cache_control so we'll set
    - 1 on tools
    - 1 on the first and last system block
    - 1 on the last even 5th message
    """
    system_messages = []
    other_messages = []
    
    for message in compiled_chat.messages:
        if isinstance(message, SystemMessage):
            if not system_messages:
                system_messages.append(SystemMessage(to_cached(message.content)))
            else:
                system_messages[0].content.append(message.content)
        else:
            other_messages.append(deepcopy(message))

    if system_messages:
        system_messages[0].content = to_cached(system_messages[0].content)

    messages = system_messages + other_messages
    last_cached = (len(messages) - 2) // 5 * 5 # never set on the last message since it's constantly changing
    messages[last_cached].content = to_cached(messages[last_cached].content)
    return ChatPromptValue(messages=messages)
    

### and then

if isinstance(model, ChatAnthropic):
    model = RunnableLambda(add_cache_control) | model
```

mrdrprofuroboros:
Hey @baskaryan, can we add this info to the `llm_output` which is meant to be unreliable provider-specific info?

baskaryan:
closing in favor of https://github.com/langchain-ai/langchain/pull/27087

baskaryan:
> Hey @baskaryan, can we merge this branch? We can explicitly warn that this data is not standardized and one could use it on their own risk of deprecation

will have a standardized format out in https://github.com/langchain-ai/langchain/pull/27087! should land and release later today

mrdrprofuroboros:
That’s great news, thank you!

baskaryan:
> That’s great news, thank you!

out in langchain-anthropic 0.2.2

baskaryan:
> > That’s great news, thank you!
> 
> out in langchain-anthropic 0.2.2

Sorry 0.2.3! Had to patch so that usage_metadata['input_tokens'] was a sum of all input tokens, including cache read and cache creation tokens

PR #25686, opened 2024-08-22
infra: xfail pydantic v2 arg to py function

baskaryan:
Issue to track: #25687

### DIFF
--- a/libs/core/tests/unit_tests/utils/test_function_calling.py
+++ b/libs/core/tests/unit_tests/utils/test_function_calling.py
@@ -296,6 +296,21 @@ def test_convert_to_openai_function(
     assert actual == runnable_expected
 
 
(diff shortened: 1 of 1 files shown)


libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -296,60 +296,59 @@ def test_convert_to_openai_function(
...
-        nested_v2_arg2: Literal["bar", "baz"] = FieldV2Maybe(
-            ..., description="one of 'bar', 'baz'"
-        )
-
-    def my_function(arg1: Nested, arg2: NestedV2) -> None:
+    def my_function(arg1: Nested) -> None:
         """dummy function"""
         pass
 
-        expected = {
```
baskaryan, line 316: this was indented so it was part of my_function and not actually tested, which is why it wasn't failing

PR #25707, opened 2024-08-23
community: remove unused verify_ssl kwarg from aiohttp request

efriis:
it's not a valid kwarg in aiohttp request


thunze:
Now `self.verify` isn't honored anymore and [SSL validation is enabled by default](https://docs.aiohttp.org/en/stable/client_reference.html) with no option for LangChain users to disable it for async requests. Maybe a better solution would've been to add `ssl=bool(self.verify)` as a keyword argument since `ssl` supersedes `verify_ssl`.

PR #25717, opened 2024-08-23
docs: remove templates

efriis:
- [x] check redirect works at template root
- [x] check redirect works within individual template page


efriis:
fyi @jacoblee93

PR #25727, opened 2024-08-24
community: fix model initialization bug for deepinfra

mo374z:
### Description
adds an init method to ChatDeepInfra to set the model_name attribute accordings to the argument
### Issue
currently, the model_name specified by the user during initialization of the ChatDeepInfra class is never set. Therefore, it always chooses the default model (meta-llama/Llama-2-70b-chat-hf, however probably since this is deprecated it always uses meta-llama/Llama-3-70b-Instruct). We stumbled across this issue and fixed it as proposed in this pull request. Feel free to change the fix according to your coding guidelines and style, this is just a proposal and we want to draw attention to this problem.
### Dependencies
no additional dependencies required

Feel free to contact me or @timo282 and @finitearth if you have any questions.


baskaryan (approved):

baskaryan:
thanks for flagging the issue @mo374z!

PR #25737, opened 2024-08-25
community[neo4j_vector]: make embedding dimension check optional

HollowMan6:
**Description:**

Starting from Neo4j 5.23 (22 August 2024), with vector-2.0 indexes, `vector.dimensions` is not required to be set, which will cause it the key not exist error in index config if it's not set.

Since the existence of vector.dimensions will only ensure additional checks, this commit turns embedding dimension check optional, and only do checks when it exists (not None).

https://neo4j.com/release-notes/database/neo4j-5/

**Twitter handle:** @HollowM186

### DIFF
--- a/libs/community/langchain_community/vectorstores/neo4j_vector.py
+++ b/libs/community/langchain_community/vectorstores/neo4j_vector.py
@@ -692,9 +692,10 @@ class Neo4jVector(VectorStore):
             self.node_label = index_information[0]["labelsOrTypes"][0]
             self.embedding_node_property = index_information[0]["properties"][0]
             self._index_type = index_information[0]["entityType"]
-            embedding_dimension = index_information[0]["options"]["indexConfig"][
-                "vector.dimensions"
-            ]
+            embedding_dimension = None
+            index_config = index_information[0]["options"]["indexConfig"]
+            if "vector.dimensions" in index_config:
+                embedding_dimension = index_config["vector.dimensions"]
 
             return embedding_dimension, index_information[0]["entityType"]
         except IndexError:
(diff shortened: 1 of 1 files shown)


baskaryan (commented):
cc @tomasonjo

tomasonjo:
Thanks for raising this. I need a couple of days to think how we want to
handle this and then I'll get back to you

V sre., 28. avg. 2024, 16:47 je oseba Bagatur ***@***.***>
napisala:

libs/community/langchain_community/vectorstores/neo4j_vector.py
```diff
@@ -1030,9 +1033,12 @@ def similarity_search_with_score_by_vector(
             base_index_query = parallel_query + (
                 f"MATCH (n:`{self.node_label}`) WHERE "
                 f"n.`{self.embedding_node_property}` IS NOT NULL AND "
-                f"size(n.`{self.embedding_node_property}`) = "
```
tomasonjo, line 1033: The embedding dimension comes from embedding model and not index definition, so this is not needed:

https://github.com/langchain-ai/langchain/blob/2ddf5b27f0e6f3c4d4a670f28a0e6f0860a6c575/libs/community/langchain_community/vectorstores/neo4j_vector.py#L576

tomasonjo (approved):

tomasonjo:
@HollowMan6 thanks for you contribution, we can merge it in as is, and then make any changes later if needed

libs/community/langchain_community/vectorstores/neo4j_vector.py
HollowMan6: Removed, thank you!

efriis:
last minute edits look ok to me - @tomasonjo if you want to take a look as well

tomasonjo:
Lgtm

V sob., 31. avg. 2024, 06:04 je oseba Erick Friis ***@***.***>
napisala:

ccurme (approved):

Issue #15168, opened 2023-12-26
Issue: Custom Mistral based LLM from API for RetrievalQA chain


hitsense:
### Issue you'd like to raise.

I would like to build RAG based on Mistral 7B model
The model is already hosted, and I provide llm_url in the custom LLM setup

I am able to make a request and get a response from the URL using the `llm._call` method, however something is wrong with the callbacks in `RetrievalQA.from_chain_type` method

It gives me below error
`'Mistral7B_LLM' object has no attribute 'callbacks'`

Am I missing anything in the below code

```
from pydantic import Extra
import requests
from typing import Any, List,Dict, Callable, Type, Mapping, Optional

from langchain.callbacks.manager import CallbackManagerForLLMRun
from langchain.llms.base import LLM, BaseLLM


class Mistral7B_LLM(LLM):

    def __init__(self):
        self.__post_init__()

    def __post_init__(self) -> None:

        def _import_mistral7B_llm() -> Any:
            from svcs.vector.src.controllers.llm.mistral7B_serving import Mistral7B_LLM

            return Mistral7B_LLM
        
        def __getattr__() -> Any:
            return Mistral7B_LLM()
        
        def get_type_to_cls_dict() -> Dict[str, Callable[[], Type[BaseLLM]]]:
            return {
                "Mistral7B_LLM": _import_mistral7B_llm,
            }
        
        __all__ = ["Mistral7B_LLM"]

    class Config:
        extra = Extra.forbid

    @property
    def _llm_type(self) -> str:
        return "Mistral7B_LLM"

    def _call(
        self,
        prompt: str,
        stop: Optional[List[str]] = None,
        run_manager: Optional[CallbackManagerForLLMRun] = None,
        max_new_tokens: Optional[int] = 156,
        temperature: Optional[float] = 0.7,
        **kwargs: Any,
    ) -> str:
        if stop is not None:
            raise ValueError("stop kwargs are not permitted.")

        payload = {
            "inputs": [prompt],
            "max_new_tokens": max_new_tokens,
            "temperature": temperature,
        }

        headers = {"Content-Type": "application/json"}


        llm_url = 'my url'

        response = requests.post(llm_url, json=payload, headers=headers, verify=False)
        response.raise_for_status()

        # print("API Response:", response.json())
        answer = response.json()["outputs"].split("[/INST]")[-1]
        return answer

    @property
    def _identifying_params(self) -> Mapping[str, Any]:
        """Get the identifying parameters."""
[...]

PR #25741, opened 2024-08-26
core: support Optional typing for callbacks in llms.py

anonstudy:
- **Description:** I noticed that the 'callbacks' parameter in llms.py was not consistently defined as optional, which could lead to type-checking errors. Although these functions will be deprecated in v1, it is still necessary to maintain consistent type definitions until then to ensure proper functionality and avoid errors.
  
  - **Issue:** the issue [#15168](https://github.com/langchain-ai/langchain/issues/15168)


ccurme:
`Callbacks` is already typed as Optional: https://github.com/langchain-ai/langchain/blob/745d1c2b8d2585d1ab76c5b231a51f15b39d990a/libs/core/langchain_core/callbacks/base.py#L1073

Issue #22135, opened 2024-05-24
Bug in Indexing Function Causes Inconsistent Document Deletion


ericvaillancourt:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Here is an example that demonstrates the problem:

If I change the `batch_size` in `api.py` to a value that is larger than the number of elements in my list, everything works fine. By default, the `batch_size` is set to 100, and only the first 100 elements are handled correctly.

```python
from langchain_community.document_loaders import TextLoader
from langchain.text_splitter import CharacterTextSplitter
from langchain_openai import OpenAIEmbeddings
from langchain_community.vectorstores import Chroma
from langchain_core.documents import Document
from langchain.indexes import SQLRecordManager, index

embeddings = OpenAIEmbeddings()

documents = []


for i in range(1, 201):
    page_content = f"data {i}"
    metadata = {"source": f"test.txt"}
    document = Document(page_content=page_content, metadata=metadata)
    documents.append(document)


collection_name = "test_index"

embedding = OpenAIEmbeddings()

vectorstore = Chroma(
    persist_directory="emb",
    embedding_function=embeddings
)
namespace = f"choma/{collection_name}"
record_manager = SQLRecordManager(
    namespace, db_url="sqlite:///record_manager_cache.sql"
)

record_manager.create_schema()

idx = index(
    documents,
    record_manager,
    vectorstore,
    cleanup="incremental",
    source_id_key="source",
)
# for the first run
# should be : {'num_added': 200, 'num_updated': 0, 'num_skipped': 0, 'num_deleted': 0}
# and that's what we get.
print(idx)
idx = index(
    documents,
    record_manager,
    vectorstore,
    cleanup="incremental",
    source_id_key="source",
)
# for the second run
# should be : {'num_added': 0, 'num_updated': 0, 'num_skipped': 200, 'num_deleted': 0}
# but we get : {'num_added': 100, 'num_updated': 0, 'num_skipped': 100, 'num_deleted': 100}
print(idx)
```

### Description

[...]

PR #25754, opened 2024-08-26
core[patch]: improve index/aindex api when batch_size<n_docs

federico-pisanu:
- **Description:** prevent index function to re-index entire source document even if nothing has changed.
- **Issue:** #22135

I worked on a solution to this issue that is a compromise between being cheap and being fast.
In the previous code, when batch_size is greater than the number of docs from a certain source almost the entire source is deleted (all documents from that source except for the documents in the first batch)
My solution deletes documents from vector store and record manager only if at least one document has changed for that source.

Hope this can help!


baskaryan (commented):
could we add unit tests for this?

federico-pisanu:
@baskaryan  yeah sure!
I checked the current tests and noticed that they where not properly testing the functionality. Patching the timestamps but not changing the patch between the "index" calls it was not properly simulating consequent calls. The test were passing but they were not testing the right thing.
Updating the tests was sufficient to check that the new code works and fixes the old behavior.

federico-pisanu:
Hi @eyurtsev can i do something to make it easier to merge?

eyurtsev:
Standby we're working on the 0.3 release so there was an effective code freeze for the past 2 weeks

eyurtsev:
Hi @federico-pisanu,

Thank you for the PR! In order to merge the changes, the code would need to do the following:

1. Apply similar changes to the `aindex` API
2. Revert modifications to existing unit tests (these modifications make it look like the PR is introducing a new bug!)
3. Add a new unit tests that cover the relevant scenario.

eyurtsev:
@federico-pisanu let me double check the existing unit tests to make sure they were wrong

eyurtsev:
@federico-pisanu OK confirmed -- no action is required from you at this time. I'll re-apply changes on aindex

federico-pisanu:
@eyurtsev Isn't it already done?

eyurtsev (approved):

eyurtsev:
@federico-pisanu yep didn't notice -- looks good added unit tests to cover the optimization -- i.e., we get different results based on which batch was mutated

PR #25764, opened 2024-08-27
Docs contri

Swastik-Swarup-Dash:
updated  Documentation Style Guide under 
>docs/docs/contributing/documentation/style_guide.mdx
for easier understanding


ccurme:
Thanks for this. Subjective opinion but I don't think this is a material change, and have a slight preference for the existing phrasing.

PR #25777, opened 2024-08-27
community: avoid double templating in `langchain_prompty`

### LINKED ISSUE #25703: prompty: Superfluous f-string templating causes prompt templaing to fail
### Checked other resources

### Example Code

Use a barebone `.prompty` file such as:

```yaml
---
name: IssuePrompt
description: A prompt used to infer the intent of a user's query
model:
  api: chat
---

user:
Tell me about {templating issues}.
````

And the following script in the same folder:

```python
from langchain_prompty import create_chat_prompt
from pathlib import Path

root_path = Path(__file__).parent
prompt_path = root_path / "issue.prompty"

templated_prompt = create_chat_prompt(prompt_path)
templated_prompt.invoke(inputs={})
```

```log
  File "issue.py", line 8, in <module>
    templated_prompt.invoke({})
  File ".venv\Lib\site-packages\langchain_core\runnables\base.py", line 4475, in invoke
    return self._call_with_config(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\langchain_core\runnables\base.py", line 1785, in _call_with_config
    context.run(
  File ".venv\Lib\site-packages\langchain_core\runnables\config.py", line 397, in call_func_with_variable_args
    return func(input, **kwargs)  # type: ignore[call-arg]
           ^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\langchain_core\runnables\base.py", line 4341, in _invoke
    output = output.invoke(
             ^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\langchain_core\prompts\base.py", line 179, in invoke
    return self._call_with_config(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\langchain_core\runnables\base.py", line 1785, in _call_with_config
    context.run(
  File ".venv\Lib\site-packages\langchain_core\runnables\config.py", line 397, in call_func_with_variable_args
    return func(input, **kwargs)  # type: ignore[call-arg]
           ^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\langchain_core\prompts\base.py", line 153, in _format_prompt_with_error_handling
    _inner_input = self._validate_input(inner_input)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv\Lib\site-packages\langchain_core\prompts\base.py", line 145, in _validate_input     
    raise KeyError(
KeyError: "Input to ChatPromptTemplate is missing variables {'templating issues'}.  Expected: ['templating issues'] Received: []"
```

### Description

Hello everyone! Thank you for your work.

I am playing around with `langchain_prompty` and noticed that messages are rendered two times, once by Prompty using Mustache, and once by langchain by treating messages as f-strings.

So you have to process all inputs to your prompts to escape brakcets for them to be ignored by the f-string templating (so `{{ }}`instead of `{ }`, which could cause issues with Mustache if the content is a key in your inputs). This is a bit hacky and brittle. Would there be a way to avoid this sedonc templating? Below is a detailled explanation of how the templating(s) happens.

 `langchain_prompty` uses Mustache templates, as discussed in this [PR introducing the package](https://github.com/langchain-ai/langchain/pull/21346#discussion_r1596035508).

This templating happens like this:
- [Prompt creation using the prompty file](https://github.com/langchain-ai/langchain/blob/99f9a664a5381d4bcd0d38397b604cd562d4452c/libs/partners/prompty/langchain_prompty/langchain.py#L16) (call to `langchain_prompty.create_chat_prompt()`)
	- Which leads to the [processing of the prompty file](https://github.com/langchain-ai/langchain/blob/99f9a664a5381d4bcd0d38397b604cd562d4452c/libs/partners/prompty/langchain_prompty/utils.py#L142)
		- Which triggers the [Mustache templating](https://github.com/langchain-ai/langchain/blob/99f9a664a5381d4bcd0d38397b604cd562d4452c/libs/partners/prompty/langchain_prompty/renderers.py#L15)

After this initial templating, [the messages are processed by Langchain](https://github.com/langchain-ai/langchain/blob/99f9a664a5381d4bcd0d38397b604cd562d4452c/libs/partners/prompty/langchain_prompty/langchain.py#L28), which by default treats them as f-strings templates ([see source code](https://github.com/langchain-ai/langchain/blob/30f1bf24ac5f62b17dd71bc341fd3dfadd03893f/libs/core/langchain_core/prompts/chat.py#L1147)). Hence the issue.

Tagging you @wayliums as I saw you worked on the prompty integration into langchain.

Thanks!

### System Info

Happens on all systems and configurations.

matthieucx:
## Description

In `langchain_prompty`, messages are templated by Prompty. However, a call to `ChatPromptTemplate` was initiating a second templating. We now convert parsed messages to `Message` objects before calling `ChatPromptTemplate`, signifying clearly that they are already templated.

We also revert #25739 , which applied to this second templating, which we now avoid, and did not fix the original issue.

## Issue

Closes #25703

### DIFF
--- a/libs/partners/prompty/langchain_prompty/langchain.py
+++ b/libs/partners/prompty/langchain_prompty/langchain.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Literal
+from typing import Any, Dict
 
 from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
 from langchain_core.runnables import Runnable, RunnableLambda
--- a/libs/partners/prompty/langchain_prompty/langchain.py
+++ b/libs/partners/prompty/langchain_prompty/langchain.py
@@ -3,6 +3,7 @@ from typing import Any, Dict, Literal
 from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
 from langchain_core.runnables import Runnable, RunnableLambda
 
+from .parsers import RoleMap
 from .utils import load, prepare
 
 
@@ -15,17 +16,17 @@ def create_chat_prompt(
     def runnable_chat_lambda(inputs: Dict[str, Any]) -> ChatPromptTemplate:
         p = load(path)
         parsed = prepare(p, inputs)
+        # Parsed messages have been templated
+        # Convert to Message objects to avoid templating attempts in ChatPromptTemplate
         lc_messages = []
         for message in parsed:
-            lc_messages.append((message["role"], message["content"]))
+            message_class = RoleMap.get_message_class(message["role"])
+            lc_messages.append(message_class(content=message["content"]))
 
         lc_messages.append(
             MessagesPlaceholder(
                 variable_name=input_name_agent_scratchpad, optional=True
             )  # type: ignore[arg-type]
-        )
-        lc_p = ChatPromptTemplate.from_messages(
-            lc_messages, template_format=template_format
         )
         lc_p = ChatPromptTemplate.from_messages(lc_messages)
         lc_p = lc_p.partial(**p.inputs)
--- a/libs/partners/prompty/langchain_prompty/langchain.py
+++ b/libs/partners/prompty/langchain_prompty/langchain.py
@@ -9,7 +9,6 @@ from .utils import load, prepare
 def create_chat_prompt(
     path: str,
     input_name_agent_scratchpad: str = "agent_scratchpad",
-    template_format: Literal["f-string", "mustache", "jinja2"] = "f-string",
 ) -> Runnable[Dict[str, Any], ChatPromptTemplate]:
     """Create a chat prompt from a Langchain schema."""
 
@@ -28,6 +27,7 @@ def create_chat_prompt(
         lc_p = ChatPromptTemplate.from_messages(
             lc_messages, template_format=template_format
         )
+        lc_p = ChatPromptTemplate.from_messages(lc_messages)
         lc_p = lc_p.partial(**p.inputs)
         return lc_p
 
--- a/libs/partners/prompty/langchain_prompty/parsers.py
+++ b/libs/partners/prompty/langchain_prompty/parsers.py
@@ -5,6 +5,28 @@ from typing import List, Union
 from pydantic import BaseModel
 
 from .core import Invoker, Prompty, SimpleModel
+from langchain_core.messages import (
+    AIMessage,
+    BaseMessage,
+    FunctionMessage,
+    HumanMessage,
+    SystemMessage,
+)
+
+class RoleMap:
+    _ROLE_MAP: dict[str, type[BaseMessage]] = {
+        "system": SystemMessage,
+        "user": HumanMessage,
+        "human": HumanMessage,
+        "assistant": AIMessage,
+        "ai": AIMessage,
+        "function": FunctionMessage,
+    }
+    ROLES = _ROLE_MAP.keys()
+
+    @classmethod
+    def get_message_class(cls, role: str) -> type[BaseMessage]:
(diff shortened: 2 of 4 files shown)


libs/partners/prompty/langchain_prompty/parsers.py
```diff
@@ -1,18 +1,41 @@
...
+    FunctionMessage,
+    HumanMessage,
+    SystemMessage,
+)
 from pydantic import BaseModel
 
 from .core import Invoker, Prompty, SimpleModel
 
 
+class RoleMap:
```
ccurme, line 17: Does this cover the full scope of roles? What about `"tool"`?

libs/partners/prompty/langchain_prompty/langchain.py
```diff
@@ -1,33 +1,34 @@
...
+            message_class = RoleMap.get_message_class(message["role"])
+            lc_messages.append(message_class(content=message["content"]))
 
         lc_messages.append(
             MessagesPlaceholder(
                 variable_name=input_name_agent_scratchpad, optional=True
             )  # type: ignore[arg-type]
         )
-        lc_p = ChatPromptTemplate.from_messages(
-            lc_messages, template_format=template_format
```
ccurme, line 29: Can you clarify why specifying `template_format` with a default of `"f-string"` is breaking? It looks like that is the default in `ChatPromptTemplate.from_messages`:
https://github.com/langchain-ai/langchain/blob/d6c4803ab068e3fb20fb84ed27f96a6ae33a02d4/libs/core/langchain_core/prompts/chat.py#L1147-L1151

Or is it just unrelated?

libs/partners/prompty/langchain_prompty/parsers.py
matthieucx: Yes, it's on par with the latest versions of Prompty, they define the [same set of roles](https://github.com/microsoft/prompty/blob/074abe7f35968e9ed10a54eeda98d203c00de3ae/runtime/prompty/prompty/parsers.py#L11)! Tools are defined in the [YAML header](https://github.com/microsoft/prompty/blob/074abe7f35968e9ed10a54eeda98d203c00de3ae/web/docs/prompty-file-spec/page.mdx#parameters-for-the-model-attribute) of the `.prompty` files, [here's an example](https://github.com/microsoft/prompty/blob/074abe7f35968e9ed10a54eeda98d203c00de3ae/runtime/prompty/tests/prompts/funcfile.prompty#L11)

libs/partners/prompty/langchain_prompty/langchain.py
matthieucx: You're right that's it's unrelated.

Here the issue is that before initializing the `ChatPromptTemplate`, Prompty has already parsed and templated the prompt using the user input.

When invoking `ChatPromptTemplate`, it tries to template the messages and interprets text inside `{ }` (or `{{ }}`) as missing variables. The fix proposed in this PR avoids this templating by converting to `Messages` objects, which are already templated.

Since we skip this second templating entirely, it does not make sense to expose one of its parameters to the user, as it won't have any effect. You can select the template engine used by Prompty directly in the `.prompty`s file YAML header.

ccurme (approved):

matthieucx:
Thanks for merging and for the time you spent on this!

ccurme:
> Thanks for merging and for the time you spent on this!

Thank you for the fix! Released in langchain-prompty 0.0.3, feel free to take a look.

PR #25796, opened 2024-08-28
[Evaluation] New Constrained COT evaluator

GMartin-dev:
This PR adds a new evaluator implemented with a new technique derived from COT.
I have been using for our use cases to reduce token length generation / cost and improve test speed.
Beyond the obvious upside of cost / speed reduction I think is important that Langchain exposes diverse prompting patterns by example.

Original Paper:
[[Concise Thoughts: Impact of Output Length on LLM Reasoning and Cost] ](https://arxiv.org/pdf/2407.19825)

>  constraining the reasoning of LLaMA2-70b to 100 words
improves the accuracy from 36.01% (CoT) to
41.07% (CCoT) on the GSM8K dataset, while
reducing the average output length by 28 words

image

image


ccurme:
Thanks for this. `prompt` is a parameter in `ContextQAEvalChain`. I don't think it makes sense to implement a new class in the codebase that inherits `ContextQAEvalChain` and adds two words to its default prompt (separately, `ContextQAEvalChain` inherits LLMChain, which is deprecated, so we should be wary of further increasing the surface area of LLMChain).

I think documentation-- e.g., in the API reference for `ContextQAEvalChain` or in a cookbook-- showing how to pass in a custom CCOT prompt would make sense. Happy to review this if you tag me.

PR #25832, opened 2024-08-29
core[minor]: Add msg content formatting util

baskaryan:
Fixes #25813 

Direct usage
```python
from langchain_core.messages import format_messages_as, AIMessage, HumanMessage, ToolMessage, SystemMessage

messages = [
    SystemMessage([{"type": "text", "text": "foo"}]),
    {"role": "user", "content": [{"type": "text", "text": "whats in this"}, {"type": "image_url", "image_url": {"url": "data:image/png;base64,'/9j/4AAQSk'"}}]},
    AIMessage("", tool_calls=[{"name": "analyze", "args": {"baz": "buz"}, "id": "1", "type": "tool_call"}]),
    ToolMessage("foobar", tool_call_id="1", name="bar"), 
    {"role": "assistant", "content": "thats nice"},
]
oai_strings = format_messages_as(messages, format="openai", text="string")
# -> [
#     SystemMessage(content='foo'),
#     HumanMessage(content=[{'type': 'text', 'text': 'whats in this'}, {'type': 'image_url', 'image_url': {'url': "data:image/png;base64,'/9j/4AAQSk'"}}]),
#     AIMessage(content='', tool_calls=[{'name': 'analyze', 'args': {'baz': 'buz'}, 'id': '1', 'type': 'tool_call'}]),
#     ToolMessage(content='foobar', name='bar', tool_call_id='1'),
#     AIMessage(content='thats nice')
# ]

anthropic_blocks = format_messages_as(messages, format="anthropic", text="block")
# -> [
#     SystemMessage(content=[{'type': 'text', 'text': 'foo'}]),
#     HumanMessage(content=[{'type': 'text', 'text': 'whats in this'}, {'type': 'image', 'source': {'type': 'base64', 'media_type': 'image/png', 'data': "'/9j/4AAQSk'"}}]),
#     AIMessage(content=[{'type': 'tool_use', 'input': {'baz': 'buz'}, 'id': '1', 'name': 'analyze'}], tool_calls=[{'name': 'analyze', 'args': {'baz': 'buz'}, 'id': '1', 'type': 'tool_call'}]),
#     HumanMessage(content=[{'type': 'tool_result', 'content': 'foobar', 'tool_use_id': '1', 'is_error': False}]),
#     AIMessage(content=[{'type': 'text', 'text': 'thats nice'}])
# ]
```

Chaining
```python
  from langchain_core.messages import format_messages
  formatter = format_messages(format="langchain-openai", text_format="string")
  llm = init_chat_model() | formatter
  llm.invoke(
      [{"role": "user", "content": "how are you"}],
      config={"model": "gpt-4o"},
  )
  # -> AIMessage(["I am good..."], ...)
  llm.invoke(
      [{"role": "user", "content": "whats your name"}],
      config={"model": "claude-3-5-sonnet-20240620"}
  )
  # -> AIMessage(["My name is...], ...)
```


libs/core/langchain_core/messages/utils.py
```diff
@@ -845,6 +909,610 @@ def list_token_counter(messages: Sequence[BaseMessage]) -> int:
...
+) -> Union[BaseMessage, List[BaseMessage]]:
+    """Convert message contents into a standard format.
+
+    .. versionadded:: 0.2.36
+
+    Args:
+        messages: Message-like object or iterable of objects whose contents are already
+            in OpenAI, Anthropic, Bedrock Converse, or VertexAI formats.
+        format: Format to convert message contents to.
+        text: How to format text contents. If ``text='string'``  then any string
```
vbarda, line 927: could we do `content_format` or something instead of `text`?

vbarda: and the above could be "provider" or something

baskaryan: content_format feels a little too broad given it only relates to text blocks — text_format?

mschoenb97IL:
Looks excellent, thank you!

mschoenb97IL:
Looks good! But does this support message chunk types? (i.e. AIMessageChunk, ToolMessageChunk)

baskaryan:
> Looks good! But does this support message chunk types? (i.e. AIMessageChunk, ToolMessageChunk)

as currently implemented it can format a chunk one at a time but it can't transform a stream on the fly. meaning if you used in a chain and you tried to stream from that chain, you would just get a single accumulated chunk out instead of token-level chunks

will think about if/how we can support transforming a stream

mschoenb97IL:
> > Looks good! But does this support message chunk types? (i.e. AIMessageChunk, ToolMessageChunk)
> 
> as currently implemented it can format a chunk one at a time but it can't transform a stream on the fly. meaning if you used in a chain and you tried to stream from that chain, you would just get a single accumulated chunk out instead of token-level chunks
> 
> will think about if/how we can support transforming a stream

So if I have a downstream application that consumes message chunks from a graph, can I apply this function to uniformize chunks across different providers?

libs/core/langchain_core/messages/utils.py
vbarda: sure!

jacoblee93:
What if you're going from more expressive to less expressive (e.g. Anthropic blocks -> OpenAI strings)?

If we are dropping expressiveness in that case, how about `coerce_messages_to` intead?

libs/core/langchain_core/messages/utils.py
```diff
@@ -845,6 +862,775 @@ def list_token_counter(messages: Sequence[BaseMessage]) -> int:
...
+                    yield msg
+
+            return RunnableGenerator(partial(transform, **kwargs), name=func.__name__)
+
+    wrapped.__doc__ = func.__doc__
+    return wrapped
+
+
+@_runnable_generator
+def format_messages_as(
```
efriis, line 937:
```suggestion
def format_messages(
```
? 🌝

libs/core/langchain_core/messages/utils.py
```diff
@@ -845,6 +862,775 @@ def list_token_counter(messages: Sequence[BaseMessage]) -> int:
...
+    messages: Union[MessageLikeRepresentation, Sequence[MessageLikeRepresentation]],
+    *,
+    format: Literal["openai", "anthropic"],
+    text_format: Literal["string", "block"],
```
efriis, line 941:
```suggestion
    format: Literal["openai", "anthropic", "openai_text", "anthropic_text"],
```
might be confusing to have `text_format` option if we ever introduce something else that doesn't offer one or the other

baskaryan:
> What if you're going from more expressive to less expressive (e.g. Anthropic blocks -> OpenAI strings)?
> 
> If we are dropping expressiveness in that case, how about `coerce_messages_to` intead?

only converts to strings if its all text blocks

libs/core/langchain_core/messages/utils.py
efriis: don't we also have a `convert_messages_to_dict` somewhere that does the same thing? Would that kind of naming make more sense?

Might make it clearer this is only meant to handle conversion to dicts (incase anyone tries to add conversions to custom sdks at some point)

efriis (approved):
light preferences commented. Also looks good as is

libs/core/langchain_core/messages/utils.py
baskaryan: something that doesn't offer string or block options?

libs/core/langchain_core/messages/utils.py
baskaryan: watcha mean meant to handle conversion to dict?

libs/core/langchain_core/messages/utils.py
efriis: yep - for example if new_model_provider_c only supports blocks and not string content in their api for some reason. Or anthropic introduces a block_2 format. Keeping it 1-dimensional is easier to maintain in that case.

libs/core/langchain_core/messages/utils.py
efriis: it formats the contents as openai/anthropic dicts, not <insert some model provider sdk> basemodel/protobuf/etc types for theirs.

mschoenb97IL:
@baskaryan Are there pending changes here? Would love to be able to use this soon. Thanks again!

mschoenb97IL:
@baskaryan Why was this closed?

Issue #24089, opened 2024-07-10
core: Fix @tool typing


baskaryan:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Type hints for `@tool` decorator are incorrect, which causes downstream type-checking errors. For example:
```python
from langchain_core.tools import tool, StructuredTool

@tool
def multiply(a: int, b: int) -> int:
    """multiply two ints"""
    return a * b

tool_: StructuredTool = multiply
```

leads to the following mypy errors:
```pycon
error: Incompatible types in assignment (expression has type "Callable[..., Any]", variable has type "StructuredTool")  [assignment]
```

eyurtsev:
This can just be BaseTool

PR #25856, opened 2024-08-29
core[fix] @tool typing

samjove:
Description: Fixes '@tool' typing addressing incompatibility in assignment
Issue: #24089


libs/core/langchain_core/tools/convert.py
```diff
@@ -10,14 +10,14 @@
...
+    *args: Union[str, Callable[..., Any], Runnable],
     return_direct: bool = False,
-    args_schema: Optional[Type] = None,
+    args_schema: Optional[Type[BaseModel]] = None,
     infer_schema: bool = True,
     response_format: Literal["content", "content_and_artifact"] = "content",
     parse_docstring: bool = False,
     error_on_invalid_docstring: bool = True,
-) -> Callable:
+) -> StructuredTool:
```
baskaryan, line 20: this should be Union[Tool, Callable] i believe
```python

@tool(return_direct=True)  # this outputs a Callable that creates a StructuredTool
def foo(...)

@tool(infer_schema=False)   # this outputs a Callable that creates a Tool
def foo(...)
```

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -101,6 +101,17 @@ async def _arun(self, arg1: int, arg2: bool, arg3: Optional[dict] = None) -> str
         raise NotImplementedError
 
 
+def test_tool_typing() -> None:
```
baskaryan, line 104: could we add a test case for each tool overload variant?

eyurtsev:
Closing this PR as it's about 4 months old, but I also think that it's OK for this to just be BaseTool.

PR #25857, opened 2024-08-29
community: attach database result id to Document object when using PGVector

kiqaps:
**Description**: retrieving Documents with `PGVector` its ID comes as `None` even if it has and ID on database (it will always have one id). I've changed the `_results_to_docs_and_scores` method to attach the database id into Documents

### DIFF
--- a/libs/community/langchain_community/vectorstores/pgvector.py
+++ b/libs/community/langchain_community/vectorstores/pgvector.py
@@ -638,6 +638,7 @@ class PGVector(VectorStore):
         docs = [
             (
                 Document(
+                    id=result.EmbeddingStore.id if result.EmbeddingStore.id else None,
                     page_content=result.EmbeddingStore.document,
                     metadata=result.EmbeddingStore.cmetadata,
                 ),


libs/community/langchain_community/vectorstores/pgvector.py
```diff
@@ -638,6 +638,7 @@ def _results_to_docs_and_scores(self, results: Any) -> List[Tuple[Document, floa
         docs = [
             (
                 Document(
+                    id=result.EmbeddingStore.id if result.EmbeddingStore.id else None,
```
ccurme, line 641: How do we know `result.EmbeddingStore` always has an `id` attribute?

For what it's worth I don't see that attribute here:
https://github.com/langchain-ai/langchain/blob/09c2d8faca42f09a924599c9d75cc38e05ba1ad4/libs/community/langchain_community/vectorstores/pgvector.py#L187

efriis:
this is a deprecated community integration in favor of the one in langchain-postgres! Feel free to resubmit there

https://github.com/langchain-ai/langchain-postgres

kiqaps:
Oh, I see... and just for information, its already fixed there: https://github.com/langchain-ai/langchain-postgres/commit/88831295acfb7bae3184c124ba82aaf2000dd3a9

So we just need to wait for a release.

Anyway, ty

PR #25862, opened 2024-08-29
langchain_openai: Add "strict" parameter to OpenAIFunctionsAgent

map9959:
- **Description:** OpenAI recently introduced a "strict" parameter for [structured outputs in their API](https://openai.com/index/introducing-structured-outputs-in-the-api/). An optional `strict` parameter has been added to `create_openai_functions_agent()` and `create_openai_tools_agent()` so developers can use this feature in those agents.

### DIFF
--- a/libs/langchain/langchain/agents/openai_functions_agent/base.py
+++ b/libs/langchain/langchain/agents/openai_functions_agent/base.py
@@ -278,7 +278,10 @@ class OpenAIFunctionsAgent(BaseSingleActionAgent):
 
 
 def create_openai_functions_agent(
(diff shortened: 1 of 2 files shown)


ccurme (approved):
Thanks for this. When I run with the functions agent I get:
> BadRequestError: Error code: 400 - {'error': {'message': "Invalid function definition for 'get_word_length': the 'strict' parameter is only permitted in 'tools'.", 'type': 'invalid_request_error', 'param': 'functions[0].strict', 'code': 'invalid_function_parameters'}}

As far as I can tell `strict` is not supported with functions, so I removed it. Curious if you're seeing the same.

PR #25876, opened 2024-08-30
community: O365Toolkit - send_event - fixed timezone error

SirNotReallySmokeAlot:
**Description**: Fixed formatting start and end time
**Issue**: The old formatting resulted everytime in an timezone error
**Dependencies**: /
**Twitter handle**: /


efriis (commented):
adding a fallback to parse the old way incase anyone is passing the timezone in the declared format `HHMM` instead of iso format `HH:MM`

efriis (approved):

Issue #25879, opened 2024-08-30
Change dimensions from None to the default number of dimensions used by the text-embedding-ada-002 model.


ValentinKovalev:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
from langchain_openai import OpenAIEmbeddings

embed = OpenAIEmbeddings()


print(embed.model)
print(embed.dimensions)
```

### Error Message and Stack Trace (if applicable)
```
text-embedding-ada-002
None
```
### Description

When using the default embedding model, the **embed.dimensions** attribute should not be **None**, but instead should default to **1536**, which is the number of dimensions for the default **text-embedding-ada-002** model.

### System Info

pip freeze | grep langchain:

langchain==0.2.14
langchain-anthropic==0.1.23
langchain-community==0.2.12
langchain-core==0.2.32
langchain-openai==0.1.20
langchain-text-splitters==0.2.2
langchainhub==0.1.20

platform: mac
python version: 3.11.9

PR #25880, opened 2024-08-30
Change default dimensions number in OpenAIEmbeddings class

ValentinKovalev:

- [ ] **PR title**: "partners: default dimension number in OpenAIEmbeddings class"

    - **Description:** Change default number of dims in OpenAIEmbeddings class
    - **Issue:** #25879
    - **Dependencies:** none
    - **Twitter handle:** @ValentinKovalev

### DIFF
--- a/libs/partners/openai/langchain_openai/embeddings/base.py
+++ b/libs/partners/openai/langchain_openai/embeddings/base.py
@@ -177,9 +177,9 @@ class OpenAIEmbeddings(BaseModel, Embeddings):
     client: Any = Field(default=None, exclude=True)  #: :meta private:
     async_client: Any = Field(default=None, exclude=True)  #: :meta private:
     model: str = "text-embedding-ada-002"
-    dimensions: Optional[int] = None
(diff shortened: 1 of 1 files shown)


efriis:
not desired as the default! This is for the embedding models that support this parameter, so defaulting to none is good (which defaults to max for those models)

If you need this in your code, you can pass dimensions=x manually :)

ValentinKovalev:
Could you please clarify the reasoning behind using the default model without specifying the dimensions parameter? I understand that hardcoding dimensions (like **1536** for the **text-embedding-ada-002** model) might not be ideal. However, I'm concerned that leaving **embed.dimensions** as **None** could lead to confusion or errors. It might be more intuitive for the default model to automatically set the correct dimensions.

efriis:
if you have a test case that's failing let me know. Using null/none just gives the max dimensions provided by the model https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-dimensions

PR #25892, opened 2024-08-30
docs: update chatbot.ipynb

wangxinalex:
- Description: Add the message history in the last example of "Managing Conversation History" chapter so that the the example returns a successful response.
- Issue: None
- Dependencies: None


ccurme:
Thanks for this. `with_message_history` is stateful and incorporates the previous chat history. This example is demonstrating the `trimmer`, and prefixes the code in that cell with `"so the model no longer knows about it:"`. So as I understand it we are deliberately demonstrating how messages are trimmed from the history.

PR #25898, opened 2024-08-30
community[feat]: Add OpenSearch Chat Message History

leedotpang:
**Description:** 
In building out a langchain powered OpenAI RAG tool at Ionic, we determined that utilizing AWS OpenSearch best fit our needs for our embeddings store. Since the service can also function as a document store without much additional effort, we also tested utilizing it for chat message history!

**Twitter handle:** @leedotpang


leedotpang:
I'll get these failures sorted out asap!

leedotpang:
@efriis did you have thoughts on this? 😄

PR #25906, opened 2024-08-31
core[feat]: Allow search_kwargs to be passed directly when invoking retriever

teenaxta:
## Description
Previously, `search_kwargs` was only supported as a class attribute. This commit extends the functionality to allow passing `search_kwargs` directly when invoking the retriever.

### Example:
```python
docs = retriever.invoke(
    input="Example input question", 
    search_kwargs={"k": 7}
)


hwchase17:
tackling more holistically in https://github.com/langchain-ai/langchain/pull/25934

Issue #25912, opened 2024-09-01
ChatLLM's are caches entire message object as string instead of storing the content


Kirushikesh:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_openai import ChatOpenAI
from langchain.globals import set_llm_cache
from langchain_community.cache import SQLiteCache
from langchain_core.messages import HumanMessage

model = ChatOpenAI(model="gpt-4o-mini")
set_llm_cache(SQLiteCache(database_path=".langchain.db"))

model.invoke([HumanMessage(content="Hi! I'm Bob")])
# Cached content is '[{"lc": 1, "type": "constructor", "id": ["langchain", "schema", "messages", "HumanMessage"], "kwargs": {"content": "Hi! I\'m Bob", "type": "human"}}]'
```

### Description

While using the Langchain library for caching, I've observed two issues:
a) The [documentation](https://python.langchain.com/v0.2/docs/integrations/llm_caching/#sqlite-cache) primarily demonstrates examples with text completion-based LLMs for the most common caching methods (InMemory and SQLite), rather than ChatLLMs, which Langchain now recommends to use in general.

b) When using ChatLLMs with caching, the cached content is stored as a string representation of the message object, instead of the actual content within the messages. This appears to be due to the [dumps()](https://github.com/langchain-ai/langchain/blob/60054db1c41571cd2aebc06c389658e90600a55b/libs/core/langchain_core/language_models/chat_models.py#L809C26-L809C31) function, which converts messages into a JSON string.

To improve this, it would be beneficial if users could provide a pre-processing function, similar to the approach used in the [GPTCache](https://gptcache.readthedocs.io/en/latest/configure_it.html#pre-process-function) library. This function would convert a user message into a string. By default, for ChatLLMs, it could extract the content of the last user message. Users could also implement custom processing functions, making it easier to debug or insert specific information into the cache.
I'd appreciate your thoughts on this suggestion.

### System Info

"pip freeze | grep langchain"
```
langchain==0.2.0
langchain-anthropic==0.1.15
langchain-community==0.0.25
langchain-core==0.1.52
langchain-ibm==0.1.7
[...]

PR #25926, opened 2024-09-01
Feat: Cache text only option

keenborder786:
### Description

This pull request addresses issue #25912, which suggests that caching only the text content of messages would be more efficient than caching the entire serialized message. 

**Summary of Changes:**

- Introduces a new attribute, `cache_text_only`, to control caching behavior.
- When `cache_text_only` is set to `True`, the cache will store only the `text` content of the message, rather than the entire serialized message.

**Rationale:**

Storing only the text content of messages can lead to significant improvements in cache efficiency and performance. By focusing on just the text, we reduce the cache size and potentially decrease retrieval times.

**Implementation Details:**

- Added the `cache_text_only` attribute to the relevant configuration.
- Updated cache handling logic to check the value of `cache_text_only` and store only the message `text` if the attribute is `True`.
- Ensured backward compatibility by defaulting `cache_text_only` to `False`, thus preserving the existing behavior.


Undertone0809:
Some benchmarking may be required to quantify the actual performance gains.

Undertone0809 (commented):
1. Performance impact: Some benchmarking may be required to quantify the actual performance gains.
2. You may want to consider some boundary cases, such as how the message content will behave if it is empty or very long.
3. Cache consistency: You need to ensure that cache consistency is not affected when switching cache_text_only Settings.
4. Documentation update: Need to add docs.

I personally do not recommend this usage, adding complexity to the API usage. Given that LangChain is a general-purpose framework, it may be more important to provide flexibility. But at the same time, we have to weigh the costs of increased complexity.


Suggest:
- Consider implementing a more general cache preprocessing hook that allows users to customize how cached content is generated. This satisfies both users who want to cache only text and users who need more control.
- At the same time, several preset caching strategies can be provided, such as "text only", "full object", etc., to meet common requirements.

keenborder786:
Okay

Get Outlook for Android<https://aka.ms/AAb9ysg>

efriis:
leaving as needs support and can return if folks have some good motivating examples.

In general, I think if you want custom caching behavior like this, it doesn't make sense to use the main LC caching abstraction. Keeping the cache contents in sync with "which caching strategy you're using" will create a lot more confusing errors than usefulness it provides imo

keenborder786:
@efriis keeping it open then.

PR #25929, opened 2024-09-02
Add Databricks integration to docs

B-Step62:
Updating the gateway pages in the documentation to name the `langchain-databricks` integration.

PR #25941, opened 2024-09-02
Fix databricks doc

B-Step62:
https://github.com/langchain-ai/langchain/pull/25929 broke the layout because of missing `:::` for the caution clause.


B-Step62:
@baskaryan Would you mind taking a look at the Vercel failure? This PR just solves a tiny syntax error in the doc.

PR #25959, opened 2024-09-03
community: Fix links in GraphVectorStore pydoc

cbornet:


libs/community/langchain_community/graph_vectorstores/__init__.py
```diff
@@ -15,23 +18,21 @@
 
 Link extractors can be used to extract links from documents.
 
-Example:
-
-.. code-block:: python
+Example::
```
cbornet, line 21: Python is the default code-block language. So the syntax can be simplified.

docs/api_reference/scripts/custom_formatter.py
```diff
@@ -17,15 +17,16 @@ def process_toc_h3_elements(html_content: str) -> str:
 
     # Process each element
     for element in toc_h3_elements:
-        element = element.a.code.span
-        # Get the text content of the element
-        content = element.get_text()
+        if element.a.code:
```
cbornet, line 20: Pyoc with multiple levels of headings create `toc-h3` elements (even if there's no `.. contents` directive)

cbornet:
Rebased.
Can this be reviewed 🙏 ?
Only easy to review docstring.

efriis (commented):
I don't think the api reference formatter change does anything different aside from increasing the cases where the job could fail. Thoughts on reverting that and just doing the docstring change?

cbornet:
I reverted the changes on the ref formatter.
Can you take a look again ?

PR #25966, opened 2024-09-03
community: [indexing] [feature] support for mssql backend in indexing

MacanPN:
# Description
Indexing requires a database access to store the current state of vector store. Currently indexing supports 2 database backends: `sqlite` and `postgresql`. This PR adds support for `mssql (SQL Server)`.

In order to meet requirements of `mssql` I had to introduce length limits on strings in `UpsertionRecord`. This is necessary because `mssql` does not support indexing / private keys on columns with VARCHAR type with undefined max length. I've opted to use very liberal limits to minimize chance that the length limits would break anything. The limits introduced are:
- uuid [80] (by default uuids are 36 chars long)
- key [80] (same as uuid)
- namespace [120] (this is used to identify the vector store / index in the database. Each database can contain states of multiple vector stores.)
- group_id [2048] (this is generally set to the document name or source url. Urls can be fairly long hence the higher limit)

Since `mssql` in `sqlalchemy` does not implement `insert.on_conflict_do_update()`, we're implementing equivalent functionality using `MERGE INTO` statement.

 **Twitter handle:** @martintriska1

@ccurme If I can be so bold, could you please also take a look at this PR? Please let me know what you think! Thank you!


eyurtsev:
mssql in this PR: https://github.com/langchain-ai/langchain/pull/25966
mysql in this PR: https://github.com/langchain-ai/langchain/pull/25815/files

---

Both PRs are blocked by:

1. Lack of integration tests
2. Updates to common schema that would affect any existing users. Is the update to the schema backwards compatible for existing users running postgres?


To unblock:

1. Need to introduce standard unit tests for SQL record manager
2. Need to add an integration test to langchain that verifies this with the other backends

Hanson-G2G:
May I ask how to do the integration testing?

eyurtsev:
Hi @Hanson-G2G and @MacanPN,

I am changing this pr to needs support. Before doing extra work, I think we should figure out whether there are other users interested in the implementation.

If so, the way that we would need to proceed:

1. We'd create a separate implementation in langchain_community for the integration. Not in langchain. The langchain implementation will only support sqlite and postgres going forward (and we'll probably change its name when there are most implementations). 
2. We would need to create a standard test suite in the standard tests (https://github.com/langchain-ai/langchain/blob/master/libs/standard-tests/langchain_standard_tests/integration_tests/)
You can take a look at base_store as an example of a simple test suite.
The idea would be to make a set of test that work just against the interface and aren't aware of any of the internal details of the record manager otherwise.
3. Use the standard test in langchain-community inside of `integration_tests`

It's not a small amount of work, so before doing any of this I think it makes sense to check that enough interest from the community. If there isn't, I think it makes sense for you to maintain your own implementation.

MacanPN:
@eyurtsev Thanks for giving  this a thought. Please suggest the future steps. Should we start an issue asking people to comment whether they'd be interested in indexer supporting other SQL dialects? 
I have already spent some time implementing the integration tests (feel free to take a look at new commits in this PR). I agree that connection to mssql database should be tested separately (in standard tests) rather then already in indexer code, however I'll have to take a better look at how exactly could everything come together.

eyurtsev:
@MacanPN Could you pitch the implementation as an idea and ask for up-votes on this PR to gauge interest from the community?

https://github.com/langchain-ai/langchain/discussions/categories/ideas

PR #25975, opened 2024-09-03
community: Change ChatOpenAI import in OctoAI chat model

sebascherny:
OctoAI model was importing OpenAI from a langchain_community object that doesn't have `bind_tool` defined so the model wasn't allowing tool choosing.


efriis:
this pattern for importing unfortunately isn't available for community integrations because of how python imports/subclasses work

if it's a pure descendant of langchain-openai, would recommend telling users to directly use ChatOpenAI with a custom base url

sebastian-cherny-toptal:
Mm okay I see, thanks for clarifying @efriis . But for example, in this case, where we want to set different properties like `is_lc_serializable` different than the original ChatOpenAI, how should we do it?
I think we could create our own open-source library but I thought that the purpose of the `langchain_community/lang_models` folder was avoiding that.

PR #25986, opened 2024-09-03
core[major]: Upgrade langchain-core to pydantic 2

eyurtsev:
This PR upgrades core to pydantic 2.

It involves a combination of manual changes together with automated code mods using gritql.

Changes and known issues:

1. Current models override __repr__ to be consistent with pydantic 1 (this will be removed in a follow up PR)
Related: https://github.com/langchain-ai/langchain/pull/25986/files#diff-e5bd296179b7a72fcd4ea5cfa28b145beaf787da057e6d122aa76ee0bb8132c9R74
2. Issue with decorator for BaseChatModel (https://github.com/langchain-ai/langchain/pull/25986/files#diff-932bf3b314b268754ef640a5b8f52da96f9024fb81dd388dcd166b5713ecdf66R202) -- cc @baskaryan 
3. `name` attribute in Base Runnable does not have a default -- was raising a pydantic warning due to override. We need to see if there's a way to fix to avoid making a breaking change for folks with custom runnables. (https://github.com/langchain-ai/langchain/pull/25986/files#diff-836773d27f8565f4dd45e9d6cf828920f89991a880c098b7511e0d3bb78a8a0dR238)
4. Likely can remove hard-coded RunnableBranch name (https://github.com/langchain-ai/langchain/pull/25986/files#diff-72894b94f70b1bfc908eb4d53f5ff90bb33bf8a4240a5e34cae48ddc62ac313aR147)
5. `model_*` namespace is reserved in pydantic. We'll need to specify `protected_namespaces`
6. create_model does not have a cached path yet
7. get_input_schema() in many places has been updated to be explicit about whether parameters are required or optional
8. injected tool args aren't picked up properly (losing type annotation)

For posterity the following gritql migrations were used:

```
engine marzano(0.1)
language python

or {
    `from $IMPORT import $...` where {
        $IMPORT <: contains `pydantic_v1`,
        $IMPORT => `pydantic`
    },
    `$X.update_forward_refs` => `$X.model_rebuild`,
  // This pattern still needs fixing as it fails (populate_by_name vs.
  // allow_populate_by_name)
  class_definition($name, $body) as $C where {
      $name <: `Config`,
      $body <: block($statements),
      $t = "",
      $statements <: some bubble($t) assignment(left=$x, right=$y) as $A where {    
        or {
            $x <: `allow_population_by_field_name` where {
                $t += `populate_by_name=$y,`
            },
            $t += `$x=$y,`
        }
      },
      $C => `model_config = ConfigDict($t)`,
      add_import(source="pydantic", name="ConfigDict")
  }
}

```

```
engine marzano(0.1)
language python

`@root_validator(pre=True)` as $decorator where {
    $decorator <: before function_definition($body, $return_type),
    $decorator => `@model_validator(mode="before")\n@classmethod`,
    add_import(source="pydantic", name="model_validator"),
    $return_type => `Any`
}
```

```
engine marzano(0.1)
language python

`@root_validator(pre=False, skip_on_failure=True)` as $decorator where {
    $decorator <: before function_definition($body, $parameters, $return_type) where {
        $body <: contains bubble or {
            `values["$Q"]` => `self.$Q`,
            `values.get("$Q")` => `(self.$Q or None)`,
            `values.get($Q, $...)` as $V where {
                $Q <: contains `"$QName"`,
                $V => `self.$QName`,
            },
            `return $Q` => `return self`
        }
    },
    $decorator => `@model_validator(mode="after")`,
    // Silly work around a bug in grit
    // Adding Self to pydantic and then will replace it with one from typing
    add_import(source="pydantic", name="model_validator"),
    $parameters => `self`,
    $return_type => `Self`
}

```

```
grit apply --language python '`Self` where { add_import(source="typing_extensions", name="Self")}'
```

### DIFF
--- a/libs/core/langchain_core/language_models/base.py
+++ b/libs/core/langchain_core/language_models/base.py
@@ -18,6 +18,7 @@ from typing import (
     Union,
 )
 
+from pydantic import BaseModel, ConfigDict, Field, validator
 from typing_extensions import TypeAlias, TypedDict
 
 from langchain_core._api import deprecated
@@ -28,7 +29,6 @@ from langchain_core.messages import (
     get_buffer_string,
 )
 from langchain_core.prompt_values import PromptValue
-from langchain_core.pydantic_v1 import BaseModel, Field, validator
 from langchain_core.runnables import Runnable, RunnableSerializable
 from langchain_core.utils import get_pydantic_field_names
 
@@ -113,7 +113,11 @@ class BaseLanguageModel(
     
     Caching is not currently supported for streaming methods of models.
     """
-    verbose: bool = Field(default_factory=_get_verbosity)
+    # Repr = False is consistent with pydantic 1 if verbose = False
+    # We can relax this for pydantic 2?
+    # TODO(Team): decide what to do here.
+    # Modified just to get unit tests to pass.
+    verbose: bool = Field(default_factory=_get_verbosity, exclude=True, repr=False)
     """Whether to print out response text."""
     callbacks: Callbacks = Field(default=None, exclude=True)
     """Callbacks to add to the run trace."""
@@ -125,6 +129,10 @@ class BaseLanguageModel(
         default=None, exclude=True
     )
     """Optional encoder to use for counting tokens."""
+
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True,
+    )
 
     @validator("verbose", pre=True, always=True, allow_reuse=True)
     def set_verbose(cls, verbose: Optional[bool]) -> bool:
--- a/libs/core/langchain_core/language_models/base.py
+++ b/libs/core/langchain_core/language_models/base.py
@@ -115,7 +115,7 @@ class BaseLanguageModel(
     """
     # Repr = False is consistent with pydantic 1 if verbose = False
     # We can relax this for pydantic 2?
-    # TODO(Team): decide what to do here.
+    # TODO(0.3): Resolve repr for verbose
     # Modified just to get unit tests to pass.
     verbose: bool = Field(default_factory=_get_verbosity, exclude=True, repr=False)
     """Whether to print out response text."""
--- a/libs/core/langchain_core/load/serializable.py
+++ b/libs/core/langchain_core/load/serializable.py
@@ -10,7 +10,7 @@ from typing import (
     cast,
 )
 
-from pydantic import BaseModel, ConfigDict  # pydantic: ignore
+from pydantic import BaseModel, ConfigDict  
 from typing_extensions import NotRequired
 
 from langchain_core.utils.pydantic import v1_repr
--- a/libs/core/langchain_core/load/serializable.py
+++ b/libs/core/langchain_core/load/serializable.py
@@ -10,9 +10,10 @@ from typing import (
     cast,
 )
 
+from pydantic import BaseModel, ConfigDict  # pydantic: ignore
 from typing_extensions import NotRequired
 
-from langchain_core.pydantic_v1 import BaseModel
+from langchain_core.utils.pydantic import v1_repr
 
 
 class BaseSerialized(TypedDict):
@@ -80,7 +81,7 @@ def try_neq_default(value: Any, key: str, model: BaseModel) -> bool:
         Exception: If the key is not in the model.
     """
     try:
-        return model.__fields__[key].get_default() != value
+        return model.model_fields[key].get_default() != value
     except Exception:
         return True
 
@@ -161,16 +162,25 @@ class Serializable(BaseModel, ABC):
         For example, for the class `langchain.llms.openai.OpenAI`, the id is
         ["langchain", "llms", "openai", "OpenAI"].
         """
-        return [*cls.get_lc_namespace(), cls.__name__]
+        # Pydantic generics change the class name. So we need to do the following
+        if (
+            "origin" in cls.__pydantic_generic_metadata__
+            and cls.__pydantic_generic_metadata__["origin"] is not None
+        ):
+            original_name = cls.__pydantic_generic_metadata__["origin"].__name__
+        else:
+            original_name = cls.__name__
+        return [*cls.get_lc_namespace(), original_name]
 
-    class Config:
-        extra = "ignore"
+    model_config = ConfigDict(
+        extra="ignore",
+    )
 
     def __repr_args__(self) -> Any:
         return [
             (k, v)
             for k, v in super().__repr_args__()
-            if (k not in self.__fields__ or try_neq_default(v, k, self))
+            if (k not in self.model_fields or try_neq_default(v, k, self))
         ]
 
     def to_json(self) -> Union[SerializedConstructor, SerializedNotImplemented]:
@@ -184,12 +194,15 @@ class Serializable(BaseModel, ABC):
 
         secrets = dict()
         # Get latest values for kwargs if there is an attribute with same name
-        lc_kwargs = {
-            k: getattr(self, k, v)
-            for k, v in self
-            if not (self.__exclude_fields__ or {}).get(k, False)  # type: ignore
-            and _is_field_useful(self, k, v)
-        }
+        lc_kwargs = {}
+        for k, v in self:
+            if not _is_field_useful(self, k, v):
+                continue
+            # Do nothing if the field is excluded
+            if k in self.model_fields and self.model_fields[k].exclude:
+                continue
+
+            lc_kwargs[k] = getattr(self, k, v)
 
         # Merge the lc_secrets and lc_attributes from every class in the MRO
         for cls in [None, *self.__class__.mro()]:
@@ -221,8 +234,10 @@ class Serializable(BaseModel, ABC):
             # that are not present in the fields.
             for key in list(secrets):
                 value = secrets[key]
-                if key in this.__fields__:
-                    secrets[this.__fields__[key].alias] = value
+                if key in this.model_fields:
+                    alias = this.model_fields[key].alias
+                    if alias is not None:
+                        secrets[alias] = value
             lc_kwargs.update(this.lc_attributes)
 
         # include all secrets, even if not specified in kwargs
@@ -244,6 +259,10 @@ class Serializable(BaseModel, ABC):
     def to_json_not_implemented(self) -> SerializedNotImplemented:
         return to_json_not_implemented(self)
 
+    def __repr__(self) -> str:
+        # TODO(0.3): Remove this override after confirming unit tests!
+        return v1_repr(self)
+
 
 def _is_field_useful(inst: Serializable, key: str, value: Any) -> bool:
     """Check if a field is useful as a constructor argument.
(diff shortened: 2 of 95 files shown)


libs/core/langchain_core/language_models/base.py
```diff
@@ -113,7 +113,11 @@ class BaseLanguageModel(
     
     Caching is not currently supported for streaming methods of models.
     """
-    verbose: bool = Field(default_factory=_get_verbosity)
+    # Repr = False is consistent with pydantic 1 if verbose = False
+    # We can relax this for pydantic 2?
+    # TODO(Team): decide what to do here.
```
baskaryan, line 118: update todo since we know we want to undo this

libs/core/langchain_core/language_models/base.py
```diff
@@ -126,6 +130,10 @@ class BaseLanguageModel(
     )
     """Optional encoder to use for counting tokens."""
 
+    model_config = ConfigDict(
```
baskaryan, line 133: is this new?

eyurtsev: Yes, it's required when using pydantic 2 since cache is an attribute on the chat model and the cache is not a base model

libs/core/langchain_core/language_models/base.py
eyurtsev, line 118:
```suggestion
    # TODO(0.3): Resolve repr for verbose
```

eyurtsev: If I'm not mistaken the issue was that this ends up affecting caching behavior with verbose=False and verbose=None being cached differently -- i'll double check, but i updated a TODO(0.3) for us to resolve

libs/core/langchain_core/tools/simple.py
```diff
@@ -155,3 +165,6 @@ def from_function(
             args_schema=args_schema,
             **kwargs,
         )
+
+
+Tool.model_rebuild()
```
baskaryan, line 170: could we add comment about what this does

eyurtsev: I'll add in all places with a code mod later on -- this is the equvalent of update_forward_refs

libs/core/langchain_core/messages/tool.py
```diff
@@ -70,6 +71,11 @@ class ToolMessage(BaseMessage):
     .. versionadded:: 0.2.24
     """
 
+    additional_kwargs: dict = Field(default_factory=dict, repr=False)
+    """Currently inherited from BaseMessage, but not used."""
+    response_metadata: dict = Field(default_factory=dict, repr=False)
```
baskaryan, line 76: i think these could technically be used

libs/core/langchain_core/output_parsers/openai_functions.py
```diff
@@ -267,11 +268,17 @@ def parse_result(self, result: List[Generation], *, partial: bool = False) -> An
         """
         _result = super().parse_result(result)
         if self.args_only:
-            pydantic_args = self.pydantic_schema.parse_raw(_result)  # type: ignore
+            if hasattr(self.pydantic_schema, "model_validate_json"):
+                pydantic_args = self.pydantic_schema.model_validate_json(_result)
+            else:
+                pydantic_args = self.pydantic_schema.parse_raw(_result)  # type: ignore
```
baskaryan, line 274: idea is we keep supporting pydantic v1 args?

libs/core/langchain_core/output_parsers/openai_tools.py
```diff
@@ -252,7 +254,7 @@ def parse_result(self, result: List[Generation], *, partial: bool = False) -> An
 class PydanticToolsParser(JsonOutputToolsParser):
     """Parse tools from OpenAI response."""
 
-    tools: List[TypeBaseModel]
+    tools: Annotated[List[TypeBaseModel], SkipValidation()]
```
baskaryan, line 257: are we getting rid of TypeBaseModel?

libs/core/langchain_core/outputs/llm_result.py
```diff
@@ -16,7 +18,9 @@ class LLMResult(BaseModel):
     wants to return.
     """
 
-    generations: List[List[Generation]]
+    generations: List[
+        List[Union[Generation, ChatGeneration, GenerationChunk, ChatGenerationChunk]]
```
baskaryan, line 22: ooc why's Generation not good enough anymore, do things get coerced to Generation?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -922,7 +935,7 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
 
     """  # noqa: E501
 
-    messages: List[MessageLike]
+    messages: Annotated[List[MessageLike], SkipValidation]
```
baskaryan, line 938: used SkipValidation() in other places, dunno if that matters

libs/core/langchain_core/runnables/base.py
```diff
@@ -236,25 +235,56 @@ def buggy_double(y: int) -> int:
...
-                return name + "_" + suffix.lower()
+                return name_ + "_" + suffix.lower()
         else:
-            return name
+            return name_
 
     @property
     def InputType(self) -> Type[Input]:
         """The type of input this Runnable accepts specified as a type annotation."""
+        # First loop through bases -- this will help generic
```
baskaryan, line 278: not sure what comment means

libs/core/langchain_core/runnables/base.py
```diff
@@ -236,25 +235,56 @@ def buggy_double(y: int) -> int:
...
+        # any pydantic models.
+        for base in self.__class__.mro():
+            if hasattr(base, "__pydantic_generic_metadata__"):
+                metadata = base.__pydantic_generic_metadata__
+                if "args" in metadata and len(metadata["args"]) == 2:
+                    return metadata["args"][0]
+
+        # then loop through __orig_bases__ -- this will Runnables that do not inherit
```
baskaryan, line 286: this will find/catch/... Runnable that...

libs/core/langchain_core/messages/tool.py
eyurtsev: OK -- we'll address in a follow up PR and update unit tests -- it affects a bunch of snapshots

libs/core/langchain_core/outputs/llm_result.py
eyurtsev: pydantic 2 is stricter with respect to union discrimination

libs/core/langchain_core/prompts/chat.py
eyurtsev: oh good catch! I wonder why this works

eyurtsev, line 938:
```suggestion
    messages: Annotated[List[MessageLike], SkipValidation()]
```

libs/core/langchain_core/output_parsers/openai_tools.py
eyurtsev: We could try to support .v1 models for a bit of time to avoid more breaking changes for users, but can also drop now

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -54,6 +54,7 @@
 )
 from langchain_core.utils.function_calling import convert_to_openai_function
 from langchain_core.utils.pydantic import PYDANTIC_MAJOR_VERSION, _create_subset_model
+from langchain_core.utils.pydantic import TypeBaseModel as TypeBaseModel
```
baskaryan, line 57: why needed

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -874,15 +875,13 @@ async def _arun(self) -> str:
 
 def test_optional_subset_model_rewrite() -> None:
     class MyModel(BaseModel):
-        a: Optional[str]
+        a: Optional[str] = None
```
baskaryan, line 878: ooc why needed

libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -450,17 +451,17 @@ def test_multiple_tool_calls() -> None:
...
             "id": messages[3].tool_call_id,
             "type": "function",
-            "function": {"name": "FakeCall", "arguments": '{"data": "ToolCall2"}'},
+            "function": {"name": "FakeCall", "arguments": '{"data":"ToolCall2"}'},
         },
         {
             "id": messages[4].tool_call_id,
             "type": "function",
-            "function": {"name": "FakeCall", "arguments": '{"data": "ToolCall3"}'},
+            "function": {"name": "FakeCall", "arguments": '{"data":"ToolCall3"}'},
```
baskaryan, line 464: is this still valid json

libs/core/tests/unit_tests/utils/test_pydantic.py
```diff
@@ -92,12 +94,12 @@ def validator(cls, v: Dict[str, Any]) -> Dict[str, Any]:
 def test_is_basemodel_subclass() -> None:
     """Test pydantic."""
     if PYDANTIC_MAJOR_VERSION == 1:
```
baskaryan, line 96: do need need to test pydantic v1 anymore?

libs/core/langchain_core/load/serializable.py
```diff
@@ -161,16 +162,25 @@ def lc_id(cls) -> List[str]:
...
-        return [*cls.get_lc_namespace(), cls.__name__]
-
-    class Config:
-        extra = "ignore"
+        # Pydantic generics change the class name. So we need to do the following
+        if (
+            "origin" in cls.__pydantic_generic_metadata__
+            and cls.__pydantic_generic_metadata__["origin"] is not None
+        ):
+            original_name = cls.__pydantic_generic_metadata__["origin"].__name__
```
baskaryan, line 170: feel like ive seen this logic in a few places, maybe worth factoring out?

libs/core/langchain_core/load/serializable.py
```diff
@@ -282,7 +316,8 @@ def _is_field_useful(inst: Serializable, key: str, value: Any) -> bool:
             except Exception as _:
                 value_neq_default = False
 
-    return field.required is True or value_is_truthy or value_neq_default
+    # If value is falsy and does not match the default
+    return value_is_truthy or value_neq_default
```
baskaryan, line 320: nit if value_is_truth we would've already returned

libs/core/langchain_core/runnables/utils.py
```diff
@@ -699,9 +705,57 @@ def include_event(self, event: StreamEvent, root_type: str) -> bool:
...
 
-class _SchemaConfig(BaseConfig):
-    arbitrary_types_allowed = True
-    frozen = True
+_SchemaConfig = ConfigDict(arbitrary_types_allowed=True, frozen=True)
+
+NO_DEFAULT = object()
+
+
+def create_base_class(
```
baskaryan, line 713: unit tests for this would be great

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -242,16 +250,16 @@ def _convert_any_typed_dicts_to_pydantic(
...
                 new_arg_type = _convert_any_typed_dicts_to_pydantic(
                     arg_type, depth=depth + 1, visited=visited
                 )
                 field_kwargs = {"default": ...}
                 if arg_desc := arg_descriptions.get(arg):
                     field_kwargs["description"] = arg_desc
-                fields[arg] = (new_arg_type, Field(**field_kwargs))
-        model = create_model(typed_dict.__name__, **fields)
+                fields[arg] = (new_arg_type, Field_v1(**field_kwargs))
+        model = create_model_v1(typed_dict.__name__, **fields)
```
baskaryan, line 262: we're still creating v1 models? is this to minimize stuff we need to test?

PR #25991, opened 2024-09-03
core[patch]: Fix injected args in tool signature

eyurtsev:
- Fix injected args in tool signature
- Fix another unit test that was using the wrong namespace import in pydantic


eyurtsev:
Force merging as it's failing due to compatibilty tests using pydantic

PR #26002, opened 2024-09-03
fix crash when using create_xml_agent with parameterless function as …

QunBB:
When using `create_xml_agent` or `create_json_chat_agent` to create a agent, and the function corresponding to the tool is a parameterless function, the `XMLAgentOutputParser` or `JSONAgentOutputParser` will parse the tool input into an empty string, `BaseTool` will parse it into a positional argument.
So, the program will crash finally because we invoke a parameterless function but with a positional argument.Specially, below code will raise StopIteration in [_parse_input](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/tools/base.py#L419)
```python
from langchain import hub
from langchain.agents import AgentExecutor, create_json_chat_agent, create_xml_agent

prompt = hub.pull("hwchase17/react-chat-json")

llm = ChatOpenAI()

# agent = create_xml_agent(llm, tools, prompt)
agent = create_json_chat_agent(llm, tools, prompt)

agent_executor.invoke(......)
```


QunBB:
@baskaryan I have added unit tests for the [related pull request](https://github.com/langchain-ai/langchain/pull/24866), please check it again.
The unit tests should raise StopIteration similarly with current code, as I mentioned above.So i think my code could be merged.

efriis:
overall I think this looks good - as a heads up we're charging towards a 0.3 release right now, so core is in code freeze until that goes out

Will mark to review after 0.3 is out!

ccurme (approved):

efriis (approved):

PR #26011, opened 2024-09-04
langchain-community[major]: Upgrade community to pydantic 2

hwchase17:
This PR upgrades langchain-community to pydantic 2.

* Most of this PR was auto-generated using code mods with gritql (https://github.com/eyurtsev/migrate-pydantic/tree/main)
* Subsequently, some code was fixed manually due to accommodate differences between pydantic 1 and 2

Breaking Changes:

-  Use TEXTEMBED_API_KEY and TEXTEMBEB_API_URL for env variables for text embed integrations:  https://github.com/langchain-ai/langchain/pull/26011/commits/cbea780492b7afde84b06f9780cdaeaa7770b2a7

Other changes:

- Added pydantic_settings as a required dependency for community. This may be removed if we have enough time to convert the dependency into an optional one.


eyurtsev:
This PR passes unit tests and extended tests across both python versions for langchain-community.

There are still (many) type checking issues that we will investigate separately.

PR #26031, opened 2024-09-04
langchain_openai: Make sure the response from the async client in the astream method of ChatOpenAI is properly awaited in case of "include_response_headers=True"

fhkingma:
- **Description:** This is a **one line change**. the `self.async_client.with_raw_response.create(**payload)` call is not properly awaited within the `_astream` method. In `_agenerate` this is done already, but likely forgotten in the other method.
  - **Issue:** Not applicable
  - **Dependencies:** No dependencies required.


ccurme (approved):

ccurme:
Thank you!

Issue #26046, opened 2024-09-04
`get_num_tokens_from_message` will cause 302 Redirect on specific image in message list


A-baoYang:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
>>> img_url="https://pdi-jtcg.zendesk.com/sc/attachments/conversations/66d6dbaf64263209ece921fd/nDdHO-089iwD8n3-EvAA-15Z/1711940417769.png"
>>> prompt_messages = [SystemMessage(content="please describe the image and tell me when I can use this")]
>>> prompt_messages.append(HumanMessagePromptTemplate.from_template(template=[{"type": "image_url", "image_url": {"url": img_url}}]))
>>> messages = ChatPromptTemplate.from_messages(prompt_messages)
>>> _msg_list = messages.format_prompt(**{}).messages
>>> llm.get_num_tokens_from_messages(_msg_list)
```

### Error Message and Stack Trace (if applicable)

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.11/site-packages/langchain_openai/chat_models/base.py", line 930, in get_num_tokens_from_messages
    image_size = _url_to_size(val["image_url"]["url"])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/langchain_openai/chat_models/base.py", line 2031, in _url_to_size
    response.raise_for_status()
  File "/usr/local/lib/python3.11/site-packages/httpx/_models.py", line 763, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Redirect response '302 Found' for url 'https://pdi-jtcg.zendesk.com/sc/attachments/conversations/66d6dbaf64263209ece921fd/nDdHO-089iwD8n3-EvAA-15Z/1711940417769.png'
Redirect location: 'https://p25.zdusercontent.com/sc/attachments/account/15330086/conversations/66d6dbaf64263209ece921fd/nDdHO-089iwD8n3-EvAA-15Z/1711940417769.png'
[...]

PR #26047, opened 2024-09-04
Solve #26046: Use requests with common headers rather than httpx to avoid 302 redirect error

A-baoYang:
When using `get_num_tokens_from_messages` function of ChatOpenAI, when there has image_url in message list, some image will cause 302 redirect error.

Change `httpx` to `requests` with headers to read the image in function `_url_to_size`


efriis:
Given openai sdk relies on httpx, let's stick with httpx.

Adding custom headers like this is inadvisable too.

If you file an issue or make a PR with an integration test that fails, that would be helpful for understanding the exact case that's failing for you!

A-baoYang:
I've mentioned the image_url that will cause error in the langchain module, but success when pass to OpenAI API.

```
image_url = "https://pdi-jtcg.zendesk.com/sc/attachments/conversations/66d6dbaf64263209ece921fd/nDdHO-089iwD8n3-EvAA-15Z/1711940417769.png"
```

The workaround is wrote here: https://github.com/langchain-ai/langchain/issues/26046

PR #26050, opened 2024-09-04
langchain[major]: Upgrade langchain to be pydantic 2 compatible

eyurtsev:
Upgrading the langchain package to be pydantic 2 compatible.

Had to remove some parts of unit tests in parsers that were relying on spying on methods since that fails with pydantic 2. The unit tests don't seem particularly good, so can be re-written at a future date.

Depends on: https://github.com/langchain-ai/langchain/pull/26057

Most of this PR was done using gritql for code mods, followed by some fixes done manually to account for changes made by pydantic


eyurtsev:
This PR was merged despite a few failing unit tests:

1. There is non-determinism in this unit test:

https://github.com/langchain-ai/langchain/blob/e19255bb3c7bfcfba492bb5b5b74a45281d8a4f7/libs/langchain/tests/unit_tests/retrievers/document_compressors/test_chain_filter.py#L24-L24

2. Extended tests still trigger some failing unit tests. We'll need to address those as well prior to release.

Issue #26058, opened 2024-09-04
Can't use mustache/jinja2 templates with an argument named "template"


ikbenale:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```py
>>> from langchain_core.prompts import PromptTemplate
>>> prompt = PromptTemplate.from_template("Say {template}")
>>> prompt.format(template="foo")
'Say foo'
>>> prompt = PromptTemplate.from_template("Say {{template}}", template_format="mustache")
>>> prompt.format(template="foo")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/alejandro/src/github.com/langchain-ai/langchain/libs/core/langchain_core/prompts/prompt.py", line 179, in format
    return DEFAULT_FORMATTER_MAPPING[self.template_format](self.template, **kwargs)
TypeError: mustache_formatter() got multiple values for argument 'template'
>>> prompt = PromptTemplate.from_template("Say {{template}}", template_format="jinja2")
>>> prompt.format(template="foo")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/alejandro/src/github.com/langchain-ai/langchain/libs/core/langchain_core/prompts/prompt.py", line 179, in format
    return DEFAULT_FORMATTER_MAPPING[self.template_format](self.template, **kwargs)
TypeError: jinja2_formatter() got multiple values for argument 'template'
```

### Error Message and Stack Trace (if applicable)

  File "/Users/alejandro/src/github.com/langchain-ai/langchain/libs/core/langchain_core/prompts/prompt.py", line 179, in format
    return DEFAULT_FORMATTER_MAPPING[self.template_format](self.template, **kwargs)
TypeError: mustache_formatter() got multiple values for argument 'template'

### Description

- Trying to format a prompt template in mustache or jinja2 with a `template` argument
- LangChain throws a `TypeError`

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 23.6.0: Mon Jul 29 21:13:00 PDT 2024; root:xnu-10063.141.2~1/RELEASE_X86_64
[...]

PR #26060, opened 2024-09-04
core: fix "template" not allowed as prompt param

ikbenale:
- **Description:**  fix "template" not allowed as prompt param
- **Issue:** #26058
- **Dependencies:** none

Additional guidelines:


efriis:
Could you add a unit test? 

I'm open to merging this with additional tests (as long as existing lint/test passes) because I don't think these two functions are used externally. However, parameter changes like this tend to create more breaking changes than they fix because anyone passing `template="my template"` intentionally would break, and the behavior is easily worked around by using a name other than `template`

Issue #26084, opened 2024-09-05
retry_min_seconds and retry_max_seconds is not work in OpenAIEmbeddings of langchain_openai


dpneko:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The `retry_max_seconds` and `retry_min_seconds` don't work in langchain_openai.OpenAIEmbeddings

``` python
from langchain_openai import OpenAIEmbeddings

embedding_function = OpenAIEmbeddings(
    model="text-embedding-ada-002",
    retry_max_seconds=40,
    retry_min_seconds=60,
    max_retries=10,
)
```

### Description

I checked the source code of langchain_openai.OpenAIEmbeddings but didn't find `retry_max_seconds` and `retry_min_seconds` used anywhere. In the deprecated class [langchain_community.embeddings.openai.OpenAIEmbeddings](https://github.com/langchain-ai/langchain/blob/de97d5064437c98f34dae0b0afa3b61162790726/libs/community/langchain_community/embeddings/openai.py#L44), ther are used in `@retry_decorator`. 

### System Info

langchain                                0.2.0rc2         
langchain-community                      0.2.0rc1        
langchain-core                           0.2.0rc1         
langchain-openai                         0.1.7          
langchain-text-splitters                 0.0.2

efriis:
discussion in #26092

efriis:
closing, built into openai sdk

PR #26092, opened 2024-09-05
Parnters: Using `embed_with_retry` and `async_embed_with_retry` for OpenAI Embedding

keenborder786:
- **Description:** As the issue points out, the attributes `retry_max_seconds` and `retry_min_seconds` are not being used in `OpenAIEmbebeddings` in `langchain_openai` partners package like they are being used in deprecated version of `OpenAIEmbebeddings`. Therefore I have switched the invocation of embedding through same function as in the deprecated version of `OpenAIEmbebeddings` which is through `embed_with_retry` and `async_embed_with_retry`.
- **Issue:**  #26084


keenborder786:
@efriis

efriis:
I believe this was removed intentionally when openai released their v1 sdk, which had retries for common errors built-in: https://github.com/openai/openai-python/tree/93aa548201af70646fe651cc01fd875f32b00202?tab=readme-ov-file#retries

Could you confirm this functionality makes sense? If so can add a comment to the attributes clarifying that those retries are handled internally

keenborder786:
@efriis sure let me check

keenborder786:
@efriis you are right, no need .

efriis:
great thanks for checking!

PR #26094, opened 2024-09-05
core[patch]: add bedrock to load.mapping

ccurme:


ccurme:
bypassing CI because failures appear unrelated: https://github.com/langchain-ai/langchain/actions/runs/10722796442/job/29734596739?pr=26095

Issue #25915, opened 2024-09-01
[BUG]: HuggingFacePipeline not initializes correctly when passing transformers pipeline


Kirushikesh:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from transformers import pipeline
from langchain_huggingface import HuggingFacePipeline

model_id = "microsoft/Phi-3.5-mini-instruct"
pipe = pipeline("text-generation", model=model_id)

llm = HuggingFacePipeline(pipeline=pipe)
llm.model_id # Output: 'gpt2'
```
While initializing the HuggingFacePipeline through model_id gives,

```python
from langchain_huggingface import ChatHuggingFace, HuggingFacePipeline

llm = HuggingFacePipeline.from_model_id(
    model_id="microsoft/Phi-3.5-mini-instruct",
    task="text-generation",
)
llm.model_id # Output is : 'microsoft/Phi-3.5-mini-instruct'
```

### Description

I am using langchain, when using HuggingFacePipeline first I initialized it with custom transformers pipeline doing that results in incorrect model name as 'gpt-2'. While using the same module and initialization through .model_id() recognizes model_id correctly as phi3. Due to the model_id mismatch the other modules are getting impacted like using ChatHuggingFace because ChatHuggingFace uses model_id to initialize the tokenizer.

### System Info

"pip freeze | grep langchain"
```
langchain==0.2.15
langchain-community==0.0.27
langchain-core==0.2.37
langchain-huggingface==0.0.3
langchain-text-splitters==0.2.2
```
platform: Linux
Python: 3.10.6

AndrewEffendi:
hi @eyurtsev 
I can confirm that this issue is still persist, Can we take on this issue? We are group of students at UofT developers interested in this issue.
I have seen that you have wrote some code in the PR that is now closed. Do you want us to add some unit test to confirm its working?

PR #26100, opened 2024-09-05
fix: add HuggingFacePipeline model_id parameter

NotShrirang:
**PR title**: "Fixes issue with model parameter not getting initialized correctly when passing transformers pipeline"
- Package: "langchain_huggingface"

- **Issue:** #25915
- **Dependencies:** None

The current version of the [code](https://github.com/langchain-ai/langchain/blob/master/libs/partners/huggingface/langchain_huggingface/llms/huggingface_pipeline.py) contains default parameter of `model_id` in `HuggingFacePipeline` class. If the user passes a transformers pipeline object, ideally the default value should get overridden, but it doesn't. This PR fixes it.

Cheers!


eyurtsev (changes requested):
Should this not be fixed via the initialiaztion path instead?

i.e., create `__init__` or `@root_validator(pre=True)` and set model_id appropriately based on whether pipeline exists or whether a user provided a value for the model id

NotShrirang:
@eyurtsev I have added the root initializer as you had suggested. Please review the changes.

NotShrirang:
Hello @eyurtsev can you review the changes?

eyurtsev:
Hi @NotShrirang, i re-merged the code against the updated master branch (which uses pydantic 2)

The original code did not look correct in terms of initialization logic since it was not handling the defaults properly -- i rewrote the code to match the description in the PR.

Given that the original PR had errors in it, could you please add a unit test before we can merge this PR that tests that initialization is working as expected.

PR #26103, opened 2024-09-05
Changed docker command of redis to stack redis

VpkPrasanna:

changed the docker command from normal redis to stack redis 
When i tried to follow the example faced an issue but with stack redis it worked as expected

    - **Description:** Docker command changed : Langchain
    - **Issue:**  With normal redis it is not working as expected so replaced with stack redis
    - **Twitter handle:** https://x.com/VpkPrasanna


efriis:
believe the old one should work too for chat message history! If you include the error I'm happy to reopen

PR #26104, opened 2024-09-05
groq, fireworks, text-splitters

ccurme:


libs/text-splitters/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "langchain-text-splitters"
-version = "0.2.4"
+version = "0.2.0.dev0"
```
baskaryan, line 7: 0.3.0.dev0

PR #26108, opened 2024-09-05
core: reduce warnings

ccurme:


libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -391,7 +391,9 @@ def my_function(arg1: Nested) -> None:
     assert actual == expected
 
 
-@pytest.mark.xfail(reason="Pydantic converts Optional[str] to str in .schema()")
+@pytest.mark.xfail(
+    reason="Pydantic converts Optional[str] to str in .model_json_schema()"
```
eyurtsev, line 395: hmm is this one still failing we the new json schema?

ccurme: yeah appears so

PR #26121, opened 2024-09-05
core[patch]: Fix regression in core

eyurtsev:
Limited to unit testing code -- did not cause any actual issues

### DIFF
--- a/libs/core/tests/unit_tests/pydantic_utils.py
+++ b/libs/core/tests/unit_tests/pydantic_utils.py
@@ -70,13 +70,14 @@ def _remove_enum_description(obj: Any) -> None:
 
 def _schema(obj: Any) -> dict:
     """Return the schema of the object."""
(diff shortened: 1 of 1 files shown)


libs/core/tests/unit_tests/pydantic_utils.py
```diff
@@ -70,13 +70,14 @@ def _remove_enum_description(obj: Any) -> None:
...
-    # Remap to old style schema
+
     if not is_basemodel_subclass(obj):
         raise TypeError(
             f"Object must be a Pydantic BaseModel subclass. Got {type(obj)}"
         )
+    # Remap to old style schema
     if not hasattr(obj, "model_json_schema"):  # V1 model
-        return obj.model_json_schema()
+        return obj.schema()
```
ccurme, line 80: thank you

PR #26126, opened 2024-09-05
core[patch]: Remove some usage of .copy() in favor of .model_copy()

eyurtsev:
Address under place where deprecated functionality is used.


eyurtsev:
Failed twice due to a flaky test -- we need to increase the time allowance for the rate limiter tests

PR #26136, opened 2024-09-06
partners[milvus]: fix integration test issues

zc277584121:
fix some integration test issues: https://github.com/langchain-ai/langchain/actions/runs/10688447230/job/29628412258


zc277584121:
there are some issues introduced in history, i fixed them:
- fix problem that can't get db_name from addr
- fix mmr search test result
- add milvus_model dependency into integration test of poetry requirement
- update milvus-lite dependency to 2.4.10(x86 linux released), which is the latest one with the array bug fixed

Issue #26085, opened 2024-09-05
[Partners-HuggingFace] Breaking Change - impossible to skip login after PR #23309


morgandiverrez:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
 HuggingFaceEndpoint(
            endpoint_url=self.setting.api_base,
            temperature=self.setting.temperature,
            repetition_penalty=self.setting.repetition_penalty,
            max_new_tokens=self.setting.max_new_tokens,
            server_kwargs={'trust_env': True},
        )
```

### Error Message and Stack Trace (if applicable)

  File "/home/Morgan.Diverrez/code/GitHub/tock-arkea/gen-ai/orchestrator-server/src/main/python/server/src/gen_ai_orchestrator/services/langchain/factories/llm/hugging_face_tgi_llm_factory.py", line 41, in get_language_model
    return HuggingFaceEndpoint(
  File "/home/Morgan.Diverrez/code/GitHub/tock-arkea/gen-ai/orchestrator-server/src/main/python/server/.venv/lib/python3.9/site-packages/pydantic/v1/main.py", line 341, in __init__
    raise validation_error
pydantic.v1.error_wrappers.ValidationError: 1 validation error for HuggingFaceEndpoint
__root__
  Did not find huggingfacehub_api_token, please add an environment variable `HUGGINGFACEHUB_API_TOKEN` which contains it, or pass `huggingfacehub_api_token` as a named parameter. (type=value_error)

### Description

After PR #23309 a few months ago, when we leave **huggingfacehub_api_token** at its default value, None, and do not set the environment variable HF_TOKEN or HUGGINGFACEHUB_API_TOKEN, the `get_from_dict_or_env` function at lines 197 and 201 also has a default value of None, similar to **huggingfacehub_api_token**. The `get_from_dict_or_env` function calls `get_from_env`, which raises an error if no token is set in the environment variable.

Therefore, we need to use another solution to retrieve environment variables.

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  #127~20.04.1-Ubuntu SMP Thu Jul 11 15:36:12 UTC 2024
[...]

PR #26153, opened 2024-09-06
core: [env.py/get_from_env] change default value None -> sentinel object

morgandiverrez:
**Description**
   Install sentinel value by default in get_from_env and get_from_dict_and_env function in env.py for let's None value by default and not raise ValueError in this case.

**Issue :**
[ [Partners-HuggingFace] Breaking Change - impossible to skip login after PR #23309 #26085](https://github.com/langchain-ai/langchain/issues/26085)

### DIFF
--- a/libs/core/langchain_core/utils/env.py
+++ b/libs/core/langchain_core/utils/env.py
@@ -59,6 +59,7 @@ def get_from_dict_or_env(
     else:
         return get_from_env(key_for_err, env_key, default=default)
 
+
 def get_from_env(
         key: str,
         env_key: str,
--- a/libs/core/langchain_core/utils/env.py
+++ b/libs/core/langchain_core/utils/env.py
@@ -55,15 +55,17 @@ def get_from_dict_or_env(
     return get_from_env(key_for_err, env_key, default=default)
 
 
-def get_from_env(key: str, env_key: str, default: Optional[str] = None) -> str:
-    """Get a value from a dictionary or an environment variable.
+_get_from_env_default_sentinel = object()
 
+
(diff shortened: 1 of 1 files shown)


eyurtsev (commented):
This may break a lot of static typing in dependent packages that assume that the return type can only be `str`. If you want to make the change you'll likely need to improve the type annotation behavior so that it captures the correct semantics

morgandiverrez:
I changed the code to only return a string as you asked. Do you think I can further improve this code?

Benvii:
Thanks @morgandiverrez , as mentionned in the issue :

> Introducing a sentinel value may be considered as a breaking change, get_from_env will then accept None as default value without raising any exception which is currently not accepted 🤔. This would assume that users rely on this exception being raised, which I my opinion shouldn't be considered as "nominal" usage of this method.

https://github.com/langchain-ai/langchain/issues/26085#issuecomment-2331968516

What do you think about this @eyurtsev ?

Benvii (approved):

eyurtsev (changes requested):
Hi thanks for the update -- this still won't work for the type annotations.

Presumably we're changing the behavior to support a default=None type. When a user passes deafult=None, the return type should be Optional[str] rather than `str`.

The code can be updated using typing.overload (https://docs.python.org/3/library/typing.html#overload)

eyurtsev:
Moved to needs status since there was no activity on this.

Developers can use the existing `from_env` and `secret_from_env` in langchain_core.utils to get factories for this use case.

eyurtsev:
Closing the PR due to lack of activity. Will be happy to re-review if folks want to tackle and can get the type annotations working correctly.

PR #26157, opened 2024-09-06
core[patch]: resolve warnings

ccurme:
Resolve a batch of warnings

### DIFF
--- a/libs/core/langchain_core/tools/base.py
+++ b/libs/core/langchain_core/tools/base.py
@@ -481,12 +481,14 @@ class ChildTool(BaseTool):
             if input_args is not None:
                 if issubclass(input_args, BaseModel):
                     result = input_args.model_validate(tool_input)
+                    result_dict = result.model_dump()
                 else:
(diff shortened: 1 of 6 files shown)


libs/core/langchain_core/tools/base.py
```diff
@@ -479,10 +479,16 @@ def _parse_input(self, tool_input: Union[str, Dict]) -> Union[str, Dict[str, Any
             return tool_input
         else:
             if input_args is not None:
-                result = input_args.parse_obj(tool_input)
+                if issubclass(input_args, BaseModel):
+                    result = input_args.model_validate(tool_input)
+                    result_dict = result.model_dump()
+                else:
```
eyurtsev, line 485: we should probably do

```python
else issubclass(input_args, pydantic.v1.BaseModel): 
  ..
else:
  raise NotImplementedError() # or assertion error
```

to be very explicit about compatibility (so we can also grep for it later on in the code)

baskaryan (approved):

PR #26161, opened 2024-09-06
core[patch]: fix deprecated pydantic code

baskaryan:


libs/core/langchain_core/tracers/langchain.py
```diff
@@ -69,11 +71,16 @@ def _get_executor() -> ThreadPoolExecutor:
...
 
 def _run_to_dict(run: Run) -> dict:
-    return {
-        **run.dict(exclude={"child_runs", "inputs", "outputs"}),
-        "inputs": run.inputs.copy() if run.inputs is not None else None,
-        "outputs": run.outputs.copy() if run.outputs is not None else None,
-    }
+    # TODO: Update once langsmith moves to Pydantic V2 and we can swap run.dict for
+    # run.model_dump
+    with warnings.catch_warnings():
```
baskaryan, line 76: should double check that this is actually needed, was messing around with a local version of langsmith that i upgraded to v2

ccurme (approved):

libs/core/langchain_core/utils/pydantic.py
```diff
@@ -134,42 +135,45 @@ def pre_init(func: Callable) -> Any:
...
-        """
-        # Insert default values
-        fields = cls.model_fields
-        for name, field_info in fields.items():
-            # Check if allow_population_by_field_name is enabled
-            # If yes, then set the field name to the alias
-            if hasattr(cls, "Config"):
-                if hasattr(cls.Config, "allow_population_by_field_name"):
-                    if cls.Config.allow_population_by_field_name:
+    with warnings.catch_warnings():
```
eyurtsev, line 138: nice!!!

libs/core/langchain_core/tracers/schemas.py
```diff
@@ -142,9 +143,14 @@ def assign_name(cls, values: dict) -> dict:
         return values
 
 
-ChainRun.update_forward_refs()
-ToolRun.update_forward_refs()
-Run.update_forward_refs()
+# TODO: Update once langsmith moves to Pydantic V2 and we can swap Run.model_rebuild
+# for Run.update_forward_refs
+with warnings.catch_warnings():
```
eyurtsev, line 148: I think you can undo these changes -- since it's using v1, but doesn't matter gith now. We'll likely need to delete schemas.py from the code base prior to full 0.3 release since the actual code also pulls in LLMResult (see on master)

eyurtsev (approved):

PR #26182, opened 2024-09-07
Add session expired retry to neo4j graph

tomasonjo:


efriis (approved):

jexp:
Can we also add retries for Deadlocks?
`Neo.TransientError.Transaction.DeadlockDetected` or in general for all _transient_ errors.

PR #25169, opened 2024-08-08
core: Add additional properties recursively for structured output(openai) when nested object is generated

fufufukakaka:
Hello. 
First of all, thank you for maintaining such a great project.

## Description
In https://github.com/langchain-ai/langchain/pull/25123, support for structured_output is added. However, `"additionalProperties": false` needs to be included at all levels when a nested object is generated.

error from current code: https://gist.github.com/fufufukakaka/e9b475300e6934853d119428e390f204
```
BadRequestError: Error code: 400 - {'error': {'message': "Invalid schema for response_format 'JokeWithEvaluation': In context=('properties', 'self_evaluation'), 'additionalProperties' is required to be supplied and to be false", 'type': 'invalid_request_error', 'param': 'response_format', 'code': None}}
```

Reference: [Introducing Structured Outputs in the API](https://openai.com/index/introducing-structured-outputs-in-the-api/)

```json
{
  "model": "gpt-4o-2024-08-06",
  "messages": [
    {
      "role": "system",
      "content": "You are a helpful math tutor."
    },
    {
      "role": "user",
      "content": "solve 8x + 31 = 2"
    }
  ],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "math_response",
      "strict": true,
      "schema": {
        "type": "object",
        "properties": {
          "steps": {
            "type": "array",
            "items": {
              "type": "object",
              "properties": {
                "explanation": {
                  "type": "string"
                },
                "output": {
                  "type": "string"
                }
              },
              "required": ["explanation", "output"],
              "additionalProperties": false
            }
          },
          "final_answer": {
            "type": "string"
          }
        },
        "required": ["steps", "final_answer"],
        "additionalProperties": false
      }
    }
  }
}
```

In the current code, `"additionalProperties": false` is only added at the last level. 
This PR introduces the `_add_additional_properties_key` function, which recursively adds `"additionalProperties": false` to the entire JSON schema for the request.

Twitter handle: `@fukkaa1225`

Thank you!

PR #26187, opened 2024-09-07
core: Add additional properties recursively for structured output(ope…

Karjan1:
Hi, thanks for amazing work! 

## Description 
Add support for Pydantic with Openai structured output. Followup on work of https://github.com/langchain-ai/langchain/pull/25169

I don't believe it's a fully ready feature, but it's good enough for our use case and might help others.  Maybe someone else will be willing to give a second look to make it fully ready.


efriis:
I'm pretty sure this case won't be necessary, and it may also conflict with some of the other pydantic work we're doing this week. Will let @ccurme comment

If you add a unit test showing a schema that doesn't work currently that would be helpful/necessary to get this merged!

Karjan1:
> I'm pretty sure this case won't be necessary, and it may also conflict with some of the other pydantic work we're doing this week. Will let @ccurme comment
> 
> If you add a unit test showing a schema that doesn't work currently that would be helpful/necessary to get this merged!

Great to hear that there is work on Pydantic integration. Honestly I don't feel like my code is added in the right place, but section responsible for Pydantic conversion was trickier to implement. I feel like it's a bigger work. So I just shared to possibly help others as a temporary solution.

ccurme:
Hi @Karjan1, thanks for the clarity around the intent of your contribution here. I think what would be most useful is a failing test case or minimal reproducible example that demonstrates the issue. We can then iterate on the details of the solution.

PR #26197, opened 2024-09-08
community: Reflection output parser

heiko-hotz:
This adds a custom chain to parse the output of the Reflection 70B model. It will extract the final answer inside `<output>` and `</output>` tags, as per https://huggingface.co/mattshumer/Reflection-Llama-3.1-70B


efriis:
closing because this isn't formatted properly as an output parser (it's formatted as a legacy chain instead). Would be great to add an integration test if you do showing how it works with one of the llm integrations that hosts reflection 70b!

PR #26199, opened 2024-09-08
docs: add ai21 tool calling example

miri-bar:
Add tool calling example to AI21 docs


efriis (dismissed):

efriis (commented):
actually this is in the wrong document - could you move to chat/ai21.ipynb instead?

efriis (approved):

PR #26203, opened 2024-09-08
community: Reflection output parser

heiko-hotz:
This adds a custom chain to parse the output of the Reflection 70B model. It will extract the final answer inside `<output>` and `</output>` tags, as per https://huggingface.co/mattshumer/Reflection-Llama-3.1-70B

I also added a test notebook as per feedback on #26197

### DIFF
--- a/libs/community/langchain_community/output_parsers/reflection_parser.py
+++ b/libs/community/langchain_community/output_parsers/reflection_parser.py
@@ -1,0 +1,40 @@
+from __future__ import annotations
+
+import re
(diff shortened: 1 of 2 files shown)


Undertone0809 (commented):
- relfection_test.ipynb could be more appropriately placed.
- It's hard for anyone to know it exists without documentation.
- Since it's not complicated to build this yourself, it's weird to build this exporter for this model alone, and the effort of checking the docs could probably be written.

PR #26206, opened 2024-09-09
add new script

WKRZY:
A new script module has been added, implementing LLM based on SiliconFlow's API. You can use the free API to learn about langchain.


efriis:
we're not open to scripts like this even if they're formatted as an LLM :)

PR #26214, opened 2024-09-09
couchbase: Add ttl support to caches & chat_message_history

nithishr:
**Description:** Add support to delete documents automatically from the caches & chat message history by adding a new optional parameter, `ttl`.


efriis:
will wait til after 0.3 is in! then will TAL

PR #26234, opened 2024-09-09
chore: typos fixed

CodeWithEmad:
While going through the chatbot tutorial, I noticed a couple of typos and grammatical issues. Also, the pip install command for langchain_community was commented out, but the document mentions installing it.

### DIFF
--- a/docs/docs/tutorials/chatbot.ipynb
+++ b/docs/docs/tutorials/chatbot.ipynb
@@ -105,7 +105,7 @@
     "\n",
     "## Quickstart\n",
     "\n",
-    "First up, let's learn how to use a language model by itself. LangChain supports many different language models that you can use interchangably - select the one you want to use below!\n",
+    "First up, let's learn how to use a language model by itself. LangChain supports many different language models that you can use interchangeably - select the one you want to use below!\n",
     "\n",
     "```{=mdx}\n",
     "import ChatModelTabs from \"@theme/ChatModelTabs\";\n",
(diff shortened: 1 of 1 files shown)


docs/docs/tutorials/chatbot.ipynb
```diff
@@ -254,14 +254,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# ! pip install langchain_community"
+    "pip install langchain_community"
```
efriis, line 257:
```suggestion
    "%pip install langchain_community"
```

docs/docs/tutorials/chatbot.ipynb
```diff
@@ -254,14 +254,14 @@
...
-    "# ! pip install langchain_community"
+    "%pip install langchain_community"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "After that, we can import the relevant classes and set up our chain which wraps the model and adds in this message history. A key part here is the function we pass into as the `get_session_history`. This function is expected to take in a `session_id` and return a Message History object. This `session_id` is used to distinguish between separate conversations, and should be passed in as part of the config when calling the new chain (we'll show how to do that)."
+    "After that, we can import the relevant classes and set up our chain which wraps the model and adds to this message history. A key part here is the function we pass into as the `get_session_history`. This function is expected to take in a `session_id` and return a Message History object. This `session_id` is used to distinguish between separate conversations, and should be passed in as part of the config when calling the new chain (we'll show how to do that)."
```
efriis, line 264:
```suggestion
    "After that, we can import the relevant classes and set up our chain which wraps the model and adds in this message history. A key part here is the function we pass into as the `get_session_history`. This function is expected to take in a `session_id` and return a Message History object. This `session_id` is used to distinguish between separate conversations, and should be passed in as part of the config when calling the new chain (we'll show how to do that)."
```

efriis (approved):

PR #26235, opened 2024-09-09
core[patch]: wrap mermaid node names w/ markdown in <p> tag

vbarda:
This fixes the issue where `__start__` and `__end__` node labels are being interpreted as markdown, as of the most recent Mermaid update

### DIFF
--- a/libs/core/tests/unit_tests/runnables/__snapshots__/test_graph.ambr
+++ b/libs/core/tests/unit_tests/runnables/__snapshots__/test_graph.ambr
@@ -1,31 +1,4 @@
 # serializer version: 1
-# name: test_double_nested_subgraph_mermaid[mermaid]
-  '''
-  %%{init: {'flowchart': {'curve': 'linear'}}}%%
-  graph TD;
-  	__start__([<p>__start__</p>]):::first
-  	parent_1(parent_1)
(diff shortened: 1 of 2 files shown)


libs/core/tests/unit_tests/runnables/__snapshots__/test_graph.ambr
```diff
@@ -1040,7 +1040,7 @@
   	PromptTemplate(PromptTemplate)
   	FakeListLLM(FakeListLLM)
   	Parallel_as_list_as_str_Input(Parallel<as_list,as_str>Input)
-  	Parallel_as_list_as_str_Output([Parallel_as_list_as_str_Output]):::last
+  	Parallel_as_list_as_str_Output([Parallel<as_list,as_str>Output]):::last
```
nfcampos, line 1043: why is this one changing?

nfcampos (approved):

libs/core/tests/unit_tests/runnables/__snapshots__/test_graph.ambr
vbarda: cause we used to do use key as display name for the last node, and we apply `_escape_node_label` on the key, but now since we're passing value explicitly, it's no longer escaped (which i think is actually better)

PR #26251, opened 2024-09-10
Switch graph - FalkorDB

galshubeli:

**Switch graph**: 

- **Feature Introduction:** Adds the switch_graph(graph_name: str) method to allow switching between graphs on the same database connection.

- **Why It's Needed:**
    Improves efficiency when working with multiple graphs under the same driver.
    Removes the need to open a new connection for each graph, reducing connection overhead.

- **Use Case:**
Ideal for applications managing several graphs simultaneously, where quick transitions between graphs are required without disconnecting or re-establishing new connections.

- **Goal:**
Enhance performance and streamline workflows by maintaining a single connection while managing different graphs.

Twitter handle: @tariyekorogha @baskaryan


eyurtsev (changes requested):
What does this change do? Why is it needed? Could you document the method and add a description in the PR summary?

galshubeli:
> What does this change do? Why is it needed? Could you document the method and add a description in the PR summary?

This change adds a method to switch between graphs on the same connection, making it more efficient to work with multiple graphs. I’ve documented the method and updated the PR summary.

libs/community/langchain_community/graphs/falkordb_graph.py
```diff
@@ -199,3 +199,19 @@ def add_graph_documents(
...
+
+        Args:
+            graph_name (str): The name of the graph to switch to.
+        """
+        self._graph = self._driver.select_graph(graph_name)
+        
+        try:
+            self.refresh_schema()
+        except Exception as e:
+            raise ValueError(f"Could not refresh schema. Error: {e}")
```
swilly22, line 217: Are there any additional state variables that need to change when switching graphs ?

galshubeli: No

PR #26255, opened 2024-09-10
community: Fix SPARQL Bug

### LINKED ISSUE #5428: parse_json_markdown is unable to parse json strings with nested triple backticks
### System Info

Langchain version 0.0.184, python 3.9.13
Function `parse_json_markdown` in langchain/output_parsers/json.py fails with input text string:
\`\`\`json
{
    "action": "Final Answer",
    "action_input": "Here's a Python script to remove backticks at the beginning and end of a string:\n\n\`\`\`python\ndef remove_backticks(s):\n    return s.strip('\`')\n\nstring_with_backticks = '\`example string\`'\nresult = remove_backticks(string_with_backticks)\nprint(result)\n\`\`\`\n\nThis script defines a function called \`remove_backticks\` that takes a string as input and returns a new string with backticks removed from the beginning and end. It then demonstrates how to use the function with an example string."
}
\`\`\`

Potential case of error:
`match.group(2)` in the function `parse_json_markdown` contains only the string up to the first occurrence of the second triple backticks:

{
    "action": "Final Answer",
    "action_input": "Here's a Python script to remove backticks at the beginning and end of a string:\n\n

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [X] Output Parsers
- [ ] Document Loaders
- [ ] Memory
- [X] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Called function `parse_json_markdown` in langchain/output_parsers/json.py with input text string:
\`\`\`json
{
    "action": "Final Answer",
    "action_input": "Here's a Python script to remove backticks at the beginning and end of a string:\n\n\`\`\`python\ndef remove_backticks(s):\n    return s.strip('\`')\n\nstring_with_backticks = '\`example string\`'\nresult = remove_backticks(string_with_backticks)\nprint(result)\n\`\`\`\n\nThis script defines a function called \`remove_backticks\` that takes a string as input and returns a new string with backticks removed from the beginning and end. It then demonstrates how to use the function with an example string."
}
\`\`\`

### Expected behavior

Function `parse_json_markdown` should return the following json string
{
"action": "Final Answer",
"action_input": "Here's a Python script to remove backticks at the beginning and end of a string:\n\n\`\`\`python\ndef remove_backticks(s):\n return s.strip('\`')\n\nstring_with_backticks = '\`example string\`'\nresult = remove_backticks(string_with_backticks)\nprint(result)\n\`\`\`\n\nThis script defines a function called \`remove_backticks\` that takes a string as input and returns a new string with backticks removed from the beginning and end. It then demonstrates how to use the function with an example string."
}

LiMingchen159:
**Description:**
This PR addresses the formatting errors in `generated`_sparql that were automatically generated by the large prediction model. The specific formatting issues have been documented by the user `i-arnab` in issue [#5428](https://github.com/langchain-ai/langchain/issues/5428). An attempt to resolve these issues was made in [#24262](https://github.com/langchain-ai/langchain/discussions/24262), but it was unsuccessful.

**Issue:**
Resolves [#5428](https://github.com/langchain-ai/langchain/issues/5428), related to [#24262](https://github.com/langchain-ai/langchain/discussions/24262)

**Dependencies:**
Python's `re` module is used for regex operations. No new external dependencies were introduced.

**Twitter handle:**
N/A


efriis:
will mark for support but I don't think this is the right approach because it's just replacing certain string formats, which is brittle

one could also imagine an llm starting an output with e.g. tick tick tick sparql or something representing the format, which would also be wrong

LiMingchen159:
> will mark for support but I don't think this is the right approach because it's just replacing certain string formats, which is brittle
> 
> one could also imagine an llm starting an output with e.g. tick tick tick sparql or something representing the format, which would also be wrong

Thank you for your feedback! In my actual use case, I’ve been working with the GPT-3.5 model, and I’ve found that around 60%-80% of the responses face this issue, where the format includes unnecessary "```". Initially, I attempted to address this by adjusting the prompt, but the results were not ideal. I also tried making a second call to the model to correct the output, but this wasn’t fully effective and consumed a lot of tokens.

Therefore, I opted for this direct approach. In my local setup, I haven’t encountered this issue again since implementing this method, and it has significantly improved the success rate of generating SPARQL queries to retrieve graph data. The solution only involves a simple cleanup of the start and end markers and does not affect the actual SPARQL content.

If you have a better approach in mind, please feel free to share, and I’d be happy to try it out and test it with my current project.

LiMingchen159:
Hey, @efriis. Do you know if this will be merged? I have to reapply the modification every time I update langchain-community.

efriis:
going to close this because it's more of a forced feature add rather than a bug.

You'll probably have more luck orchestrating this SPARQL generation in langgraph, such that it's easy to add this cleaning logic to your flow as well. If just for you, would recommend investigating that!

If you want to reopen, here's what I'd suggest:
- add a new input param to the chain like `postprocess_function: Optional[RunnableLambda]` that defaults to None
- if that's passed, run the cleaning function

rather than hardcoding a specific postprocessing function for the llm call.

Additionally, just swapping out the llm for any runnable would work for this too, as you could pass `ChatModel | RunnableLambda` as the model to get the outputs cleaned

PR #26265, opened 2024-09-10
community: remove sambaverse

jhpiedrahitao:
removing Sambaverse llm model and references given is not available after Sep/10/2024

image


jhpiedrahitao:
@ccurme sorry for tagging you, can you take a look at this, we are removing the Sambaverse model given As part of our ongoing efforts to improve and expand our offerings, we will deprecate our existing product Sambaverse on 10th Sept 2024.

* We will include the new [sambanova cloud](https://cloud.sambanova.ai/) chat model integration in our next PR (WIP)

libs/community/langchain_community/llms/sambanova.py
```diff
@@ -9,464 +9,6 @@
...
-                "modelName": sambaverse_model_name,
-            },
-            json=data,
-            stream=True,
-        )
-        for chunk in SVEndpointHandler._process_streaming_response(response):
-            yield chunk
-
-
-class Sambaverse(LLM):
```
ccurme, line 165: (nit) Might consider leaving a placeholder here that raises an informative error message if folks try to import it, letting them know what they should be using / plan to use instead.

ccurme (approved):

Issue #26250, opened 2024-09-10
schema_json fail with Pydantic BaseModel in langchain-core==0.3.0.dev4


liuhetian:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code
``` python
from langchain_openai import ChatOpenAI
model_chat = ChatOpenAI(api_key='foo')
print(model_chat.get_output_schema().schema_json()[:5])  # ok 

from pydantic import BaseModel
from typing import Any
class Foo(BaseModel):
    bar: Any
model_chat.with_structured_output(Foo, method='json_schema', strict=True).get_output_schema().schema_json()  
# PydanticUserError
```

### Error Message and Stack Trace (if applicable)

PydanticUserError: `_oai_structured_outputs_parser_output` is not fully defined; you should define `PydanticBaseModel`, then call `_oai_structured_outputs_parser_output.model_rebuild()`.

### Description

meet this question when using langserve `add_routes(app, mychain, path='/faq')`

### System Info

langchain-core==0.3.0.dev4
langchain-openai==0.2.0.dev2

pydantic==2.9.1
pydantic-extra-types==2.5.0
pydantic-settings==2.2.1
pydantic_core==2.23.3

python=3.12

eyurtsev:
After fix it doesn't crash, but also doesn't return the expected schema, so the fix is only partial.

Users will still need to overwrite it using with_types (which is generally recommended)

I'll double check, but believe this is actually not a regression, but an existing issue in 0.2x as well

PR #26267, opened 2024-09-10
core[patch]: Propagate module name to create model

eyurtsev:
* This allows pydantic to correctly resolve annotations necessary for building pydantic models dynamically.
* Makes a small fix for RunnableWithMessageHistory which was fetching the OutputType from the  RunnableLambda that was yielding another RunnableLambda. This doesn't propagate the output of the RunnableAssign fully (i.e., with concrete type information etc.)

Resolves issue: https://github.com/langchain-ai/langchain/issues/26250

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -348,6 +348,7 @@ class Runnable(Generic[Input, Output], ABC):
         return create_model(
             self.get_name("Input"),
             __root__=root_type,
+            __module_name=self.__class__.__module__,
         )
 
     def get_input_jsonschema(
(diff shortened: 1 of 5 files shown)


libs/core/langchain_core/runnables/base.py
```diff
@@ -348,6 +348,7 @@ def get_input_schema(
         return create_model(
             self.get_name("Input"),
             __root__=root_type,
+            __module_name=self.__class__.__module__,
```
baskaryan, line 351: some comments would help, dont have an intuition for why this is needed

eyurtsev: added comments

nfcampos (approved):

PR #26269, opened 2024-09-10
huggingface[patch]: add integration tests

ccurme:
Add standard tests for ChatHuggingFace. About half of these fail currently.


ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain/actions/runs/10798250107/job/29951337149

PR #26283, opened 2024-09-10
core[patch]: manually coerce ToolMessage args

baskaryan:


docs/docs/how_to/extraction_examples.ipynb
baskaryan, line 1: @ccurme seeing the output of some cells as bunch of empty lists `people=[]` not sure if that's intentional

libs/core/langchain_core/messages/tool.py
```diff
@@ -82,15 +83,21 @@ def get_lc_namespace(cls) -> List[str]:
         Default is ["langchain", "schema", "messages"]."""
         return ["langchain", "schema", "messages"]
 
+    @model_validator(mode="before")
+    @classmethod
+    def coerce_args(cls, values: dict) -> dict:
+        content = values["content"]
```
eyurtsev, line 89: Since this is the response from an invocation and is expected to go to a chat model I'm OK with a more generic str() invocation regardless of type

try:
   str(content)
except ValueError:
   some useful error

eyurtsev (approved):

docs/docs/how_to/extraction_examples.ipynb
ccurme: Took a look at this. I think code is fine. The point of this example is to show how few-shot examples can remove false positives. Unfortunately (?) models have improved and this example no longer generates false positives.

PR #26297, opened 2024-09-11
experimental: Handle possible KeyError in 'convert_to_graph_documents' in llm.py

olliequ:
It's possible some returned dicts don't have the "head" or "tail_type" key which results in a KeyError if not handled (which it currently isn't). So, ignore dicts returned that displays this behaviour, as well as those who have `head` as "". I noticed `rel` dicts that have `head` as "" (empty string) are usually dicts which are text duplicates of a **previous** `rel`, hence why they can be safely ignored.


efriis:

that being said, this pr would need some more detail in order to get reviewed. ideally add some unit tests for the issue it fixes!

PR #26307, opened 2024-09-11
openai[patch] get output_type when using with_structured_output 

liuhetian:
- This allows pydantic to correctly resolve annotations necessary when using openai new param `json_schema`

Resolves issue: #26250


eyurtsev (approved):

eyurtsev:
Would need to take a more careful look to see if that's the appropriate place to set the output_type... current code seems to ignore the possibility that the schema is a dict

eyurtsev:
@baskaryan tagging you in case this is a quick fix for you

baskaryan (approved):
awesome, thank you!

PR #26327, opened 2024-09-11
core[patch]: fix regression in convert_to_openai_tool with instances of Tool

ccurme:
```python
from langchain_core.tools import Tool

def my_function(x: int) -> int:
    return x + 2

tool = Tool(
    name="tool_name",
    func=my_function,
    description="test description",
)
convert_to_openai_tool(tool)
```

Current:
```
{'type': 'function',
 'function': {'name': 'tool_name',
  'description': 'test description',
  'parameters': {'type': 'object',
   'properties': {'args': {'type': 'array', 'items': {}},
    'config': {'type': 'object',
     'properties': {'tags': {'type': 'array', 'items': {'type': 'string'}},
      'metadata': {'type': 'object'},
      'callbacks': {'anyOf': [{'type': 'array', 'items': {}}, {}]},
      'run_name': {'type': 'string'},
      'max_concurrency': {'type': 'integer'},
      'recursion_limit': {'type': 'integer'},
      'configurable': {'type': 'object'},
      'run_id': {'type': 'string', 'format': 'uuid'}}},
    'kwargs': {'type': 'object'}},
   'required': ['config']}}}
```

Here:
```
{'type': 'function',
 'function': {'name': 'tool_name',
  'description': 'test description',
  'parameters': {'properties': {'__arg1': {'title': '__arg1',
     'type': 'string'}},
   'required': ['__arg1'],
   'type': 'object'}}}
```

### DIFF
--- a/libs/core/langchain_core/utils/function_calling.py
+++ b/libs/core/langchain_core/utils/function_calling.py
@@ -280,7 +280,10 @@ def format_tool_to_openai_function(tool: BaseTool) -> FunctionDescription:
     Returns:
         The function description.
     """
-    if tool.tool_call_schema:
+    from langchain_core.tools import simple
+
+    is_simple_oai_tool = isinstance(tool, simple.Tool) and not tool.args_schema
+    if tool.tool_call_schema and not is_simple_oai_tool:
         return convert_pydantic_to_openai_function(
             tool.tool_call_schema, name=tool.name, description=tool.description
         )
--- a/libs/core/tests/unit_tests/utils/test_function_calling.py
+++ b/libs/core/tests/unit_tests/utils/test_function_calling.py
@@ -37,7 +37,7 @@ except ImportError:
 from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
 from langchain_core.pydantic_v1 import BaseModel, Field
 from langchain_core.runnables import Runnable, RunnableLambda
-from langchain_core.tools import BaseTool, Tool, tool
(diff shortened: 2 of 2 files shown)


libs/core/langchain_core/utils/function_calling.py
```diff
@@ -280,7 +280,10 @@ def format_tool_to_openai_function(tool: BaseTool) -> FunctionDescription:
     Returns:
         The function description.
     """
-    if tool.tool_call_schema:
```
ccurme, line 283: Unclear to me when this would be falsey

libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -295,6 +295,27 @@ def test_convert_to_openai_function(
     runnable_expected["parameters"] = parameters
     assert actual == runnable_expected
 
+    # Test simple Tool
+    def my_function(input_string: str) -> str:
```
hinthornw, line 299: Do we have a check that if we did this with structured tool / decorator it actually uses the argument name?

For prompting purposes it matters

hinthornw (approved):

libs/core/tests/unit_tests/utils/test_function_calling.py
ccurme: Might not be following but is this what you're after? https://github.com/langchain-ai/langchain/pull/26327/commits/84322fda9cd0f0e98041165ae8ddcc6d7d3a4ed0

PR #26334, opened 2024-09-11
core[patch]: Replace BaseMessage with AnyMessage

eyurtsev:
Replace BaseMessage with AnyMessage for serialization purposes


eyurtsev:
Going to close for now and fix just in langserve since it appears only during serde

PR #26336, opened 2024-09-11
core[patch]: support pydantic v1 annotations in tool arguments

ccurme:
If all pydantic annotations in function signature are V1, use V1 `validate_arguments`.

### DIFF
--- a/libs/core/langchain_core/tools/base.py
+++ b/libs/core/langchain_core/tools/base.py
@@ -174,7 +174,9 @@ def _is_pydantic_annotation(annotation: Any, pydantic_version: str = "v2") -> bo
         return False
 
 
-def _function_annotations_are_pydantic_v1(signature: inspect.Signature) -> bool:
+def _function_annotations_are_pydantic_v1(
+    signature: inspect.Signature, func: Callable
+) -> bool:
     """Determine if all Pydantic annotations in a function signature are from V1."""
     any_v1_annotations = any(
         _is_pydantic_annotation(parameter.annotation, pydantic_version="v1")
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/tools/base.py
```diff
@@ -208,16 +231,19 @@ def create_schema_from_function(
...
-    with warnings.catch_warnings():
-        # We are using deprecated functionality here.
-        # This code should be re-written to simply construct a pydantic model
-        # using inspect.signature and create_model.
-        warnings.simplefilter("ignore", category=PydanticDeprecationWarning)
-        validated = validate_arguments(func, config=_SchemaConfig)  # type: ignore
-
     sig = inspect.signature(func)
 
+    if _function_annotations_are_pydantic_v1(sig):
```
eyurtsev, line 236: What happens if there's mixed v1 and v2 namespaces? I'd love to raise an assertion error here if it doesn't work and be explicit given that this code is in core

libs/core/langchain_core/tools/base.py
```diff
@@ -164,6 +165,35 @@ def _infer_arg_descriptions(
     return description, arg_descriptions
 
 
+def _is_pydantic_annotation(annotation: Any, pydantic_version: str = "v2") -> bool:
+    """Determine if a type annotation is a Pydantic model."""
+    base_model_class = BaseModelV1 if pydantic_version == "v1" else BaseModel
+    try:
+        return issubclass(annotation, base_model_class)
+    except TypeError:
```
ccurme, line 173: @eyurtsev this look correct?

eyurtsev (approved):

PR #26339, opened 2024-09-11
rfc: run docs gha

baskaryan:


.github/workflows/update_notebooks.yml
```diff
@@ -0,0 +1,78 @@
...
+        MISTRAL_API_KEY: ${{ secrets.MISTRAL_API_KEY }}
+        COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
+        NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
+        GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
+        GOOGLE_SEARCH_API_KEY: ${{ secrets.GOOGLE_SEARCH_API_KEY }}
+        GOOGLE_CSE_ID: ${{ secrets.GOOGLE_CSE_ID }}
+        # TODO: Remove this once 0.3 is relased
+        PIP_PRE: 1
+      run: |
+        poetry run jupyter nbconvert --to notebook --execute --allow-errors docs/docs/how_to/*.ipynb;
```
eyurtsev, line 57: Could you document what this does?

Why are we allowing for errors?

efriis: If I had to guess, this is temporary until we fix em

If not, we should make this temporary until we fix the notebooks intentional error cells to mark them as "error allowed" (I think jupyter has a concept of this)

efriis (approved):

PR #26343, opened 2024-09-11
core: deal with tool errors properly

isahers1:
Always raise ToolExceptions


libs/core/langchain_core/tools/base.py
```diff
@@ -725,50 +736,61 @@ async def arun(
...
-            if not self.handle_validation_error:
+                status = "error"
+            except ToolException as e:
+                if not self.handle_tool_error:
+                    error_to_raise = e
+                else:
+                    content = _handle_tool_error(e, flag=self.handle_tool_error)
+                status = "error"
+            except Exception as e:
+                raise ToolException(str(e)) from e
```
baskaryan, line 784: i think we want this to occur before the except in 777, so that it can catch it and do error handling

PR #26345, opened 2024-09-11
core[major]: Add restrictions on create_model field names to match pydantic constraints

eyurtsev:
Pydantic 2 is stricter in terms of which field names are allowed in pydantic models.

This PR results in the following breaking changes:

These will raise ValueErrors: 

```python
ChatPromptTemplate([("system", "{_private}")]).get_input_schema()
ChatPromptTemplate([("system","{model_json_schema}")]).get_input_schema()
```

This PR should properly suppress warnings for the following cases:

```python
ChatPromptTemplate([("system", "{schema}")]).get_input_schema()
ChatPromptTemplate([("system","{model_id}")]).get_input_schema()
```

### DIFF
--- a/libs/core/langchain_core/runnables/utils.py
+++ b/libs/core/langchain_core/runnables/utils.py
@@ -6,7 +6,6 @@ import ast
 import asyncio
 import inspect
 import textwrap
-import warnings
 from functools import lru_cache
 from inspect import signature
 from itertools import groupby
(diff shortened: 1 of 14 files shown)


libs/core/langchain_core/runnables/utils.py
```diff
@@ -26,24 +25,17 @@
...
-    GenerateJsonSchema,
-    JsonSchemaMode,
-)
-from pydantic.v1 import BaseModel as BaseModelV1
 from typing_extensions import TypeGuard
 
 from langchain_core.runnables.schema import StreamEvent
 
+# Re-export create-model for backwards compatibility
+from langchain_core.utils.pydantic import create_model as create_model
```
eyurtsev, line 37: retains backwards compatibility for create_model

PR #26350, opened 2024-09-11
docs[community]: Fix raw string in docstring

### LINKED ISSUE #26212: SyntaxWarning: invalid escape sequence '\s'
### Checked other resources

### Example Code

```shell
mypy .
```

```shell
./python3.12/site-packages/langchain_community/document_loaders/recursive_url_loader.py:54: SyntaxWarning: invalid escape sequence '\s'
  """Recursively load all child links from a root URL.
```

image

### Description

When I run mypy for the first time (if `.mypy_cache` exists, it must be deleted), I always get this warning.
I searched for this error message, and it might only appear in 3.12.

Do we need to add `r""` to that docstring? Will there be any side effects?

### System Info

langchain==0.2.16
langchain-community==0.2.16
langchain-core==0.2.38

platform==linux
python-version==3.12.5

tibor-reiss:
Fixes #26212: replaced the raw string with backslashes. Alternative: raw-stringif the full docstring.

### DIFF
--- a/libs/community/langchain_community/document_loaders/recursive_url_loader.py
+++ b/libs/community/langchain_community/document_loaders/recursive_url_loader.py
@@ -227,7 +227,7 @@ class RecursiveUrlLoader(BaseLoader):
                 "https://docs.python.org/3.9/",
                 prevent_outside=True,
                 base_url="https://docs.python.org",
-                link_regex='<a\\s+(?:[^>]*?\\s+)?href="([^"]*(?=index)[^"]*)"',
+                link_regex=r'<a\\s+(?:[^>]*?\\s+)?href="([^"]*(?=index)[^"]*)"',
                 exclude_dirs=['https://docs.python.org/3.9/faq']
             )
             docs = loader.load()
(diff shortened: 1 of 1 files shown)


efriis:
will comment in issue. These strings are identical, so won't have any impact on folks, but sounds like they might be running a very old, or a newer+buggy version of mypy

libs/community/langchain_community/document_loaders/recursive_url_loader.py
```diff
@@ -227,7 +227,7 @@ def simple_metadata_extractor(
                 "https://docs.python.org/3.9/",
                 prevent_outside=True,
                 base_url="https://docs.python.org",
-                link_regex=r'<a\s+(?:[^>]*?\s+)?href="([^"]*(?=index)[^"]*)"',
+                link_regex='<a\\s+(?:[^>]*?\\s+)?href="([^"]*(?=index)[^"]*)"',
```
efriis, line 230:
```suggestion
                link_regex=r'<a\\s+(?:[^>]*?\\s+)?href="([^"]*(?=index)[^"]*)"',
```

efriis (commented):
Ah missed that this was embedded in a docstring sorry about that! I think this is needed to keep the code block accurate (the docs will render a single backslash, which means the string needs to be defined as an r-string)

tibor-reiss:
> Ah missed that this was embedded in a docstring sorry about that! I think this is needed to keep the code block accurate (the docs will render a single backslash, which means the string needs to be defined as an r-string)

@efriis I am surprised that despite the double backslash you still need the `r` to mark it as raw string...

PR #26398, opened 2024-09-12
docs[experimental]: Make docs clearer and add min_chunk_size

### LINKED ISSUE #26171: What units to use for threshold amount?
### URL

https://python.langchain.com/v0.2/docs/how_to/semantic-chunker/

### Checklist

- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

It seems that units for threshold-type = "percentage" are out of a hundred, i.e., 85.0 not 0.85, and this is also unclear for the other threshold types, "gradient," and "interquartile." 

### Idea or request for content:

Also, Semantic Chunker really needs a min and max chunk size. I am getting chunks of a single word, and chunks that exceed the OpenAI limit. Thanks for all the great work on LangChain.

tibor-reiss:
Fixes #26171:

- added some clarification text for the keyword argument `breakpoint_threshold_amount`
- added min_chunk_size: together with `breakpoint_threshold_amount`, too small/big chunk sizes can be avoided

Note: the langchain-experimental was moved to a separate repo, so only the doc change stays here.

### DIFF
--- a/docs/docs/how_to/semantic-chunker.ipynb
+++ b/docs/docs/how_to/semantic-chunker.ipynb
@@ -125,9 +125,11 @@
     "\n",
     "There are a few ways to determine what that threshold is, which are controlled by the `breakpoint_threshold_type` kwarg.\n",
     "\n",
+    "Note: if the resulting chunk sizes are too small/big, the additional kwargs `breakpoint_threshold_amount` and `min_chunk_size` can be used for adjustments.\n",
+    "\n",
     "### Percentile\n",
     "\n",
-    "The default way to split is based on percentile. In this method, all differences between sentences are calculated, and then any difference greater than the X percentile is split."
+    "The default way to split is based on percentile. In this method, all differences between sentences are calculated, and then any difference greater than the X percentile is split. The default value for X is 95.0 and can be adjusted by the keyword argument `breakpoint_threshold_amount` which expects a number between 0.0 and 100.0."
    ]
   },
   {
@@ -186,7 +188,7 @@
    "source": [
     "### Standard Deviation\n",
     "\n",
-    "In this method, any difference greater than X standard deviations is split."
+    "In this method, any difference greater than X standard deviations is split. The default value for X is 3.0 and can be adjusted by the keyword argument `breakpoint_threshold_amount`."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


efriis (commented):
could you add some unit tests? I believe this has a bug where it will drop chunks if some of the final chunks are shorter than min_chunk_size

This is also relatively achievable as a postprocessing step, so not confident this needs to be part of this implementation in particular. 

e.g. 

```
def merge_short_docs(docs, min_chunk_size = 20):
    rtn = []
    doc = None
    for d in docs:
        if doc is None:
            doc = d
        else:
            # use first doc's metadata
            doc = Document(doc.page_content + d.page_content, metadata=doc.metadata)

        if len(doc.page_content) < min_chunk_size:
            continue
        rtn.append(doc)
        doc = None
    if doc is not None:
        rtn.append(doc)
    return rtn
```

note this allows you to configure metadata handling as well if you want to merge in a custom way.

Thoughts on documenting this instead of adding the param?

tibor-reiss:
@efriis sure, I will try to add some tests. However, since start_index is not incremented, the second if will catch this:
```
# The last group, if any sentences remain
if start_index < len(sentences):
```

tibor-reiss:
@efriis Test added. The current implementation does not miss the last sentence (or what is left) and does the same thing as your snippet just in less lines of code :)

However, I like the idea of adding a postprocessing step. On the other hand, it's not that simple to cover all edge cases - e.g. if the last chunk is smaller than `min_chunk_size`, it still remains as such. But if I would add it to the previous, then that previous chunk could grow too much, and so on. That was also the reason I did not add a `max_chunk_size`, because this can be already controlled with `breakpoint_threshold_amount`.
Let me know what you think.

tibor-reiss:
@efriis is there a way to restart vercel, please? Or how can I access the logs/build from this workflow? I am getting 404.

tibor-reiss:
Hi @efriis, as requested, I moved the code change to the other repo, leaving just the doc change here.

tibor-reiss:
Friendly ping @baskaryan @eyurtsev
The corresponding has been live in experimental for some time now

efriis (approved):

PR #26404, opened 2024-09-12
core: Cache serialized manifest for llms, chat models and prompts

nfcampos:


eyurtsev (approved):
Looks great! I think this will significantly speed up the majority of usage

PR #26414, opened 2024-09-13
support nested dicts

hwchase17:
needed for thinking tokens


nfcampos (approved):

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -335,6 +335,33 @@ def _convert_chunk_to_generation_chunk(
...
+        return new_usage + overall_token_usage
+    elif isinstance(new_usage, dict):
+        if not isinstance(overall_token_usage, dict):
+            raise ValueError(
+                f"Got different types for token usage: "
+                f"{type(new_usage)} and {type(overall_token_usage)}"
+            )
+        return {
+            k: _update_token_usage(overall_token_usage.get(k, 0), v)
+            for k, v in new_usage.items()
```
athyuttamre, line 358: I would suggest inspecting the value to ensure it's an integer. The OpenAI API may add additional fields within `usage` that are not integers in the future.

efriis: Believe this actually handles what you're asking for! Recursively calls the same function, the adding only happens if it's an int. Otherwise hits the warning "else" case at the bottom

efriis:
released in langchain-openai==0.1.24

community release will wait til tomorrow - github having some issues with actions, and there are some dependency failures in experimental that we'll debug before releasing

libs/partners/openai/langchain_openai/chat_models/base.py
arpanptnk85: Agreed with @efriis , The best way to handle this by checking the instance types programmatically for `int` and `dict`. And the `else` case will then take care of warnings.

athyuttamre: Makes sense, thanks! I don't think you need to warn in the else condition though — we may add non-integer fields in `usage` in the future, and that's not a breaking change or unexpected.

PR #26492, opened 2024-09-14
community: Update UC toolkit documentation to use LangGraph APIs, and fix pydantic error

smurching:
- **Description:** Update UC toolkit documentation to show an example of using recommended LangGraph agent APIs before the existing LangChain AgentExecutor example, and fixes a `pydantic` error message due to a missing type definition in the example notebook. Tested by manually running the updated example notebook 


libs/community/langchain_community/tools/databricks/tool.py
```diff
@@ -9,8 +9,9 @@
 from langchain_core.tools.base import BaseToolkit
 from typing_extensions import Self
 
+from databricks.sdk import WorkspaceClient
```
smurching, line 12: This seems necessary to fix a pydantic forward reference issue; without it, calling `UCFunctionToolkit.include(...)` throws with


```
PydanticUserError: UCFunctionToolkit is not fully defined; you should define WorkspaceClient, then call UCFunctionToolkit.model_rebuild().
```

docs/docs/integrations/tools/databricks.ipynb
```diff
@@ -54,11 +62,12 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 9,
    "metadata": {},
    "outputs": [],
    "source": [
     "from langchain_community.tools.databricks import UCFunctionToolkit\n",
+    "from databricks.sdk import WorkspaceClient\n",
```
smurching, line 70: We should be able to remove this import after the change in `tool.py` below is released

efriis (commented):
Hey there! I think the code change would be more relevant in the `langchain-databricks` integration here, rather than the deprecated community one:

https://github.com/langchain-ai/langchain-databricks

efriis:
Closing because this import format isn't compatible with community's optional parameters, and would welcome this class in the langchain-databricks package, and we can deprecate this one!

PR #26527, opened 2024-09-16
standard-tests[patch]: add async test for structured output

ccurme:


ccurme:
Run of scheduled tests off this branch passes: https://github.com/langchain-ai/langchain/actions/runs/10886576245

PR #26529, opened 2024-09-16
core[minor]: change how tool error handling works

isahers1:
- error handling can now be passed through decorator
- all errors thrown will be ValidationErrors (tool call was formatted incorrectly) or ToolExceptions (any error that occurs while running the tool)

### DIFF
--- a/libs/core/langchain_core/tools/base.py
+++ b/libs/core/langchain_core/tools/base.py
@@ -663,7 +663,7 @@ class ChildTool(BaseTool):
             except ToolException as e:
                 raise e
             except Exception as e:
-                raise ToolException(str(e)) from e
+                raise ToolException(repr(e)) from e
             except KeyboardInterrupt as e:
                 error_to_raise = e
                 status = "error"
@@ -786,7 +786,7 @@ class ChildTool(BaseTool):
             except ToolException as e:
                 raise e
             except Exception as e:
-                raise ToolException(str(e)) from e
+                raise ToolException(repr(e)) from e
         except ToolException as e:
             if not self.handle_tool_error:
                 error_to_raise = e
--- a/libs/core/langchain_core/tools/base.py
+++ b/libs/core/langchain_core/tools/base.py
@@ -90,26 +90,6 @@ def _get_annotation_description(arg_type: Type) -> str | None:
                 return annotation
     return None
 
-
-def _get_filtered_args(
-    inferred_model: Type[BaseModel],
-    func: Callable,
-    *,
-    filter_args: Sequence[str],
-    include_injected: bool = True,
-) -> dict:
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/tools/base.py
```diff
@@ -649,43 +649,50 @@ def run(
...
-            else:
-                content = response
-            status = "success"
-        except ValidationError as e:
-            if not self.handle_validation_error:
+                status = "error"
+            except ToolException as e:
+                raise e
+            except Exception as e:
+                raise ToolException(str(e)) from e
```
efriis, line 686: do we want to format this any differently? I believe this has impact on agent performance

libs/core/langchain_core/tools/base.py
```diff
@@ -649,43 +649,50 @@ def run(
...
+            except KeyboardInterrupt as e:
```
efriis, line 687: does this not log as an exception? Will this ever get hit after `except Exception`?

efriis (commented):
Believe this will technically be breaking for anyone doing error handling on their tools currently - could also be viewed as a fix. Will let @baskaryan comment

libs/core/langchain_core/tools/base.py
isahers1: keyboard interrupt is treated differently at least according my basic testing with this tool:
```@tool
def my_tool(x: int):
    """My tool"""
    time.sleep(100)
    return x*5```
    
I was able to run and interrupt as normal

File: libs/core/langchain_core/tools/base.py (hunk 1 above)
Comment (isahers1) [reply libs/core/langchain_core/tools/base.py:686]: I don't understand why this impacts performance - but happy to reformat to whatever just lmk what you had in mind

File: libs/core/langchain_core/tools/convert.py [hunk 3]
```diff
@@ -42,6 +48,14 @@ def tool(
         error_on_invalid_docstring: if ``parse_docstring`` is provided, configure
             whether to raise ValueError on invalid Google Style docstrings.
             Defaults to True.
+        handle_tool_error: Handle the content of the ToolException thrown. If False
+              do nothing, If True returns 'Tool execution error'. If string then
+              returns that string directly, if callable converts error to string to be returned.
+              Defaults to False.
+          handle_validation_error: Handle the content of the ValidationError thrown. If False
```
vbarda, line 55: nit - docstring misaligned

vbarda: also, i think the param description is a bit confusing (which is true for both the original description in the `BaseTool` class as well) - this is not handling the content of the error, but content of `ToolMessage` returned if the error is thrown

PR #26535, opened 2024-09-16
docs: show how to use langchain-cli for migration

eyurtsev:
Update v0.3 instructions a bit


docs/docs/versions/v0_3/index.mdx
```diff
@@ -205,3 +207,48 @@ class FooParser(BaseOutputParser):
 
 FooParser.model_rebuild()
 ```
+
+
+## Migrate using langchain-cli
```
baskaryan, line 212: should this be in the "How to update your code" section?

docs/docs/versions/v0_3/index.mdx
```diff
@@ -205,3 +207,48 @@ class FooParser(BaseOutputParser):
...
+
+The `langchain-cli` can help migrate your code to new imports automatically. As of 0.0.31, the `langchain-cli` relies on [gritql](https://about.grit.io/) for applying code mods.
+
+## Installation
+
+```bash
+pip install langchain-cli
```
baskaryan, line 219:
```suggestion
pip install -U langchain-cli
```

docs/docs/versions/v0_3/index.mdx
```diff
@@ -205,3 +207,48 @@ class FooParser(BaseOutputParser):
...
+langchain-cli --version # <-- Make sure the version is at least 0.0.31
+```
+
+## Usage
+
+Given that the migration script is not perfect, you should make sure you have a backup of your code first (e.g., using version control like `git`).
+
+You will need to run the migration script **twice** as it only applies one import replacement per run.
```
baskaryan, line 227: should we just run it twice for users?

eyurtsev: Not enough of a friction point to do this I think

docs/docs/versions/v0_3/index.mdx
eyurtsev: I think let's reference, but not move since it's just one part of how to update yoru code and doesn't handle most of the work associated with the pydantic migration

baskaryan: maybe worth making that explicit here then? that this doesn't handle pydantic migration

baskaryan (approved):

PR #26545, opened 2024-09-16
docs: fixed issue in chatbot.ipynb

markperfect:
**Description:** The tutorial would return `ValueError: SystemMessage should be the first in the history` since the prompt wasn't defined
**Issue:** N/A
**Dependencies:** N/A


efriis:
existing notebook runs for me. `prompt` is defined higher up (same with `model`)!

PR #26560, opened 2024-09-17
community: Better document Unity Catalog tools usage with Databricks chat models

smurching:
## Description

**Description:** Better documents usage of Unity Catalog tools with Databricks chat models in langchain
**Dependencies:** No new dependencies are required

## Tests
Tested manually by running the updated `.ipynb`


efriis:
I believe this should also be using stuff from langchain-ai/langchain-databricks but can discuss on the other PR too!

efriis:
Given these should work equivalently to other tools, I don't think this makes great sense to document specially here. Let's stick with the components being documented separately!

Happy to slack more, and per other pr, let's move the UC stuff into the integration package.

PR #26562, opened 2024-09-17
Fixed adding float values into DynamoDB

VpkPrasanna:

- [x] **PR title**: Add float Message into Dynamo DB
  -  community
  - Example: "community: Chat Message History 

- [x] **PR message**: 
    - **Description:** pushing float values into dynamo db creates error , solved that by converting to str type
    - **Issue:** Float values are not getting pushed
    - **Twitter handle:** VpkPrasanna
    
    
    Have added an utility function for str conversion , let me know where to place it happy to do an commit.
    
    This PR is from an discussion of #26543
    
    @hwchase17 @baskaryan @efriis


Undertone0809 (changes requested):
image

Need to format the code.

If you have any questions, feel free to ask.

VpkPrasanna:
> image
> 
> Need to format the code.
> 
> If you have any questions, feel free to ask.

How to verify it locally before i push ? and one more thing the issues shows only formatting but is there any other changes do i need to add because i have added an utility function to fix the issue and it is in the same file do i need to place it in other location or it is fine ?

VpkPrasanna:
hi @Undertone0809 i have fixed the issue , can you review it again ?

salisbbn:
Hi @VpkPrasanna  / @Undertone0809 -- my team is also affected by this issue. Any update here regarding merging this fix in?

VpkPrasanna:
> Hi @VpkPrasanna / @Undertone0809 -- my team is also affected by this issue. Any update here regarding merging this fix in?

There is no response / update from the maaintainers

efriis (commented):
hey there! sorry for the delay. Could we gate this behind an optional constructor variable to `coerce_float_to_decimal`? I don't love the `Decimal(str(item))` bit, but it's ok if it's off by default, and folks can enable it!

Would likely be managed out of the `langchain-ai/langchain-aws` repo

salisbbn:
When saving floating point values, boto3+DynamoDB doesn't support the `float` type at all though (only `Decimal` types, and will throw otherwise). So, hiding this behind an optional flag only serves to keep the current, errant behavior in place. Is there a good reason for that? 

I was also thinking about the `str` conversion as well, but then I realized it's likely there to prevent floating-point precision errors.

efriis:
for the same reason that aws doesn't add this conversion to their sdks - it's unexpected behavior for people to store `float` datatypes, and when you retrieve them, they're `Decimal` types

I agree this is useful, so I think gating behind an attribute is good. And in the event Dynamo adds this feature to their sdk or service, it won't be a breaking change for folks to turn this off!

ccurme (approved):

ccurme:
Gated behind a `coerce_float_to_decimal` parameter with default False for now. Suggest adding to [docs](https://python.langchain.com/docs/integrations/memory/aws_dynamodb/) to assist discoverability.

VpkPrasanna:
Thank you @ccurme for merging the PR finally

salisbbn:
Hi @VpkPrasanna -- have you been using this implementation? When I try to use it, I'm able to have the float values coerced and saved to DDB, but when the messages are subsequently queried from DDB and sent as message history to Bedrock, the Bedrock services throws the following error:

`Invalid type for parameter type: <class 'decimal.Decimal'>, valid types: <class 'str'>, <class 'int'>, <class 'bool'>, <class 'float'>, <class 'list'>, <class 'dict'>`

I think for this feature to be helpful, the `coerce_float_to_decimal` flag (or another flag) will also need to coercing Decimal values back to float values when messages are retrieved. What do you think?

MatheusRDG:
@salisbbn same here, you found any solution?

salisbbn:
Sorry, no, we stopped using Langchain and rolled our own lightweight orchestrator.

PR #26587, opened 2024-09-17
anthropic[patch]: fix tool call and tool res image_url handling

baskaryan:


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -194,65 +194,80 @@ def _format_messages(
...
-                else [{"type": "text", "text": message.content}]
-            )
-            # Note: Anthropic can't have invalid tool calls as presently defined,
-            # since the model already returns dicts args not JSON strings, and invalid
-            # tool calls are those with invalid JSON for args.
-            content += _lc_tool_calls_to_anthropic_tool_use_blocks(message.tool_calls)
         else:
             content = message.content
 
+        if isinstance(message, AIMessage) and message.tool_calls:
```
ccurme, line 252: (nit)
```suggestion
        # Ensure all tool_calls have a tool_use content block
        if isinstance(message, AIMessage) and message.tool_calls:
```

ccurme (approved):

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -194,65 +194,81 @@ def _format_messages(
...
+                                    for k, v in block.items()
                                     if k in ("type", "text", "cache_control")
                                 }
                             )
+                    elif block["type"] == "tool_result":
+                        tool_content = _format_messages(
+                            [HumanMessage(block["content"])]
+                        )[1][0]["content"]
+                        content.append({**block, **{"content": tool_content}})
                     else:
```
eyurtsev, line 242: would be nice to get rid of the `else` branches if at all possible (and instead have an elif with an instance check)

eyurtsev (approved):

libs/partners/anthropic/langchain_anthropic/chat_models.py
baskaryan: there's no other known types so alternative would be to raise an error on else

Issue #26588, opened 2024-09-17
Issue with callback manager for verbose output


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Following this tutorial: https://python.langchain.com/v0.1/docs/modules/agents/quick_start/#create-the-agent

![image](https://github.com/user-attachments/assets/8e789de5-8774-44cf-a6c3-1b2b6bcdf356)

keenborder786:
@eyurtsev I can take this.

eyurtsev:
identical to: https://github.com/langchain-ai/langchain/issues/26773

PR #26593, opened 2024-09-17
core[patch]: On Chain Start Fix for `Chain` Class

keenborder786:
- **Issue:** #26588


keenborder786:
@eyurtsev please review

eyurtsev (changes requested):
HI @keenborder786 thanks for the PR! this is not the correct fix. We removed the serialization representation on purpose. The tracer itself should be fixed to deal with no representation being passed.

eyurtsev:
Here's some context: 
https://github.com/langchain-ai/langchain/pull/26270
https://github.com/langchain-ai/langchain/pull/26576

eyurtsev:
@keenborder786 we'll also need a unit test to be added for the PR to be merged

keenborder786:
@eyurtsev okay understood, let me do it.

keenborder786:
@eyurtsev review please

keenborder786:
@eyurtsev I have implemented the changes

libs/langchain/langchain/chains/base.py
```diff
@@ -147,11 +147,12 @@ def invoke(
             self.metadata,
         )
         new_arg_supported = inspect.signature(self._call).parameters.get("run_manager")
-
         run_manager = callback_manager.on_chain_start(
             None,
             inputs,
             run_id,
+            class_name=self.name,
```
eyurtsev, line 154: This is a big decision! 

`on_chain_start` here is an event dispatcher. By updating the information sent here the code sets the shape of dispatched on chain start events.

i.e., we should probably not be including `class_name` and `lc_id` in here, instead everything can be placed in the `name` logic if any additional attributes are needed

libs/core/langchain_core/callbacks/stdout.py
```diff
@@ -32,8 +32,14 @@ def on_chain_start(
             inputs (Dict[str, Any]): The inputs to the chain.
             **kwargs (Any): Additional keyword arguments.
         """
-        class_name = serialized.get("name", serialized.get("id", ["<unknown>"])[-1])
-        print(f"\n\n\033[1m> Entering new {class_name} chain...\033[0m")  # noqa: T201
+        if "name" in kwargs:
```
eyurtsev, line 35: We can fetch the `name`  from kwargs here if present and otherwise fallback to serialized representation

eyurtsev (approved):

libs/langchain/langchain/chains/base.py
eyurtsev: I'm going to remove these additions since they're not necessary or desirable

eyurtsev:
Should resolve: https://github.com/langchain-ai/langchain/issues/26773

slyin87:
@eyurtsev sorry to ask, so which version should I use to have this fix in place? How do I figure out based on this PR?

PR #26599, opened 2024-09-17
openai[patch]: enable Azure structured output, parallel_tool_calls=Fa…

baskaryan:
…lse, tool_choice=required

response_format=json_schema, tool_choice=required, parallel_tool_calls are all supported for gpt-4o on azure.

### DIFF
--- a/libs/partners/openai/tests/integration_tests/chat_models/test_azure_standard.py
+++ b/libs/partners/openai/tests/integration_tests/chat_models/test_azure_standard.py
@@ -9,9 +9,6 @@ from langchain_standard_tests.integration_tests import ChatModelIntegrationTests
 
 from langchain_openai import AzureChatOpenAI
 
-OPENAI_API_VERSION = os.environ.get("AZURE_OPENAI_API_VERSION", "")
-OPENAI_API_BASE = os.environ.get("AZURE_OPENAI_API_BASE", "")
-OPENAI_API_KEY = os.environ.get("AZURE_OPENAI_API_KEY", "")
 DEPLOYMENT_NAME = os.environ.get(
     "AZURE_OPENAI_DEPLOYMENT_NAME",
     os.environ.get("AZURE_OPENAI_CHAT_DEPLOYMENT_NAME", ""),
@@ -25,12 +22,11 @@ class TestOpenAIStandard(ChatModelIntegrationTests):
 
     @property
     def chat_model_params(self) -> dict:
-        return {
-            "deployment_name": DEPLOYMENT_NAME,
-            "openai_api_version": OPENAI_API_VERSION,
-            "azure_endpoint": OPENAI_API_BASE,
-            "api_key": OPENAI_API_KEY,
-        }
+        return {"deployment_name": DEPLOYMENT_NAME}
+
+    @property
+    def supports_image_inputs(self) -> bool:
+        return True
 
     @pytest.mark.xfail(reason="Not yet supported.")
     def test_usage_metadata_streaming(self, model: BaseChatModel) -> None:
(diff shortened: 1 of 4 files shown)


Skar0:
Hi @baskaryan, thank you for addressing this issue and for your work on this! I tested the changes, and they work as expected for the `json_schema` part 😃

I noticed that in your fix, `AzureChatOpenAI._create_chat_result` now passes an `openai.BaseModel` to `BaseChatOpenAI._create_chat_result` instead of the `dict` resulting from `response.model_dump()`, as it did before. While this does resolve the issue for which I proposed a fix in #26086, it requires both `BaseChatOpenAI._create_chat_result` and `AzureChatOpenAI._create_chat_result` to transform the response from an `openai.BaseModel` into a `dict`.

In #26086, I proposed modifying `BaseChatOpenAI._create_chat_result` to consistently work with the `dict` version of the response, which is always created anyways. This change eliminates the need to convert the response into a `dict` multiple times when calling `AzureChatOpenAI._create_chat_result` and I believe it simplifies the logic and complexity within these methods. I believe that this would complement your changes and improve code clarity. Of course, this is assuming that my proposed changes are correct and align with your goals for the implementation of that method.

Thanks again for your work and for considering my contribution! 😉

libs/partners/openai/tests/integration_tests/chat_models/test_azure_standard.py
```diff
@@ -9,27 +9,39 @@
 
 from langchain_openai import AzureChatOpenAI
 
-OPENAI_API_VERSION = os.environ.get("AZURE_OPENAI_API_VERSION", "")
```
ccurme, line 12: Ran OpenAI integration tests on this branch and got failures related to env vars: https://github.com/langchain-ai/langchain/actions/runs/10924401318/job/30323353395

Also getting some failures here in json mode on master, suspect unrelated. Unable to reproduce so far.
```
FAILED tests/integration_tests/chat_models/test_azure.py::test_json_mode_async - openai.LengthFinishReasonError: Could not parse response content as the length limit was reached - CompletionUsage(completion_tokens=10, prompt_tokens=18, total_tokens=28, completion_tokens_details=None)
```

baskaryan:
> Hi @baskaryan, thank you for addressing this issue and for your work on this! I tested the changes, and they work as expected for the `json_schema` part 😃
> 
> I noticed that in your fix, `AzureChatOpenAI._create_chat_result` now passes an `openai.BaseModel` to `BaseChatOpenAI._create_chat_result` instead of the `dict` resulting from `response.model_dump()`, as it did before. While this does resolve the issue for which I proposed a fix in #26086, it requires both `BaseChatOpenAI._create_chat_result` and `AzureChatOpenAI._create_chat_result` to transform the response from an `openai.BaseModel` into a `dict`.
> 
> In #26086, I proposed modifying `BaseChatOpenAI._create_chat_result` to consistently work with the `dict` version of the response, which is always created anyways. This change eliminates the need to convert the response into a `dict` multiple times when calling `AzureChatOpenAI._create_chat_result` and I believe it simplifies the logic and complexity within these methods. I believe that this would complement your changes and improve code clarity. Of course, this is assuming that my proposed changes are correct and align with your goals for the implementation of that method.
> 
> Thanks again for your work and for considering my contribution! 😉

the issue is that we want to extract `response.choices[0].message.parsed` before response is converted to a dict, as parsed can itself be a pydantic object and we don't want it converted to a dict

libs/partners/openai/tests/integration_tests/chat_models/test_azure_standard.py
baskaryan: huh thought id set that env var, will double check

PR #26601, opened 2024-09-18
community: arxiv and pubmed tools description fix

ivanmilevtues:
## Tool description updated for Langchain Community ArXiv and PubMed tool

- **Description**: this PR contains changes in the description of the ArXiv and PubMed tool so that the functionality and usage explanations is better aligned for the LLM.
- **Issue.**: The issue with the descriptions is that they are incomplete. **PubMed** tool: the description does not provide the input grammer for the query parameter, so a lot of the functionality is not available to the end user. **ArXiV** tool: the description in this case does not provide which day today is which confuses the model for the relevance of papers.
- **Dependencies**: no new dependencies are required for this change
- **Twitter handle**: @ivan_milev21

**Tests** - As this kind of bug fix is not really unit testable, I will provide one example for each tool. The examples show the erroneous tool calls. The bug here is that when the description is not correct the LLM would be unable to provide correct query parameter. 

For **ArXiv**:
Query: 'Can you provide a summary of the recent findings in recent findings in Physics?'
Before description change:
  tool call arguments:  {"query":"Physics"}
  answer: 'Published: 2008-11-21\nTitle: Review on recent developments...'
After description change:
  tool call arguments: {"query":"Physics 2024"}
  answer: 'Published: 2024-05-14\nTitle: Status of Dark Photons...'

For **PubMed**:
Query: Can you list some papers for cancer from 2020 using RNA technology?
Before description change:
  tool call arguments: {"query":"cancer RNA technology 2020"}
  answer: ''Published: 2024-09-15\nTitle: A novel small molecule ZYZ384...'
After description change:
  tool call arguments: {"query":"cancer AND RNA technology AND (\"2020/01/01\"[Date - Publication] : \"2020/12/31\"[Date - Publication])"}
  answer: ''Published: 2020-11-26\nTitle: TUG1 long non-coding RNA...'

Additional guidelines:


libs/community/langchain_community/tools/arxiv/tool.py
```diff
@@ -25,7 +26,8 @@ class ArxivQueryRun(BaseTool):
         "Computer Science, Quantitative Biology, Quantitative Finance, Statistics, "
         "Electrical Engineering, and Economics "
         "from scientific articles on arxiv.org. "
-        "Input should be a search query."
+        "Input should be a search query. If there is something related to date please use date format and not words! "
+        f"Today's date is {datetime.today().strftime('%Y-%m-%d')}."
```
eyurtsev, line 30: This doesn't work on any production server that runs for more than a single day

eyurtsev (changes requested):

eyurtsev:
Users should probably sub-class and write an appropriate description for their use case or create the tool dynamically as needed

libs/community/langchain_community/tools/arxiv/tool.py
ivanmilevtues: Thanks for the quick response. You are right so for now I am reverting my changes.

On another note the input query grammar for the tool is not specified, but this I will leave for now and maybe create a separate pull request in future.

ivanmilevtues:
Thanks for the quick response @eyurtsev. The comment you made for sub-classing is valid if the user wants to tailor the tool usage for his concrete needs.

However my change in the PubMed tool allows people to utilize the tool in the general case. The PubMed tool works with the PubMed's APi which expects concrete syntax in order to query the Database, if you just give it free form text it would only search in the paper's title. Without this change people can use the tool only to query based on the title, which again is not described in the tool description.

I believe that my fix allows the users to utilize the tool's functionality while now they can query only by one field.

PR #26608, opened 2024-09-18
core: Put Python version as a project requirement so it is considered by ruff

cbornet:
Ruff doesn't know about the python version in `[tool.poetry.dependencies]`. It can get it from `project.requires-python`.

Notes:
* poetry seems to have issues getting the python constraints from `requires-python` and using `python` in per dependency constraints. So I had to duplicate the info. I will open an issue on poetry.
* `inspect.isclass()` doesn't work correctly with `GenericAlias` (`list[...]`, `dict[..., ...]`) on Python <3.11 so I added some `not isinstance(type, GenericAlias)` checks:

Python 3.11
```pycon
>>> import inspect
>>> inspect.isclass(list)
True
>>> inspect.isclass(list[str])
False
```

Python 3.9
```pycon
>>> import inspect
>>> inspect.isclass(list)
True
>>> inspect.isclass(list[str])
True
```

### DIFF
--- a/libs/core/langchain_core/_api/beta_decorator.py
+++ b/libs/core/langchain_core/_api/beta_decorator.py
@@ -14,7 +14,8 @@ import contextlib
 import functools
 import inspect
 import warnings
-from typing import Any, Callable, Generator, Type, TypeVar, Union, cast
+from collections.abc import Generator
+from typing import Any, Callable, TypeVar, Union, cast
 
 from langchain_core._api.internal import is_caller_internal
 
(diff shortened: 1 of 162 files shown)


eyurtsev (approved):

efriis:
reverting the pyproject change because core isn't installable by `uv` anymore as a result

I think if we completely switch metadata over to project from tool.poetry that will help, but docs build currently can't run

I think extended tests pass because none of the uv install steps actually install core - they just affect envs that have core already installed from poetry

PR #26625, opened 2024-09-18
core: Deprecate BaseMemory

prcastro:
**Description**
Deprecate BaseMemory, because:
* LangChain developers don't recommend using it
* No other parts of the core code use it
* There is no documentation on how to use it (apart from a vague description on the docstring)
* It doesn't seem to integrate well with LCEL
* There are no tests covering it

I wonder if I can deprecate the whole `langchain.memory` package as well. If not, we can move this to `langchain.memory` and deprecate it in core.

**Issues**
* https://github.com/langchain-ai/langchain/discussions/25904#discussion-7114933
* https://www.reddit.com/r/LangChain/comments/1fju4sp/chat_history_vs_memory_whats_the_difference/

### DIFF
--- a/libs/core/langchain_core/memory.py
+++ b/libs/core/langchain_core/memory.py
@@ -19,7 +19,12 @@ from langchain_core._api import deprecated
 from langchain_core.load.serializable import Serializable
 from langchain_core.runnables import run_in_executor
 
-@deprecated("0.3.0", alternative="langchain_core.chat_history.BaseChatMessageHistory", removal="1.0")
+
+@deprecated(
+    "0.3.0",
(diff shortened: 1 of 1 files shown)


eyurtsev (changes requested):
Hi @prcastro thanks for the PR. 

Deprecations are usually done with appropriate backing documentation. We're in the process of writing it and will then make appropriate PRs to deprecate.

There's still use code that relies on these and functionality typically cannot be deprecated without a migration strategy for existing code.

eyurtsev:
We'll do the deprecation once the migration guides are up (likely beginning of next week)

PR #26633, opened 2024-09-18
core: remove [project] tag from pyproject

efriis:
makes core incompatible with uv installs


eyurtsev:
cc @cbornet the tag was breaking docs build with uv

efriis:
comment in #26608

efriis:
looks like recommendation is to use `tool.ruff` target-version! https://docs.astral.sh/ruff/settings/#target-version

PR #26634, opened 2024-09-18
core: use ruff.target-version instead

efriis:
tested on one of the replacement cases and seems to work! 
ScreenShot 2024-09-18 at 02 02 43PM


efriis:
cc @cbornet @eyurtsev

efriis:
related to #26633 #26608

cbornet:
👀 too bad we can’t have all tools work correctly with the project section 🤷‍♂️

PR #26649, opened 2024-09-19
[community] Fix WorkspaceClient error with pydantic validation

serena-ruan:

Fix error like
image

After the fix:
image

Additional guidelines:


efriis (commented):
I'm pretty sure we don't want to introduce this pattern into community.

Let's change the type definitions to `Any` on the WorkspaceClient functions/attributes.

You can strictly type it in the partner package instead!

efriis (approved):

PR #26653, opened 2024-09-19
community: refactor opensearch query constructor to use wildcard instead of match in the contain comparator

quintinoandre:
- **Description:** Changed the comparator to use a wildcard query instead of match. This modification allows for partial text matching on analyzed fields, which improves the flexibility of the search by performing full-text searches that aren't limited to exact matches.
- **Issue:** The previous implementation used a match query, which performs exact matches on analyzed fields. This approach limited the search capabilities by requiring the query terms to align with the indexed text. The modification to use a wildcard query instead addresses this limitation. The wildcard query allows for partial text matching, which means the search can return results even if only a portion of the term matches the text. This makes the search more flexible and suitable for use cases where exact matches aren't necessary or expected, enabling broader full-text searches across analyzed fields.
In short, the problem was that match queries were too restrictive, and the change to wildcard queries enhances the ability to perform partial matches.
- **Dependencies:** none
- **Twitter handle:** @Andre_Q_Pereira


quintinoandre:
@eyurtsev can you review this merge request please. Thank you.

quintinoandre:
@eyurtsev eyurtsev or @ccurme ccurme can you review this merge request please. Thank you.

ccurme (approved):
Arguably a breaking change but IMO this is a bug fix given the operator is `CONTAIN`.

PR #26660, opened 2024-09-19
core: Improve support for id in VectorStore

cbornet:


libs/core/langchain_core/vectorstores/base.py
```diff
@@ -61,17 +61,16 @@ def add_texts(
         self,
         texts: Iterable[str],
         metadatas: Optional[list[dict]] = None,
-        # One of the kwargs should be `ids` which is a list of ids
-        # associated with the texts.
-        # This is not yet enforced in the type signature for backwards compatibility
-        # with existing implementations.
+        *,
+        ids: Optional[Iterable[str]] = None,
```
cbornet, line 65: `ids` is kw-only and optional so it doesn't break compatibility

libs/core/langchain_core/vectorstores/base.py
```diff
@@ -827,6 +836,15 @@ def from_documents(
         """
         texts = [d.page_content for d in documents]
         metadatas = [d.metadata for d in documents]
+
+        if "ids" not in kwargs:
```
cbornet, line 840: We use the same logic in `from_documents` as in `add_documents`

libs/core/langchain_core/vectorstores/base.py
```diff
@@ -61,17 +61,16 @@ def add_texts(
         self,
         texts: Iterable[str],
         metadatas: Optional[list[dict]] = None,
-        # One of the kwargs should be `ids` which is a list of ids
-        # associated with the texts.
-        # This is not yet enforced in the type signature for backwards compatibility
-        # with existing implementations.
+        *,
+        ids: Optional[list[str]] = None,
```
cbornet, line 65: `ids` is kw-only and optional so it only breaks compatibility if implementations have an `ids` param which is not compatible with `Optional[list[str]]`.
Making it positional would break compatibility and can be considered later.
Is it acceptable ?

libs/core/langchain_core/vectorstores/base.py
cbornet, line 840: Should we add deprecation warnings that `ids` should be passed in `Document` and not as a param ?

eyurtsev: The optional id in the document is optional now and not all implementations honor them. (if I had to guess most don't yet)

So far I've been interpreting these as precedence rules -- ids specified as part of the function signature take precedence over ids specified as part of the document.

libs/core/langchain_core/vectorstores/base.py
eyurtsev: Yes makes a ton of sense, probably Optional[Sequence[str]].

I was going to write a code mod at some point to update this across the entire code base to make it explicit

eyurtsev:
@cbornet sorry i'm falling a bit behind will try to catch up on review tomorrow and will finish reviewing this PR

libs/core/langchain_core/vectorstores/base.py
cbornet: Well, ideally, this would be `Optional[Iterable[str]]` but a lot of implementations have already used `Optional[list[str]]` so both `Optional[Iterable[str]]` and `Optional[Sequence[str]]` are breaking to them.
So we would have to update them to support `Optional[Iterable[str]]`. Do you prefer that we do this update ?

eyurtsev: iterable forces the implementation to create pagination logic which I don't think will work well for the given abstraction. It forces the implementations to paginate which means that they will likely all paginate differently and set bad default batch sizes. Lots of opportunities for silent failures on the ingestion path

cbornet: I’m not sure to follow.
My opinion would be that all inputs should be Iterables (texts, metadata and ids) so that you can zip them without materializing.
Currently the fact that metadata is a list is a problem imo as you can’t load big collections and you have to do batching on the user side.
If implementations don’t want to bother with zipping they could always call `metadatas=list(metadatas)` and be back to using materialized lists.

cbornet: Or you mean it’s intentional to force the user-side to materialize and do the batching ?

eyurtsev: I don't know what the original intention was of the current interface, but it's definitely problematic since it's not self consistent (metadata vs. texts).

> Or you mean it’s intentional to force the user-side to materialize and do the batching ?

I think it's **better** to have the batching on the user side b/c we're dealing with so many different implementations.

The concern is that it's another place where implementations can mess something up:
1. Incorrect batching logic (e.g., zip short-circuits and will fail silently if inputs are of different lengths)
2. Abstract interface does not have a batch_size parameter (so batch size control would be inconsistent or missing)
3. Implementations may attempt to guess the batch_size which they can't do correctly:
-- batch size considerations depend on size of average document + on what batch is optimal for embeddings 

If we had just a few implementations I think iterables would be great, but given the amount I'd prefer that we left pagination is something done on user side to reduce the likelihood of bugs

libs/core/langchain_core/vectorstores/base.py
```diff
@@ -848,6 +866,15 @@ async def afrom_documents(
         """
         texts = [d.page_content for d in documents]
         metadatas = [d.metadata for d in documents]
+
+        if "ids" not in kwargs:
```
eyurtsev, line 870: nice thank you!

libs/core/langchain_core/vectorstores/base.py
```diff
@@ -99,10 +98,14 @@ def add_texts(
...
+            ids_: Iterator[Optional[str]] = iter(ids) if ids else cycle([None])
             docs = [
-                Document(page_content=text, metadata=metadata_)
-                for text, metadata_ in zip(texts, metadatas_)
+                Document(id=id_, page_content=text, metadata=metadata_)
+                for text, metadata_, id_ in zip(texts, metadatas_, ids_)
             ]
+            if ids is not None:
+                # For backward compatibility
+                kwargs["ids"] = ids
```
eyurtsev, line 108: Any chance that the `ids` was passed as an `iter` in any existing code? If so, we should probably pick ids from the `Document` object?

eyurtsev (approved):
I think this looks good -- we may need a defensive way of handling any existing code that's passing ids as an iterable?

libs/core/langchain_core/vectorstores/base.py
cbornet: Yes I agree that batching is hard to get right. I will still regret that it's not possible to handle it in the vectorstore.

> Incorrect batching logic (e.g., zip short-circuits and will fail silently if inputs are of different lengths)

Python 3.11 has a strict param for zip. and we could have helpers for a `strict_zip` function.

> Abstract interface does not have a batch_size parameter (so batch size control would be inconsistent or missing)

Yes batch_size would have to be detected by the impl.

> Implementations may attempt to guess the batch_size which they can't do correctly:
-- batch size considerations depend on size of average document + on what batch is optimal for embeddings

I tink it would be possible to detect in the impl.
Ideally the Embeddings would handle the batching for their own usage (with an appropriate API that accepts an iterable an returns an iterator). And batching for the persistence could be completely separated from the one for the Embeddings and be optimized for the storage.

Anyway, metadatas is already a list and this PR does the same for ids. So it's just for the sake of discussion 😄

libs/core/langchain_core/vectorstores/base.py
cbornet: If `ids` was an `iter` in existing code, it means that `add_documents` accepts it. So it shouldn't cause any problem ?

cbornet: If `ids` was passed as an `iter` in existing code, it means that `add_documents` accepts it. So it shouldn't cause any problem ?
We are really just reproducing the previous behavior of having `ids` in the kwargs, whatever its value or its type.

eyurtsev: (last question) can merge if OK

does the code not exhaust the iterable here:

```python
            docs = [
                Document(id=id_, page_content=text, metadata=metadata_)
                for text, metadata_, id_ in zip(texts, metadatas_, ids_)
            ]


            if ids is not None:
                # For backward compatibility
                kwargs["ids"] = ids # <-- is it not exhaustted?
```

eyurtsev: OOOH ids_ and id we're all good

cbornet: Good catch! This would indeed be a problem. For typing too.
I checked all implementations and actually none that implements `add_documents` uses ids as an iter.
So I believe it's fine ?

eyurtsev (approved):

PR #26670, opened 2024-09-19
unstructured[patch]: support loading URLs

ccurme:
`unstructured.partition.auto.partition` supports a `url` kwarg, but `url` in `UnstructuredLoader.__init__` is reserved for the server URL. Here we add a `web_url` kwarg that is passed to the partition kwargs:
```python
self.unstructured_kwargs["url"] = web_url
```


Coniferish:
LGTM. Only comment would be to potentially add an example to `docs/docs/integrations/document_loaders/unstructured_file.ipynb`, but I don't think that's a blocker necessarily.

ccurme:
> LGTM. Only comment would be to potentially add an example to `docs/docs/integrations/document_loaders/unstructured_file.ipynb`, but I don't think that's a blocker necessarily.

Thanks, added!

PR #26671, opened 2024-09-19
all: Remove special numpy resolution for Python 3.8

cbornet:


efriis (commented):
Not confident we want this because it will break langchain+community compatibility for non 3.12 users that have stricter than desired numpy requirements (under 1.26). Agree it's cleaner though!

Is having this special case creating issues?

efriis:
going to close this for now. This also decreases compatibility for 3.9-3.11 users!

cbornet:
🤔 I don’t see why users would have constraints to not use numpy 1.26 if they accept 1.24…

PR #26672, opened 2024-09-19
core: Simplify pydantic version constraints

cbornet:
Or is there an interest in keeping these constraints ?


efriis (commented):
yep we're keeping 2.5/2.6 compatibility for all other than 3.12 intentionally!

cbornet:
Can I ask why ?
I don’t see why users would have constraints to not use 2.7 if they accept 2.5…

PR #26689, opened 2024-09-20
Avoid copying runs

hinthornw:
Also, re-unify run trees. Use a single shared client. Etc.


libs/community/langchain_community/tools/zenguard/tool.py
```diff
@@ -91,7 +91,7 @@ def _run(
             response = requests.post(
                 self._ZENGUARD_API_URL_ROOT + postfix,
                 json=json,
-                headers={"x-api-key": self.zenguard_api_key},
+                headers={"x-api-key": cast(str, self.zenguard_api_key)},
```
hinthornw, line 94: IDK why mypy was flagging these completely unrelated files.

libs/core/langchain_core/tracers/langchain.py
```diff
@@ -112,6 +109,11 @@ def __init__(
         self.tags = tags or []
         self.latest_run: Optional[Run] = None
 
+    def _start_trace(self, run: Run) -> None:
+        super()._start_trace(run)
+        if run._client is None:
+            run._client = self.client
```
hinthornw, line 115: This is so that if you were defining the client solely via a LangChainTracer we will hand off that client to the nested run tree

libs/community/langchain_community/tools/zenguard/tool.py
eyurtsev: the cast is wrong since zenguard_api_key can be a None here -- is this catching it on CI or only on your local machine?

libs/community/pyproject.toml
```diff
@@ -42,7 +42,7 @@ aiohttp = "^3.8.3"
 tenacity = "^8.1.0,!=8.4.0"
 dataclasses-json = ">= 0.5.7, < 0.7"
 pydantic-settings = "^2.4.0"
-langsmith = "^0.1.112"
+langsmith = "^0.1.125"
```
eyurtsev, line 45: why is a bump in community necessary?

libs/core/langchain_core/runnables/config.py
```diff
@@ -131,17 +123,27 @@ def _set_config_context(config: RunnableConfig) -> None:
...
 
     var_child_runnable_config.set(config)
-    if hasattr(RunTree, "from_runnable_config"):
-        # import _set_tracing_context, get_tracing_context
-        rt = RunTree.from_runnable_config(dict(config))
-        tc = run_helpers.get_tracing_context()
-        run_helpers._set_tracing_context({**tc, "parent": rt})
+    if (
+        (cb := config.get("callbacks"))
+        and (prid := getattr(cb, "parent_run_id", None))  # Is callback manager
```
eyurtsev, line 131: I'd recommend keeping the same variable name as the name that we're looking up to keep the mental burden low.

`prid` is faster to read, but harder to read since now i have to think about what this variable name means

prid -> parent_id

t -> handler 

etc

libs/community/langchain_community/tools/zenguard/tool.py
hinthornw: CI started flagging these files despite me not modifying them

libs/core/langchain_core/tracers/langchain.py
```diff
@@ -49,17 +50,13 @@ def log_error_once(method: str, exception: Exception) -> None:
 
 def wait_for_all_tracers() -> None:
     """Wait for all tracers to finish."""
-    global _CLIENT
-    if _CLIENT is not None and _CLIENT.tracing_queue is not None:
-        _CLIENT.tracing_queue.join()
+    if rt._CLIENT is not None and rt._CLIENT.tracing_queue is not None:
```
eyurtsev, line 53: oh nice

libs/core/langchain_core/tracers/langchain.py
eyurtsev: any situations when one might create two tracers with different clients and pass them as callbacks?

eyurtsev (approved):

libs/core/langchain_core/tracers/langchain.py
eyurtsev: I assume no easy way to catch that as an issue to raise an error?

hinthornw: hmm I kinda would prefer if we still traced something / didn't raise an error that could potentially interrupt their execution flow? It's kinda hard to catch correctly

PR #26712, opened 2024-09-20
ci: Use output-format=github for ruff in CI

cbornet:


efriis:
does it not already do this?

efriis:
looks like it already does this unless I'm mistaken: https://github.com/langchain-ai/langchain/pull/26725/files

cbornet:
It does ! Good catch !

PR #26717, opened 2024-09-20
core[patch],langchain[patch]: deprecate memory and entity abstractions and implementations

eyurtsev:
This PR deprecates the old memory, entity abstractions and implementations


eyurtsev:
Following up on https://github.com/langchain-ai/langchain/pull/26625

cc @prcastro

PR #26742, opened 2024-09-21
community: Remove backticks from query executed by SQL Database Agent if they are there

iboates:
**Description:** Fix a really trivial but wasteful problem with the SQL query generation refinement loop

**Issue:** Some LLMs have a really hard time with NOT putting markdown-style backticks around code blocks, no matter how hard you try to prompt it out of them. With ChatGPT 4o I was seeing it frequently get stuck in very long loops because of this, wasting precious tokens and also degrading the result as it hopelessly shuffled around the actual SQL (which was often actually correct!

This is really trivial to fix, just check the first and last few characters for the markdown backtick structure and get rid of it if it is there.

**Dependencies:** None

I don't have a twitter handle but a link to my Github would be nice!

### DIFF
--- a/libs/community/langchain_community/utilities/sql_database.py
+++ b/libs/community/langchain_community/utilities/sql_database.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from logging import getLogger
 from typing import Any, Dict, Iterable, List, Literal, Optional, Sequence, Union
 
 import sqlalchemy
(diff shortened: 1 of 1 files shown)


iboates:
@baskaryan @efriis @eyurtsev @ccurme @vbarda @hwchase17 it said to ping you if nobody reviews the PR in a few days

eyurtsev (changes requested):
This utility knows nothing about LLMs. I don't think it's reasonable to have any clean up code here for incorrect commands.

iboates:
> This utility knows nothing about LLMs. I don't think it's reasonable to have any clean up code here for incorrect commands.

Do you have a suggestion on where it would be better implemented?

iboates:
@eyurtsev This issue is still causing me quite a lot of grief, the only way I have been able to deal with it effectively is by using my own fork with the fix applied. I am really quite motivated to fix this more permanently, but I am not sure where would be a more appropriate place. The reason I put it where I did is because it is the "last stop" before anything gets executed on the database, so I can be guaranteed that it will be applied.

If you could tell me where would be a more appropriate place to put this, I am happy to do so.

PR #26752, opened 2024-09-22
Partners: voyageai - introducing the new voyageai models

fzowl:
**Description:**
Introducing new VoyageAI models


libs/partners/voyageai/langchain_voyageai/embeddings.py
```diff
@@ -55,7 +55,15 @@ def default_values(cls, values: dict) -> Any:
...
-            values["batch_size"] = 72 if model in ["voyage-2", "voyage-02"] else 7
+            values["batch_size"] = (
+                72
+                if model in ["voyage-2", "voyage-02"]
+                else (
+                    30
+                    if model == "voyage-3-lite"
+                    else (10 if model == "voyage-3" else 7)
+                )
+            )
```
efriis, line 66: any chance this is set in the sdk as well? if so can we just make this `None` and rely on the sdk for defaults?

efriis (commented):
looks good! one question

libs/partners/voyageai/langchain_voyageai/embeddings.py
fzowl: @efriis Our SDK is thin and it is just a wrapper around the API. We let our clients to do the batching and do the batching in the integrations - wherever it is supported.

efriis:
merging from #26907 because don't have write acccess to branch

PR #26767, opened 2024-09-23
langchain_chroma[patch]: updated example for get documents with where clause

RidhamGolakiya:
Example updated for vectorstore ChromaDB.

If we want to apply multiple filters then ChromaDB supports filters like this:
Reference: [ChromaDB filters](https://cookbook.chromadb.dev/core/filters/)

Thank you.


libs/partners/chroma/langchain_chroma/vectorstores.py
```diff
@@ -941,7 +941,7 @@ def get(
         Args:
             ids: The ids of the embeddings to get. Optional.
             where: A Where type dict used to filter results by.
-                   E.g. `{"color" : "red", "price": 4.20}`. Optional.
+                   E.g. `{"$and": [{"color": "red"}, {"price": 4.20}]}` Optional.
```
eyurtsev, line 944: Does this not work?

```
`{"color" : "red", "price": 4.20}`.
```

It's supported by a lot of integrations to mean `$and`

ridhaminfyom: No this is not works with chromaDB. If you want to compare multiple argument then you compulsory need to use "$and".

eyurtsev (approved):

PR #26769, opened 2024-09-23
Enhance BM25SparseEmbedding initialization

FouadLbakali:
- [ ] **PR title**: "Enhance BM25SparseEmbedding initialization"

- [ ] **PR message**:
    - **Description:** This update enhances the `BM25SparseEmbedding` class by allowing the initialization to optionally load a model from a specified path, in addition to fitting it with a provided corpus. This makes the embedding model more flexible for different use cases.
    - **Issue:** Not applicable

### DIFF
--- a/libs/partners/milvus/langchain_milvus/utils/sparse.py
+++ b/libs/partners/milvus/langchain_milvus/utils/sparse.py
@@ -53,3 +53,9 @@ class BM25SparseEmbedding(BaseSparseEmbedding):
         for col_index, value in zip(col_indices, non_zero_values):
             result_dict[col_index] = value
         return result_dict
+
+    def load(self, path: str):
+        self.bm25_ef.load(path)
+
+    def save(self, path: str):
+        self.bm25_ef.save(path)
--- a/libs/partners/milvus/langchain_milvus/utils/sparse.py
+++ b/libs/partners/milvus/langchain_milvus/utils/sparse.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 from scipy.sparse import csr_array  # type: ignore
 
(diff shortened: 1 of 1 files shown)


libs/partners/milvus/langchain_milvus/utils/sparse.py
```diff
@@ -29,15 +29,18 @@ class BM25SparseEmbedding(BaseSparseEmbedding):
     https://milvus.io/docs/embed-with-bm25.md
     """
 
-    def __init__(self, corpus: List[str], language: str = "en"):
+    def __init__(self, corpus: List[str] = None, path: str = None, language: str = "en"):
```
eyurtsev, line 32: This is a breaking change use `*` to make and force it to be a last argument so that it's invoked by name.

eyurtsev, line 32: List type needs to be updated to an optional

libs/partners/milvus/langchain_milvus/utils/sparse.py
```diff
@@ -29,15 +29,18 @@ class BM25SparseEmbedding(BaseSparseEmbedding):
...
             build_default_analyzer,
         )
 
         self.analyzer = build_default_analyzer(language=language)
         self.bm25_ef = BM25EmbeddingFunction(self.analyzer, num_workers=1)
-        self.bm25_ef.fit(corpus)
+        if corpus is not None:
+            self.bm25_ef.fit(corpus)
+        elif path is not None:
+            self.bm25_ef.load(path)
```
eyurtsev, line 43: Need to validate what happens if neither is provided or if both are provided

eyurtsev (changes requested):
A few different bugs in this PR, left comments

efriis:
You can find the integration package you would want to edit here: https://github.com/langchain-ai/langchain-milvus

sorry for the inconvenience!

PR #26775, opened 2024-09-23
Update neo4j vector procedures

tomasonjo:


libs/community/langchain_community/vectorstores/neo4j_vector.py
```diff
@@ -753,18 +765,14 @@ def create_new_index(self) -> None:
         to create a new vector index in Neo4j.
         """
         index_query = (
-            "CALL db.index.vector.createNodeIndex("
-            "$index_name,"
-            "$node_label,"
-            "$embedding_node_property,"
-            "toInteger($embedding_dimension),"
-            "$similarity_metric )"
+            f"CREATE VECTOR INDEX {self.index_name} "
```
jexp, line 768: we should add `IF NOT EXISTS`

jexp:
should we do for rel-index too?

libs/community/langchain_community/vectorstores/neo4j_vector.py
tomasonjo: I deliberately didnt add it as I want an error here if it already exists as otherwise there could be sth wrong upstream and we wouldnt know

tomasonjo: added

tomasonjo:
No, we don't support importing data for relationship vector index, so we don't need it.

libs/community/langchain_community/vectorstores/neo4j_vector.py
```diff
@@ -753,18 +765,14 @@ def create_new_index(self) -> None:
...
         """
         index_query = (
-            "CALL db.index.vector.createNodeIndex("
-            "$index_name,"
-            "$node_label,"
-            "$embedding_node_property,"
-            "toInteger($embedding_dimension),"
-            "$similarity_metric )"
+            f"CREATE VECTOR INDEX {self.index_name} IF NOT EXISTS "
+            f"FOR (m:{self.node_label}) ON m.`{self.embedding_node_property}` "
```
jexp, line 769: perhaps backticks for node-label tool.

jexp (approved):
LGTM

efriis (approved):

PR #26783, opened 2024-09-23
experimental[major]: CVE-2024-46946 fix

mercyspirit:
Description: Resolve CVE-2024-46946 by switching out sympify with parse_expr with a very specific allowed set of operations.

https://nvd.nist.gov/vuln/detail/cve-2024-46946

Sympify uses eval which makes it vulnerable to code execution. parse_expr is limited to specific expressions.

Bandit results
image


libs/experimental/langchain_experimental/llm_symbolic_math/base.py
```diff
@@ -66,7 +66,52 @@ def _evaluate_expression(self, expression: str) -> str:
...
+                "cosh": sympy.cosh,
+                "tanh": sympy.tanh,
+                "asinh": sympy.asinh,
+                "acosh": sympy.acosh,
+                "atanh": sympy.atanh,
+                # Exponentials and logarithms
+                "exp": sympy.exp,
+                "log": sympy.log,
+                "ln": sympy.log,  # natural log (alias)
+                "log10": sympy.log,  # log base 10 (use sympy.log)
```
eyurtsev, line 91: both log10 and ln are aliased to sympy.log

eyurtsev (commented):
This looks great! Would you be able to check the issue with log10 vs. ln?

libs/experimental/langchain_experimental/llm_symbolic_math/base.py
mercyspirit: changed so that log would be a user defined base (default 10), ln will be regular sympy log which defaults to ln. And log 10 to base 10

libs/experimental/langchain_experimental/llm_symbolic_math/base.py
```diff
@@ -66,7 +66,52 @@ def _evaluate_expression(self, expression: str) -> str:
...
+                # Hyperbolic functions
+                "sinh": sympy.sinh,
+                "cosh": sympy.cosh,
+                "tanh": sympy.tanh,
+                "asinh": sympy.asinh,
+                "acosh": sympy.acosh,
+                "atanh": sympy.atanh,
+                # Exponentials and logarithms
+                "exp": sympy.exp,
+                "log": lambda x, base=10: sympy.log(x, base),
```
eyurtsev, line 89: This should honor the default with `log` being `ln` rather than `log10`

```suggestion
                "log": sympy.log,
```

eyurtsev (approved):

eyurtsev:
@mercyspirit I'm going to make the fix backwards compatible for users that are using this code for prototyping since this fix will break a lot of functionality otherwise.

dzubke:
This PR resolves this discussion: https://github.com/langchain-ai/langchain/discussions/26720.

PR #26790, opened 2024-09-23
docs[patch]: update chatbot memory how-to

vbarda:


eyurtsev:
@vbarda I think we need the same link at the top of the how-to guide that @ccurme added to his tutorial that explains that the guide was previously written using RunnableWithMessageHistory, and links to some explanation about why we're changing

we can punt on the explanation for now so we can do in one place

vbarda:
> @vbarda I think we need the same link at the top of the how-to guide that @ccurme added to his tutorial that explains that the guide was previously written using RunnableWithMessageHistory, and links to some explanation about why we're changing
> 
> we can punt on the explanation for now so we can do in one place

yup, i will add!

eyurtsev (approved):

PR #26793, opened 2024-09-23
docs: update trim messages notebook

eyurtsev:
Update trim messages notebook to include common use cases and explain what the desired behavior is

### DIFF
--- a/docs/docs/how_to/trim_messages.ipynb
+++ b/docs/docs/how_to/trim_messages.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "b5ee5b75-6876-4d62-9ade-5a7a808ae5a2",
+   "id": "eaad9a82-0592-4315-9931-0621054bdd0e",
    "metadata": {},
    "source": [
     "# How to trim messages\n",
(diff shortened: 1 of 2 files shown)


docs/docs/how_to/trim_messages.ipynb
```diff
@@ -70,73 +114,123 @@
     "    HumanMessage(\"what do you call a speechless parrot\"),\n",
     "]\n",
     "\n",
+    "\n",
     "trim_messages(\n",
     "    messages,\n",
-    "    max_tokens=45,\n",
+    "    # Keep the last <= n_count tokens of the messages.\n",
     "    strategy=\"last\",\n",
+    "    # highlight-start\n",
```
ccurme, line 122: very nit but I was a little confused as to why these particular parameters are highlighted. another option is to just set
```python
token_counter = ChatOpenAI(model="gpt-4o")  # can also pass a custom token counter
max_tokens=45  # adjust to desired history length
```
right above `trim_messages(...)`

ccurme (commented):
Guide still features `RunnableWithMessageHistory` at the end. Do we intend to replace that?

docs/docs/how_to/trim_messages.ipynb
eyurtsev: Got it

eyurtsev:
I'll merge as is, but we can update afterwards -- the role of the highlighting is to prevent users from copy pasta without adjusting the parameters appropriately

Issue #26763, opened 2024-09-23
Bug Report: astream_events fails when tool returns a dictionary


sharrajesh:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

## Description
When using `astream_events` with an agent that has multiple tools, if one of the tools returns a dictionary containing an answer and source documents, the streaming process fails. However, the same setup works correctly in non-streaming mode.

## Steps to Reproduce
1. Set up an agent with multiple tools, including one that returns a dictionary.
2. Use `astream_events` to stream the agent's output.
3. Observe the error when the tool returning a dictionary is called.

## Code to Reproduce

```python
import asyncio
import json
import os
import random
import warnings
from typing import Type, Dict, Any

import boto3
from dotenv import find_dotenv, load_dotenv
from langchain.agents import create_react_agent, AgentExecutor
from langchain.prompts import PromptTemplate
from langchain.tools import BaseTool
from langchain_aws import ChatBedrock
from langsmith import Client
from pydantic import BaseModel, Field

# Suppress warnings
warnings.filterwarnings("ignore")

load_dotenv(find_dotenv())

reproduce_problem = True

boto3_session = boto3.Session(
    aws_access_key_id=os.environ["BEDROCK_AWS_ACCESS_KEY"],
    aws_secret_access_key=os.environ["BEDROCK_AWS_ACCESS_SECRET"],
    region_name=os.environ["BEDROCK_AWS_REGION"],
)
model = ChatBedrock(
    model_id="anthropic.claude-3-5-sonnet-20240620-v1:0",
    client=boto3_session.client("bedrock-runtime"),
    streaming=True,
    model_kwargs={
        "temperature": 0,
    },
    tags=["agent"],
    metadata={"streaming": True},
    callbacks=None,
)


class CatHidingInput(BaseModel):
    query: str = Field(default="", description="Query about where the cat is hiding")


class CatHidingTool(BaseTool):
    name = "where_cat_is_hiding"
    description = "Use this tool to find out where the cat is hiding right now."
    args_schema: Type[BaseModel] = CatHidingInput

    async def _arun(self, query: str) -> str:
        return random.choice(["under the bed", "on the shelf"])

[...]

PR #26794, opened 2024-09-23
core[patch]: Agent Observations should be cast to string to avoid errors

keenborder786:
- **Description:**  When the user calls `astream_events` for `AgentExecutor`, it executes `__aiter__`, and for every chunk it gets from `agent_executor`, it calls its `messages` property. However, when the `messages` property of `AgentStep` is called, the following function is invoked: `_convert_agent_observation_to_messages`, which converts observations to `HumanMessage`. The problem is that `HumanMessage` only expects a string in its content, which was not happening, and therefore the error was raised, as pointed out in the issue below.
    - **Issue:** #26763


keenborder786:
@eyurtsev

libs/core/langchain_core/agents.py
```diff
@@ -189,10 +189,18 @@ def _convert_agent_observation_to_messages(
...
         AIMessage that corresponds to the original tool invocation.
     """
+
     if isinstance(agent_action, AgentActionMessageLog):
         return [_create_function_message(agent_action, observation)]
     else:
-        return [HumanMessage(content=observation)]
+        if not isinstance(observation, str):
+            try:
+                content = json.dumps(observation, ensure_ascii=False)
```
eyurtsev, line 198: Could you add a comment to explain why json.dumps is chosen as the default strategy vs. str? ( I'm actually not sure which one should be used -- is there a good way to think about this?)

keenborder786: okay let me think about this.

sharrajesh: fwiw i added these gists to repro the issue

https://github.com/langchain-ai/langchain/issues/26763#issuecomment-2367315600

this issue happens with tools with dict when agent was created with create_react_agent https://gist.github.com/sharrajesh/765c0b6edfe991363675f45d467e3c93

but does not happen with other variant of agent creation create_openai_tools_agent which i cannot change without introducing risk
https://gist.github.com/sharrajesh/1080af5a95ae9d7b83a8da46597b68e1

i can give you branch a try to make sure it really works if it helps?

keenborder786: Yes don't worry @sharrajesh all of these will get resolved.

eyurtsev:
@keenborder786 benchmarking total time isn't relevant (at least not directly). If the encoding affected the number of cycles that the agent had to do because it affected the chat models ability to correctly use the encoded output, then sure that's a relevant consideration, but that's not what's measured in these benchmarks. 

The way to think about json.dumps vs. str is based on:  For things typically returned from tools, how should that output be cast into string format to be best consumed by a chat model. This mean that the chat model would need to act on that information correctly.

Why did you chose json.dumps for this PR? Is there another part in langchain code that has similar logic?

---

As an side: The benchmarks themselves are completely dominated by noise from fluctuations in chat model run times. 

If you care you can verify this by calculating the standard error of the mean (which is the standard deviation divded by the square root of the number of observations) -- https://en.wikipedia.org/wiki/Standard_error, it'll give you a sense about the uncertainty in our measurement of the mean.

sharrajesh:
please also try running create_react_agent version of this code which is causing the issue 

https://gist.github.com/sharrajesh/765c0b6edfe991363675f45d467e3c93

keenborder786:
> @keenborder786 benchmarking total time isn't relevant (at least not directly). If the encoding affected the number of cycles that the agent had to do because it affected the chat models ability to correctly use the encoded output, then sure that's a relevant consideration, but that's not what's measured in these benchmarks.
> 
> The way to think about json.dumps vs. str is based on: For things typically returned from tools, how should that output be cast into string format to be best consumed by a chat model. This mean that the chat model would need to act on that information correctly.
> 
> Why did you chose json.dumps for this PR? Is there another part in langchain code that has similar logic?
> 
> As an side: The benchmarks themselves are completely dominated by noise from fluctuations in chat model run times.
> 
> If you care you can verify this by calculating the standard error of the mean (which is the standard deviation divded by the square root of the number of observations) -- https://en.wikipedia.org/wiki/Standard_error, it'll give you a sense about the uncertainty in our measurement of the mean.

@eyurtsev okay

keenborder786:
> > @keenborder786 benchmarking total time isn't relevant (at least not directly). If the encoding affected the number of cycles that the agent had to do because it affected the chat models ability to correctly use the encoded output, then sure that's a relevant consideration, but that's not what's measured in these benchmarks.
> > The way to think about json.dumps vs. str is based on: For things typically returned from tools, how should that output be cast into string format to be best consumed by a chat model. This mean that the chat model would need to act on that information correctly.
> > Why did you chose json.dumps for this PR? Is there another part in langchain code that has similar logic?
> > As an side: The benchmarks themselves are completely dominated by noise from fluctuations in chat model run times.
> > If you care you can verify this by calculating the standard error of the mean (which is the standard deviation divded by the square root of the number of observations) -- https://en.wikipedia.org/wiki/Standard_error, it'll give you a sense about the uncertainty in our measurement of the mean.
> 

> @eyurtsev . 

-  You are right but then I guess benchmarking is actually futile, because if our concern is `how chat model would need to act on that information correctly` , then we can always use `json.dumps`, and if it throws an error then we can use `str` to cast the obervation.
- Yes I choosed json.dumps because of `_create_function_message` but without the exception but I guess we can use the exact same logic when casting: https://github.com/langchain-ai/langchain/blob/b2c315997c5a3b0664797ae712e23f69f9b5f1ab/libs/core/langchain_core/agents.py#L212, It should being used by `_create_function_message`.

eyurtsev:
> Yes I choosed json.dumps because of _create_function_message but without the exception but I guess we can use the exact same logic when casting (I just made a commit, please see):

Got it -- I didn't see it from the github diff

eyurtsev (approved):

libs/core/langchain_core/agents.py
sharrajesh: Team, when you do expect this fix will be merged?

PR #26797, opened 2024-09-24
core: Run LangChainTracer inline

nfcampos:
- this flag ensures the tracer always runs in the same thread as the run being traced for both sync and async runs
- pro: less chance for ordering bugs and other oddities
- blocking the event loop is not a concern given all code in the tracer holds the GIL anyway


hinthornw (approved):

eyurtsev:
Aren't there network requests being done which would result in the gil being released?

nfcampos:
The langchain tracer doesn't make any network requests inline

eyurtsev:
Got it -- won't hurt to add a doc-string in run_inline explaining why that's OK. Knowing that this is OK depends on understanding what's happening inside the langsmith client

PR #26829, opened 2024-09-24
core: Add ruff rules for comprehensions (C4)

cbornet:


libs/core/langchain_core/output_parsers/json.py
```diff
@@ -108,7 +108,7 @@ def get_format_instructions(self) -> str:
             return "Return a JSON object."
         else:
             # Copy schema to avoid altering original Pydantic schema.
-            schema = {k: v for k, v in self._get_schema(self.pydantic_object).items()}
+            schema = dict(self._get_schema(self.pydantic_object).items())
```
eyurtsev, line 111: hmm why is this not .copy()

eyurtsev (approved):

eyurtsev:
cc @efriis if you want to take another look prior to merging. I think this looks good, code is cleaner and likely at some point will result in a tiny code speed up

efriis (approved):

PR #26836, opened 2024-09-24
core: Add various ruff rules

cbornet:
Adds
- ASYNC
- COM
- DJ
- EXE
- FLY
- FURB
- ICN
- INT
- LOG
- NPY
- PD
- Q
- RSE
- SLOT
- T10
- TID
- YTT


libs/core/langchain_core/runnables/graph_mermaid.py
```diff
@@ -291,9 +292,14 @@ async def _render_mermaid_using_pyppeteer(
     img_bytes = await page.screenshot({"fullPage": False})
     await browser.close()
 
+    def write_to_file(path: str, bytes: bytes) -> None:
```
eyurtsev, line 295: is this from a ruff rule?

eyurtsev:
@cbornet looks good -- what fraction of these have corresponding auto-formatting if any?

libs/core/langchain_core/runnables/graph_mermaid.py
cbornet: Yes because we are in an async context so we can’t use blocking `open()` and `write()` directly (ruff rule ASYNC)

cbornet:
Auto-fix is mostly `raise SomeError()` -> `raise SomeError`.
The rest is manual.

efriis:
DJ - why do we need django lint rules?

TID - nice! looks like we can enforce some of the import greps with this instead of the scripts

PR #26842, opened 2024-09-25
community[patch]: callback before yield for friendli

subroy13:
**Description:** Moves yield to after callback for `_stream` and `_astream` function for the friendli model in the community package
**Issue:** #16913


eyurtsev (approved):

eyurtsev:
Thank you!

PR #26889, opened 2024-09-26
core: impl DocumentCompressorSequence

A0nameless0man:
## Description

Implement the `DocumentCompressorSequence` class and add the `__or__` method to `BaseDocumentCompressor`.

This change allows for chaining different compressors. For example, this PR enables combining `BCERerank` with `LLMLinguaCompressor`, allowing documents to be filtered and streamlined simultaneously.

## Issue

Not applicable

## Dependencies

No dependencies are required for this change.

## Twitter Handle

Not applicable


eyurtsev (changes requested):
Hi @A0nameless0man, thanks for the PR contribution. I don't think this is a feature we'd be interested in adding for now since it can be easily handled in user code.

PR #26910, opened 2024-09-26
OCI Generative AI tool calling bug fix

raveharpaz:
- [x] **PR title**: 
  "community: OCI Generative AI tool calling bug fix 

- [x] **PR message**: 
    - **Description:** bug fix for streaming chat responses with tool calls. Update to PR 24693
    - **Issue:** chat response content is repeated when streaming
    - **Dependencies:** NA
    - **Twitter handle:** NA

- [x] **Add tests and docs**: NA

- [x] **Lint and test**: make format, make lint and make test we run successfully


libs/community/langchain_community/chat_models/oci_generative_ai.py
```diff
@@ -158,7 +158,10 @@ def chat_response_to_text(self, response: Any) -> str:
 
     def chat_stream_to_text(self, event_data: Dict) -> str:
         if "text" in event_data:
-            return event_data["text"]
+            if "finishedReason" in event_data or "toolCalls" in event_data:
+                return ""
+            else:
+                return event_data["text"]
```
efriis, line 164:
```suggestion
            return event_data.get("text")
```

efriis (commented):
what's the error? could you add a unit test that triggers the case you're reporting?

libs/community/langchain_community/chat_models/oci_generative_ai.py
YouNeedCryDear: @efriis It is actually not a key error. It is the mechanism of our API. When finishedReason and toolCalls exists in the response, "text" field will output all the previous content generated and combine them together. This will make the streaming result to be duplicated two to three times.

efriis (approved):

PR #26917, opened 2024-09-26
openai[patch]: use max_completion_tokens in place of max_tokens

ccurme:
`max_tokens` is deprecated: https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens


ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain/actions/runs/11060036967

DuhastMish:
How soon should we expect these changes?

efriis (approved):

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -435,7 +435,7 @@ class BaseChatOpenAI(BaseChatModel):
     """Number of chat completions to generate for each prompt."""
     top_p: Optional[float] = None
     """Total probability mass of tokens to consider at each step."""
-    max_tokens: Optional[int] = None
+    max_tokens: Optional[int] = Field(default=None, alias="max_completion_tokens")
```
baskaryan, line 438: do we need to update azure openai as well? if someone is passing in max_completion_tokens to azure today this change will update so instead max_tokens is passed in

Issue #6900, opened 2023-06-29
load_prompt Unable to set encoding for JSON files


jiaxin576:
### System Info

llm = OpenAI(temperature=0)
chain = LLMChain(llm=llm, prompt=load_prompt("coolapk_prompt.json"))

config = json.load(f)
UnicodeDecodeError: 'gbk' codec can't decode byte 0xaf in position 108: illegal multibyte sequence

### Who can help?

llm = OpenAI(temperature=0)
chain = LLMChain(llm=llm, prompt=load_prompt("coolapk_prompt.json"))

### Information

- [ ] The official example notebooks/scripts
- [X] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [X] Prompts / Prompt Templates / Prompt Selectors
- [ ] Output Parsers
- [ ] Document Loaders
- [ ] Vector Stores / Retrievers
- [ ] Memory
- [ ] Agents / Agent Executors
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

 
llm = OpenAI(temperature=0)
chain = LLMChain(llm=llm, prompt=load_prompt("coolapk_prompt.json"))

### Expected behavior

add set encoding for JSON files

yuhenggh:
This indeed is a problem. By default, this method uses system encoding to decode when loading template files, which would inevitably lead to errors for non-utf-8 encoded files. It would be ideal to include the autodetect_encoding=True option or provide an encoding parameter. I kindly ask you to consider adding these features to enhance flexibility and user experience. For modifications, you can refer to responses to similar issues at https://github.com/langchain-ai/langchain/issues/9005.

Jasper-zh:
Has this problem been solved? can load_prompt () specify utf-8?

josebenitezg:
same question here

Leonx01:
same question

zakki:
Fixed by https://github.com/langchain-ai/langchain/pull/21559

PR #26935, opened 2024-09-27
core: Add encoding config to json for load_prompt()

zakki:
- **Description:**

This PR lets you specify file encoding of "*_path" in the config.
If an encoding is set in the config, it will be used when opening the file.
Otherwise, the default encoding will be applied.

The default encoding in Python differs depending on the environment.
On Windows, it is the ANSI code page (ex: "cp932").

- **Issue:** https://github.com/langchain-ai/langchain/issues/6900
- **Twitter handle:** https://x.com/k_matsuzaki

### DIFF
--- a/libs/core/langchain_core/prompts/loading.py
+++ b/libs/core/langchain_core/prompts/loading.py
@@ -51,9 +51,13 @@ def _load_template(var_name: str, config: dict) -> dict:
             )
         # Pop the template path from the config.
         template_path = Path(config.pop(f"{var_name}_path"))
+        if f"{var_name}_encoding" in config:
+            encoding = config.pop(f"{var_name}_encoding")
+        else:
+            encoding = None
         # Load the template.
         if template_path.suffix == ".txt":
-            with open(template_path) as f:
+            with open(template_path, encoding=encoding) as f:
                 template = f.read()
         else:
             raise ValueError
--- a/libs/core/tests/unit_tests/examples/simple_prompt_with_template_file_cp932.json
+++ b/libs/core/tests/unit_tests/examples/simple_prompt_with_template_file_cp932.json
@@ -1,0 +1,6 @@
+{
+    "_type": "prompt",
+    "input_variables": ["adjective", "content"],
+    "template_path": "simple_template_cp932.txt",
+    "template_encoding": "cp932"
+}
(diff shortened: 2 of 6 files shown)


zakki:
@hwchase17 @efriis Could you please review this?
This patch addresses an issue similar to #21559. 
While #21559 fixes json encoding, this patch focuses on the encoding of text files as specified within a prompt json.

eyurtsev:
@zakki i don't think that this we should be supporting multiple encodings. I'd rather just specify the encoding explicitly to be utf-8 and remove all the OS dependence

eyurtsev (changes requested):
We should hard-code the encoding as utf-8 in open I think.

Just need to figure out what to do this as it could be a breaking change to folks on windows

zakki:
I don't think hard-coding UTF-8 is a bad idea, and in the long term, the problem will be solved by Python 3.15 with PEP-686.
If it's my personal execution environment, I can also handle it with environment variables `PYTHONUTF8=1`.

However, in the short term, the problem is the environment dependency when distributing applications.

eyurtsev:
@zakki also I wanted to confirm the current PR only handles the decoding path? What code path takes care of the encoding path so users on windows that may use cp1252 or some other encoding end up creating appropriate json files?

> However, in the short term, the problem is the environment dependency when distributing applications.

Sorry could you paraphrase, are you arguing for your solution vs. the hard-coding UTF-8 everywhere?

zakki:
> @zakki also I wanted to confirm the current PR only handles the decoding path? What code path takes care of the encoding path so users on windows that may use cp1252 or some other encoding end up creating appropriate json files?

For my use case, I create the output file in my own python code, not within langchain framework code.

Considering completeness for other use cases, it may be necessary to explicitly specify the encoding of the output file. 
Perhaps this creates cp1252 json on windows for example.

https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/callbacks/sagemaker_callback.py#L24

> Sorry could you paraphrase, are you arguing for your solution vs. the hard-coding UTF-8 everywhere?

I think  it would be reasonable to enforce UTF-8 as a breaking change in version 0.4.x.

eyurtsev:
Going to close this PR!

We'll force all file writes/reads to be utf-8 by default and enforce via linting, but only on next major breaking release

PR #26944, opened 2024-09-27
infra: add CI job for running tutorial notebooks

ccurme:


.github/workflows/run_notebooks.yml
```diff
@@ -0,0 +1,58 @@
+name: Run notebooks
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 13 * * *'
+
+env:
+  PYTHON_VERSION: "3.11"
```
efriis, line 9: might be worth parameterizing this to make sure they run in 3.9 + 3.12 mostly

.github/workflows/run_notebooks.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Run notebooks
+        env:
+          # these won't actually be used because of the VCR cassettes
+          # but need to set them to avoid triggering getpass()
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          MISTRAL_API_KEY: ${{ secrets.MISTRAL_API_KEY }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          TAVILY_API_KEY: ${{ secrets.TAVILY_API_KEY }}
+        run: |
+          ./docs/scripts/execute_notebooks.sh
```
efriis, line 58: if we know they won't be needed, could we skip populating the real secrets and instead just run `export ANTHROPIC_API_KEY=fake-api-key` commands?

docs/scripts/prepare_notebooks_for_ci.py
```diff
@@ -0,0 +1,181 @@
...
+CASSETTES_PATH = os.path.join(DOCS_PATH, "cassettes")
+
+# TODO: populate if needed
+NOTEBOOKS_NO_CASSETTES = [
+    "docs/docs/tutorials/retrievers.ipynb",  # TODO: fix non-determinism
+]
+
+NOTEBOOKS_NO_EXECUTION = [
+    "docs/docs/tutorials/graph.ipynb",  # Requires local graph db running
+    "docs/docs/tutorials/local_rag.ipynb",  # Local LLMs
```
efriis, line 22: ooc why can't this be run off a cassette?

docs/scripts/execute_notebooks.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Read the list of notebooks to skip from the JSON file
+SKIP_NOTEBOOKS=$(python -c "import json; print('\n'.join(json.load(open('docs/notebooks_no_execution.json'))))")
+
+# Function to execute a single notebook
+execute_notebook() {
+    file="$1"
+    echo "Starting execution of $file"
+    start_time=$(date +%s)
+    if ! output=$(time poetry run jupyter execute "$file" 2>&1); then
```
efriis, line 11: might be worth switching to `jupyter nbconvert --to notebook --execute $file` for later customizability if we ever want to process the outputs / look at the file

doesn't need to block this PR - this looks like it only gets errors from the jupyter runtime and might be tricky to get warnings/incorrect outputs from cells?

docs/scripts/execute_notebooks.sh
```diff
@@ -0,0 +1,31 @@
...
+    fi
+    end_time=$(date +%s)
+    execution_time=$((end_time - start_time))
+    echo "Finished $file. Execution time: $execution_time seconds"
+}
+
+export -f execute_notebook
+
+# Find all notebooks and filter out those in the skip list
+notebooks=$(find docs/docs/tutorials -name "*.ipynb" | grep -v ".ipynb_checkpoints" | grep -vFf <(echo "$SKIP_NOTEBOOKS"))
```
efriis, line 26: if there's a way to parameterize this, might make this CI step more customizable in future if we want to use it to run e.g. individual notebooks

efriis (approved):

.github/workflows/run_notebooks.yml
ccurme: Will do separately as it's easier to debug workflow dispatch after merging the new workflow to master.

.github/workflows/run_notebooks.yml
ccurme: Deleted this comment as these actually do get used for notebooks where we skip cassettes.

docs/scripts/prepare_notebooks_for_ci.py
ccurme: Local LLM notebook I think may not make http calls, there's just significant overhead running local models (e.g., ollama). LMK if I've misunderstood.

docs/scripts/execute_notebooks.sh
ccurme: Will look into this after merging so we can debug workflow dispatch.

.github/workflows/run_notebooks.yml
ccurme: Added python version as a parameter in https://github.com/langchain-ai/langchain/pull/26956

docs/scripts/execute_notebooks.sh
ccurme: Added in https://github.com/langchain-ai/langchain/pull/26956

PR #26965, opened 2024-09-28
core: Add ruff rules for error messages (EM)

cbornet:
All auto-fixes


efriis (commented):
Why is is undesirable to have f-strings in error messages?

cbornet:
It comes from this rule: https://docs.astral.sh/ruff/rules/f-string-in-exception/

efriis (approved):

efriis:
got it. works for me

PR #26993, opened 2024-09-30
all: Add pre-commit hook

cbornet:
This calls `make format` on projects that have modified files.
So `poetry install --with lint` must have been done for those projects.


efriis (commented):
I'm lightly against this and open to being convinced. In general, I feel like encouraging small commits (even with lint/format errors) is good, even if it means running `make format` at the end.

Should we instead document how to add it for people who want it?

cbornet:
Using the pre commit hook is opt-in.
You have to install it with `pre-commit install`.
So it’s up to the user taste to use it or not.

efriis:
in that case could you 

1. move it into the scripts/ directory (to not clutter the root)
2. add to contributing code documentation an optional step using `pre-commit install --config configs/.pre-commit-config.yaml`?

cbornet:
> move it into the scripts/ directory (to not clutter the root)

OK but that makes it less discoverable. People used to pre-commit that would see the file in the root directly know they can use it. Otherwise they have to read the setup doc which I believe most of the time they won't do (I didn't do it when starting contributing to LangChain, I saw a` poetry.lock` file in the root and so I ran `poetry install`. I also saw a Makefile in the root and knew there would be interesting stuff there)

efriis:
given the lack of engagement from folks here or in issues/discussions on precommit hooks I'm going to close this - it increases maintenance surface area (e.g. devcontainers and docker stuff that is stale now)

happy to look into it again as part of a broader contributor experience sprint. agreed figuring out linting/formatting could be easier. don't think hardcoding some of the libraries like this is the way.

efriis:
have seen enough engagement here in the last few weeks that it's probably worth it. Some interesting learnings though

- make doesn't work on windows machines - is there a way we can offer some different compatibility via the precommit hook
- is there a way to have the hooks "discovered" for each of the libs/partners dirs as well

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+      name: format langchain
+      language: system
+      entry: make -C libs/langchain format
+      files: ^libs/langchain/(langchain|tests)/
+      pass_filenames: false
+    - id: standard-tests
+      name: format standard-tests
+      language: system
+      entry: make -C libs/standard-tests format
+      files: ^libs/standard-tests/(langchain_standard_tests|tests)/
```
efriis, line 26: might be worth adding the whole directory because changes to pyproject.toml/poetry.lock/makefile could make changes too, and this is what we detect in CI

efriis (commented):
is there a way we can include the partners/ directories as well? Ideally in an auto-detected manner - ideally would extend to the main libs too. Not sure if this is possible in precommit / how much additional scripting would be needed. If helpful I could try to make the script at `check_diff.py` work for checking for directory changes in the same way we do in CI

.pre-commit-config.yaml
cbornet: OK. I'll update

cbornet:
> is there a way we can include the partners/ directories as well?

Yes, I'll add them.

.pre-commit-config.yaml
cbornet: Done.

cbornet:
> is there a way we can include the partners/ directories as well?

Done

PR #26998, opened 2024-09-30
core: add **kwargs to index and aindex functions for custom vector_field support

Jalmeida1994:
Added `**kwargs` parameters to the `index` and `aindex` functions in `libs/core/langchain_core/indexing/api.py`. This allows users to pass additional arguments to the `add_documents` and `aadd_documents` methods, enabling the specification of a custom `vector_field`. For example, users can now use `vector_field="embedding"` when indexing documents in `OpenSearchVectorStore`


libs/core/langchain_core/indexing/api.py
```diff
@@ -198,6 +198,7 @@ def index(
     source_id_key: Union[str, Callable[[Document], str], None] = None,
     cleanup_batch_size: int = 1_000,
     force_update: bool = False,
+    **kwargs: Any,
```
eyurtsev, line 201: Don't use `**` here to avoid confusion of which namespace kwargs belong (e.g., what if we need kwargs for delete at some point or else add other parameters to index in the future).

Instead can call is `upsert_kwargs`. Please document which API whey will be passed to in the doc-string

eyurtsev (changes requested):
Added a comment to explain how to change the variable name. 

Will need a unit test as well before this can be merged

libs/core/langchain_core/indexing/api.py
Jalmeida1994: Thank you for the feedback. I've implemented the changes as suggested:

1. Replaced `**kwargs` with a specific `upsert_kwargs: Optional[dict[str, Any]] = None` parameter.
2. Updated the docstring to clarify that these kwargs will be passed to the `add_documents` and `aadd_documents` method of the VectorStore.
3. Added an example in the docstring to demonstrate how `upsert_kwargs` can be used, specifically mentioning the `vector_field` use case.

Jalmeida1994: Also added some unit tests for `index` and `aindex` with `upsert_kwargs`.

libs/core/langchain_core/indexing/api.py
```diff
@@ -363,7 +368,10 @@ def index(
...
             if isinstance(destination, VectorStore):
                 destination.add_documents(
-                    docs_to_index, ids=uids, batch_size=batch_size
+                    docs_to_index,
+                    ids=uids,
+                    batch_size=batch_size,
+                    **(upsert_kwargs or {}),
                 )
             elif isinstance(destination, DocumentIndex):
                 destination.upsert(docs_to_index)
```
eyurtsev, line 377: They also need to be added here -- I'll push a change in a bit

libs/core/tests/unit_tests/indexing/test_indexing.py
```diff
@@ -1728,3 +1729,180 @@ async def test_incremental_aindexing_with_batch_size_with_optimization(
...
+    assert upsert_spy.call_args.kwargs == upsert_kwargs
+
+
+async def test_aindex_with_upsert_kwargs(
+    arecord_manager: InMemoryRecordManager, upserting_vector_store: InMemoryVectorStore
+) -> None:
+    """Test async indexing with upsert_kwargs parameter."""
+    mock_aadd_documents = AsyncMock()
+
+    with patch.object(upserting_vector_store, "aadd_documents", mock_aadd_documents):
```
eyurtsev, line 1865: I find usage of `spy` here to be a bit easier on readability

eyurtsev (approved):

libs/core/tests/unit_tests/indexing/test_indexing.py
Jalmeida1994: Thanks for the feedback!

PR #27014, opened 2024-10-01
community[patch]: Fix handling of a few integrations that use private attribute

eyurtsev:
In pydantic 2 model_validator(mode="before") and root_validator(pre=True) ignore any private attributes that may be passed through values


ccurme:
Will need to move this to https://github.com/langchain-ai/langchain-community

PR #27020, opened 2024-10-01
[fix] error on invoke HuggingFaceEndpoint Bad request: following model_kwargs

Warra07:
- [x] **fix invoke error on HuggingFaceEndpoint**: "fix invoke error HuggingFaceEndpoint Bad request"

    - **Description:** ability to set 'stop_sequences' parameter to None on models that doesn't uses it
    - **Issue:**  from [18321](https://github.com/langchain-ai/langchain/issues/18321) When trying to use invoke on some models with HuggingFaceEndpoint, users may get the following error:
Bad request: The following model_kwargs are not used by the model: ['return_full_text', 'stop', 'watermark', 'stop_sequences']
this is due to the fact that the call function send those parameters to the huggingface  inferenceClient and huggingface raise an error if those parameters are not used, in this specific case, the user should set the unused parameters to None on the invoker.
however for the specific case of 'stop_sequences' and 'stop', if we set them up to None an error occured because the call was still trying to iterate through the stop_sequences parameter and concatenate it
    - **Dependencies:** No new dependency

- [x] **Add tests and docs**: modified  integration_tests/llms/test_huggingface_endpoint.py  file to reflect the change, tests did not work properly previously.
also added a note on HuggingFaceEndpoint docstring to inform users.


efriis:
hey there! this change is against the deprecated community version of the huggingface integration. Please try the `langchain-huggingface` package in `libs/partners/huggingface`, and if the error appears there, make the change against that package.

Warra07:
oh sorry i missed it, i think the error will still occure in the partner package, i'll dig into it later :)

PR #27025, opened 2024-10-01
Opensearch embedding

komikndr:
- **Description:** Adding embedding feature for opensearch based on ML pluggin sentece transformer. Example is in "docs/docs/integrations/text_embedding/opensearch.ipynb"
- **Dependencies:** None
- **Twitter handle:** @kxnlxia


libs/community/tests/integration_tests/embeddings/test_opensearch.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.fixture
+def model_id() -> str:
+    """Fixture to provide model ID."""
+    return "some-model-id"
+
+
+@pytest.fixture
+def client() -> OpenSearch:
+    """Fixture to provide OpenSearch client connection."""
+    return OpenSearch(
```
pberger514, line 18: Does this assume that an OpenSearch domain instance exists at localhost? If so I think this should be mocked.

komikndr: Yeah, I forgot about the mock and just assumed localhost. Well, back to fixing my test code. Thanks for the insight!

Btw, if i may ask, is it be better for me to upload the docs for the embedding after the code passing, or just upload it together with the code?

PR #27030, opened 2024-10-01
core: add parser for numeric comma separated output

marcelomachado:
- [ ] **PR title**: "core: add parser for numeric comma separated output""

- [ ] **PR message**: 
    - **Description:** Adding class to handle numeric comma separated output.
    - **Issue:** N/A
    - **Dependencies:** None
    - **Twitter handle:** @celow_m


efriis (commented):
I think it will make more sense to just use a comma list output parser and then cast the output elements as ints/floats

ScreenShot 2024-10-08 at 10 52 52AM


let me know what you think

marcelomachado:
Hi @efriis, thank you so much for your feedback.

I believe there is value in having this as a separate class, as it allows us to address potential formatting issues the LLM might generate in advance. For example, I've encountered cases where LLMs use commas instead of dots to separate decimals or sometimes insert commas to denote thousands. These could be handled within the class.

Additionally, this isn't just a parser—it also contains other helpful methods, such as `get_format_instructions`, which can be used programmatically. In the pipeline I'm building, I’m unable to provide the kind of instructive prompt you mentioned. Instead, I rely on parsers, which are selected dynamically, and we use the instructions from the `get_format_instructions` method. While I understand that I can extend this to my own project, I thought it might be useful to share this as a common method for the community, particularly for those who may face similar situations.

marcelomachado:
Hi @efriis, does that make sense?

efriis:
got it - in this case, I think we'll pass on adding this output parser to core. in particular, having too many classes that do slightly different things is confusing to users, and we're trying to reduce rather than proliferate that at the moment.

Appreciate the thought!

PR #27039, opened 2024-10-02
community: add PanelCallbackHandler

ahuang11:
Description:

Migrates https://panel.holoviz.org/reference/chat/PanelCallbackHandler.html over to langchain-community.

Dependencies: `panel`
Twitter: IAteAnDrew1


ahuang11:
Can you take a look at this PR @vbarda? Thanks!

ahuang11:
@ccurme can I have some more time to gather upvotes? Thanks!

efriis:
hey! lots of CI issues, and I don't think this makes sense to duplicate the code here (in general we're more encouraging of externally-packaged/documented callback handlers like you have already!)

if I added a section to the callbacks list on "external callbacks" that included a link to [your linked docs page](https://panel.holoviz.org/reference/chat/PanelCallbackHandler.html) would that work?

ahuang11:
> lots of CI issues

I can fix those; seems like it's just long lines

> I don't think this makes sense to duplicate the code

We are actually planning to deprecate this in Panel https://github.com/holoviz/panel/issues/7470

I think we'd prefer the callback handler to live in the langchain community.

ahuang11:
Can I have this reviewed? Thanks!

efriis:
oops sorry for the delay! we're actually not taking on new callback integrations (see [this doc](https://python.langchain.com/docs/contributing/how_to/integrations/) ) - mostly because of maintenance overhead if people make PRs against a callback like this, which tend to be difficult to review, as callbacks tend to rely much more on the external integration.

Given it's already in our docs, we're happy to accept updates to the docs pointing the integration to how it's best to install! Let me know if that makes sense

If it works for you, you're also welcome to publish a separate langchain-panel package that includes the code such that it doesn't live in either main package

PR #27066, opened 2024-10-03
fix:Specified the model that solves langchain moonshotChat exmaple di…

Swastik-Swarup-Dash:
resolving#27058


newday1:
sorry it didn't work . The error is

PR #27068, opened 2024-10-03
community : added support for pickle networkx graph import

benx13:
- **Description:** added support for pickle networkx graph import
    - **Dependencies:** pickle


efriis (commented):
Don't think this adds much - you can add `pickle.load(open(pickle_path, 'rb'))` instead, and pickle.load is a dangerous operation if called on untrusted code.

Would prefer to let users who want pickle functionality like this to write the pickle commands themselves so they understand the security implications of what they're writing!

PR #27071, opened 2024-10-03
community: add bm25 retriever model save and load method

i-w-a:
Description
This PR adds functionality for saving and loading retrievers to the` BM25Retriever` class. This feature, already present in the `TFIDFRetriever`, is extremely useful and necessary for the `BM25Retriever` as well.

https://api.python.langchain.com/en/latest/_modules/langchain_community/retrievers/tfidf.html#TFIDFRetriever:~:text=docs%5D%20%20%20%20%40classmethod-,def%20load_local(,-cls%2C%0A%20%20%20%20%20%20%20%20folder_path

Dependencies
No additional dependencies are required for this change.

Twitter handle
@i_w__a

### DIFF
--- a/libs/community/langchain_community/retrievers/bm25.py
+++ b/libs/community/langchain_community/retrievers/bm25.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import joblib
 import pickle
 from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, List, Optional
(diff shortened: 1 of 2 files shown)


efriis:
hey there! we're not merging in new pickle save/load functionality due to security concerns around it. appreciate the dangerous flag on this, but even with that we try to reserver to experimental!

feel free to document using `pickle.loads` or `pickle.dumps` in the documentation externally, as there's not a strong reason this needs to be a defined method on the class instead of just loading/dumping the retriever object!

PR #27072, opened 2024-10-03
core[patch]: add UsageMetadata details

baskaryan:


libs/core/langchain_core/messages/ai.py
```diff
@@ -29,6 +29,62 @@
...
+            {
+                "audio": 10,
+                "cache_creation": 200,
+                "cache_read": 100,
+            }
+    """
+
+    audio: int
+    """Audio input tokens."""
+    cache_creation: int
```
jacoblee93, line 50: For OpenAI this will be `prompt_tokens  - prompt_tokens_details.cached_tokens`?

baskaryan: i'm not actually sure if cached_tokens represents cache read or cache creation or a union of the two for oai, their docs aren't very clear

"cached_tokens: integer

Cached tokens present in the prompt."

baskaryan: my understanding was its cache_read but will try to double check

baskaryan: ok yea it represents cache_read. there isn't really a concept of cache_creation for oai since they do it automatically and don't charge extra for it, so think that would just be left out

eyurtsev (approved):
looks good!

libs/core/langchain_core/messages/ai.py
```diff
@@ -29,6 +29,62 @@
...
+
+    Example:
+
+        .. code-block:: python
+
+            {
+                "audio": 10,
+                "cache_creation": 200,
+                "cache_read": 100,
+            }
```
eyurtsev, line 45: add ..version-added information

libs/core/langchain_core/messages/ai.py
```diff
@@ -39,18 +95,29 @@ class UsageMetadata(TypedDict):
...
     input_tokens: int
-    """Count of input (or prompt) tokens."""
+    """Count of input (or prompt) tokens. Sum of all input token types."""
     output_tokens: int
-    """Count of output (or completion) tokens."""
+    """Count of output (or completion) tokens. Sum of all output token types."""
     total_tokens: int
-    """Total token count."""
+    """Total token count. Sum of input_tokens + output_tokens."""
+    input_token_details: NotRequired[InputTokenDetails]
```
eyurtsev, line 119: add a doc-string here with version added?

libs/standard-tests/langchain_standard_tests/unit_tests/chat_models.py
```diff
@@ -138,6 +138,10 @@ def supports_anthropic_inputs(self) -> bool:
     def supports_image_tool_message(self) -> bool:
         return False
 
+    @property
+    def supported_usage_metadata_details(self) -> List[str]:
```
eyurtsev, line 142: List[Listeral[...]] so folks don't have to guess the names?

baskaryan: smort

PR #27081, opened 2024-10-03
community: Add instructions in InfinityEmbeddings

baptiste-pasquier:
**Description:**

Add instructions for embedding with Infinity:

```python
embed_instruction: str = ""
query_instruction: str = ""
```

Allows to pass instructions like "passage: " or "query: ". Defaults to "" (no instructions)

### DIFF
--- a/libs/community/langchain_community/embeddings/infinity.py
+++ b/libs/community/langchain_community/embeddings/infinity.py
@@ -43,6 +43,11 @@ class InfinityEmbeddings(BaseModel, Embeddings):
     client: Any = None  #: :meta private:
     """Infinity client."""
 
+    embed_instruction: str = "passage: "
+    """Instruction used to embed documents."""
+    query_instruction: str = "query: "
+    """Instruction used to embed the query."""
+
     # LLM call kwargs
     model_config = ConfigDict(
         extra="forbid",
@@ -73,7 +78,7 @@ class InfinityEmbeddings(BaseModel, Embeddings):
         """
         embeddings = self.client.embed(
             model=self.model,
-            texts=texts,
+            texts=[self.embed_instruction + text for text in texts],
         )
         return embeddings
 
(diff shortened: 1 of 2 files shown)


baptiste-pasquier:
What do you think @michaelfeil ?

libs/community/langchain_community/embeddings/infinity.py
```diff
@@ -43,6 +43,11 @@ class InfinityEmbeddings(BaseModel, Embeddings):
     client: Any = None  #: :meta private:
     """Infinity client."""
 
+    embed_instruction: str = "passage: "
```
michaelfeil, line 46: Should default to ""

michaelfeil (approved):
approved with changes

libs/community/langchain_community/embeddings/infinity.py
baptiste-pasquier: default to "" for document and also query ?

baptiste-pasquier: For BGE, the instruction for query is `Represent this sentence for searching relevant passages:` 
https://huggingface.co/BAAI/bge-large-en-v1.5

michaelfeil: Incorrect, see their repo: `update embedding model: release bge-*-v1.5 embedding model to alleviate the issue of the similarity distribution, and enhance its retrieval ability without instruction.` - by adding random instruction very deep in langchain, users of langchain will become frustrated, as they do not understand why their retrieval accuracy dropped. 

Please remove both instructions, thank you.

baptiste-pasquier: Thanks for the explanation, I set the defaults to ""

efriis:
this is better done externally when you have the documents! This implementation will also prepend the instructions many times if you call the query functions I think

PR #27088, opened 2024-10-03
First draft of concept pages

rlancemartin:


docs/docs/concepts.mdx
```diff
@@ -522,120 +522,6 @@ for modifying **multiple** key-value pairs at once:
 
 For key-value store implementations, see [this section](/docs/integrations/stores/).
 
-### Tools
```
eyurtsev, line 525: @rlancemartin shall we keep the section headings everywhere to maintain a glossary on the main concepts page?

We can remove the beefy content, but maintain headings so no links break

docs/docs/concepts/tool_calling.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+While tool calling is not universal, many model providers support it.
+
+See our [model integration page](/docs/integrations/chat/) for a list of models that support tool calling.
+
+LangChain provides a standardized interface for tool calling that is consistent across different models.
+
+`ChatModel.bind_tools()` is a method for specifying which tools are available for a model to call. 
+
+If a model has been initialized as `llm` without tools, we can bind tools to it as a list:
```
eyurtsev, line 37: Do we need to be more precise about chat model vs. llm terminology? We use them interchangeably in docs, but can be confusing?

We could also opt for favoring `model`?

docs/docs/concepts/tool_calling.mdx
```diff
@@ -0,0 +1,133 @@
...
+The arguments can be easily extracted and passed to the tool directly:
+
+```python
+args = tool_call.tool_calls[0]['args']
+result = multiply(**args)
+```
+
+For more details on usage, see our [how-to guides](docs/how_to/#tools)!
+
+## Common patterns
```
eyurtsev, line 104: This section is a nice addition, not sure if i've seen this discussion before.

docs/docs/concepts/tool_calling.mdx
```diff
@@ -0,0 +1,133 @@
+# Tool Calling
```
eyurtsev, line 1: We need a URL that we can link to that explains what a Tool is / brief example of definition/ brief example of how it can be used.

I think we need another page for tools specifically, or else retitle this page to `# Tools and have a tool calling section somewhere within it?

Some of the questions that the conceptual guide should address:

1) What is a tool?
2) Where is the tool getting executed? (maybe does a tool code need to be executed?)
3) How does one define a tool? (We have 4-5 ways of doing it); e.g., @tool  decorator, StructuredTool.from_function..... We don't need to do it in detail, but can link to the how-to guide, but just explain high level what the differences are.

docs/docs/concepts/structured_output.mdx
```diff
@@ -0,0 +1,120 @@
+# Structured output
+
+## Motivation 
+
+While many AI applications, such as chatbots, typically respond in natural language, there are scenarios where we need models to output in a structured format.
```
eyurtsev, line 5: maybe this?
```suggestion
While many AI applications, such as chatbots, typically respond in natural language, there are scenarios where we need models to output a response matching a specific structured format. 
```

docs/docs/concepts/structured_output.mdx
```diff
@@ -0,0 +1,120 @@
+# Structured output
+
+## Motivation 
+
+While many AI applications, such as chatbots, typically respond in natural language, there are scenarios where we need models to output in a structured format. 
+
+This is what we call structured output. Structured output is particularly useful when:
```
eyurtsev, line 7: nit: What does the "This is" referring to? (I think this sentence is basically saying "This structure format is what we call structured output"

docs/docs/concepts/structured_output.mdx
```diff
@@ -0,0 +1,120 @@
+# Structured output
```
eyurtsev, line 1: We have some conceptual material here:

https://python.langchain.com/v0.1/docs/modules/model_io/chat/structured_output/
https://python.langchain.com/v0.1/docs/use_cases/extraction/

docs/docs/concepts/retrieval_unstructured.mdx
```diff
@@ -0,0 +1,160 @@
+# Retrieval for Unstructured Data
+
+## Motivation
+
+Language models (LLMs) are trained on vast but fixed datasets, which limits their ability to access up-to-date or domain-specific information. 
+
+To enhance their performance on specific tasks, we can augment their knowledge using retrieval systems.
```
eyurtsev, line 7: `retrieval system` is one technique for increasing performance on specific task (another technique is just prompting).

Perhaps re-sequence here to something like this:

1. Can enhance performance on specific task by presenting the models with relevant information as part of their prompt
2. One technique to do so is by retrieving relevant information using a search engine.

docs/docs/concepts/retrieval_unstructured.mdx
```diff
@@ -0,0 +1,160 @@
...
+3. Reduced hallucination by grounding responses in retrieved facts
+4. Cost-effective alternative to fine-tuning for factual recall
+
+## Un-Structured Data
+
+Vector stores are a type of retrieval system that work with unstructured data. 
+
+Unstructured documents (e.g., raw text, tables, images) are often embedded into a compressed representation (e.g., a vector) that can used directly for search based upon similarity to an input query.
+
+## Query Translation
```
eyurtsev, line 24: Query translation -> Query analysis?

translation could mean: 
* translate from english to french
* translate from raw language to SQL

----

In case there's any relevant information here: https://python.langchain.com/v0.1/docs/use_cases/query_analysis/

docs/docs/concepts/retrieval_unstructured.mdx
```diff
@@ -0,0 +1,160 @@
...
+|---------------|-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [Multi-query](/docs/how_to/MultiQueryRetriever/)   | When you need to cover multiple perspectives of a question. | Rewrite the user question from multiple perspectives, retrieve documents for each rewritten question, return the unique documents for all queries. |
+| [Decomposition](https://github.com/langchain-ai/rag-from-scratch/blob/main/rag_from_scratch_5_to_9.ipynb) | When a question can be broken down into smaller subproblems. | Decompose a question into a set of subproblems / questions, which can either be solved sequentially (use the answer from first + retrieval to answer the second) or in parallel (consolidate each answer into final answer).                                                           |
```
eyurtsev, line 39: Could we link to any guides that exist as part of the main documentation?

See list here:
https://python.langchain.com/v0.1/docs/use_cases/query_analysis/#techniques

For example:
https://python.langchain.com/v0.1/docs/use_cases/query_analysis/techniques/decomposition/

We'd need to identify whether these are present in the 0.3 version of the docs, and if not incorporate them, into the main documentation.

Generally, we should not link to any documentation that's outside of the main docs b/c maintenance becomes very difficult + we're losing theming, linking to API refs etc.

The only exception is langgraph documentation, which we'll add more support to for now (and potentially merge in the future)

docs/docs/concepts/retrieval_unstructured.mdx
```diff
@@ -0,0 +1,160 @@
...
+| [Step-back](https://github.com/langchain-ai/rag-from-scratch/blob/main/rag_from_scratch_5_to_9.ipynb)     | When a higher-level conceptual understanding is required. | First prompt the LLM to ask a generic step-back question about higher-level concepts or principles, and retrieve relevant facts about them. Use this grounding to help answer the user question. [Paper](https://arxiv.org/pdf/2310.06117).                                            |
+| [HyDE](https://github.com/langchain-ai/rag-from-scratch/blob/main/rag_from_scratch_5_to_9.ipynb)          | If you have challenges retrieving relevant documents using the raw user inputs. | Use an LLM to convert questions into hypothetical documents that answer the question. Use the embedded hypothetical documents to retrieve real documents with the premise that doc-doc similarity search can produce more relevant matches. [Paper](https://arxiv.org/abs/2212.10496). |
+
+:::tip
+
+See our RAG from Scratch videos for a few different specific approaches:
```
eyurtsev, line 45: Love the videos for the conceptual stuff. Do we have any concerns about the code part being out of date?

docs/docs/concepts/retrieval_unstructured.mdx
```diff
@@ -0,0 +1,160 @@
...
+- [Multi-query](https://youtu.be/JChPi0CRnDY?feature=shared)
+- [Decomposition](https://youtu.be/h0OPWlEOank?feature=shared)
+- [Step-back](https://youtu.be/xn1jEjRyJ2U?feature=shared)
+- [HyDE](https://youtu.be/SaDzIVkYqyY?feature=shared)
+
+:::
+
+## Indexing Strategies
+
+Documents needed to be indexed, frequently using embedding models to [compress the semantic information in documents to fixed-size vectors](/docs/concepts/#embedding-models).
```
eyurtsev, line 55: - Has the term "Documents" been defined yet?
- Should we use the more general term of "Content" or something along those lines so it's clear that conceptually some ideas carry to other modalities?

docs/docs/concepts/retrieval_unstructured.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+Many RAG approaches focus on splitting documents into chunks and retrieving some number based on similarity to an input question for the LLM. 
+
+But chunk size and chunk number can be difficult to set and affect results if they do not provide full context for the LLM to answer a question. 
+
+Furthermore, LLMs are increasingly capable of processing millions of tokens.
```
eyurtsev, line 61: Latency is important for many applications. 

Pulling in the content of the top 10 search hits can add significant latency. 

In an extreme case, all 10 hits come from different documents.

In many domains, document length can run into the hundreds of pages (i've dozens if not hundreds of thousand page long documents in finance)

Even if the llm context window were unlimited (and could incorporate information from the context perfectly), implementation details still introduce a tradeoff in terms of latency.

Is a user willing to wait for an extra 5-60 seconds to get an answer from a chat system? This ends up being very use case dependent.

docs/docs/concepts/retrieval_unstructured.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+There are a few ways to improve the quality of similarity search. 
+
+Embedding models compress text into fixed-length (vector) representations that capture the semantic content of the document. 
+
+This compression is useful for search / retrieval, but puts a heavy burden on that single vector representation to capture the semantic nuance / detail of the document. 
+
+In some cases, irrelevant or redundant content can dilute the semantic usefulness of the embedding.
+
+[ColBERT](https://docs.google.com/presentation/d/1IRhAdGjIevrrotdplHNcc4aXgIYyKamUKTWtB3m3aMU/edit?usp=sharing) is an interesting approach to address this with a higher granularity embeddings: (1) produce a contextually influenced embedding for each token in the document and query, (2) score similarity between each query token and all document tokens, (3) take the max, (4) do this for all query tokens, and (5) take the sum of the max scores (in step 3) for all query tokens to get a query-document similarity score; this token-wise scoring can yield strong results.
```
eyurtsev, line 93: This section is describing how to implement the algorithm, but it's not explaining what the algorithm does at a conceptual level. I think most readers will gloss over this description as a result since it's not conceptual.

It would be helpful to have a sentence in the spirit of: a variation that generates multiple embeddings for the query, and retrieves closest documents to any of these embeddings

docs/docs/concepts/retrieval_semi_structured.mdx
```diff
@@ -0,0 +1,45 @@
+# Retrieval for Semi-Structured Data
+
+## Motivation
+
+Language models (LLMs) are trained on vast but fixed datasets, which limits their ability to access up-to-date or domain-specific information. 
+
+To enhance their performance on specific tasks, we can augment their knowledge using retrieval systems.
```
eyurtsev, line 7: "retrieval systems" => external information (see other related comment)

docs/docs/concepts/tool_calling.mdx
```diff
@@ -0,0 +1,133 @@
...
+```
+
+#### Tool
+
+Of course, need to specify a `tool` that the model can call.
+
+In LangChain, any function can be bound as a tool.
+
+```python
+def multiply(a: int, b: int) -> int:
```
eyurtsev, line 50: I believe we usually need a `@tool` decorator (since some code probably assumes that the tool is a runnable) at invocation time

```suggestion
def multiply(a: int, b: int) -> int:
```

docs/docs/concepts/tool_calling.mdx
rlancemartin: this is good point that i'm also wary of -- 
1/ maybe we defer to `model` by default
2/ conceptual guide on `llm` and `chat model` call out each specifically

docs/docs/concepts/tool_calling.mdx
rlancemartin: Ya -- we use this same framing in the course.

docs/docs/concepts/tool_calling.mdx
rlancemartin: that's true. let me try to add as a section on this page.

rlancemartin: also noted in Slack discussion. this is a good thing to flesh out -- 
https://langchain.slack.com/archives/C04GWPE38LV/p1728407895162059

PR #27099, opened 2024-10-04
Update README.md for Tutorial to Usecase url

bhadreshpsavani:
Fixed tutorial URL since earlier Tutorial URL was pointing to usecase age which does not have any detail it should  redirect to correct URL page


ccurme (approved):
Thank you!

PR #27103, opened 2024-10-04
[community] [feature]: Implementation of excel parser and including it in o365 loader

MacanPN:
# Description
Currently there is no implementation of `.xlsx` parser and consequently `FileSystemBlobLoader` as well as all loaders derived from `O365BaseLoader` are limited to `.doc`, `.docx`, `.pdf` and `.txt`. This PR uses _unstructured_ to implement `ExcelParser` parser very similar to `MsWordParser` (with just an extra bit of post processing).

## Dependencies: 
- *htmltabletomd*

## Tests
I've implemented a unit test for the `.xlsx` parser however it depends on `unstructured` module and I have a problem getting any version of unstructured to `pyproject.toml` that would work and didn't break something somewhere. Please let me know if anyone is willing to help me with that.

**Twitter handle: @martintriska1**

Is anyone willing to review this please? Thanks! @baskaryan, @eyurtsev, @ccurme

### DIFF
--- a/libs/community/langchain_community/document_loaders/parsers/excel.py
+++ b/libs/community/langchain_community/document_loaders/parsers/excel.py
@@ -1,0 +1,72 @@
+from typing import Iterator, Sequence
+
+from langchain_core.documents import Document
+
+from langchain_community.document_loaders.base import BaseBlobParser
+from langchain_community.document_loaders.blob_loaders import Blob
+
+
+class ExcelParser(BaseBlobParser):
+    """Parse the Microsoft Excel documents from a blob."""
+
+    def post_process_unstructured_elements(
+        self, elements: Sequence[str], source: str
+    ) -> Iterator[Document]:
+        try:
+            import htmltabletomd
+        except ImportError as e:
+            raise ImportError(
+                "Could not import htmltabletomd, please install with `pip install "
+                "htmltabletomd`."
+            ) from e
+        last_page = None
+        for element in elements:
+            if element.metadata.page_number != last_page:
+                if last_page:
+                    metadata = {"source": source, "title": element.metadata.page_name}  # type: ignore[attr-defined]
+                    yield Document(page_content=page_text, metadata=metadata)
+                page_text = f"# {element.metadata.page_name}\n"
+                last_page = element.metadata.page_number
+
+            if type(element).__name__ == "Title":
+                page_text += f"## {element.text}\n"
+            elif type(element).__name__ == "Table":
+                page_text += (
+                    htmltabletomd.convert_table(element.metadata.text_as_html) + "\n"
+                )
+            else:
+                page_text += f"{element.text}\n"
+        # yield last page
+        metadata = {"source": source, "title": element.metadata.page_name}  # type: ignore[attr-defined]
+        yield Document(page_content=page_text, metadata=metadata)
+
+    def lazy_parse(self, blob: Blob) -> Iterator[Document]:  # type: ignore[valid-type]
+        """Parse a Microsoft Excel document into the Document iterator.
+
+        Args:
+            blob: The blob to parse.
+
+        Returns: An iterator of Documents.
+
+        """
+        try:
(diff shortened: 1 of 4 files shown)


petergoldstein:
XLSX files are important for a variety of enterprise use cases.  Having a parser will be very handy for some of our current challenges.

avalon-swain:
This will be helpful

libs/community/langchain_community/document_loaders/parsers/excel.py
```diff
@@ -0,0 +1,89 @@
...
+        """Parse a Microsoft Excel document into the Document iterator.
+
+        Args:
+            blob: The blob to parse.
+
+        Returns: An iterator of Documents.
+
+        """
+        try:
+            from unstructured.partition.xlsx import partition_xlsx
```
ccurme, line 33: Thanks for this. How do you feel about using [langchain-unstructured](https://github.com/langchain-ai/langchain-unstructured/tree/main/libs/unstructured) here? Are there important differences in the resulting `Document` objects?

MacanPN: @ccurme Actually in the meantime, I've written up [this](https://github.com/langchain-ai/langchain/pull/27749) wrapper that allows to turn a `DocumentLoader` into `Parser`. Excel parser is already implemented [here](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/document_loaders/excel.py) so I'd abandon this PR in favor of merging the `DocumentLoaderAsParser` wrapper together with some more changes to `UnstructuredExcelLoader` (or possibly `langchain-unstructured.document_loaders.UnstructuredLoader`). What's the current thinking? Is the implementation in community being deprecated? (In community there is implementation with inheritance `UnstructuredExcelLoader : UnstructuredFileLoader : UnstructuredBaseLoader : BaseLoader` where as in `langchain-unstructured` there is independent implementation only inheriting directly from `BaseLoader`.

MacanPN: @ccurme I see that the [DocumentLoaderAsParser](https://github.com/langchain-ai/langchain/pull/27749) was assigned status _needs support_. I would appreciate if you could take a look at that one in context of being substitute for this PR. Thanks!

MacanPN: @ccurme I'm thinking I might not have made clear enough about why having `BlobParser` rather than `DocumentLoader` is important. Please let me breifly explain:
- Document loaders function as `<whatever> -> documents`
- BlobParsers process `blob -> documents`.
- Many document loaders actually implement logic until they have a blob. Then the actual parsing is done by calling parser with similar name. As an example you can look at `AmazonTextractPDFLoader` and `AmazonTextractPDFParser`.
- Some document loaders may want to process numerous filetypes. An example is a `SharePointLoader` that is supposed to fetch all parsable files, parse them and produce documents. Until recently it processed only 3 file types `doc, docx, pdf` and ignored everything else. Recently I got [this pr](https://github.com/langchain-ai/langchain/pull/27716) merged that enables a user to pass in dict mapping file types (or mime types) to parsers. Now one can easily use `SharePointLoader` with any files where suitable parser is available.
- Unfortunatelly many document loaders do not define `BlobParser` but rather handle everything directly inside the loader class. `UnstructuredLoader` is one of those. 
- Originally I went ahead and implemented separate `ExcelParser` (this PR). However this duplicated some of the code. That could be mitigated by replacing the part of the code inside `UnstructuredExcelLoader` responsible for parsing the file with calling the `Parser`. This would basically bring it to agreement with how things are done for ex. in PDF Loaders.
- However I found out that I'd have to do this shuffling for a fairly long list of document loaders. Given how much time such effort would require, I opted to create `DocumentLoaderAsParser` wrapper. This wrapper can be used on any DocumentLoader that accepts `file_path` argument and turns it into `BaseBlobParser`. 

Hope this makes sense.

PR #27109, opened 2024-10-04
anthropic[patch]: correctly handle tool msg with empty list

baskaryan:


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -118,9 +118,13 @@ def _merge_messages(
     for curr in messages:
         curr = curr.model_copy(deep=True)
         if isinstance(curr, ToolMessage):
-            if isinstance(curr.content, list) and all(
-                isinstance(block, dict) and block.get("type") == "tool_result"
-                for block in curr.content
+            if (
+                isinstance(curr.content, list)
+                and curr.content
```
efriis, line 123: nit: maybe worth adding a code comment for posterity

efriis (approved):
looks good. should we add testing for this to standard tests instead? seems like an edge case we should test across all providers

PR #27110, opened 2024-10-04
core,integrations[minor]: Dont error on fields in model_kwargs

baskaryan:
Given the current erroring behavior, every time we've moved a kwarg from model_kwargs and made it its own field that was a breaking change. Updating this behavior to support the old instantiations / serializations.

Assuming build_extra_kwargs was not something that itself is being used externally and needs to be kept backwards compatible


libs/partners/anthropic/tests/unit_tests/test_chat_models.py
```diff
@@ -59,8 +59,10 @@ def test_anthropic_model_kwargs() -> None:
 
 @pytest.mark.requires("anthropic")
 def test_anthropic_invalid_model_kwargs() -> None:
-    with pytest.raises(ValueError):
-        ChatAnthropic(model="foo", model_kwargs={"max_tokens_to_sample": 5})  # type: ignore[call-arg]
+    llm = ChatAnthropic(model="foo", model_kwargs={"max_tokens_to_sample": 5})  # type: ignore[call-arg]
```
eyurtsev, line 62: test no longer valid?

PR #27123, opened 2024-10-04
Docs: Add Comet Opik Update

sherpan:
-  **Description: Updated Comet's Docs for its new integration with Langchian via Opik 
- **Issue: None
- **Dependencies: None
 - **Twitter handle:  [@Cometml ](https://x.com/Cometml)


efriis:
pretty sure this is documenting a different package that's not a langchain integration - would recommend hosting these docs in your product!

sherpan:
Hi @efriis ,

Our product already has an older version that's mentioned on the Langchain docs. We just want to update the docs so its pointing to the  newer version of our offering!

PR #27136, opened 2024-10-05
community: Added splitting the Input to the allowed length of the Jina API.

krfdv:
- Description: Added splitting of input into blocks of no more than 2048 length in JinaEmbeddings._embed 
- Issue: The error in the response from Jina is that the input cannot be more than 2048 length if the input is longer


efriis:
hey! I don't think this makes sense because this would cause you to get a larger number of embeddings in return than the number of texts passed in.

I would also guess that Jina's api is based on tokens and not python `len()`

If you want to do splitting like this, you should split up your documents beforehand to the desired lengths with a text splitter!

PR #27138, opened 2024-10-05
feat: pass tool call id to tool runs

ha3:
Passes `tool_call_id` to the tools run during agent's execution. Without this, due to [this check](https://github.com/langchain-ai/langchain/blob/7a07196df683582c783edf164bfb6fe813135169/libs/core/langchain_core/tools/base.py#L884C5-L884C20), one can't access to the artifacts (if provided) from the result of the tool run.

### DIFF
--- a/libs/langchain/langchain/agents/agent.py
+++ b/libs/langchain/langchain/agents/agent.py
@@ -1444,6 +1444,7 @@ class AgentExecutor(Chain):
                 verbose=self.verbose,
                 color=color,
                 callbacks=run_manager.get_child() if run_manager else None,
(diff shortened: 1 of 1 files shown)


eyurtsev:
@ha3 we recommend that new code that involves agents be written using langgraph.

We'll fix bugs in the given agent executor if they arise, but generally reluctant to add new functionality, as the langchain agent executor is not the best way to write agent code.

If you'd like to get this PR merged, could you please:
* explain why we need to enhance the agent executor instead of using a langgraph agent?
* please add a unit test that covers the relevant scenario

PR #27151, opened 2024-10-06
community: verbose CSVLoader exception raise

ghost:
### Description:
Fix CSVLoader does not raise ValueError for missing metadata columns + tests

### Twitter handle:
@RaphaelGozlan1


efriis (commented):
I like this change in principle, but unfortunately it might have to wait for a minor bump because it will break people's code who are catching `RuntimeError` currently, and this change would cause `ValueError` to be thrown. Will discuss with the team and get back to you with either a PR merge or close.

If we do close, will add it to the list to consider for 0.4 (or more likely 1.0 further in the future)

ccurme:
Thank you for this! Closing but I've documented this PR in an (internal) list of desired breaking changes and can re-open in next version bump.

PR #27160, opened 2024-10-07
community: improve retrieval speed of BM25Retriever

jiroshimaya:
# PR Message

## Overview
- Significantly improved search speed of BM25Retriever (approximately 50 times faster for 100k documents retrieval).
- Initialization time increases by 2-3 times, but this is a one-time cost.
- Slight differences in search results may occur, but these are negligible for practical use.
- No API changes, so existing code can be used as-is.
- The scipy library is now required as an additional dependency, which can be installed via pip.

If this modification aligns with the principles of langchain, I would be grateful if it could be merged.

## Cause
In rank_bm25 used by BM25Retriever, BM25 scores are calculated using dictionary operations for each query, which may be causing the slow speed.

## Solution and Implementation
TFIDFRetriever is fast because it pre-computes weight matrices for the corpus and uses matrix operations to calculate similarities during searches.
I implemented a similar approach for BM25Retriever. Specifically, I made the following changes:
- Extended rank_bm25.BM25Okapi to add a method for converting sentences into weight vectors
  - Scipy library (installable via pip) is additionally used for efficient storage and computation of sparse matrices.
- In BM25Retriever.from_texts, created BM25 weight vectors for each document in the corpus and saved them as properties
- In BM25Retriever._get_relevant_documents, calculated BM25 scores as the dot product of the query word frequency vector and the corpus document weight vectors.

## Evaluation
When evaluated on a corpus of 100,000 documents, initialization became 2-3 times slower, but searching became about 50 times faster.
Since initialization can be completed before the user starts, search speed is more likely to affect user experience. Therefore, this improvement is considered useful for many use cases.

As the search process was changed, the BM25 scores no longer completely match those from before the change, but the difference was very small. Specifically, while the average value of the top 100 BM25 scores was around 30, the absolute difference in BM25 scores between the conventional method and the improved method was on the order of 10 to the minus 16th power.
There was also a slight impact on search results, but it seems to be at a negligible level for practical use. Specifically, when comparing the top 100 search results of BM25Retriever.invoke for 100 queries between the conventional and improved methods, 94 queries had completely matching top 100 search results. Of the remaining 6 queries, one had a difference in the 6th place result, and the other 5 all had differences in the 50th place result. Therefore, for applications like RAG that prioritize top search results, this difference appears to have minimal impact.

The source code used for evaluation, detailed results, and analysis are available for viewing at the following repository:
https://github.com/jiroshimaya/fast-langchain-sparse-retriever

# Checklist
- [x] **PR message**
  - PR message was written considering guidelines. 
  - no tests and docs added 
- [x] **Lint and test**

---
twitter handle: @shimajiroxyz

### DIFF
--- a/libs/community/langchain_community/retrievers/bm25.py
+++ b/libs/community/langchain_community/retrievers/bm25.py
@@ -1,17 +1,32 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Dict, Iterable, List, Optional
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional
 
 from langchain_core.callbacks import CallbackManagerForRetrieverRun
 from langchain_core.documents import Document
 from langchain_core.retrievers import BaseRetriever
 from pydantic import ConfigDict, Field
 
+if TYPE_CHECKING:
+    from scipy.sparse import csr_matrix
+
 
 def default_preprocessing_func(text: str) -> List[str]:
     return text.split()
 
-def create_bm25_vectorizer(corpus, **bm25_params):
+
+def create_bm25_vectorizer(corpus: Iterable[Iterable[str]], **bm25_params: Any) -> Any:
+    """
+    Create a BM25 vectorizer from the given corpus.
+
+    Args:
+        corpus: An iterable of tokenized documents, where each document is an iterable of strings.
+        **bm25_params: Additional parameters to be passed to the BM25Okapi constructor.
+
+    Returns:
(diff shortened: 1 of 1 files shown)


jiroshimaya:
I am closing this PR on my own accord.
In this PR, the score calculation is implemented on the langchain side for speed improvement. However, after some time and reconsideration, I realized that implementing it independently without using the standard API of rankbm25 could lead to unnecessarily high maintenance costs, which is not desirable.
If the maintainers are interested in this PR, feel free to reopen it.

PR #27163, opened 2024-10-07
Fix issue with _strip function extraction when both `sql and SQLQuery…

EricZhouaaaa:
I made the following changes to the `create_sql_query_chain` function to address issues with handling complex output formats:

### Problem Description
The original `create_sql_query_chain` function could not correctly extract SQL queries when both a ```sql code block and a `SQLQuery:` prefix were present in the output. This led to incomplete or incorrect extraction, especially when handling complex inputs with mixed output formats.

### Changes Made
1. **Prioritized Extraction from Code Blocks:**
   - Modified the function's logic to prioritize extracting SQL content from ```sql code blocks when both the code block and `SQLQuery:` prefix are present. This is because SQL in code blocks is typically more structured and explicit.

2. **Improved Regular Expression Matching:**
   - Enhanced the regular expressions used to match ```sql code blocks and `SQLQuery:` prefixes.
   - The new logic ensures that when there is a code block nested after a `SQLQuery:` prefix, it is also correctly extracted.

3. **Optimized Processing Order:**
   - Adjusted the processing order: first, it checks for ```sql code blocks, then the `SQLQuery:` prefix, and finally processes cases with neither.
   - This approach ensures that the extraction logic is more robust and can better handle mixed formats.

### Effects of the Changes
- The `create_sql_query_chain` function is now better at handling complex inputs that contain both ```sql code blocks and `SQLQuery:` prefixes.
- The function performs more reliably across different LLM output formats, reducing errors or incomplete extractions due to complex output formats.
- The improved logic provides users with more accurate SQL extraction capabilities.

### Additional Note
The example provided in the official documentation ([link to tutorial](https://python.langchain.com/docs/tutorials/sql_qa/)) 
 chapter Convert question to SQL querypreviously failed to run successfully. With the changes I made, it now runs correctly.

### DIFF
--- a/libs/langchain/langchain/chains/sql_database/query.py
+++ b/libs/langchain/langchain/chains/sql_database/query.py
@@ -13,7 +13,47 @@ if TYPE_CHECKING:
     from langchain_community.utilities.sql_database import SQLDatabase
 
 
+import re
+
+
 def _strip(text: str) -> str:
+    """
+    Extracts the pure SQL query from the LLM output.
+
+    Supports the following formats:
+    1. Contains a ```sql``` code block
+    2. Starts with "SQLQuery:"
+    3. Contains only an SQL statement
+
+    If both a ```sql``` code block and an "SQLQuery:" prefix are present, the SQL code block is prioritized.
+
+    Args:
+        text (str): The original output from the LLM.
+
+    Returns:
+        str: The extracted SQL query.
+    """
+    # Try to match all ```sql ... ``` code blocks
+    code_block_pattern = re.compile(r"```sql\s*(.*?)\s*```", re.DOTALL | re.IGNORECASE)
+    code_blocks = code_block_pattern.findall(text)
+    if code_blocks:
+        # If multiple code blocks are found, assume the first one is the desired SQL query
+        return code_blocks[0].strip()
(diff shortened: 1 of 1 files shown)


libs/langchain/langchain/chains/sql_database/query.py
```diff
@@ -13,7 +13,47 @@
     from langchain_community.utilities.sql_database import SQLDatabase
 
 
+import re
+
+
 def _strip(text: str) -> str:
```
eyurtsev, line 19: Currently this can be done in user code by suffixing a custom strip using LCEL composition.

I'm moving this PR into needs support.

If it gets enough up-votes, to be merged, it'll require:

1. unit tests
2. allowing to pass a custom strip function as part of the API itself (and maybe improving a default implementation together with a good explanation of the what the default implementation does / why its needed)

eyurtsev: The main problem is that it doesn't seem like this solution is general enough to be useful across many situations, so users will still likely need to use tackle a post-processor using LCEL composition even if this PR were to be merged

efriis:
closing with no upvotes or tests

PR #27173, opened 2024-10-07
Docs: conceptual docs batch 1

eyurtsev:
Re-organizing some of the content involves runnables/lcel/streaming into conceptual guides.

Conceptual guides added:
- [x] Runnables
- [x] LCEL
- [x] Chat Models
- [x] LLM
- [x] async
- [x] Messages
- [x] Chat History
- [x] Multimodality
- [x] Tokenization

Outstanding:
- [ ] Callbacks/Tracers
- [ ] Streaming
- [ ] Tool Creation
- [ ] Document Loading

Other conceptual guides are placeholders to make sure that no existing links breaks.

Some high level re-organization:

* Introduce the Runnable interface prior to LCEL (since those are two distinct concepts)
* Cross-link as much related content as possible (including how-to guides)


docs/docs/concepts/lcel.mdx
```diff
@@ -0,0 +1,106 @@
+# LangChain Expression Language (LCEL)
```
eyurtsev, line 1: This is WIP -- I want to reduce the Why LCEL section so it's faster to read tthrough

will expand later more detail about composition primitives etc

PR #27181, opened 2024-10-08
docs: fixed broken API reference link for StructuredTool.from_function

wkdghdus:
Fix broken API reference link for StructuredTool.from_function


eyurtsev (approved):
thank you

eyurtsev:
@wkdghdus thank you for the PR. 

We're also in the process of updating some of the tool documentation. In general, we're likely going to recommend just using the
`@tool` decorator instead of `StructuredTool.from_function` if you don't need to maintain state, and likely `BaseTool` if you do need to maintain state.

`@tool` decorator may also start replacing BaseTool since we should be able to apply it on methods / classes

eyurtsev:
cc @rlancemartin

PR #27197, opened 2024-10-08
infra: test 3.13 ci

efriis:


pyproject.toml
```diff
@@ -11,55 +11,16 @@ repository = "https://www.github.com/langchain-ai/langchain"
...
-langchain-experimental = { git = "https://github.com/langchain-ai/langchain-experimental.git", subdirectory = "libs/experimental" }
-langchain-anthropic = { path = "libs/partners/anthropic", develop = true }
-langchain-chroma = { path = "libs/partners/chroma", develop = true }
-langchain-mistralai = { path = "libs/partners/mistralai", develop = true }
-langgraph = { git = "https://github.com/langchain-ai/langgraph.git", subdirectory = "libs/langgraph" }
-jupyter = "^1.1.1"
-click = "^8.1.7"
-faiss-cpu = "^1.7.4"
-pypdf = "^5.0.0"
-vcrpy = "^6.0.1"
```
ccurme, line 58: these dependencies are needed for [notebook tests](https://github.com/langchain-ai/langchain/actions/workflows/run_notebooks.yml)

libs/community/langchain_community/agent_toolkits/nla/tool.py
```diff
@@ -13,7 +13,7 @@
 from langchain_community.utilities.requests import Requests
 
 
-class NLATool(Tool):
+class NLATool(Tool):  # type: ignore[override]
```
ccurme, line 16: ooc what is driving the need for the new `# type: ignores`? is it mypy getting updated?

efriis: Should we instead `@override` overridden methods? I believe that might be what it's trying to enforce

efriis: (happy to merge this now and do a sprint on fixing ignores like this)

libs/community/langchain_community/chat_models/azureml_endpoint.py
```diff
@@ -304,36 +304,38 @@ def _stream(
             "http_client": None,
         }
 
-        client = openai.OpenAI(**client_params)
+        client = openai.OpenAI(**client_params)  # type: ignore[arg-type, arg-type, arg-type, arg-type, arg-type, arg-type]
```
efriis, line 307: what does this mean?

efriis (commented):
lgtm

not sure I understand all the type annotations. another option would be to upper-bound mypy at the old locked version, get 3.13 in, and then work on the mypy upgrade separately. I believe they're 2 different things (3.13 was more tests, mypy upgrade was more community lint)

libs/community/langchain_community/agent_toolkits/nla/tool.py
baskaryan: @ccurme im not actually sure, afaict it was same mypy version. but issue was only happening in py3.13 so maybe im missing something

PR #27203, opened 2024-10-08
core[patch]: utils for adding/subtracting usage metadata

baskaryan:


efriis (approved):

libs/core/langchain_core/utils/usage.py
```diff
@@ -0,0 +1,80 @@
+import operator
+from typing import Callable, Optional, cast
+
+from langchain_core.messages.ai import UsageMetadata
+
+
+def add_usage(
```
eyurtsev, line 7: doc-strings would be meowed very much

PR #27206, opened 2024-10-08
openai[patch]: default temp=1 for o1

baskaryan:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -480,6 +480,15 @@ def build_extra(cls, values: Dict[str, Any]) -> Any:
         values = _build_model_kwargs(values, all_required_field_names)
         return values
 
+    @model_validator(mode="before")
+    @classmethod
+    def validate_temperature(cls, values: Dict[str, Any]) -> Any:
+        """Currently o1 models only allow temperature=1."""
+        model = values.get("model_name") or values.get("model") or ""
+        if "o1" in model and "temperature" not in values:
```
efriis, line 488:
```suggestion
        if model.startswith("o1-") and "temperature" not in values:
```

efriis: also should we handle the 0.7 default here so it's less confusing pydantic validator/default wise, and remove default on attribute above?

baskaryan: if we remove default or make it Optional might lead to linting errors for folks

baskaryan: (re o1 vs o1- as prefix, presumably openai will soon release just an "o1" model)

efriis (approved):
smh kids these days calling code improvements "fmt"

PR #27211, opened 2024-10-08
core: ruff encoding rule preview

efriis:


libs/core/langchain_core/rate_limiters.py
```diff
@@ -248,7 +248,7 @@ async def aacquire(self, *, blocking: bool = True) -> bool:
         if not blocking:
             return self._consume()
 
-        while not self._consume():
+        while not self._consume():  # noqa: ASYNC110
```
efriis, line 251: @eugene to confirm we don't want to reformat this as an asyncio.Event() rn

eyurtsev:
I need to review this to determine whether we're worried about the changes breaking any existing serialized data

eyurtsev (approved):
@efriis this is technically a breaking change for any user that has currently data that was serialized and saved on their machine in a non-utf 8 format.

PR #27228, opened 2024-10-09
community: Replace pypdf2 import with pypdf

Cokral:
Replace the dependency PyPDF2 with PyPDF. PyPDF2 is an old fork from PyPDF which hasn't been updated for 11years.

### DIFF
--- a/libs/community/langchain_community/document_loaders/googledrive.py
+++ b/libs/community/langchain_community/document_loaders/googledrive.py
@@ -335,7 +335,7 @@ class GoogleDriveLoader(BaseLoader, BaseModel):
             return docs
 
         else:
-            from PyPDF2 import PdfReader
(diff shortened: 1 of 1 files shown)


leo-gan:
It makes sense. But Don't we need to validate it somehow?

Cokral:
What do you think we should validate? I don't think we want to test that external packages work as expected, no?

ulgens:
Where `PyPDF2` coming in the first place? I couldn't find it as dependency.

efriis:
Hey there! This PR is against a deprecated community integration that has moved to a dedicated integration package.

I'll close it, and feel free to reopen against the integration package! https://github.com/langchain-ai/langchain-google

Fwiw I think this is a good change. Although PyPDF's docs don't make it easy to figure out which import to use!

PR #27241, opened 2024-10-10
docs: Migrate Makefile to Compose V2 syntax

jbding:
docs: Migrate Makefile to Compose V2 syntax
Description：The Makefile now reflects the change from using Compose V1 (docker-compose) to Compose V2 (docker compose).


efriis:
these aren't used by many people - I'd be more tempted to remove than upgrade!

PR #27078, opened 2024-10-03
community: Cassandra Vector Store: extend metadata-related methods

hemidactylus:
**Description:** this PR adds a set of methods to deal with metadata associated to the vector store entries. These, while essential to the Graph-related extension of the `Cassandra` vector store, are also useful in themselves. These are (all come in their sync+async versions):

- `[a]delete_by_metadata_filter`
- `[a]replace_metadata`
- `[a]get_by_document_id`
- `[a]metadata_search`

Additionally, a `[a]similarity_search_with_embedding_id_by_vector` method is introduced to better serve the store's internal working (esp. related to reranking logic).

**Issue:** no issue number, but now all Document's returned bear their `.id` consistently (as a consequence of a slight refactoring in how the raw entries read from DB are made back into `Document` instances).

**Dependencies:** (no new deps: packaging comes through langchain-core already; `cassio` is now required to be version 0.1.10+)

**Add tests and docs**
Added integration tests for the relevant newly-introduced methods.
(Docs will be updated in a separate PR).

**Lint and test** Lint and (updated) test all pass.

PR #27253, opened 2024-10-10
community: Cassandra Vector Store: modernize implementation

epinzur:
**Description:** 

This PR updates `CassandraGraphVectorStore` to be based off `CassandraVectorStore`, instead of using a custom CQL implementation. This allows users using a `CassandraVectorStore` to upgrade to a `GraphVectorStore` without having to change their database schema or re-embed documents.

This PR also updates the documentation of the `GraphVectorStore` base class and contains native async implementations for the standard graph methods: `traversal_search` and `mmr_traversal_search` in `CassandraVectorStore`.

**Issue:** No issue number.

**Dependencies:** https://github.com/langchain-ai/langchain/pull/27078 (already-merged)

**Lint and test**: 
- Lint and tests all pass, including existing `CassandraGraphVectorStore` tests.
- Also added numerous additional tests based of the tests in `langchain-astradb` which cover many more scenarios than the existing tests for `Cassandra` and `CassandraGraphVectorStore`

** BREAKING CHANGE**

Note that this is a breaking change for existing users of `CassandraGraphVectorStore`. They will need to wipe their database table and restart.  

However:
- The interfaces have not changed. Just the underlying storage mechanism.
- Any one using `langchain_community.vectorstores.Cassandra` can instead use `langchain_community.graph_vectorstores.CassandraGraphVectorStore` and they will gain Graph capabilities without having to re-embed their existing documents. This is the primary goal of this PR.


eyurtsev:
cc @cbornet  or @hemidactylus

@epinzur thank you for the contribution. I tagged the main maintainers for this integration. I took a quick glance, from a first pass it's not clear to me what this refactoring achieves. It does have a number of breaking changes as far as I can tell -- which could break existing users code.

hemidactylus:
@eyurtsev the author of this PR is a co-worker of mine (and of @cbornet's), we actually discussed several of the points addressed in this PR while it was being written.
The main assumption behind the changes here is twofold:

1. we posit that current usage of the Cassandra graph vector store (the one prior to this PR) is essentially nil, especially in production. Aided by the fact that the graph layer is still marked as beta, we consider the pros vastly outnumber the cons.
2. The pros: this PR makes the graph store something built on top of the (non-graph) "just vector store", `community.vectorstores.Cassandra`. This means, in particular, users would be able to "promote" their regular vector store into a graph store, without having to re-compute the embedding or re-ingest the stored entries - the graph structure consisting of certain metadata tags with the mechanism already supported by the graph store. To make this easily workable now - and maintainable in the future - some consistency between methods, signatures etc had to be reached.

efriis (approved):

PR #27267, opened 2024-10-11
[Community]update HF pipeline's StopSequenceCriteria

openvino-dev-samples:
To support Stop Sequence


efriis (commented):
hey! going to close this because this seems focused on supporting `stop_sequences: str` which is not supported (it should be `stop_sequences: List[str]`)

if you want to resubmit, could you add a unit test to `langchain-huggingface` in `libs/partners/huggingface` that fails currently, and add a description for what you're fixing?

PR #27278, opened 2024-10-11
docs: long context embedding example

leo-gan:
Addressed #25714 aka `Late Chunking`:
- created an example of the Late Chinking
- added a reference to this example in the `How-to guides` index page

### DIFF
--- a/docs/docs/how_to/index.mdx
+++ b/docs/docs/how_to/index.mdx
@@ -156,6 +156,7 @@ What LangChain calls [LLMs](/docs/concepts/#llms) are older forms of language mo
 
 - [How to: embed text data](/docs/how_to/embed_text)
 - [How to: cache embedding results](/docs/how_to/caching_embeddings)
+- [How to: long context embedding](/docs/how_to/long_context_embedding)
 
 ### Vector stores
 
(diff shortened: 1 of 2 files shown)


eyurtsev:
Hi @leo-gan , thank you for the contribution -- the how go guide isn't using langchain anywhere as far as I can tell so feels odd including this as a technique -- neither the text splitters nor existing embedding models are used. 

Can we illustrate the same ideas using existing text splitters or embedding models?

eyurtsev:
We should probably avoid including a how-to guide until we have proper conceptual information for this (i.e., we're re-working the concept pages during the next 2 weeks).

I'm vary of adding anything that's not actually captured by the concept hierarchy -- there's value in the documentation scope also not being too broad and cohesive

For example, users reading about this are assumed to understand what the technique is, whether they should use it or when they should use it, what kind of performance gains should one expect etc

leo-gan:
@eyurtsev I will rework it with LangChain text splitters and embedding. Yeah.
I'm not sure where this example should be placed. Please advise.

efriis:
@eyurtsev maybe somewhere in the conceptual docs?

although in general I'm not sure this belongs in the langchain docs

leo-gan:
It is one of techniques with a confusing name (it is not "long context embedding" exactly). I'm closing it.

PR #27327, opened 2024-10-14
community: fixes json loader not getting texts with json standard

TonyBotongChu:
This PR fixes JSONLoader._get_text not converting objects to json string correctly.
If an object is serializable and is not a dict, JSONLoader will use python built-in str() method to convert it to string. This may cause object converted to strings not following json standard. For example, a list will be converted to string with single quotes, and if json.loads try to load this string, it will cause error.


ccurme:
Tests failing due to how the new implementation handles string content.

TonyBotongChu:
Fixed the string content problem. And I also modified file_path in test_json_loader.py, to make sure tests will work on windows.

efriis (approved):

PR #27336, opened 2024-10-14
Langchain azure proposal one (single package)

marlenezw:
This PR is a draft of what `langchain_azure` could look like. This is a draft PR mainly for discussion purposes. 
Imports and names need to be changed for things to work as expected. 

This would allow users to `pip install langchain-azure`


efriis:
handled in external repo I believe (lmk if this shoudl have been kept open!)

PR #27351, opened 2024-10-15
community: add `AzureContentSafetyTextTool`

Sheepsta300:
community: add `AzureContentSafetyTextTool`

 - **Description:** With the ever-growing number of Azure AI services available, not all have been integrated with LangChain yet. This tool aims to expose one part of the Azure AI content safety service and allows customers to have input text get assessed against four criteria,  based on Azure's Content Safety Policy. As not all AI models may have their own inbuilt content safety feature this tool can be used in conjuction with other models as an agent, to assess and determine what to do with harmful content.
  
- **Dependencies:** no new dependencies for the tool, but azure.ai.contentsafety has been added to the extended-tests dependencies file.

  1. 3 unit tests have been supplied, mocking the underlying client. One testing the tools initialisation, one testing when harmful content is passed and one testing when no harmful content is passed.
  2. an example notebook showing its use has been created in the appropriate docs/tools folder, using headers enforced by the LangChain CI.

**Lint and test**: All linting passed.


docs/docs/integrations/tools/azure_content_safety.ipynb
```diff
@@ -0,0 +1,375 @@
+{
```
vbarda, line 1: this is a nice guide! few comments:

- could you add a URL to the azure context safety API / docs? (i see you have it in the docstring)
- could you add `getpass()` for environment variables to allow users to set them manually, like here https://python.langchain.com/docs/integrations/tools/bing_search/
- for the agent example, could you please use 

```python

agent = create_react_agent(model, tools)
agent.invoke({"messages": [("user",  f"Can you check the following text for harmful content : {input}")]})
```
this will also allow you to get rid of the structured chat prompt
- could you remove the chaining section?

libs/community/langchain_community/tools/azure_ai_services/content_safety.py
```diff
@@ -0,0 +1,125 @@
...
+        content_safety_key (str):
+            The API key used to authenticate requests with Azure Content Safety API.
+        content_safety_endpoint (str):
+            The endpoint URL for the Azure Content Safety API.
+        content_safety_client (Any):
+            An instance of the Azure Content Safety Client used for making API
+            requests.
+    """
+
+    content_safety_key: str = ""  #: :meta private:
```
vbarda, line 34: could we remove empty string defaults and make these required?

vbarda (commented):
Thanks for your contribution! Looks good to me -- added a few comments

vbarda:
@Sheepsta300 would you be able to make the changes?

Sheepsta300:
Sorry for taking a while, things have been busy lately! I've made all the changes you suggested I believe. By removing the ##Chaining header from the docs this does cause a linting error in the integration docs lint as it expects this header.

Let me know if anything else needs to be changed.

ccurme:
@Sheepsta300 thank you for iterating on this.

We are actively moving away from adding new integrations to community in favor of separating them into [integration packages](https://python.langchain.com/docs/concepts/architecture/#integration-packages). A package for Azure was recently added ([langchain-azure](https://github.com/langchain-ai/langchain-azure)).

Advantages of doing this include:

- Dependencies can be managed instead of catching import errors and recommending pip installs.
- Changes can be properly versioned.
- We can run integration tests.

I think the right thing to do is to implement this in `langchain-azure`. If we merge it into community it is liable to be deprecated once there's equivalent functionality in `langchain-azure`.

I notice you've made a few contributions around Azure into LangChain. I think these efforts could do a lot to boost the functionality of the Azure integration package.

Let me know what you think!

Sheepsta300:
Totally understand. I'll open a PR to the langchain-azure integration package soon, thanks for the info!

ccurme:
Closing for now.

PR #27368, opened 2024-10-15
core: add flake8-bandit (S) ruff rules to core

efriis:


libs/core/pyproject.toml
```diff
@@ -72,6 +72,7 @@ select = [
     "PIE",
     "Q",
     "RSE",
+    "S",
```
eyurtsev, line 75:
```suggestion
    "S", # https://docs.astral.sh/ruff/rules/#flake8-bandit-s
```

eyurtsev (approved):

Issue #27269, opened 2024-10-11
[Azure CosmosDB NoSQL] similarity_search_with_score function returning CosmosHttpResponseError "bad request"


peacemaker360:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Sample for reproducing the issue
```python
import os
from dotenv import load_dotenv
from azure.cosmos import CosmosClient, PartitionKey
from langchain_community.vectorstores.azure_cosmos_db_no_sql import (
    AzureCosmosDBNoSqlVectorSearch,
)
from langchain_openai import OpenAIEmbeddings

# from azure_cosmos_db_no_sql import AzureCosmosDBNoSqlVectorSearch # workarround with modifications to code

_ = load_dotenv()

HOST = os.environ['AZURE_COSMOS_URL']
KEY = os.environ['AZURE_COSMOS_KEY']

cosmos_client = CosmosClient(HOST, KEY)
database_name = "db"
container_name = "container1"
partition_key = PartitionKey(path="/id")
cosmos_container_properties = {"partition_key": partition_key}

openai_embeddings = OpenAIEmbeddings(model="text-embedding-3-large")

vectorstore = AzureCosmosDBNoSqlVectorSearch(
    cosmos_client=cosmos_client,
    embedding=openai_embeddings,
    vector_embedding_policy={
        "vectorEmbeddings": [
            {
                "path": "/embedding",
                "dataType": "float32",
                "distanceFunction": "cosine",
                "dimensions": 3072,
            }
        ]
    },
    indexing_policy={
        "indexingMode": "consistent",
        "includedPaths": [{"path": "/*"}],
        "excludedPaths": [{"path": '/"_etag"/?'}],
        "vectorIndexes": [{"path": "/embedding", "type": "quantizedFlat"}],
    },
    cosmos_container_properties=cosmos_container_properties,
    cosmos_database_properties={'id': database_name},
    database_name=database_name,
    container_name=container_name,
    create_container=False,
)

query = "What were the compute requirements for training GPT 4"

results = vectorstore.similarity_search_with_score(
    query=query,
)
````

### Error Message and Stack Trace (if applicable)

{
[...]

Issue #25468, opened 2024-08-16
Syntax error, incorrect syntax near '{'."} for Azure Cosmos DB No SQL tutorial


hepengfe:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Following tutorial here https://python.langchain.com/v0.2/docs/integrations/vectorstores/azure_cosmos_db_no_sql/

After fixing the code as instructed here by adding 
```python
cosmos_database_properties={
        "id" : Config.COSMOS_DB_NAME
    }
```
Another issue occur and the message indicates some syntax error.

### Error Message and Stack Trace (if applicable)

page_content='Direct Preference Optimization:
Your Language Model is Secretly a Reward Model
Rafael Rafailov∗†Archit Sharma∗†Eric Mitchell∗†
Stefano Ermon†‡Christopher D. Manning†Chelsea Finn†
†Stanford University‡CZ Biohub
{rafailov,architsh,eric.mitchell}@cs.stanford.edu
Abstract
While large-scale unsupervised language models (LMs) learn broad world knowl-
edge and some reasoning skills, achieving precise control of their behavior is
difficult due to the completely unsupervised nature of their training. Existing
methods for gaining such steerability collect human labels of the relative quality of
model generations and fine-tune the unsupervised LM to align with these prefer-
ences, often with reinforcement learning from human feedback (RLHF). However,
RLHF is a complex and often unstable procedure, first fitting a reward model that
reflects the human preferences, and then fine-tuning the large unsupervised LM
using reinforcement learning to maximize this estimated reward without drifting' metadata={'source': 'https://arxiv.org/pdf/2305.18290', 'page': 0}
Traceback (most recent call last):
  File "/home/pii/adabit/proj/proj/src/db/cosmos_db_langchain.py", line 114, in <module>
    results = vectorstore.similarity_search(query)
  File "/home/pii/miniconda3/envs/proj/lib/python3.9/site-packages/langchain_community/vectorstores/azure_cosmos_db_no_sql.py", line 338, in similarity_search
    docs_and_scores = self.similarity_search_with_score(
[...]

PR #27377, opened 2024-10-15
community: fix @embeddingKey in azure cosmos db no sql

DBPhoenix:
I will keep this PR as small as the changes made.

**Description:** fixes a fatal bug syntax error in AzureCosmosDBNoSqlVectorSearch
**Issue:** #27269 #25468

### DIFF
--- a/libs/community/langchain_community/vectorstores/azure_cosmos_db_no_sql.py
+++ b/libs/community/langchain_community/vectorstores/azure_cosmos_db_no_sql.py
@@ -274,15 +274,15 @@ class AzureCosmosDBNoSqlVectorSearch(VectorStore):
             query += "TOP @limit "
 
         query += (
-            "c.id, c.{}, c.text, c.metadata, "
-            "VectorDistance(c.@embeddingKey, @embeddings) AS SimilarityScore FROM c"
+            "c.id, c[@embeddingKey], c.text, c.metadata, "
+            "VectorDistance(c[@embeddingKey], @embeddings) AS SimilarityScore FROM c"
         )
 
         # Add where_clause if specified
         if pre_filter is not None and pre_filter.get("where_clause") is not None:
             query += " {}".format(pre_filter["where_clause"])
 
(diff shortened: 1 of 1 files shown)


codenamics:
when it will be merged?

DBPhoenix:
> when it will be merged?

I am no expert so anyone should feel free to correct me. But it seems the status "Needs support" means that the PR is on hold until it reaches more engagement. Therefore check the changes out, confirm that they work for you, and use the +1 on the post to voice your opinion.

libs/community/langchain_community/vectorstores/azure_cosmos_db_no_sql.py
```diff
@@ -274,15 +274,15 @@ def _similarity_search_with_score(
...
+            "c.id, c[@embeddingKey], c.text, c.metadata, "
+            "VectorDistance(c[@embeddingKey], @embeddings) AS SimilarityScore FROM c"
         )
 
         # Add where_clause if specified
         if pre_filter is not None and pre_filter.get("where_clause") is not None:
             query += " {}".format(pre_filter["where_clause"])
 
-        query += " ORDER BY VectorDistance(c.@embeddingKey, @embeddings)"
+        query += " ORDER BY VectorDistance(c[@embeddingKey], @embeddings)"
```
ccurme, line 285: I'm not very familiar with Azure CDB NoSQL but is `c.@embeddingKey` equivalent to `c[@embeddingKey]`? Why make this change?

ccurme (approved):

libs/community/langchain_community/vectorstores/azure_cosmos_db_no_sql.py
DBPhoenix: As far as I know, the dot syntax is not supported for variable expressions. You are therefore forced to use brackets if you want to protect yourself against injections. CDB NoSQL seems quite restrictive in its capabilities here. But I do not have greater proficiency in it either, so there may be smarter ways about it.

Issue #27365, opened 2024-10-15
When unicode character is in API key, non-specific error is returned (instead of invalid API key)


jasonkaedingrhino:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
from langchain_openai import ChatOpenAI
from langchain_core.messages import HumanMessage

#######################
#### CONFIGURATION ####
#######################

configs = dict(
    model = 'gpt-4-turbo-preview',
    openai_api_key = 'здравейздравейздравейздравей',  # This is the bad one
)

client = ChatOpenAI(**configs)

######################
###### RUN TEST ######
######################

message = 'Hello!'

response = client.invoke([HumanMessage(content=message)])

print(response)
```

### Error Message and Stack Trace (if applicable)

```
Traceback (most recent call last):
  File "C:\Users\jason\Documents\source\bug-test\venv\working\langchain-api-key-bug\test-basic.py", line 23, in <module>
    response = client.invoke([HumanMessage(content=message)])
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jason\Documents\source\bug-test\venv\Lib\site-packages\langchain_core\language_models\chat_models.py", line 286, in invoke
    self.generate_prompt(
  File "C:\Users\jason\Documents\source\bug-test\venv\Lib\site-packages\langchain_core\language_models\chat_models.py", line 786, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jason\Documents\source\bug-test\venv\Lib\site-packages\langchain_core\language_models\chat_models.py", line 643, in generate
    raise e
  File "C:\Users\jason\Documents\source\bug-test\venv\Lib\site-packages\langchain_core\language_models\chat_models.py", line 633, in generate
    self._generate_with_cache(
[...]

PR #27378, opened 2024-10-15
partners: validate api key

AndrewEffendi:
fixes issue #27365
Adds an error message notifying the user that they are using an invalid character in their api key

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -6,6 +6,7 @@ import base64
 import json
 import logging
 import os
+import re
 import sys
 import warnings
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


efriis:
would rather not in case openai changes their api key format!

jasonkaedingrhino:
> would rather not in case openai changes their api key format!

@efriis 
I agree that a regex test is too heavy-handed. However, would it be appropriate to "check" the API key from an encoding perspective, since just blindly passing strings down into the HTTP client is where we end up with esoteric HTTP client errors about encodings?

AndrewEffendi:
What about instead of using regex, we check if it raise a UnicodeEncodeError when we try to encode the api key?
```python
if self.openai_api_key:
    api_key = self.openai_api_key.get_secret_value()
    try:
        api_key.encode('ascii')
    except UnicodeEncodeError:
        raise ValueError("Invalid API key: Only ASCII characters are allowed.")
```

jasonkaedingrhino:
FWIW - I have since realized that the stack trace is actually pointing back down to `openai` library directly.

I decided to open an [issue](https://github.com/openai/openai-python/issues/1793) there.

Issue #27325, opened 2024-10-14
langchain_openai reports error `role Input should be a valid string` when connecting to specific LLM API


Moskize91:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_openai import ChatOpenAI

api_key = "XXXXXXXX"
model_name="gpt-3.5-turbo"
base_url = "https://XXXXXX" # specific LLM vendor
llm = ChatOpenAI(
    api_key=api_key,
    base_url=base_url
    model=model_name,
)
llm.invoke("hello world")
```

### Error Message and Stack Trace (if applicable)

Traceback (most recent call last):
  File "/Users/taozeyu/codes/github.com/moksize91/llm-inception/server/llm/node.py", line 320, in _invoke
    resp_content = chain.invoke(
                   ^^^^^^^^^^^^^
  File "/Users/taozeyu/codes/github.com/moksize91/llm-inception/.venv/lib/python3.12/site-packages/langchain_core/runnables/base.py", line 3022, in invoke
    input = context.run(step.invoke, input, config, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/taozeyu/codes/github.com/moksize91/llm-inception/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 286, in invoke
    self.generate_prompt(
  File "/Users/taozeyu/codes/github.com/moksize91/llm-inception/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 786, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/taozeyu/codes/github.com/moksize91/llm-inception/.venv/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 643, in generate
    raise e
[...]

PR #27398, opened 2024-10-16
[Partner-OpenAI]: Default Role to `assistant` if not provided in `_convert_dict_to_message`

keenborder786:
- **Description:** Some vendors might not provide a role in the message, resulting in a null value and causing a validation error when converting it to a message. If the role is not provided, I will assume it should be `assistant`, indicating that it is coming from AI.
- **Issue:** #27325


eyurtsev (changes requested):
HI @keenborder786, I am not convinced yet that we want to make this change. Let's get more evidence from the community to see that this is an issue with other LLM providers. I don't love changing correct code into incorrect code just b/c some provider got the wrong format specification. This can lead to other subtle errors down the road.


----

Also, we generally expect all patches to be accompanied by unit-tests, especially in important integration packages like openai.

PR #27406, opened 2024-10-16
Generate vectors with only positive values:langchain_community.embeddings

yorek:
**Description:** 
Generate vectors for which cosine distance will always be between 0 and 1, to comply with similarity_search_with_relevance_scores requirement of returning only values between 0 and 1

**Issue:** 
N/A

**Dependencies:** 
None


efriis:
hey! some embedding functions return negative values, so the fake implementation shouldn't be prevented from either.

if you need positive for a specific case, you could implement your own embedding class, or you could add a param that forces this to the class

Issue #26617, opened 2024-09-18
OpenAI Chat isn't forward compatible with OpenAI API


adubovik:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

requirements.txt:
```
langchain==0.3.0
langchain-core==0.3.1
langchain-openai==0.2.0
openai==1.46.0
httpx==0.27.2
```

test.py:
```python
from langchain_openai import ChatOpenAI
from langchain_openai.chat_models.base import _convert_dict_to_message, _convert_message_to_dict
from langchain_core.messages import HumanMessage, BaseMessage
import httpx
from pydantic import SecretStr


class MockClient(httpx.Client):
    def send(self, request, **kwargs):
        # !!! No extra_request field
        print(f"Request: {request.content.decode()}")

        message = {
            "role": "assistant",
            "content": "answer",
            "extra_response": "extra_response",
        }

        # !!! No extra_response field
        print(f"Response message dict: {_convert_dict_to_message(message)}")

        return httpx.Response(
            request=request,
            status_code=200,
            headers={"Content-Type": "application/json"},
            json={"choices": [{"index": 0, "message": message}]},
        )


llm = ChatOpenAI(api_key=SecretStr("-"), http_client=MockClient())

request_message = HumanMessage(content="question", extra_request="extra_request")

# !!! No extra_request field
print(f"Request message dict: {_convert_message_to_dict(request_message)}")

output = llm.generate(messages=[[request_message]])

response_message: BaseMessage = output.generations[0][0].message

# !!! No extra_response field
print(f"Response message: {response_message}")
```

```sh
> python test.py
Request message dict: {'content': 'question', 'role': 'user'}
Request: {"messages": [{"content": "question", "role": "user"}], "model": "gpt-3.5-turbo", "n": 1, "stream": false, "temperature": 0.7}
Response message dict: content='answer' additional_kwargs={} response_metadata={}
[...]

PR #27441, opened 2024-10-17
openai: convert* now forwards compatible

mmmelkadri:
**Description:** Despite being able to add extra fields to a message using a message's constructor:

`ToolMessage(content="foo", tool_call_id="bar", extra_field="extra")`

Both `_convert_dict_to_message` and `_convert_message_to_dict` discard any extra fields. This PR fixes this, allowing for users to add extra fields that may be supported by an OpenAI API but not in LangChain yet, making these functions forward compatible.

**Issue:** #26617 

**Tests:** Unit tests have been added testing converting to and from dictionaries for messages.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -106,21 +106,26 @@ def _convert_dict_to_message(_dict: Mapping[str, Any]) -> BaseMessage:
     Returns:
         The LangChain message.
     """
-    role = _dict.get("role")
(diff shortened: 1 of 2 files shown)


efriis (commented):
I'd be ok doing this under a flag on ChatOpenAI, but unfortunately as-is, this is a breaking change. Especially when passing messages between multiple providers. E.g. an AI message that comes back from Anthropic might have metadata attached that isn't compatible with openai.

If there are specific params we're missing, let me know!

PR #27449, opened 2024-10-18
community: Outlines integration

shroominic:
In collaboration with @rlouf I build an [outlines](https://dottxt-ai.github.io/outlines/latest/) integration for langchain!

I think this is really useful for doing any type of structured output locally.
[Dottxt](https://dottxt.co) spend alot of work optimising this process at a lower level ([outlines-core](https://pypi.org/project/outlines-core/0.1.14/) written in rust) so I think this is a better alternative over all current approaches in langchain to do structured output.
It also implements the `.with_structured_output` method so it should be a drop in replacement for a lot of applications.

The integration includes:
- **Outlines LLM class**
- **ChatOutlines class**
- **Tutorial Cookbooks**
- **Documentation Page**
- **Validation and error messages** 
- **Exposes Outlines Structured output features**
- **Support for multiple backends**
- **Integration and Unit Tests**

Dependencies: `outlines` + additional (depending on backend used)

I am not sure if the unit-tests comply with all requirements, if not I suggest to just remove them since I don't see a useful way to do it differently.

### Quick overview:

Chat Models:
image

Structured Output:
image


shroominic:
Either I am stupid or the `Integration docs lint / build (pull_request)` check is broken...
It showes ValueError: Document /home/runner/work/langchain/langchain/docs/docs/integrations/chat/outlines.ipynb is missing headers:
- ### Installation
But the header is definitlely there!

shroominic:
The integration tests are currently running local models to check if everything is working. Is this possible in the testing environment or should the tests be written differently?

cpfiffer:
> Either I am stupid or the `Integration docs lint / build (pull_request)` check is broken... It showes ValueError: Document /home/runner/work/langchain/langchain/docs/docs/integrations/chat/outlines.ipynb is missing headers:
> 
> * ### Installation
> 
> But the header is definitlely there!

I'm surprised by this too -- the header is included for sure.

cpfiffer:
Ah, I see. I believe the order of credentials and installation should be switched, as per [this doc](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/chat.ipynb).

shroominic:
> Ah, I see. I believe the order of credentials and installation should be switched, as per [this doc](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/chat.ipynb).

I tried this and run the script locally but still didn't work ...

cpfiffer:
Could y'all wait a bit before merging? I want some Outlines people from @dottxt-ai to take a look at this to see how we feel about the interface.

rlouf:
Looks good to me 🫡

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+        raise ValueError("Tokenizer not found")
+
+    def _convert_messages_to_openai_format(
+        self, messages: list[BaseMessage]
+    ) -> list[dict]:
+        return [convert_message_to_dict(message) for message in messages]
+
+    def _convert_messages_to_prompt(self, messages: list[BaseMessage]) -> str:
+        """Convert a list of messages to a single prompt."""
+        # if hasattr(self.tokenizer, "apply_chat_template"):
```
vbarda, line 343: could you please remove the commented code?

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+            model_card = ModelCard.load(f"{repo_creator}/{gguf_repo_name}")
+            if hasattr(model_card.data, "base_model"):
+                model_name = model_card.data.base_model
+            else:
+                raise ValueError(f"Base model name not found for {self.model}")
+        else:
+            model_name = self.model
+
+        from transformers import AutoTokenizer
+        # from llama_cpp.llama_tokenizer import LlamaHFTokenizer
```
vbarda, line 360: same here

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> ChatResult:
+        params = {**self._default_params, **kwargs}
+        if stop:
+            params["stop_at"] = stop
+
+        prompt = self._convert_messages_to_prompt(messages)
+        # todo? stop = return assistant token mask
```
vbarda, line 470: could you please turn this into a more descriptive TODO?

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+        if backend == "llamacpp":
+            check_packages_installed([("llama-cpp-python", "llama_cpp")])
+            if ".gguf" in model:
+                creator, repo_name, file_name = model.split("/", 2)
+                repo_id = f"{creator}/{repo_name}"
+            else:
+                raise ValueError("GGUF file_name must be provided for llama.cpp.")
+            self.client = models.llamacpp(repo_id, file_name, **self.model_kwargs)
+        elif backend == "transformers":
+            check_packages_installed(["transformers", "torch", "datasets"])
```
vbarda, line 247: doesn't transformers install `torch` by default? (and maybe `datasets` too)

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+            "regex": self.regex,
+            "type_constraints": self.type_constraints,
+            "json_schema": self.json_schema,
+            "grammar": self.grammar,
+            **self._default_params,
+        }
+
+    @property
+    def _generator(self) -> Any:
+        from outlines import generate
```
vbarda, line 298: let's move this to top-level imports

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+            return self.custom_generator
+        constraints = [
+            self.regex,
+            self.type_constraints,
+            self.json_schema,
+            self.grammar,
+        ]
+
+        num_constraints = sum(constraint is not None for constraint in constraints)
+        if num_constraints != 1 and num_constraints != 0:
```
vbarda, line 310: we're already doing this validation in `validate_environment` -- could we remove?

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+        if self.type_constraints:
+            return generate.format(self.client, python_type=self.type_constraints)
+        if self.json_schema:
+            return generate.json(self.client, schema_object=self.json_schema)
+        if self.grammar:
+            return generate.cfg(self.client, cfg_str=self.grammar)
+        return generate.text(self.client)
+
+    @property
+    def tokenizer(self) -> Any:
```
vbarda, line 326: why is this needed? as opposed to accessing `self.client.tokenizer`?

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+        )
+
+    def bind_tools(
+        self,
+        tools: Sequence[Dict[str, Any] | type | Callable[..., Any] | BaseTool],
+        *,
+        tool_choice: Optional[Union[Dict, bool, str]] = None,
+        **kwargs: Any,
+    ) -> Runnable[
+        PromptValue
```
vbarda, line 376: can we use `from langchain_core.language_models import LanguageModelInput` instead

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+        return super().bind_tools(tools=formatted_tools, **kwargs)
+
+    def with_structured_output(
+        self,
+        schema: Dict | type,
+        *,
+        include_raw: bool = False,
+        **kwargs: Any,
+    ) -> Runnable[
+        Union[
```
vbarda, line 426: let's use from `langchain_core.language_models import LanguageModelInput`

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+            )
+        else:
+            parser = JsonOutputParser()
+
+        if include_raw:
+            return self | {
+                "raw": RunnablePassthrough(),
+                "parsed": parser,
+                "parsing_error": RunnableLambda(
+                    lambda x: None,  # todo add errors
```
vbarda, line 452: let's remove and just keep a TODO

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+                    )
+                tool_choice = formatted_tools[0]
+
+        kwargs["tool_choice"] = tool_choice
+        formatted_tools = [convert_to_openai_tool(tool) for tool in tools]
+        return super().bind_tools(tools=formatted_tools, **kwargs)
+
+    def with_structured_output(
+        self,
+        schema: Dict | type,
```
vbarda, line 421: could we add a better type annotation here? e.g. https://github.com/langchain-ai/langchain/blob/master/libs/partners/openai/langchain_openai/chat_models/base.py#L1142

libs/community/langchain_community/chat_models/outlines.py
```diff
@@ -0,0 +1,564 @@
...
+                creator, repo_name, file_name = model.split("/", 2)
+                repo_id = f"{creator}/{repo_name}"
+            else:
+                raise ValueError("GGUF file_name must be provided for llama.cpp.")
+            self.client = models.llamacpp(repo_id, file_name, **self.model_kwargs)
+        elif backend == "transformers":
+            check_packages_installed(["transformers", "torch", "datasets"])
+            self.client = models.transformers(model_name=model, **self.model_kwargs)
+        elif backend == "transformers_vision":
+            from transformers import LlavaNextForConditionalGeneration
```
vbarda, line 250: should probably be imported inside the `if` below

vbarda (commented):
Thanks for the contribution -- looks good overall! Added a few comments

vbarda:
Could you add some unittests as well?

shroominic:
> Could you add some unittests as well?

I added unit-tests before but removed them because I could not find a useful way to do them without requiring network connection since outlines requires local models and you initialize the chat model with a hugging face model id.
https://github.com/langchain-ai/langchain/blob/30071e3182459caa975c16a84b3e22ab7e2a94d3/libs/community/tests/unit_tests/llms/test_outlines.py

libs/community/langchain_community/chat_models/outlines.py
shroominic: https://github.com/huggingface/transformers/blob/b53e44e847230a39caecc320527776ed80a3ceff/setup.py#L279-L287
they are extras

libs/community/langchain_community/chat_models/outlines.py
shroominic: people can still change the constraints on the chat model instance

libs/community/langchain_community/chat_models/outlines.py
shroominic: I think it can be removed was not sure about the stop token but everything works

libs/community/langchain_community/chat_models/outlines.py
shroominic: i saw in all other chat models that client libraries were never imported at the top level why here?

libs/community/langchain_community/chat_models/outlines.py
vbarda: makes sense -- could we use a shared util then?

vbarda:
> > Could you add some unittests as well?
> 
> I added unit-tests before but removed them because I could not find a useful way to do them without requiring network connection since outlines requires local models and you initialize the chat model with a hugging face model id. https://github.com/langchain-ai/langchain/blob/30071e3182459caa975c16a84b3e22ab7e2a94d3/libs/community/tests/unit_tests/llms/test_outlines.py

yea, you would probably need to do some mocking for the model interfaces to avoid network connections or using local models. primarily it would be nice to test different scenarios for constraints validation etc. you could also factor some of those out outside the classes and that would also make it easier to test

libs/community/langchain_community/chat_models/outlines.py
vbarda: ah you're correct actually, in `langchain-community` this is the right way to do it

shroominic:
> > > Could you add some unittests as well?
> > 
> > 
> > I added unit-tests before but removed them because I could not find a useful way to do them without requiring network connection since outlines requires local models and you initialize the chat model with a hugging face model id. https://github.com/langchain-ai/langchain/blob/30071e3182459caa975c16a84b3e22ab7e2a94d3/libs/community/tests/unit_tests/llms/test_outlines.py
> 
> yea, you would probably need to do some mocking for the model interfaces to avoid network connections or using local models. primarily it would be nice to test different scenarios for constraints validation etc. you could also factor some of those out outside the classes and that would also make it easier to test

✅

vbarda:
@shroominic thanks for adding the tests! i think it's almost ready to merge -- docs page for LLMs is failing (the chat model one is correct)

`ValueError: Document /home/runner/work/langchain/langchain/docs/docs/integrations/llms/outlines.ipynb is missing headers`

shroominic:
everything passing 🙌

PR #27460, opened 2024-10-18
core[patch]: Improve type checking for the tool decorator

nvachhar:
**Description:**

When annotating a function with the @tool decorator, the symbol should have type BaseTool. The previous type annotations did not convey that to type checkers. This patch creates 4 overloads for the tool function for the 4 different use cases.

1. @tool decorator with no arguments
2. @tool decorator with only keyword arguments
3. @tool decorator with a name argument (and possibly keyword arguments)
4. Invoking tool as function with a name and runnable positional arguments

The main function is updated to match the overloads. The changes are 100% backwards compatible (all existing calls should continue to work, just with better type annotations).

**Twitter handle:** @nvachhar

### DIFF
--- a/libs/core/langchain_core/tools/convert.py
+++ b/libs/core/langchain_core/tools/convert.py
@@ -1,5 +1,5 @@
 import inspect
-from typing import Any, Callable, Literal, Optional, Union, get_type_hints
+from typing import Any, Callable, Literal, Optional, Union, get_type_hints, overload
 
 from pydantic import BaseModel, Field, create_model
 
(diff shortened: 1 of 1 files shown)


eyurtsev:
@nvachhar thank you very much! I will take a look earliest Monday, but this is something that I've wanted to fix for a while!

eyurtsev:
@navalchand okay this looks good overall (don't worry about fixing lint -- i'll take care of it)... i need to confirm that support for runnables is indeed intentional

nvachhar:
@eyurtsev - thank you! Looking forward to having this merged in.

eyurtsev (approved):

eyurtsev:
Made a small update to doc-strings and renamed some internal variables in attempt to make it easier to follow what's going on inside the decorator, not sure if I was entirely successful

libs/core/langchain_core/tools/convert.py
```diff
@@ -263,37 +275,61 @@ def invoke_wrapper(
...
-        return _make_tool
+        return _tool_factory
 
     if len(args) != 0:
-        msg = "Too many arguments for tool decorator"
+        # Triggered if a user attempts to use positional arguments that
+        # do not exist in the function signature
+        # e.g., @tool("name", runnable, "extra_arg")
+        # Here, "extra_arg" is not a valid argument
+        msg = "Too many arguments for tool decorator. A decorator "
```
nvachhar, line 285: Looks like this message got cut off.

nvachhar:
Thanks for merging this!

PR #27462, opened 2024-10-18
community: fix query method of OntotextGraphDBGraph for ASK, DESCRIBE and CONSTRUCT type

harrytran001:
**Description:**
I am trying to fix a query method of OntotextGraphDBGraph since it can handle `SELECT` query type only

**Issue:**
As I tried QA on SparQL Database with a dataset from [Github](https://github.com/lilyhoanghg/MCQA). There are many Yes/No questions which could be answer by `ASK` query. However, ASK query was always giving an empty.

**Replicate Issue**
I tried to replicate the issue on Langchain SparQL dataset below
```python
graph = OntotextGraphDBGraph(
  query_endpoint="http://localhost:7200/repositories/langchain",
  query_ontology="CONSTRUCT {?s ?p ?o}"
  "FROM <https://swapi.co/ontology/> WHERE {?s ?p ?o}",
)

query_result = graph.query(
    "PREFIX : <https://swapi.co/vocabulary/>"
    "PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>"
    "ASK {"
    "  ?species a :Species ;"
    '    rdfs:label "Besalisk" ;'
    "    :averageHeight 178.0 ;"
    '    :averageLifespan "75" ;'
    '    :eyeColor "yellow" .'
    "}"
)

# query_result is []
```
The result must be `True` instead of an empty array, which later will be passed to GraphQA chain and receive undesirable results. 

**Underlying Issue**
The underlying issue was that there are 4 types query in SparQL: `SELECT`, `ASK` , `DESCRIBE` and `CONSTRUCT`. However, only `SELECT` query type was handled in OntotextGraphDBGraph.
As per `rdflib` documentation, there are 3 types of responses, as below.

Screenshot 2024-10-19 at 12 24 44 AM

Hence, I tried to handle the case where the query type is `ASK` or `DESCRIBE` or `CONSTRUCT`.

*Dependencies:*
None

- [x] **Add tests**

- [x] **Lint and test**:


harrytran001:
@eyurtsev Could you help me to review. Thanks.

efriis:
hey sorry for the delay! This one looks like it's a bit brittle based on the query parsing, so it probably isn't mergable in its current form. I think for custom parsing of different query types, it's probably better to execute queries manually with e.g. `ontotextgraph.graph.query(...)` and parse them yourself. depending on the query you're generating, it could be a lot of different shapes on the output.

If you work for ontotextgraph also - we'd love to migrate this into a dedicated integration package

PR #27466, opened 2024-10-18
text-splitters: refactor if-elif to mapping

Zephyruss1:
**Description:** Using mapping to accelerate process


eyurtsev (changes requested):
is this a meaningful change in terms of any bottlenecks?

OK making a change if it speed something up in a meaningful way, but we don't want to make changes that aren't backed by evidence or actual need (consumes time from maintainers to review / merge, and increases risk of something breaking)

Zephyruss1:
Hello @eyurtsev,

I used mapping to avoid unnecessary transitions between the (if-elif-else) blocks. I believe that using mapping in this case is beneficial because there are no complex statements to check. Even if the difference is slight, mapping operates faster.

How can I provide you with evidence to verify this?

Issue #27500, opened 2024-10-21
StructuredTool.from_function does not support functools.partial


coolbeevip:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
def test_structured_tool_from_function_partial() -> None:
    """Test that structured tools can be created from a partial function."""

    def func(bar: str, baz: str=None) -> str:
        """Docstring

        Args:
            bar: str
            baz: str
        """
        return bar + baz

    structured_tool = StructuredTool.from_function(
        name="tool",
        description="A tool",
        func=partial(func, baz="foo"),
    )
    assert structured_tool.invoke({"bar": "bar"}) == "barfoo"
```

### Error Message and Stack Trace (if applicable)

```
tests/unit_tests/test_tools.py:484 (test_structured_tool_from_function_partial)
def test_structured_tool_from_function_partial() -> None:
        """Test that structured tools can be created from a partial function."""
    
        def func(bar: str, baz: str=None) -> str:
            """Docstring
    
            Args:
                bar: str
                baz: str
            """
            return bar + baz
    
>       structured_tool = StructuredTool.from_function(
            name="tool",
            description="A tool",
            func=partial(func, baz="foo"),
        )

test_tools.py:497: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../langchain_core/tools/structured.py:181: in from_function
    args_schema = create_schema_from_function(
../../langchain_core/tools/base.py:249: in create_schema_from_function
    validated = validate_arguments(func, config=_SchemaConfig)  # type: ignore
../../../../venv/lib/python3.11/site-packages/pydantic/deprecated/decorator.py:64: in validate_arguments
    return validate(func)
../../../../venv/lib/python3.11/site-packages/pydantic/deprecated/decorator.py:51: in validate
    vd = ValidatedFunction(_func, config)
../../../../venv/lib/python3.11/site-packages/pydantic/deprecated/decorator.py:94: in __init__
    type_hints = _typing_extra.get_type_hints(function, include_extras=True)
[...]

PR #27501, opened 2024-10-21
core: Add support for creating structured tools from partial functions

coolbeevip:
- **Description:** StructuredTool.from_function should check the type of func before parsing function.
- **Issue:** #27500
- **Twitter handle:** coolbeevip

### DIFF
--- a/libs/core/tests/unit_tests/test_tools.py
+++ b/libs/core/tests/unit_tests/test_tools.py
@@ -483,6 +483,82 @@ def test_structured_tool_lambda_multi_args_schema() -> None:
     assert tool.args == expected_args
 
 
+def test_structured_tool_from_function_partial_docstring() -> None:
+    """Test that structured tools can be created from partial functions."""
+
+    def foo(bar: int, baz: str) -> str:
+        """Docstring
+
+        Args:
+            bar: the bar value
+            baz: the baz value
+        """
+        raise NotImplementedError
+
+    structured_tool = StructuredTool.from_function(func=partial(foo, baz="foo"))
+    assert structured_tool.name == "foo"
+    assert structured_tool.args == {
+        "bar": {"title": "Bar", "type": "integer"},
+        "baz": {"title": "Baz", "type": "string"},
+    }
+
+    assert _schema(structured_tool.args_schema) == {
+        "properties": {
+            "bar": {"title": "Bar", "type": "integer"},
+            "baz": {"title": "Baz", "type": "string"},
+        },
+        "description": inspect.getdoc(foo),
+        "title": "foo",
+        "type": "object",
+        "required": ["bar", "baz"],
+    }
+
+    assert foo.__doc__ is not None
+    assert structured_tool.description == textwrap.dedent(foo.__doc__.strip())
+
+
+def test_structured_tool_from_function_partial_docstring_complex_args() -> None:
+    """Test that structured tools can be created from partial functions."""
+
+    def foo(bar: int, baz: list[str]) -> str:
+        """Docstring
+
+        Args:
+            bar: int
+            baz: List[str]
+        """
+        raise NotImplementedError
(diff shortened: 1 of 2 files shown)


efriis (commented):
can you add to the test to confirm it can autopopulate

- name from function name
- description from docstring

and humor me because I haven't used partial as much - could we write an explicit test for decorated partial functions?

coolbeevip:
> can you add to the test to confirm it can autopopulate
> 
> * name from function name
> * description from docstring

Thank you for your suggestion. To maintain consistency, I will refer to the following test cases in order to add the corresponding partial test cases.

- test_structured_tool_from_function_partial_docstring
- test_structured_tool_from_function_partial_docstring_complex_args

> and humor me because I haven't used partial as much - could we write an explicit test for decorated partial functions?

I would like to clarify what you mean by "explicit test for decorated partial functions." Could you please provide more details or examples?

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -483,6 +483,102 @@ def test_structured_tool_lambda_multi_args_schema() -> None:
...
+            bar: the bar value
+            baz: the baz value
+        """
+        raise NotImplementedError
+
+    structured_tool = StructuredTool.from_function(func=partial(foo, baz="foo"))
+    assert structured_tool.name == "foo"
+    assert structured_tool.args == {
+        "bar": {"title": "Bar", "type": "integer"},
+        "baz": {"title": "Baz", "type": "string"},
```
baskaryan, line 502: don't we want baz to not appear as an  arg since its already been set?

coolbeevip: I'm looking to assign the same tool to several agents, ensuring that some of the parameters are preset, whereas others are dynamically determined by a large model each time an agent uses this tool.

Also, I found similar test scenarios in the test cases; unfortunately, right now, the `StructuredTool.from_function` method doesn't support them yet.

https://github.com/langchain-ai/langchain/blob/master/libs/core/tests/unit_tests/test_tools.py#L486

coolbeevip: > don't we want baz to not appear as an arg since its already been set?

Hi, just checking in on this PR. Any feedback? Thanks!

ccurme: @coolbeevip I think some more work is needed on this PR. IMO it doesn't make sense for the args schema of the tool to still retain the partialed-out parameter. In your first two tests, you can pass the test by just setting `func=foo`.

We document some other ways to do this in [this guide](https://python.langchain.com/docs/how_to/tool_runtime/) as well.

coolbeevip: Thank you for your guidance. The InjectedToolArg can indeed address my use case. 

However, in my situation, I already know the value of a parameter that needs to be passed when defining the function, rather than when calling the tool. 

Supporting partial functions would be more flexible, as this could eliminate the need for repeated function definitions.

Regardless, I appreciate your response. Please feel free to close this PR anytime.

coolbeevip: > @coolbeevip I think some more work is needed on this PR. IMO it doesn't make sense for the args schema of the tool to still retain the partialed-out parameter. In your first two tests, you can pass the test by just setting `func=foo`.
> 
> We document some other ways to do this in [this guide](https://python.langchain.com/docs/how_to/tool_runtime/) as well.

Thank you for your guidance. The InjectedToolArg can indeed address my use case.

However, in my situation, I already know the value of a parameter that needs to be passed when defining the function, rather than when calling the tool.

Supporting partial functions would be more flexible, as this could eliminate the need for repeated function definitions.

Regardless, I appreciate your response. Please feel free to close this PR anytime.

Issue #27505, opened 2024-10-21
HuggingFaceEmbeddings takes the default model name and reinitializes self.client even while passing in client parameter.


PraNavKumAr01:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_huggingface import HuggingFaceEmbeddings
from sentence_transformers import SentenceTransformer

model = SentenceTransformer(
   "all-MiniLM-L6-v2",
   backend="onnx",
   model_kwargs={
      "file_name": "model_qint8_avx512_vnni.onnx",
      "provider" : "CPUExecutionProvider"
  },
)

embeddings = HuggingFaceEmbeddings(client = model)
```

### Description

The HuggingFaceEmbedding class has a client parameter, which lets you pass in a pre-loaded model. But right now in the code, even after passing the client parameter, it has a DEFAULT_MODEL_NAME = "sentence-transformers/all-mpnet-base-v2" set and it starts downloading a loading that model
The reason for this is i think that 
```python
self.client = sentence_transformers.SentenceTransformer(
                self.model_name, cache_folder=self.cache_folder, **self.model_kwargs
            )
```
The self.client is initialized without checking if client has been passed into it or not, and regardless creates a new client loading the default model

This can be easily fixed by doing this
```python
if not self.client: # Add a check to see if self.client has been passed, if yes no need to reinitialize the client
            self.client = sentence_transformers.SentenceTransformer(
                self.model_name, cache_folder=self.cache_folder, **self.model_kwargs
            )
```
Ill be happy to open a pr, fix this small issue then merge. Let me know how to proceed with this.

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 23.5.0: Wed May  1 20:19:05 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T8112
> Python Version:  3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:41:52) [Clang 15.0.7 ]

Package Information
-------------------
> langchain_core: 0.3.12
> langchain: 0.2.0
> langchain_community: 0.0.38
> langsmith: 0.1.136
[...]

PR #27507, opened 2024-10-21
langchain_huggingface: Fix default initialization of self.client even after passing client as a parameter in HuggingFaceEmbeddings

PraNavKumAr01:
**Description:** Currently, the HuggingFaceEmbeddings class reinitializes self.client with the DEFAULT_MODEL_NAME even if client is passed as a parameter
**Issue:** #27505 
**Dependencies:** None
**Twitter handle:** PranavAI_ML


vbarda:
@PraNavKumAr01 thank you for your contribution! the `client` field is private and shouldn't be set directly by users -- here is a PR that fixes that https://github.com/langchain-ai/langchain/pull/27522

will close this one in the meantime

Issue #26777, opened 2024-09-23
I am getting pydantic Validation error when expecting tools in a response


tkmptech:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The Following code:
```python
from datetime import datetime
from typing import List

from langchain_core.messages import (
    SystemMessage,
    HumanMessage,
    ToolMessage,
    trim_messages,
)
from langchain_core.tools import tool
from langchain_openai import ChatOpenAI

# from openai import BaseModel
from pydantic.v1 import BaseModel, Field

## My messages in this format are accepted
 def build_message_list(self) -> List:
        # Convert internal history to a list of SystemMessage, HumanMessage, AIMessage, ToolMessage
        messages = [
            SystemMessage(
                content=""" \n Do Not make Vague assumptions use 
        appropriate tools provided whenever required to extract new or next Question."""
            )
        ]
        for entry in self.history:
            if entry["role"] == "human":
                messages.append(HumanMessage(content=entry["content"]))
            elif entry["role"] == "assistant":
                messages.append(AIMessage(content=entry["content"]))
            elif entry["role"] == "tool":
                messages.append(ToolMessage(content=entry["content"]))
        return messages
# this is the function where i should expect a AIMessage.
  def invoke_tool_or_model(self, messages) -> dict:
      """Handles whether to invoke a tool or continue with the LLM."""
      last_message = messages[-1]
      if hasattr(last_message, "tool_calls") and last_message.tool_calls:
          return self.call_tools(last_message.tool_calls)
      else:
          prompt = self.format_trimmed_history(messages)

          response = self.llm.invoke(prompt)
          return response
          
```

### Error Message and Stack Trace (if applicable)

the json response is giving the error inside the "tool_calls" where the accepted format is not decoded.

```
D:\Work\Company_Product_Brainstorming\Sample_projects\GPTInterviewer\Final\envinter\Lib\site-packages\pydantic\main.py:390: UserWarning: Pydantic serializer warnings:
  Expected `str` but got `dict` with value `{'category': 'math'}` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_python(
[...]

PR #27523, opened 2024-10-21
core: Handle arguments returned by Text Generation Inference's Messages API

Simon-Stone:
core: Handle tool arguments returned by Text Generation Inference's messages API

- **Description:** The parser for OpenAI tool calls expects the function arguments to be returned as a string. This is true for OpenAI's own API, but not necessarily for OpenAI-compatible APIs like Text Generation Inference's (TGI) Messages API. TGI returns the arguments as a dictionary. Since some OpenAI-compatible tools refer to OpenAI's client library (which handles TGI's tool response correctly), it makes sense to also support them through LangChain's OpenAI wrappers. This PR checks the type of the returned arguments and correctly processes them accordingly.
- **Issue:** #26777

### DIFF
--- a/libs/core/langchain_core/output_parsers/openai_tools.py
+++ b/libs/core/langchain_core/output_parsers/openai_tools.py
@@ -39,25 +39,35 @@ def parse_tool_call(
     """
     if "function" not in raw_tool_call:
         return None
-    if partial:
-        try:
-            function_args = parse_partial_json(
-                raw_tool_call["function"]["arguments"], strict=strict
-            )
-        except (JSONDecodeError, TypeError):  # None args raise TypeError
-            return None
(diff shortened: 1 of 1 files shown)


efriis:
hey! i believe this would be an issue with TGI not actually being openai-compatible

would recommend using the integration for TGI in langchain-huggingface instead!

Simon-Stone:
Thanks for taking a look at this!

Unfortunately, there is no integration for TGI's Messages API in `langchain-huggingface`, best I can tell. TGI also [officially refers](https://huggingface.co/docs/text-generation-inference/en/basic_tutorials/using_guidance#openai-integration) to OpenAI's client library. Other serving platforms may also benefit from this flexibility (the linked issue uses koboldcpp, for example).

Simon-Stone:
I have started discussion #27561 since this is touching multiple packages and may be more about finding the right tool for the job.

PR #27547, opened 2024-10-22
fix(core): support `handle_tool_error=(Exception, ...)` tuple

baskaryan:
Current handle_tool_error functionality doesn't support handling any exceptions outside of ToolException, meaning user has very little actual control over what gets handled (if the tool implementation doesn't raise ToolException then there's not much handling they can do). This pr proposes allowing users to specify the exception types they want handled, either directly via tuple[type[Exception]] or via the signature of Callable handler.

### DIFF
--- a/libs/core/langchain_core/tools/base.py
+++ b/libs/core/langchain_core/tools/base.py
@@ -19,8 +19,6 @@ from typing import (
     TypeVar,
     Union,
     cast,
-    get_args,
-    get_origin,
     get_type_hints,
 )
 
@@ -37,6 +35,7 @@ from pydantic import (
 from pydantic.v1 import BaseModel as BaseModelV1
 from pydantic.v1 import ValidationError as ValidationErrorV1
 from pydantic.v1 import validate_arguments as validate_arguments_v1
+from typing_extensions import get_args, get_origin
 
 from langchain_core._api import deprecated
 from langchain_core.callbacks import (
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/tools/base.py
```diff
@@ -399,9 +398,9 @@ class ChildTool(BaseTool):
     You can use these to eg identify a specific instance of a tool with its use case.
     """
 
-    handle_tool_error: Optional[Union[bool, str, Callable[[ToolException], str]]] = (
-        False
-    )
+    handle_tool_error: Union[
```
vbarda, line 401: i think it is confusing to put a tuple of exceptions to handle into `handle_tool_errors` -- currently it defines how we're populating the `ToolMessage.content`, so it would be unintuitive to me as a user that i would need to pass exceptions here (as opposed to a separate kwarg). besides, what if i want to customize the `ToolMessage.content` AND the exceptions that are handled?

libs/core/langchain_core/tools/base.py
```diff
@@ -1050,3 +1078,24 @@ class BaseToolkit(BaseModel, ABC):
     @abstractmethod
     def get_tools(self) -> list[BaseTool]:
         """Get the tools in the toolkit."""
+
+
+def _infer_handled_types(handler: Callable[..., str]) -> tuple[type[Exception]]:
```
vbarda, line 1083: this is nice!

libs/core/langchain_core/tools/base.py
baskaryan: then you use the Callable

vbarda:
what happens if a user wants to customize the error handling behavior for different exceptions?

baskaryan:
> what happens if a user wants to customize the error handling behavior for different exceptions?

pass in a function (callable)

vbarda:
> > what happens if a user wants to customize the error handling behavior for different exceptions?
> 
> pass in a function (callable)

ah, yes, that makes sense

libs/core/langchain_core/tools/base.py
```diff
@@ -842,19 +864,25 @@ def _handle_validation_error(
 
 
 def _handle_tool_error(
-    e: ToolException,
+    e: Exception,
     *,
-    flag: Optional[Union[Literal[True], str, Callable[[ToolException], str]]],
+    flag: Union[
+        None,
+        bool,
```
vbarda, line 871: (nit) shouldn't this still be a `Literal[True]`?

libs/core/langchain_core/tools/base.py
eyurtsev: I'm closing this PR. I think that we should've kept this interface minimal and closer to an `on_error` callback with a callable (disable all other functionality)

PR #27564, opened 2024-10-22
bagatur nits

baskaryan:


docs/docs/concepts/callbacks.mdx
```diff
@@ -4,13 +4,11 @@
 - [Runnable interface](/docs/concepts/#runnable-interface)
 :::
 
-## Overview
```
eyurtsev, line 7: IMO these really break navigation if they're missing w/ respect to the right side bar -- i found it very frustrating not being able to view the top of the page from the navigation bar

eyurtsev (approved):

PR #27585, opened 2024-10-23
langchain-openai: add check for none values when summing token usage

yahya-mouman:
**Description:** Fixes None addition issues when an empty value is passed on

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -604,9 +604,9 @@ class BaseChatOpenAI(BaseChatModel):
                 continue
             token_usage = output["token_usage"]
             if token_usage is not None:
-                if v is None:
-                    continue
                 for k, v in token_usage.items():
+                    if v is None:
+                        continue
                     if k in overall_token_usage:
                         overall_token_usage[k] = _update_token_usage(
                             overall_token_usage[k], v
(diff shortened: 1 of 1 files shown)


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -605,6 +605,8 @@ def _combine_llm_outputs(self, llm_outputs: List[Optional[dict]]) -> dict:
             token_usage = output["token_usage"]
             if token_usage is not None:
                 for k, v in token_usage.items():
+                    if v is None:
```
eyurtsev, line 608: what would happen if `v` were a string due to a bad API response? Instead of checking what it is, should we be checking what it's not? is `v` expected to be an `int` usually?

eyurtsev:
@baskaryan I think i've seen this pop up in other places w/ None values being returned by some model providers

libs/partners/openai/langchain_openai/chat_models/base.py
yahya-mouman: I believe `v` is already checked inside _update_token_usage(). `v` can be a list or a dict or an int. However we don't check for it being a None value which breaks the code when we try to add up None + None

baskaryan: do you know what it looks like when it's a list?

PR #27604, opened 2024-10-24
community: Add `@mozilla/readability` document transformer

CNSeniorious000:
### Description

`langchain-js` already has a useful document transformer that use `@mozilla/readability` to extract main content of a web page heuristically. [[docs]](https://js.langchain.com/docs/integrations/document_transformers/mozilla_readability/) [[source]](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/document_transformers/mozilla_readability.ts)

This PR introduces a new `ReadabilityTransformer` class to the `langchain_community/document_transformers`, which class leverages the `python-readability` library to do the same thing.

### Dependencies:

[`python-readability`](https://pypi.org/project/python-readability/) — a Standalone Python wrapper for `@mozilla/readability`

Mention that no nodejs environment is needed. In regular CPython distributions, [python-readability](https://github.com/CNSeniorious000/python-readability) requires [PythonMonkey](https://github.com/Distributive-Network/PythonMonkey) to interpret JavaScript, and in [Pyodide](https://github.com/pyodide/pyodide), it uses the native JavaScript environment. So this package is available even if the user deploys langchain apps on [Cloudflare Workers](https://developers.cloudflare.com/workers/languages/python/).

### DIFF
--- a/libs/community/extended_testing_deps.txt
+++ b/libs/community/extended_testing_deps.txt
@@ -70,6 +70,7 @@ pymupdf>=1.22.3,<2
 pypdf>=3.4.0,<5
 pypdfium2>=4.10.0,<5
 pyspark>=3.4.0,<4
+python-readability>=0.1.2
 rank-bm25>=0.2.2,<0.3
 rapidfuzz>=3.1.1,<4
 rapidocr-onnxruntime>=1.3.2,<2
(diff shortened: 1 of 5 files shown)


ccurme (commented):
Thanks for this @CNSeniorious000.

It's unclear to me that the demand for this is high enough to justify the additional maintenance burden.

Would you be interested in publishing an OSS integration package (e.g., `langchain-readability` or similar)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the package, and you have control over versioning.

Let me know what you think!

PR #25756, opened 2024-08-26
core: added an extra input param to RunnableWithMessageHistory

Abd-elr4hman:
# Description 
Added a new input argument to RunnableWithMessageHistory: extra_input_keys, the extra_input_keys will define prompt keys ignored by the get_input_schema method.
## The issue it solves
The current RunnableWithMessageHistory definition only makes use of an input key from the prompt, adds it to the input schema and ignores others. 
to illustrate:
* if I replicate the following example from [RunnableWithMessageHistory](https://python.langchain.com/v0.1/docs/expression_language/how_to/message_history/) docs :
```
from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
from langchain_openai.chat_models import ChatOpenAI
from langchain_community.chat_message_histories import ChatMessageHistory
from langchain_core.chat_history import BaseChatMessageHistory
from langchain_core.runnables.history import RunnableWithMessageHistory

model = ChatOpenAI()
prompt = ChatPromptTemplate.from_messages(
    [
        (
            "system",
            "You're an assistant who's good at {ability}. Respond in 20 words or fewer",
        ),
        MessagesPlaceholder(variable_name="history"),
        ("human", "{input}"),
    ]
)
runnable = prompt | model

store = {}

def get_session_history(session_id: str) -> BaseChatMessageHistory:
    if session_id not in store:
        store[session_id] = ChatMessageHistory()
    return store[session_id]


with_message_history = RunnableWithMessageHistory(
    runnable,
    get_session_history,
    input_messages_key="input",
    history_messages_key="history",
)
```
and then run
```
with_message_history.get_input_schema().schema_json()
```
I get the following output
```
{
  "title": "RunnableWithChatHistoryInput",
  "type": "object",
  "properties": {
    "input": {
      "title": "Input",
      "anyOf": [
        {
          "type": "string"
        },
        {
          "$ref": "#/definitions/BaseMessage"
        },
        {
          "type": "array",
          "items": {
            "$ref": "#/definitions/BaseMessage"
          }
        }
      ]
    }
  },...
```
**With only one property(input)** on the input schema, **despite the prompt containing others (ability)**

## Further issue with langserve
when I try to use RunnableWithMessageHistory with langserve in the example below
```
from pydantic.v1 import BaseModel
from fastapi import FastAPI
from fastapi.responses import RedirectResponse
from langserve import add_routes
from chain import with_message_history

app = FastAPI()

@app.get("/")
async def redirect_root_to_docs():
[...]

Issue #24370, opened 2024-07-17
KeyError with RunnableWithMessageHistory and LangServe: Missing Variable


norisuke3:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Description
When starting langserve with the code below and accessing it via `RemoteRunnable`, I encounter a `KeyError: "Input to ChatPromptTemplate is missing variables {'language'}. Expected: ['history', 'input', 'language'] Received: ['input', 'history']"`.

### Example Code
```python
from langchain_openai import ChatOpenAI
from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
from langchain_core.runnables.history import RunnableWithMessageHistory
from langchain_core.chat_history import BaseChatMessageHistory
from langchain_community.chat_message_histories import ChatMessageHistory
from langserve import add_routes
from fastapi import FastAPI
import uvicorn

app = FastAPI()

store = {}

def get_session_history(session_id: str) -> BaseChatMessageHistory:
    if session_id not in store:
        store[session_id] = ChatMessageHistory()
    return store[session_id]

prompt = ChatPromptTemplate.from_messages(
    [
        (
            "system",
            "You're an assistant who speaks in {language}. Respond in 20 words or fewer.",
        ),
        MessagesPlaceholder(variable_name="history"),
        ("human", "{input}"),
    ]
)

model = ChatOpenAI(model="gpt-3.5-turbo-0125")

runnable = prompt | model

runnable_with_history = RunnableWithMessageHistory(
    runnable,
    get_session_history,
    input_messages_key="input",
    history_messages_key="history",
)

chain = runnable_with_history

add_routes(app, chain, path="/test")

uvicorn.run(app, host="0.0.0.0", port=8000)
```
### Code for RemoteRunnable:
```python
from langserve import RemoteRunnable

rr = RemoteRunnable("http://localhost:8000/test/")
rr.invoke(
    {"language": "Italian", "input": "what's my name?"},
    config={"configurable": {"session_id": "1"}},
)
```
[...]

PR #27619, opened 2024-10-24
feat(core): modified `RunnableWithMessageHistory` `get_input_schema` to make use of the underlying runnable input keys

Abd-elr4hman:
# Description 
Modified RunnableWithMessageHistory get_input_schema to make use of the underlying runnable input keys.
This pull request a continuation of  https://github.com/langchain-ai/langchain/pull/25756 after langchain migration to pydantic v2.

## The issue it solves
The current RunnableWithMessageHistory definition only makes use of an input key from the prompt, adds it to the input schema and ignores others. 
to illustrate:
* if I replicate the following example from [RunnableWithMessageHistory](https://python.langchain.com/v0.1/docs/expression_language/how_to/message_history/) docs :
```
from langchain_community.chat_message_histories import ChatMessageHistory
from langchain_core.chat_history import BaseChatMessageHistory
from langchain_core.runnables.history import RunnableWithMessageHistory

model = ChatOpenAI()
    [
        (
            "system",
            "You're an assistant who's good at {ability}. Respond in 20 words or fewer",
        ),
        MessagesPlaceholder(variable_name="history"),
        ("human", "{input}"),
    ]
)
runnable = prompt | model

store = {}

def get_session_history(session_id: str) -> BaseChatMessageHistory:
    if session_id not in store:
        store[session_id] = ChatMessageHistory()
    return store[session_id]


with_message_history = RunnableWithMessageHistory(
    runnable,
    get_session_history,
    input_messages_key="input",
    history_messages_key="history",
)
```
and then run
```
with_message_history.get_input_schema().schema_json()
```
I get the following output
```
{
  "title": "RunnableWithChatHistoryInput",
  "type": "object",
  "properties": {
    "input": {
      "title": "Input",
      "anyOf": [
        {
          "type": "string"
        },
        {
          "$ref": "#/definitions/BaseMessage"
        },
        {
          "type": "array",
          "items": {
            "$ref": "#/definitions/BaseMessage"
          }
        }
      ]
    }
  },...
```
**With only one property(input)** on the input schema, **despite the prompt containing others (ability)**

## Further issue with langserve
when I try to use RunnableWithMessageHistory with langserve in the example below
```
from pydantic.v1 import BaseModel
from fastapi import FastAPI
from fastapi.responses import RedirectResponse
from langserve import add_routes
from chain import with_message_history

app = FastAPI()

@app.get("/")
async def redirect_root_to_docs():
    return RedirectResponse("/docs")

# Edit this to add the chain you want to add
add_routes(app, with_message_history)

if __name__ == "__main__":
    import uvicorn
    uvicorn.run(app, host="0.0.0.0", port=8000)
```
The fastapi docs shows the input schema without "ability" in the example above
Screenshot 2024-08-26 at 6 16 18 PM

  - **Issue:** #24370 
  - **Discussions:** 
  #22468 
  #22464

### DIFF
--- a/libs/core/langchain_core/runnables/history.py
+++ b/libs/core/langchain_core/runnables/history.py
@@ -12,6 +12,7 @@ from typing import (
 )
 
 from pydantic import BaseModel
+from pydantic_core import PydanticUndefined
 from typing_extensions import override
 
 from langchain_core.chat_history import BaseChatMessageHistory
@@ -235,6 +236,7 @@ class RunnableWithMessageHistory(RunnableBindingBase):
     output_messages_key: Optional[str] = None
     history_messages_key: Optional[str] = None
     history_factory_config: Sequence[ConfigurableFieldSpec]
+    _underlying_runnable: Runnable
 
     @classmethod
     def get_lc_namespace(cls) -> list[str]:
@@ -363,6 +365,7 @@ class RunnableWithMessageHistory(RunnableBindingBase):
             **kwargs,
         )
         self._history_chain = history_chain
+        self._underlying_runnable = runnable
 
     @property
     def config_specs(self) -> list[ConfigurableFieldSpec]:
@@ -390,6 +393,25 @@ class RunnableWithMessageHistory(RunnableBindingBase):
                 module_name=self.__class__.__module__,
                 root=(Sequence[BaseMessage], ...),
             )
+
+        underlying_input_schema = self._underlying_runnable.get_input_schema()
+        for (
+            field_name,
+            field_info,
+        ) in underlying_input_schema.model_fields.items():
+            if field_info.is_required():
+                if (
+                    field_name != self.input_messages_key
+                    and field_name != self.history_messages_key
+                ):
+                    fields[field_name] = (field_info.annotation, ...)
+            else:
+                if field_info.default is not PydanticUndefined:
+                    fields[field_name] = (field_info.annotation, field_info.default)
+                else:  # Default factory
+                    # We'll skip supporting this for now
+                    continue
+
         return create_model_v2(  # type: ignore[call-overload]
             "RunnableWithChatHistoryInput",
             field_definitions=fields,
--- a/libs/core/langchain_core/runnables/history.py
+++ b/libs/core/langchain_core/runnables/history.py
@@ -11,7 +11,7 @@ from typing import (
     Union,
 )
 
-from pydantic import BaseModel
+from pydantic import BaseModel, RootModel
 from pydantic_core import PydanticUndefined
 from typing_extensions import override
 
@@ -395,6 +395,13 @@ class RunnableWithMessageHistory(RunnableBindingBase):
             )
 
         underlying_input_schema = self._underlying_runnable.get_input_schema()
+        if issubclass(underlying_input_schema, RootModel):
+            return create_model_v2(  # type: ignore[call-overload]
+                "RunnableWithChatHistoryInput",
+                field_definitions=fields,
+                module_name=self.__class__.__module__,
+            )
+
         for (
             field_name,
             field_info,
--- a/libs/core/tests/unit_tests/runnables/test_history.py
+++ b/libs/core/tests/unit_tests/runnables/test_history.py
@@ -2,7 +2,6 @@ from collections.abc import Sequence
 from typing import Any, Callable, Optional, Union
 
 import pytest
-from pydantic import BaseModel
 
 from langchain_core.callbacks import (
     CallbackManagerForLLMRun,
@@ -12,7 +11,11 @@ from langchain_core.language_models.chat_models import BaseChatModel
 from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage
 from langchain_core.outputs import ChatGeneration, ChatResult
 from langchain_core.runnables import Runnable
-from langchain_core.runnables.base import RunnableBinding, RunnableLambda
+from langchain_core.runnables.base import (
+    RunnableBinding,
+    RunnableLambda,
+    RunnableParallel,
+)
 from langchain_core.runnables.config import RunnableConfig
 from langchain_core.runnables.history import RunnableWithMessageHistory
 from langchain_core.runnables.utils import ConfigurableFieldSpec, Input, Output
@@ -422,22 +425,17 @@ async def test_output_dict_async() -> None:
 
 
 def test_get_input_schema_input_dict() -> None:
-    from langchain_core.prompts import (
-        ChatPromptTemplate,
-        HumanMessagePromptTemplate,
-    )
+    from operator import itemgetter
 
-    runnable = ChatPromptTemplate.from_messages(
-        messages=[
-            SystemMessage(content="You are a nice assistant."),
-            HumanMessagePromptTemplate.from_template(
-                "ability: {ability}, input:{input}, history:{history}"
-            ),
-        ]
+    runnable = RunnableParallel(
+        {
+            "input": itemgetter("input"),
+            "ability": itemgetter("ability"),
+            "history": itemgetter("history"),
+        }
     )
 
     get_session_history = _get_get_session_history()
-
     with_history = RunnableWithMessageHistory(
         runnable,
         get_session_history,
@@ -456,7 +454,7 @@ def test_get_input_schema_input_dict() -> None:
                 ],
                 "title": "Input",
             },
-            "ability": {"title": "Ability", "type": "string"},
+            "ability": {"title": "Ability"},
         },
         "required": ["input", "ability"],
         "title": "RunnableWithChatHistoryInput",
--- a/libs/core/tests/unit_tests/runnables/test_history.py
+++ b/libs/core/tests/unit_tests/runnables/test_history.py
@@ -422,27 +422,22 @@ async def test_output_dict_async() -> None:
 
 
 def test_get_input_schema_input_dict() -> None:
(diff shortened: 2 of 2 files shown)


eyurtsev:
@Abd-elr4hman this looks good. Could you create a new test and keep the old one as well?

We want to make sure that test coverage is such that there are no regressions in the future and to verify that this change only fixes bugs and doesn't introduce new issues

Abd-elr4hman:
## The old test doesn't pass...

```python
runnable = RunnableLambda(
        lambda input: {
            "output": [
                AIMessage(
                    content="you said: "
                    + "\n".join(
                        [
                            str(m.content)
                            for m in input["history"]
                            if isinstance(m, HumanMessage)
                        ]
                        + [input["input"]]
                    )
                )
            ]
        }
    )
```

The reason is the input schema of the runnable used in the test is subclass of RootModel, and later in the test function input and history key is supplied:

```python
with_history = RunnableWithMessageHistory(
        runnable,
        get_session_history,
        input_messages_key="input",
        history_messages_key="history",
        output_messages_key="output",
    )
```

The class expects the input and history keys to be used only if the underlying runnable accepts a dict (subclass of BaseModel and not subclass of RootModel), from [docs](https://api.python.langchain.com/en/latest/runnables/langchain_core.runnables.history.RunnableWithMessageHistory.html) in the class definition:
```
input_messages_key – Must be specified if the base runnable accepts a dict as input. The key in the input dict that contains the messages.

history_messages_key – Must be specified if the base runnable accepts a dict as input and expects a separate key for historical messages.
```
The test itself supplied an underlying runnable with schema of RootModel and tested if the class treated it as if the underlying runnable accepts a dict.

The test still passed because it didn't need to account for the underlying runnable or check its schema.

in practice a runnable like:
```python
runnable = RunnableLambda(
        lambda input: {
            "output": [
                AIMessage(
                    content="you said: "
                    + "\n".join(
                        [
                            str(m.content)
                            for m in input["history"]
                            if isinstance(m, HumanMessage)
                        ]
                        + [input["input"]]
                    )
                )
            ]
        }
    )
```
should not be used with input and history keys.

## Solution?
it might be a good idea to test the underlying runnable schema type early on and alert the user if input and history keys are supplied where they shouldn't.

I hope my thoughts on this are clear, let me know if you think I missed anything

eyurtsev:
Hi @Abd-elr4hman, 

I don't think I understand the explanation.

The original runnable in the unit test looks like a Runnable that should be supported. 

It accepts a dict as an input (`input`) and two keys are expected to be present ("history" and "input"):

```python
    runnable = RunnableLambda(
        lambda input: {
            "output": [
                AIMessage(
                    content="you said: "
                            + "\n".join(
                        [
                            str(m.content)
                            for m in input["history"]
                            if isinstance(m, HumanMessage)
                        ]
                        + [input["input"]]
                    )
                )
            ]
        }
    )
```


As far as I can tell RunnableWithMessageHistory is correctly specified based on the runnable above

```python
    with_history = RunnableWithMessageHistory(
        runnable,
        get_session_history,
        input_messages_key="input",
        history_messages_key="history",
        output_messages_key="output",
    )
```

This looks like a valid test case to me, and we shouldn't be replacing it.

It's definitely OK to update the assertion in the test to avoid an indirect test through the pydantic model:

```
    assert _schema(with_history.get_input_schema()) ==  {} <-- Update the dictionary with the expected result
```

Abd-elr4hman:
You are correct the runnable accepts a dict.

I thought that if a runnable accepts a dict, its input schema had to indicate that. which is clearly not the case.

First: the runnable does in fact accept a dict.

second: its input schema doesn't have any information that I can use to deduce that it takes a dict input.
here is the runnable and its input json schema:
```python
    runnable = RunnableLambda(
        lambda input: {
            "output": [
                AIMessage(
                    content="you said: "
                            + "\n".join(
                        [
                            str(m.content)
                            for m in input["history"]
                            if isinstance(m, HumanMessage)
                        ]
                        + [input["input"]]
                    )
                )
            ]
        }
    )
```
```
{'title': 'RunnableLambdaInput'}
```

My mistake:
I thought an input runnable had to have information about its input fields in its schema. Example:
```python
from operator import itemgetter

    runnable = RunnableParallel(
        {
            "input": itemgetter("input"),
            "ability": itemgetter("ability"),
            "history": itemgetter("history"),
        }
    )
```
```
{
   "properties":{
      "input":{
         "title":"Input"
      },
      "ability":{
         "title":"Ability"
      },
      "history":{
         "title":"History"
      }
   },
   "required":[
      "input",
      "ability",
      "history"
   ],
   "title":"RunnableParallel<input,ability,history>Input",
   "type":"object"
}
```
The above example input schema has field "type":"object" which indicates that the schema describes a key value structure.. and its keys are listed in "properties"

Abd-elr4hman:
why the test case fails?
```python
class RunnableWithChatHistoryInput(BaseModel):
        input: Union[str, BaseMessage, Sequence[BaseMessage]]

    runnable = RunnableLambda(
        lambda input: {
            "output": [
                AIMessage(
                    content="you said: "
                    + "\n".join(
                        [
                            str(m.content)
                            for m in input["history"]
                            if isinstance(m, HumanMessage)
                        ]
                        + [input["input"]]
                    )
                )
            ]
        }
    )
    get_session_history = _get_get_session_history()
    with_history = RunnableWithMessageHistory(
        runnable,
        get_session_history,
        input_messages_key="input",
        history_messages_key="history",
        output_messages_key="output",
    )
```
because since 
```
input_messages_key="input"
history_messages_key="history"
```
are supplied, the class checks the runnable for its input schema and finds only root, since its input schema:
```
{'title': 'RunnableLambdaInput'}
```
are a subclass of RootModel. resulting in an RunnableWithHistory schema that looks as follows:
```
{
   "properties":{
      "input":{
         "anyOf":[
            {
               "type":"string"
            },
            {
               "$ref":"#/definitions/BaseMessage"
            },
            {
               "items":{
                  "$ref":"#/definitions/BaseMessage"
               },
               "type":"array"
            }
         ],
         "title":"Input"
      },
      "root":{
         "title":"Root"
      }
   },
   "required":[
      "input",
      "root"
   ],
   "title":"RunnableWithChatHistoryInput",
   "type":"object",
   "definitions":{
      "BaseMessage":{
         "additionalProperties":true,
         "description":"Base abstract message class.\n\nMessages are the inputs and outputs of ChatModels.",
         "properties":{
            "content":{
               "anyOf":[
                  {
                     "type":"string"
                  },
                  {
                     "items":{
                        "anyOf":[
                           {
                              "type":"string"
                           },
                           {
                              "type":"object"
                           }
                        ]
                     },
                     "type":"array"
                  }
               ],
               "title":"Content"
            },
            "additional_kwargs":{
               "title":"Additional Kwargs",
               "type":"object"
            },
            "response_metadata":{
               "title":"Response Metadata",
               "type":"object"
            },
            "type":{
               "title":"Type",
               "type":"string"
            },
            "name":{
               "anyOf":[
                  {
                     "type":"string"
                  },
                  {
                     "type":"null"
                  }
               ],
               "default":"None",
               "title":"Name"
            },
            "id":{
               "anyOf":[
                  {
                     "type":"string"
                  },
                  {
                     "type":"null"
                  }
               ],
               "default":"None",
               "title":"Id"
            }
         },
         "required":[
            "content",
            "type"
         ],
         "title":"BaseMessage",
         "type":"object"
      }
   }
}
```
with "type":"object" and "root" on properties.

Abd-elr4hman:
I believe this can be avoided with a check like:
```python
def get_input_schema(
    ) -> type[BaseModel]:

        fields: dict = {}
        if self.input_messages_key and self.history_messages_key:
            fields[self.input_messages_key] = (
                Union[str, BaseMessage, Sequence[BaseMessage]],
                ...,
            )
        elif self.input_messages_key:
            fields[self.input_messages_key] = (Sequence[BaseMessage], ...)
        else:
            return create_model_v2(
                "RunnableWithChatHistoryInput",
                module_name=self.__class__.__module__,
                root=(Sequence[BaseMessage], ...),
            )

        underlying_input_schema = self._underlying_runnable.get_input_schema()
        for (
            field_name,
            field_info,
        ) in underlying_input_schema.model_fields.items():
            if field_info.is_required():
                # check here ------------------------------------------------>
                if (
                    field_name != self.input_messages_key
                    and field_name != self.history_messages_key
                    and field_name != "root"
                ):
                    fields[field_name] = (field_info.annotation, ...)
            else:
                if field_info.default is not PydanticUndefined:
                    fields[field_name] = (field_info.annotation, field_info.default)
                else:  # Default factory
                    # We'll skip supporting this for now
                    continue

        return create_model_v2(  # type: ignore[call-overload]
            "RunnableWithChatHistoryInput",
            field_definitions=fields,
            module_name=self.__class__.__module__,
        )
```
But it might be better to make sure that runnable passed to RunnableWithMessageHistory specify their inputs on the input schema.

looking forward to hearing your opinion.

Abd-elr4hman:
hi @eyurtsev 
Could you please review the latest changes? Thanks.

eyurtsev:
Hi @Abd-elr4hman, apologies this was never merged. RunnableWithMessageHistory isn't really the recommended way to add persistence anymore. Almost all persistence is not working based on LangGraph checkpointing which is significantly better (both more optimized, simpler and can support a lot of things except for message history). 

I'm closing this PR as i think that the risk that this code breaks user code outweighs benefits from the feature. Given that:
* This is no longer the best way to add memory for langchain applications
* users have a work-around by explicitly specifying the inputs / outputs if they really want to use RunnableWithMessageHistory if they're still using it.

PR #27495, opened 2024-10-20
core: fix Image prompt template hardcoded template format

chunkanglu:
Fixes #27411 

**Description:** Adds `template_format` to the `ImagePromptTemplate` class and updates passing in the `template_format` parameter from ChatPromptTemplate instead of the hardcoded "f-string".
Also updated docs and typing related to `template_format` to be more up-to-date and specific.

**Dependencies:** None

**Add tests and docs**: Added unit tests to validate fix. Needed to update `test_chat` snapshot due to adding new attribute `template_format` in `ImagePromptTemplate`.

PR #27629, opened 2024-10-25
core: remove mustache in extended deps

eyurtsev:
Remove mustache from extended deps -- we vendor the mustache implementation


eyurtsev:
https://github.com/langchain-ai/langchain/pull/27495/files

PR #27678, opened 2024-10-28
langchain: Replace lxml and XSLT with BeautifulSoup in HTMLHeaderTextSplitter for Improved Large HTML File Processing

### LINKED ISSUE #13149: HTMLHeaderTextSplitter won't run (maxHead)
### System Info

Notebook with latest langchain

### Who can help?

_No response_

### Information

- [ ] My own modified scripts

### Related Components

- [ ] LLMs/Chat Models
- [ ] Embedding Models
- [ ] Output Parsers
- [X] Document Loaders
- [ ] Memory
- [ ] Tools / Toolkits
- [ ] Chains
- [ ] Callbacks/Tracing
- [ ] Async

### Reproduction

Trying in notebook the HTMLHeaderTextSplitter
```jupyter-notebook
from langchain.text_splitter import RecursiveCharacterTextSplitter

headers_to_split_on = [
    ("h1", "Header 1"),
    ("h2", "Header 2"),
    ("h3", "Header 3"),
    ("h4", "Header 4"),
]

html_splitter = HTMLHeaderTextSplitter(headers_to_split_on=headers_to_split_on)

html_header_splits = html_splitter.split_text_from_file("/content/X.html")

chunk_size = 500
chunk_overlap = 30
text_splitter = RecursiveCharacterTextSplitter(
    chunk_size=chunk_size, chunk_overlap=chunk_overlap
)

Split
splits = text_splitter.split_documents(html_header_splits)
splits[80:85]`

``` 
```jupyter-notebook


---------------------------------------------------------------------------
XSLTApplyError                            Traceback (most recent call last)
[<ipython-input-54-bd3edea942d1>](https://localhost:8080/#) in <cell line: 12>()
     10 html_splitter = HTMLHeaderTextSplitter(headers_to_split_on=headers_to_split_on)
     11 
---> 12 html_header_splits = html_splitter.split_text_from_file("/content/X.html")
     13 
     14 chunk_size = 500

[/usr/local/lib/python3.10/dist-packages/langchain/text_splitter.py](https://localhost:8080/#) in split_text_from_file(self, file)
    586         xslt_tree = etree.parse(xslt_path)
    587         transform = etree.XSLT(xslt_tree)
--> 588         result = transform(tree)
    589         result_dom = etree.fromstring(str(result))
    590 

src/lxml/xslt.pxi in lxml.etree.XSLT.__call__()

XSLTApplyError: maxHead``` 
Is the HTML just to large to be handled by the textsplitter?

### Expected behavior

Load the html..

### PR DESCRIPTION
**Pull request by AhmedTammaa**
This pull request updates the `HTMLHeaderTextSplitter` by replacing the `split_text_from_file` method's implementation. The original method used `lxml` and XSLT for processing HTML files, which caused `lxml.etree.xsltapplyerror maxhead` when handling large HTML documents due to limitations in the XSLT processor. Fixes #13149 

By switching to BeautifulSoup (`bs4`), we achieve:

- **Improved Performance and Reliability:** BeautifulSoup efficiently processes large HTML files without the errors associated with `lxml` and XSLT.
- **Simplified Dependencies:** Removes the dependency on `lxml` and external XSLT files, relying instead on the widely used `beautifulsoup4` library.
- **Maintained Functionality:** The new method replicates the original behavior, ensuring compatibility with existing code and preserving the extraction of content and metadata.

**Issue:**

This change addresses issues related to processing large HTML files with the existing `HTMLHeaderTextSplitter` implementation. It resolves problems where users encounter lxml.etree.xsltapplyerror maxhead due to large HTML documents.

**Dependencies:**

- **BeautifulSoup (`beautifulsoup4`):** The `beautifulsoup4` library is now used for parsing HTML content.
  - Installation: `pip install beautifulsoup4`

**Code Changes:**

Updated the `split_text_from_file` method in `HTMLHeaderTextSplitter` as follows:

```python
def split_text_from_file(self, file: Any) -> List[Document]:
    """Split HTML file using BeautifulSoup.

    Args:
        file: HTML file path or file-like object.

    Returns:
        List of Document objects with page_content and metadata.
    """
    from bs4 import BeautifulSoup
    from langchain.docstore.document import Document
    import bs4

    # Read the HTML content from the file or file-like object
    if isinstance(file, str):
        with open(file, 'r', encoding='utf-8') as f:
            html_content = f.read()
    else:
        # Assuming file is a file-like object
        html_content = file.read()

    # Parse the HTML content using BeautifulSoup
    soup = BeautifulSoup(html_content, 'html.parser')

    # Extract the header tags and their corresponding metadata keys
    headers_to_split_on = [tag[0] for tag in self.headers_to_split_on]
    header_mapping = dict(self.headers_to_split_on)

    documents = []

    # Find the body of the document
    body = soup.body if soup.body else soup

    # Find all header tags in the order they appear
    all_headers = body.find_all(headers_to_split_on)

    # If there's content before the first header, collect it
    first_header = all_headers[0] if all_headers else None
    if first_header:
        pre_header_content = ''
        for elem in first_header.find_all_previous():
            if isinstance(elem, bs4.Tag):
                text = elem.get_text(separator=' ', strip=True)
                if text:
                    pre_header_content = text + ' ' + pre_header_content
        if pre_header_content.strip():
            documents.append(Document(
                page_content=pre_header_content.strip(),
                metadata={}  # No metadata since there's no header
            ))
    else:
        # If no headers are found, return the whole content
        full_text = body.get_text(separator=' ', strip=True)
        if full_text.strip():
            documents.append(Document(
                page_content=full_text.strip(),
                metadata={}
            ))
        return documents

    # Process each header and its associated content
    for header in all_headers:
        current_metadata = {}
        header_name = header.name
        header_text = header.get_text(separator=' ', strip=True)
        current_metadata[header_mapping[header_name]] = header_text

        # Collect all sibling elements until the next header of the same or higher level
        content_elements = []
        for sibling in header.find_next_siblings():
            if sibling.name in headers_to_split_on:
                # Stop at the next header
                break
            if isinstance(sibling, bs4.Tag):
                content_elements.append(sibling)

        # Get the text content of the collected elements
        current_content = ''
        for elem in content_elements:
            text = elem.get_text(separator=' ', strip=True)
            if text:
                current_content += text + ' '

        # Create a Document if there is content
        if current_content.strip():
            documents.append(Document(
                page_content=current_content.strip(),
                metadata=current_metadata.copy()
            ))
        else:
            # If there's no content, but we have metadata, still create a Document
            documents.append(Document(
                page_content='',
                metadata=current_metadata.copy()
            ))

    return documents
```


AhmedTammaa:
Hi @eyurtsev 

Could you take a look at this, please?

eyurtsev:
@AhmedTammaa this is a great change -- it'll make it much easier to understand what's going on.

Given that you're already taken a deep dive into this code, would you be able to help define the precise semantics for what the splitter does?

Based on the code:

```python
class HTMLHeaderTextSplitter:
    """
    Splitting HTML files based on specified headers.
    Requires lxml package.
    """
    def __init__(
        self,
        headers_to_split_on: List[Tuple[str, str]],
        return_each_element: bool = False,
    ):
        """Create a new HTMLHeaderTextSplitter.
        Args:
            headers_to_split_on: list of tuples of headers we want to track mapped to
                (arbitrary) keys for metadata. Allowed header values: h1, h2, h3, h4,
                h5, h6 e.g. [("h1", "Header 1"), ("h2", "Header 2)].
            return_each_element: Return each element w/ associated headers.
        """
```

It's a bit hard to understand from the doc-string, but what's the actual of each document chunk?
* What does the content for a given header contain. (Does the `h1` header contain the `h2` header is we're splitting on h1 and h2)
* Can we split only on h2, but not h1? 
* what metadata does each document contain?

AhmedTammaa:
Hi @eyurtsev 

Thanks for the feedback! Here’s what’s going on:

- The splitter takes an HTML page and creates `Document` objects, each representing a chunk of text under a certain “header context.”
- Before the first header appears, any text is grouped into a “pre-header” chunk with no metadata.
- Once we hit a header (like `<h1>`, `<h2>`, etc.), we finalize the previous chunk and start a new one. The headers you choose in `headers_to_split_on` determine what levels we actually split on and what gets stored in the metadata. For example:
  - `[("h1", "Header 1"), ("h2", "Header 2")]` means `<h1>` text goes into `"Header 1"` in metadata and `<h2>` text goes into `"Header 2"`.
- If you only pick `h2` and `h3` to split on, the code ignores `h1` for splitting and metadata. It just won’t create a new chunk at `h1`.
- Each `Document`’s `page_content` is basically the combined text of `<p>` elements (and whatever tags you specify) under the current headers. Its `metadata` is a dict of header levels (e.g. `{"Header 1": "Some Title", "Header 2": "Some Subsection"}`).


## I have also modified the provided code and made further tests:
I prepared this small notebook on Google Colab to show you use cases old vs new [code](https://colab.research.google.com/drive/15BaYKXJ6_HpuYrrniH-eUrhSAYr5jMCz?usp=sharing)

It is public and anyone with the link can see it. 

### The last experiment shows the main reason for change where the old class fails with larger files
I tested it on a generated HTML file

If you have any more questions please let me know. 
Thanks for your effort in Langchain it makes my life easier and that's why I am eager to contribute to it :)

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -6,7 +6,9 @@
 from typing import Any, Dict, Iterable, List, Optional, Tuple, TypedDict, cast
 
 import requests
-from langchain_core.documents import Document
+from bs4 import BeautifulSoup
```
eyurtsev, line 9: Move these imports into local scope. Most of langchain implementations have to avoid assuming that a particular 3rd party package is installed on a user's machine

```python
from bs4 import BeautifulSoup
from bs4.element import Tag
```
This import is incorrect, use the `langchain_core.documents import Document` import that you had before

```python
from langchain.docstore.document import Document
```

eyurtsev:
@AhmedTammaa OK this makes sense and will be very happy to merge the code. if we can do the following:

1. Add a unit test for at least one, ideally two more test cases:

A) split on h1 only (but document contains h2 / h3 tags)
B) split on h1 only (but document does not contain any header tags)

The existing unit tests are here: https://github.com/langchain-ai/langchain/blob/master/libs/text-splitters/tests/unit_tests/test_text_splitters.py#L1638-L1638

You can use `pytest.parameterize()` and I'd suggest just asking chatgpt (or your favorite chat model) to write it. It'll help make sure that the semantics are well defined.

2) Address the issue w/ the imports (flagged via a comment)

3) If you're able improve the doc-string for the API reference (here too you can probably ask your favorite chat model to update the API reference using google style doc-string and the content of your explanation + test cases)

AhmedTammaa:
Alright @eyurtsev 

I will be working on that. I will do a lot of testing before pushing. I am currently studying the old behaviour deeply, so I can reproduce it (ideally, exactly the same) or with minor changes. 

The way I will be testing is asking an llm to generate some sample documents and I compare old output to new output which should be the same (at the moment it isn't). 

Or,

There can be more flexibility?

eyurtsev:
> I compare old output to new output which should be the same (at the moment it isn't).

Likely OK if the output isn't the same. Based on how the splitter is supposed to work, it sounds like we can well define what the behavior should be for each of the test cases.

So let's create some simple test cases, and verify that the output we get is as expected.

AhmedTammaa:
Hi @eyurtsev,

Thanks for answering my query.

Currently, I have the class stable and it passes all of tests. I am doing code-cleaning and performance optimization.

The current class also, as the former, deals with the document as a "tree" and no recursion so we can avoid future problems like the `recursion limit exceeded` error.


Here is a summary of the behaviour and the added test cases.

### **Summary of `HTMLHeaderTextSplitter` Behavior**

1. **Document Chunk Content:**
   - **Headers:** Each specified header (e.g., `<h1>`, `<h2>`) becomes a separate `Document` containing the header text.
   - **Content:** Text following a header up to the next header of the same or higher level is grouped into a `Document`. When splitting on multiple headers (e.g., `<h1>` and `<h2>`), higher-level headers do not contain lower-level ones; instead, metadata reflects the hierarchy.

2. **Splitting Only on Specific Headers:**
   - You can choose to split on any combination of headers. For example, splitting only on `<h2>` while ignoring `<h1>` is supported by specifying only `("h2", "Header 2")` in `headers_to_split_on`.

3. **Metadata in Each Document:**
   - **Structure:** Each `Document` includes metadata mapping the specified headers to their respective texts.
   - **Hierarchy:** For nested headers, metadata accumulates to reflect the document structure. For example, a `<h2>` under a `<h1>` will have both `Header 1` and `Header 2` in its metadata.

### **Test Cases Overview**

- **Test Case A:**  
  *Splits on `<h1>`, `<h2>`, and `<h3>` within a nested structure, ensuring each header and its content are correctly segmented with appropriate metadata.*

- **Test Case B:**  
  *Splits on `<h1>` only in a document that contains no headers, resulting in a single aggregated `Document` without metadata.*

- **Added Test Cases**
I have added edge cases where elements are on different depths, nested structure and etc.
Here is the full test set now:
```python
from typing import List, Tuple

import pytest
from langchain.docstore.document import Document
from loguru import logger

from new_code import HTMLHeaderTextSplitter
from old_code import OldHTMLHeaderTextSplitter  # Imported for potential future use
from typing import Any

@pytest.fixture
def splitter_factory() -> HTMLHeaderTextSplitter:
    """
    Fixture to create an HTMLHeaderTextSplitter instance with given headers.
    This factory allows dynamic creation of splitters with different headers.
    """
    def _create_splitter(headers_to_split_on: List[Tuple[str, str]]) -> HTMLHeaderTextSplitter:
        return HTMLHeaderTextSplitter(headers_to_split_on=headers_to_split_on)
    return _create_splitter


@pytest.mark.parametrize(
    "headers_to_split_on, html_input, expected_documents, test_case",
    [
        (
            # Test Case 1: Split on h1 and h2
            [("h1", "Header 1"), ("h2", "Header 2")],
            """
            <html>
                <body>
                    <h1>Introduction</h1>
                    <p>This is the introduction.</p>
                    <h2>Background</h2>
                    <p>Background information.</p>
                    <h1>Conclusion</h1>
                    <p>Final thoughts.</p>
                </body>
            </html>
            """,
            [
                Document(
                    page_content="Introduction",
                    metadata={"Header 1": "Introduction"}
                ),
                Document(
                    page_content="This is the introduction.",
                    metadata={"Header 1": "Introduction"}
                ),
                Document(
                    page_content="Background",
                    metadata={
                        "Header 1": "Introduction",
                        "Header 2": "Background"
                    }
                ),
                Document(
                    page_content="Background information.",
                    metadata={
                        "Header 1": "Introduction",
                        "Header 2": "Background"
                    }
                ),
                Document(
                    page_content="Conclusion",
                    metadata={"Header 1": "Conclusion"}
                ),
                Document(
                    page_content="Final thoughts.",
                    metadata={"Header 1": "Conclusion"}
                )
            ],
            "Simple headers and paragraphs"
        ),
        (
            # Test Case 2: Nested headers with h1, h2, and h3
            [("h1", "Header 1"), ("h2", "Header 2"), ("h3", "Header 3")],
            """
            <html>
                <body>
                    <div>
                        <h1>Main Title</h1>
                        <div>
                            <h2>Subsection</h2>
                            <p>Details of subsection.</p>
                            <div>
                                <h3>Sub-subsection</h3>
                                <p>More details.</p>
                            </div>
                        </div>
                    </div>
                    <h1>Another Main Title</h1>
                    <p>Content under another main title.</p>
                </body>
            </html>
            """,
            [
                Document(
                    page_content="Main Title",
                    metadata={"Header 1": "Main Title"}
                ),
                Document(
                    page_content="Subsection",
                    metadata={
                        "Header 1": "Main Title",
                        "Header 2": "Subsection"
                    }
                ),
                Document(
                    page_content="Details of subsection.",
                    metadata={
                        "Header 1": "Main Title",
                        "Header 2": "Subsection"
                    }
                ),
                Document(
                    page_content="Sub-subsection",
                    metadata={
                        "Header 1": "Main Title",
                        "Header 2": "Subsection",
                        "Header 3": "Sub-subsection"
                    }
                ),
                Document(
                    page_content="More details.",
                    metadata={
                        "Header 1": "Main Title",
                        "Header 2": "Subsection",
                        "Header 3": "Sub-subsection"
                    }
                ),
                Document(
                    page_content="Another Main Title",
                    metadata={"Header 1": "Another Main Title"}
                ),
                Document(
                    page_content="Content under another main title.",
                    metadata={"Header 1": "Another Main Title"}
                )
            ],
            "Nested headers with h1, h2, and h3"
        ),
        (
            # Test Case 3: No headers
            [("h1", "Header 1")],
            """
            <html>
                <body>
                    <p>Paragraph one.</p>
                    <p>Paragraph two.</p>
                    <div>
                        <p>Paragraph three.</p>
                    </div>
                </body>
            </html>
            """,
            [
                Document(
                    page_content="Paragraph one.  \nParagraph two.  \nParagraph three.",
                    metadata={}
                )
            ],
            "No headers present"
        ),
        (
            # Test Case 4: Multiple headers of the same level
            [("h1", "Header 1")],
            """
            <html>
                <body>
                    <h1>Chapter 1</h1>
                    <p>Content of chapter 1.</p>
                    <h1>Chapter 2</h1>
                    <p>Content of chapter 2.</p>
                    <h1>Chapter 3</h1>
                    <p>Content of chapter 3.</p>
                </body>
            </html>
            """,
            [
                Document(
                    page_content="Chapter 1",
                    metadata={"Header 1": "Chapter 1"}
                ),
                Document(
                    page_content="Content of chapter 1.",
                    metadata={"Header 1": "Chapter 1"}
                ),
                Document(
                    page_content="Chapter 2",
                    metadata={"Header 1": "Chapter 2"}
                ),
                Document(
                    page_content="Content of chapter 2.",
                    metadata={"Header 1": "Chapter 2"}
                ),
                Document(
                    page_content="Chapter 3",
                    metadata={"Header 1": "Chapter 3"}
                ),
                Document(
                    page_content="Content of chapter 3.",
                    metadata={"Header 1": "Chapter 3"}
                )
            ],
            "Multiple headers of the same level"
        ),
        (
            # Test Case 5: Headers with no content
            [("h1", "Header 1"), ("h2", "Header 2")],
            """
            <html>
                <body>
                    <h1>Header 1</h1>
                    <h2>Header 2</h2>
                    <h1>Header 3</h1>
                </body>
            </html>
            """,
            [
                Document(
                    page_content="Header 1",
                    metadata={"Header 1": "Header 1"}
                ),
                Document(
                    page_content="Header 2",
                    metadata={
                        "Header 1": "Header 1",
                        "Header 2": "Header 2"
                    }
                ),
                Document(
                    page_content="Header 3",
                    metadata={"Header 1": "Header 3"}
                )
            ],
            "Headers with no associated content"
        ),
    ]
)
def test_html_header_text_splitter(
    splitter_factory: Any,
    headers_to_split_on: List[Tuple[str, str]],
    html_input: str,
    expected_documents: List[Document],
    test_case: str
):
    """
    Parametrized test for HTMLHeaderTextSplitter.

    Args:
        splitter_factory (Callable): Fixture to create splitter instances.
        headers_to_split_on (List[Tuple[str, str]]): Headers to split on.
        html_input (str): The HTML content to split.
        expected_documents (List[Document]): The expected list of Documents after splitting.
        test_case (str): Description of the test case.
    """
    logger.info(f"Running Test Case: {test_case}")
    splitter = splitter_factory(headers_to_split_on=headers_to_split_on)
    docs = splitter.split_text(html_input)

    logger.info("Generated Documents:")
    for idx, doc in enumerate(docs, start=1):
        logger.info(f"Document {idx}: {doc}")

    assert len(docs) == len(expected_documents), (
        f"Test Case '{test_case}' Failed: Number of documents mismatch. "
        f"Expected {len(expected_documents)}, got {len(docs)}."
    )
    for idx, (doc, expected) in enumerate(zip(docs, expected_documents), start=1):
        assert doc.page_content == expected.page_content, (
            f"Test Case '{test_case}' Failed at Document {idx}: "
            f"Content mismatch.\nExpected: {expected.page_content}\nGot: {doc.page_content}"
        )
        assert doc.metadata == expected.metadata, (
            f"Test Case '{test_case}' Failed at Document {idx}: "
            f"Metadata mismatch.\nExpected: {expected.metadata}\nGot: {doc.metadata}"
        )
    logger.info(f"Test Case '{test_case}' Passed.")


@pytest.mark.parametrize(
    "headers_to_split_on, html_content, expected_output, test_case",
    [
        (
            # Test Case A: Split on h1 and h2 with h3 in content
            [("h1", "Header 1"), ("h2", "Header 2"), ("h3", "Header 3")],
            """
            <!DOCTYPE html>
            <html>
            <body>
                <div>
                    <h1>Foo</h1>
                    <p>Some intro text about Foo.</p>
                    <div>
                        <h2>Bar main section</h2>
                        <p>Some intro text about Bar.</p>
                        <h3>Bar subsection 1</h3>
                        <p>Some text about the first subtopic of Bar.</p>
                        <h3>Bar subsection 2</h3>
                        <p>Some text about the second subtopic of Bar.</p>
                    </div>
                    <div>
                        <h2>Baz</h2>
                        <p>Some text about Baz</p>
                    </div>
                    <br>
                    <p>Some concluding text about Foo</p>
                </div>
            </body>
            </html>
            """,
            [
                Document(
                    metadata={'Header 1': 'Foo'},
                    page_content='Foo'
                ),
                Document(
                    metadata={'Header 1': 'Foo'},
                    page_content='Some intro text about Foo.'
                ),
                Document(
                    metadata={'Header 1': 'Foo', 'Header 2': 'Bar main section'},
                    page_content='Bar main section'
                ),
                Document(
                    metadata={'Header 1': 'Foo', 'Header 2': 'Bar main section'},
                    page_content='Some intro text about Bar.'
                ),
                Document(
                    metadata={
                        'Header 1': 'Foo',
                        'Header 2': 'Bar main section',
                        'Header 3': 'Bar subsection 1'
                    },
                    page_content='Bar subsection 1'
                ),
                Document(
                    metadata={
                        'Header 1': 'Foo',
                        'Header 2': 'Bar main section',
                        'Header 3': 'Bar subsection 1'
                    },
                    page_content='Some text about the first subtopic of Bar.'
                ),
                Document(
                    metadata={
                        'Header 1': 'Foo',
                        'Header 2': 'Bar main section',
                        'Header 3': 'Bar subsection 2'
                    },
                    page_content='Bar subsection 2'
                ),
                Document(
                    metadata={
                        'Header 1': 'Foo',
                        'Header 2': 'Bar main section',
                        'Header 3': 'Bar subsection 2'
                    },
                    page_content='Some text about the second subtopic of Bar.'
                ),
                Document(
                    metadata={'Header 1': 'Foo', 'Header 2': 'Baz'},
                    page_content='Baz'
                ),
                Document(
                    metadata={'Header 1': 'Foo'},
                    page_content='Some text about Baz  \nSome concluding text about Foo'
                )
            ],
            "Test Case A: Split on h1, h2, and h3 with nested headers"
        ),
        (
            # Test Case B: Split on h1 only without any headers
            [("h1", "Header 1")],
            """
            <html>
                <body>
                    <p>Paragraph one.</p>
                    <p>Paragraph two.</p>
                    <p>Paragraph three.</p>
                </body>
            </html>
            """,
            [
                Document(
                    metadata={},
                    page_content='Paragraph one.  \nParagraph two.  \nParagraph three.'
                )
            ],
            "Test Case B: Split on h1 only without any headers"
        )
    ]
)
def test_additional_html_header_text_splitter(
    splitter_factory: Any,
    headers_to_split_on: List[Tuple[str, str]],
    html_content: str,
    expected_output: List[Document],
    test_case: str
):
    """
    Additional parametrized tests for HTMLHeaderTextSplitter.

    Args:
        splitter_factory (Callable): Fixture to create splitter instances.
        headers_to_split_on (List[Tuple[str, str]]): Headers to split on.
        html_content (str): The HTML content to split.
        expected_output (List[Document]): The expected list of Documents after splitting.
        test_case (str): Description of the test case.
    """
    logger.info(f"Running {test_case}")
    splitter = splitter_factory(headers_to_split_on=headers_to_split_on)
    docs = splitter.split_text(html_content)

    logger.info("Generated Documents:")
    for idx, doc in enumerate(docs, start=1):
        logger.info(f"Document {idx}: {doc}")

    assert len(docs) == len(expected_output), (
        f"{test_case} Failed: Number of documents mismatch. "
        f"Expected {len(expected_output)}, got {len(docs)}."
    )
    for idx, (doc, expected) in enumerate(zip(docs, expected_output), start=1):
        assert doc.page_content == expected.page_content, (
            f"{test_case} Failed at Document {idx}: "
            f"Content mismatch.\nExpected: {expected.page_content}\nGot: {doc.page_content}"
        )
        assert doc.metadata == expected.metadata, (
            f"{test_case} Failed at Document {idx}: "
            f"Metadata mismatch.\nExpected: {expected.metadata}\nGot: {doc.metadata}"
        )
    logger.info(f"{test_case} Passed.")


@pytest.mark.parametrize(
    "headers_to_split_on, html_content, expected_output, test_case",
    [
        (
            # Test Case C: Split on h1, h2, and h3 with no headers present
            [("h1", "Header 1"), ("h2", "Header 2"), ("h3", "Header 3")],
            """
            <html>
                <body>
                    <p>Just some random text without headers.</p>
                    <div>
                        <span>More text here.</span>
                    </div>
                </body>
            </html>
            """,
            [
                Document(
                    page_content='Just some random text without headers.  \nMore text here.',
                    metadata={}
                )
            ],
            "Test Case C: Split on h1, h2, and h3 without any headers"
        )
    ]
)
def test_no_headers_with_multiple_splitters(
    splitter_factory: Any,
    headers_to_split_on: List[Tuple[str, str]],
    html_content: str,
    expected_output: List[Document],
    test_case: str
):
    """
    Additional test case where headers to split on include h3, but no headers are present in the HTML.

    Args:
        splitter_factory (Callable): Fixture to create splitter instances.
        headers_to_split_on (List[Tuple[str, str]]): Headers to split on.
        html_content (str): The HTML content to split.
        expected_output (List[Document]): The expected list of Documents after splitting.
        test_case (str): Description of the test case.
    """
    logger.info(f"Running {test_case}")
    splitter = splitter_factory(headers_to_split_on=headers_to_split_on)
    docs = splitter.split_text(html_content)

    logger.info("Generated Documents:")
    for idx, doc in enumerate(docs, start=1):
        logger.info(f"Document {idx}: {doc}")

    assert len(docs) == len(expected_output), (
        f"{test_case} Failed: Number of documents mismatch. "
        f"Expected {len(expected_output)}, got {len(docs)}."
    )
    for idx, (doc, expected) in enumerate(zip(docs, expected_output), start=1):
        assert doc.page_content == expected.page_content, (
            f"{test_case} Failed at Document {idx}: "
            f"Content mismatch.\nExpected: {expected.page_content}\nGot: {doc.page_content}"
        )
        assert doc.metadata == expected.metadata, (
            f"{test_case} Failed at Document {idx}: "
            f"Metadata mismatch.\nExpected: {expected.metadata}\nGot: {doc.metadata}"
        )
    logger.info(f"{test_case} Passed.")
```

Currently, I have the class is stable and passes all of them. I am doing code-cleaning and doing performance optimization. 

The current class also deals with the document as a "tree" and no recursion so we can avoid future problems like the recursion limit exceeded.

AhmedTammaa:
@eyurtsev So I have committed the new updates. Please let me know if there are further modifications we need to make.

AhmedTammaa:
@eyurtsev, I need some help here. I have added bs4 to the poetry .toml file and updated the poetry lock. However, it still fails tests because `no module named bs4`. I checked this link: https://python.langchain.com/docs/contributing/how_to/code/setup/#testing

the `extended_testing_deps.txt` in text-splitters already contains bs4, likely because the section splitter class uses it. So, there must be something I have overlooked.

libs/text-splitters/pyproject.toml
```diff
@@ -37,6 +37,7 @@ ignore_missing_imports = "True"
 [tool.poetry.dependencies]
 python = ">=3.9,<4.0"
 langchain-core = "^0.3.25"
+beautifulsoup4 = "^4.12.3"
```
eyurtsev, line 40: You can remove this and revert the change to the lock file. We treat almost all deps as optional.

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -1,16 +1,76 @@
+'''
+This module provides classes for splitting HTML content into structured
+Document objects based on specified header tags.
+
+    ElementType(TypedDict): A typed dictionary representing an element type.
```
eyurtsev, line 5: We don't want to document any of these -- the API reference does this automatically by scanning the code. Poke around the code base e.g. look at the langchain core package to see how things are typically documented.

We follow Google document style for docstrings but without documenting the type as it's already documented via type hints

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -20,149 +80,451 @@ class ElementType(TypedDict):
...
+    name: str
+    tag_type: str
+    content: str
+    level: int
+    dom_depth: int
+    parent: Optional[Node] = field(default=None)
+
+class HTMLHeaderTextSplitter:
+    '''
+    HTMLHeaderTextSplitter is a class designed to split HTML content into
```
eyurtsev, line 106: What we want the docstring to do is explain what the code does. No user (or developer) is interested in reading the list of methods.. 

They care about what this splitter is supposed to do, what are the semantics, what's inside the generated docs etc

Importantly they want to learn this from as little text as possible. So the explanation needs to contain only what's essential and nothing more.

AhmedTammaa:
Hi @eyurtsev, 

I have committed the required changes from your comments. Please let me know if further modifications are required.

AhmedTammaa:
Hi @eyurtsev, I think you can have a look now. I have made the required changes. I figured out how to make the dependency optional by looking at other scripts in the repo. I made the documentation much clearer.

I have learnt a lot from this one. It is my first contribution to a project on such a scale, and I hope it gets merged. I am eager to contribute more in the future with the experience I have gained from this one.

Lastly, if we require any more changes please let me know.

libs/text-splitters/tests/unit_tests/test_text_splitters.py
```diff
@@ -2039,49 +2039,476 @@ def test_haskell_code_splitter() -> None:
...
+                    <h2>Background</h2>
+                    <p>Background information.</p>
+                    <h1>Conclusion</h1>
+                    <p>Final thoughts.</p>
+                </body>
+            </html>
+            """,
+            [
+                Document(
+                    page_content="Introduction", metadata={"Header 1": "Introduction"}
```
eyurtsev, line 2080: > Headers: Each specified header (e.g., <h1>, <h2>) becomes a separate Document containing the header text.

This is different from the behavior before. Which situation is this useful for? Is this for a case where an HTML document only contains the <h1> header and no other content inside the header?

----

How would this be used downstream?

Should we include a boolean flag to toggle this behavior?

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -33,149 +34,443 @@ class ElementType(TypedDict):
...
+        if isinstance(file, str):
+            with open(file, "r", encoding="utf-8") as f:
+                html_content = f.read()
+        else:
+            html_content = file.read()
 
-        # create filter and mapping for header metadata
-        header_filter = [header[0] for header in self.headers_to_split_on]
-        header_mapping = dict(self.headers_to_split_on)
+        soup = BeautifulSoup(html_content, "html.parser")
```
eyurtsev, line 353: perhaps move the parsing code that starts from in-memory content into a separate function, so that the parsing logic doesn't have to go through a file?

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -33,149 +34,443 @@ class ElementType(TypedDict):
     metadata: Dict[str, str]
 
 
+@dataclass
+class Node:
```
eyurtsev, line 38: Can we not rely on bs4 internal representation and avoid introducing new primitives in the implementation?

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -33,149 +34,443 @@ class ElementType(TypedDict):
...
+            The depth of the element in the DOM tree.
+        """
+        depth = 0
+        for _ in element.parents:
+            depth += 1
+        return depth
 
-    def split_text_from_url(self, url: str, **kwargs: Any) -> List[Document]:
-        """Split HTML from web URL.
+    def _build_tree(self, elements: Any) -> None:
```
eyurtsev, line 185: This shouldn't be necessary

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -3,6 +3,7 @@
 import copy
```
eyurtsev, line 3: The implementation feels very complex right now, is there a way to simplify it? Otherwise it doesn't help that much with maintaining.

What I imagine is having some variable to for booking keeping purposes that keeps track of the current dom_path in a representation that captures what we care about; e.g., dom_path = [('h1', character position of h1), ('h2', character position of h2)]

1. We'd walk through the tree and update the dom_path appropriately as we encounter header elements
2. When we detect a meaningful change in the dom_path:

```python
old_dom_path = [('h1',  142), ('h2', 382)]
new dom path  = [('h1',  142), ('h2', 582)]
```

We know that we hit a change in the document structure (e.g., here the h2 header changed). 

And based on that we can take an action (like yield the current document, and yield a new document for the heading h2)

In between changes, we just accumulate page_content into the currently running document.

```
dom_path = []
current_content = []

walk through tree 
   new_dom_path = dom_path.copy() + appropriate update if needed
   compare new dom path to dom path, 
    if no change
     current_content += ... 
   else: (change)
     yield Document(current_content, metadata=..)
    current_content = ""
    # and maybe yield a document for the change in the dom content)
```

This implementation can probably fit in a single function that's ~20 lines long and avoid having to rebuild anything?

libs/text-splitters/tests/unit_tests/test_text_splitters.py
AhmedTammaa: When I was checking old behaviour. Sometimes it create a document contains only the header text and then the next one contains the whats in between. So, I wrote it to be consistently creating a document contains the header text and then next one contains the the content inside. There is no strong reason for doing it like that. So, I can change it.

libs/text-splitters/langchain_text_splitters/html.py
AhmedTammaa: Yes, You are absolutely right. There can be maintainability issue due to the over complicated approach. I am currently working on simplifying it as you suggested. I won't rebuild a tree or use the Node class to rely only on bs4.

AhmedTammaa:
@eyurtsev I have simplified the implementation following your suggestions. Now, 
- No nodes or data class for them.
- No tree building, instead, splitting on the fly.
- The overall code is much more maintainable than before.

Please let me know if we need to do further modification.

eyurtsev:
I @AhmedTammaa sorry for the delay will review in a bit -- I was out for holidays for a couple of weeks

AhmedTammaa:
hi @eyurtsev.

I know that you are probably having a lot of code to review since the last break. This is just a gentle reminder for this PR :)

eyurtsev (approved):

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -34,148 +34,291 @@ class ElementType(TypedDict):
...
             ) from e
-        # use lxml library to parse html document and return xml ElementTree
-        # Explicitly encoding in utf-8 allows non-English
-        # html files to be processed without garbled characters
-        parser = etree.HTMLParser(encoding="utf-8")
-        tree = etree.parse(file, parser)
-
-        # document transformation for "structure-aware" chunking is handled with xsl.
-        # see comments in html_chunks_with_headers.xslt for more detailed information.
-        xslt_path = pathlib.Path(__file__).parent / "xsl/html_chunks_with_headers.xslt"
```
eyurtsev, line 126: We should probably delete this file as well (can be done in a separate PR)

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -34,148 +34,291 @@ class ElementType(TypedDict):
...
+            file: A file path or a file-like object containing HTML content.
+
+        Returns:
+            A list of split Document objects.
+        """
+        if isinstance(file, str):
+            with open(file, "r", encoding="utf-8") as f:
+                html_content = f.read()
+        else:
+            html_content = file.read()
```
eyurtsev, line 222: Implementation looks like it's working based on unit tests, so I think we're good.

I suspect that there's a way to further simplify, I would've tried the following:

1. create a private method that accepts html source and yields a generator over documents
2. the method itself would use a queue to do a tree traversal and have some variable to keep track of dom path and current content associated with the given dom path.

eyurtsev (approved):

libs/text-splitters/langchain_text_splitters/html.py
AhmedTammaa: Sure, I will try to enhance it even further and create a new PR. The new PR will also delete the file you suggested be removed.

Thank you very much for your invaluable review of my code. I definitely learnt, and I am eager to contribute more to Langchain!!

Issue #26971, opened 2024-09-29
Token-level streaming is unavailable in langchain_ollama.ChatOllama when tools are bound


u-haru:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### without bind_tools (works correctly):
```python
from langchain_ollama import ChatOllama
llm = ChatOllama(model="llama3.1", temperature=0)
for chunk in llm.stream("Tell me a joke"):
    print(chunk.content, end="|", flush=True)
```
stdout:
```
Here|'s| one|:

|What| do| you| call| a| fake| nood|le|?

|An| imp|asta|!

|Hope| that| made| you| laugh|!| Do| you| want| to| hear| another| one|?||
```

### with bind_tools:
```python
from langchain_ollama import ChatOllama
llm = ChatOllama(model="llama3.1", temperature=0).bind_tools([])
for chunk in llm.stream("Tell me a joke"):
    print(chunk.content, end="|", flush=True)
```
stdout:
```
Here's one:

What do you call a fake noodle?

An impasta!

Hope that made you laugh! Do you want to hear another one?|
```

### Description

I'm trying to use local LLMs with both tool calling and streaming.

I tested ChatOllama with some simple tools (like getting the current date), and then realized that streaming cannot be used when tools are bound, even if the tools list is empty.

I'm not aware if all models work with tool calling and streaming, but ChatGPT and Gemini seem to handle this correctly.

To confirm this, I ran the same prompt and tools using the Ollama library, and it worked as expected.

```python
import ollama
stream = ollama.chat(
	model="llama3.1",
	messages=[{'role': 'user', 'content': 'Tell me a joke'}],
	options={"temperature":0},
	stream=True,
	tools=[],
)
for chunk in stream:
	print(chunk['message']['content'], end='|', flush=True)
```

```
Here|'s| one|:

|What| do| you| call| a| fake| nood|le|?

|An| imp|asta|!

|Hope| that| made| you| laugh|!| Do| you| want| to| hear| another| one|?||
```

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.19045
[...]

PR #27689, opened 2024-10-28
partners/ollama: Enabled Token Level Streaming when Using Bind Tools for ChatOllama

ElhamBadri2411:
**Description:** The issue concerns the unexpected behavior observed using the bind_tools method in LangChain's ChatOllama. When tools are not bound, the llm.stream() method works as expected, returning incremental chunks of content, which is crucial for real-time applications such as conversational agents and live feedback systems. However, when bind_tools([]) is used, the streaming behavior changes, causing the output to be delivered in full chunks rather than incrementally. This change negatively impacts the user experience by breaking the real-time nature of the streaming mechanism.
**Issue:** #26971


ccurme (commented):
Does tool calling still work following this change?

I'm finding issues with tool calling integration tests when running locally (e.g., [this test](https://github.com/langchain-ai/langchain/blob/8a5807a6b4a99d79fe9d740958ffb5d692a972f1/libs/standard-tests/langchain_standard_tests/integration_tests/chat_models.py#L268)).

Unfortunately we don't run these in CI for Ollama. You can run locally via
```
cd libs/partners/ollama
poetry run python -m pytest tests/integration_tests/test_chat_models.py
```

ElhamBadri2411:
Hey, the tool calling does pass now, we will fix the out of date branches soon!

libs/partners/ollama/langchain_ollama/chat_models.py
```diff
@@ -476,15 +476,16 @@ async def _acreate_chat_stream(
 
         params["options"]["stop"] = stop
         if "tools" in kwargs:
-            yield await self._async_client.chat(
+            async for part in await self._async_client.chat(
```
ccurme, line 479: Do we need to also fix the async streaming case? We unfortunately don't cover async streaming with tools in our standard tests.

libs/partners/ollama/langchain_ollama/chat_models.py
```diff
@@ -514,15 +515,26 @@ def _create_chat_stream(
...
-            yield self._client.chat(
-                model=params["model"],
-                messages=ollama_messages,
-                stream=False,
-                options=Options(**params["options"]),
-                keep_alive=params["keep_alive"],
-                format=params["format"],
-                tools=kwargs["tools"],
-            )
+            if len(kwargs["tools"]) == 0:
```
ccurme, line 518: (nit) Wondering if there are any options for simplifying this:
- change `if "tools" in kwargs` to `if kwargs.get("tools")`, which is Falsey for empty list; or
- There's a method [_should_stream](https://github.com/langchain-ai/langchain/blob/00e7b2dada9085aad12ec47e7bd84089bad1d2bc/libs/core/langchain_core/language_models/chat_models.py#L362C21-L362C35) which is called on both `.stream` and `.astream` that controls whether we delegate to `.invoke`. Could potentially override this on ChatOllama. See example here: https://github.com/langchain-ai/langchain/blob/00e7b2dada9085aad12ec47e7bd84089bad1d2bc/libs/partners/openai/langchain_openai/chat_models/base.py#L982-L1002

ccurme (commented):
Thanks for your patience with the review. Only blocking question is on async streaming case.

libs/partners/ollama/langchain_ollama/chat_models.py
```diff
@@ -513,25 +514,26 @@ def _create_chat_stream(
...
-                model=params["model"],
-                messages=ollama_messages,
-                stream=False,
-                options=Options(**params["options"]),
-                keep_alive=params["keep_alive"],
-                format=params["format"],
-                tools=kwargs["tools"],
-            )
+
+        tools = kwargs.get("tools", None)
```
4meyDam1e, line 518: Hello, I've added the change you asked for to simplify this part of the code.

ccurme: Thank you!

libs/partners/ollama/langchain_ollama/chat_models.py
ccurme: I added a standard test for async tool calling this morning in https://github.com/langchain-ai/langchain/pull/28133. You can run it against Ollama via
```
python -m pytest tests/integration_tests/test_chat_models.py::TestChatOllama::test_tool_calling_async
```
It appeared broken here due to changing `stream` to True in the async case for tool calling. I pushed an update to fix the test. Confirmed no new failures across standard tests.

ccurme (approved):
Updated the async case. Let me know if you see any issues.

Thanks!

PR #27710, opened 2024-10-29
community: CrateDB: Vector Store

amotl:
### About
  - **Description:** Vector Store adapter for CrateDB.
  - **Coming from:** https://github.com/crate-workbench/langchain/pull/1
  - **Documentation:** GH-27713
  - **Addressed to:** @eyurtsev

### Status
- We are considering the patch ready for review and merging, with a few spots to be handled on a later iteration.
- Please let us know if you want to see any other details to be addressed before the initial merge.
- A few backlog items have been collected here: https://github.com/crate-workbench/langchain/issues/30.

### Sandbox
A little walkthrough how to exercise the software tests on your workstation.
```shell
docker run --rm -it --name=cratedb \
  --publish=4200:4200 --publish=5432:5432 --env=CRATE_HEAP_SIZE=2g \
  crate:latest -Cdiscovery.type=single-node
```
```shell
git clone https://github.com/crate-workbench/langchain.git --branch=cratedb-up/1/vector-store
cd langchain
uv venv
source .venv/bin/activate
cd libs/community
uv pip install --upgrade --prerelease=allow --editable=. poetry sqlalchemy-cratedb
poetry install --no-interaction --no-ansi --with dev,test,test_integration
```
```shell
pytest -vvv tests/integration_tests/vectorstores/test_cratedb.py
```

### Trivia
The CrateDB implementation is heavily based on PGVector's, with a few adjustments. Previous generalizations and improvements to PGVector have been submitted the other day already.

- GH-16243
- GH-16244

PR #27712, opened 2024-10-29
community: CrateDB: Conversational Memory

amotl:
### About
  - **Description:** Conversational Memory adapter for CrateDB.
  - **Coming from:** https://github.com/crate-workbench/langchain/pull/1
  - **Documentation:** GH-27713
  - **Addressed to:** @eyurtsev


amotl:
## Problem
With the current status of the patch, `pytest -vvv tests/integration_tests -k cratedb` trips like this when aiming to import the conversational memory adapter module from `langchain.memory`:
> ImportError: cannot import name 'CrateDBChatMessageHistory' from 'langchain.memory.chat_message_histories'
>
> -- https://github.com/langchain-ai/langchain/actions/runs/11583845932/job/32249838349?pr=27712#step:5:45

When trying to satisfy this admonition by adding import stubs to `libs/langchain`, like outline here,
```diff
(.venv311) sink:community amo$ git diff
diff --git a/libs/langchain/langchain/memory/chat_message_histories/__init__.py b/libs/langchain/langchain/memory/chat_message_histories/__init__.py
index 91910137f..ff95fd215 100644
--- a/libs/langchain/langchain/memory/chat_message_histories/__init__.py
+++ b/libs/langchain/langchain/memory/chat_message_histories/__init__.py
@@ -8,6 +8,7 @@ if TYPE_CHECKING:
         CassandraChatMessageHistory,
         ChatMessageHistory,
         CosmosDBChatMessageHistory,
+        CrateDBChatMessageHistory,
         DynamoDBChatMessageHistory,
         ElasticsearchChatMessageHistory,
         FileChatMessageHistory,
@@ -34,6 +35,7 @@ DEPRECATED_LOOKUP = {
     "CassandraChatMessageHistory": "langchain_community.chat_message_histories",
     "ChatMessageHistory": "langchain_community.chat_message_histories",
     "CosmosDBChatMessageHistory": "langchain_community.chat_message_histories",
+    "CrateDBChatMessageHistory": "langchain_community.chat_message_histories",
     "DynamoDBChatMessageHistory": "langchain_community.chat_message_histories",
     "ElasticsearchChatMessageHistory": "langchain_community.chat_message_histories",
     "FileChatMessageHistory": "langchain_community.chat_message_histories",
@@ -65,6 +67,7 @@ __all__ = [
     "CassandraChatMessageHistory",
     "ChatMessageHistory",
     "CosmosDBChatMessageHistory",
+    "CrateDBChatMessageHistory",
     "DynamoDBChatMessageHistory",
     "ElasticsearchChatMessageHistory",
     "FileChatMessageHistory",
```

... those tests succeed, but then CI trips on a protection policy like this:

> FAILED tests/unit_tests/test_imports.py::test_no_more_changes_to_proxy_community - AssertionError: If you're triggering this test, you're likely adding a new import to the langchain package that is importing something from langchain_community. This test is meant to catch such such imports as they are officially DEPRECATED. Please do not add any new imports from langchain_community to the langchain package. 
>
> -- https://github.com/langchain-ai/langchain/actions/runs/11583340242/job/32248332900?pr=27712#step:5:895

## Solution
Ah right, just updating `evil_magic_number` works well to make CI succeed. ;]

libs/community/langchain_community/chat_message_histories/cratedb.py
```diff
@@ -0,0 +1,109 @@
...
+    SQLChatMessageHistory,
+)
+
+
+def create_message_model(table_name, DynamicBase):  # type: ignore
+    """
+    Create a message model for a given table name.
+
+    This is a specialized version for CrateDB for generating integer-based primary keys.
+    TODO: Find a way to converge CrateDB's generate_random_uuid() into a variant
```
kneth, line 18: Can you do something like:

```python
b = bytearray()
b.extend(map(ord, uuid))
i = int.from_bytes(b, "big")
```

amotl: Thanks for spotting this. I think the TODO comment is from a previous version of the code. I think the canonical form to define primary keys with SQLAlchemy and CrateDB to synthesize the almost-equivalent of "autoincrement" columns is just that how it is written down here:

```python
sa.Column(sa.BigInteger, primary_key=True, server_default=sa.func.now())
```

FYI, FWIW: If you can't afford to define/override the SQLAlchemy ORM model definitions, but have to support a column definition using `primary_key=True, autoincrement=True`, and can't change it, there is a support method mentioned at [Synthetic Autoincrement using Timestamps](https://cratedb.com/docs/sqlalchemy-cratedb/support.html#synthetic-autoincrement-using-timestamps).

amotl: I've just removed the inline TODO comment. Thanks again.

kneth (approved):
I think it would be good to have an additional reviewer

efriis:
see #27710

PR #27716, opened 2024-10-29
community: Allow other than default parsers in SharePointLoader and OneDriveLoader

MacanPN:
### Currently `O365BaseLoader` (and consequently both derived loaders) are limited to `pdf`, `doc`, `docx` files.
- **Solution: here we introduce _handlers_ attribute that allows for custom handlers to be passed in. This is done in _dict_ form:**

**Example:**
```python
from langchain_community.document_loaders.parsers.documentloader_adapter import DocumentLoaderAsParser
# PR for DocumentLoaderAsParser here: https://github.com/langchain-ai/langchain/pull/27749
from langchain_community.document_loaders.excel import UnstructuredExcelLoader

xlsx_parser = DocumentLoaderAsParser(UnstructuredExcelLoader, mode="paged")

# create dictionary mapping file types to handlers (parsers)
handlers = {
    "doc": MsWordParser()
    "pdf": PDFMinerParser()
    "txt": TextParser()
    "xlsx": xlsx_parser
}
loader = SharePointLoader(document_library_id="...",
                            handlers=handlers # pass handlers to SharePointLoader
                            )
documents = loader.load()

# works the same in OneDriveLoader
loader = OneDriveLoader(document_library_id="...",
                            handlers=handlers
                            )
```
This dictionary is then passed to `MimeTypeBasedParser` same as in the [current implementation](https://github.com/langchain-ai/langchain/blob/5a2cfb49e045988d290a1c7e3a0c589d6b371694/libs/community/langchain_community/document_loaders/parsers/registry.py#L13). 

### Currently `SharePointLoader` and `OneDriveLoader` are separate loaders that both inherit from `O365BaseLoader`
However both of these implement the same functionality. The only differences are:
  - `SharePointLoader` requires argument `document_library_id` whereas `OneDriveLoader` requires `drive_id`. These are just different names for the same thing.
  - `SharePointLoader` implements significantly more features.
- **Solution: `OneDriveLoader` is replaced with an empty shell just renaming `drive_id` to `document_library_id` and inheriting from `SharePointLoader`**

**Dependencies:** None
**Twitter handle:** @martintriska1


libs/community/langchain_community/document_loaders/base_o365.py
```diff
@@ -46,24 +48,27 @@ class _O365TokenStorage(BaseSettings):
...
 
-class _FileType(str, Enum):
-    DOC = "doc"
-    DOCX = "docx"
-    PDF = "pdf"
+def fetch_mime_types(file_types: Sequence[str]) -> Dict[str, str]:
+    """Fetch the mime types for the specified file types."""
+    mime_types_mapping = {}
+    for ext in file_types:
+        mime_type, _ = mimetypes.guess_type(f"file.{ext}")
```
vbarda, line 55: could we just use a simple dictionary lookup instead?

vbarda: (and same for below)

libs/community/langchain_community/document_loaders/base_o365.py
```diff
@@ -78,16 +83,84 @@ class O365BaseLoader(BaseLoader, BaseModel):
     """Number of bytes to retrieve from each api call to the server. int or 'auto'."""
     recursive: bool = False
     """Should the loader recursively load subfolders?"""
+    handlers: Optional[Dict[str, Any]] = {}
```
vbarda, line 86: for the keys (file types) let's either keep the original `FileType` enum, or use a `Literal["doc", "pdf", "txt"]` for the annotation

libs/community/langchain_community/document_loaders/onedrive.py
```diff
@@ -1,94 +1,16 @@
...
-        for subfolder in subfolders:
-            try:
-                subfolder_drive = list(filter(lambda x: subfolder in x.name, items))[0]
-                items = subfolder_drive.get_items()
-            except (IndexError, AttributeError):
-                raise FileNotFoundError("Path {} not exist.".format(self.folder_path))
-        return subfolder_drive
+    """The ID of the OneDrive drive to load data from."""
 
-    def lazy_load(self) -> Iterator[Document]:
```
vbarda, line 76: is this removed because the one in `SharePointLoader` is more comprehensive?

libs/community/langchain_community/document_loaders/onedrive.py
```diff
@@ -1,94 +1,16 @@
...
-    from O365.drive import Drive, Folder
-
-logger = logging.getLogger(__name__)
+from langchain_community.document_loaders import SharePointLoader
 
 
-class OneDriveLoader(O365BaseLoader):
-    """Load from `Microsoft OneDrive`."""
+class OneDriveLoader(SharePointLoader):
+    """Load documents from OneDrive. Uses `SharePointLoader` under the hood."""
```
vbarda, line 9: why remove `Microsoft`?

vbarda (commented):
Thanks for your contribution! 

Looks good overall, but my biggest question is: can't we just allow a single optional `MimeTypeBasedParser` as an input to this document loader? I think it could simplify a lot of this logic. The user could then do something like:

```python
blob_parser = MimeTypeBasedParser({
    "application/msword": MsWordParser(),
    "application/pdf": PDFMinerParser(),
    ""audio/mpeg": OpenAIWhisperParser()
})
loader = OneDriveLoader(..., blob_parser=blob_parser)
```

libs/community/langchain_community/document_loaders/base_o365.py
vbarda: (pending the higher level comment about simply passing a parser)

libs/community/langchain_community/document_loaders/onedrive.py
vbarda: (saw you mention this in the PR description)

libs/community/langchain_community/document_loaders/base_o365.py
MacanPN: Who's going to maintain that lookup dict? Under the hood that is exactly what `mimetypes` does, except that library _specializes_ in this so their lookup will be always more up to date and complete. Why not offload the conversion to a library?

libs/community/langchain_community/document_loaders/onedrive.py
MacanPN: I rewrote the whole comment and ended up leaving out "Microsoft". I can put that back.

MacanPN:
> Thanks for your contribution!
> 
> Looks good overall, but my biggest question is: can't we just allow a single optional `MimeTypeBasedParser` as an input to this document loader? I think it could simplify a lot of this logic. The user could then do something like:
> 
> ```python
> blob_parser = MimeTypeBasedParser({
>     "application/msword": MsWordParser(),
>     "application/pdf": PDFMinerParser(),
>     ""audio/mpeg": OpenAIWhisperParser()
> })
> loader = OneDriveLoader(..., blob_parser=blob_parser)
> ```

This is a good point. I'll change the implementation so we're passing in `Parser` rather than `handlers` dictionary.

MacanPN:
In general the extra logic implemented in this PR is not about instantiation of `MimeTypeBasedParser` (that is 1 line thing) but rather about supporting arbitrary file types (rather than just 3 enumerated types).

libs/community/langchain_community/document_loaders/base_o365.py
vbarda: fair point!

libs/community/langchain_community/document_loaders/base_o365.py
```diff
@@ -78,16 +83,84 @@ class O365BaseLoader(BaseLoader, BaseModel):
...
 
-    @property
-    @abstractmethod
-    def _file_types(self) -> Sequence[_FileType]:
-        """Return supported file types."""
+            self._blob_parser = MimeTypeBasedParser(
+                handlers=mime_handlers, fallback_parser=None
+            )
+        else:
+            from langchain_community.document_loaders.parsers.registry import get_parser
```
vbarda, line 150: let's import at the top of the file

vbarda (approved):

vbarda:
> > Thanks for your contribution!
> > Looks good overall, but my biggest question is: can't we just allow a single optional `MimeTypeBasedParser` as an input to this document loader? I think it could simplify a lot of this logic. The user could then do something like:
> > ```python
> > blob_parser = MimeTypeBasedParser({
> >     "application/msword": MsWordParser(),
> >     "application/pdf": PDFMinerParser(),
> >     ""audio/mpeg": OpenAIWhisperParser()
> > })
> > loader = OneDriveLoader(..., blob_parser=blob_parser)
> > ```
> 
> This is a good point. I've been thinking about this but consider:
> 
> * We still need to implement functions for `file extension <-> mime type` conversions.
> * We still need to set up `_file_types` and `_mime_types` properties (for interface compatibility)
> * We still need to provide default parser.
>   So we wouldn't really simplify the logic that much. What the implementation has on top is option to pass in file types (file extensions) instead of mime types, which I think is a real quality of life improvement. Some mime types can be not as straight forward as the ones in the example above, such as:
> * `.docx`: `application/vnd.openxmlformats-officedocument.wordprocessingml.document`
> * `xlsx`: `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`
> 
> Please also consider that one could not simply pass in general `BaseBlobParser` but it would have to be `MimeTypeBasedParser` (because we are looking at `handlers` to determine whether the file should be downloaded and parsed or not).

Makes sense, thanks for the follow up! Approved pending two small nits. IMO this is the key argument in favor of this change:

> What the implementation has on top is option to pass in file types (file extensions) instead of mime types, which I think is a real quality of life improvement. Some mime types can be not as straight forward as the ones in the example above, such as:
> .docx: application/vnd.openxmlformats-officedocument.wordprocessingml.document
> xlsx: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet

libs/community/langchain_community/document_loaders/onedrive.py
MacanPN: Done

libs/community/langchain_community/document_loaders/base_o365.py
MacanPN: Done

MacanPN:
@vbarda Minor nitpicks addressed. Are we good to merge?

MacanPN:
@vbarda I can't merge it myself. Can you please merge the PR?
image

rasc0der:
Thanks for your contribution!

PR #27729, opened 2024-10-30
Support message trimming on single messages

hinthornw:
Permit trimming message lists of length 1


libs/core/tests/unit_tests/messages/test_utils.py
```diff
@@ -301,6 +301,24 @@ def test_trim_messages_last_40_include_system_allow_partial_start_on_human() ->
     assert _MESSAGES_TO_TRIM == _MESSAGES_TO_TRIM_COPY
 
 
+def test_trim_messages_allow_partial_one_message() -> None:
```
baskaryan, line 304: could we add check for strategy="last" too

PR #27730, opened 2024-10-30
rfc: AIMessage.parsed and with_structured_output(..., tools=[])

baskaryan:


baskaryan:
enables this type of usage

Screenshot 2024-10-29 at 7 54 26 PM

PR #27749, opened 2024-10-30
community: DocumentLoaderAsParser wrapper

MacanPN:
## Description

This pull request introduces the `DocumentLoaderAsParser` class, which acts as an adapter to transform document loaders into parsers within the LangChain framework. The class enables document loaders that accept a `file_path` parameter to be utilized as blob parsers. This is particularly useful for integrating various document loading capabilities seamlessly into the LangChain ecosystem.

When merged in together with PR https://github.com/langchain-ai/langchain/pull/27716 It opens options for `SharePointLoader` / `OneDriveLoader` to process any filetype that has a document loader.

### Features

- **Flexible Parsing**: The `DocumentLoaderAsParser` class can adapt any document loader that meets the criteria of accepting a `file_path` argument, allowing for lazy parsing of documents.
- **Compatibility**: The class has been designed to work with various document loaders, making it versatile for different use cases.

### Usage Example

To use the `DocumentLoaderAsParser`, you would initialize it with a suitable document loader class and any required parameters. Here’s an example of how to do this with the `UnstructuredExcelLoader`:

```python
from langchain_community.document_loaders.blob_loaders import Blob
from langchain_community.document_loaders.parsers.documentloader_adapter import DocumentLoaderAsParser
from langchain_community.document_loaders.excel import UnstructuredExcelLoader

# Initialize the parser adapter with UnstructuredExcelLoader
xlsx_parser = DocumentLoaderAsParser(UnstructuredExcelLoader, mode="paged")

# Use parser, for ex. pass it to MimeTypeBasedParser
MimeTypeBasedParser(
    handlers={
        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": xlsx_parser
    }
)
```

- **Dependencies:** None
- **Twitter handle:** @martintriska1


MacanPN:
I've written down am extra explanation of why one might need to mutate Document Loader into parser. If you're unsure about that, please keep reading:
- Document loaders function as `<whatever> -> documents`
- BlobParsers process `blob -> documents`.
- Many document loaders actually implement logic until they have a blob. Then the actual parsing is done by calling parser with similar name. As an example you can look at `AmazonTextractPDFLoader` and `AmazonTextractPDFParser`.
- Some document loaders may want to process numerous filetypes. An example is a `SharePointLoader` that is supposed to fetch all parsable files, parse them and produce documents. Until recently it processed only 3 file types `doc, docx, pdf` and ignored everything else. Recently I got [this pr](https://github.com/langchain-ai/langchain/pull/27716) merged that enables a user to pass in dict mapping file types (or mime types) to parsers. Now one can easily use `SharePointLoader` with any files where suitable parser is available.
- Unfortunatelly many document loaders do not define `BlobParser` but rather handle everything directly inside the loader class. `UnstructuredLoader` is one of those. 
- Originally I went ahead and implemented separate `ExcelParser` (this PR). However this duplicated some of the code. That could be mitigated by replacing the part of the code inside `UnstructuredExcelLoader` responsible for parsing the file with calling the `Parser`. This would basically bring it to agreement with how things are done for ex. in PDF Loaders.
- However I found out that I'd have to do this shuffling for a fairly long list of document loaders. Given how much time such effort would require, I opted to create `DocumentLoaderAsParser` wrapper. This wrapper can be used on any DocumentLoader that accepts `file_path` argument and turns it into `BaseBlobParser`.

So the idea is to be able to do:
```
xlsx_parser = DocumentLoaderAsParser(UnstructuredExcelLoader, mode="paged")
mp3_parser = DocumentLoaderAsParser(GoogleSpeechToTextLoader, project_id="...")

# create dictionary mapping file types to handlers (parsers)
handlers = {
    "xlsx": xlsx_parser,
    "mp3": mp3_parser
}
loader = SharePointLoader(document_library_id="...",
                            handlers=handlers # pass handlers to SharePointLoader
                            )
documents = loader.load()
```

Hope this makes sense.

MacanPN:
@vbarda Please take a look :)

MacanPN:
@efriis Would you mind taking a look yourself please? Seems like @vbarda might not have the throughput right now? Thank you and please let me know if there is anything I can do on my part to move this forward.

libs/community/langchain_community/document_loaders/parsers/documentloader_adapter.py
```diff
@@ -0,0 +1,61 @@
...
+                f"{document_loader_class.__name__} does not accept `file_path`."
+                "Only document loaders with `file_path` parameter"
+                "can be morphed into a parser."
+            )
+
+    def lazy_parse(self, blob: Blob) -> Iterator[Document]:
+        """
+        Use underlying DocumentLoader to lazily parse the blob.
+        """
+        doc_loader = self.DocumentLoaderClass(
```
vbarda, line 58: why not just create this on init?

libs/community/langchain_community/document_loaders/parsers/documentloader_adapter.py
```diff
@@ -0,0 +1,61 @@
...
+        # Ensure the document loader class has a `file_path` parameter
+        init_signature = inspect.signature(document_loader_class.__init__)
+        if "file_path" not in init_signature.parameters:
+            raise TypeError(
+                f"{document_loader_class.__name__} does not accept `file_path`."
+                "Only document loaders with `file_path` parameter"
+                "can be morphed into a parser."
+            )
+
+    def lazy_parse(self, blob: Blob) -> Iterator[Document]:
```
vbarda, line 54: i think we should add `parse` as well

libs/community/langchain_community/document_loaders/parsers/documentloader_adapter.py
```diff
@@ -0,0 +1,61 @@
...
+            excel_parser = DocumentLoaderAsParser(UnstructuredExcelLoader, mode="elements")
+            ```
+        """  # noqa: E501
+        super().__init__()
+        self.DocumentLoaderClass = document_loader_class
+        self.document_loader_kwargs = kwargs
+
+        # Ensure the document loader class has a `file_path` parameter
+        init_signature = inspect.signature(document_loader_class.__init__)
+        if "file_path" not in init_signature.parameters:
```
vbarda, line 47: is this a sufficient condition for being able to convert?

vbarda (commented):
apologies for the delay -- looks good high level, added some questions

ccurme (commented):
Thanks for this.

1. Can we add tests?
2. Thinking through alternatives, does it make sense to add a "sub" document loader class as an optional attribute of `O365BaseLoader` or `SharePointLoader`? Introducing a new abstraction brings some risk if we don't get the interface right (or it's not as generic / extendable as we thought).
3. If we were to introduce a new class or method, consider decorating it with a `@beta` parameter if we're not 100% on the API (see example [here](https://github.com/langchain-ai/langchain/blob/f5f53d1101ea73d8465deb7d73b0a4e70bb556e7/libs/community/langchain_community/graph_vectorstores/base.py#L708)).

Regarding unstructured: I would favor use of the `langchain-unstructured` package over the integrations in community. The community integrations are older and could be deprecated but we have not put in the work to verify that we don't lose functionality in `langchain-unstructured`.

libs/community/langchain_community/document_loaders/parsers/documentloader_adapter.py
MacanPN: Because on init, you don't know the `file_path`. The file path is pulled from `blob` when `lazy_parse` is called. This is consistent with all other parsers.

libs/community/langchain_community/document_loaders/parsers/documentloader_adapter.py
MacanPN: Yes, that is the only condition we're adding. Of course any number of problems can arise if the provided document loader class cannot load provided file, however that would be handled inside of given document loader.

libs/community/langchain_community/document_loaders/parsers/documentloader_adapter.py
MacanPN: `parse` method is created inside `BaseBlobParser` [here](https://github.com/langchain-ai/langchain/blob/65deeddd5dfec5d51f33ebc961f09c2e47a8f064/libs/core/langchain_core/document_loaders/base.py#L112). calling `.parse()` will end up calling `list()` on `.lazy_parse()`

MacanPN (commented):
Answered code comments. Please take a look

MacanPN:
> Thanks for this.
> 
>     1. Can we add tests?

I've had tests in [previous commits](https://github.com/langchain-ai/langchain/pull/27749/commits/3efd2b179c0f54b2d99be04caaafa9e5c6d59ea1#diff-9b012a9c1ec582e18b537a5448df73ada0fe9f1e5bc62dfe7609f0dcb6b90704) that I had rolled back. The problem was that for the tests to run, I needed to add `unstructured` module to requirements. I was getting:
`INTERNALERROR> Failed: Package `unstructured` is not installed but is required for extended tests. Please install the given package and try again.`
Back when working on this, I remember developers were asked not to add dependencies to pyproject.toml? I see there is new guidance now, I'll try to get the test working and will report back shortly.

> 
>     2. Thinking through alternatives, does it make sense to add a "sub" document loader class as an optional attribute of `O365BaseLoader` or `SharePointLoader`? Introducing a new abstraction brings some risk if we don't get the interface right (or it's not as generic / extendable as we thought).

This was my original thought. However after spending some time on this issue, I found this approach much better. Except for cleaner code, it should help with reusability. 
In general, this is very simple concept. Many document loaders accept `file_path` and return list of documents. Parsers accept `blob` and return list of documents. All we need to do is to open the `file_path` as `blob` and call corresponding loader. There are classes that require `parser(s)` to be passed in. We can have either universal adapter, or we can go hunting for piece of code where `parser` is used and update the code so `document_loader` could be used instead.
> 
>     3. If we were to introduce a new class or method, consider decorating it with a `@beta` parameter if we're not 100% on the API (see example [here](https://github.com/langchain-ai/langchain/blob/f5f53d1101ea73d8465deb7d73b0a4e70bb556e7/libs/community/langchain_community/graph_vectorstores/base.py#L708)).

I can add `@beta` parameter. What version should I say that it will be added in?

> 
> 
> Regarding unstructured: I would favor use of the `langchain-unstructured` package over the integrations in community. The community integrations are older and could be deprecated but we have not put in the work to verify that we don't lose functionality in `langchain-unstructured`.
Got it, I will be using `langchain-unstructured` going forward. It should not have material influence on this PR since `DocumentLoaderAsParser` can be used on either (any many other loaders)

libs/community/langchain_community/document_loaders/parsers/documentloader_adapter.py
```diff
@@ -0,0 +1,61 @@
...
+                "Only document loaders with `file_path` parameter"
+                "can be morphed into a parser."
+            )
+
+    def lazy_parse(self, blob: Blob) -> Iterator[Document]:
+        """
+        Use underlying DocumentLoader to lazily parse the blob.
+        """
+        doc_loader = self.DocumentLoaderClass(
+            file_path=blob.path, **self.document_loader_kwargs
```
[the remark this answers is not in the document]
MacanPN: Yes, you're right - that would be the ideal solution. On the O365 doc loader side I've already prepared the ground in [this PR](https://github.com/langchain-ai/langchain/pull/27716). The problem is that there is a lot of document loaders that do not have that separation. Therefore there is no blob parser to point to. I actually wrote up a wrapper for *unstructured* to serve as `.xlsx` parser [here](https://github.com/langchain-ai/langchain/pull/27103) but
1. People were confused why are we re-implementing something already done in the unstructured document loader
2. There is large number of document loaders that are doing this. I don't really fency trying to go inside all of them and try to separate out the parser from the blob loader. I'm all for marking this class as a temporary fix and if someone finds time to refactor all those doc loaders, we can retire the class...

Hope that makes sense!

MacanPN:
ps: sorry for the mess in last 2 commits, I screwed up a rebase onto master and had to reset and rebase again...

efriis (commented):
Do you have a list of the loaders this would be relevant for?

Would have a strong preference for refactoring those to be `BaseBlobParser` objects, and creating new replacement classes that are just `GenericLoader` with a filesystem loader to maintain same interface

PR #27756, opened 2024-10-30
core[patch]: make oai tool description optional

baskaryan:


ccurme (approved):
Curious when these became optional (Dec '23? https://platform.openai.com/docs/changelog)

PR #27761, opened 2024-10-30
core[patch]: make get_all_basemodel_annotations public

baskaryan:


vbarda (approved):

libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -1891,6 +1891,13 @@ async def test_prompt_with_chat_model_async(
     )
 
 
+@pytest.mark.skipif(
```
baskaryan, line 1894: cc @nfcampos

PR #27787, opened 2024-10-31
Improvement [docs] Improve api docs

ZhangShenao:
- Add missing param
- Remove unused param


ccurme (approved):
Thanks for this. I kept `tool_choice` in the docstring for `ChatMistralAI.bind_tools` as it's a valid / important kwarg. May be better to make it explicit but IMO it's helpful in the API ref. Let me know if you disagree.

ZhangShenao:
> Thanks for this. I kept `tool_choice` in the docstring for `ChatMistralAI.bind_tools` as it's a valid / important kwarg. May be better to make it explicit but IMO it's helpful in the API ref. Let me know if you disagree.

No problem, I agree with this.

PR #27797, opened 2024-10-31
core[patch]: rm image loading

baskaryan:
Patch for v0.3.x branch


libs/core/langchain_core/prompts/image.py
```diff
@@ -33,6 +32,12 @@ def __init__(self, **kwargs: Any) -> None:
                 f" Found: {overlap}"
             )
             raise ValueError(msg)
+        if "path" in kwargs["input_variables"]:
```
efriis, line 35: this will never get hit right? will always fail overlap test above?

efriis: I actually think not important to raise here because this is just about variable names - not what gets passed in where

should we instead validate the `template` dict to make sure it doesn't have a path key?

baskaryan: good point re overlap

efriis (approved):

libs/core/langchain_core/utils/image.py
```diff
@@ -1,29 +1,8 @@
...
-
-    Returns:
-        The data URL of the image.
-    """
-    encoding = encode_image(image_path)
-    mime_type = mimetypes.guess_type(image_path)[0]
-    return f"data:{mime_type};base64,{encoding}"
+def __getattr__(name: str) -> Any:
+    if name in ("encode_image", "image_to_data_url"):
+        msg = f"'{name}' has been removed for security reasons."
```
eyurtsev, line 6: This is very cryptic!

1) why did we remove encode_image / image_to_data_url are those not fine if users just use them as a utility? (i also get that they're really light weight so maybe we don't need them.)
2) The error message is cryptic. Could we provide at least one more detail? It's not clear why the utility functions themselves would be a security concern.


The main concern is that this going to break user code, so would be ideal to provide information. Including the version as of which this change was made

bachp:
Is there any information why this was removed? As this breaks already working code in the 0.3 branch it would be nice to at least know why this is a security issue.

baskaryan:
hi @bachp, good callout

You can create langchain_core.prompts.ImagePromptTemplate's (and by extension the langchain_core.prompts.ChatPromptTemplate's) with input variables that make it possible for the prompt template to read any user-specified path from the server file system. If the outputs of the prompt template are exposed to the user (directly or e.g. by being passed to a model, whose outputs are then passed to user) this allows the user to read the contents of local files.

```python
from langchain_core.prompts import ChatPromptTemplate, HumanMessagePromptTemplate

prompt = ChatPromptTemplate([
  HumanMessagePromptTemplate.from_template([{"type": "image_url", "image_url": {"path": "{image_path}"}}])
])

# input any file path. note it does not need to be an image file.
prompt.invoke({"image_path": "/path/to/private/file/on/server.xyz"})

# output contains base64 encoded str contents of the file
# -> ChatPromptValue(messages=[HumanMessage(content=[{"type": "image_url", "image_url": {"url": "data:{mime_type...};base64,{encoding...}"}])])

# # using with a model

llm = init_chat_model("gpt-4o-mini")
chain = prompt | llm
# note the file does need to be an image for the model to respond
chain.invoke({"image_path": "/path/to/private/file/on/server.jpg"})
```

If an application allows users to specify prompt templates or inputs to prompt templates with image url variables, and the outputs of those templates are exposed in any way (directly in e.g. traces or via downstream model outputs), then users will be able to read arbitrary files on host machines.

Hope this helps!

Issue #27790, opened 2024-10-31
Cannot create MistralAI embeddings from pdf or urls


PabloVD:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_chroma import Chroma
from langchain_mistralai import MistralAIEmbeddings
from langchain_text_splitters import RecursiveCharacterTextSplitter
from langchain_community.document_loaders import PyPDFLoader
import requests
from pathlib import Path

# Get data from url
url = 'https://camels.readthedocs.io/_/downloads/en/latest/pdf/'
r = requests.get(url, stream=True)
document_path = Path('data.pdf')

document_path.write_bytes(r.content)
# document_path = "camels-readthedocs-io-en-latest.pdf"
loader = PyPDFLoader(document_path)
docs = loader.load()

# Split text
text_splitter = RecursiveCharacterTextSplitter(chunk_size=1000, chunk_overlap=200)
splits = text_splitter.split_documents(docs)

# Create vector store
embeddings = MistralAIEmbeddings()
vectorstore = Chroma.from_documents(documents=splits, embedding=embeddings)
```

### Error Message and Stack Trace (if applicable)

```
An error occurred with MistralAI: 'data'
Traceback (most recent call last):
  File "/home/tda/GenAICourse/RAG/ragbot_langchain/embeddings.py", line 24, in <module>
    vectorstore = Chroma.from_documents(documents=splits, embedding=embeddings)
  File "/home/tda/miniconda3/envs/ragbot2/lib/python3.10/site-packages/langchain_chroma/vectorstores.py", line 1128, in from_documents
    return cls.from_texts(
  File "/home/tda/miniconda3/envs/ragbot2/lib/python3.10/site-packages/langchain_chroma/vectorstores.py", line 1089, in from_texts
    chroma_collection.add_texts(texts=texts, metadatas=metadatas, ids=ids)
  File "/home/tda/miniconda3/envs/ragbot2/lib/python3.10/site-packages/langchain_chroma/vectorstores.py", line 508, in add_texts
    embeddings = self._embedding_function.embed_documents(texts)
[...]

PR #27818, opened 2024-11-01
mistral[minor]: Added Retrying Mechanism in case of Request Rate Limit Error for `MistralAIEmbeddings`

keenborder786:
- **Description:**: In the event of a Rate Limit Error from the MistralAI server, the response JSON raises a KeyError. To address this, a simple retry mechanism has been implemented to handle cases where the request limit is exceeded. 
  - **Issue:** #27790


keenborder786:
@eyurtsev

libs/partners/mistralai/langchain_mistralai/embeddings.py
```diff
@@ -209,16 +211,27 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
             List of embeddings, one for each text.
         """
         try:
-            batch_responses = (
-                self.client.post(
+            batch_responses = []
+
+            @retry(
+                retry=retry_if_exception_type(Exception),
```
eyurtsev, line 217: Generally, retries should never be implemented on 4xx errors (except for 408 and 429). e.g., 403 should not be retried by default

* What do we do in other parts of the code? Perhaps there's a better example that can be adopted?
* What do other models do in the code base in terms of exposing the retry parameters so users can adjust? (e.g.,what if someone wants to have the first retry after 1 second rather than 30 seconds?)

libs/partners/mistralai/langchain_mistralai/embeddings.py
```diff
@@ -209,16 +211,27 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
...
                     url="/embeddings",
                     json=dict(
                         model=self.model,
                         input=batch,
                     ),
                 )
-                for batch in self._get_batches(texts)
-            )
+                if response.status_code == 429:
+                    raise Exception("Requests rate limit exceeded")
```
eyurtsev, line 230: This code takes an exception that was good and informative and turns it into one that's a broad Exception of type Exception -- this is usually not a good pattern for exception handling. Stack trace will be partially lost, the exception type is less specific etc.

eyurtsev (changes requested):
Makes sense to add retry mechanism. Added a few questions to see if we can improve how it's configured

libs/partners/mistralai/langchain_mistralai/embeddings.py
keenborder786: - @eyurtsev it is a 429 error therefore retrying makes sense.
- I was not able to find any related example.
- Yes I can make the wait and stop seconds as parameters.

libs/partners/mistralai/langchain_mistralai/embeddings.py
keenborder786: - Their were no specific exception being raised to being with. But I can change it from a general Exception.

keenborder786:
@eyurtsev please check now

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev this is really important, please give feedback if needed

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev

libs/partners/mistralai/langchain_mistralai/embeddings.py
```diff
@@ -215,16 +220,27 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
...
+                response = self.client.post(
                     url="/embeddings",
                     json=dict(
                         model=self.model,
                         input=batch,
                     ),
                 )
-                for batch in self._get_batches(texts)
-            )
+                if response.status_code == 429:
```
eyurtsev, line 238: any reason not to use  `raise_for_status`?  We're trying not to drop the original exception which might have useful information inside it

libs/partners/mistralai/langchain_mistralai/embeddings.py
```diff
@@ -215,16 +220,27 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
             List of embeddings, one for each text.
         """
         try:
-            batch_responses = (
-                self.client.post(
+            batch_responses = []
+
+            @retry(
+                retry=retry_if_exception_type(httpx.TimeoutException),
```
eyurtsev, line 226: (no need to change if you don't want) This is OK b/c it's probably the dominant failure mode

But it's very common to retry 5xx errors as well. And 408 (request timeout)

eyurtsev (approved):

eyurtsev:
@keenborder786 sorry for how long it took. Pushed a minor change to raise on error, so originally information isn't lost

PR #27820, opened 2024-11-01
docs: replace invoke with run for PubMed query execution 

francemazzi:
Updated to use run instead of invoke in PubmedQueryRun to prevent execution errors like "object has no attribute 'invoke'"" and ensure compatibility with the latest API syntax.

### DIFF
--- a/docs/docs/integrations/tools/pubmed.ipynb
+++ b/docs/docs/integrations/tools/pubmed.ipynb
@@ -60,7 +60,7 @@
     }
    ],
    "source": [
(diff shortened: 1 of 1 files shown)


ccurme:
[PubmedQueryRun](https://python.langchain.com/api_reference/community/tools/langchain_community.tools.pubmed.tool.PubmedQueryRun.html) subclasses BaseTool, so it implements `invoke`. That is the latest and recommended syntax. Can you confirm what langchain library versions you are using?

PR #27827, opened 2024-11-01
community: chroma error patch(attribute changed on chroma)

shjunn:
There was a change of attribute name which was "max_batch_size". It's now "get_max_batch_size" method.
I want to use "create_batches" which is right down below.

Please check this PR link.
reference: https://github.com/chroma-core/chroma/pull/2305


efriis (commented):
this is unfortunately a breaking change caused by a breaking change in their client

should this just be an OR case to make it work for both?

shjunn:
OR case can be a safe way but their client changed the attribute to method because of https://github.com/chroma-core/chroma/issues/2403 (Preventing a max_batch_size not to be set.)
Once they did it, I believe they will not revert the name.
If this code could be another bug, I respect your suggestion and I'll edit this code.
Please let me know

efriis:
old versions of chroma that have this function are still available, and we don't want to break those folks' code

shjunn:
Thanks for letting me know.

efriis (approved):

PR #27846, opened 2024-11-01
community: fixed bug in GraphVectorStoreRetriever

epinzur:
Description:

This fixes an issue that mistakenly created in https://github.com/langchain-ai/langchain/pull/27253. The issue currently exists only in `langchain-community==0.3.4`.

Test cases were added to prevent this issue in the future.


libs/community/langchain_community/graph_vectorstores/base.py
```diff
@@ -837,8 +838,8 @@ class GraphVectorStoreRetriever(VectorStoreRetriever):
         retriever = graph_vectorstore.as_retriever(search_kwargs={"score_threshold": 0.5})
     """  # noqa: E501
 
-    vector_store: GraphVectorStore
-    """GraphVectorStore to use for retrieval."""
+    vectorstore: VectorStore
```
epinzur, line 841: changing this from `vectorstore` to `vector_store` created the bug.

libs/community/langchain_community/graph_vectorstores/base.py
```diff
@@ -849,14 +850,20 @@ class GraphVectorStoreRetriever(VectorStoreRetriever):
         "mmr_traversal",
     )
 
+    @property
+    def graph_vectorstore(self) -> GraphVectorStore:
+        return cast(GraphVectorStore, self.vectorstore)
+
```
epinzur, line 856: I had to add this cast property to get the 3.13 lint to pass. If there is a better way to do this, please let me know.

libs/community/tests/integration_tests/graph_vectorstores/test_cassandra.py
efriis, line 1: jfyi the community integration tests don't run in CI ever, so might be more effective to add this to the astradb ci instead

efriis (approved):

Issue #27826, opened 2024-11-01
anthropic_api_key not used for ChatLiteLLM


chenzimin:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code:

```
from langchain_community.chat_models import ChatLiteLLM
from langchain_core.messages import HumanMessage

chat = ChatLiteLLM(model="claude-3-haiku-20240307", anthropic_api_key="...")

messages = [
    HumanMessage(
        content="Translate this sentence from English to French. I love programming."
    )
]
chat(messages)
```

will raise an error
```
AuthenticationError: litellm.AuthenticationError: Missing Anthropic API Key - A call is being made to anthropic but no key is set either in the environment variables or via params. Please set `ANTHROPIC_API_KEY` in your environment vars
```

However, setting the environment variable will work
```
import os
from langchain_community.chat_models import ChatLiteLLM
from langchain_core.messages import HumanMessage
os.environ["ANTROPIC_API_KEY"] = "xxx"

chat = ChatLiteLLM(model="claude-3-haiku-20240307")

messages = [
    HumanMessage(
        content="Translate this sentence from English to French. I love programming."
    )
]
chat(messages)
```

### Error Message and Stack Trace (if applicable)

Full stack trace
```
/usr/local/lib/python3.10/dist-packages/langchain_core/_api/deprecation.py in warning_emitting_wrapper(*args, **kwargs)
    180                 warned = True
    181                 emit_warning()
--> 182             return wrapped(*args, **kwargs)
    183 
    184         async def awarning_emitting_wrapper(*args: Any, **kwargs: Any) -> Any:

/usr/local/lib/python3.10/dist-packages/langchain_core/language_models/chat_models.py in __call__(self, messages, stop, callbacks, **kwargs)
   1015         **kwargs: Any,
   1016     ) -> BaseMessage:
-> 1017         generation = self.generate(
   1018             [messages], stop=stop, callbacks=callbacks, **kwargs
   1019         ).generations[0][0]

/usr/local/lib/python3.10/dist-packages/langchain_core/language_models/chat_models.py in generate(self, messages, stop, callbacks, tags, metadata, run_name, run_id, **kwargs)
[...]

PR #27851, opened 2024-11-01
Set api_key param to litellm client based on providers api keys

jamesev15:
- **PR title**: "ChatLiteLLM: set api_key to litellm client"

-  **PR message**:
    - **Description:** The litellm client requires an api_key parameter, but it currently does not take any provider-specific API keys. I propose using the available provider API key as the api_key for litellm.
    - **Issue:** the issue #27826


efriis:
this will introduce a new bug where it'll pass the incorrect api key for any models after the first one that has an api key

chenzimin:
@efriis How about removing all existing `*_api_key` (since they are not used), and adding `api_key` (for litellm, only `api_key` is needed) to pass to litellm?

PR #27859, opened 2024-11-02
community: use api token from Replicate constructor for service access

fayvor:
**Description:**
- In the Replicate LLM, when a replicate_api_token is passed into the Replicate constructor, use it to create a private Client. Use that instead of the default Client to access the service for things like model lookups.
- Test that the api token is used by the model to access the service.

**Issue:**
https://github.com/langchain-ai/langchain/issues/27861

**Dependencies:**
N/A

**Twitter handle:**
N/A

### DIFF
--- a/libs/community/pyproject.toml
+++ b/libs/community/pyproject.toml
@@ -56,9 +56,9 @@ select = [ "E", "F", "I", "T201",]
 omit = [ "tests/*",]
 
 [tool.pytest.ini_options]
-addopts = "--strict-markers --strict-config --durations=5 --snapshot-warn-unused -vv"
+addopts = "--strict-markers --strict-config --durations=5 -vv"
 markers = [ "requires: mark tests as requiring a specific library", "scheduled: mark tests to run in scheduled testing", "compile: mark placeholder test used to compile integration tests without running them",]
-asyncio_mode = "auto"
+# asyncio_mode = "auto"
 filterwarnings = [ "ignore::langchain_core._api.beta_decorator.LangChainBetaWarning", "ignore::langchain_core._api.deprecation.LangChainDeprecationWarning:test", "ignore::langchain_core._api.deprecation.LangChainPendingDeprecationWarning:test",]
 
 [tool.poetry.group.test]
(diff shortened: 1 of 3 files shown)


libs/community/pyproject.toml
efriis, line 1: revert

efriis (commented):
why were the pytest options changed?

libs/community/pyproject.toml
```diff
@@ -56,9 +56,9 @@ select = [ "E", "F", "I", "T201",]
 omit = [ "tests/*",]
 
 [tool.pytest.ini_options]
-addopts = "--strict-markers --strict-config --durations=5 --snapshot-warn-unused -vv"
+addopts = "--strict-markers --strict-config --durations=5 -vv"
```
efriis, line 59:
```suggestion
addopts = "--strict-markers --strict-config --durations=5 --snapshot-warn-unused -vv"
```

libs/community/pyproject.toml
```diff
@@ -56,9 +56,9 @@ select = [ "E", "F", "I", "T201",]
...
 markers = [ "requires: mark tests as requiring a specific library", "scheduled: mark tests to run in scheduled testing", "compile: mark placeholder test used to compile integration tests without running them",]
-asyncio_mode = "auto"
+# asyncio_mode = "auto"
```
efriis, line 61:
```suggestion
asyncio_mode = "auto"
```

efriis (commented):
I believe this is unfortunately a breaking change for folks on the v0 replicate client. We have 2 options:

1. we can close this PR and would recommend implementing a dedicated integration package, such that breaking changes can be tolerated moving forward. See these [docs](https://python.langchain.com/docs/contributing/how_to/integrations/) 
2. we can update the docs to show that this community Replicate integration is meant to be used with `replicate==0.x`
3. we can add some if/else statements to the integration to tolerate both the v0 and v1 sdk

let me know what you prefer! I'd be strongly in favor of [1]!
5.

efriis:
let me know what you think! For [1] and [2], I think it makes sense to close this PR and reopen separately as a docs change. For [3] could iterate here. If you think [3] is the way forward and you have time to implement that compatibility before the holidays, let me know in the next day or two so we can keep this open! Otherwise we can close this and reopen when those changes have gone in.

fayvor:
Hi! Thanks for the review. I won't have time to implement [1] but I will add some if/else switches for [3] this week, and you can make the call as to whether to merge.

efriis:
great let's go for [3] here this week! If you don't get to it that's ok - I'll close and you're welcome to reopen a pr when it's in!

PR #27874, opened 2024-11-04
community: Ontotext GraphDB Graph & QA Chain use SPARQLWrapper instead of rdflib

nelly-hateva:
- **Description:** Ontotext GraphDB  & QA Chain use the SPARQLWrapper instead of rdflib:
  - the client can now use any authentication method supported by GraphDB, not just basic auth
  - some SPARQL queries are not valid, but the rdflib doesn't raise an exception and swallows the server error message, so instead of validating the queries on the client side with rdflib, we now validate them on the server and pass a good error message to the LLM
- **Issue:** N/A
- **Dependencies:** SPARQLWrapper
- **Twitter handle:** @OntotextGraphDB


efriis (commented):
so unfortunately this change is a breaking change for existing users, so this would have to be done in a new integration, which is actually the best path forward for this integration! Added bonus because it seems you work on it :)

we'd love to get an integration package out! Future PRs against langchain would just be {docs updates, as well as registering your package in `libs/packages.yml`, deprecating this community integration in favor of your integration package}

efriis:
going to close with the above recommendation - as mentioned feel free to leave quesitons about the process on that doc so others can see them!

PR #27885, opened 2024-11-04
community: correct return type of get_files_from_directory in github tool

lorenzb07:
### About:
- **Description:** the _get_files_from_directory_ method return a string, but it's used in other methods that expect a List[str]
- **Issue:** None
- **Dependencies:** None

This pull request import a new method _list_files_ with the old logic of _get_files_from_directory_, but it return a List[str] at the end.
The behavior of _ get_files_from_directory_ is not changed.


efriis:
hey! could you add a unit test that shows the failure case?

efriis (approved):

Issue #1559, opened 2023-03-09
{tool_name} is not a valid tool, try another one.


gcsun:
I encapsulated an agent into a tool ，and load it into another agent 。
when the second agent run, all the tools become invalid.
After calling any tool, the output is “*** is not a valid tool, try another one.

why! agent cannot be a tool?

Ashok93:
I got the same error while trying `gpt-3.5-turbo` model with Python REPL tool. But when I changed my model the tool was used.
I am using langchain version `0.0.107`

disarmyouwitha:
#'text-davinci-003'
Action: Requests
Action Input: https://splatoonwiki.org/wiki/Splatoon_3
Observation: <!DOCTYPE html>

#'gpt-3.5-turbo',
Action: Use Requests
Action Input: "https://splatoonwiki.org/wiki/.96_Gal#Splatoon_3"
Observation: Use Requests is not a valid tool, try another one.

It seems ChatGPT is responding, "Use Requests" instead of "Requests" like davinci.

alvinvogelzang:
I'm having the same issue when using `gpt-3.5-turbo`. This is what I receive:

```
Observation: Use the `value_counts()` method on the "Hoe voel je je?" column to count the number of occurrences of each value. is not a valid tool, try another one.
```

Anyone who has an idea on how to fix this?

bockaerts:
```
from langchain.agents import load_tools
from langchain.agents import initialize_agent
from langchain.chat_models import ChatOpenAI
from langchain.llms import OpenAI

llm = ChatOpenAI(temperature=0,model_name='gpt-3.5-turbo')
#llm = OpenAI(temperature=0)
tools = load_tools(["python_repl"], llm=llm)
agent = initialize_agent(tools, llm, agent="zero-shot-react-description", verbose=True)
agent.run("Can i print a number in python?")
```

```
Same problem hereTraceback (most recent call last):
  File "C:\Users\Name\Desktop\project\test.py", line 10, in <module>
    agent.run("Can i print a number in python?")
  File "C:\Users\Name\AppData\Roaming\Python\Python311\site-packages\langchain\chains\base.py", line 213, in run
    return self(args[0])[self.output_keys[0]]
           ^^^^^^^^^^^^^
  File "C:\Users\Name\AppData\Roaming\Python\Python311\site-packages\langchain\chains\base.py", line 116, in __call__
    raise e
[...]

PR #27889, opened 2024-11-04
fixed issue #1559

siddarthreddygsr:
resolved issue https://github.com/langchain-ai/langchain/issues/1559

### DIFF
--- a/libs/langchain/langchain/agents/agent.py
+++ b/libs/langchain/langchain/agents/agent.py
@@ -1430,6 +1430,8 @@ class AgentExecutor(Chain):
     ) -> AgentStep:
         if run_manager:
             run_manager.on_agent_action(agent_action, color="green")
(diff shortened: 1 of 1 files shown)


eyurtsev:
Hi @siddarthreddygsr, thank you for the patch, but it's better to not make this modification.

The main recommendation for users is to use langgraph for agents. You can use `create_react_agent` to create pre-built agents.  The agents leverage tool calling capabilities of chat models and will be more robust.

In general, no patch PRs against langchain and/or langchain-core will not be accepted without unit-tests. In this case, however, it's not clear that the patch is meaningful -- any code changes that focus on adjusting the prompt or making it slightly more robust can't guarantee significant performance increase. It's better to rely on chat models tool calling capabilities and/or provide users with the ability to adjust / manipulate the prompt / tool results as they see needed. Both of these are easy enough to achieve with langgraph.

Issue #26860, opened 2024-09-25
Choma similarity_search_with_score (and similar methods) don't populate the document.id property of the returned documents.


KrisTC:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

I want to get the document id from the documents I find with `similarity_search_with_score` but the id's aren't being set to the documents when they are created.

Here is my test code:

```python
def test_basic_operations_with_langchain():
    token = os.environ.get('CHROMA_TOKEN')
    client = chromadb.HttpClient(
        settings=Settings(chroma_client_auth_provider="chromadb.auth.token_authn.TokenAuthClientProvider",
                          chroma_client_auth_credentials=token))
    test_collection_name = "test_collection"

    try:
        client.delete_collection(test_collection_name)
    except:  # noqa: E722
        pass

    collection = client.get_or_create_collection(test_collection_name)
    assert collection.name == test_collection_name
    assert collection.count() == 0

    EMBEDDING_MODEL_NAME = "sentence-transformers/all-mpnet-base-v2"
    EMBEDDING_MODEL_CHUNK_SIZE = 384

    embedding_model = HuggingFaceEmbeddings(
        model_name=EMBEDDING_MODEL_NAME,
        # multi_process=True,   # I found this causes crashes and slowness
        model_kwargs={"device": "cuda" if torch.cuda.is_available() else "cpu"},
        encode_kwargs={
            "normalize_embeddings": True
        },  # set True for cosine similarity
    )
    vector_db = Chroma(
        embedding_function=embedding_model,
        client=client,
        collection_name=test_collection_name,
        collection_metadata={"hnsw:space": "cosine"},
    )
    # Make new uid
    id = str(uuid.uuid4())
    test_doc = Document("hello world")
    test_doc.id = id
    test_doc.metadata["test"] = "test"
    vector_db.add_documents([test_doc])
    assert collection.count() == 1

    docs = vector_db.similarity_search_with_score("hello world", 1)
    assert len(docs) == 1
    assert docs[0][0].page_content == "hello world"
    assert docs[0][0].id == id

    client.delete_collection(test_collection_name)
```

### Error Message and Stack Trace (if applicable)

Test output:
```
[...]

PR #27893, opened 2024-11-04
chroma[minor]: Add id attribute to Document returned by search results

kwei-zhang:
- Description
   We added the document id when documents are created in function ```similarity_search_with_score``` such that enable retrieve documents again from the vector store using the same id.

   Sample code: 

Before change
```
def _results_to_docs_and_scores(results: Any) -> List[Tuple[Document, float]]:
    return [
        # TODO: Chroma can do batch querying,
        # we shouldn't hard code to the 1st result
        (Document(page_content=result[0], metadata=result[1] or {}), result[2])
        for result in zip(
            results["documents"][0],
            results["metadatas"][0],
            results["distances"][0],
        )
    ]
```

After change
```
def _results_to_docs_and_scores(results: Any) -> List[Tuple[Document, float]]:
    return [
        # TODO: Chroma can do batch querying,
        # we shouldn't hard code to the 1st result
        (Document(id =result[0], page_content=result[1], metadata=result[2] or {}), result[3])
        for result in zip(
            results["ids"][0],
            results["documents"][0],
            results["metadatas"][0],
            results["distances"][0],
        )
    ]
```
 - Issue
   The PR is corresponding to [this issue](https://github.com/langchain-ai/langchain/issues/26860).

Test added:

[test_vectorstores.py](https://github.com/chaunguyenm/langchain/blob/chroma_add_id/libs/partners/chroma/tests/unit_tests/test_vectorstores.py)

- [✓] **Lint and test**: Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified. See contribution guidelines for more: https://python.langchain.com/docs/contributing/


KevinZJN:
duplicate pr

PR #27899, opened 2024-11-05
standard-tests: add tools standard tests

efriis:


libs/standard-tests/langchain_standard_tests/unit_tests/tools.py
```diff
@@ -0,0 +1,71 @@
...
+                actual = getattr(tool, k)
+                if isinstance(actual, SecretStr):
+                    actual = actual.get_secret_value()
+                assert actual == expected
+
+    def test_has_name(self, tool: BaseTool) -> None:
+        assert tool.name
+
+    def test_has_description(self, tool: BaseTool) -> None:
+        assert tool.description
```
baskaryan, line 59: dunno that we need to enforce this, descriptions aren't requried by openai

libs/standard-tests/langchain_standard_tests/integration_tests/tools.py
```diff
@@ -0,0 +1,43 @@
...
+        elif tool.response_format == "content_and_artifact":
+            # should be (content, artifact)
+            assert isinstance(result, tuple)
+            assert len(result) == 2
+            content, artifact = result
+
+            assert artifact  # artifact can be anything, but shouldn't be none
+
+        # check content is a valid ToolMessage content
+        assert isinstance(content, (str, list))
```
baskaryan, line 21: if self.tool_invoke_params_example isn't a ToolCall this isn't guaranteed. should probable wrap tool_invoke_params_example in a ToolCall (if needed) before passing to invoke

PR #27930, opened 2024-11-06
community[docs]: modify parameter for the LoRA adapter on the vllm page

e7217:
**Description:** 
This PR modifies the documentation regarding the configuration of the VLLM with the LoRA adapter. The updates aim to provide clear instructions for users on how to set up the LoRA adapter when using the VLLM.

- before
```python
VLLM(..., enable_lora=True)
```
- after
```python
VLLM(..., 
    vllm_kwargs={
        "enable_lora": True
    }
)
```
This change clarifies that users should use the vllm_kwargs to enable the LoRA adapter.


e7217:
@ccurme Could you please check this PR? Thank you.

ccurme (approved):

Issue #27485, opened 2024-10-19
Hallucination with Ollama vision models (Llava, Bakllava)


ahmedkooli:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
import base64
from io import BytesIO

from IPython.display import HTML, display
from PIL import Image
from langchain_ollama import OllamaLLM


def convert_to_base64(pil_image):
    """
    Convert PIL images to Base64 encoded strings

    :param pil_image: PIL image
    :return: Re-sized Base64 string
    """

    buffered = BytesIO()
    pil_image.save(buffered, format="JPEG")  # You can change the format if needed
    img_str = base64.b64encode(buffered.getvalue()).decode("utf-8")
    return img_str


def plt_img_base64(img_base64):
    """
    Display base64 encoded string as image

    :param img_base64:  Base64 string
    """
    # Create an HTML img tag with the base64 string as the source
    image_html = f'<img src="data:image/jpeg;base64,{img_base64}" />'
    # Display the image by rendering the HTML
    display(HTML(image_html))


file_path = "./test.jpg"
pil_image = Image.open(file_path)
image_b64 = convert_to_base64(pil_image)
plt_img_base64(image_b64)


llm = OllamaLLM(model="llava")
llm_with_image_context = llm.bind(images=[image_b64])
res = llm_with_image_context.invoke("Describe this image")
```

### Description

Hey, I was trying to use vision models with Ollama and I based myself on [this tutorial](https://python.langchain.com/docs/integrations/llms/ollama/#multi-modal) found on an official Langchain page.

Unfortunately, it seems that the model is not seeing the image that I'm passing as it is hallucinating every time and outputting a very different result at each run (for example: _**The image depicts a vibrant and colorful scene, featuring a tree with lush green leaves. The tree is adorned with various decorations, including what appears to be colorful baubles and lights [...]**_ for the image of food provided below).

I tried with multiple models (Llava and Bakllava), images and prompts, but the result is always wrong. 

Note that when I try using Ollama directly in the CLI it works:
```bash
ollama run llava "describe this image ./test.jpg"
```
results in "_**Added image './test.jpg'
[...]

PR #27933, opened 2024-11-06
partners: fix for hallucination of OllamaLLM when using image inputs

VedatAlpGoktepe:
- **Description:** Added an `images` property to the `OllamaLLM` class to take in a list of strings (base64 image data or image paths) to pass to the model together with the prompt.
- **Issue:** #27485 
- **Dependencies:** No dependencies required for the change.


efriis (commented):
hey! we have this as a feature of chat models. LLM abstraction is just str in, str completion out.

feel free to resubmit as a feature of ChatOllama instead!

VedatAlpGoktepe:
Hi @efriis ,

We decided to add this fix because the OllamaLLM documentation on Langchain included an example for multimodal input that used `.bind()` to add images to prompts, but the instructions provided in the documentation were not functioning as expected.

If the `.bind()` method is valid, would it make sense to submit a pull request that removes the alternative method of adding images via `kwargs` in `.invoke()`? Or is multimodal (image) input in LLMs entirely out of scope?

Issue #27909, opened 2024-11-05
Bug in langchain_community/callbacks/tracers/wandb.py:  'NoneType' object has no attribute 'items'


holiholiholiholi:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

python code
```python


import os
import wandb
from langchain.agents import AgentType, initialize_agent, load_tools
from langchain_openai import OpenAI
from contextvars import ContextVar
from typing import (
    Optional,
)
from langchain_core.tracers.context import register_configure_hook
from langchain_community.callbacks.tracers.wandb import WandbTracer


os.environ["LANGCHAIN_WANDB_TRACING"] = "true"
os.environ["WANDB_PROJECT"] = 'set your wand project'

wandb_tracing_callback_var: ContextVar[Optional[WandbTracer]] = ContextVar(
    "tracing_wandb_callback", default=None
)
register_configure_hook(
    wandb_tracing_callback_var, True, WandbTracer, "LANGCHAIN_WANDB_TRACING"
)

llm = OpenAI(temperature=0)
tools = load_tools(["llm-math"], llm=llm)

agent = initialize_agent(
    tools, llm, agent=AgentType.ZERO_SHOT_REACT_DESCRIPTION, verbose=False
)

agent.run("What is 2 raised to .123243 power?")

wandb.finish()

```

### Error Message and Stack Trace (if applicable)

`wandb: ERROR WARNING: Failed to serialize model: 'NoneType' object has no attribute 'items'`

### Description

I'm trying to use langchain wandb lib and there is a bug in the model: 
`langchain_community/callbacks/tracers/wandb.py` Line: 202 (version langchain-community@0.3.3)

```python

# def transform_run(run: Dict[str, Any]) -> Dict[str, Any]:
#         ...
#        transformed_dict = transform_serialized(run)

#        serialized = transformed_dict.pop("serialized")
       
        for k, v in serialized.items():
#            transformed_dict[k] = v
#       ...
```

where the object `serialized` is None!

### System Info

System Information
------------------
> OS:  Darwin
[...]

PR #27947, opened 2024-11-06
Community: In case of `serialized` is NoneType return empty Dict for `transform_run` of WanDB Tracer

keenborder786:
- **Description:**  Solves the following issue:
- **Issue:** #27909


keenborder786:
@eyurtsev

libs/community/langchain_community/callbacks/tracers/wandb.py
```diff
@@ -191,14 +191,16 @@ def transform_serialized(serialized: Dict[str, Any]) -> Dict[str, Any]:
         serialized = remove_exact_and_partial_keys(serialized)
         return serialized
 
-    def transform_run(run: Dict[str, Any]) -> Dict[str, Any]:
+    def transform_run(run: Dict[str, Any]) -> Dict:
```
eyurtsev, line 194:
```suggestion
    def transform_run(run: Dict[str, Any]) -> Dict[str, Any]:
```

libs/community/langchain_community/callbacks/tracers/wandb.py
```diff
@@ -191,14 +191,16 @@ def transform_serialized(serialized: Dict[str, Any]) -> Dict[str, Any]:
...
         """Transforms a run dictionary to be compatible with WBTraceTree.
         :param run: The run dictionary to transform.
         :return: The transformed run dictionary.
         """
         transformed_dict = transform_serialized(run)
 
         serialized = transformed_dict.pop("serialized")
+        if not serialized:
+            return {}
```
eyurtsev, line 203: This fix isn't correct as far as I can tell -- at least from reading this function code alone (i.e., 
without reading the contents of transform_serialized, there's no way to to determine that this code is written correctly).

Either missing doc-string to explain why empty dict OK or else need to handle case when name / kind are defined but serialized is not available

-----

cc @parambharat

eyurtsev:
@parambharat please review or advice on what you want to do w/ respect to users using the community wandb tracer

libs/community/langchain_community/callbacks/tracers/wandb.py
keenborder786: @eyurtsev I was not able to find any other way to make it work

libs/community/langchain_community/callbacks/tracers/wandb.py
keenborder786: @eyurtsev because we are returning an empty dict otherwise the linter raises an error

keenborder786:
@parambharat

keenborder786:
@parambharat

ccurme:
Closing but will re-open pending review from @parambharat.

PR #27952, opened 2024-11-07
templates,docs: leave templates in v0.2

efriis:
all template installs will now have to declare `--branch v0.2` to make clear they aren't compatible with langchain 0.3 (most have a pydantic v1 setup). e.g.

```
langchain-cli app add pirate-speak --branch v0.2
```


ccurme (approved):
One thing cookbooks do cover that is useful to some is multi-modal RAG. We might consider adding a how-to guide on that.

PR #27968, opened 2024-11-07
langchain_chroma: fixed integration tests

epinzur:
Description:
* I'm planning to add `Document.id` support to the Chroma VectorStore, but first I wanted to make sure all the integration tests were passing first.  They weren't.  This PR fixes the broken tests.  
* I found 2 issues:
  * This change (from a year ago, exactly :) ) for supporting multi-modal embeddings: https://docs.trychroma.com/deployment/migration#migration-to-0.4.16---november-7,-2023
  * This change https://github.com/langchain-ai/langchain/pull/27827 due to an update in the chroma client.
  
Also ran `format` and `lint` on the changes.

Note: I am not a member of the Chroma team.

### DIFF
--- a/libs/partners/chroma/tests/integration_tests/test_vectorstores.py
+++ b/libs/partners/chroma/tests/integration_tests/test_vectorstores.py
@@ -1,12 +1,16 @@
 """Test Chroma functionality."""
 
 import uuid
-from typing import Generator
+from typing import (
+    Generator,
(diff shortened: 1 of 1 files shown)


efriis (commented):
not compatible with chroma 0.4. Is there a way to retain compatibility, or worth removing that compatibility?

PR #27972, opened 2024-11-07
docs: new stack diagram

efriis:


README.md
efriis, line 1: this adds dark mode support

libs/community/README.md
efriis, line 1: didn't add dark mode here because pypi won't render it

PR #27983, opened 2024-11-08
openai[patch]: default to invoke on o1 stream()

baskaryan:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -999,6 +963,28 @@ def get_num_tokens_from_messages(self, messages: List[BaseMessage]) -> int:
...
+        async_api: bool,
+        run_manager: Optional[
+            Union[CallbackManagerForLLMRun, AsyncCallbackManagerForLLMRun]
+        ] = None,
+        response_format: Optional[Union[dict, type]] = None,
+        **kwargs: Any,
+    ) -> bool:
+        if isinstance(response_format, type) and is_basemodel_subclass(response_format):
+            # TODO: Add support for streaming with Pydantic response_format.
+            warnings.warn("Streaming with Pydantic response_format not yet supported.")
```
ccurme, line 978: should we also emit a warning in case of `o1`, for clarity about why results aren't streamed?

ccurme (approved):

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: maybe. just seems like model is something you're more likely to let end users configure (eg the way langsmith playground does) in which case warnings would be annoying and unhelpful

Issue #27264, opened 2024-10-11
SqliteCache fails with ChatOpenAI.with_structured_output(method="json_schema")


bjschnei:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python

from langchain.globals import set_llm_cache
from langchain_community.cache import SQLiteCache
from langchain_core.messages import SystemMessage
from langchain_openai import ChatOpenAI
from pydantic import BaseModel

from dotenv import load_dotenv

_ = load_dotenv()

class Limerick(BaseModel):
    limerick: str

def main():
    set_llm_cache(SQLiteCache())
    llm = ChatOpenAI(model_name="gpt-4o-2024-08-06")
    structured_llm = llm.with_structured_output(Limerick, method="json_schema")
    for _ in range(2):
        result = structured_llm.invoke([SystemMessage(content="Write a limerick")])
        print(result)

if __name__ == "__main__":
    main()
```

### Error Message and Stack Trace (if applicable)

  File "/usr/local/lib/python3.12/site-packages/langchain_core/runnables/base.py", line 3022, in invoke
    input = context.run(step.invoke, input, config, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/langchain_core/runnables/base.py", line 5354, in invoke
    return self.bound.invoke(
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 286, in invoke
    self.generate_prompt(
  File "/usr/local/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 786, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 643, in generate
    raise e
[...]

PR #27989, opened 2024-11-08
Core: Handle Failed Cache Reads For Non-Serializable Objects

FarhanChowdhury248:
This PR fixes the issue mentioned in #27264.

### Bug

Many caching implementations for chat model results (as defined in `langchain_core/caches.py`) require cached objects to be `Serializable` in order to properly perform the cache read and write. In cases where they are not, the lookup may fail when the cache attempts to de-serialize the cached data into a list of `ChatGeneration` objects. On such failure, many caches in `langchain_core\caches.py` fail over to returning a list of `Generation` objects that contain the serialized data strings.

When these are read by `BaseChatModel` in `_generate_with_cache` or `_agenerate_with_cache`, the cached result is fed into a `ChatGeneration` object, which does not accept `Generation` objects, leading to the error.

### Fix

Add a check in `_generate_with_cache` and `_agenerate_with_cache` to verify the cached results are of type `ChatGeneration[]`. If not, skip the lookup and treat the case as an uncached one.

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -820,7 +820,11 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
                 # of Generation objects in this case, which is invalid
                 # for ChatResult, so we check and treat this as an
                 # uncached case.
-                if isinstance(cache_val, list) and len(cache_val) > 0 and isinstance(cache_val[0], ChatGeneration):
+                if (
+                    isinstance(cache_val, list)
+                    and len(cache_val) > 0
+                    and isinstance(cache_val[0], ChatGeneration)
+                ):
                     return ChatResult(generations=cache_val)
             elif self.cache is None:
                 pass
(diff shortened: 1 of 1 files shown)


FarhanChowdhury248:
Where do tests for `BaseChatModel` go? I was hoping to add the case defined in #27264 as a test case for this, but could not find any tests that test `BaseChatModel` directly.

FarhanChowdhury248:
@jacoblee93 Do you mind having a look at this?

FarhanChowdhury248:
@baskaryan Can someone have a look at this?

efriis:
hey there! Sorry for the delay on this one. Could you add some tests either to this, or even better to some of the integration packages where you're seeing this failure mode with InMemoryCache? I don't fully understand why this needs to be implemented on BaseChatModel, instead of fixing on a specific cache that has a bug.

FarhanChowdhury248:
@efriis Thanks for taking the time to look into it. The fix is on BaseChatModel because the root cause of the issue has to do with how the cache reads are converted into a ChatResult. Any cache implementation that tries to read an unsupported object from its serialized version will run into this error.

Do you know where the ideal place to add tests for this? I looked into the test directories but found no designated directory/file for tests that focus on (de)serialization.

efriis:
if this is the fix that's necessary, would recommend adding to `libs/core/tests/unit_tests/fake/test_fake_chat_model.py` or even a new one of `libs/core/tests/unit_tests/fake/test_fake_chat_model_with_cache.py` using the InMemoryCache

this will help me understand what the issue is

efriis:
@FarhanChowdhury248 let me know if you're interested in writing some of those unit tests! Otherwise I'll close later this week (and folks are always welcome to reopen once those are up demonstrating the issue)

FarhanChowdhury248:
Hey @efriis. I plan on getting on this after tomorrow, when my last exam is done. Sorry I haven't been able to keep up with this.

efriis:
closing as stale and feel free to reopen if you get around to it!

eschbach-mak:
This issue still persists for me

ccurme:
This should resolve: https://github.com/langchain-ai/langchain/pull/29940

PR #27991, opened 2024-11-08
core[patch]: support numpy 2

ccurme:


baskaryan (approved):

eyurtsev (approved):

efriis (approved):

ghinch:
Hi, I can see this patch updated the numpy requirement for `langchain-core` to now be inclusive of the 2.x.x versions, but the `langchain` and `langchain-community` modules were left at <2. Do new issues/PRs need to be set to update those to match?

ulgens:
+1 for @ghinch 's comment. This PR doesn't exactly provide numpy>=2 support as is. How should we proceed for the proper support?

PR #28003, opened 2024-11-09
community: fix duplicate content

Renzo-vS:
Thank you for reading my first PR!

**Description:**
Deduplicate content in AzureSearch vectorstore.
Currently, by default, the content of the retrieval is placed both in metadata and page_content of a Document.
This PR removes the content from metadata, and leaves it in page_content.

**Issue:**:
Previously, the content was popped from result before metadata was populated.
In #25828 , the order was changed which leads to a response with duplicated content.
This was not the intention of that PR and seems undesirable.

Looking forward to seeing my contribution in the next version!

Cheers, 
Renzo


eyurtsev:
If you're doing a bug fix, could you include a test showing what's fixed?

efriis (approved):

Issue #27346, opened 2024-10-14
ChatPromtpTemplate doesn't accept PDF data as bytes


alandema:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
pdf_data = base64.b64encode(httpx.get("https://dagrs.berkeley.edu/sites/default/files/2020-01/sample.pdf").content).decode("utf-8")

prompt = ChatPromptTemplate([
    ("system", "You are a helpful assistant "),
    ("human", [
        {"type": "media", "mime_type": "application/pdf", "data": pdf_data},
        {"type": "text", "text": "{user_input}"}
    ])
])
```

### Error Message and Stack Trace (if applicable)

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], [line 12](vscode-notebook-cell:?execution_count=4&line=12)
      [1](vscode-notebook-cell:?execution_count=4&line=1) model = ChatGoogleGenerativeAI(
      [2](vscode-notebook-cell:?execution_count=4&line=2)     model="gemini-1.5-flash",
      [3](vscode-notebook-cell:?execution_count=4&line=3)     temperature=0.1,
   (...)
      [6](vscode-notebook-cell:?execution_count=4&line=6)     max_retries=2
      [7](vscode-notebook-cell:?execution_count=4&line=7) )
      [9](vscode-notebook-cell:?execution_count=4&line=9) pdf_data = base64.b64encode(
     [10](vscode-notebook-cell:?execution_count=4&line=10)     httpx.get("http://www.dagrs.berkley.edu/sites/default/files/2020-01/sample.pdf").content).decode("utf-8")
---> [12](vscode-notebook-cell:?execution_count=4&line=12) prompt = ChatPromptTemplate([
     [13](vscode-notebook-cell:?execution_count=4&line=13)     ("system", "You are a helpful assistant "),
     [14](vscode-notebook-cell:?execution_count=4&line=14)     ("human", [
     [15](vscode-notebook-cell:?execution_count=4&line=15)         {"type": "media", "mime_type": "application/pdf", "data": pdf_data},
[...]

PR #28011, opened 2024-11-09
core: fix ChatPromptTemplate doesn't accept PDF data as bytes

rkwan05:
**Description:** 
- Functionality (_chat.py_): added if statement in ChatPromptTemplate's `from_template()` that catches data in mime_type: {pdf_data} format
    - extracts text from pdf and inserts it into the prompt
- Unit Test (_test_chat.py_): added test that passes in pdf data to ChatPromptTemplate

**Issue:** #27346
**Dependencies:** 
- base64: Decodes pdf data in `chat.py > extract_pdf_text()` and encodes pdf data in `text_chat.py > test_create_pdf_chat_prompt()`
- io: Creates an in-memory binary stream to handle pdf data in `chat.py > extract_pdf_text()`
- pypdf: To import PDFReader to read the pdf file in `chat.py > extract_pdf_text()`

**Lint and test:**

Ran make format, make lint, and make test from the root of `libs\core`. All checks passed successfully.


libs/core/tests/unit_tests/prompts/test_chat.py
```diff
@@ -36,6 +36,36 @@
...
+    with open(file_path, "rb") as file:
+        file_data = file.read()
+
+    pdf_data = base64.b64encode(file_data).decode("utf-8")
+    prompt = ChatPromptTemplate(
+        [
+            (
+                "human",
+                [
+                    {"type": "media", "mime_type": "application/pdf", "data": pdf_data},
```
efriis, line 55: which chat model supports this?

rkwan05: Hi Efriis, I just wanted to make sure that we understood the question correctly! If you’re asking which chat models this implementation would be compatible with, it should work with all chat models because it is sending the pdf data as text. But please let me know if you were looking for something else!

efriis:
closing because this isn't desirable in the prompt template. in general, the right process here would be parse the pdfs using something like [this](https://python.langchain.com/docs/how_to/document_loader_pdf/), and plumb that text into your prompt

this implementation would import PyPDF parsing for anyone trying to use a prompt template, which isn't mergable

PR #28024, opened 2024-11-11
feat(core): Allow nested prompt templates

dmenini:
This PR allows to use nested prompt templates. 
* `partial` now checks if the partial variable is a `PromptTemplate` and it will propagate partial variables to it
* `format` will first format the nested prompts and the root prompt

In my application, we allow users to write part of prompts (i.e. specific instructions), which may contain input variables. After this change, it will be possible for us to instantiate those instructions as `PromptTemplate`, and pass them to the main prompts as partial variables. 

N.B. I know that there is PipelinePromptTemplate that does something similar, but it doesn't support partial prompts as of now, and imo it should not be needed if the `PromptTemplate` itself can handle nested prompts.

Thanks for your feedback!

@hwchase17 @baskaryan

### DIFF
--- a/libs/core/langchain_core/prompts/prompt.py
+++ b/libs/core/langchain_core/prompts/prompt.py
@@ -8,6 +8,7 @@ from typing import Any, Optional, Union
 
 from pydantic import BaseModel, model_validator
 
+from langchain_core.prompts import BasePromptTemplate
 from langchain_core.prompts.string import (
     DEFAULT_FORMATTER_MAPPING,
     PromptTemplateFormat,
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/prompts/prompt.py
```diff
@@ -104,12 +105,24 @@ def pre_init_validation(cls, values: dict) -> Any:
...
+            nested_partial_vars = {
+                key
+                for partial_var in values["partial_variables"].values()
+                if isinstance(partial_var, BasePromptTemplate)
+                for key in partial_var.partial_variables
+            }
+
+            # Filter template variables based on
+            # partial_variables and nested_partial_vars
             values["input_variables"] = [
```
dmenini, line 119: This is the only point I found a bit counterintuitive. Imo if the user passes `input_variables` to the `PromptTemplate` we should not overwrite them with values from the template. This complicates the logic as I had to make sure that the nested partial variables are correctly excluded, to not break backward compatibility.

ccurme (commented):
Thanks for this. The changes look largely OK (e.g., non-breaking) but IMO it's unclear that the added complexity here is worth it. This is ultimately somewhat niche functionality and we don't have any documentation of what this buys over doing the formatting yourself, e.g.,
```python
template = "This is a {foo} test."
template_nested = "{bar}"

new_prompt = PromptTemplate(
    template=template.format(foo=template_nested)
).partial(bar="bar")
```

ccurme:
Closing due to inactivity.

PR #28032, opened 2024-11-11
partners: add xAI chat integration

vbarda:


libs/partners/xai/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Installation and Setup
+
+- Install the LangChain partner package
+
+```bash
+pip install -U langchain-xai
+```
+
+- Get your xAI api key from the [xAI Dashboard](https://x.ai/api-keys) and set it as an environment variable (`XAI_API_KEY`)
```
jimmyfraiture, line 13: Can you replace with https://console.x.ai

libs/partners/xai/README.md
```diff
@@ -0,0 +1,17 @@
+# langchain-xai
+
+This package contains the LangChain integrations for [xAI](https://x.ai/) through their [APIs](https://docs.x.ai/).
```
jimmyfraiture, line 3: https://x.ai/ -> https://console.x.ai

libs/partners/xai/README.md
vbarda: yup!

libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -0,0 +1,349 @@
...
+    """  # noqa: E501
+
+    @property
+    def lc_secrets(self) -> Dict[str, str]:
+        """A map of constructor argument names to secret ids.
+
+        For example,
+            {"xai_api_key": "XAI_API_KEY"}
+        """
+        return {"xai_api_key": "XAI_API_KEY"}
```
baskaryan, line 262: is this stored as xai_api_key or openai_api_key as currently implemented? if mypy won't yell about it we may want to do something like
```python
openai_api_key: Optional[SecretStr] = None
xai_api_key: SecretStr = Field(..., alias="api_key")

@model_validator(before=True):
def validate_api_key(values: dict) -> dict:
  if values.get("openai_api_key") and not (values.get("xai_api_key") or values.get("api_key")):
    values["xai_api_key"] = values.pop("openai_api_key")
  return values
```

libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -0,0 +1,349 @@
...
+    Automatically read from env variable `XAI_API_KEY` if not provided.
+    """
+    xai_api_base: str = Field(
+        default_factory=from_env("XAI_API_BASE", default="https://api.x.ai/v1/"),
+        alias="base_url",
+    )
+
+    model_config = ConfigDict(
+        populate_by_name=True,
+    )
```
baskaryan, line 312: nit: could we put these above any methods/properties

libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -0,0 +1,349 @@
...
+
+    def _get_ls_params(
+        self, stop: Optional[List[str]] = None, **kwargs: Any
+    ) -> LangSmithParams:
+        """Get the parameters used to invoke the model."""
+        params = super()._get_ls_params(stop=stop, **kwargs)
+        params["ls_provider"] = "xai"
+        return params
+
+    model_name: str = Field(default="grok-beta", alias="model")
```
baskaryan, line 295: would prefer to not have a default model, this will go out of date pretty soon and changing it is breaking

libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -0,0 +1,349 @@
...
+    """Model name to use."""
+    xai_api_key: Optional[SecretStr] = Field(
```
baskaryan, line 297: may need to update openai_api_key param to not accept alias
```python
openai_api_key: Optional[SecretStr] = None
```

libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -0,0 +1,349 @@
...
+        alias="api_key",
+        default_factory=secret_from_env("XAI_API_KEY", default=None),
+    )
+    """xAI API key.
+
+    Automatically read from env variable `XAI_API_KEY` if not provided.
+    """
+    xai_api_base: str = Field(
```
baskaryan, line 305: same with openai_api_base and alias

libs/partners/xai/tests/integration_tests/test_chat_models.py
```diff
@@ -0,0 +1,128 @@
+import pytest  # type: ignore[import-not-found]
+from langchain_core.messages import BaseMessage, HumanMessage, SystemMessage
+
+from langchain_xai import ChatXAI
+
+
+def test_chat_xai_model() -> None:
```
baskaryan, line 7: unit test

libs/partners/xai/tests/integration_tests/test_chat_models.py
baskaryan, line 1: think all of these should either be covered by standard integration tests or turned into unit tests

libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -0,0 +1,351 @@
...
+        default_factory=from_env("XAI_API_BASE", default="https://api.x.ai/v1/"),
```
baskaryan, line 266: nit: imo doesn't need to be env configurable

baskaryan: can always introduce later if it'd be useful, but would prefer to minimize num of env vars we have so there's fewer backwards compatibility considerations in future

libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -0,0 +1,351 @@
...
+
+        For example,
+            {"xai_api_key": "XAI_API_KEY"}
+        """
+        return {"xai_api_key": "XAI_API_KEY"}
+
+    @classmethod
+    def get_lc_namespace(cls) -> List[str]:
+        """Get the namespace of the langchain object."""
+        return ["langchain", "chat_models", "xai"]
```
baskaryan, line 288: think we're going with actual import paths from here on out
```suggestion
        return ["langchain_xai", "chat_models"]
```

baskaryan: should also add langchain_xai to https://github.com/langchain-ai/langchain/blob/e29073669616b3802d5208a79ae510c40ce541e5/libs/core/langchain_core/load/load.py#L15

baskaryan: also think we need to update is_lc_serializable to return True

baskaryan:
lgtm aside from serialization comments (note will need to rerun unit tests with snapshot update `make test_watch` to generate snapshots)

baskaryan (approved):

MrForExample:
Support for grok-vision-beta?

PR #28076, opened 2024-11-13
docs: langchain-ai21 update ai21 docs

miri-bar:

Update docs to match latest langchain-ai21 release.


docs/docs/integrations/llms/ai21.ipynb
```diff
@@ -1,221 +0,0 @@
-{
```
eyurtsev, line 1: @efriis do we need to set up redirects here?

efriis (commented):
@miri-bar can we update either the provider page or the individual pages (if ok keeping them) with a note that there used to be embeddings/llms, and the endpoints have been turned off?

I can set up some redirects from the deleted pages to make sure the links don't break

miri-bar:
@efriis Sure, no objection from my end to leave them and add a redirect.
I've added to the provider page a section for the deprecated endpoints.
LMK what you think

PR #28097, opened 2024-11-13
community: Remove all other keys in ChatLiteLLM and add api_key

chenzimin:

- **PR title**: "community: Remove all other keys in ChatLiteLLM and add api_key"

- **PR message**: Currently, no api_key are passed to LiteLLM, and LiteLLM only takes on api_key parameter. Therefore I removed all current `*_api_key` attributes (They are not used), and added `api_key` that is passed to ChatLiteLLM.
  - Should fix issue #27826

### DIFF
--- a/libs/community/langchain_community/chat_models/litellm.py
+++ b/libs/community/langchain_community/chat_models/litellm.py
@@ -54,7 +54,7 @@ from langchain_core.outputs import (
 )
 from langchain_core.runnables import Runnable
 from langchain_core.tools import BaseTool
-from langchain_core.utils import pre_init
+from langchain_core.utils import get_from_dict_or_env, pre_init
 from langchain_core.utils.function_calling import convert_to_openai_tool
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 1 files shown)


efriis (commented):
hey there! is this a recent change in litellm that merits this change / is it possible this will break something for current users of this + an old version of lite llm?

chenzimin:
I am not sure, in the current litellm documentation it states that you can provide different provider keys by (https://docs.litellm.ai/docs/set_keys#litellm-variables):

```
litellm.openai_key = "sk-OpenAIKey"
response = litellm.completion(messages=messages, model="gpt-3.5-turbo")

# anthropic call
litellm.anthropic_key = "sk-AnthropicKey"
response = litellm.completion(messages=messages, model="claude-2")
```

Even from the initial version of litellm (back in Nov 2023), they are called `openai_key`, not `openai_api_key`, https://github.com/BerriAI/litellm/blob/1b4aadbb253043378fd58205dc66f4fcc9339976/litellm/__init__.py#L8

But `api_key` is checked for all providers.

Plus I have not found where `openai_api_key`, `openai_api_key` ... are used in the current ChatliteLLM.

ccurme (approved):
@chenzimin from what I can tell, you're correct that the named API keys are not used and it is defensible to delete them.

I think it's preferable to just fix them in this case instead of deleting the fields, in case users are setting them somewhere.

I've updated the PR to support the `api_key` field in addition to the named keys. Let me know if you see any issues with this.

Thanks!

PR #28144, opened 2024-11-15
docs:Moved AWS tab ahead in the list as per integration telemetry

3coins:
Moving ahead per integration telemetry


efriis (approved):
should we also update to use `ChatBedrockConverse(model="...")` instead, rather than the shim from `ChatBedrock(model="...", use_converse=True)`?

3coins:
> should we also update to use ChatBedrockConverse(model="...") instead

We want to move to one `ChatBedrock` class, so users don't have to rename in their code. We will likely end up making ChatBedrockConverse --> ChatBedrock in the next minor release.

PR #28151, opened 2024-11-16
community: Apache AGE wrapper additional edge cases.

GMartin-dev:
Description: 
Current AGEGraph() implementation does some custom wrapping for graph queries. The method here is _wrap_query() as it parse the field from the original query to add some SQL context to it.
This improves the current parsing logic to cover additional edge cases that are added to the test coverage, basically if any Node property name or value has the "return" literal in it will break the graph / SQL query.
We discovered this while dealing with real world datasets, is not an uncommon scenario and I think it needs to be covered.


GMartin-dev:
@eyurtsev Just a ping just in case this PR fell through the cracks

ccurme (approved):

PR #28156, opened 2024-11-16
docs: Added missing installation for required packages in tutorial notebooks

amrzv:
**Description:** some of the required packages are missing in the installation cell in tutorial notebooks. So I added required packages to installation cell or created latter one if it was not presented in the notebook at all.

Tested in colab: "Kernel" -> "Run all cells". All the notebooks under `docs/tutorials` run as expected without `ModuleNotFoundError` error.

### DIFF
--- a/docs/docs/tutorials/chatbot.ipynb
+++ b/docs/docs/tutorials/chatbot.ipynb
@@ -124,15 +124,6 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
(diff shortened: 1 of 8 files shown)


ccurme (approved):
For most of these there are pip and conda commands for installing the libraries formatted in markdown in a previous cell.

See this [rendered docs page](https://python.langchain.com/docs/tutorials/chatbot/) as an example. Also, when you select a chat model from the tabs it provides installation instructions. They are similarly formatted in markdown which is why they don't get executed. So we should typically not hardcode `langchain-openai` into cells.

PR #28167, opened 2024-11-17
community: Add support for cross-region inference profile IDs in Bedrock Anthropic Claude token cost calculation

lakinduboteju:
This change modifies the token cost calculation logic to support cross-region inference profile IDs for Anthropic Claude models. Instead of explicitly listing all regional variants of new inference profile IDs in the cost dictionaries, the code now extracts a base model ID from the input model ID (or inference profile ID), making it more maintainable and automatically supporting new regional variants.

These inference profile IDs follow the format: `<region>.<vendor>.<model-name>` (e.g., `us.anthropic.claude-3-haiku-xxx`, `eu.anthropic.claude-3-sonnet-xxx`).

Cross-region inference profiles are system-defined identifiers that enable distributing model inference requests across multiple AWS regions. They help manage unplanned traffic bursts and enhance resilience during peak demands without additional routing costs.

References for Amazon Bedrock's cross-region inference profiles:-
- https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html
- https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html


libs/community/langchain_community/callbacks/bedrock_anthropic_callback.py
```diff
@@ -28,15 +28,25 @@
...
     prompt_tokens: int, completion_tokens: int, model_id: Union[str, None]
 ) -> float:
+    if model_id:
+        # The model ID can be a cross-region (system-defined) inference profile ID,
+        # which has a prefix indicating the region (e.g., 'us', 'eu') but
+        # shares the same token costs as the "base model".
+        # By extracting the "base model ID", by taking the last two segments
+        # of the model ID, we can map cross-region inference profile IDs to
+        # their corresponding cost entries.
+        base_model_id = model_id.split(".")[-2] + "." + model_id.split(".")[-1]
```
efriis, line 38: this seems rather brittle. @3coins is there a better way of going about this? Also any interest in moving this callback to the langchain-aws package?

3coins: Agree that this is not ideal, but there is no better way at the moment to parse these values. One option could be look for each pattern of model id supported by invoke, and based on the pattern do something specific, but this list is rather large. Take a look at the `modelId` field in the invoke API documentation.
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html

*Pattern: ^(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:(([0-9]{12}:custom-model/[a-z0-9-]{1,63}[.]{1}[a-z0-9-]{1,63}/[a-z0-9]{12})|(:foundation-model/[a-z0-9-]{1,63}[.]{1}[a-z0-9-]{1,63}([.:]?[a-z0-9-]{1,63}))|([0-9]{12}:imported-model/[a-z0-9]{12})|([0-9]{12}:provisioned-model/[a-z0-9]{12})|([0-9]{12}:(inference-profile|application-inference-profile)/[a-zA-Z0-9-:.]+)))|([a-z0-9-]{1,63}[.]{1}[a-z0-9-]{1,63}([.:]?[a-z0-9-]{1,63}))|(([0-9a-zA-Z][_-]?)+)|([a-zA-Z0-9-:.]+)$|(^(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:prompt/[0-9a-zA-Z]{10}(?::[0-9]{1,5})?))$|(^arn:aws:sagemaker:[a-z0-9-]+:[0-9]{12}:endpoint/[a-zA-Z0-9-]+$)|(^arn:aws(-[^:]+)?:bedrock:([0-9a-z-]{1,20}):([0-9]{12}):default-prompt-router/[a-zA-Z0-9-:.]+$)$*

Would support moving this to langchain-aws where we can support these requests better.

efriis: sounds good will merge this for now, and let me know when we can deprecate the community callback!

efriis (approved):

PR #28170, opened 2024-11-17
core: fix support for Pydantic Field information parameters in convert_to_openai_function

### LINKED ISSUE #26819: When using `convert_to_openai_function`, if a type hint of field is an Enum, its `description` will be lost.
### Checked other resources

### Example Code

```py
from enum import StrEnum

class WithDesciptionModel(BaseModel):
    foo: str = Field(..., description="This is a foo field")


class MyEnum(StrEnum):
    NAME = "name"
    AGE = "age"


class WithoutDesciptionModel(BaseModel):
    foo: MyEnum = Field(..., description="This is a foo field")


# The description of foo will be kept.
print(convert_to_openai_function(WithDesciptionModel))
# The description of foo will be lost.
print(convert_to_openai_function(WithoutDesciptionModel))


# print(WithDesciptionModel.model_json_schema())
# print(WithoutDesciptionModel.model_json_schema())
```

```py
{'name': 'WithDesciptionModel', 'description': '', 'parameters': {'properties': {'foo': {'description': 'This is a foo field', 'type': 'string'}}, 'required': ['foo'], 'type': 'object'}}
{'name': 'WithoutDesciptionModel', 'description': '', 'parameters': {'properties': {'foo': {'enum': ['name', 'age'], 'type': 'string'}}, 'required': ['foo'], 'type': 'object'}}
```

### Description

If the type hint in a `BaseModel` includes an `Enum`, its description will be lost.

### System Info

langchain-core==0.3.5
pydantic==2.9.2

python-version==3.12.6

ArmandoRJr:
## Description
This is being submitted for a school project as part of a team of 5. Team members are @wilso929, @ArmandoRJr, @LoudLiang, @SheeperGit, @JuefeiLu. 

Initial context is in the following issue: https://github.com/langchain-ai/langchain/issues/26819

This pull request fixes support for Pydantic Field information parameters (e.g. `description`) when using custom/non-primitive types for `convert_to_openai_function` (e.g. Python's `Enum`). 

## Issues
- Closes https://github.com/langchain-ai/langchain/issues/26819

## Dependencies
No extra dependencies needed for this bug fix.

## Git commits
We'll gladly squash our current commits if necessary. Please let us know if this is adequate, or if this will be done in the merging process anyway.


libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -459,6 +461,128 @@ def my_function(arg1: Nested) -> None:
     assert actual == expected
 
 
+@pytest.mark.skipif(
+    tuple(map(int, PYDANTICVERSION.split("."))) < (2, 9, 0),
```
efriis, line 465: can we switch this condition to use `packaging.Version` instead?

efriis (commented):
in general a feature like this should pass pydantic tests in all pydantic versions we support (2.5+). Why don't these tests pass below 2.9?

libs/core/tests/unit_tests/utils/test_function_calling.py
efriis: (ignore - this should not be a condition)

efriis:
will close this for now and you're welcome to reopen if you find a way to make this pass for all supported pydantic versions! If it's only partial support, I'm inclined to keep this off master until we have a solution that passes across all. Issues that only come up in certain pydantic versions are some of the most difficult to debug.

If you pick it back up, happy to discuss (sorry for the delay this past month), and there are 2 strategies would recommend looking into

1. (best) find a way of looking through the object that is consistent across pydantic versions
2. (ok) write some if/else statements in the implementation depending on pydantic version that make the test pass equivalently across all versions

Issue #24856, opened 2024-07-31
Standardize Embeddings Docs


efriis:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

## Issue

To make our Embeddings integrations as easy to use as possible we need to make sure the docs for them are thorough and standardized. There are two parts to this: updating the embeddings docstrings and updating the actual integration docs.

This needs to be done for each embeddings integration, ideally with one PR per embedding provider.

Related to broader issues #21983 and #22005.

## Docstrings

Each Embeddings class docstring should have the sections shown in the [Appendix](#appendix) below. The sections should have input and output code blocks when relevant.

To build a preview of the API docs for the package you're working on run (from root of repo):

```bash
make api_docs_clean; make api_docs_quick_preview API_PKG=openai
```

where `API_PKG=` should be the parent directory that houses the edited package (e.g. community, openai, anthropic, huggingface, together, mistralai, groq, fireworks, etc.). This should be quite fast for all the partner packages.

## Doc pages

Each Embeddings [docs page](https://python.langchain.com/v0.2/docs/integrations/text_embedding/) should follow [this template](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/text_embedding.ipynb). 

- [ ] TODO(Erick): populate a complete example

You can use the `langchain-cli` to quickly get started with a new chat model integration docs page (run from root of repo):

```bash
poetry run pip install -e libs/cli
poetry run langchain-cli integration create-doc --name "foo-bar" --name-class FooBar --component-type Embeddings --destination-dir ./docs/docs/integrations/text_embedding/
```

where `--name` is the integration package name without the "langchain-" prefix and `--name-class` is the class name without the "Embedding" prefix. This will create a template doc with some autopopulated fields at docs/docs/integrations/text_embedding/foo_bar.ipynb.

To build a preview of the docs you can run (from root):

```bash
make docs_clean
make docs_build
cd docs/build/output-new
yarn
yarn start
```

## Appendix

Expected sections for the Embedding class docstring.

`__package_name___`: This is the full name of the package (e.g., `langchain-anthropic`)
`__ModuleName__` : This is the CamelCase name of the partner (e.g., `Anthropic`)
`__MODULE_NAME__`: [SCREAMING_SNAKE_CASE](https://en.wikipedia.org/wiki/Snake_case) name of the partner (e.g., ANTHROPIC)

Y

```python
[...]

PR #28172, opened 2024-11-18
docs: Standardize DashScopeEmbeddings docstring

maang-h:
- **Description:** Standardize `DashScopeEmbeddings` 
  -  standardize docsting, the issue #24856
  - standardized model init arg names, the issue #20085


efriis:
I don't think dashscope is the same as tongyi?

PR #28197, opened 2024-11-18
openai: release 0.2.9, o1 streaming

efriis:


libs/partners/openai/poetry.lock
ulgens, line 1: Wouldn't it be better to compile these changes with `--no-update`?

Issue #28176, opened 2024-11-18
with_structured_output not working with OpenAI ChatLiteLLM


chenzimin:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code

```
from langchain_community.chat_models import ChatLiteLLM
from langchain_core.messages import HumanMessage
from pydantic import BaseModel, Field
import os

os.environ["OPENAI_API_KEY"] = "xxx"

class Joke(BaseModel):
    setup: str = Field(description="The setup of the joke")
    punchline: str = Field(description="The punchline to the joke")

model = ChatLiteLLM(model="gpt-4o")
structured_llm = model.with_structured_output(Joke)
structured_llm.invoke("Tell me a joke about cats")
```

will raise

```
BadRequestError: litellm.BadRequestError: OpenAIException - Error code: 400 - {'error': {'message': "Invalid value: 'any'. Supported values are: 'none', 'auto', and 'required'.", 'type': 'invalid_request_error', 'param': 'tool_choice', 'code': 'invalid_value'}}
```

### Error Message and Stack Trace (if applicable)

```
[/usr/local/lib/python3.10/dist-packages/litellm/llms/OpenAI/openai.py](https://localhost:8080/#) in completion(self, model_response, timeout, optional_params, logging_obj, model, messages, print_verbose, api_key, api_base, acompletion, litellm_params, logger_fn, headers, custom_prompt_dict, client, organization, custom_llm_provider, drop_params)
    789                         headers, response = (
--> 790                             self.make_sync_openai_chat_completion_request(
    791                                 openai_client=openai_client,

[/usr/local/lib/python3.10/dist-packages/litellm/llms/OpenAI/openai.py](https://localhost:8080/#) in make_sync_openai_chat_completion_request(self, openai_client, data, timeout)
    650             else:
--> 651                 raise e
    652 

[/usr/local/lib/python3.10/dist-packages/litellm/llms/OpenAI/openai.py](https://localhost:8080/#) in make_sync_openai_chat_completion_request(self, openai_client, data, timeout)
    632         try:
--> 633             raw_response = openai_client.chat.completions.with_raw_response.create(
    634                 **data, timeout=timeout

[...]

PR #28198, opened 2024-11-19
Invalid `tool_choice` being passed to `ChatLiteLLM`

keenborder786:
- **Description:** Invalid `tool_choice` is given to `ChatLiteLLM` to `bind_tools` due to it's parent's class default value being pass through `with_structured_output`.
- **Issue:** #28176


keenborder786:
@efriis

keenborder786:
@efriis any updates please

ccurme (commented):
Issue https://github.com/langchain-ai/langchain/issues/28176 mentions that things work fine with Anthropic. I think OpenAI uses ["required"](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice) instead of "any" for this purpose.

It's not clear to me that we should categorically change everything to "auto", since this makes things worse for other providers that do support "any". We might special-case specific providers like OpenAI instead.

[LiteLLM docs](https://docs.litellm.ai/docs/completion/input) also suggest that dict format should be compatible. We handle this in some other providers, could do something similar here. See for example: https://github.com/langchain-ai/langchain/blob/ec205fcee093e6bea3e8a5c36a2e81f9a3fe803b/libs/partners/openai/langchain_openai/chat_models/base.py#L1108-L1118

keenborder786:
@ccurme I verified each of the Providers and OpenAI is the only exception for when tool_choice is `any`. Therefore I have changed my code to account for this i.e switch to `required` if the model is from OpenAI. Also if it is a bool value than in case of OpenAI it switches to `required` as well.

keenborder786:
@ccurme  To verify that call is actually coming from OpenAI, given ChatLiteLLM I could only use `model` or `model_name`

keenborder786:
@ccurme  does this look good to you?

keenborder786:
@ccurme

tvvignesh:
@ccurme Can this be reviewed and merged please if it works - I am facing the same issue as well when trying to invoke tool in OpenAI from langchain with ChatLiteLLM

keenborder786:
@ccurme

ccurme (approved):

danielbichuetti:
Hello! I wanted to let you know that this issue also appears when using VertexAI. Here’s the error message I’m getting:

`litellm.UnsupportedParamsError: VertexAI doesn't support tool_choice=any. Supported tool_choice values=['auto', 'required', json object].`

Drummersbrother:
The list of openai models should be updated, newer model strings like `o3-mini` need the same treatment (i understand that there's not much other than this list of model strings available to check if model is from openai at the moment)

PR #28212, opened 2024-11-19
community: tongyi multimodal response format fix to support langchain

achenglike:
- **Description:** The multimodal(tongyi) response format "message": {"role": "assistant", "content": [{"text": "图像"}]}}]} is not compatible with LangChain.
- **Dependencies:** No

### DIFF
--- a/libs/community/langchain_community/chat_models/tongyi.py
+++ b/libs/community/langchain_community/chat_models/tongyi.py
@@ -558,19 +558,22 @@ class ChatTongyi(BaseChatModel):
                     yield check_response(resp)
 
         return _stream_completion_with_retry(**kwargs)
-        
+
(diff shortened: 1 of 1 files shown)


efriis (commented):
at a minimum let's make this a hidden `_fix...` method so it doesn't show in the api docs. Can we handle this inline instead though? I feel like message parsing should maybe also happen in other places than streaming...

achenglike:
> at a minimum let's make this a hidden `_fix...` method so it doesn't show in the api docs. Can we handle this inline instead though? I feel like message parsing should maybe also happen in other places than streaming...
Thank you for your advice. I have created a new PR #28645 (using inline). By the way, I only encountered this issue with the streaming implementation, as Tongyi uses different packages for streaming and one-shot responses.

PR #28228, opened 2024-11-20
community: fix some features on Naver ChatModel & embedding model

hyper-clova:
# Description

- adding stopReason to response_metadata to call stream and astream
- excluding NCP_APIGW_API_KEY input required validation
- to remove warning Field "model_name" has conflict with protected namespace "model_".

cc. @vbarda

### DIFF
--- a/libs/community/langchain_community/chat_models/naver.py
+++ b/libs/community/langchain_community/chat_models/naver.py
@@ -51,6 +51,12 @@ def _convert_chunk_to_message_chunk(
     role = message.get("role")
     content = message.get("content") or ""
 
+    if sse.event == "result":
+        response_metadata = {}
+        if "stopReason" in sse_data:
+            response_metadata['stopReason'] = sse_data["stopReason"]
+        return AIMessageChunk(content='', response_metadata=response_metadata)
+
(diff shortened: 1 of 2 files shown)


efriis (approved):
hey! any interest in deprecating this and moving to a self-managed partner package via this guide?

very similar to your original PR, just would be in a repo under your management instead of ours: https://python.langchain.com/docs/contributing/how_to/integrations/

efriis:
(this is a new guide by the way - sorry it wasn't available during the initial integration - we're learning as a team!)

hyper-clova:
@efriis , I waited for the updates because the docs were in draft version at the time you left the comments. Now I see the final version and will look into it.

For clarification, do you have plan for deprecating langchain-community then? (if yes, in what is the expected timeline?)

From our perspective, we just announced to our users about our integration availability this month, so we want to maintain in this way for a moment. (Also we have several upcoming minor fixes, wishing to be updated via community package like this issue.)

But as we are preparing a new API spec in next year (about 2Q), it might be ok then to migrating to a self-managed partner package as you mentioned. Would like to hear your thought on whether it is ok with this timeline.

Issue #26029, opened 2024-09-04
safe_mode Parameter in ChatMistralAI Class Should Not Be Set to False by default or safe_prompt body parameter not sent to mistral api


hadi-ilies:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
        llm = ChatMistralAI(
            client=httpx.Client(
                base_url=llm_creds.api_base,
                headers={
                    "Content-Type": "application/json",
                    "Accept": "application/json",
                },
                timeout=120,
            ),
            temperature=temperature,
            max_tokens=max_tokens,
            model_name="mistral-large",
            safe_mode=False,
            mistral_api_key=llm_creds.api_key,
            max_retries=5,
        ), None
```

### Error Message and Stack Trace (if applicable)

error 422 from Mistral API called through langchain

### Description

I am encountering an issue with the ChatMistralAI class where the `safe_mode` parameter, which corresponds to the `safe_prompt` parameter in the [Mistral documentation](https://docs.mistral.ai/capabilities/guardrailing/), is set to False by default or sent by default to false. According to the Mistral API the safe_prompt parameter cannot be set to False. If this parameter is set to False, the API returns a 422 error. to disable the safe_prompt feature it should not be sent

## Steps to Reproduce:
Initialize an instance of the ChatMistralAI class with the default settings.
Call a method that triggers an API request with the safe_prompt parameter set to False.
Observe the 422 error returned by the Mistral API.
Expected Behavior:
The safe_mode parameter should not be set to False by default or sent by default. Instead, it should either be set to True by default or omitted from the request altogether, unless explicitly set by the user.

## Actual Behavior:
When the safe_mode parameter is set to False (the current default) or sent by default, the Mistral API returns a 422 error, indicating that the safe_prompt parameter should not be False.

## Suggested Fix:
Update the ChatMistralAI class to ensure that the safe_prompt body parameter is omitted from the request unless explicitly specified by the user with `safe_mode` enabled.

### System Info

mac and linux, and last version of langchain

VaibhavLakshmiS:
[...]

PR #28233, opened 2024-11-20
partners: fix of issue #26029

blaufink:
- Description: Azure AI takes an issue with the safe_mode parameter being set to False instead of None. Therefore, this PR changes the default value of safe_mode from False to None. This results in it being filtered out before the request is sent - avoind the extra-parameter issue described below.

- Issue: #26029

- Dependencies: /


ccurme:
I may be confused but Mistral's API documents `safe_prompt` has having a default of false: https://docs.mistral.ai/api/#tag/chat/operation/chat_completion_v1_chat_completions_post

Integration tests appear to be [passing on master](https://github.com/langchain-ai/langchain/actions/runs/11960168645/job/33343568313) (though they also pass on this branch).

Let me know what else we can try to reproduce this.

efriis:
@blaufink I think this is a difference between how mistral's own api performs being different from the azure ai hosted ones.

The Azure AI team is working on an integration currently - that will be the best thing to use once it's out

I'll modify this to no longer disagree with how mistral's own apis accept this parameter in the docs (shouldn't be null)

efriis:
oh I see we filter it already - yeah this works for me

efriis (approved):

Issue #28237, opened 2024-11-20
Investigate tool message usage with mistral


eyurtsev:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

```from langchain_core.messages import HumanMessage, ToolMessage
from langchain_core.tools import tool
from langchain_mistralai import ChatMistralAI


@tool
def example_tool(action: str):
    """Perform an example action"""
    print(f"Performing action: {action}")
    return f"Action {action} completed!"


llm = ChatMistralAI(model="mistral-large-latest", api_key="",
                    temperature=0).bind_tools([example_tool])

human_message = HumanMessage(role="user", content="perform action A")

result = llm.invoke([human_message])

tool_call = result.additional_kwargs["tool_calls"][0]
print(f"Tool Call: {tool_call['function']['arguments']}")

tool_result = example_tool.invoke("A")

tool_message = ToolMessage(
    content=tool_result,
    tool_call_id=tool_call["id"]
)

conversation = [
    human_message,
    result,
    tool_message
]

response = llm.invoke(conversation)

print(response.content)
```

```
    raise httpx.HTTPStatusError(
httpx.HTTPStatusError: Error response 500 while fetching https://api.mistral.ai/v1/chat/completions: {"object":"error","message":"Service unavailable.","type":"internal_server_error","param":null,"code":"3000"} (edited) 
```

eyurtsev:
https://github.com/langchain-ai/langchain/pull/28238/files

eyurtsev:
Will be released in langchain-mistral 0.2.2

PR #28238, opened 2024-11-20
mistral[patch]: Propagate tool call id

eyurtsev:
mistralai-large-2411 requires tool call id

Older models accept tool call id if its provided

mistral-large-2407 
mistral-large-2402


eyurtsev:
Fixes issue https://github.com/langchain-ai/langchain/issues/28237

PR #28243, opened 2024-11-20
infra: remove stale dockerfiles from repo

efriis:
deleting the following docker things from monorepo. they aren't currently usable because of old dependencies, and I'd rather avoid people using them / having to maintain them

- /docker
  - this folder has a compose file that spins up postgres,pgvector (separate from postgres and very stale version),mongo instance with default user/password that we've gotten security pings about before. not worth having
  - also spins up a custom dockerfile with onttotext/graphdb - not even sure what that is
- /libs/langchain/dockerfile + dev.dockerfile
  - super old poetry version, doesn't implement the right thing anymore
- .github/workflows/_release_docker.yml, langchain_release_docker.yml
  - not used anymore, not worth having an alternate release path


louislung:
`.devcontainer` should be removed as well as it is not usable anymore since `libs/langchain/dev.Dockerfile` is removed in this PR, it creates confusion

hesreallyhim:
@louislung i just came across this issue - i'm not expert at docker, how do i set up a container (e.g. in VS Code), since it seems the docs are now stale?

PR #28267, opened 2024-11-21
core[patch]: support final AIMessage responses in `tool_example_to_messages`

ccurme:
We have a test [test_structured_few_shot_examples](https://github.com/langchain-ai/langchain/blob/ad4333ca032033097c663dfe818c5c892c368bd6/libs/standard-tests/langchain_tests/integration_tests/chat_models.py#L546) in standard integration tests that implements a version of tool-calling few shot examples that works with ~all tested providers. The formulation supported by ~all providers is: `human message, tool call, tool message, AI reponse`.

 Here we update `langchain_core.utils.function_calling.tool_example_to_messages` to support this formulation.

The `tool_example_to_messages` util is undocumented outside of our API reference. IMO, if we are testing that this function works across all providers, it can be helpful to feature it in our guides. The structured few-shot examples we document at the moment require users to implement this function and can be simplified.

### DIFF
--- a/libs/core/langchain_core/utils/function_calling.py
+++ b/libs/core/langchain_core/utils/function_calling.py
@@ -495,19 +495,24 @@ def convert_to_openai_tool(
 
 
 def tool_example_to_messages(
-    input: str, tool_calls: list[BaseModel], tool_outputs: Optional[list[str]] = None
+    input: str,
+    tool_calls: list[BaseModel],
+    tool_outputs: Optional[list[str]] = None,
+    ai_response: Optional[str] = None,
 ) -> list[BaseMessage]:
     """Convert an example into a list of messages that can be fed into an LLM.
 
     This code is an adapter that converts a single example to a list of messages
     that can be fed into a chat model.
 
-    The list of messages per example corresponds to:
+    The list of messages per example by default corresponds to:
 
     1) HumanMessage: contains the content from which content should be extracted.
     2) AIMessage: contains the extracted information from the model
     3) ToolMessage: contains confirmation to the model that the model requested a tool
         correctly.
+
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/utils/function_calling.py
```diff
@@ -495,20 +495,25 @@ def convert_to_openai_tool(
 
 
 def tool_example_to_messages(
```
ccurme, line 497: This utility was added [8 months ago](https://github.com/langchain-ai/langchain/pull/18602) but is not documented. Thoughts on retroactively marking it `@beta`?

eyurtsev: Sure I don't mind!

hmm tool calls are also limited to BaseModel / should probably supported typed dicts

eyurtsev (approved):

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -495,20 +495,25 @@ def convert_to_openai_tool(
...
-    input: str, tool_calls: list[BaseModel], tool_outputs: Optional[list[str]] = None
+    input: str,
+    tool_calls: list[BaseModel],
+    tool_outputs: Optional[list[str]] = None,
```
eyurtsev, line 500: Missing `*` operator :(

libs/core/langchain_core/utils/function_calling.py
ccurme: Marked `beta`

---

Looked into supporting TypedDict but we need a way of getting the tool name from an **instance** of TypedDict

```python


class Apple(TypedDict):
    color: str


tool_call = Apple(color="red")
tool_call.__class__.__name__
```
```
'dict'
```

PR #28290, opened 2024-11-22
partners[chroma]: add retrieval of embedding vectors

mspronesti:
This PR adds an additional method to `Chroma` to retrieve the embedding vectors, besides the most relevant Documents. This is sometimes of use when you need to run a postprocessing algorithm on the retrieved results based on the vectors, which has been the case for me lately.

Example issue (discussion) requesting this change: https://github.com/langchain-ai/langchain/discussions/20383


mspronesti:
@erikinfo @ccurme

libs/partners/chroma/tests/integration_tests/test_vectorstores.py
```diff
@@ -92,6 +92,24 @@ def test_chroma_with_metadatas_with_scores() -> None:
     assert output == [(Document(page_content="foo", metadata={"page": "0"}), 0.0)]
 
 
+def test_chroma_with_metadatas_with_vectors() -> None:
+    """Test end to end construction and scored search."""
+    texts = ["foo", "bar", "baz"]
+    metadatas = [{"page": str(i)} for i in range(len(texts))]
+    embeddings = FakeEmbeddings()
```
ccurme, line 99: Does this test pass? Do we need [ConsistentFakeEmbeddings](https://github.com/langchain-ai/langchain/blob/6a7155e2bac99c6825d8de20b920b711d9567778/libs/partners/chroma/tests/integration_tests/fake_embeddings.py#L33C7-L33C31) instead of `FakeEmbeddings`?

ccurme (commented):
LGTM aside for question on integration test.

libs/partners/chroma/tests/integration_tests/test_vectorstores.py
mspronesti: It seems to be passing in my locale env. Does it fail for you?

ccurme (approved):

libs/partners/chroma/tests/integration_tests/test_vectorstores.py
ccurme: Took a look at this. The `FakeEmbeddings` in Chroma is returning a fixed vector for each query. So the test isn't very stringent. Updated it to use `ConsistentFakeEmbeddings`, which is different for each query.

PR #28297, opened 2024-11-22
community[patch]: fix errors introduced by pydantic 2.10

ccurme:


libs/community/langchain_community/llms/exllamav2.py
```diff
@@ -43,7 +43,7 @@ class ExLlamaV2(LLM):
     # Langchain parameters
     logfunc: Callable = print
 
-    stop_sequences: List[str] = Field("")
+    stop_sequences: List[str] = Field([""])
```
eyurtsev, line 46: This should be Field([])

happychen666:
Name: pydantic
Version: 2.10.4
this version still has the bug of PydanticUserError: `SQLDatabaseToolkit` is not fully defined; you should define `Callbacks`, then call `SQLDatabaseToolkit.model_rebuild()`.

PR #28298, opened 2024-11-22
core[patch]: Update unit tests to work for pydantic 2.10

eyurtsev:
* Bump pydantic to 2.10 in lock file
* Update unit tests


eyurtsev:
handled in a separate PR

Issue #26147, opened 2024-09-06
Add support for non-OpenAI models on Azure


baskaryan:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Models like llama-3 are now hosted on azure.

ElhamBadri2411:
Hey! We are a group of students from UofT and were wondering if we could take a look at this issue.

hschmied:
it is still relevant imo; there's workarounds/partial solutions, but accessing/using llama & mistral models hosted on azure is still tricky.

PR #28301, opened 2024-11-22
partners/mistral ai: Added support for mistral chat models

ElhamBadri2411:
The main objective of this issue is to extend LangChain's capabilities by supporting Mistral chat models hosted on Azure. While LangChain already supports OpenAI models deployed via Azure, it currently lacks support for non-OpenAI models such as Mistral. 

Issue: #26147


efriis (commented):
isn't `ChatMistralAI` compatible with azure with a custom base url?

Would prefer to use that in the interim, and the azure folks are working on a package that will include this as well!

efriis:
would also recommend checking out `langchain-azure-ai`! https://pypi.org/project/langchain-azure-ai/

will close this for now, and let me know if either of those work for you

PR #28320, opened 2024-11-23
partners: update pinecone client

aulorbe:
This PR updates the Pinecone client to `5.4.0`, as well as its dependencies (`pinecone-plugin-inference` and `pinecone-plugin-interface`). 

Note: `pinecone-client` is now simply called `pinecone`.

**Question for reviewer(s):** should this PR also update the `pinecone` dep in [the root dir's `poetry.lock` file](https://github.com/langchain-ai/langchain/blob/master/poetry.lock#L6729)? Was unsure. (I don't believe so b/c it seems pinned to a lower version likely based on 3rd-party deps (e.g. Unstructured).)

--
TW: @audrey_sage_

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208693659122374


libs/partners/pinecone/pyproject.toml
```diff
@@ -21,15 +21,12 @@ disallow_untyped_defs = "True"
...
-[[tool.poetry.dependencies.numpy]]
-version = "^1"
-python = "<3.12"
-
-[[tool.poetry.dependencies.numpy]]
-version = "^1.26.0"
-python = ">=3.12"
+numpy = [
+    { version = "^1", python = "<3.12" },
+    { version = "^1.26.0", python = ">=3.12" }
```
aulorbe, line 28: Just a structural change to make things cleaner

libs/partners/pinecone/langchain_pinecone/vectorstores.py
```diff
@@ -74,6 +74,7 @@ class PineconeVectorStore(VectorStore):
                     dimension=1536,
                     metric="cosine",
                     spec=ServerlessSpec(cloud="aws", region="us-east-1"),
+                    deletion_protection="disabled",  # Defaults to "enabled"
```
jhamon, line 77: The default behavior is `deletion_protection="disabled"` So this comment is incorrect.

libs/partners/pinecone/pyproject.toml
```diff
@@ -21,15 +21,12 @@ disallow_untyped_defs = "True"
 [tool.poetry.dependencies]
 python = ">=3.9,<3.13"
 langchain-core = "^0.3.15"
-pinecone-client = "^5.0.0"
+pinecone = ">=5.4.0"
```
jhamon, line 24: This should be `"^5.4.0"`. `>=5.4.0` would include future breaking changes which may not be compatible.

Poetry docs have a nice explanation of how these various syntax elements in version numbers are interpreted. https://python-poetry.org/docs/dependency-specification/

jhamon (approved):
Thanks for taking this on, Audrey. There are a few small things I would address.

libs/partners/pinecone/pyproject.toml
aulorbe: Ah thank you so much!

libs/partners/pinecone/langchain_pinecone/vectorstores.py
aulorbe: Eek, silly oversight. Thank you!

efriis (approved):

libs/partners/pinecone/langchain_pinecone/embeddings.py
```diff
@@ -65,6 +65,19 @@ class PineconeEmbeddings(BaseModel, Embeddings):
...
 
+    @property
+    def async_client(self) -> aiohttp.ClientSession:
+        """Lazily initialize the async client."""
+        if self._async_client is None:
+            self._async_client = aiohttp.ClientSession(
+                headers={
+                    "Api-Key": self.pinecone_api_key.get_secret_value(),
+                    "Content-Type": "application/json",
+                    "X-Pinecone-API-Version": "2024-07",
```
jhamon, line 76: You should probably update this API version header while you're here.

PR #28325, opened 2024-11-24
Community: Adding bulk_size as a setable param for OpenSearchVectorSearch

manukychen:
Description:
When using langchain.retrievers.parent_document_retriever.py with vectorstore is OpenSearchVectorSearch, I found that the bulk_size param I passed into OpenSearchVectorSearch class did not work on my ParentDocumentRetriever.add_documents() function correctly, it will be overwrite with int 500 the function which OpenSearchVectorSearch class had (e.g., add_texts(), add_embeddings()...).

So I made this PR requset to fix this, thanks!


libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -618,7 +615,6 @@ def add_embeddings(
         text_embeddings: Iterable[Tuple[str, List[float]]],
         metadatas: Optional[List[dict]] = None,
         ids: Optional[List[str]] = None,
-        bulk_size: int = 500,
```
efriis, line 621: breaking change - can we keep this, and use the passed-in as an override? can still default to the `self.bulk_size` behavior if it's None

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -596,7 +594,6 @@ async def aadd_texts(
         texts: Iterable[str],
         metadatas: Optional[List[dict]] = None,
         ids: Optional[List[str]] = None,
-        bulk_size: int = 500,
```
efriis, line 599: break

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -1085,7 +1081,6 @@ def from_texts(
         texts: List[str],
         embedding: Embeddings,
         metadatas: Optional[List[dict]] = None,
-        bulk_size: int = 500,
```
efriis, line 1088: break

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -1150,7 +1145,6 @@ async def afrom_texts(
...
```
efriis, line 1153: break

efriis (commented):
more breaking changes to the interface

could you revert these to something like `bulk_size: Optional[int] = None`, and replace usage of bulk size with `bulk_size if bulk_size is not None else self.bulk_size`?

manukychen:
Hi @efriis , thanks for reviewing my request, really appreciate it. :)
I've followed your suggestions and modified the code, unless I misinterpreted anything.
If you have any new questions or suggestions, please let me know. Thanks again!

efriis (approved):

epinzur:
This change seems to break `from_documents()`.

Now the bulk_size param is required on the call to prevent an exception.

```
vector_store = OpenSearchVectorSearch.from_documents(
    documents=animals,
    opensearch_url="localhost:9200",
    index_name="animals",
    embedding=embeddings,
    bulk_size=500,  # newly required
)
```

without specifying the param, you get this error:
```
...
/langchain_community/vectorstores/opensearch_vector_search.py:1362) )

AttributeError: type object 'OpenSearchVectorSearch' has no attribute 'bulk_size'
```

epinzur:
looks like there is an existing issue and fix: https://github.com/langchain-ai/langchain/issues/29071

PR #28346, opened 2024-11-26
Community: Load a PDF document as per its Table of Contents

vineetm:
**Description:** Many PDFs such as textbooks come with a rich Table of Contents (also known as bookmarks). It would be helpful for a RAG system to see the entire content for a ToC entry in tandem, rather than limiting it to a single PDF page!

This PR adds adds a new PDF document loader which loads content of PDF by using its Table of Contents

Many PDFs have detailed Table of Contents. For example [this book](https://open.umn.edu/opentextbooks/textbooks/computer-systems-security-planning-for-success) with the [pdf](https://web.njit.edu/~rt494/security/security.pdf) has this Table of Contents structure... 
Screenshot 2024-11-26 at 1 30 37 PM

[PyPfium2Loader](https://github.com/langchain-ai/langchain/blob/aa2c17b56c60012de35aabd054c3cf0fa04a0b20/libs/community/langchain_community/document_loaders/pdf.py#L260) breaks the flow of content which logically continues on the next page.

This PR, loads content based on Toc entry. It can

1. Extract logical chunks within a single page
2. Extract logical chunks of document spanning across multiple pages
3. Is not limited by the page scope


vineetm:
Many PDFs have detailed Table of Contents. For example [this book](https://open.umn.edu/opentextbooks/textbooks/computer-systems-security-planning-for-success) with the [pdf](https://web.njit.edu/~rt494/security/security.pdf) has this Table of Contents structure... 
Screenshot 2024-11-26 at 1 30 37 PM

[PyPffium2Loader](https://github.com/langchain-ai/langchain/blob/aa2c17b56c60012de35aabd054c3cf0fa04a0b20/libs/community/langchain_community/document_loaders/pdf.py#L260) which breaks the flow of content which logically continues on the next page as it is part of the same ToC entry.

This PR, loads content based on Toc entry

efriis:
I don't fully understand what this does (it looks like the docs page you added isn't actually populated), but I think it's unnecessary to add another PDF document loader because we have so many. Depending what the goal of loading in the TOC in this way is, couldn't this be added as some kind of postprocessing on documents in a guide instead?

ccurme:
Closing with recommendation of implementing as:

- A guide (if possible to do downstream of `PyPDFium2Parser`);
- Possibly a mode / parameter of the existing `PyPDFium2Parser`.

PR #28364, opened 2024-11-26
ollama[patch]: support ollama 0.4

ccurme:
v0.4 of the Python SDK is already installed via the lock file in CI, but our current implementation is not compatible with it.

This also addresses an issue introduced in https://github.com/langchain-ai/langchain/pull/28299. @RyanMagnuson would you mind explaining the motivation for that change? From what I can tell the Ollama SDK [does not support kwargs](https://github.com/ollama/ollama-python/blob/6c44bb272976b9dc8d24d3a3fd913fd39b83394b/ollama/_client.py#L286). Previously, unsupported kwargs were ignored, but they currently raise `TypeError`.

Some of LangChain's standard test suite expects `tool_choice` to be supported, so here we catch it in `bind_tools` so it is ignored and not passed through to the client.

### DIFF
--- a/libs/partners/ollama/langchain_ollama/chat_models.py
+++ b/libs/partners/ollama/langchain_ollama/chat_models.py
@@ -66,8 +66,8 @@ def _get_tool_calls_from_response(
     """Get tool calls from ollama response."""
     tool_calls = []
     if "message" in response:
-        if "tool_calls" in response["message"]:
-            for tc in response["message"]["tool_calls"]:
+        if raw_tool_calls := response["message"].get("tool_calls"):
+            for tc in raw_tool_calls:
                 tool_calls.append(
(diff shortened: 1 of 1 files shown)


RyanMagnuson:
@ccurme For sure. I encountered some weird behavior while trying to work with multimodal input to an Ollama model following this [tutorial](https://python.langchain.com/docs/integrations/llms/ollama/). I then switched to using chatollama, and encountered an error in the serialization of the model. At this point I was in contact with @efriis, who I believe worked with the ollama folks to debug and make the change in https://github.com/langchain-ai/langchain/pull/28299. 

ollama version: `0.4.3`
langchain-ollama:`0.2.0`

Initial behavior following tutorial:
image


Error serializing model:
image

PR #28367, opened 2024-11-26
langchain[patch]: init_chat_model provider in model string

baskaryan:
```python
llm = init_chat_model("openai:gpt-4o")
```

### DIFF
--- a/libs/langchain/langchain/chat_models/base.py
+++ b/libs/langchain/langchain/chat_models/base.py
@@ -468,6 +468,7 @@ def _parse_model(model: str, model_provider: Optional[str]) -> Tuple[str, str]:
     model_provider = model_provider.replace("-", "_").lower()
     return model, model_provider
 
+
 def _check_pkg(pkg: str) -> None:
     if not util.find_spec(pkg):
         pkg_kebab = pkg.replace("_", "-")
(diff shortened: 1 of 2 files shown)


libs/langchain/langchain/chat_models/base.py
```diff
@@ -461,6 +455,19 @@ def _attempt_infer_model_provider(model_name: str) -> Optional[str]:
         return None
 
 
+def _parse_model(model: str, model_provider: Optional[str]) -> Tuple[str, str]:
+    if ":" in model and model.split(":")[0] in _SUPPORTED_PROVIDERS:
```
hinthornw, line 459: If model_provider is present I think we should give that priority

hinthornw:
```python
     if model_provider is None and ":" in model:
          _provider, _model = model.split(":", maxsplit=1)
          if _provider in _SUPPORTED_PROVIDERS:
              model_provider, model = _provider, _model
      model_provider = model_provider
      if not model_provider:
          raise ValueError(
              f"Unable to infer model provider for {model=}, please specify "
              f"model_provider directly."
          )
      model_provider = model_provider.replace("-", "_").lower()
      return model, model_provider
````

**Review by nfcampos — APPROVED** [review]

File: libs/langchain/langchain/chat_models/base.py [hunk 2]
```diff
@@ -461,6 +455,24 @@ def _attempt_infer_model_provider(model_name: str) -> Optional[str]:
         return None
 
 
+def _parse_model(model: str, model_provider: Optional[str]) -> Tuple[str, str]:
+    if (
+        not model_provider
+        and ":" in model
+        and model.split(":")[0] in _SUPPORTED_PROVIDERS
+    ):
+        model_provider = model.split(":")[0]
```
hinthornw, line 464: ooc why split multiple times

PR #28370, opened 2024-11-26
Init embeddings

hinthornw:


libs/langchain/langchain/embeddings/base.py
```diff
@@ -1,4 +1,226 @@
...
+        return HuggingFaceEmbeddings(model_name=model_name, **kwargs)
+    else:
+        raise ValueError(
+            f"Provider '{provider}' is not supported.\n"
+            f"Supported providers and their required packages:\n"
+            f"{_get_provider_list()}"
+        )
+
+
+@functools.lru_cache
```
baskaryan, line 153: should we set max size

libs/langchain/langchain/embeddings/base.py
```diff
@@ -1,4 +1,226 @@
...
+def _check_pkg(pkg: str) -> None:
+    """Check if a package is installed."""
+    if not util.find_spec(pkg):
+        raise ImportError(
+            f"Could not import {pkg} python package. "
+            f"Please install it with `pip install {pkg}`"
+        )
+
+
+def embedding_model(
```
baskaryan, line 163: in a vacuum i do like this name more but i wonder if we should stick to init_embedding_model for consistency

libs/langchain/langchain/embeddings/base.py
```diff
@@ -1,4 +1,226 @@
...
+    Raises:
+        ValueError: If the model provider is not supported or cannot be determined
+        ImportError: If the required provider package is not installed
+
+    Examples:
+        >>> # Using a model string
+        >>> model = init_embedding_model("openai:text-embedding-3-small")
+        >>> model.embed_query("Hello, world!")
+        
+        >>> # Using explicit provider
```
baskaryan, line 200: would double check that these render correctly in api ref, not sure they will

libs/langchain/langchain/embeddings/base.py
baskaryan: also might want to mark as beta initially

baskaryan (commented):
slick

libs/langchain/langchain/embeddings/base.py
hinthornw: ya i'm fine with it

libs/langchain/langchain/embeddings/base.py
hinthornw: sure! though I hope people don't spam the provider list

libs/langchain/langchain/embeddings/base.py
hinthornw: wlil fix

libs/langchain/langchain/embeddings/base.py
```diff
@@ -1,4 +1,235 @@
...
+    .. versionadded:: 0.3.9
+    """
+    if not model:
+        providers = _SUPPORTED_PROVIDERS.keys()
+        raise ValueError(
+            "Must specify model name. "
+            f"Supported providers are: {', '.join(providers)}"
+        )
+
+    return _init_embeddings_helper(model, provider=provider, **kwargs)
```
baskaryan, line 229: ubernit, helper doesn't really feel needed here

baskaryan (approved):

PR #28269, opened 2024-11-21
docs: correct HuggingFaceEmbeddings documentation model param

manuelescobar-dev:
- **Description:** Corrected the parameter name in the HuggingFaceEmbeddings documentation under integrations/text_embedding/ from model to model_name to align with the actual code usage in the langchain_huggingface package.
- **Issue:** Fixes #28231
- **Dependencies:** None

PR #28374, opened 2024-11-27
partners: add 'model' alias for consistency in embedding classes

manuelescobar-dev:
**Description:**  This PR introduces a `model` alias for the embedding classes that contain the attribute `model_name`, to ensure consistency across the codebase, as suggested by a moderator in a previous PR. The change aligns the usage of attribute names across the project (see for example [here](https://github.com/langchain-ai/langchain/blob/65deeddd5dfec5d51f33ebc961f09c2e47a8f064/libs/partners/groq/langchain_groq/chat_models.py#L304)).
**Issue:**  This PR addresses the suggestion from the review of issue #28269.
**Dependencies:**  None


eyurtsev:
Needs this: 

```
    model_config = ConfigDict(
        populate_by_name=True,
    )```

otherwise it's a breaking change. I'll add

File: libs/partners/qdrant/langchain_qdrant/fastembed_sparse.py [hunk 1]
```diff
@@ -8,7 +10,7 @@ class FastEmbedSparse(SparseEmbeddings):
 
     def __init__(
         self,
-        model_name: str = "Qdrant/bm25",
+        model_name: str = Field(default="Qdrant/bm25", alias="model"),
```
eyurtsev, line 13: This is not a pydantic model -- that's not going to work, so revering changes here

eyurtsev (approved):

PR #28376, opened 2024-11-27
fix: Set default search_type to 'similarity' in as_retriever method of AzureSearch

TamagoTorisugi:
**Description**
This PR updates the `as_retriever` method in the `AzureSearch` to ensure that the `search_type` parameter defaults to 'similarity' when not explicitly provided. 

Previously, if the `search_type` was omitted, it did not default to any specific value. So it was inherited from `AzureSearchVectorStoreRetriever`, which defaults to 'hybrid'.

This change ensures that the intended default behavior aligns with the expected usage.

**Issue**
No specific issue was found related to this change.

**Dependencies**
No new dependencies are introduced with this change.


efriis (commented):
I believe the `search_type` must be a string, and it defaults to "hybrid", so I don't think this PR makes sense

efriis:
i.e. if you look at the default values in the `__init__` of `AzureSearch`: https://python.langchain.com/api_reference/community/vectorstores/langchain_community.vectorstores.azuresearch.AzureSearch.html#azuresearch

TamagoTorisugi:
Thank you for your comment, @efriis .

The intent of this PR is to align with the comment and documentation for `search_type` of `as_retriever` which states "Can be "similarity" (default), "hybrid", or "semantic_hybrid". 
https://python.langchain.com/api_reference/community/vectorstores/langchain_community.vectorstores.azuresearch.AzureSearch.html#langchain_community.vectorstores.azuresearch.AzureSearch.as_retriever

As you've correctly pointed out, the default value of `search_type` is "hybrid". Therefore, I think we have two options: either merge this PR or update the comment and documentation.

efriis (approved):

Issue #27868, opened 2024-11-03
PipelinePromptTemplate doesn't support `partial`


dmenini:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
      # Example from docs: https://python.langchain.com/v0.1/docs/modules/model_io/prompts/composition/#using-pipelineprompt
      full_template = """{introduction}
      {example}
      {start}"""
      full_prompt = PromptTemplate.from_template(full_template)
  
      introduction_template = """You are impersonating {person}."""
      introduction_prompt = PromptTemplate.from_template(introduction_template)
      
      example_template = """Here's an example of an interaction:
      Q: {example_q}
      A: {example_a}"""
      example_prompt = PromptTemplate.from_template(example_template)
  
      start_template = """Now, do this for real!
      Q: {input}
      A:"""
      start_prompt = PromptTemplate.from_template(start_template)
  
      input_prompts = [
          ("introduction", introduction_prompt),
          ("example", example_prompt),
          ("start", start_prompt),
      ]
      pipeline_prompt = PipelinePromptTemplate(
          final_prompt=full_prompt, pipeline_prompts=input_prompts
      )
  
      pipeline_prompt.partial(person="Elon Musk")
```

### Error Message and Stack Trace (if applicable)

```python
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../.venv/lib/python3.12/site-packages/langchain_core/prompts/base.py:277: in partial
    return type(self)(**prompt_dict)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = PipelinePromptTemplate(input_variables=['person', 'input', 'example_q', 'example_a'], input_types={}, partial_variable...ariables=['input'], input_types={}, partial_variables={}, template='Now, do this for real!\n    Q: {input}\n    A:'))])
args = ()
[...]

PR #28377, opened 2024-11-27
Core: Added partial functionality to PipelinePromptTemplate

Aarya2004:
**Description:** Added extra logic in the PipelinePromptTemplate to support partial assignments primarily by overriding the partial method from BasePromptTemplate
**Issue:** Fixes issue #27868
**Dependencies:** No dependencies required or new dependencies added

**Add tests**
Added test in test_pipeline_prompt.py to check for this case.

**Lint and test**: Done


Aarya2004:
Hello @efriis! Sorry for the ping, but you had reviewed a PR related to Prompt Templates recently (#28011) so I was hoping you would be able to review this too. Let me know if you have any questions, thank you!

efriis (commented):
I think this is ok. Quickly discussing with team to see what we want to do with PipelinePromptTemplate because it's not that intuitive of an abstraction, and it might make sense to just implement as a RunnableSequence of RunnableAssigns instead...

i.e. I think this does the same thing from my understanding:
```
keys_and_prompts: list[tuple[str, PromptTemplate]] = ...


chain = RunnablePassthrough()
for key, prompt in keys_and_prompts:
  chain = chain | RunnablePassthrough.assign(key, prompt)

chain = chain | lambda x: x[keys_and_prompts[-1][0]]
```

Aarya2004:
Sounds good, I wasn't sure if this complied with how PipelinePromptTemplate is 'supposed' to be used but I thought it made sense as a functionality given that the PromptTemplate class implements this function (and PipelinePromptTemplate used a list of these PromptTemplates for its functionality) but I do see how it makes sense through your code. Thanks for taking a look at the code, let me know what you and the team decide!

efriis:
Are you using it personally? If so could you check if the code snippet solves your use case?

Aarya2004:
No I'm not, I'm just trying to fix the issue #27868

efriis:
got it. I can respond there, and I'll close this because it introduces a new bug, where `partial` implements a different interface (stateful mutation instead of stateless wrap)

Aarya2004:
Thank you!

PR #28382, opened 2024-11-27
doc-loader: retain Azure Doc Intelligence API metadata in Document parser

jmohren:
**Description**:
This PR modifies the doc_intelligence.py parser in the community package to include all metadata returned by the Azure Doc Intelligence API in the Document object. Previously, only the parsed content (markdown) was retained, while other important metadata such as bounding boxes (bboxes) for images and tables was discarded. These image bboxes are crucial for supporting use cases like multi-modal RAG workflows when using Azure Doc Intelligence.

The change ensures that all information returned by the Azure Doc Intelligence API is preserved by setting the metadata attribute of the Document object to the entire result returned by the API, rather than an empty dictionary. This extends the parser's utility for complex use cases without breaking existing functionality.

**Issue**:
This change does not address a specific issue number, but it resolves a critical limitation in supporting multimodal workflows when using the LangChain wrapper for the Azure API.

**Dependencies**:
No additional dependencies are required for this change.

### DIFF
--- a/libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
+++ b/libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
@@ -71,7 +71,7 @@ class AzureAIDocumentIntelligenceParser(BaseBlobParser):
             yield d
 
     def _generate_docs_single(self, result: Any) -> Iterator[Document]:
-        yield Document(page_content=result.content, metadata={})
+        yield Document(page_content=result.content, metadata=result)
 
     def lazy_parse(self, blob: Blob) -> Iterator[Document]:
         """Lazily parse the blob."""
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
```diff
@@ -71,7 +71,7 @@ def _generate_docs_page(self, result: Any) -> Iterator[Document]:
             yield d
 
     def _generate_docs_single(self, result: Any) -> Iterator[Document]:
-        yield Document(page_content=result.content, metadata={})
+        yield Document(page_content=result.content, metadata=result)
```
ccurme, line 74: is `result` always a dict? I see it typed as `Any`.

jmohren: No you are right, it is AnalyzeResult which has an as_dict method. I will adjust accordingly, thanks for pointing out!
image

jmohren: fixed!

jmohren: @ccurme I kindly ask if there is an update in the review process :)

ccurme (approved):

PR #28393, opened 2024-11-27
nomic[patch]: run python 3.13 in CI

ccurme:


ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain/actions/runs/12055002100/job/33614354193

Issue #27988, opened 2024-11-08
Youtube requires login to view videoDetails


ymcki:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

I was trying to run this jupyter notebook
https://github.com/langchain-ai/rag-from-scratch/blob/main/rag_from_scratch_10_and_11.ipynb

Then it crashes at the beginning

```
from langchain_community.document_loaders import YoutubeLoader

docs = YoutubeLoader.from_youtube_url(
    "https://www.youtube.com/watch?v=pbAd8O1Lvm4", add_video_info=True,
).load()

print(docs[0].metadata)
```

### Error Message and Stack Trace (if applicable)

Traceback (most recent call last):
  File "/home/user/anaconda3/envs/ai/lib/python3.10/site-packages/pytube/__main__.py", line 341, in title
    self._title = self.vid_info['videoDetails']['title']
KeyError: 'videoDetails'

### Description

After I dig more into why it crashed, I find that it is due to this video requires login.

By modifying line 322 of
langchain_community/document_loaders/youtube.py
from 
yt = YouTube(f"https://www.youtube.com/watch?v={self.video_id}")
to
yt = YouTube(f"https://www.youtube.com/watch?v={self.video_id}", use_oauth=True, allow_oauth_cache=True)

Then it asks me to login and then it works.

I think this "bug" can be fixed by adding these two arguments to YoutubeLoader.from_youtube_url and pass them to pytube.

### System Info

langchain 0.3.4
langchain-community 0.3.3

Harsimran-19:
Can I work on this issue?

sonalshreya:
I want to work on this issue

Patotricks15:
The same problem here trying to develop a langchain project using YoutubeLoader

PR #28427, opened 2024-11-29
Langchain Community: Added OAuth Support to YoutubeLoader for restricted videos

ashvini8:
## Description
(This PR has contributions from @khushiDesai,  @anushak18, and @ssumaiyaahmed).

This PR addresses **Issue #27988** which addresses the bug where YoutubeLoader fails when attempting to load videos that require login credentials. The issue seemed to occur due to the lack of OAuth support in the pytube integration.

This pull request adds the following:
- Added ```use_oauth``` and ```allow_oauth_cache``` as optional arguments to ```YoutubeLoader.from_youtube_url``` in ```youtube.py```

## Issues
**Issue #27988**: Youtube requires login to view videoDetails

## Testing
We added the following tests in ```test_youtube.py```:
- ```def test_from_youtube_url``` checks that the ```from_youtube_url``` method correctly creates a YoutubeLoader and extracts the correct video information from various YouTube URL formats while also checks the OAuth settings
- ```def test_oauth_cache``` checks that the OAuth caching works properly and prevents unnecessary login prompts when loading a video using the YoutubeLoader


libs/community/langchain_community/document_loaders/youtube.py
```diff
@@ -163,6 +163,9 @@ def __init__(
         transcript_format: TranscriptFormat = TranscriptFormat.TEXT,
         continue_on_failure: bool = False,
         chunk_size_seconds: int = 120,
+        use_oauth: bool = True,
```
baskaryan, line 166: setting to True by default is a breaking change, lets set to False

ccurme:
Closing as author is unresponsive but happy to re-open if we want to keep iterating on this.

Issue #27717, opened 2024-10-29
Voice Input Support for Ollama Models


efriis:
### Discussed in https://github.com/langchain-ai/langchain/discussions/27404

<div type='discussions-op-text'>

<sup>Originally posted by **kodychik** October 16, 2024</sup>
### Checked

- [X] I searched existing ideas and did not find a similar one
- [X] I added a very descriptive title
- [X] I've clearly described the feature request and motivation for it

# Feature request

We (a team of CS students at the University of Toronto) propose that we add voice input support to LangChain's Ollama models. 

# Motivation

LangChain currently supports the best models via Ollama integration but lacks the ability to accept voice inputs on these Ollama models. This limitation restricts its use in voice-enabled applications such as virtual assistants, voice-controlled systems, and accessibility tools. This enhancement will enable developers to build applications that can process spoken language, expanding the ways users can interact with LangChain-powered systems.

# Proposal

## Feasibility Analysis
Feasible, it involves:
● Speech-to-Text Conversion: Using a speech recognition engine to transcribe voice
inputs into text that the language model can process.
● Integration with Existing Pipelines: Modifying or extending existing chains to include a
speech-to-text (STT) component before processing inputs with the LLM.
● Modular Implementation: Leveraging LangChain's modular architecture to add this
functionality without significant changes to existing code.

## Outline of Changes
Existing Architecture Overview
LangChain's architecture consists of:
● LLMs (Language Models): Interfaces to language models via Ollama.
● Chains: Sequences of components (e.g., prompt templates, LLMs) that process inputs
and generate outputs.
● Agents: Systems that use LLMs to perform tasks by making decisions and possibly
interacting with tools.
● Retrievers and VectorStores: Components used in Retrieval-Augmented Generation
(RAG) pipelines to fetch relevant information.

## Proposed Solution
Introduce a Speech-to-Text Component that converts voice inputs into text, integrating
seamlessly with existing LangChain chains and agents.
1. User Interaction: User provides voice input via microphone.
2. Speech-to-Text Conversion:
○ The STT component transcribes the voice input into text.
3. Text Processing:
○ The transcribed text is passed to existing LangChain chains or agents.
4. LLM Response:
○ The LLM generates a response based on the input text.
5. Output Delivery:
○ The response is delivered to the user (could be text or converted back to
speech).

## Files to Modify and Create
New Files:
● speech_to_text.py: Implements the SpeechToTextConverter class.
Files to Modify:
[...]

PR #28428, opened 2024-11-30
community: Voice Input Support For Ollama Models

faizannaseerr:
Tool to provide voice input support for Ollama models.

Description: Added voice input tool for Ollama model that holds two classes that extend BaseBlobParser & BaseTool respectively, to allow for voice input (pre-saved or on the spot) on various Ollama model chains, as the user requires. Details on how to utilize the tool (e.g. use cases) can be found in the docs/integrations/tools folder in 'ollama_voice_input.ipynb'.

Issue: #27717 

Dependencies:  'sounddevice', 'scipy' for voice recording (not anymore as per my knowledge, except for note below)

Twitter Handle: @faizannaseerr

Lint & Test: I ran lint & tests locally, and everything seemed fine as per my knowledge.

Added Test & Docs: Unit tests for modules and docs on how to use the tool have been added.

Additional Notes: 

- There is a unit test for long audio input that should use an audio file named 'long_audio.wav' in the langchain_community/tests/examples file. I have not added this file yet as I could not find any long, free-to-use, audio input file online and therefore can leave this unit test out if required, or can use it if a long audio input file is found (that leads to several Document objects after the parse)
- Will fix imports shortly if required (still some being re-imported from LangChain, not sure if they need to be altered or not)


efriis:
going to close because the ci issues stem from dependencies we don't want to package here. The description alludes to this , and if you want to re-contribute, would recommend reading our docs on optional dependencies, as well as some of the Whisper parsers in community (instead of a tool implemented here)

PR #28434, opened 2024-11-30
docs: fixed a variable name in embeddings section

rcocetta:
This is a simple change for a variable name in the Embeddings section of this document: https://python.langchain.com/docs/tutorials/retrievers/#embeddings .

The variable name `embeddings_model` seems to be wrong and doesn't match what follows in the template: `embeddings`.


ccurme (approved):
Thank you!

PR #28437, opened 2024-12-01
docs: Fixed broken link in the warning message to @tool API Reference…

DangerousPotential:
Fixed the broken hyperlink in the warning of docstring section to the correct `@tool` API reference


DangerousPotential:
Bump #28042 which also modifies the scale factor from a `str` to `int`

baskaryan (approved):

Issue #28436, opened 2024-12-01
DOC: HuggingFaceEmbeddings use `FP16` by nested loop


MoosaTae:
### URL

https://python.langchain.com/docs/integrations/text_embedding/bge_huggingface/

### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:
```py
model_name = "BAAI/bge-small-en"
model_kwargs={
      "device": device,
      "torch_dtype": torch.float16
  },
encode_kwargs = {"normalize_embeddings": True}
hf = HuggingFaceBgeEmbeddings(
    model_name=model_name, model_kwargs=model_kwargs, encode_kwargs=encode_kwargs
)
```

cause to error

```
_embeddings_instance = HuggingFaceBgeEmbeddings(
    |   File "/Users/tae/mambaforge/envs/api/lib/python3.9/site-packages/langchain_huggingface/embeddings/huggingface.py", line 58, in __init__
    |     self._client = sentence_transformers.SentenceTransformer(
    | TypeError: __init__() got an unexpected keyword argument 'torch_dtype'
```

MoosaTae:
```py
model_name = "BAAI/bge-small-en"
model_kwargs={
      "device": device,
      "model_kwargs": {"torch_dtype": torch.float16},
  },
encode_kwargs = {"normalize_embeddings": True}
hf = HuggingFaceBgeEmbeddings(
    model_name=model_name, model_kwargs=model_kwargs, encode_kwargs=encode_kwargs
)
```


can you right now but it need to parse to be nested loop, should it fix document or code?

keenborder786:
@MoosaTae I have created a PR which resolves this issue.

PR #28439, opened 2024-12-01
Passing the `model_kwargs` correctly while maintaing backward compatability

keenborder786:
- **Description:** `Model_Kwargs` was not being passed correctly to `sentence_transformers.SentenceTransformer` which has been corrected while maintaing backward compatability
- **Issue:** #28436


keenborder786:
@ccurme

keenborder786:
@ccurme

keenborder786:
@ccurme

efriis (commented):
@keenborder786  ooc do you understand why this BgeEmbeddings are a different class? Should we deprecate this integration in favor of using `langchain_huggingface.HuggingFaceEmbeddings`?

keenborder786:
> @keenborder786 ooc do you understand why this BgeEmbeddings are a different class? Should we deprecate this integration in favor of using `langchain_huggingface.HuggingFaceEmbeddings`?

Yes, you are right, we should deprecate `HuggingFaceBgeEmbeddings` because I see no difference except for the fact certain instructions are being concatenated as a suffix to the given text which needs to be embedded. 

@efriis should I raise the deprecation warning in the same PR.

efriis:
yes please. I would actually prefer to not modify the behavior as well just in case anyone is passing kwargs in a way they're not expecting, and folks can migrate for the correct behavior!

keenborder786:
@efriis done

efriis (approved):

sam-bercovici:
The HuggingFaceBgeEmbeddings accepts:
query_instruction = "search_query:",
embed_instruction = "search_document:"

The HuggingFaceEmbeddings does NOT accept those parameters.

How can the HuggingFaceEmbeddings used instead of HuggingFaceBgeEmbeddings

sam-bercovici:
The HuggingFaceBgeEmbeddings accepts:
query_instruction = "search_query:",
embed_instruction = "search_document:"

The HuggingFaceEmbeddings does NOT accept those parameters.

How can the HuggingFaceEmbeddings used instead of HuggingFaceBgeEmbeddings

Issue #25696, opened 2024-08-23
ReAct agent creating own "User Input" and ending up stopping agent with iteration limit


Rohannain:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

# Agent and Executor setup
```python 
agent = create_react_agent(llm, tools, full_prompt)
agent_executor = AgentExecutor(agent=agent, tools=tools, verbose=True, handle_parsing_errors=True)
langchain.debug = True```

### Error Message and Stack Trace (if applicable)

Agent stopped due to iteration limit exceeded.

### Description

I was using langchain.agents AgentExecutor, create_react_agent. I got the output for the user input in my chain but it is creating its own user input and keep answering it with the given available tools.
<img width="1512" alt="Screenshot 2024-08-23 at 4 08 40 PM" src="https://github.com/user-attachments/assets/1b9b8537-8ab2-49de-806f-0f85f438b688">


### System Info

langchain==0.2.11
langchain-community==0.2.7
langchain-core==0.2.24
langchain-experimental==0.0.62
langchain-huggingface==0.0.3
langchain-openai==0.1.16
langchain-text-splitters==0.2.1

platform mac

dnspython==2.6.1
opencv-python==4.10.0.82
python-dateutil==2.9.0.post0
python-dotenv==1.0.1
python-multipart==0.0.9

Ahonanhin:
Hello, I’m interested in working on this issue as part of a project for my computer science course at the University of Toronto. Could you please let me know if it’s available and if there are any specific considerations I should be aware of? Thanks in advance!

Ahonanhin:
Hi @Rohannain ,

We are interested in working on the issue you’ve reported where the agent creates its own user input and exceeds the iteration limit. We were able to reproduce the error using a different prompt, which suggests that the issue might be related to how the agent handles prompts in general.

To help us investigate further, could you please provide some additional information?

- Are there any custom configurations or parameters you’ve applied that might affect the agent’s behavior?
- Could you share the specific code snippets you used to set up the `AgentExecutor` and `create_react_agent`?

andrewaucie:
[...]

PR #28441, opened 2024-12-02
Add safeguards for infinite loops and unresolvable tasks

Matthewi1012:
Description

In the issue referenced below, the agent generates it's own "User Input" prompts during execution. Rather than using provided tools it requests input data from the user whick it cannot receive. This results in Inifinite loops, ultimately reaching the iteration limit, and provides the following error message. "Agent stopped due to iteration limit". In this PR we provided a proposed solution with associated unit and integration tests, for the specific instances of infinite loops from user input.

Issue: https://github.com/langchain-ai/langchain/issues/25696

Dependencies:
No additional dependencies

Twitter Handle:
None


efriis:
hey there! I believe this would be better implemented by the user in a langgraph agent: https://langchain-ai.github.io/langgraph/tutorials/introduction/

This out of the box AgentExecutor does have shortcomings, hence the greater degree of control in langgraph. I don't think this change to AgentExecutor and supporting data objects is desirable, so I will close this PR

PR #23729, opened 2024-07-01
Community[minor]: Update VDMS vectorstore

cwlacewe:
**Description:** 
- This PR exposes some functions in VDMS vectorstore, updates VDMS related notebooks, updates tests, and upgrade version of VDMS (>=0.0.20)

**Issue:** N/A

**Dependencies:** 
- Update vdms>=0.0.20

PR #28443, opened 2024-12-02
community: update VDMS vectorstore with latest VDMS release

cwlacewe:
**Description:** 
- Update to use changes made to VDMS v2.10.0 and include some optimizations
- Use test suite for testing as suggested in [PR #23729](
https://github.com/langchain-ai/langchain/pull/23729#discussion_r1674363313).
- Reran VDMS related notebooks using changes

**Issue:** N/A

**Dependencies:** N/A


cwlacewe:
@efriis, @ccurme,
Could someone please review this PR?

libs/community/langchain_community/vectorstores/vdms.py
```diff
@@ -171,586 +129,664 @@ class VDMS(VectorStore):
...
 
+    @override
     def __init__(
         self,
         client: vdms.vdms,
         *,
-        embedding: Optional[Embeddings] = None,
-        collection_name: str = DEFAULT_COLLECTION_NAME,  # DescriptorSet name
-        distance_strategy: DISTANCE_METRICS = "L2",
+        embedding: Embeddings,
```
baskaryan, line 137: breaking change

baskaryan (commented):
thanks for the update!

before fully reviewing, could you provide a bit more context? seems like there's some breaking changes, are these necessary for compatibility with the new vdms release? or are those two separate things

libs/community/langchain_community/vectorstores/vdms.py
cwlacewe: @baskaryan, this change wasn't necessary. I reverted it

libs/community/langchain_community/vectorstores/vdms.py
```diff
@@ -171,586 +129,664 @@ class VDMS(VectorStore):
...
+        filter: Optional[dict[str, Any]] = None,
         **kwargs: Any,
-    ) -> List[str]:
-        """Run videos through the embeddings and add to the vectorstore.
+    ) -> list[Tuple[Document, float]]:
+        """Return docs and their similarity scores on a scale from 0 to 1."""
+        if self.override_relevance_score_fn is not None:
+            kwargs["normalize_distance"] = False
+        # else:
+        #     kwargs["normalize_distance"] = True
```
ccurme, line 532: Should we handle the `else` condition here?

libs/community/langchain_community/vectorstores/vdms.py
```diff
@@ -171,586 +129,664 @@ class VDMS(VectorStore):
...
+        vdms_store._len_check_if_sized(ids, texts, "ids", "texts")
+        vdms_store._len_check_if_sized(ids, metadatas, "ids", "metadatas")
 
-        inserted_ids = self.add_from(
-            texts=texts,
-            embeddings=embeddings,
-            ids=ids,
-            metadatas=metadatas,
-            batch_size=batch_size,
+        # Remove IDs if exist-TEST_REMOVAL
```
ccurme, line 727: What is meant by `-TEST_REMOVAL`?

ccurme (commented):
@cwlacewe I'm going to close this for now with the recommendation that you / Intel implement the VDMS vector store in a new package (e.g., `langchain-vdms` or `langchain-intel`).

We've written a walkthrough on this process here and are happy to answer your questions as you go through it:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the code, and you have control over versioning.

I think this route is preferable to reviewing a large code change (thousands of lines), only to have the implementation deprecated if/when the standalone package is released.

Let me know what you think!

cwlacewe:
@ccurme is there a way to get this one integrated as we have customers waiting and we can work on the package for new year?

cwlacewe:
Hi @ccurme, yes the existing implementation in langchain-community is functional. This PR includes a new class to centralize the private functions.  Are these the breaking changes you're referring to or was it something else?

PR #28444, opened 2024-12-02
docs: update "language" model to "chat" model and added hyperlink to …

DangerousPotential:
Edited "language model" to "chat model" and added hyperlink to the relevant concept page.


baskaryan (commented):
technically prompt templates can be used for either LLMs (text completion models) or chat models

Issue #28450, opened 2024-12-02
DOC: Managing Conversation History


CC-KEH:
### URL

https://python.langchain.com/docs/tutorials/chatbot/#managing-conversation-history

### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

![image](https://github.com/user-attachments/assets/300c8115-57b4-4241-b659-8fed707c9691)

**Problem in the Example:**  
- **query1**: `"What is my name?"`
  thread_id: `abc567`
- **query2**: `"What math problem did I ask?"`
  thread_id: `abc678`

The change in `thread_id` is incorrect, as both queries should belong to the same conversation.

**Intended Behavior to Demonstrate:**  
- After trimming the chat history, the context for `query1` is discarded. The model cannot answer it.  
- The context for `query2` is retained, so the model can still respond to it.

**Suggested Fix:**  
Use the same `thread_id` for both queries to accurately reflect context retention within a single thread.

asdkfjsd:
Hello, I am interested in taking on this task. However, I have a couple of questions to ensure I understand the scope and requirements correctly.

Is it sufficient to only modify this particular text and unify the thread_id ( "abc567" and "abc678" ) across the context?
Do I need to check all similar pages across the documentation to address this issue, or is the change limited to a specific page or section?

I am eager to assist with this task and would appreciate any guidance you can provide, thank you very much.

efriis:
hey! just ran the existing code, and it's working as expected. if we use the same thread id, then it'll duplicate some of the messages, so the intention is actually to create a new thread (with the same conversation history in `messages`)

PR #28470, opened 2024-12-03
docs: Correct thread_id consistency for conversation history example

asdkfjsd:

**Description:** Updated the documentation to ensure that both queries in the conversation history example share the same thread_id, reflecting a single conversation context. This change aligns with the intended behavior where the model should discard the context for query1 after trimming the chat history and retain the context for query2. 

**Issue:** DOC: Managing Conversation History #28450

**Dependencies:** No additional dependencies required for this change.

### DIFF
--- a/docs/docs/tutorials/chatbot.ipynb
+++ b/docs/docs/tutorials/chatbot.ipynb
@@ -847,7 +847,7 @@
     }
    ],
    "source": [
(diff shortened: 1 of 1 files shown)


efriis:
hey! just ran the existing code, and it's working as expected. if we use the same thread id, then it'll duplicate some of the messages, so the intention is actually to create a new thread (with the same conversation history in `messages`)

Issue #28476, opened 2024-12-03
[Bug] Azure cosmos db no sql vector store similarity search method "mmr"


wassim-mechergui-shift:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python

from langchain_openai import AzureOpenAIEmbeddings
from langchain_community.vectorstores.azure_cosmos_db import AzureCosmosDBVectorSearch

azure_endpoint = "here_goes_azure_endpoint" 
openai_api_version = "2023-12-01-preview"
openai_api_key = "here_goes_api_key" 

embeddings = AzureOpenAIEmbeddings(
                    model="text-embedding-ada-002",
                    deployment="text-embedding-ada-002",
                    azure_endpoint=azure_endpoint,
                    openai_api_version=openai_api_version,
                    openai_api_key=openai_api_key
                )

azure_cosmos_db_connection_string = "here_goes_connection string" 
namespace = "here_goes_namespace" 
index_name = "here_goes_index_name" 

vs = AzureCosmosDBVectorSearch.from_connection_string(azure_cosmos_db_connection_string,
                                                                  namespace, embeddings,
                                                                  index_name=index_name)
docs_with_scores = vs.max_marginal_relevance_search(query, k=8)
```

### Error Message and Stack Trace (if applicable)

```txt 

File \site-packages\langchain_community\vectorstores\azure_cosmos_db_no_sql.py:363, in AzureCosmosDBNoSqlVectorSearch.max_marginal_relevance_search_by_vector(self, embedding, k, fetch_k, lambda_mult, **kwargs)
    [361](/langchain_community/vectorstores/azure_cosmos_db_no_sql.py:361) if kwargs["with_embedding"]:
    [362](/langchain_community/vectorstores/azure_cosmos_db_no_sql.py:362)     with_embedding = kwargs["with_embedding"]
--> [363](/langchain_community/vectorstores/azure_cosmos_db_no_sql.py:363) docs = self._similarity_search_with_score(
    [364](/langchain_community/vectorstores/azure_cosmos_db_no_sql.py:364)     embeddings=embedding,
[...]

Issue #26097, opened 2024-09-05
Azure Cosmos DB  _similarity_search_with_score fails with KeyError: 'metadata'


nataliia-obraztsova:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python

from langchain_openai import AzureOpenAIEmbeddings
from langchain_community.vectorstores.azure_cosmos_db import AzureCosmosDBVectorSearch

azure_endpoint = "here_goes_azure_endpoint" 
openai_api_version = "2023-12-01-preview"
openai_api_key = "here_goes_api_key" 

embeddings = AzureOpenAIEmbeddings(
                    model="text-embedding-ada-002",
                    deployment="text-embedding-ada-002",
                    azure_endpoint=azure_endpoint,
                    openai_api_version=openai_api_version,
                    openai_api_key=openai_api_key
                )

azure_cosmos_db_connection_string = "here_goes_connection string" 
namespace = "here_goes_namespace" 
index_name = "here_goes_index_name" 

vs = AzureCosmosDBVectorSearch.from_connection_string(azure_cosmos_db_connection_string,
                                                                  namespace, embeddings,
                                                                  index_name=index_name)
docs_with_scores = vs.similarity_search_with_score(query, k=8)
```

### Error Message and Stack Trace (if applicable)

  File "/.../utils/index_manager.py", line 187, in similarity_search
    docs_with_scores = vs.similarity_search_with_score(query, k=k_similar_vectors)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../lib/python3.11/site-packages/langchain_community/vectorstores/azure_cosmos_db.py", line 541, in similarity_search_with_score
    docs = self._similarity_search_with_score(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../lib/python3.11/site-packages/langchain_community/vectorstores/azure_cosmos_db.py", line 463, in _similarity_search_with_score
    metadata = document_object_field.pop("metadata")
[...]

PR #28479, opened 2024-12-03
[Community][fix] Fix Azure cosmos db no SQL similarity search with score and mmr 

wassim-mechergui-shift:
**Issues**:
* similarity search with score not working for `VectorStoreRetriever` instanciated from `AzureCosmosDBNoSqlVectorSearch` because it requires an implementation of  `__select_relevance_score_fn_`. 
* mmr retrieval is not working because the fetching of metadata field from item to document metadata is currently bugged : (issues: https://github.com/langchain-ai/langchain/issues/26097, https://github.com/langchain-ai/langchain/issues/28476)

**Changes**:
* VectorStoreRetriever fix:
  *  For the similarity search from `VectorStoreRetriever` a rather typical implementation was added ([example for Qdrant](https://github.com/langchain-ai/langchain/blob/ff675c11f6814e88aa338dc3668aa84d1ade3d19/libs/community/langchain_community/vectorstores/qdrant.py#L1928))
* with_embeddings to similarity search:
  * Fix for the with_embedding keyword) due to a change made here (https://github.com/langchain-ai/langchain/issues/26097), item no longer has `self._embedding_key` and has "$1" (that represents it's order in the parameters). Since Cosmos DB’s SQL API does not support parameterized aliases or dynamic aliasing within the query, we chose to name it 'embeddingKey'
  * optimization : if with_embedding is not True, we don't need to query the embeddings. Querying the embeddings is heavy and makes requests much slower. This was changed
* Changed initialisation of vector policy (from which we get the embedding key and the distance function)
  * if user specifies a dict, we check it has the correct form
  * if user specifies the container creation if it doesn't exist option, we check this is not null
  * for the final user vector policy property
    *  if the container exposes one, we use that one  (and raise a warning if the user specified policy has differences)
    * if the container doesn't expose one, we use the user exposed one (with an error in case nothing was specified)

These changes were tested locally.


libs/community/langchain_community/vectorstores/azure_cosmos_db_no_sql.py
```diff
@@ -121,6 +121,9 @@ def __init__(
         self._embedding_key = self._vector_embedding_policy["vectorEmbeddings"][0][
             "path"
         ][1:]
+        self._distance_strategy = self._vector_embedding_policy["vectorEmbeddings"][0][
+            "distanceFunction"
```
baskaryan, line 125: is this guaranteed to always be specified? seems like we only check that self._vector_embedding_policy isn't empty if self._create_container is True

wassim-mechergui-shift: Actually, this assumption is not always valid 
> Once created, it's possible to query for these info in the properties of the container. (through container.read())

It seems there is a bug in azure cosmos db no sql but if the container was the result of a partition_key migration then it seems the information on the vector policy is lost (so not possible to read through `container.read()`). However, the container still fully supports vector search !

I'll readapt the implementation to account for this case.

I'm sorry for the long post I wanted to be thorooughly descriptive and argumentative in my implementation choices!

ccurme (commented):
@wassim-mechergui-shift it looks like there's some redundancy with https://github.com/langchain-ai/langchain/pull/28716.

Closing for now but happy to re-open if there are differences in functionality and you'd like to merge with latest changes (ok to open another PR as well).

wassim-mechergui-shift:
Hi @ccurme ! Thanks for the comment. I didn't notice the other PR.

I will test when I get time from my side if the bugs I mentionned here were addressed in that PR (and hopefully no additional bugs were added). If so I'll merge here with the latest changes

PR #28487, opened 2024-12-03
tests[patch]: populate API reference for chat models

ccurme:
Populate API reference for test class properties and test methods for chat models.

Also `pytest.skip` some tests that were previously passed if features are not supported.

### DIFF
--- a/libs/standard-tests/langchain_tests/integration_tests/chat_models.py
+++ b/libs/standard-tests/langchain_tests/integration_tests/chat_models.py
@@ -295,8 +295,8 @@ class ChatModelIntegrationTests(ChatModelTests):
         .. dropdown:: Configuration
 
             By default, this test is run.
-            To disable this feature, set `returns_usage_metadata` to False in your test
-            class:
+            To disable this feature, set `returns_usage_metadata` to False in your
+            test class:
 
             .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


libs/standard-tests/langchain_tests/integration_tests/chat_models.py
```diff
@@ -74,7 +74,7 @@ def _validate_tool_call_message_no_args(message: BaseMessage) -> None:
 
 class ChatModelIntegrationTests(ChatModelTests):
     @property
-    def standard_chat_model_params(self) -> dict:
+    def _standard_chat_model_params(self) -> dict:
```
efriis, line 77: is there a way to configure this via docstring, or does it have to be by name? just to avoid having to update all the implementations

ccurme: From what I can tell the only integration that implements it is AWS. I added `:meta private:` to docstrings but it still seems to show up in the attributes table. Going with that for now.

I'm actually unclear on what is intended for this property. It is always combined with `chat_model_params`, so my guess was that we were testing a set of standard params (temperature, max_tokens, etc) that we want all models to support, and users should only overwrite `chat_model_params`. If that's correct, then we should update AWS to not override `standard_chat_model_params`.

Otherwise, we should document when/why this should be overwritten (as opposed to specifying things in `chat_model_params`).

PR #28505, opened 2024-12-04
core: add new clean up strategy "scoped_full" to indexing

KeiichiHirobe:
~Note that this PR is now Draft, so I didn't add change to `aindex` function and didn't add test codes for my change.
After we have an agreement on the direction, I will add commits.~

`batch_size` is very difficult to decide because setting a large number like >10000 will impact VectorDB and RecordManager, while setting a small number will delete records unnecessarily, leading to redundant work, as the `IMPORTANT` section says. 
On the other hand, we can't use `full` because the loader returns just a subset of the dataset in our use case.

I guess many people are in the same situation as us.

So, as one of the possible solutions for it, I would like to introduce a new argument, `scoped_full_cleanup`.
This argument will be valid only when `claneup` is Full. If True, Full cleanup deletes all documents that haven't been updated AND that are associated with source ids that were seen during indexing. Default is False.

This change keeps backward compatibility.


libs/core/langchain_core/indexing/api.py
```diff
@@ -215,10 +216,6 @@ def index(
      are not able to specify the uid of the document.
 
     IMPORTANT:
-       * if auto_cleanup is set to True, the loader should be returning
```
KeiichiHirobe, line 218: BTW, `auto_cleanup` doesn't exist now.

eyurtsev: This looks incorrectly documented due to a refactor in names.

`auto_cleanup` sounds like `full` mode?

libs/core/langchain_core/indexing/api.py
```diff
@@ -259,6 +256,11 @@ def index(
                        specify a custom vector_field:
                        upsert_kwargs={"vector_field": "embedding"}
             .. versionadded:: 0.3.10
+        scoped_full_cleanup: This argument will be valid only when `claneup` is Full.
```
eyurtsev, line 259: How about we turn this into another clean up mode so we don't introduce another parameter that only works conditionally?

This looks like `cleanup == 'full_scoped'` or something like that.

And we'd document:

1. In the important section -- that full_scoped is a solution to problem w/ batches being a best effort
2. This solution keeps track of the source ids in memory (probably fine for most use cases in terms of memory consumption) -- would require parallelizing for 10M+ docs anyway

eyurtsev (commented):
This looks very reasonable!

If you'd like to merge:

1. Could you introduce this as a new cleanup mode?
2. Add unit-tests

libs/core/langchain_core/indexing/api.py
KeiichiHirobe: Agree. I will change it to `full` later.

libs/core/langchain_core/indexing/api.py
```diff
@@ -245,6 +250,9 @@ def index(
                     during this run of indexing.
                     Clean up runs after all documents have been indexed.
                     This means that users may see duplicated content during indexing.
+            - Scoped_Full: Similar to Full, but only deletes all documents
```
eyurtsev, line 253:
```suggestion
            - scoped_full: Similar to Full, but only deletes all documents
```

libs/core/langchain_core/indexing/api.py
```diff
@@ -475,10 +492,23 @@ async def aindex(
...
+         supposed to.
+       * In incremental mode, if documents associated with a particular
+         source id appear across different batches, the indexing API
+         will do some redundant work. This will still result in the
+         correct end state of the index, but will unfortunately not be
+         100% efficient. For example, if a given document is split into 15
+         chunks, and we index them using a batch size of 5, we'll have 3 batches
+         all with the same source id. In general, to avoid doing too much
+         redundant work select as big a batch size as possible.
+       * scoped_full mode may be a good option if you have difficulty decidng
```
eyurtsev, line 507:
```suggestion
       * scoped_full mode may be a good option if you have difficulty deciding
```

KeiichiHirobe:
@eyurtsev Thank you for your review. I've just pushed 1 commit: https://github.com/langchain-ai/langchain/pull/28505/commits/32c3f03a89fe2fb5de780ccf78d92f5d11982647

eyurtsev:
Pushed a minor doc-string update

eyurtsev (approved):

eyurtsev:
@KeiichiHirobe thank you looks great!

PR #28514, opened 2024-12-04
community: add trust_env at web_base_loader

dongs0104:
- **Description:** I am working to address a similar issue to the one mentioned in https://github.com/langchain-ai/langchain/pull/19499. Specifically, there is a problem with the Webbase loader used in open-webui, where it fails to load the proxy configuration. This PR aims to resolve that issue.


docs/docs/integrations/document_loaders/web_base.ipynb
```diff
@@ -66,7 +66,10 @@
    "source": [
     "from langchain_community.document_loaders import WebBaseLoader\n",
     "\n",
-    "loader = WebBaseLoader(\"https://www.espn.com/\")"
+    "loader = WebBaseLoader(\"https://www.espn.com/\")\n",
```
hwchase17, line 69: i dont think this should be documented here, it should just be prevalent in the reference docs

libs/community/langchain_community/document_loaders/web_base.py
```diff
@@ -128,6 +128,7 @@ def __init__(
         session: Any = None,
         *,
         show_progress: bool = True,
+        trust_env: bool = False,
```
hwchase17, line 131: needs documentation of this in the docstring so it shows up in api reference

ccurme (approved):

PR #28516, opened 2024-12-04
chroma[patch]: add `get_by_ids` and fix bug

ccurme:
- Run standard integration tests in Chroma
- Add `get_by_ids` method
- Fix bug in `add_texts`: if a list of `ids` is passed but any of them are None, Chroma will raise an exception. Here we assign a uuid.


ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain/actions/runs/12166370920/job/33932456500

PR #28574, opened 2024-12-06
Switch graphqa example in docs to langgraph

tomasonjo:


baskaryan:
why delete the how-tos?

tomasonjo:
They are now part of the langgraph example

V pon., 9. dec. 2024, 02:32 je oseba Bagatur ***@***.***>
napisala:

ccurme:
@tomasonjo thanks very much for this. I think we should either:

1. Keep the how-to guides;
2. Add re-directs to specific areas of the tutorial via Vercel: https://github.com/langchain-ai/langchain/blob/6e6061fe738323d8cd3e1d1702affdfb974118f9/docs/vercel.json#L57-L60

Is it easy to do the mapping in (2)?

tomasonjo:
The other two examples are for the "deprecated" CypherQAChain, which we mention in the main docs, but say it's much better to use langgraph. In the langraph we add steps on dynamic fewshot examples to text2cypher, do relationship direction validation, and also mapping to database values. So separate docs without the context of the whole langgraph might not be the best, and I also am not a fan of promoting CypherQAChain and how to optimize it. Just use langgraph

PR #28587, opened 2024-12-06
Fixes to _merge in core utils when e.g. index changes in token count response. It'l…

stenvala:
Description:

Merging fails when there is integer, or float, where value changes.

Issue: For example in streaming chat bedrock there is case where `right` equals to 
```
{'inputTextTokenCount': None, 'completionReason': None, 'totalOutputTextTokenCount': None, 'index': 1}
```
and `merged` to 
```
{'inputTextTokenCount': None, 'completionReason': None, 'totalOutputTextTokenCount': None, 'index': 0}
```
In this case error is thrown. There is currently no reported bug on this, but this is happening.

Dependencies: N/A


libs/core/tests/unit_tests/utils/test_utils.py
```diff
@@ -111,10 +111,18 @@ def test_check_package_version(
             {"a": [{"idx": 0, "b": "f"}]},
             {"a": [{"idx": 0, "b": "{"}, {"idx": 0, "b": "f"}]},
         ),
+        # int replacing
+        (
+            {"index": 1, "b": "a"},
+            {"index": 5, "b": "b"},
+            {"index": 5, "b": "ab"},
```
baskaryan, line 118: mm is this always the right behavior? if two things have different indexes then presumably they shouldn't be merged?

in general its not clear how to "merge" ints, which is why its not supported. certain ints should be added (eg incremental token counts), certain ints should be max/min-ed (eg timestamps), certain ones should be replaced (eg total token counts). there's no general way to handle ints

stenvala: This is good question. Very hard to say, but now the streaming pi with Bedrock fails due to this issue that index changes.

stenvala: I'm new here and don't udnerstand why CI fails (test-doc-imports skipped, perhaps normal) and Vercel deployment failed. I don't know what this means.

ccurme:
@stenvala thanks for this. Would you mind linking an issue or PR in `langchain-aws` that includes a minimal reproducible example of the problem you are encountering? Then we can devise an appropriate solution, as it's not clear that the issue is in `langchain-core`.

PR #28594, opened 2024-12-06
tests[patch]: update API ref for chat models

ccurme:


libs/standard-tests/langchain_tests/integration_tests/chat_models.py
```diff
@@ -145,6 +145,14 @@ def chat_model_params(self) -> dict:
         By default, this is determined by whether the chat model's `bind_tools` method
         is overridden. It typically does not need to be overridden on the test class.
 
+        Example override:
```
efriis, line 148: nice explicit

libs/standard-tests/langchain_tests/integration_tests/chat_models.py
```diff
@@ -1323,8 +1331,8 @@ def has_tool_calling(self) -> bool:
             .. code-block:: python
 
                 @pytest.mark.xfail(reason=("Not implemented."))
-                def test_tool_message_histories_string_content(self, model: BaseChatModel) -> None:
-                    super().test_tool_message_histories_string_content(model)
+                def test_tool_message_histories_string_content(self, *args: Any) -> None:
```
efriis, line 1334: smart - should we pass **kwargs too? Just in case we add kwargs for some reason in future

efriis (approved):

PR #28605, opened 2024-12-07
core[patch]: return ToolMessage from tool

baskaryan:


libs/core/langchain_core/tools/base.py
```diff
@@ -954,10 +972,31 @@ class InjectedToolArg:
     """Annotation for a Tool arg that is **not** meant to be generated by a model."""
 
 
-def _is_injected_arg_type(type_: type) -> bool:
+class InjectedToolCallId(InjectedToolArg):
```
vbarda, line 975: could we export this one from `langchain_core.tools`?

nfcampos (approved):

libs/core/langchain_core/messages/tool.py
```diff
@@ -9,7 +9,11 @@
 from langchain_core.utils._merge import merge_dicts, merge_obj
 
 
-class ToolMessage(BaseMessage):
+class ToolOutputMixin:
```
eyurtsev, line 12: Could you define what returning directly means in the doc-string?

libs/core/langchain_core/tools/base.py
```diff
@@ -512,9 +514,39 @@ def _parse_input(self, tool_input: Union[str, dict]) -> Union[str, dict[str, Any
...
+                            if tool_call_id is None:
+                                msg = (
+                                    "When tool includes an InjectedToolCallId "
+                                    "argument, tool must always be invoked with a full "
+                                    "model ToolCall of the form: {'args': {...}, "
+                                    "'name': '...', 'type': 'tool_call', "
+                                    "'tool_call_id': '...'}"
+                                )
+                                raise ValueError(msg)
+                            tool_input[k] = tool_call_id
```
nielsbox, line 531: Writing to tool_input seems to indirectly break with_structured_output. Can you do a deepcopy of the tool_input before modifying it?

Issue #28579, opened 2024-12-06
ValueError: Self query retriever with Vector Store type <class 'langchain_weaviate.vectorstores.WeaviateVectorStore'> not supported.


RuiDevAltar:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
import weaviate
from langchain.schema import Document
from langchain_weaviate.vectorstores import WeaviateVectorStore
from langchain.retrievers.self_query.base import SelfQueryRetriever
from langchain.chains.query_constructor.base import AttributeInfo
from weaviate.classes.init import Auth

client = weaviate.connect_to_weaviate_cloud(
    cluster_url=weaviate_url,
    auth_credentials=Auth.api_key(weaviate_api_key),
)

documents = [
    Document(
        page_content="Weavite is a powerful vector database for embeddings.",
        metadata={"author": "John Doe", "category": "Technology"}
    ),
    Document(
        page_content="Python is a versatile programming language used for various applications.",
        metadata={"author": "Jane Smith", "category": "Programming"}
    ),
    Document(
        page_content="Machine learning enables computers to learn from data.",
        metadata={"author": "Alice Johnson", "category": "AI"}
    )
]

db_from_docs = WeaviateVectorStore.from_documents(documents, embeddings, client=client)

pdf_metadata_field_info = [
    AttributeInfo(name="source", description="The PDF file the chunk is from", type="string"),
    AttributeInfo(name="page", description="The page number from the PDF", type="integer")
]

search_kwargs = {"k": 6}

data_retriever = SelfQueryRetriever.from_llm(
    llm, db_from_docs, "Company information from PDF files", pdf_metadata_field_info, search_kwargs=search_kwargs
)
```

### Error Message and Stack Trace (if applicable)

File /opt/conda/lib/python3.11/site-packages/langchain/retrievers/self_query/base.py:333, in SelfQueryRetriever.from_llm(cls, llm, vectorstore, document_contents, metadata_field_info, structured_query_translator, chain_kwargs, enable_limit, use_original_query, **kwargs)
[...]

PR #28608, opened 2024-12-07
Import for `WeaviateVectorStore` from langchain_weaviate in `_get_builtin_translator`

keenborder786:
- **Description:** The `WeaviateVectorStore` from `langchain_weaviate` was missing in `_get_builtin_translator` which have been added in this PR.
  - **Issue:** #28579


libs/langchain/langchain/retrievers/self_query/base.py
```diff
@@ -94,6 +94,7 @@ def _get_builtin_translator(vectorstore: VectorStore) -> Visitor:
     from langchain_community.vectorstores import (
         Pinecone as CommunityPinecone,
     )
+    from langchain_weaviate.vectorstores import WeaviateVectorStore
```
baskaryan, line 97: can we conditionally import the way we do with elasticsearch in L141

keenborder786: Sure

keenborder786: Done

efriis:
I think this may have been handled in a different PR simultaneously sorry about that! The remaining change seems unrelated. Feel free to reopen with a diff title if important

PR #28615, opened 2024-12-08
[Community]: Fixed `ArxivAPIWrapper` (corrected fitz attribute and not using `load_max_docs`)

keenborder786:
- **Description:** `ArxivAPIWrapper` was not using `load_max_docs` which have been fixed, and also fitz.fitz was raising an attribute error which has also been fixed in this PR.


keenborder786:
@ccurme

hwchase17:
this should use load_max_documents in lazy_load - otherwise, we do the work of loading all the docs and filter after, thats not very efficient

Also - fitz.fitz works with older versions of pymupdf. we need to fix dependencies

PR #28643, opened 2024-12-10
core: add yaml friendly dict as message-like for chat prompt templates

ianchi:
Current way of defining ChatPromptTemplate is not friendly to have them stored in local files as yaml. You can do it as a list of lists, which doesn't look good or intuitive.

This PR adds a variation that works nicely for that and doesn't diverge from current way.

#### File template

```yaml
# prompt_tamplate.yaml

- system: |
    You are a helpful AI bot. Your name is {name}.
- placeholder: chat_history
- human: |
    {user_input}
```
#### Code
```python

messages = yaml.safe_load(file)
template = ChatPromptTemplate(messages)
```

Main divergence is that placeholder variables are without curly braces to avoid forcing quotes on the string, as curly braces have meaning in yaml.

### DIFF
--- a/libs/core/langchain_core/prompts/chat.py
+++ b/libs/core/langchain_core/prompts/chat.py
@@ -829,6 +829,7 @@ MessageLikeRepresentation = Union[
         Union[str, type],
         Union[str, list[dict], list[object]],
     ],
+    dict[str, Union[str, dict[str, Union[str, bool]], list[Union[str, bool]]]],
     str,
 ]
 
@@ -1445,6 +1446,8 @@ def _convert_to_message(
     - 2-tuple of (role string, template); e.g., ("human", "{user_input}")
     - 2-tuple of (message class, template)
     - string: shorthand for ("human", template); e.g., "{user_input}"
+    - dict with a single key with the role and template as value;
+         e.g., {"human": "{user_input}"}
 
     Args:
         message: a representation of a message in one of the supported formats.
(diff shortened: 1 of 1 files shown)


efriis (commented):
hey! we'd prefer to stick with just `("human", "content")` rather than introduce yet another one like `{"human": "message"}`. Any reason this is necessary in particular?

ianchi:
Hi @efriis, the idea behind this format is purely to have a way to keep templates in a separate file, on their own, for easier versions control and sharing.
To be easy to read and edit by humans the format is yaml. It has great integration with IDEs and tooling and syntax highlighting.

I definitely don’t want templates in the middle of python code. If you only consider Python code, then I agree that the current format is enough.

But the current format is not intuitive, and doesn’t look good in editors:

### Proposed

```yaml
# prompt_tamplate.yaml

- system: |
    You are a helpful AI bot. Your name is {name}.
- placeholder: chat_history
- human: |
    {user_input}
```

### Current

```yaml
# prompt_tamplate.yaml

- 
  - system
  - You are a helpful AI bot. Your name is {name}.
- 
  - placeholder
  - “{chat_history}”
- 
  - human
  - “{user_input}”
```

I think it is quite apparent that the proposed one is much cleaner and easier to read.
Besides it enables to build a schema so that IDEs help in editing an validation 

I already use the proposed template yaml, with an intermediate function to do the conversion.
I thought others might also find the format useful and so shared an adaptation of that function in this PR.

efriis:
got it - would recommend writing some code that parses your yaml file into tuples! something like 
``` messages = [next(v.keys()): next(v.values()) for v in my_dict_list]```

then you can store on disk and make the messages work throughout langchain! introducing a new format like this PR does involves more work than this PR does, and I think this is better achieved in user code like above.

Thanks!

Issue #28634, opened 2024-12-09
Web base loader's fetch pages does not provide way to assign query parameters


bluearrow98:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python 
from langchain_community.document_loaders import WebBaseLoader


web_loader = WebBaseLoader()
pages = await web_loader.fetch_all(urls) # urls being a list of webpages link

```

### Description

The source code assumes that fetching the pages requires no authentication and/or query parameters. This is highly unlikely in practical scenarios where page contents are often protected. The scrape function seems to be using requests_kwargs argument, but the same functionality is also needed for fetch. Currently the source code of fetch function is:

```python 
async def _fetch(
        self, url: str, retries: int = 3, cooldown: int = 2, backoff: float = 1.5
    ) -> str:
        async with aiohttp.ClientSession() as session:
            for i in range(retries):
                try:
                    kwargs: Dict = dict(
                        headers=self.session.headers,
                        cookies=self.session.cookies.get_dict(),
                    )
                    if not self.session.verify:
                        kwargs["ssl"] = False
                    async with session.get(url, **kwargs) as response:
                        if self.raise_for_status:
                            response.raise_for_status()
                        return await response.text()
                except aiohttp.ClientConnectionError as e:
                    if i == retries - 1:
                        raise
                    else:
                        logger.warning(
                            f"Error fetching {url} with attempt "
                            f"{i + 1}/{retries}: {e}. Retrying..."
                        )
                        await asyncio.sleep(cooldown * backoff**i)
        raise ValueError("retry count exceeded")
```
the additional kwargs only takes in headers and cookies, but not parameters or authentication variables. This should be a necessary basic requirement. 

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 22.6.0: Mon Feb 19 19:43:41 PST 2024; root:xnu-8796.141.3.704.6~1/RELEASE_ARM64_T8103
> Python Version:  3.12.7 (main, Oct  1 2024, 02:05:46) [Clang 15.0.0 (clang-1500.1.0.2.5)]

Package Information
-------------------
> langchain_core: 0.3.7
[...]

PR #28646, opened 2024-12-10
[Community]: `requests_kwargs` not being used in _fetch

keenborder786:
- **Description:** `requests_kwargs` is not being passed to `_fetch` which is fetching pages asynchronously.  In this PR, making sure that we are passing `requests_kwargs` to `_fetch` just like `_scrape`.
- **Issue:** #28634


efriis (approved):
made sure to merge them to avoid errors if they have the same key

PR #28652, opened 2024-12-10
Refactoring all PDF loader and parser

pprados:
WIP


efriis:
Hey @pprados! I don't think this work is going to result in a PR that is reviewable if it's only partially done and already adding 7000 lines.

What is your goal in this work?

efriis:
ah got it - is there an issue or discussion of proposed changes? It might be easier to discuss ideas than these code changes

pprados:
@efriis 
90% of my customers work with PDF files and don't have a satisfactory solution at the moment. They cobble together solutions outside langchain (pdf processing outside loaders/parsers), sacrificing a good part of the benefits of this framework. Seeing the same problems over and over again, and the same bad solutions, I couldn't let them go on like that. I had to deal with the problem in the best possible way, for my customers and all LangChain users.

efriis:
will let you and eugene discuss in the time you have scheduled.

efriis:
hey @pprados! Would it be ok if we closed this until it was ready for review in january? If you need a PR for tracking changes while you're working, would recommend opening one against your own fork's master branch

Let me know! We're cleaning up our review process in target of faster review times and generally improved contributor experiences, and one of the aspects of it is trying to keep the inbox clear.

PR #28659, opened 2024-12-11
community[patch]: fix QuerySQLDatabaseTool name

baskaryan:


ccurme (approved):
Should we update `langchain` as well in this PR?

efriis (approved):

efriis:
@ccurme as in the backwards compat stubs? I care less there. Game for either. This seems like a good change in community regardless

Issue #24545, opened 2024-07-23
Chroma - wrong relevance scores.


Saniel0:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from typing import List, Tuple
from langchain_community.vectorstores import Chroma
from langchain_core.documents import Document

from my_embeddings import my_embeddings

vectorStore = Chroma(
    collection_name="products",
    embedding_function=my_embeddings
    persist_directory="./database",
)

# these two functions should give the same result, but the relevance scores are different
def get_similar_docs1(sku: str, count: int) -> List[Tuple[Document, float]]:
    base_query = vectorStore.get(ids=sku).get("documents")[0]
    return vectorStore.similarity_search_with_relevance_scores(query=base_query, k=(count + 1))[1:]

def get_similar_docs2(sku: str, count: int) -> List[Tuple[Document, float]]:
    base_vector = vectorStore.get(ids=sku, include=["embeddings"]).get("embeddings")[0]
    return vectorStore.similarity_search_by_vector_with_relevance_scores(embedding=base_vector, k=(count + 1))[1:]
```

### Description

I am writing a function, that finds ```count``` number of the most simillar document to the document with id ```sku```. 

I started with the first function and it works as expected. I then tried to rewrite the function, so that it retrieves the embedding vector so I do not have to calculate it again. This returns same documents as the first function (also in the same order), but the relevance scores are completely different. Firstly, it seems that the most relevant result now has the lowest relevance score, but even if I do ```(1 - score)``` I do not get the same score as in the first function.

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  #38-Ubuntu SMP PREEMPT_DYNAMIC Fri Jun  7 15:25:01 UTC 2024
> Python Version:  3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0]

Package Information
-------------------
> langchain_core: 0.2.13
> langchain: 0.2.7
> langchain_community: 0.2.7
> langsmith: 0.1.85
> langchain_huggingface: 0.0.3
> langchain_text_splitters: 0.2.2
> langchainhub: 0.1.20
> langgraph: 0.1.7

[...]

PR #28669, opened 2024-12-11
community: rename method 'similarity_search_by_vector_with_relevance_scores' for clarity

sorbose:
**Description:**

In the current version, the similarity_search_by_vector_with_relevance_scores method in langchain_chroma.vectorstores.Chroma returns scores where "lower score represents more similarity". However, the similarity_search_with_relevance_scores method in the parent class langchain_core.vectorstores.base.VectorStore returns scores where "0 is dissimilar, and 1 is most similar" ([source](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores/base.py#L538)).

This discrepancy between the two methods' score conventions can confuse developers and lead to errors. Developers may mix up the relevance_scores in vector and text searches, resulting in opposite outcomes.

To resolve this issue and avoid confusion, I propose renaming Chroma’s similarity_search_by_vector_with_relevance_scores method to similarity_search_by_vector_with_scores. This change aligns with the scoring convention of the similarity_search_with_score method in the VectorStore class ([source](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores/base.py#L450)), where the score represents distance.

**Issue:**
https://github.com/langchain-ai/langchain/issues/24545

**Dependencies:**
No new dependencies

**Twitter handle:** N/A

### DIFF
--- a/libs/community/langchain_community/vectorstores/chroma.py
+++ b/libs/community/langchain_community/vectorstores/chroma.py
@@ -377,7 +377,7 @@ class Chroma(VectorStore):
         )
         return _results_to_docs(results)
 
-    def similarity_search_by_vector_with_relevance_scores(
+    def similarity_search_by_vector_with_scores(
         self,
         embedding: List[float],
         k: int = DEFAULT_K,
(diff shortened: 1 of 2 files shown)


efriis:
unfortunately this is a breaking change, so we can't do this type of thing until 1.0 :(

PR #28670, opened 2024-12-11
community: fixed bug in model2vec embedding code

Pringled:
This PR fixes a bug with the current implementation for Model2Vec embeddings where `embed_documents` does not work as expected.

- **Description**: the current implementation uses `encode_as_sequence` for encoding documents. This is incorrect, as `encode_as_sequence` creates token embeddings and not mean embeddings. The normal `encode` function handles both single and batched inputs and should be used instead. The return type was also incorrect, as encode returns a NumPy array. This PR converts the embedding to a list so that the output is consistent with the Embeddings ABC.


efriis (approved):
thanks!

https://github.com/MinishLab/model2vec/blob/main/model2vec/model.py#L230

PR #28695, opened 2024-12-12
core: improved method tools

### LINKED ISSUE #27471: make sure that @tool decorator can be applied to a method (regression?)

ethanglide:
Add functionality for transforming methods and classmethods into tools. This is an improvement over the implementation in https://github.com/langchain-ai/langchain/pull/28160.

Resolves https://github.com/langchain-ai/langchain/issues/27471

Example:
```py
class A:
    def __init__(self, c: int):
        self.c = c

    @MethodTool
    def foo(self, a: int, b: int) -> int:
        """Add two numbers to c."""
        return a + b + self.c

a = A(10)
a.foo.invoke({"a": 1, "b": 2}) # 13
```

Example with `classmethod`:
```py
class A:
    c = 10
  
    @MethodTool
    @classmethod
    def foo(cls, a: int, b: int) -> int:
        """Add two numbers to c."""
        return a + b + cls.c

A.foo.invoke({"a": 1, "b": 2}) # 13
```

This implementation requires the programmer to specify `@MethodTool` rather than using the `@tool` decorator, which allows for a much better implementation which gets over a lot of shortcomings of my previous attempt in https://github.com/langchain-ai/langchain/pull/28160. Some improvements are as follows:

- The first parameter of the method is not required to be named `self` or `cls`, it can be any name
- The argument name `outer_instance` is no longer a "claimed" parameter name. Parameters can be of any name
- `classmethod`s are fully supported


libs/core/langchain_core/tools/__init__.py
```diff
@@ -41,6 +41,7 @@
 from langchain_core.tools.base import (
     create_schema_from_function as create_schema_from_function,
 )
+from langchain_core.tools.convert import MethodTool as MethodTool
```
efriis, line 44:
```suggestion
from langchain_core.tools.convert import MethodTool
```

libs/core/langchain_core/tools/convert.py
```diff
@@ -421,3 +429,15 @@ def invoke_wrapper(callbacks: Optional[Callbacks] = None, **kwargs: Any) -> Any:
             description=description,
             args_schema=args_schema,
         )
+
+
+class MethodTool:
```
efriis, line 434: can we make this a function instead of a class?

libs/core/langchain_core/tools/structured.py
```diff
@@ -41,6 +42,9 @@ class StructuredTool(BaseTool):
     """The function to run when the tool is called."""
     coroutine: Optional[Callable[..., Awaitable[Any]]] = None
     """The asynchronous version of the function."""
+    outer_instance: Optional[Any] = None
```
efriis, line 45: Let's try to avoid storing this if possible. Could we instead set `.func` to a callable that handles the plumbing for the user?

efriis: (in other words, let's not modify `StructuredTool`)

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2174,3 +2175,124 @@ def foo(x: int) -> Bar:
...
+
+
+def test_method_tool_self_ref() -> None:
+    """Test that a method tool can reference self."""
+
+    class A:
+        def __init__(self, c: int):
+            self.c = c
+
+        @MethodTool
```
efriis, line 2187:
```suggestion
        @methodtool
```
or `toolmethod` - @baskaryan do you have a preference?

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2174,3 +2175,124 @@ def foo(x: int) -> Bar:
...
+        def foo(self, a: int, b: int) -> int:
+            """Add two numbers to c."""
+            return a + b + self.c
+
+    a = A(10)
+    assert a.foo.invoke({"a": 1, "b": 2}) == 13
```
efriis, line 2193: good - let's also do some asserts on name/description/args schema

libs/core/tests/unit_tests/test_tools.py
efriis, line 1: good tests. can we add some that call "cls" and "self" something different in the implementation (e.g. `c` and `s`)

libs/core/langchain_core/tools/convert.py
efriis: I see - this maybe solves the issue of typing. going to suggest some ideas - not sure if they work

libs/core/langchain_core/tools/convert.py
```diff
@@ -421,3 +429,15 @@ def invoke_wrapper(callbacks: Optional[Callbacks] = None, **kwargs: Any) -> Any:
...
+    """A descriptor that converts a method into a tool."""
+
+    def __init__(self, func: Union[Callable, classmethod]) -> None:
+        self.func = func
+
+    def __get__(self, instance: Any, owner: Any) -> BaseTool:
+        if isinstance(self.func, classmethod):
+            return tool(self.func.__func__, outer_instance=owner)
+        return tool(self.func, outer_instance=instance)
```
efriis, line 443:
```suggestion
def toolmethod(f: callable) -> property:
    """A decorator that converts a method into a tool."""
    def inner_f(outer_instance) -> BaseTool:
        tool_f = functools.partial(f, outer_instance) # this substitutes in for self
        return tool(tool_f)
            
    return property(inner_f)
```

efriis: this leaves out what was being handled by owner vs instance for `outer_instance` - what is that doing for us?

efriis (commented):
thanks! a few suggestions and questions

libs/core/tests/unit_tests/test_tools.py
ethanglide: I did have some but the linter (I think ruff) was upset at this. Is there any way to get it to ignore this for the sake of having it in a test case?

libs/core/langchain_core/tools/convert.py
ethanglide: That was to allow the support for classmethods. With the way the `__get__` descriptor works `owner` is for the class itself and `instance` is for the instance of the class.

libs/core/tests/unit_tests/test_tools.py
ethanglide: Was getting linter errors for this, which will not be a problem if changing the decorator to a function rather than a class is required.

libs/core/langchain_core/tools/__init__.py
ethanglide: Linter (ruff) requires this redundant alias

libs/core/langchain_core/tools/structured.py
ethanglide: Yes

libs/core/tests/unit_tests/test_tools.py
ethanglide: Will do

libs/core/langchain_core/tools/convert.py
ethanglide: The class is needed for classmethod support. I have created a wrapper function `methodtool` that simply returns `MethodTool(func)`. How's that?

libs/core/tests/unit_tests/test_tools.py
ethanglide: Have added one for this, let me know if it was exhaustive enough

libs/core/langchain_core/tools/convert.py
efriis: why is it needed for classmethod support?

libs/core/tests/unit_tests/test_tools.py
efriis: you can add `#noqa: N804` or `#noqa: N805` for classmethod and method respectively: https://docs.astral.sh/ruff/rules/#pep8-naming-n

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2174,3 +2175,146 @@ def foo(x: int) -> Bar:
...
+        def __init__(self, c: int):
+            self.c = c
+
+        @methodtool
+        def foo(self, a: int, b: int) -> int:
+            """Add two numbers to c."""
+            return a + b + self.c
+
+    a = A(10)
+    assert a.foo.invoke({"a": 1, "b": 2}) == 13
```
efriis, line 2193:
```suggestion
    assert a.foo.invoke({"a": 1, "b": 2}) == 13
    assert self.args_schema == {...}
    assert self.name == "foo"
    assert self.description == "Add two numbers to c."
```

libs/core/tests/unit_tests/test_tools.py
efriis: added as a code suggestion to be clear what this was suggesting - would be great to get these asserted across all cases!

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2174,3 +2175,146 @@ def foo(x: int) -> Bar:
...
+    class A:
+        c = 5
+
+        @methodtool
+        @classmethod
+        def foo(cls, a: int, b: int) -> int:
+            """Add two numbers to c."""
+            return a + b + cls.c
+
+    assert "cls" not in A.foo.args
```
efriis, line 2320:
```suggestion
    assert "cls" not in A.foo.args
    
class A:
    ...

class TestStandardAUnit(BaseToolUnitTests):
    ...
    
class TestStandardAIntegration(BaseToolIntegrationTests):
    ...
```

would also be good to implement some of these as standard tests - it tests a few of the exhaustive cases: https://python.langchain.com/docs/contributing/how_to/integrations/standard_tests/

sorry this is dragging on a bit! appreciate your help!

libs/core/tests/unit_tests/test_tools.py
ethanglide: Thank you for the clarification, will get that done.

efriis:
also before starting too much work @baskaryan had a suggestion for how to document this behavior without needing code changes. He smartly suggested that the underlying issue with decorating methods with anything is that it incorrectly decorates `Class.method` instead of the object's `self.method` instance, so it may be better to just document that instead, with a way of adding a separate property/method that generates that for you if you so wish

ethanglide:
> also before starting too much work @baskaryan had a suggestion for how to document this behavior without needing code changes. He smartly suggested that the underlying issue with decorating methods with anything is that it incorrectly decorates `Class.method` instead of the object's `self.method` instance, so it may be better to just document that instead, with a way of adding a separate property/method that generates that for you if you so wish

Sorry, I don't think I understand. I get that the decorator does decorate `Class.method` instead of `self.method`, but with the way things are currently set up even if you decorated the method within the constructor (which would decorate `self.method`) you would still be unable to use the tool because of the conflicting names of the `self` parameter. There would need to still be code changes.

However, since we know we can create this decorator which does the job we want and works with a very similar interface to the current `@tool` decorator, why not do it?

libs/core/tests/unit_tests/test_tools.py
ethanglide: Thank you! This will be in the next push

libs/core/tests/unit_tests/test_tools.py
ethanglide: This seems cool, I've never heard of this before and would like to give it a try since it could make the tests more concise. But I am having trouble putting them into the `test_tools.py` file. The issue comes from being unable to import the `ToolsUnitTests` class.

baskaryan:
hey @ethanglide, appreciate the pr! i have a strong bias against adding more interfaces that sound/seem very similar and have similar functionality. thoughts on documenting this type of usage of the existing @tool instead of introducing another decorator?
```python
from langchain_core.tools import tool, BaseTool

class A:
    def __init__(self, c: int):
        self.c = c

    def _foo(self, a: int, b: int) -> int:
        """Add two numbers to c."""
        return a + b + self.c
  
    @property
    def foo(self) -> BaseTool:
        return tool(self._foo)

a = A(10)
a.foo.invoke({"a": 1, "b": 2}) # 13
```

efriis:
closing in favor of @baskaryan's suggestion

really appreciate the thought on this ethan!

Issue #28447, opened 2024-12-02
[Bug] [core/indexer] PR #25754 seems like introducing bugs


KeiichiHirobe:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

None

### Error Message and Stack Trace (if applicable)

None

### Description

https://github.com/langchain-ai/langchain/pull/25754 seems like introducing bugs.

Let's say we have 3 records `a`, `b`, `c` with the same source_id, and someone will delete the record `c`.
The indexer should delete a record for `c` from vector database and record manager, but it won't after the PR was merged because `docs_to_index` doesn't include the source_id.

### System Info

None

eyurtsev:
Confirmed

eyurtsev:
@KeiichiHirobe thank you for the simple example!

KeiichiHirobe:
This has been fixed, so I've just closed this

PR #28702, opened 2024-12-13
[core/indexer] Reverts PR #25754 and add unit tests

KeiichiHirobe:
I reported the bug 2 weeks ago here: https://github.com/langchain-ai/langchain/issues/28447

I believe this is a critical bug for the indexer, so I submitted a PR to revert the change and added unit tests to prevent similar bugs from being introduced in the future.

@eyurtsev Could you check this?

### DIFF
--- a/libs/core/tests/unit_tests/indexing/test_indexing.py
+++ b/libs/core/tests/unit_tests/indexing/test_indexing.py
@@ -629,7 +629,6 @@ def test_incremental_delete(
         "This is another document.",
     }
 
-
 def test_incremental_indexing_with_batch_size(
     record_manager: InMemoryRecordManager, vector_store: InMemoryVectorStore
 ) -> None:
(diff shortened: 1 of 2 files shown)


libs/core/tests/unit_tests/indexing/test_indexing.py
```diff
@@ -630,26 +630,18 @@ def test_incremental_delete(
     }
 
 
-def test_incremental_indexing_with_batch_size(
+def test_incremental_delete_with_same_source(
```
KeiichiHirobe, line 633: Unfortunately, the diff in `test_indexing.py` is very confusing, but `test_incremental_delete_with_same_source` is the only test case I newly created.

libs/core/tests/unit_tests/indexing/test_indexing.py
```diff
@@ -688,11 +689,10 @@ def test_incremental_indexing_with_batch_size(
             vector_store,
             cleanup="incremental",
             source_id_key="source",
-            batch_size=2,
         ) == {
             "num_added": 0,
-            "num_deleted": 0,
-            "num_skipped": 4,
+            "num_deleted": 1,
+            "num_skipped": 1,
```
KeiichiHirobe, line 695: I confirmed that the original logic returned `num_deleted` = 0

eyurtsev:
Related to PR: https://github.com/langchain-ai/langchain/pull/25754

PR #28712, opened 2024-12-13
core[patch]: Update error message in indexing code for unreachable code assertion

eyurtsev:
Minor update for error message that should never be triggered


eyurtsev:
cc @KeiichiHirobe

PR #28715, opened 2024-12-13
community: fix an issue with deepinfra integration

JustNawaf:

- [x] **PR title**: langchain: add URL parameter to ChatDeepInfra class

- [x] **PR message**: add URL parameter to ChatDeepInfra class
    - **Description:** This PR introduces a url parameter to the ChatDeepInfra class in LangChain, allowing users to specify a custom URL. Previously, the URL for the DeepInfra API was hardcoded to "https://stage.api.deepinfra.com/v1/openai/chat/completions", which caused issues when the staging endpoint was not functional. The _url method was updated to return the value from the url parameter, enabling greater flexibility and addressing the problem. out!


JustNawaf:
@efriis Could you kindly review this PR?

efriis (approved):
breaking change makes sense in this case

PR #28728, opened 2024-12-15
community: FAISS vectorstore - consistent Document id field

nhols:
make sure id field of Documents in `FAISS` docstore have the same id as values in `index_to_docstore_id`, implement `get_by_ids` method


efriis (approved):

efriis:
thanks!

PR #28740, opened 2024-12-16
partners/voyageai: enable setting output dimension

omriel1:
Voyage has introduced voyage-3-large and voyage-code-3, which feature different output dimensions by leveraging a technique called "Matryoshka Embeddings" (see blog  - https://blog.voyageai.com/2024/12/04/voyage-code-3/).
These two models are available in various sizes: [256, 512, 1024, 2048] (https://docs.voyageai.com/docs/embeddings#model-choices).

This PR adds the option to set the required output dimension.


libs/partners/voyageai/pyproject.toml
```diff
@@ -19,9 +19,9 @@ disallow_untyped_defs = "True"
 "Release Notes" = "https://github.com/langchain-ai/langchain/releases?q=tag%3A%22langchain-voyageai%3D%3D0%22&expanded=true"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<4.0"
+python = ">=3.9,<3.13"
```
ccurme, line 22: What is the status of python 3.13 support before and after this PR?

libs/partners/voyageai/langchain_voyageai/embeddings.py
```diff
@@ -105,22 +107,26 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
...
-            )
-
+            r = self._client.embed(
+                texts[i : i + self.batch_size],
+                model=self.model,
+                input_type="document",
+                truncation=self.truncation,
+                output_dimension=self.output_dimension,
+            ).embeddings
+            embeddings.extend(cast(Iterable[List[float]], r))
```
ccurme, line 117: why are all these casts necessary?

libs/partners/voyageai/langchain_voyageai/embeddings.py
```diff
@@ -37,8 +37,10 @@ class VoyageAIEmbeddings(BaseModel, Embeddings):
     _aclient: voyageai.client_async.AsyncClient = PrivateAttr()
     model: str
     batch_size: int
+
+    output_dimension: Optional[Literal[256, 512, 1024, 2048]] = None
     show_progress_bar: bool = False
-    truncation: Optional[bool] = None
+    truncation: bool = True
```
ccurme, line 43: is changing this default a breaking change for any possible usage of `langchain-voyageai` prior to this PR?

libs/partners/voyageai/langchain_voyageai/embeddings.py
omriel1: As for the newest models (`voyage-3-large` and `voyage-code-3`) they also let you define the `output_dtype` and you can choose for example `float` or `int8`, `binary`. For anything that is not `float` we get an `int`:
1. https://github.com/voyage-ai/voyageai-python/blob/012d16bd708c797766beb26555e56b10db8d09fd/voyageai/object/embeddings.py#L8
2. https://github.com/voyage-ai/voyageai-python/blob/012d16bd708c797766beb26555e56b10db8d09fd/tests/test_client.py#L105

The default (as we use now) outputs `float`, but still I had to make this casting

libs/partners/voyageai/langchain_voyageai/embeddings.py
omriel1: No - it seems was `True` anyway - https://github.com/voyage-ai/voyageai-python/blob/012d16bd708c797766beb26555e56b10db8d09fd/voyageai/client.py#L51

libs/partners/voyageai/pyproject.toml
omriel1: About a month ago, when they moved to `0.3.0` they have changed the dependency from: `^3.7.1` to: `^3.9,<3.13`, but here the requirement where: `>=3.9,<4.0`.
So I wonder how can I approach that, as otherwise it seems we cannot use any version above `0.3.*` version.

ccurme (approved):

omriel1:
Hi!
@ccurme  when do you plan to release this change as part of `langchain-voyageai`? Thanks!

ccurme:
> Hi! @ccurme when do you plan to release this change as part of `langchain-voyageai`? Thanks!

Thanks for the ping, just released in `langchain-voyageai==0.1.4`.

PR #28744, opened 2024-12-16
text-splitter: support for trust_remote_code and not loading model in SentenceTransformersTokenTextSplitter

anchit2000:

- **Description:** Trust remote code parameter was not there due to which we could not load external models, also there is no need to load the model and increase the memory footprint.
- **Dependencies:** Transformers
- **Twitter handle:** @anchit2000

Additional guidelines:

### DIFF
--- a/libs/text-splitters/langchain_text_splitters/sentence_transformers.py
+++ b/libs/text-splitters/langchain_text_splitters/sentence_transformers.py
@@ -1,41 +1,47 @@
 from __future__ import annotations
 
 from typing import Any, List, Optional, cast
-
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Apologies if this is ignorant but isn't changing the underlying library for the tokenization a breaking change?

Also, why are we deleting docstrings in other methods?

Will re-open if needed / I've misunderstood.

anchit2000:
This is important for not downloading and loading the model when it’s not needed while tokenising.

anchit2000:
Also, we can’t use models on the hub where trust remote code parameter is required.

PR #28751, opened 2024-12-16
docs, external: introduce `langchain-localai`

mkhludnev:

Referring to https://github.com/mkhludnev/langchain-localai


efriis (approved):

ccurme:
@mkhludnev we should also update the providers page: https://python.langchain.com/docs/integrations/providers/localai/

mkhludnev:
> @mkhludnev we should also update the providers page: https://python.langchain.com/docs/integrations/providers/localai/

Thanks for reminder, @ccurme! Here we go #28855

PR #28760, opened 2024-12-17
community: support Confluence cookies

soulmachine:
**Description**: Some confluence instances don't support personal access token, then cookie is a convenient way to authenticate. This PR adds support for Confluence cookies.

**Twitter handle**: soulmachine


libs/community/langchain_community/document_loaders/confluence.py
```diff
@@ -152,6 +154,7 @@ def __init__(
         session: Optional[requests.Session] = None,
         oauth2: Optional[dict] = None,
         token: Optional[str] = None,
+        cookies: Optional[dict] = None,
```
ccurme, line 157: Can we make this parameter keyword-only (move below the `*`), in case users are passing in parameters positionally?

soulmachine: Make sense, making `cookies` a keyword-only parameter is a backward compatible way, although semantically `cookies` are far away with authentication related parameters(`api_key`, `oauth2`, `token`)

soulmachine: Done

ccurme (approved):

PR #28785, opened 2024-12-18
langchain_community: Add default None values to DocumentAttributeValue class properties

y1u0d2a1i:
**Description**: 
This PR addresses an issue where the DocumentAttributeValue class properties did not have default values of None. By explicitly setting the Optional attributes (DateValue, LongValue, StringListValue, and StringValue) to default to None, this change ensures the class functions as expected when no value is provided for these attributes.

**Changes Made**:
Added default None values to the following properties of the DocumentAttributeValue class:
DateValue
LongValue
StringListValue
StringValue
Removed the invalid argument extra="allow" from the BaseModel inheritance.
Dependencies: None.

**Twitter handle (optional)**: @__korikori1021

**Checklist**
- [x] Verified that KendraRetriever works as expected after the changes.


ccurme (approved):
Recommend using the [langchain-aws](https://github.com/langchain-ai/langchain-aws) package, where this fix is already implemented.

PR #28786, opened 2024-12-18
langchain_community: Add default value 0.0 to min_score_confidence in AmazonKendraRetriever

y1u0d2a1i:
**Description**:
This PR introduces a default value of 0.0 for the min_score_confidence parameter in the AmazonKendraRetriever class. Previously, min_score_confidence did not have a default value, and it was not documented, requiring developers to read the code to understand how to use the class effectively. This caused confusion and made it difficult to execute AmazonKendraRetriever without prior knowledge of the parameter.

**With this change**:

A default value of 0.0 ensures that all documents, regardless of their confidence score, are included in the results unless a stricter threshold is explicitly specified.
The min_score_confidence parameter is now easier to understand and use, improving the developer experience.

**Rationale**:

Setting a default of 0.0 maintains backward compatibility and aligns with the behavior where no confidence filtering is applied.
Clearly documenting the default behavior ensures developers can utilize AmazonKendraRetriever without needing to dig into the codebase.
Changes Made:

Added min_score_confidence: Annotated[Optional[float], Field(ge=0.0, le=1.0)] = 0.0 to the AmazonKendraRetriever class.
Updated comments to explain the purpose and default value of min_score_confidence.

**Checklist**:
 Verified that AmazonKendraRetriever works as expected with the default value of min_score_confidence=0.0.

### DIFF
--- a/libs/community/langchain_community/retrievers/kendra.py
+++ b/libs/community/langchain_community/retrievers/kendra.py
@@ -355,11 +355,10 @@ class AmazonKendraRetriever(BaseRetriever):
 
         user_context: Provides information about the user context
             See: https://docs.aws.amazon.com/kendra/latest/APIReference
-        
-        min_score_confidence: The minimum confidence score threshold for 
-            filtering documents. Documents with a confidence score 
-            below this value will be excluded from the results. 
 
+        min_score_confidence: The minimum confidence score threshold for
+            filtering documents. Documents with a confidence score
+            below this value will be excluded from the results.
     Example:
         .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this.

Important: the current default is a null value. If this value is used, how does the service behave?

If a field is optional for an API, the desire is often to not override its default and just accept whatever default value the API implements.

If the null default is not valid, I can see the need to fix that.

In any case, if a change is needed, it should be implemented in [langchain-aws](https://github.com/langchain-ai/langchain-aws). I recommend opening an issue or PR there.

PR #28791, opened 2024-12-18
openai: developer message just openai

efriis:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -493,6 +496,20 @@ def build_extra(cls, values: Dict[str, Any]) -> Any:
...
 
+    @model_validator(mode="before")
+    @classmethod
+    def validate_system_message_role(cls, values: Dict[str, Any]) -> Any:
+        """Ensure that the system message role is correctly set for the model."""
+        if "system_message_role" in values:
+            return values
+
+        model = values.get("model_name") or values.get("model") or ""
+        if model.startswith("o1"):
```
eyurtsev, line 507: this change is not future compatible -- it'll force users to update their client when a new model gets released (or update their code w/ system_message_role attribute)

PR #28823, opened 2024-12-19
anthropic: less pydantic for client

efriis:


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -541,9 +541,6 @@ class Joke(BaseModel):
         populate_by_name=True,
     )
 
-    _client: anthropic.Client = PrivateAttr(default=None)  # type: ignore[assignment]
-    _async_client: anthropic.AsyncClient = PrivateAttr(default=None)  # type: ignore[assignment]
-
```
efriis, line 546: these create mypy issues because we want the type to be not-None, but by default it's None

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -677,9 +672,15 @@ def post_init(self) -> Self:
         if self.default_request_timeout is None or self.default_request_timeout > 0:
             client_params["timeout"] = self.default_request_timeout
 
-        self._client = anthropic.Client(**client_params)
-        self._async_client = anthropic.AsyncClient(**client_params)
-        return self
+        return client_params
+
+    @cached_property
+    def _client(self) -> anthropic.Client:
```
efriis, line 678: if we don't have it at top-level, it doesn't matter, and also removes a validator

eyurtsev (approved):

PR #28825, opened 2024-12-19
openai[patch]: type reasoning_effort

baskaryan:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -454,6 +454,16 @@ class BaseChatOpenAI(BaseChatModel):
     """Total probability mass of tokens to consider at each step."""
     max_tokens: Optional[int] = Field(default=None)
     """Maximum number of tokens to generate."""
+    reasoning_effort: Optional[str] = None
```
ccurme, line 457: should we type as literal?

ccurme (approved):
I'm ignorant on this but what is state of o1 support in AzureChatOpenAI?

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: doesn't seem super high value? esp if we care about being forwards compatible

baskaryan:
> I'm ignorant on this but what is state of o1 support in AzureChatOpenAI?

not sure, but nothing gets passed unless explicitly specified

Issue #28813, opened 2024-12-19
"TypeError in embed_documents: Nested Embedding Structure (List[List[float]]) Causes Failure in LlamaCppEmbeddings with langchain-community v0.3.12"


Forgotten-Forever:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain.document_loaders import TextLoader
from langchain.text_splitter import CharacterTextSplitter
from langchain.embeddings import LlamaCppEmbeddings

# Load a text file
loader = TextLoader("./docs/raw.txt")
docs = loader.load()

# Split text into smaller chunks
text_splitter = CharacterTextSplitter(chunk_size=10, chunk_overlap=0)
texts = text_splitter.split_documents(docs)

# Prepare input for embedding
embeddings = LlamaCppEmbeddings(model_path="./models/llama-1B-q8_0.gguf")
_texts = [doc.page_content for doc in texts]

# Attempt to embed documents
embedded_texts = embeddings.embed_documents(_texts)

print(len(embedded_texts), len(embedded_texts[0]))

### Error Message and Stack Trace (if applicable)

[list(map(float, e["embedding"])) for e in embeddings["data"]]
PyDev console: starting.
Traceback (most recent call last):
  File "D:\Pycharm\PyCharm Community Edition 2023.2.3\plugins\python-ce\helpers\pydev\_pydevd_bundle\pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "<input>", line 1, in <listcomp>
TypeError: float() argument must be a string or a real number, not 'list'

### Description

When using langchain-community's LlamaCppEmbeddings with llama-cpp-python, the embed_documents method fails with a TypeError when processing certain input texts. The issue arises because the returned embedding structure from llama_cpp is unexpectedly nested (List[List[float]]), but embed_documents assumes a flat structure (List[float]).

Environment
Python version: 3.10
langchain-community: v0.3.12
llama-cpp-python: v0.3.5
Model: llama-1B-q8_0.gguf  --> llama-3.2-1B

Expected Behavior
[...]

PR #28827, opened 2024-12-19
Community: LlamaCppEmbeddings `embed_documents` and `embed_query`

keenborder786:
- **Description:** `embed_documents` and `embed_query` was throwing off the error as stated in the issue. The issue was that `Llama` client is returning the embeddings in a nested list which is not being accounted for in the current implementation and therefore the stated error is being raised.
- **Issue:** #28813


ccurme (commented):
Could you add a test?

keenborder786:
@ccurme OKAY

keenborder786:
@ccurme please review

libs/community/langchain_community/embeddings/llamacpp.py
```diff
@@ -116,7 +117,14 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
             List of embeddings, one for each text.
         """
         embeddings = self.client.create_embedding(texts)
-        return [list(map(float, e["embedding"])) for e in embeddings["data"]]
+        if not isinstance(embeddings["data"][0]["embedding"][0], list):
```
ccurme, line 120: is it possible for this line to raise IndexError where it previously would not? e.g., if `texts` is empty or contains empty strings.

ccurme (commented):
Thank you!

libs/community/langchain_community/embeddings/llamacpp.py
keenborder786: ok

keenborder786:
@ccurme please review

libs/community/langchain_community/embeddings/llamacpp.py
ccurme: Sorry for the confusion. I didn't mean to *request* that we raise an error where it previously would not.

From an ignorant reading of the code (i.e., no knowledge of the behavior of the client) the update introduced the potential to raise IndexError where it previously wouldn't. I've pushed a change that passes your tests. Let me know if you see any issues with it.

ccurme (approved):

PR #28844, opened 2024-12-20
langchain, community: Found problems in calling openai

reatang:
Description:

Problems found during use:

1. OpenAI requires that the last message can only be user or tool
 error info
```
openai.BadRequestError: Error code: 400 - {'chat_id': '', 'message': 'last message role must be user or tool', 'stream': False, 'error_code': 0, 'refer_blocks': None}
```

2. OpenAI's call_tools may return a function without a name
trace info
```
https://smith.langchain.com/public/b8f96ad1-1c14-4846-95b6-bd129ba72a03/r
```

There is no need to merge. I believe the official has a better solution.

### DIFF
--- a/libs/community/langchain_community/agent_toolkits/sql/base.py
+++ b/libs/community/langchain_community/agent_toolkits/sql/base.py
@@ -190,9 +190,9 @@ def create_sql_agent(
         if prompt is None:
             messages: List = [
                 SystemMessage(content=cast(str, prefix)),
-                HumanMessagePromptTemplate.from_template("{input}"),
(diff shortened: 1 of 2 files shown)


ccurme:
Can you post a full reproducible example?

I suspect something else is going on. The agent scratchpad should be getting formatted into tool messages, so we shouldn't need additional human messages.

From your LangSmith trace, the input query should be after the system message and before the AIMessage tool calls. It looks like there is no human message following the system message, so unclear how the model should know what tool calls to generate.

PR #28846, opened 2024-12-20
partners: allow to set Prefix in AIMessage (for MistralAI)

sdeparte:
**Description:**

Added ability to set `prefix` attribute to prevent error : 
```
httpx.HTTPStatusError: Error response 400 while fetching https://api.mistral.ai/v1/chat/completions: {"object":"error","message":"Expected last role User or Tool (or Assistant with prefix True) for serving but got assistant","type":"invalid_request_error","param":null,"code":null}
```

### DIFF
--- a/libs/partners/mistralai/langchain_mistralai/chat_models.py
+++ b/libs/partners/mistralai/langchain_mistralai/chat_models.py
@@ -345,6 +345,8 @@ def _convert_message_to_mistral_chat_message(
             message_dict["content"] = ""
         else:
             message_dict["content"] = message.content
+        if "prefix" in message.additional_kwargs:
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Thanks for this, can you add a test?

ccurme (approved):
Thank you!

Could be worth a follow-up adding this to the docs [here](https://github.com/langchain-ai/langchain/blob/master/docs/docs/integrations/chat/mistralai.ipynb), and/or the API reference (would probably require first fleshing out the Mistral API ref following the template [here](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/integration_template/chat_models.py)).

PR #28849, opened 2024-12-20
Docs: Add langgraph to installation section for Rag tutorial

ahmadelmalah:
**Issue**:
This tutorial depends on langgraph, however Langgraph is not mentioned on the installation section for the tutorial, which raises an error when copying and pasting the code snippets as following:
image

**Solution**:
Just adding langgraph package to installation section, for both pip and Conda tabs as this tutorial requires it.


ccurme (approved):
Thank you!

PR #28861, opened 2024-12-21
core[patch]: remove redundant imports

WrRan:
`Graph` has been imported at Line: 62

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -534,8 +534,6 @@ class Runnable(Generic[Input, Output], ABC):
 
     def get_graph(self, config: Optional[RunnableConfig] = None) -> Graph:
         """Return a graph representation of this Runnable."""
-        from langchain_core.runnables.graph import Graph
-
         graph = Graph()
         try:
             input_node = graph.add_node(self.get_input_schema(config))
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/utils/pydantic.py
```diff
@@ -279,7 +279,7 @@ def _create_subset_model_v2(
     fn_description: Optional[str] = None,
 ) -> type[pydantic.BaseModel]:
     """Create a pydantic model with a subset of the model fields."""
-    from pydantic import ConfigDict, create_model
```
WrRan, line 282: same issue. ConfigDict is imported at Line: 24

libs/core/langchain_core/runnables/base.py
```diff
@@ -534,8 +534,6 @@ def get_config_jsonschema(
 
     def get_graph(self, config: Optional[RunnableConfig] = None) -> Graph:
         """Return a graph representation of this Runnable."""
-        from langchain_core.runnables.graph import Graph
```
ccurme, line 537: Looks like the module-level import was originally just for type checking; that was later changed but this import was not removed https://github.com/langchain-ai/langchain/commit/7d5800ee515bf0cf38b97d09abb6857b767ab05f

libs/core/langchain_core/utils/pydantic.py
ccurme: This was added in https://github.com/langchain-ai/langchain/pull/27045, when the module-level import already existed, possibly to be explicit vs. the other `_v1` function.

cc @eyurtsev

ccurme (approved):

PR #28865, opened 2024-12-21
integration:langchain_xai

Archkon:
Adding a default parameter for model_name when no model provided may cause problems.

### DIFF
--- a/libs/partners/xai/langchain_xai/chat_models.py
+++ b/libs/partners/xai/langchain_xai/chat_models.py
@@ -252,7 +252,7 @@ class ChatXAI(BaseChatOpenAI):  # type: ignore[override]
 
     """  # noqa: E501
 
-    model_name: str = Field(alias="model")
+    model_name: str = Field(default="grok-beta", alias="model")
     """Model name to use."""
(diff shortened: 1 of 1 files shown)


libs/partners/xai/langchain_xai/chat_models.py
```diff
@@ -266,7 +266,6 @@ class Joke(BaseModel):
     """Base URL path for API requests."""
 
     openai_api_key: Optional[SecretStr] = None
-    openai_api_base: Optional[str] = None
```
ccurme, line 269: Why are we deleting public attributes?

ccurme (commented):
Thanks for this. We don't recommend setting default models. The reason is that over time the defaults will become out of date (e.g., we will be defaulting to an old or discontinued model), and updating the default can surprise users with changes in behavior.

libs/partners/xai/langchain_xai/chat_models.py
Archkon: Because the integration has already defined an xai_api_base and this class  inherites from langchain_openai integration  which actually defined enough public attributes.

PR #28879, opened 2024-12-22
Using latest audio transcribe from openai.

BirenMer:

- [ ] **openai bug fix in audio.py**: "community: found a bug while testing langchain in another repo found the bug in the main langchian repo as well hence correcting it"

### DIFF
--- a/libs/community/langchain_community/document_loaders/parsers/audio.py
+++ b/libs/community/langchain_community/document_loaders/parsers/audio.py
@@ -329,7 +329,8 @@ class OpenAIWhisperParser(BaseBlobParser):
                             model="whisper-1", file=file_obj, **self._create_params
                         )
                     else:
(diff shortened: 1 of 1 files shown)


BirenMer:
Changing the below line of code : 
transcript = openai.Audio.transcribe("whisper-1", file_obj)  # type: ignore[attr-defined]

As above functionality is no longer supported by openai.
To make the code more consistent and stable.

The new line is as follows:
transcript = openai.audio.transcriptions.create(
                            model="whisper-1",
                            file=file_obj, )

ccurme (commented):
You should only hit this line if you are using an old (pre-1.0) version of the openai sdk. For these versions the line looks correct:

https://github.com/openai/openai-python/blob/f7ccce126325ea35b6e5224ab954652c97a74896/openai/api_resources/audio.py#L46

Let me know if I've misunderstood something.

PR #28899, opened 2024-12-24
Update chroma.ipynb, Add langchain openai to the pip install in the chroma docs

Latticeworks1:
Description: added langchain openai to the pip
Issue: Example notebook will not run without this added
Dependencies: langchain openai


ccurme (commented):
The pip install for the embedding model is displayed dynamically based on what embedding you select. If you view the [rendered docs page](https://python.langchain.com/docs/integrations/vectorstores/chroma/), the pip install for langchain-openai is already included if you select OpenAI embeddings.

PR #28902, opened 2024-12-24
Docs: Add 'Optional' to installation section to fix an issue

ahmadelmalah:
Problem:
"Optional" object is used in one example without importing, which raises the following error when copying the example into IDE or Jupyter Lab
image

Solution:
Just importing Optional from typing_extensions module, this solves the problem!


efriis:
I'm guessing you didn't run the above code because `Optional` is imported towards the top of the doc!

ahmadelmalah:
Hi @efriis , thanks for your review!
A good guess and this is exactly why I created the PR!

This is not a tutorial where we follow instructions from top to bottom, it's a how-to guide where we select the part we need to apply.

Each method should be an independent working example, because they're different methods serving different purposes. 

So, in my case I don't need to copy Pydantic example from above so that TypedDict or JSON work.

and btw, this is good for consistency, as you can see on the other section "Choosing between multiple schemas" we imported Optional again on this line "from typing import Optional, Union" although we imported it early, because they're independent code blocks which makes more sense.

Thanks!

efriis:
works for me - let's switch it to import from the same package instead of typing_extensions here though

PR #28903, opened 2024-12-24
core: allow artifact in create_retriever_tool

ianchi:
Add option to return content and artifacts, to also be able to access the full info of the retrieved documents.

They are returned as a list of dicts in the `artifacts` property if parameter `response_format` is set to `"content_and_artifact"`.

Defaults to `"content"` to keep current behavior.


libs/core/langchain_core/tools/retriever.py
```diff
@@ -41,12 +46,18 @@ async def _aget_relevant_documents(
...
+    response_format: Literal["content", "content_and_artifact"] = "content",
+) -> Union[str, tuple[str, list[dict[str, Any]]]]:
     docs = await retriever.ainvoke(query, config={"callbacks": callbacks})
-    return document_separator.join(
+    content = document_separator.join(
         [await aformat_document(doc, document_prompt) for doc in docs]
     )
 
+    if response_format == "content_and_artifact":
+        return (content, [doc.model_dump() for doc in docs])
```
efriis, line 57: might make sense to just return `docs` as the artifact - any reason to dump here in particular?

libs/core/langchain_core/tools/retriever.py
```diff
@@ -28,11 +28,16 @@ def _get_relevant_documents(
...
-) -> str:
+    response_format: Literal["content", "content_and_artifact"] = "content",
+) -> Union[str, tuple[str, list[dict[str, Any]]]]:
     docs = retriever.invoke(query, config={"callbacks": callbacks})
-    return document_separator.join(
+    content = document_separator.join(
         format_document(doc, document_prompt) for doc in docs
     )
+    if response_format == "content_and_artifact":
+        return (content, [doc.model_dump() for doc in docs])
```
efriis, line 38: same q as on async one

libs/core/langchain_core/tools/retriever.py
```diff
@@ -55,6 +66,7 @@ def create_retriever_tool(
     *,
     document_prompt: Optional[BasePromptTemplate] = None,
     document_separator: str = "\n\n",
+    response_format: Literal["content", "content_and_artifact"] = "content",
```
efriis, line 69: needs to be added to docstring

efriis (commented):
2 questions/suggestions, and will need some unit tests to be mergable

libs/core/langchain_core/tools/retriever.py
ianchi: The reason is to always have the same type after deserialization.

As `artifact` has type `Any` when deserializing a `ToolMessage` from json or any other source (api call, DB, etc), Pydantic won't be able to do any "magic" to regenerate the artifact as a `list[Document]` but it will be a plain `list[dict]`.

So, to be consistent and always have a dict I opted to dump.
Otherwise you need to add custom logic when deserializing to have consistent types.

libs/core/langchain_core/tools/retriever.py
ianchi: Same reasoning as above

libs/core/langchain_core/tools/retriever.py
efriis: Got it. In this case it makes sense to keep it as a `List[Document]` given it's a langchain utility and leave ser/deser to the user.

Adjusting the philosophy from wrapper classes to dicts like this would be a separate discussion item for future versions of langchain, and would likely focus on more object types than just documents! For now, more important to stay consistent than to switch to dicts just here.

efriis (approved):

PR #28909, opened 2024-12-24
Docs: Updaing 'JSON Schema' code block output

ahmadelmalah:
Out seems outdate, I ran the example several times and this is the updated output, one key-value pair was missing!
image


ccurme (approved):
I just ran the cell twice. First time there was no rating, the second time there was. It's not a required field so it's reasonable for the model to leave it out. But it seems to be populated in the other examples so OK to update for consistency's sake.

PR #28915, opened 2024-12-25
core: fix tool_call function

Lycnkd:
- **Description:** For some LLM APIs, the current tool_call function is too vulnerable, causing validation errors within the chain or graph invocation, such as `tool_calls=[ { "name": "function_name", "args": '{}', "id": "tool_call_id", "type": "tool_call", } ]`, where args here is a string that resembles a dictionary, rather than the standard dictionary type. This issue is encountered when using the Qwen2.5 API and cannot be resolved by adding exception handling within the user-constructed chain or graph, as it is an error within LangChain itself. This update proactively catches dictionary-like structures in args if they are strings, which resolves the issue.

### DIFF
--- a/libs/core/langchain_core/messages/tool.py
+++ b/libs/core/langchain_core/messages/tool.py
@@ -210,17 +210,18 @@ class ToolCall(TypedDict):
 
 
 def tool_call(*, name: str, args: dict[str, Any], id: Optional[str]) -> ToolCall:
-    if isinstance(args, str):
-        try:
-            # Extract JSON-like dictionary from string using regex
-            match = re.search(r'\{.*\}', args)
-            if match:
+    try:
+        return ToolCall(name=name, args=args, id=id, type="tool_call")
+    except:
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this.

The [ToolCall](https://python.langchain.com/api_reference/core/messages/langchain_core.messages.tool.ToolCall.html) here is a langchain abstraction. Its `args` should always be a dict. Each chat model is responsible for creating a valid tool call. It's possible that there's a bug in the provider you're using (e.g., it looks like `ChatTongyi` [doesn't check](https://github.com/langchain-ai/langchain/blob/d26c102a5a02108481ed966071ddb2f825298f42/libs/community/langchain_community/chat_models/tongyi.py#L110C41-L110C46) for string `args` where maybe it should).

See [here](https://github.com/langchain-ai/langchain/blob/d26c102a5a02108481ed966071ddb2f825298f42/libs/partners/openai/langchain_openai/chat_models/base.py#L121-L141) for an example.

PR #28916, opened 2024-12-25
docs: Update VectorStoreTabs.js

bongsang:
- Title: Fix typo to correct "embedding" to "embeddings" in PGVector initialization example

- Problem: There is a typo in the example code for initializing the PGVector class. The current parameter "embedding" is incorrect as the class expects "embeddings".

- Correction: The corrected code snippet is:

vector_store = PGVector(
    embeddings=embeddings,
    collection_name="my_docs",
    connection="postgresql+psycopg://...",
)


ccurme (approved):
Thank you!

PR #28920, opened 2024-12-25
community[patch]/chat_models/bedrock.py - added yield after callback of on_llm_new_token

meenakshiruia:

    - **Description:** Invoke callback for on_llm_new_token prior to yielding tokens for bedrock.py
    - **Issue:** #16913


ccurme (commented):
Thank you. This is [already implemented](https://github.com/langchain-ai/langchain-aws/blob/94cb00b5a841a66bd04aae498b0053ad085e4ce3/libs/aws/langchain_aws/chat_models/bedrock.py#L510-L514) in the [langchain-aws](https://github.com/langchain-ai/langchain-aws/tree/main) version. You should be seeing warnings directing you to that version, away from the langchain-community version. There is a large gap in functionality between the two.

PR #28928, opened 2024-12-26
community: add modelscope endpoint

Yunnglin:
## Description

Add ModelScope inference API endpoint for both LLMs and ChatModels.

ModelScope is a leading platform to bridge model checkpoints and model applications. It offers the infrastructure for sharing open models and facilitating model-centric development (https://github.com/modelscope).


ccurme (commented):
Hey! This adds a net-new community integration, which has been replaced by dedicated integration packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-modelscope`) instead? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the models, and you have control over versioning.

Docs would continue to be maintained in the `langchain` repo. I'll close this PR, and would recommend reopening with just docs updates, as well as registering your package in libs/packages.yml! We'll be able to review simple PRs that only modify these two things much faster :)

Let me know what you think!

Yunnglin:
We have already integrated the modelscope embedding at https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/embeddings/modelscope_hub.py. What should we do next? Should we still release a new package and include the embedding in it?

Issue #28944, opened 2024-12-27
Add "developer" role support to ChatPromptTempkate


baskaryan:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Add support for 
```python
ChatPromptTemplate([("developer", "...")])
```

keenborder786:
@baskaryan please check, the PR.

PR #28794, opened 2024-12-18
Handle OpenAI developer msg 

baskaryan:
- Convert developer openai messages to SystemMessage
- store additional_kwargs={"__openai_role__": "developer"} so that the correct role can be reconstructed if needed
- update ChatOpenAI to read in openai_role

PR #28946, opened 2024-12-27
(Core): Developer support in `ChatPromptTemplate`

keenborder786:
- **Description:** Developer Role Added in `ChatPromptTemplate`
- **Issue:** #28944


keenborder786:
@baskaryan please check, I will update the tests further once you have given me the confirmation that I am on right path

ccurme (commented):
Thanks for this.

Given developer messages are largely equivalent to system messages at this point we opted to not (yet) introduce a new message type.

See https://github.com/langchain-ai/langchain/pull/28794

PR #28947, opened 2024-12-27
feat(openai): Update with_structured_output default for OpenAI

jacoblee93:
Should be accompanied by a minor bump

Should we also set `strict=True` by default?


libs/standard-tests/langchain_tests/integration_tests/chat_models.py
```diff
@@ -1310,6 +1325,24 @@ class Joke(BaseModel):
         joke_result = chat.invoke("Give me a joke about cats, include the punchline.")
         assert isinstance(joke_result, Joke)
 
+        # Schema
```
ccurme, line 1328: These new tests fail unless `strict` is False.

https://github.com/langchain-ai/langchain/pull/29075 is a hack (?) that will pass them if `strict` is True. But `strict=True` will break things in other ways (e.g., schemas with nested TypedDicts).

libs/partners/openai/tests/integration_tests/chat_models/test_azure_standard.py
```diff
@@ -55,6 +55,10 @@ def chat_model_params(self) -> dict:
             "azure_endpoint": OPENAI_API_BASE,
         }
 
+    @property
+    def structured_output_kwargs(self) -> dict:
+        return {"method": "function_calling"}
```
ccurme, line 60: Added this property to chat model standard tests. Azure legacy model tests don't support `method="json_schema"`.

ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain/actions/runs/12678107252

libs/partners/openai/tests/integration_tests/chat_models/test_azure_standard.py
efriis: do we actually want this to be configurable, or should we just have standard tests test the default?

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -820,20 +821,18 @@ class magic_function(BaseModel):
 
 
 @pytest.mark.parametrize(
-    ("model", "method", "strict"),
-    [("gpt-4o", "function_calling", True), ("gpt-4o-2024-08-06", "json_schema", None)],
+    ("model", "method"),
+    [("gpt-4o", "function_calling"), ("gpt-4o-2024-08-06", "json_schema")],
```
efriis, line 825: given strict is passed in should this still test both?

libs/partners/openai/tests/integration_tests/chat_models/test_azure_standard.py
ccurme: `TestAzureOpenAIStandardLegacy` breaks with default. Seems wasteful to xfail its structured output tests instead of running them with the supported method. Can also delete `TestAzureOpenAIStandardLegacy` but that didn't feel great either. lmk if there are other options.

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
ccurme: `strict` defaults to False, so most other tests cover it. I added some "invalid" schemas to the standard test as well (`test_structured_output_optional_param`) to add coverage there.

PR #28972, opened 2024-12-30
text-splitters: Add JSFrameworkTextSplitter for Handling JavaScript Framework Code

BenStuk:
## Description
This pull request introduces a new text splitter, `JSFrameworkTextSplitter`, to the Langchain library. The `JSFrameworkTextSplitter` extends the `RecursiveCharacterTextSplitter` to handle JavaScript framework code effectively, including React (JSX), Vue, and Svelte. It identifies and utilizes framework-specific component tags and syntax elements as splitting points, alongside standard JavaScript syntax. This ensures that code is divided at natural boundaries, enhancing the parsing and processing of JavaScript and framework-specific code.

### Key Features
- Supports React (JSX), Vue, and Svelte frameworks.
- Identifies and uses framework-specific tags and syntax elements as natural splitting points.
- Extends the existing `RecursiveCharacterTextSplitter` for seamless integration.

## Issue
No specific issue addressed.

## Dependencies
No additional dependencies required.


ccurme (commented):
Can this be implemented as a new case in the [get_separators_for_language](https://github.com/langchain-ai/langchain/blob/d26c102a5a02108481ed966071ddb2f825298f42/libs/text-splitters/langchain_text_splitters/character.py#L148) method?

BenStuk:
> Can this be implemented as a new case in the [get_separators_for_language](https://github.com/langchain-ai/langchain/blob/d26c102a5a02108481ed966071ddb2f825298f42/libs/text-splitters/langchain_text_splitters/character.py#L148) method?

hi @ccurme, we need the special method to handle React component tags (such as `<MyComponent>`) in a dynamic way, a fixed list of separators wouldn't be able to recognize them.

libs/text-splitters/langchain_text_splitters/jsx.py
```diff
@@ -0,0 +1,124 @@
...
+        separators = (
+            self._separators
+            + js_separators
+            + component_separators
+            + ["<>", "\n\n", "&&\n", "||\n"]
+        )
+        self._separators = separators
+
+        # Split the text using the separators
+        chunks = super().split_text(text)
```
ccurme, line 108: Could you comment why the logic around chunk overlaps below is necessary following the call to `super()`?

libs/text-splitters/langchain_text_splitters/jsx.py
```diff
@@ -0,0 +1,124 @@
...
+        - Handling chunk overlap if enabled
+
+        Args:
+            text: String containing code to split
+
+        Returns:
+            List of text chunks split on component and JS boundaries
+        """
+        # Extract unique opening component tags using regex
+        component_tags = list(
```
ccurme, line 58: CI is failing on `test_vue_text_splitter`, and I'm finding that this test is flaky when I run it locally (it fails around 50% of the time).

I think it is due to the use of `list(set(...))` here and on line 73, since this is not deterministic. Using `sorted(set(...))` will likely fix it if that makes sense to you.

ccurme (commented):
Apologies for the delay in review.

ccurme:
Hi @BenStuk, are you interested in continuing to iterate on this?

BenStuk:
Hey @ccurme, thanks for the reminder and for your inputs! Let me know if you have any questions about the new functionality or if there's anything else you'd like to see in this PR.

ccurme:
> Hey @ccurme, thanks for the reminder and for your inputs! Let me know if you have any questions about the new functionality or if there's anything else you'd like to see in this PR.

Thanks! I left some comments earlier-- most pressing is a flaky test (was failing ~50% of the time for me locally).

libs/text-splitters/langchain_text_splitters/jsx.py
BenStuk: This is fixed, all the test and linting problems should be gone

libs/text-splitters/langchain_text_splitters/jsx.py
BenStuk: Thanks @ccurme, you're absolutely right. The chunk overlap handling in the split_text() method is redundant since it's already being handled by the parent class. I have removed this redundant implementation to avoid potential double-overlapping of chunks. Thanks!

ccurme (approved):

PR #29002, opened 2025-01-03
Support for Gemini Tool Calling and Correct usage metadata in ChatOpenAI

Gaggi72:
So, to handle both the conditions, added an OR condition which will check whether the call has any of it.

- **Package:** Langchain-openai

- **Description:** Support for Gemini Tool Calling and Correct usage metadata in ChatOpenAI
- **Issue:** When calling Gemini via `ChatOpenAI` the `tool_calls` comes as `toolCalls` and metadata usage like `prompt_tokens` comes as `promptTokens`.
- **Dependencies:** N/A

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -120,7 +120,7 @@ def _convert_dict_to_message(_dict: Mapping[str, Any]) -> BaseMessage:
             additional_kwargs["function_call"] = dict(function_call)
         tool_calls = []
         invalid_tool_calls = []
-        if raw_tool_calls := _dict.get("tool_calls"):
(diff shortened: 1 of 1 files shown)


ccurme (commented):
What is the motivation for interacting with these models via ChatOpenAI instead of a google integration, such as [langchain-google-vertexai](https://github.com/langchain-ai/langchain-google/tree/main/libs/vertexai)?

Gaggi72:
@ccurme There was a functionality given by google to use openai package for gemini model and if there is support with openai package then why do we need another library such as langchain-google-vertexai.

Please refer to below link provided by google - 
https://ai.google.dev/gemini-api/docs/openai

ccurme:
Hi @Gaggi72, this should now be fixed in Google's API. Could you please check? Can re-open if needed.

PR #29007, opened 2025-01-03
community: Add "headers" parameter support to OpenAPI tools

unkindypie:
- **Description:** Allow OpenAPI agent to use request headers. 
It adds the following:
   * request tool now supports passing of the headers parameter. Headers are optional not to introduce breaking changes. Request headers are merged with headers passed to RequestWrapper class overwriting duplicate names.
   * OpenAPI planner prompts were updated to support new parameter.
   * `reduce_openapi_spec` now has new optional parameter "remove_optional", which allows to control whether optional parameters will be added to reduced API specification. This is required for the API I am integrating langchain with and often needed in real world scenarios.
   * new functionality for requests tool covered with tests
   * docstrings updated

### DIFF
--- a/libs/community/langchain_community/agent_toolkits/openapi/spec.py
+++ b/libs/community/langchain_community/agent_toolkits/openapi/spec.py
@@ -23,7 +23,9 @@ class ReducedOpenAPISpec:
     endpoints: List[Tuple[str, str, dict]]
 
 
-def reduce_openapi_spec(spec: dict, dereference: bool = True) -> ReducedOpenAPISpec:
+def reduce_openapi_spec(
+    spec: dict, dereference: bool = True, remove_optional: bool = False
+) -> ReducedOpenAPISpec:
     """Simplify/distill/minify a spec somehow.
 
     I want a smaller target for retrieval and (more importantly)
@@ -63,7 +65,7 @@ def reduce_openapi_spec(spec: dict, dereference: bool = True) -> ReducedOpenAPIS
             out["parameters"] = [
                 parameter
                 for parameter in docs.get("parameters", [])
-                if parameter.get("required")
+                if not remove_optional or parameter.get("required")
             ]
         if "200" in docs["responses"]:
             out["responses"] = docs["responses"]["200"]
--- a/libs/community/langchain_community/agent_toolkits/openapi/planner.py
+++ b/libs/community/langchain_community/agent_toolkits/openapi/planner.py
@@ -87,7 +87,10 @@ class RequestsGetToolWithParsing(BaseRequestsTool, BaseTool):  # type: ignore[ov
             raise e
         data_params = data.get("params")
         response: str = cast(
-            str, self.requests_wrapper.get(data["url"], params=data_params)
+            str,
+            self.requests_wrapper.get(
+                data["url"], data.get("headers", {}), params=data_params
+            ),
         )
         response = response[: self.response_length]
         return self.llm_chain.predict(
--- a/libs/community/langchain_community/agent_toolkits/openapi/planner_prompt.py
+++ b/libs/community/langchain_community/agent_toolkits/openapi/planner_prompt.py
@@ -151,6 +151,7 @@ REQUESTS_GET_TOOL_DESCRIPTION = """Use this to GET content from a website.
 Input to the tool should be a json string with 3 keys: "url", "params" and "output_instructions".
 The value of "url" should be a string. 
 The value of "params" should be a dict of the needed and available parameters from the OpenAPI spec related to the endpoint. 
+The value of "headers" should be a dict of the needed (e.g. if required only in certain case based on description) and available parameters from the OpenAPI spec related to the endpoint.
 If parameters are not needed, or not available, leave it empty.
 The value of "output_instructions" should be instructions on what information to extract from the response, 
 for example the id(s) for a resource(s) that the GET request fetches.
--- a/libs/community/langchain_community/tools/requests/tool.py
+++ b/libs/community/langchain_community/tools/requests/tool.py
@@ -4,6 +4,7 @@
 import json
 from typing import Any, Dict, Optional, Union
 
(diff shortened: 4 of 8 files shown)


unkindypie:
Example repo which uses headers plus new parameter successfully  
https://github.com/unkindypie/langchain-playground/blob/master/openapi_agent.py

unkindypie:
Example repo which uses new features, although client code won't differ from the current version except for new parameter in the reduce_openapi_spec function https://github.com/unkindypie/langchain-playground/blob/master/openapi_agent.py

libs/community/langchain_community/agent_toolkits/openapi/spec.py
```diff
@@ -23,7 +23,9 @@ class ReducedOpenAPISpec:
     endpoints: List[Tuple[str, str, dict]]
 
 
-def reduce_openapi_spec(spec: dict, dereference: bool = True) -> ReducedOpenAPISpec:
+def reduce_openapi_spec(
+    spec: dict, dereference: bool = True, remove_optional: bool = False
```
ccurme, line 27: Does defaulting to False change behavior for existing users?

Currently we are effectively removing optional by checking `parameter.get("required")`, right?

ccurme (commented):
Thanks for this. I'm wary of the current implementation because all LLMs currently using the requests tools will need to begin populating (empty) headers in their tool calls.

Is it desirable for the LLM to generate headers? Curious about the context. If they can be programmatically generated outside the LLM, we have a guide on passing those to tools [here](https://python.langchain.com/docs/how_to/tool_runtime/).

libs/community/langchain_community/agent_toolkits/openapi/spec.py
unkindypie: No, False will behave as it was working before my changes and True will keep the optional params which is really useful in my case
> Currently we are effectively removing optional by checking parameter.get("required"), right?
Right, we just remove anything which is not required

unkindypie:
> Thanks for this. I'm wary of the current implementation because all LLMs currently using the requests tools will need to begin populating (empty) headers in their tool calls.

Yes, but in theory nothing should change as the description of the tool is updated. I've tried on my end and it worked on GPT 4.0

> Is it desirable for the LLM to generate headers? Curious about the context. If they can be programmatically generated outside the LLM, we have a guide on passing those to tools [here](https://python.langchain.com/docs/how_to/tool_runtime/).

It's desirable in my case because headers must be passed only in a certain case based on description of a header which an LLM should determine, here is an example from our Swagger, but thank you for sharing.
image
After my changes to the package works quite cool, LLM reads from what endpoint it needs to get ID based on a description from an optional header, adds it to the path and makes an additional request

unkindypie:
@ccurme  Also regarding filling out empty headers problem, for this purpose I've made it optional and made it default to {}, so there must be no changes required for the existing code

unkindypie:
Hello @ccurme, following up on the status of the PR.
We really need this for one of the projects, would not be good if we leave there my fork, ideally to use mainstream branch
Thank you

libs/community/langchain_community/agent_toolkits/openapi/spec.py
ccurme: Not sure if I'm missing something but this change is not backward-compatible. Here is an example:

```python
docs = {
    "description": "description_value",
    "parameters": [
        {"name": "parameter_1", "required": True},
        {"name": "parameter_2"}
    ]
}
```

Before:
```python
def reduce_endpoint_docs(docs: dict) -> dict:
    out = {}
    if docs.get("description"):
        out["description"] = docs.get("description")
    if docs.get("parameters"):
        out["parameters"] = [
            parameter
            for parameter in docs.get("parameters", [])
            if parameter.get("required")
        ]
    print(out)


reduce_endpoint_docs(docs)
```
```
{'description': 'description_value', 'parameters': [{'name': 'parameter_1', 'required': True}]}
```

After:
```python
remove_optional = False

def reduce_endpoint_docs(docs: dict) -> dict:
    out = {}
    if docs.get("description"):
        out["description"] = docs.get("description")
    if docs.get("parameters"):
        out["parameters"] = [
            parameter
            for parameter in docs.get("parameters", [])
            if not remove_optional or parameter.get("required")
        ]
    print(out)

reduce_endpoint_docs(docs)
```
```
{'description': 'description_value', 'parameters': [{'name': 'parameter_1', 'required': True}, {'name': 'parameter_2'}]}
```

ccurme (commented):
Unfortunately there are too many breaking changes on a somewhat fundamental tool to merge this. This will break any basic usage of the requests tools:

```python
from langchain_community.tools import RequestsGetTool
from langchain_community.utilities.requests import TextRequestsWrapper


tool = RequestsGetTool(
    requests_wrapper=TextRequestsWrapper(headers={}),
    allow_dangerous_requests=True,
)

# This now raises TypeError:
tool.invoke({"url": "https://jsonplaceholder.typicode.com/posts?_limit=2"})
```
This change is not mitigated by the fact that tools have schemas and we are prompting LLMs to use different schemas. We can't make assumptions about the capabilities of all models to adapt to the change in prompts + schemas.

There are likely some workarounds for your own purposes, e.g.,
- A two-step process that first generates a header, then uses existing requests tools;
- Writing a new tool that expects headers and whatever other information needed + routes to existing requests tools.

unkindypie:
I see @ccurme, this is understandable. In this cause, could you please explain more about 

> 
Writing a new tool that expects headers and whatever other information needed + routes to existing requests tools.  

How do I overwrite what tool will the OpenAPI agent use, is there a parameter I can pass? Also something important, there were certain changes I made to OpenAPI agent prompts to teach it on how to use headers, so there I must update OpenAPI agent too and not only the tool.
I think what can I do is to create another tool, e.g. requests-v2 and migrate OpenAPI agent to is, and push it to the main package. Will it be accepted to be merged to the main repo?

Issue #26918, opened 2024-09-26
Deprecation wrapper poor performance (latency)


MartinGotelli:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

pip install snakeviz

script.py
```python
from pydantic import BaseModel
import cProfile
from langchain_openai import ChatOpenAI

class Model(BaseModel):
    value: str

profiler = cProfile.Profile()
profiler.enabled()
model = ChatOpenAI(model="gpt-4o", api_key="key")
model.with_structured_output(Model)
profiler.disable()
profile.dump_stats("make_martin_happy.prof")
```

python script.py
snakeviz make_martin_happy.prof

### Description

We are trying to build a performant model that calls gpt-4o-mini, and the main issue is latency. While trying to improve latency, I discovered that the check of the module from the inspect library can take around 0.25s. It's not a lot, but it adds to the total.

I don't want you to "fix" inspect.get_module. But I would like to have an alternative to disable LangChain warnings during the execution of my code.

<img width="1295" alt="image" src="https://github.com/user-attachments/assets/21e1a51e-b76f-4332-84be-33f52cda3c41">

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 23.6.0: Wed Jul 31 20:49:39 PDT 2024; root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T6000
> Python Version:  3.11.4 (main, Nov 30 2023, 11:49:37) [Clang 15.0.0 (clang-1500.0.40.1)]

Package Information
-------------------
> langchain_core: 0.3.2
> langchain: 0.3.0
> langchain_community: 0.3.0
> langsmith: 0.1.127
> langchain_aws: 0.2.1
> langchain_openai: 0.2.0
> langchain_postgres: 0.0.12
> langchain_text_splitters: 0.3.0
> langgraph: 0.2.27

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp: 3.10.5
> async-timeout: Installed. No version info available.
> boto3: 1.34.162
> dataclasses-json: 0.6.7
> httpx: 0.27.2
[...]

PR #29015, opened 2025-01-03
core[patch]: remove deprecated functions from tool binding hotpath

ccurme:
(Inspired by https://github.com/langchain-ai/langchain/issues/26918)

We rely on some deprecated public functions in the hot path for tool binding (`convert_pydantic_to_openai_function`, `convert_python_function_to_openai_function`, and `format_tool_to_openai_function`). My understanding is that what is deprecated is not the functionality they implement, but use of them in the public API -- we expect to continue to rely on them.

Here we update these functions to be private and not deprecated. We keep the public, deprecated functions as simple wrappers that can be safely deleted.

The `@deprecated` wrapper adds considerable latency due to its use of the `inspect` module. This update speeds up `bind_tools` by a factor of ~100x:

Before:

Screenshot 2025-01-03 at 11 22 55 AM

After:

Screenshot 2025-01-03 at 11 23 41 AM


baskaryan (approved):

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -75,31 +75,14 @@ def _rm_titles(kv: dict, prev_key: str = "") -> dict:
...
-        name: The name of the function. If not provided, the title of the schema will be
-            used.
-        description: The description of the function. If not provided, the description
-            of the schema will be used.
-        rm_titles: Whether to remove titles from the schema. Defaults to True.
-
-    Returns:
-        The function description.
-    """
+    """Converts a Pydantic model to a function description for the OpenAI API."""
```
efriis, line 85:
```suggestion
    """Converts a Pydantic model to a function description for the OpenAI API.
    Args:
        model: The Pydantic model to convert.
        name: The name of the function. If not provided, the title of the schema will be
            used.
        description: The description of the function. If not provided, the description
            of the schema will be used.
        rm_titles: Whether to remove titles from the schema. Defaults to True.
    Returns:
        The function description.
    """
```

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -121,6 +104,36 @@ def convert_pydantic_to_openai_function(
...
+        rm_titles: Whether to remove titles from the schema. Defaults to True.
+
+    Returns:
+        The function description.
+    """
+    return _convert_pydantic_to_openai_function(
+        model, name=name, description=description, rm_titles=rm_titles
+    )
+
+
```
efriis, line 136:
```suggestion
convert_pydantic_to_openai_function = deprecated(
    "0.1.16",
    alternative="langchain_core.utils.function_calling.convert_to_openai_function()",
    removal="1.0",
)(_convert_pydantic_to_openai_function)
```

efriis (approved):
small change for simplicity / non duplicate docstring

PR #29044, opened 2025-01-06
openai[patch]: support streaming with json_schema response format

ccurme:
- Stream JSON string content. Final chunk includes parsed representation (following OpenAI [docs](https://platform.openai.com/docs/guides/structured-outputs#streaming)).
- Mildly (?) breaking change: if you were using streaming with `response_format` before, usage metadata will disappear unless you set `stream_usage=True`.

## Response format

Before:

Screenshot 2025-01-06 at 11 59 01 AM

After:

Screenshot 2025-01-06 at 11 58 13 AM

## with_structured_output

For pydantic output, behavior of `with_structured_output` is unchanged (except for warning disappearing), because we pluck the parsed representation straight from OpenAI, and OpenAI doesn't return it until the stream is completed. Open to alternatives (e.g., parsing from content or intermediate dict chunks generated by OpenAI).

Before:

Screenshot 2025-01-06 at 12 38 11 PM

After:

Screenshot 2025-01-06 at 12 38 58 PM

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -318,8 +318,10 @@ def _convert_delta_to_message_chunk(
 def _convert_chunk_to_generation_chunk(
     chunk: dict, default_chunk_class: Type, base_generation_info: Optional[Dict]
 ) -> Optional[ChatGenerationChunk]:
+    if chunk.get("type") == "content.delta":
+        return None
     token_usage = chunk.get("usage")
-    choices = chunk.get("choices", [])
+    choices = chunk.get("choices", []) or chunk.get("snapshot", {}).get("choices", [])
 
     usage_metadata: Optional[UsageMetadata] = (
         _create_usage_metadata(token_usage) if token_usage else None
@@ -332,12 +334,20 @@ def _convert_chunk_to_generation_chunk(
         return generation_chunk
 
     choice = choices[0]
-    if choice["delta"] is None:
-        return None
+    if chunk.get("type") == "chunk":
+        refusal = choice.get("message", {}).get("refusal")
+        content = choice.get("message", {}).get("content")
+        message_chunk = AIMessageChunk(
+            content,
+            additional_kwargs={"refusal": refusal},
+        )
+    else:
+        if choice["delta"] is None:
+            return None
 
-    message_chunk = _convert_delta_to_message_chunk(
-        choice["delta"], default_chunk_class
-    )
+        message_chunk = _convert_delta_to_message_chunk(
+            choice["delta"], default_chunk_class
+        )
     generation_info = {**base_generation_info} if base_generation_info else {}
 
     if finish_reason := choice.get("finish_reason"):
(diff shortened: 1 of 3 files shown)


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1531,6 +1565,25 @@ def _filter_disabled_params(self, **kwargs: Any) -> Dict[str, Any]:
...
+        self, completion: openai.BaseModel
+    ) -> ChatGenerationChunk:
+        """Get chunk from completion (e.g., from final completion of a stream)."""
+        chat_result = self._create_chat_result(completion)
+        chat_message = chat_result.generations[0].message
+        if isinstance(chat_message, AIMessage):
+            usage_metadata = chat_message.usage_metadata
+        else:
+            usage_metadata = None
+        message = AIMessageChunk(
```
baskaryan, line 1578: i think we have a message to message chunk util

ccurme: Thanks, I poked around and found:

- [message_chunk_to_message](https://github.com/langchain-ai/langchain/blob/6a152ce245f3c89ef340cebf6faebdd58353f337/libs/core/langchain_core/messages/utils.py#L181) (I want to go the other way)
- A private function [_msg_to_chunk](https://github.com/langchain-ai/langchain/blob/6a152ce245f3c89ef340cebf6faebdd58353f337/libs/core/langchain_core/messages/utils.py#L1330) -- this function actually loses the parsed representation because it casts to/from dict. It's used in `merge_message_runs` which isn't used in monorepo, I see it's used in langchain-aws.

Let me know if you think it's important to add a util or I've missed it. Note I actually overwrite `content` to be `""` so it's not a perfect copy.

marcammann:
We just tried moving to 0.3.x and noticed that this latest change causes the tool_calls to be duplicated when the chunks are being aggregated.
We're using tool_calls and response_format at the same time using `bind` (skipping the convenience functions) and parsing the output ourselves.

I'm currently patching the `_get_generation_chunk_from_completion` function, adding the following:

```
if "tool_calls" in chat_message.additional_kwargs:
    chat_message.additional_kwargs.pop("tool_calls")
```

I'm not sure if this would cause any unintended side effects, but it's essentially removing the parts that are already sent as partial chunks prior. Without this, the tool call id's and names would end up concatenated and thus duplicated: "get_weatherget_weather". Let me know and I can open a PR.

ccurme:
> We just tried moving to 0.3.x and noticed that this latest change causes the tool_calls to be duplicated when the chunks are being aggregated. We're using `tool_calls` and `response_format` at the same time using `bind` (skipping the convenience functions) and parsing the output ourselves.
> 
> I'm currently patching the `_get_generation_chunk_from_completion` function, adding the following:
> 
> ```
> if "tool_calls" in chat_message.additional_kwargs:
>     chat_message.additional_kwargs.pop("tool_calls")
> ```
> 
> I'm not sure if this would cause any unintended side effects, but it's essentially removing the parts that are already sent as partial chunks prior. Without this, the tool call id's and names would end up concatenated and thus duplicated: "get_weatherget_weather". Let me know and I can open a PR.

Thanks @marcammann, your description of the issue makes sense. If you PR and tag me on it I'll prioritize.

PR #29049, opened 2025-01-06
rfc: used AIMessage.parsed

baskaryan:


libs/core/langchain_core/output_parsers/base.py
```diff
@@ -155,6 +169,8 @@ def _type(self) -> str:
                     return "boolean_output_parser"
     """  # noqa: E501
 
+    return_message: bool = False
```
ccurme, line 172: do we need to constrain this based on output type of parser? above docstring showcases a boolean parser for example, which will lead to boolean values of `AIMessage.parsed`.

ccurme (commented):
Do we want to default to return message False for initial roll-out?

ccurme:
Would also be good to showcase relevance to https://github.com/langchain-ai/langchain/pull/25338.

Issue #29034, opened 2025-01-05
ChatPromptTemplate.format_messages does not handle f-strings inherited from BaseMessage, such as SystemMessage, etc


TsinlingLi:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain_core.messages import HumanMessage, SystemMessage, AIMessage
from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder

ChatPromptTemplate.from_messages([
    SystemMessage("You are a helpful assistant. Answer all questions to the best of your ability in {language}"),
    MessagesPlaceholder(variable_name="messages")
]).invoke({"language": "English", "messages": [HumanMessage("hello")]})

### Error Message and Stack Trace (if applicable)

Unexpected results

### Description

result，language Not replaced:
ChatPromptValue(messages=[SystemMessage(content='You are a helpful assistant. Answer all questions to the best of your ability in {language}', additional_kwargs={}, response_metadata={}), HumanMessage(content='hello', additional_kwargs={}, response_metadata={})])

because，No processing of objects integrated from BaseMessage

class ChatPromptTemplate(ChatPromptTemplate):
    def format_messages(self, **kwargs: Any) -> list[BaseMessage]:
        kwargs = self._merge_partial_and_user_variables(**kwargs)
        result = []
        for message_template in self.messages:
            if isinstance(message_template, BaseMessage):
                # No processing of objects integrated from BaseMessage
                result.extend([message_template])
            elif isinstance(
                    message_template, (BaseMessagePromptTemplate, BaseChatPromptTemplate)
            ):
                message = message_template.format_messages(**kwargs)
                result.extend(message)
            else:
                msg = f"Unexpected input: {message_template}"
                raise ValueError(msg)
        return result

### System Info

python==3.12.7
langchain==0.3.14

sirine707:
Hi there ,To fix this, you need to modify the format_messages method to process f-strings here's the updated code let me know if iam not correct !
`class ChatPromptTemplate(ChatPromptTemplate):
    def format_messages(self, **kwargs: Any) -> list[BaseMessage]:
        kwargs = self._merge_partial_and_user_variables(**kwargs)
[...]

PR #29059, opened 2025-01-07
Fix template variable handling in ChatPromptTemplate BaseMessage objects #29034

sandy-sachin7:
Title: core: fix template variable handling in ChatPromptTemplate BaseMessage objects

**Description:**
Fixes a bug in ChatPromptTemplate where template variables in BaseMessage objects (like SystemMessage) were not being processed. Previously, template variables in direct BaseMessage instances would be left unformatted, while variables in MessagePromptTemplate instances were processed correctly. This PR modifies the `format_messages` method to properly handle template variables in all message types.

Example of the bug:
```python
template = ChatPromptTemplate.from_messages([
    SystemMessage("Answer in {language}"),  # Variable not processed
    HumanMessagePromptTemplate.from_template("Hello")  # Works fine
])
```

Before fix:
```python
result = template.invoke({"language": "English"})
# Output: SystemMessage(content='Answer in {language}')  # Variable not replaced
```

After fix:
```python
result = template.invoke({"language": "English"})
# Output: SystemMessage(content='Answer in English')  # Variable properly replaced
```

The fix adds template variable processing for BaseMessage objects while maintaining backward compatibility and existing functionality for other message types.

Key changes:
- Modified `format_messages` method to detect and process template variables in BaseMessage content
- Added template variable processing using PromptTemplate for BaseMessage instances
- Implemented proper error handling for missing variables or formatting failures
- Added comprehensive tests for template variable processing in all message types
- Maintained backward compatibility with existing usage patterns

Implementation details:
- Uses PromptTemplate for consistent variable formatting
- Only attempts formatting when template variables are detected
- Preserves original message if formatting fails
- Maintains all existing functionality for MessagePromptTemplate instances
@baskaryan @hwchase17 @efriis

### DIFF
--- a/libs/core/langchain_core/prompts/chat.py
+++ b/libs/core/langchain_core/prompts/chat.py
@@ -1219,7 +1219,7 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
         """
         kwargs = self._merge_partial_and_user_variables(**kwargs)
         result = []
-        
+
         for message_template in self.messages:
             if isinstance(message_template, BaseMessage):
                 # Check if the content contains any template variables
--- a/libs/core/langchain_core/prompts/chat.py
+++ b/libs/core/langchain_core/prompts/chat.py
@@ -1244,11 +1244,15 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
                     # If no template variables, use original message
                     result.append(message_template)
                     
-            elif isinstance(message_template, (BaseMessagePromptTemplate, BaseChatPromptTemplate)):
+            elif isinstance(
+                message_template, 
+                (BaseMessagePromptTemplate, BaseChatPromptTemplate)
+            ):
                 message = message_template.format_messages(**kwargs)
                 result.extend(message)
             else:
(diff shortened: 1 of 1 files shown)


SANTHOSH-SACHIN:
Those Lint errors can be ignored (wouldn't impact the codebase at all) . Regarding the Make test at prompty , the error isn't from my change rather it is because
```
E   ImportError: Error importing numpy: you should not try to import numpy from
E           its source directory; please exit the numpy source tree, and relaunch
E           your python interpreter from there.
```


Probably the way the test environment is set up @baskaryan @efriis @hwchase17

ccurme (commented):
Motivation for this is unclear. Why not just use `SystemMessagePromptTemplate`? Or simpler:
```python
ChatPromptTemplate([
    ("system", "Answer in {language}"),
    ("user", "Hello")
])
```
The existing implementation doesn't propagate variables to the input schema-- the `template` in your example has `input_variables == []`.

Also, in the case that users wish to pass text with brackets in their prompts, they might (reasonably) use a message instead of a message template. This has the potential to break that pattern as well.

PR #29063, opened 2025-01-07
community[minor]: Refactoring PyMuPDF parser, loader and add image blob parsers

pprados:
* Adds BlobParsers for images. These implementations can take an image and produce one or more documents per image. This interface can be used for exposing OCR capabilities.
* Update PyMuPDFParser and Loader to standardize metadata, handle images, improve table extraction etc.

- **Twitter handle:** pprados

This is one part of a larger Pull Request (PR) that is too large to be submitted all at once.
This specific part focuses to prepare the update of all parsers.

For more details, see [PR 28970](https://github.com/langchain-ai/langchain/pull/28970).


pprados:
@eyurtsev I rebase the code with master ;-)

eyurtsev (commented):
Great will take a look in the AM

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -46,6 +58,119 @@
...
 
+logger = logging.getLogger(__name__)
+
+_format_image_str = "\n\n{image_text}\n\n"
+_join_images = "\n"
+_join_tables = "\n"
+_default_page_delimitor = "\n\f"
+
+
+def purge_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
```
eyurtsev, line 69: could we keep it private if it's only used in this file?

```suggestion
def _purge_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
```

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -46,6 +58,119 @@
...
+        k = k.lower()
+        if k in ["creationdate", "moddate"]:
+            try:
+                new_metadata[k] = datetime.strptime(
+                    v.replace("'", ""), "D:%Y%m%d%H%M%S%z"
+                ).isoformat("T")
+            except ValueError:
+                new_metadata[k] = v
+        elif k in map_key:
+            # Normliaze key with others PDF parser
```
eyurtsev, line 98:
```suggestion
            # Normalize key with other PDF parsers
```

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +203,192 @@ def extract_from_images_with_rapidocr(
...
+        chat = model
+        for image in images:
+            image_bytes = io.BytesIO()
+            Image.fromarray(image).save(image_bytes, format="PNG")
+            img_base64 = base64.b64encode(image_bytes.getvalue()).decode("utf-8")
+            msg = chat.invoke(
+                [
+                    HumanMessage(
+                        content=[
+                            {"type": "text", "text": prompt.format()},
```
eyurtsev, line 338:
```suggestion
                            {"type": "text", "text": prompt },
```

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +203,192 @@ def extract_from_images_with_rapidocr(
...
+    def _convert_images_to_description(
+        images: Iterable[np.ndarray],
+    ) -> Iterator[str]:
+        try:
+            from PIL import Image
+        except ImportError:
+            raise ImportError(
+                "`PIL` package not found, please install it with `pip install pillow`"
+            )
+        chat = model
```
eyurtsev, line 329: may as well keep model throughout?

```suggestion
```

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +203,192 @@ def extract_from_images_with_rapidocr(
...
+        for image in images:
+            image_bytes = io.BytesIO()
+            Image.fromarray(image).save(image_bytes, format="PNG")
+            img_base64 = base64.b64encode(image_bytes.getvalue()).decode("utf-8")
+            msg = chat.invoke(
```
eyurtsev, line 334:
```suggestion
            msg = model.invoke(
```

eyurtsev: nit: this can be optimized to .batch() later on (not super improtant right now)

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +203,192 @@ def extract_from_images_with_rapidocr(
...
+            logger.debug("Tesseract text: %s", result.replace("\n", "\\n"))
+            yield result
+
+    _convert_images_to_text.creator = (  # type: ignore[attr-defined]
+        convert_images_to_text_with_tesseract
+    )
+    return _convert_images_to_text
+
+
+_prompt_images_to_description = PromptTemplate.from_template(
```
eyurtsev, line 294: Better to use a string here since .format() isn't used in a useful way

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +203,192 @@ def extract_from_images_with_rapidocr(
...
+                        result = result.replace("]", r"\\]")
+                        result = f"![{result}](.)"
+                    elif format == "html":
+                        result = f'<img alt="{html.escape(result, quote=True)}" />'
+                logger.debug("RapidOCR text: %s", result.replace("\n", "\\n"))
+                yield result
+            else:
+                yield ""
+
+    _convert_images_to_text.creator = (  # type: ignore[attr-defined]
```
eyurtsev, line 249: let's avoid assigning attributes to functions

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -270,93 +581,338 @@ def get_image(layout_object: Any) -> Any:
...
+                    DEFAULT_JOIN_TOLERANCE,
+                    DEFAULT_MIN_WORDS_HORIZONTAL,
+                    DEFAULT_MIN_WORDS_VERTICAL,
+                    DEFAULT_SNAP_TOLERANCE,
                 )
-                for page in doc
-            ]
 
-    def _get_page_content(self, doc: fitz.Document, page: fitz.Page, blob: Blob) -> str:
+                self.extract_tables_settings = {
```
eyurtsev, line 721: how were these chosen is it possible to add a comment?

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +203,192 @@ def extract_from_images_with_rapidocr(
...
+    These summaries will be embedded and used to retrieve the raw image. \
+    Give a concise summary of the image that is well optimized for retrieval \
+    and extract all the text from the image."""
+)
+
+
+def convert_images_to_description(
+    model: BaseChatModel,
+    *,
+    prompt: BasePromptTemplate = _prompt_images_to_description,
```
eyurtsev, line 305:
```suggestion
    prompt: str = _prompt_images_to_description,
```

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +203,192 @@ def extract_from_images_with_rapidocr(
     return text
 
 
+# Type to change the function to convert images to text.
+CONVERT_IMAGE_TO_TEXT = Optional[Callable[[Iterable[np.ndarray]], Iterator[str]]]
```
eyurtsev, line 207: MAJOR:

Why not use an ImageBlobParser w/ the regular Blob to Document interface. it'll allow reusing the image logic for images that do not originate from pdfs (e.g., to re-use for a web crawler)

---

A PDF parser doesn't would accept a parser as part of the initializer

```python
class PDFParser(...):
   def __Init__(self, ...  *, ..., image_blob_parser: Optional[BlobParser] = None):
      pass
```

If the image_pdf_parser is provided, then it'll be used for OCR purposes.

eyurtsev (commented):
Left two major comment, a few stylistic comments and some nits.

Let's tackle the two major comments:

1. Define the standardized structure of metadata
2. Create a dedicated ImageParser which is a blob parser

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -46,6 +58,119 @@
...
+_join_images = "\n"
+_join_tables = "\n"
+_default_page_delimitor = "\n\f"
+
+
+def purge_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
+    """
+    Purge metadata from unwanted keys and normalize key names.
+
+    Args:
```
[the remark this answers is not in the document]
pprados: 1. This method take all metadata, and normalize the key name for all metadata (lowercase, without '/' prefix).
2. It is not intended to manage the list of metadata to be implemented. In fact, this must be indicated somewhere, via a verification function for example (I've added this in a new function.).

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: done

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: I've done it! There are now 3 ImageBlogParsers!

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -46,6 +58,119 @@
...
+logger = logging.getLogger(__name__)
+
+_format_image_str = "\n\n{image_text}\n\n"
+_join_images = "\n"
+_join_tables = "\n"
+_default_page_delimitor = "\n\f"
+
+
+def purge_metadata(metadata: dict[str, Any]) -> dict[str, Any]:
+    """
```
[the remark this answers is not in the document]
pprados: 1. This method take all metadata, and normalize the key name for all metadata (lowercase, without '/' prefix).
2. It is not intended to manage the list of metadata to be implemented. In fact, this must be indicated somewhere, via a verification function for example (I add it).

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -46,6 +58,119 @@
...
+    new_metadata: dict[str, Any] = {}
+    map_key = {
+        "page_count": "total_pages",
+        "file_path": "source",
+    }
+    for k, v in metadata.items():
+        if type(v) not in [str, int]:
+            v = str(v)
+        if k.startswith("/"):
```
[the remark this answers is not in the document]
pprados: No, it's for the metadata key, not values. Some PDF parseur use key like /creationDate

libs/community/tests/integration_tests/document_loaders/parsers/test_pdf_parsers.py
```diff
@@ -123,11 +129,158 @@ def test_extract_images_text_from_pdf_pdfminerparser() -> None:
...
+            if pattern:
+                matches = re.findall(pattern, doc.page_content)
+                if matches:
+                    tables.extend(matches)
+        if extract_tables:
+            assert len(tables) >= 1
+        else:
+            assert not len(tables)
+
+    os.environ["SCARF_NO_ANALYTICS"] = "false"
```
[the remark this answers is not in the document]
pprados: I remove this

libs/community/tests/integration_tests/document_loaders/test_pdf.py
```diff
@@ -230,3 +206,34 @@ def test_amazontextract_loader_failures() -> None:
...
+    "parser_factory,params",
+    [
+        ("PyMuPDFLoader", {}),
+    ],
+)
+def test_standard_parameters(
+    parser_factory: str,
+    params: dict,
+) -> None:
+    loader_class = getattr(pdf_loaders, parser_factory)
```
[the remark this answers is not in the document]
pprados: If I do this, the test name loses explanation.

libs/community/tests/integration_tests/document_loaders/parsers/test_pdf_parsers.py
```diff
@@ -123,11 +129,158 @@ def test_extract_images_text_from_pdf_pdfminerparser() -> None:
...
+    os.environ["DO_NOT_TRACK"] = "true"
+
+    def images_to_text(images: list[np.ndarray]) -> Iterator[str]:
+        return iter(["<!-- image -->"] * len(images))
+
+    parser_class = getattr(pdf_parsers, parser_factory)
```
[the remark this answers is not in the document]
pprados: If I do this, the test name loses explanation.

libs/community/tests/unit_tests/document_loaders/parsers/test_pdf_parsers.py
```diff
@@ -75,14 +89,29 @@ def test_pdfminer_parser() -> None:
...
+    ],
+)
+def test_parsers(
+    parser_factory: str,
+    require: str,
+    params: dict[str, Any],
+) -> None:
+    try:
+        require = require.replace("-", "")
+        importlib.import_module(require, package=None)
```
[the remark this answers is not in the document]
pprados: If I do this, the test name loses explanation.

libs/community/tests/integration_tests/document_loaders/parsers/test_pdf_parsers.py
```diff
@@ -123,11 +129,158 @@ def test_extract_images_text_from_pdf_pdfminerparser() -> None:
...
+                    r"|[^\n]*))\n){2,})"
+                )
+            else:
+                pattern = None
+            if pattern:
+                matches = re.findall(pattern, doc.page_content)
+                if matches:
+                    tables.extend(matches)
+        if extract_tables:
+            assert len(tables) >= 1
```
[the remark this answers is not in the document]
pprados: It can be different, as some will only deal with the first, others with the second, and so on.

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
...
+                    elif self.format == "html":
+                        content = f'<img alt="{html.escape(content, quote=True)}" />'
+                logger.debug("Image text: %s", content.replace("\n", "\\n"))
+                yield Document(
+                    page_content=content,
+                    metadata={"source": blob.source},
+                )
+        except ImportError:
+            raise ImportError(
+                "`rapidocr-onnxruntime` package not found, please install it with "
```
eyurtsev, line 54: nit: incorrect import error

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
...
+                if content:
+                    if self.format == "markdown":
+                        content = content.replace("]", r"\\]")
+                        content = f"![{content}](.)"
+                    elif self.format == "html":
+                        content = f'<img alt="{html.escape(content, quote=True)}" />'
+                logger.debug("Image text: %s", content.replace("\n", "\\n"))
+                yield Document(
+                    page_content=content,
+                    metadata={"source": blob.source},
```
eyurtsev, line 50: Propagate blob metadata as well?

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
...
+_prompt_images_to_description = (
+    "You are an assistant tasked with summarizing "
+    "images for retrieval. "
+    "These summaries will be embedded and used to retrieve the raw image. "
+    "Give a concise summary of the image that is well optimized for retrieval "
+    "and extract all the text from the image."
+)
+
+
+class MultimodalBlobParser(ImageBlobParser):
```
eyurtsev, line 116: `LLMImageBlobParser` or something like that? 

Goal is to communicate that this is being done by a multi modal **llm**

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
...
+            content = ("\n".join([text[1] for text in ocr_result])).strip()
+        return content
+
+
+class TesseractBlobParser(ImageBlobParser):
+    def __init__(
+        self,
+        *,
+        format: Literal["text", "markdown", "html"] = "text",
+        langs: list[str] = ["eng"],
```
eyurtsev, line 91: tuple -- should not be a mutable default

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
...
+from langchain_community.document_loaders.blob_loaders import Blob
+
+logger = logging.getLogger(__name__)
+
+
+class ImageBlobParser(BaseBlobParser):
+    def __init__(
+        self,
+        *,
+        format: Literal["text", "markdown", "html"] = "text",
```
eyurtsev, line 26: This format is fairly surprising to see as part of the API -- but I think I'm OK with it.

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
...
+if TYPE_CHECKING:
+    from PIL.Image import Image
+
+from langchain_community.document_loaders.base import BaseBlobParser
+from langchain_community.document_loaders.blob_loaders import Blob
+
+logger = logging.getLogger(__name__)
+
+
+class ImageBlobParser(BaseBlobParser):
```
eyurtsev, line 22: Could we rename to BaseImageBlobParser or mark as private so it's clear that i's abstract

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
...
+            import pytesseract
+        except ImportError:
+            raise ImportError(
+                "`pytesseract` package not found, please install it with "
+                "`pip install pytesseract`"
+            )
+        return pytesseract.image_to_string(img, lang="+".join(self.langs)).strip()
+
+
+_prompt_images_to_description = (
```
eyurtsev, line 107: nit: style to follow google conventions a bit more closely

```suggestion
_PROMPT_IMAGES_TO_DESCRIPTION  = (
```

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,149 @@
+import base64
```
eyurtsev, line 1: I think this makes sense -- would you be willing to do a documentation pass for the API reference for this file? you could push the entire code through chat gpt and ask for google style doc-strings. It'll probably do a reasonable

libs/community/langchain_community/document_loaders/pdf.py
```diff
@@ -412,51 +418,127 @@ def lazy_load(self) -> Iterator[Document]:
...
+            extract_tables=extract_tables,
+            extract_tables_settings=extract_tables_settings,
         )
+
+    def lazy_load(self) -> Iterator[Document]:
+        """Lazy load given path as pages or single document (see `mode`).
+        Insert image, if possible, between two paragraphs.
+        In this way, a paragraph can be continued on the next page.
+        """
+        parser = self.parser
```
eyurtsev, line 535: the change in `load` in this file is still a breaking change .. see comment on the left hand side of the PR review

pprados: I'm sorry, I can't find the comment on the left hand side of the PR review.
If your remarque is for "Insert image, if possible, between two paragraphs.", yes.

As I explained in my original PR, this is necessary to avoid memory loss.
> Even though Document has a page_content parameter (rather than text or body), we believe it’s not good practice to work with pages. Indeed, this approach creates memory gaps in RAG projects. If a paragraph spans two pages, the beginning of the paragraph is at the end of one page, while the rest is at the start of the next. With a page-based approach, there will be two separate chunks, each containing part of a sentence. The corresponding vectors won’t be relevant. These chunks are unlikely to be selected when there’s a question specifically about the split paragraph. If one of the chunks is selected, there’s little chance the LLM can answer the question. This issue is worsened by the injection of headers, footers (if parsers haven’t properly removed them), images, or tables at the end of a page, as most current implementations tend to do.

There are also modifications with each new version of the parsers.
What do you recommend?

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,279 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class BaseImageBlobParser(BaseBlobParser):
+    """
+    Abstract base class for parsing image blobs into text.
+
+    Attributes:
+        format (Literal["text", "markdown", "html"]):
```
eyurtsev, line 28: Users will not understand what this means from looking at the API reference docs. Could we improve until this is self-explanation from just looking at the API Reference?

Perhaps a type update to: `Optional[Literal['markdown-link', 'html-img']]` and make the literal names more specific?

For example,

* None = return the content as is
* `markdown-link` = wrap the content into a markdown link, w/ link pointing to... 
* `html-img` = wrap the content as the `alt` text of an  tag and link to ...

Also consider a clash against a potential future implementation: The new implementation accepts a screenshot of an **entire** PDF page and create a markdown representation of it. (i.e., the output is `markdown` here, but the has a very different meaning now and we don't want to rewrap it all as a markdown link).

Perhaps renaming `format` to `wrap_as` would be helpful for communicating what this does?

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +88,148 @@ def extract_from_images_with_rapidocr(
...
+        all_extras = ""
+        str_extras = "\n\n".join(filter(lambda x: x, extras))
+        if str_extras:
+            all_extras = _PARAGRAPH_DELIMITOR[-1] + str_extras
+        all_text = text_from_page + all_extras
+
+    return all_text
+
+
+class ImagesPdfParser(BaseBlobParser):
```
eyurtsev, line 214: * Could we remove the intermediate abstraction? It doesn't serve any user facing purpose in this case -- our unit tests can check that parsers conform to an interface?
* Why are extract_images and images_parser definable independently? Why not use just images_parser: Optional[BaseImageBlobParser] = None for any new implementations?

libs/community/langchain_community/document_loaders/pdf.py
eyurtsev: I'm referring to the `kwargs` disappearing from `load`

```python
def load(self, **kwargs: Any) -> list[Document]:
  pass
```

We can try to maintain the load w/ kwargs and emit a deprecation warning on kwargs being passed at run time OR just restore kwargs for now

----

By the way `Document.page_content` is a misnomer -- it should've been named `content` so as not to indicate that it has to be a page of text. A document is a container for text + metadata

pprados: Ok.

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: - This is because, when other PRs are published with other parsers, this class will become meaningful and will be shared with all image management-compatible parsers.
- It's for compatibility reasons. I prefer to have only ` images_parser: Optional[BaseImageBlobParser] = None`

It is possible to declare `extract_images` deprecated, with, implicitly, the use of `RapidOCRBlobParser()` if, and only if, `extract_images=True`, with an alert.

What do you think?

pprados:
@eyurtsev 
Currently, core sources generate lint errors. There's nothing I can do about it.
```
[ "langchain_core" = "" ] || poetry run ruff format langchain_core --diff
1 file would be reformatted, 162 files already formatted
--- langchain_core/documents/base.py
+++ langchain_core/documents/base.py
```

libs/community/langchain_community/document_loaders/parsers/images.py
pprados: I'm not sure `wrap_as` is clearer than `format`.

libs/community/langchain_community/document_loaders/parsers/pdf.py
eyurtsev: > This is because, when 10 other PRs are published with other parsers, this class will become meaningful and will be shared with all image management-compatible parsers.

A few issues with the intermediate abstraction:

* PDF parsers be implemented as clients with the server performing logic that includes OCR etc (their init is different, but they still parse images). In this case, the name intermediate abstraction makes confusing not helpful.
* This opens the door to a TablesPDFParser as an abstract interface (e.g.,, to swap out the strategy for interpreting a table) , and then our parsers would start implementing mixins.
* It doesn't really optimize for users -- it's an extra abstraction learn about and creates more depth in the inheritance structure.
* This interfaces doubles down on `extract_images` and `images_parser` convention even though both parameters do not need to be specified at the same time.
* You have excellent standardized tests that can be used to also verify that __init__ interface is standardized for any registered parser

> It's for compatibility reasons. I prefer to have only  images_parser: Optional[BaseImageBlobParser] = None

Maybe we just roll it out where it's needed?

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,315 @@
...
+        Args:
+            format (Literal["text", "markdown-link", "html-img"]):
+              The format for the parsed output.
+              - "text" = return the content as is
+              - "markdown-link" = wrap the content into an image markdown link, w/ link
+              pointing to (`![body)(#)`]
+              - "html-img" = wrap the content as the `alt` text of an tag and link to
+              (`<img alt="{body}" src="#"/>`)
+        """
+        self.format = format or "text"
```
eyurtsev, line 49:
```suggestion
        self.format = format
```

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,315 @@
...
+
+    @abstractmethod
+    def _analyze_image(self, img: "Image") -> str:
+        """
+        Abstract method to analyze an image and extract textual content.
+
+        Args:
+            img (Image):
```
eyurtsev, line 57: nit: this is stylistic... but the presence of type annotations in the doc-string is for a time before type annotations were a thing in python. With type annotations present, documenting this again in the doc-string is at best duplication of information, at worst it's more stuff that goes out of sync 

---

Also we try to follow google style doc-strings (https://google.github.io/styleguide/pyguide.html

There's a convention for when `newline` is used

```suggestion
            img: The image to be analyzed
```

eyurtsev (commented):
Let's remove the intermediate abstraction for now for ImagesPDFParser -- we can introduce it if the need becomes self-evident in a future PR.

Otherwise I think this PR is in a pretty good state -- I'm going to do another pass tomorrow when I'm a bit fresher to make sure I haven't missed any potential breaking changes, if not we'll aim to merge.

eyurtsev:
Also if you can try to match the style for google style doc strings

libs/community/langchain_community/document_loaders/parsers/images.py
pprados: I've added true html and markdown formats where possible.

eyurtsev:
Hi sorry i didn't yet get to it today but it's at top of review queue so i'll push to tomorrow in the AM.

If I end up needing to make doc-string stylistic changes or expand the doc-strings do you want to review before I merge or just merge to unblock other PRs?

eyurtsev:
> @eyurtsev Currently, core sources generate lint errors. There's nothing I can do about it.
> 
> ```
> [ "langchain_core" = "" ] || poetry run ruff format langchain_core --diff
> 1 file would be reformatted, 162 files already formatted
> --- langchain_core/documents/base.py
> +++ langchain_core/documents/base.py
> ```

Don't worry about i'll try to resolve

pprados:
I think it's good now. I have no more ideas for improvement. 
It will be easier for the other PDF parsers (which I have to adjust according to our modifications). Thanks for your time.

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,313 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        format: Literal[
+            "text", "markdown-img", "html-img", "markdown", "html"
+        ] = "text",
+        model: BaseChatModel,
+        prompt: BasePromptTemplate = _PROMPT_IMAGES_TO_DESCRIPTION,
```
eyurtsev, line 265: IMO just keep the original prompt that you had -- this prompt will almost certainly need to be modified, and if people are already doing that they'll also modify the format regardless

eyurtsev (commented):
@pprados OK I think w/ the new updates we need to simplify the base images blob parser abstraction.

libs/community/langchain_community/document_loaders/pdf.py
```diff
@@ -429,44 +432,131 @@ def lazy_load(self) -> Iterator[Document]:
...
         self,
         file_path: Union[str, PurePath],
         *,
-        headers: Optional[dict] = None,
+        password: Optional[str] = None,
+        mode: Literal["single", "page"] = "page",
+        pages_delimiter: str = _DEFAULT_PAGES_DELIMITER,
         extract_images: bool = False,
+        images_parser: Optional[BaseImageBlobParser] = None,
+        extract_tables: Union[Literal["csv", "markdown", "html"], None] = None,
```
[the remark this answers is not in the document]
pprados: I use the name `images_inner_format` to group all image parameters with the same prefix.

libs/community/langchain_community/document_loaders/parsers/images.py
pprados: I agree

pprados:
yum is deprecated and replaced by dnf.
But, il doc/Makefile, `yum` is used.
I can not install `yum` on Ubuntu.
It's difficult for me to fix a bug in the documentation

libs/community/langchain_community/document_loaders/parsers/images.py
```diff
@@ -0,0 +1,239 @@
...
+if TYPE_CHECKING:
+    from PIL.Image import Image
+
+logger = logging.getLogger(__name__)
+
+
+class BaseImageBlobParser(BaseBlobParser):
+    """Abstract base class for parsing image blobs into text."""
+
+    def __init__(
```
eyurtsev, line 25: I believe we can remove the initializer now

pprados: done

eyurtsev (approved):

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -78,6 +89,156 @@ def extract_from_images_with_rapidocr(
...
+    map_key = {
+        "page_count": "total_pages",
+        "file_path": "source",
+    }
+    for k, v in metadata.items():
+        if type(v) not in [str, int]:
+            v = str(v)
+        if k.startswith("/"):
+            k = k[1:]
+        k = k.lower()
```
ccurme, line 164: Did this break user workflows? https://github.com/langchain-ai/langchain/issues/29470

PR #29065, opened 2025-01-07
core: Handle unterminated escape character when parsing partial JSON

balvisio:
**Description**
Currently, when parsing a partial JSON, if a string ends with the escape character, the whole key/value is removed. For example:

```
>>> from langchain_core.utils.json import parse_partial_json
>>> my_str = '{"foo": "bar", "baz": "qux\\'
>>> 
>>> parse_partial_json(my_str)
{'foo': 'bar'}
```

My expectation (and with this fix) would be for `parse_partial_json()` to return:
```
>>> from langchain_core.utils.json import parse_partial_json
>>> 
>>> my_str = '{"foo": "bar", "baz": "qux\\'
>>> parse_partial_json(my_str)
{'foo': 'bar', 'baz': 'qux'}
```

Notes:
1. It could be argued that current behavior is still desired.
2. I have experienced this issue when the streaming output from an LLM and the chunk happens to end with `\\`
3. I haven't included tests. Will do if change is accepted.
4. This is specially troublesome when this function is used by
https://github.com/langchain-ai/langchain/blob/187131c55c1b788da38124c6e7917151249746d6/libs/core/langchain_core/output_parsers/transform.py#L111

since what happens is that, for example, if the received sequence of chunks are: `{"foo": "b` , `ar\\` : 

Then, the result of calling `self.parse_result()` is:
```
{"foo": "b"}
```
and the second time:
```
{}
```


eyurtsev (commented):
Hi @balvisio thanks for looking into this. Could you add a unit test for this scenario?

balvisio:
@eyurtsev Done

efriis (approved):
makes sense

balvisio:
Hello @efriis , can this be merged? not sure about the failing test.

PR #29078, opened 2025-01-07
core: basemessage.text()

efriis:


libs/core/langchain_core/messages/base.py
```diff
@@ -92,6 +92,28 @@ def get_lc_namespace(cls) -> list[str]:
...
+        """
+        if isinstance(self.content, str):
+            return self.content
+
+        # must be a list
+        blocks = [
+            block
+            for block in self.content
+            if isinstance(block, str)
+            or block["type"] == "text"
```
hinthornw, line 110: Should we be safer about this or are we actually consistent enough to know it will have "type"?

efriis: made safer ty

ccurme (commented):
Looks good. Nicer than having to use StrOutputParser. For streaming this will work with Anthropic but haven't looked at any other providers.

How it works for StrOutputParser: https://github.com/langchain-ai/langchain/blob/c5bee0a544206e8851adf2f811151a6876322e51/libs/core/langchain_core/outputs/chat_generation.py#L50-L63

libs/core/langchain_core/messages/base.py
```diff
@@ -92,6 +92,27 @@ def get_lc_namespace(cls) -> list[str]:
         """
         return ["langchain", "schema", "messages"]
 
+    def text(self) -> str:
```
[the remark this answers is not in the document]
efriis: originally had it as an `@property` but thought it might be more forward-compatible with config options we might want to add in the future like `custom_separator: str = ""` or `image_placeholder: str = ""`

ccurme (approved):
Would be good to document a streaming example, e.g.,
```python
for chunk in llm.stream(...):
    print(chunk.text(), end="|")
```
Should target being able to replace this page: https://python.langchain.com/docs/how_to/output_parser_string/

libs/core/tests/unit_tests/test_messages.py
```diff
@@ -1017,3 +1017,55 @@ def test_tool_message_tool_call_id() -> None:
...
+    # content: [empty], [single element], [multiple elements]
+    # content dict types: [text], [not text], [no type]
+
+    assert HumanMessage(content="foo").text() == "foo"
+    assert AIMessage(content=[]).text() == ""
+    assert AIMessage(content=["foo", "bar"]).text() == "foobar"
+    assert (
+        AIMessage(
+            content=[
+                {"type": "text", "text": "<thinking>thinking...</thinking>"},
```
eyurtsev, line 1035: add unit test for ai message w/ tool call and no text content?

PR #29103, opened 2025-01-08
langchain: allow runnables as prompt

efriis:


eyurtsev:
@efriis consider adding a PR summary so folks can see what the feature is, especially if we intend to merge

Issue #29133, opened 2025-01-10
AzureOpenAIWhisperParser fails with DefaultAzureCredential due to incorrect token parameter


carloscaicedos:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_community.document_loaders.parsers.audio import AzureOpenAIWhisperParser
from langchain_core.documents.base import Blob
from azure.identity import DefaultAzureCredential, get_bearer_token_provider

scope = "https://cognitiveservices.azure.com/.default"

token_provider = get_bearer_token_provider(DefaultAzureCredential(), scope)

parser = AzureOpenAIWhisperParser(
    azure_endpoint="https://example.openai.azure.com/",
    api_version="2024-06-01",
    deployment_name="whisper",
    azure_ad_token_provider=token_provider,
)

audio_path = "audios/sample.mp3"
audio_blob = Blob(path=audio_path)

documents = parser.lazy_parse(blob=audio_blob)

for doc in documents:
    print(doc.page_content)
```

### Error Message and Stack Trace (if applicable)

---------------------------------------------------------------------------
AuthenticationError                       Traceback (most recent call last)
Cell In[1], line 21
     17 audio_blob = Blob(path=audio_path)
     19 documents = parser.lazy_parse(blob=audio_blob)
---> 21 for doc in documents:
     22     print(doc.page_content)

File d:\Repos\Python\OpenAI\.venv\Lib\site-packages\langchain_community\document_loaders\parsers\audio.py:205, in AzureOpenAIWhisperParser.lazy_parse(self, blob)
    203 try:
    204     if is_openai_v1():
--> 205         transcript = self._client.audio.transcriptions.create(
    206             model=self.deployment_name,
    207             file=file_obj,
    208             **self._create_params,
    209         )
    210     else:
    211         transcript = self._client.Audio.transcribe(
    212             model=self.deployment_name,
    213             deployment_id=self.deployment_name,
    214             file=file_obj,
    215             **self._create_params,
    216         )

[...]

PR #29135, opened 2025-01-10
[Community]: AzureOpenAIWhisperParser Authenication Fix

keenborder786:
- **Description:** `AzureOpenAIWhisperParser` authentication fix as stated in the issue.
- **Issue:** #29133


ccurme:
@Sheepsta300 would you mind taking a look at this? Two-line change but want to make sure it's correct. This integration is not that old afaik.

keenborder786:
@ccurme is this looking good?

Sheepsta300:
Thanks for the heads up, sorry and my bad for overlooking this but seems like an easy fix on the AD token property name. The suggested code change looks all good to me.

ccurme (approved):

Issue #29130, opened 2025-01-10
Presigned URLs can become invalid in `LakeFSLoader.load` when Unstructured is slow


dhdaines:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

Unfortunately this will require you to have a LakeFS configuration so it isn't that straightforward to reproduce (and may also depend on your specific LakeFS configuration)... but basically just call `ls_objects` with `presign=True` then wait a while... and then try to access one of the URLs (which is what the `LakeFSLoader` does internally).

```python
import dotenv
import os
import requests
import time
from langchain_community.document_loaders.lakefs import LakeFSClient

client = LakeFSClient(lakefs_access_key=lakefs_access_key,
                      lakefs_secret_key=lakefs_secret_key,
                      lakefs_endpoint=lakefs_endpoint)
objs = client.ls_objects(repo, ref, path, presign=True)
path, url = objs[0]
response = requests.get(url)
response.raise_for_status()
time.sleep(1200)
response = requests.get(url)
response.raise_for_status()
```

### Error Message and Stack Trace (if applicable)

```console
Traceback (most recent call last):
  File "lakefs_bug.py", line 24, in <module>
    response.raise_for_status()
  File ".venv/lib/python3.10/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: http://172.17.0.1:9200/mybucket/data/gb80naqqr9gs72ue9qu0/ctvfgfaqr9gs72ue9qv0?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=minioadmin%2F20250110%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20250110T151448Z&X-Amz-Expires=900&X-Amz-SignedHeaders=host&x-id=GetObject&X-Amz-Signature=89f8f515de3ee05589a4d7880b5a4a21957d2e3498ed702c436a9564fe8db285
```

### Description

[...]

PR #29137, opened 2025-01-10
[feat]: Parallel Execution for Loading of Objects in `LakeFSLoader`

keenborder786:
- **Description:** Rather than Loading of Objects Sequentially which causes too much delay in loading and causes the Signed URL to expire, using ProcessesPoolExecutor to load in parallel.
  - **Issue:** #29130


ccurme (commented):
Not sure it's worth parallelization in all cases, especially due to memory trade-off flagged by the poster of the original issue.

Might be worth testing with the [langchain-unstructured](https://pypi.org/project/langchain-unstructured/) package, which is intended to absorb all Unstructured functionality implemented in langchain-community.

PR #29154, opened 2025-01-11
community: Fix rank-llm import paths for new 0.20.3 version

tymzar:
# **PR title**: "community: Fix rank-llm import paths for new 0.20.3 version"  
  - The "community" package is being modified to handle updated import paths for the new `rank-llm` version.

---

## Description
This PR updates the import paths for the `rank-llm` package to account for changes introduced in version `0.20.3`. The changes ensure compatibility with both pre- and post-revamp versions of `rank-llm`, specifically version `0.12.8`. Conditional imports are introduced based on the detected version of `rank-llm` to handle different path structures for `VicunaReranker`, `ZephyrReranker`, and `SafeOpenai`.

## Issue
RankLLMRerank usage throws an error when used GPT (not only) when rank-llm version is > 0.12.8 - #29156 

## Dependencies
This change relies on the `packaging` and `pkg_resources` libraries to handle version checks.

## Twitter handle
@tymzar

### DIFF
--- a/libs/community/langchain_community/document_compressors/rankllm_rerank.py
+++ b/libs/community/langchain_community/document_compressors/rankllm_rerank.py
@@ -2,12 +2,14 @@ from __future__ import annotations
 
 from copy import deepcopy
 from enum import Enum
+from importlib.metadata import version
 from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence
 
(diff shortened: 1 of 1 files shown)


ccurme (approved):
Thanks for maintaining backward compatibility!

PR #29169, opened 2025-01-13
fixed a bug in streaming functionality

WSQsGithub:
- **PR title**: "community: fix bug in Perplexity streaming functionality"

    - **Description:** This PR resolves a bug in the streaming functionality of the Perplexity model. The issue was caused by the stream parameter being passed multiple times, resulting in the error: TypeError: openai.resources.chat.completions.Completions.create() got multiple values for keyword argument 'stream'. A single line was added to remove the conflicting stream argument to resolve the issue.
    - **Issue:** No specific issue linked.
    - **Dependencies:** none
    - **Twitter handle:** none


libs/community/langchain_community/chat_models/perplexity.py
```diff
@@ -191,6 +191,8 @@ def _convert_delta_to_message_chunk(
             additional_kwargs["function_call"] = function_call
         if _dict.get("tool_calls"):
             additional_kwargs["tool_calls"] = _dict["tool_calls"]
+        if _dict.get("citation"):
+            additional_kwargs["citation"] = _dict["citation"]
```
ccurme, line 195: Does this play nice with accumulating `AIMessageChunks`? i.e.,

```python
full = None
for chunk in llm.stream(query):
```

It looks like we have no integration tests for Perplexity. Might be worth adding some [here](https://github.com/langchain-ai/langchain/tree/master/libs/community/tests/integration_tests/chat_models). These don't run in CI but are useful for other developers working on this integration.

ccurme (commented):
Thanks! Left some comments (it looks like maintainers don't have [permission to make changes](https://python.langchain.com/docs/contributing/reference/faq/#how-do-i-allow-maintainers-to-edit-my-pr) on this PR).

libs/community/langchain_community/chat_models/perplexity.py
WSQsGithub: This line does not work, there not "citation" or "citations" attribution in "_dict".
I have removed it from the code.
The real ones that take effect are: 
```python
citations = chunk.get("citations", [])
            choice = chunk["choices"][0]
            chunk = self._convert_delta_to_message_chunk(
                choice["delta"], default_chunk_class
            )
            chunk.additional_kwargs["citations"] = citations
```

PR #29176, opened 2025-01-13
community: added citation to streaming

WSQsGithub:
**PR message**:  Added citation for streaming
    - **Description:** Added citation for streaming
    - **Issue:**  none
    - **Dependencies:** none
    - **Twitter handle:** none


libs/community/langchain_community/chat_models/perplexity.py
```diff
@@ -228,10 +228,12 @@ def _stream(
                 chunk = chunk.dict()
             if len(chunk["choices"]) == 0:
                 continue
+            citations = chunk.get("citations", [])
```
efriis, line 231: i believe citations come back with every chunk, so you'll have to only return them on the first chunk, or otherwise detect that the citations are identical to the ones already outputted.

Otherwise citations will be repeated when merging AIMessageChunk objects

efriis (commented):
Thanks!

could you add a unit test testing the chunk merging behavior as well?

here's a screenshot of example outputs if helpful re: citation merging
ScreenShot 2025-01-13 at 07 38 19PM

ccurme (commented):
Closing in favor of https://github.com/langchain-ai/langchain/pull/29273

PR #29183, opened 2025-01-13
Docs: Documentation update on PreFilters in Azure Cosmos DB No SQL

Fbelo:
**Description:**

The current live documentation for Azure Cosmos DB NoSQL integration is showing an incorrect method for setting pre-filters when using multiple vector search options. The existing example uses a raw dictionary approach:

```
pre_filter = {
    "conditions": [
        {"property": "metadata.page", "operator": "$eq", "value": 0},
    ],
}
```
**Issue:**

This method is not compatible with the latest implementation in the azure_cosmos_db_no_sql.py module. The correct approach involves using the PreFilter and Condition classes provided by the module.

**Fix Implemented:**

The corrected approach using the PreFilter and Condition classes is now as follows:
```
pre_filter = PreFilter(
        conditions=[
            Condition(
                property=f"metadata.{'company_id' if company_id else 'opportunity_id'}",
                operator="$eq",
                value=str(company_id if company_id else opportunity_id)
            )
        ]
    )
```

**Changes Made:**

- Updated every instance of the outdated pre-filter dictionary in the specific .ipynb file to use the new PreFilter and Condition classes.
- Adjusted imports to include the usage of PreFilter and Condition classes.


efriis (commented):
I believe the intention is to offer an interface that doesn't require importing custom classes.

cc @aayush3011 - I think this would be better as a bugfix to the cosmos integration instead of updating docs

marlenezw:
Yeah, actually @aayush3011 if this looks like something that needs to be changed can we add it to the new integration?

efriis:
merging this in the interim, and when this is in the new package would recommend switching back in the docs!

PR #29199, opened 2025-01-14
core: Cache RunnableLambda __repr__

cbornet:
`RunnableLambda`'s `__repr__` may do costly OS operation by calling `get_lambda_source`.
So it's better to cache it.
See #29043


libs/core/langchain_core/runnables/base.py
```diff
@@ -4351,6 +4351,8 @@ def __init__(
         except AttributeError:
             pass
 
+        self.repr: str | None = None
```
ccurme, line 4354: should this be private? i.e., `self._repr`

cbornet: Done in https://github.com/langchain-ai/langchain/pull/29199/commits/f99af0568f0547981eb5a515ab553a0ff04a55d6

libs/core/langchain_core/runnables/base.py
```diff
@@ -4529,14 +4531,18 @@ def __eq__(self, other: Any) -> bool:
...
         """A string representation of this Runnable."""
-        if hasattr(self, "func") and isinstance(self.func, itemgetter):
-            return f"RunnableLambda({str(self.func)[len('operator.'):]})"
-        elif hasattr(self, "func"):
-            return f"RunnableLambda({get_lambda_source(self.func) or '...'})"
-        elif hasattr(self, "afunc"):
-            return f"RunnableLambda(afunc={get_lambda_source(self.afunc) or '...'})"
-        else:
-            return "RunnableLambda(...)"
+        if self._repr is None:
```
efriis, line 4534: similar comment to last one - why not decorate with `@functools.lru_cache` instead?

cbornet: Using `lru_cache` on instance methods can lead to memory leaks (as `self` is hard referenced by the cache)
See https://docs.astral.sh/ruff/rules/cached-instance-method/
Do you prefer the solution with an external function or the attribute memoizer ?

eyurtsev: TIL

ccurme (approved):

PR #29200, opened 2025-01-14
core: Cache RunnableLambda deps

cbornet:
`RunnableLambda`'s `deps` may do costly OS operation by calling `get_function_nonlocals`.
So it's better to cache it.
See #29043


efriis (commented):
why not just do a `@functools.cached_property`?

cbornet:
> why not just do a @functools.cached_property?

Of course, much better!
Done.

ccurme (approved):

PR #29201, opened 2025-01-14
core: Bump ruff version to 0.9

cbornet:
Also run some preview autofix and formatting


libs/core/pyproject.toml
```diff
@@ -82,7 +82,7 @@ classmethod-decorators = [ "classmethod", "langchain_core.utils.pydantic.pre_ini
 "scripts/**" = [ "S",]
 
 [tool.poetry.group.lint.dependencies]
-ruff = "^0.5"
+ruff = "~0.9.1"
```
efriis, line 85:
```suggestion
ruff = "^0.9.1"
```

efriis (approved):

PR #29205, opened 2025-01-14
docs: Remove redundant "%"

junhe:
Before this commit, the copied command can't be used directly.

### DIFF
--- a/docs/docs/tutorials/classification.ipynb
+++ b/docs/docs/tutorials/classification.ipynb
@@ -49,7 +49,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Can you clarify the issue here? Running this in a Jupyter notebook should run the pip install.

junhe:
> Can you clarify the issue here? Running this in a Jupyter notebook should run the pip install.

It was just that there is a redundant `%`. This is unwanted because, for example, when users click "Copy" button, the `%` will also be copied, which makes the command unable to run directly in the terminal.

ccurme (approved):

PR #29206, opened 2025-01-14
community: Bump ruff version to 0.9

cbornet:


libs/community/pyproject.toml
```diff
@@ -119,7 +120,7 @@ pytest-vcr = "^1.0.2"
 vcrpy = "^6"
 
 [tool.poetry.group.lint.dependencies]
-ruff = "^0.5"
+ruff = "~0.9.1"
```
efriis, line 123:
```suggestion
ruff = "^0.9.1"
```

cbornet: Ruff’s versioning policy is to add new rules and change formatting between minor versions. That’s why I think pinning the minor is better here.

efriis (approved):

PR #29207, opened 2025-01-14
community: Additional parameters for OpenAIAssistantV2Runnable 

pm390:
**Description:** Added Additional parameters that could be useful for usage of OpenAIAssistantV2Runnable.

This change is thought to allow langchain users to set parameters that cannot be set using assistants UI (max_completion_tokens,max_prompt_tokens,parallel_tool_calls) and parameters that could be useful for experimenting like top_p and temperature. 

This PR originated from the need of using parallel_tool_calls in langchain, this parameter is very important in openAI assistants because without this parameter set to False strict mode is not respected by OpenAI Assistants (https://platform.openai.com/docs/guides/function-calling#parallel-function-calling).

> Note: Currently, if the model calls multiple functions in one turn then strict mode will be disabled for those calls.

**Issue:** None
**Dependencies:** openai

### DIFF
--- a/libs/langchain/langchain/agents/openai_assistant/base.py
+++ b/libs/langchain/langchain/agents/openai_assistant/base.py
@@ -293,7 +293,8 @@ class OpenAIAssistantRunnable(RunnableSerializable[Dict, OutputType]):
                 instructions: Additional run instructions.
                 model: Override Assistant model for this run.
                 tools: Override Assistant tools for this run.
-                parallel_tool_calls: Allow Assistant to disallow parallel_tool_calls.
+                parallel_tool_calls: Allow Assistant to set parallel_tool_calls
+                    for this run.
                 top_p: Override Assistant top_p for this run.
                 temperature: Override Assistant temperature for this run.
                 max_completion_tokens: Allow setting max_completion_tokens for this run.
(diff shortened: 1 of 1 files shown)


pm390:
Actually...
I was just testing around the change, I noticed that the run_metadata parameter does not exist (and openAI doesn't have kwargs on their side) so invoking the assistant with the parameter set results in (if using async version):
`TypeError: AsyncThreads.create_and_run() got an unexpected keyword argument 'run_metadata'`

I'll try get a minimal reproducer and open an Issue later this week.

ccurme:
> Actually... I was just testing around the change, I noticed that the run_metadata parameter does not exist (and openAI doesn't have kwargs on their side) so invoking the assistant with the parameter set results in (if using async version): `TypeError: AsyncThreads.create_and_run() got an unexpected keyword argument 'run_metadata'`
> 
> The correct parameter should be `metadata`. I'll try get a minimal reproducer and open an Issue later this week.

Thanks @pm390. From your perspective, is this PR good to go? i.e., the issue with `run_metadata` is separate? Or should we hold off?

pm390:
@ccurme from my perspective we should treat the issue as a separate topic (adding a new functionality support should be separated from fixing an issue), so this PR is good to go.

ccurme (approved):

pm390:
Reproducer created and issue opened [29287](https://github.com/langchain-ai/langchain/issues/29287).

PR #29211, opened 2025-01-14
langchain: Bump ruff version to 0.9

cbornet:


libs/langchain/pyproject.toml
```diff
@@ -119,7 +120,7 @@ cassio = "^0.1.0"
 langchainhub = "^0.1.16"
 
 [tool.poetry.group.lint.dependencies]
-ruff = "^0.5"
+ruff = "~0.9.1"
```
efriis, line 123: let's keep this as `^0.9` to avoid auto updating and pulling in breaking changes

cbornet: Breaking changes can happen between minor versions in ruff (new rules, different formatting).
`^0.9` is equivalent to `>=0.9, <1.0.0` whereas `~0.9.1` is equivalent to `>=0.9, <0.10.0`. So I think the latter is better.

efriis: TIL they're actually equivalent in this case https://python-poetry.org/docs/dependency-specification/#caret-requirements

we tend to use `^` across other dependencies so will keep there

efriis (approved):

libs/langchain/pyproject.toml
cbornet: Oh yes for version <1, caret rules are different ! (TIL too 😄)

Issue #29227, opened 2025-01-15
SQLDatabase in langchain_community.utilities raises Exception due to type mismatch when connecting with Snowflake


SyedBaqarAbbas:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code can be used to reproduce the error:

```python

from langchain_community.utilities import SQLDatabase

# Fill in snowflake credentials
SNOWFLAKE_ACCOUNT = ""
SNOWFLAKE_USERNAME = ""
SNOWFLAKE_PASSWORD = ""
SNOWFLAKE_WAREHOUSE = ""
SNOWFLAKE_DATABASE = ""
SNOWFLAKE_SCHEMA = ""
SNOWFLAKE_ROLE = ""  # Optional field

# Build the connection string
connection_string = (
    f"snowflake://{SNOWFLAKE_USERNAME}:{SNOWFLAKE_PASSWORD}@{SNOWFLAKE_ACCOUNT}/"
    f"{SNOWFLAKE_DATABASE}/{SNOWFLAKE_SCHEMA}?warehouse={SNOWFLAKE_WAREHOUSE}"
)

# Add role if provided
if SNOWFLAKE_ROLE:
    connection_string += f"&role={SNOWFLAKE_ROLE}"

# Connect to database using langchain's wrapper
sql_db = SQLDatabase.from_uri(database_uri=connection_string)
```

### Error Description
I encountered the following error:

```
TypeError: unsupported operand type(s) for +: 'dict_keys' and 'list'
```

### Relevant Library Versions
- **snowflake-sqlalchemy**: 1.7.2  
- **snowflake-connector-python**: 3.12.4  
- **sqlalchemy**: 2.0.20  
- **langchain_community**: 0.3.14  

### Description

I am trying to use the langchain's wrapper for sqlachemy (langchain_community.utilities.SQLDatabase) to connect to a snowflake database.

While connecting, instead of a successfully connecting, I am facing the error `TypeError: unsupported operand type(s) for +: 'dict_keys' and 'list'`

### System Info

System Information
------------------
> OS:  Linux
[...]

PR #29229, opened 2025-01-15
[fix] Convert table names to list for compatibility in SQLDatabase

SyedBaqarAbbas:
- [langchain_community.utilities.SQLDatabase] **[fix] Convert table names to list for compatibility in SQLDatabase**: 
  - The issue #29227 is being fixed here
  - The "package" modified is community
  - The issue lied in this block of code:
https://github.com/langchain-ai/langchain/blob/44b41b699c3815206413125ec58b6cca601ee438/libs/community/langchain_community/utilities/sql_database.py#L72-L77

- [langchain_community.utilities.SQLDatabase] **[fix] Convert table names to list for compatibility in SQLDatabase**: 
    - **Description:** When the SQLDatabase is initialized, it runs a code `self._inspector.get_table_names(schema=schema)` which expects an output of list. However, with some connectors (such as snowflake) the data type returned could be another iterable. This results in a type error when concatenating the table_names to view_names. I have added explicit type casting to prevent this.
    - **Issue:** The issue #29227 is being fixed here
    - **Dependencies:** None
    - **Twitter handle:** @BaqarAbbas2001

## Additional Information
When the following method is called for a Snowflake database:
https://github.com/langchain-ai/langchain/blob/44b41b699c3815206413125ec58b6cca601ee438/libs/community/langchain_community/utilities/sql_database.py#L75

Snowflake under the hood calls:
```python
from snowflake.sqlalchemy.snowdialect import SnowflakeDialect
SnowflakeDialect.get_table_names
```

This method returns a `dict_keys()` object which is incompatible to concatenate with a list and results in a `TypeError`

### Relevant Library Versions
- **snowflake-sqlalchemy**: 1.7.2  
- **snowflake-connector-python**: 3.12.4  
- **sqlalchemy**: 2.0.20  
- **langchain_community**: 0.3.14

### DIFF
--- a/libs/community/langchain_community/utilities/sql_database.py
+++ b/libs/community/langchain_community/utilities/sql_database.py
@@ -72,7 +72,7 @@ class SQLDatabase:
         # including view support by adding the views as well as tables to the all
         # tables list if view_support is True
         self._all_tables = set(
-            self._inspector.get_table_names(schema=schema)
+            list(self._inspector.get_table_names(schema=schema))
             + (self._inspector.get_view_names(schema=schema) if view_support else [])
         )
 


SyedBaqarAbbas:
While this was [fixed very recently in snowflake-alchemy ](https://github.com/snowflakedb/snowflake-sqlalchemy/commit/d84484e148f97586086504f53acf095b67af76af), most systems running the older versions will still be facing the issue. This PR offers backward compatibility for most of the connectors present and offers robustness if other sqlalchemy engines return an output other than `list` in their `get_table_names`.

ccurme (approved):

PR #29240, opened 2025-01-15
core[patch]: make BaseMessage subscriptable

baskaryan:
reduce cognitive overload for users of having to know when messages are dicts (usually as inputs) and when they're BaseMessages (usually as outputs)


nhuang-lc (approved):
Nice ty!

Issue #29125, opened 2025-01-10
 MistralAIEmbeddings retry decorator catches wrong exception type for rate limiting


Gero-Gi:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_mistralai import MistralAIEmbeddings
    embeddings = MistralAIEmbeddings(
    model="mistral-embed",
    mistral_api_key="your-api-key",
    wait_time=30 # Should be used for rate limit retries
)
```

### Making multiple requests in quick succession triggers rate limits
```python
for i in range(10):
    # When hitting rate limits, raises httpx.HTTPStatusError without retrying
    embeddings.embed_query("Test string")
```

### Error Message and Stack Trace (if applicable)

  File "/usr/local/lib/python3.10/site-packages/langchain_core/language_models/chat_models.py", line 286, in invoke
    self.generate_prompt(
  File "/usr/local/lib/python3.10/site-packages/langchain_core/language_models/chat_models.py", line 786, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/langchain_core/language_models/chat_models.py", line 643, in generate
    raise e
  File "/usr/local/lib/python3.10/site-packages/langchain_core/language_models/chat_models.py", line 633, in generate
    self._generate_with_cache(
  File "/usr/local/lib/python3.10/site-packages/langchain_core/language_models/chat_models.py", line 851, in _generate_with_cache
    result = self._generate(
  File "/usr/local/lib/python3.10/site-packages/langchain_mistralai/chat_models.py", line 545, in _generate
    response = self.completion_with_retry(
  File "/usr/local/lib/python3.10/site-packages/langchain_mistralai/chat_models.py", line 464, in completion_with_retry
    rtn = _completion_with_retry(**kwargs)
  File "/usr/local/lib/python3.10/site-packages/langchain_mistralai/chat_models.py", line 461, in _completion_with_retry
    _raise_on_error(response)
  File "/usr/local/lib/python3.10/site-packages/langchain_mistralai/chat_models.py", line 170, in _raise_on_error
[...]

PR #29242, opened 2025-01-16
[MistralAI] Improve MistralAIEmbeddings

ZhangShenao:
- Add static method decorator for method.
- Add expected exception for retry decorator

#29125

### DIFF
--- a/libs/partners/mistralai/langchain_mistralai/embeddings.py
+++ b/libs/partners/mistralai/langchain_mistralai/embeddings.py
@@ -32,7 +32,8 @@ documents/chunks)"""
 class DummyTokenizer:
     """Dummy tokenizer for when tokenizer cannot be accessed (e.g., via Huggingface)"""
 
-    def encode_batch(self, texts: List[str]) -> List[List[str]]:
+    @staticmethod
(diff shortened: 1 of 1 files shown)


libs/partners/mistralai/langchain_mistralai/embeddings.py
```diff
@@ -223,7 +224,9 @@ def embed_documents(self, texts: List[str]) -> List[List[float]]:
             batch_responses = []
 
             @retry(
-                retry=retry_if_exception_type(httpx.TimeoutException),
+                retry=retry_if_exception_type(
+                    (httpx.TimeoutException, httpx.HTTPStatusError)
```
ccurme, line 228: Will any other error codes raise HTTPStatusError?

Httpx [docs](https://www.python-httpx.org/exceptions/) say "4xx or 5xx". Can we catch 429 errors specifically?

ZhangShenao: The httpx library does not provide an exception definition for TooManyRequests, which is included in the HTTPStatus Error

PR #29249, opened 2025-01-16
 [langchain_community.llms.xinference]: Rewrite _stream() method and support stream() method in xinference.py

TheSongg:
- [ ] **PR title**:  [langchain_community.llms.xinference]: Rewrite _stream() method and support stream() method in xinference.py

- [ ] **PR message**:
- Rewrite the _stream method so that the chain.stream() can be used to return data streams.

        chain = prompt | llm
        chain.stream(input=user_input)

- [ ] **tests**:

       from langchain_community.llms import Xinference
       llm = Xinference(
         server_url="http://0.0.0.0:9997", # replace your xinference server url
         model_uid={model_uid}  # replace model_uid with the model UID return from launching the model
         stream=True
         )
       prompt = PromptTemplate(input=['country'], template="Q: where can we visit in the capital of {country}? A:")
       chain = prompt | llm
       chain.stream(input={'country': 'France'})

### DIFF
--- a/libs/community/langchain_community/llms/xinference.py
+++ b/libs/community/langchain_community/llms/xinference.py
@@ -85,21 +85,21 @@ class Xinference(LLM):
             prompt="Q: where can we visit in the capital of France? A:",
             generate_config={"max_tokens": 1024, "stream": True},
         )
-    
+
     Example:
 
     .. code-block:: python
-    
+
         from langchain_community.llms import Xinference
         from langchain.prompts import PromptTemplate
 
         llm = Xinference(
             server_url="http://0.0.0.0:9997",
-            model_uid = {model_uid}, # replace model_uid with the model UID return from launching the model
-            stream = True
+            model_uid={model_uid}, # replace model_uid with the model UID return from launching the model
+            stream=True
         )
         prompt = PromptTemplate(
-            input=['country'], 
+            input=['country'],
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/llms/xinference.py
```diff
@@ -216,3 +249,68 @@ def _stream_generate(
...
+        return GenerationChunk(
+            text=token,
+            generation_info=dict(
+                finish_reason=stream_response["choices"][0].get("finish_reason", None),
+                logprobs=stream_response["choices"][0].get("logprobs", None),
+            ),
+        )
+
+    @staticmethod
+    def create_stream(
```
ccurme, line 311: why make this method public?

and maybe a nitpick but IMO `yield from model.generate(...)` in `_create_generate_stream` is simple enough

ccurme (commented):
You'll need to run `make format` following the [contribution guidelines](https://python.langchain.com/docs/contributing/how_to/code/setup/#formatting-and-linting).

libs/community/langchain_community/llms/xinference.py
TheSongg: Previously, the create_stream(...) method was complicated, and I forget to combine it after optimization. Now I combined these two methods.

PR #29268, opened 2025-01-17
core: Add ruff rules TC

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc
Some fixes done for TC001,TC002 and TC003 but these rules are excluded since they don't play well with Pydantic.


efriis (commented):
given the frequency with which this needs to be `noqa` for pydantic, I think this should be closed. Thoughts?

cbornet:
> given the frequency with which this needs to be noqa for pydantic, I think this should be closed. Thoughts?

I excluded TC001,TC002 and TC003 so there won't be conflicts with Pydantic.
The other TC rules are still relevant and useful.

ccurme (approved):

PR #29271, opened 2025-01-17
core: Add ruff rules ANN (type annotations)

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-annotations-ann
The interest compared to only mypy is that ruff is very fast at detecting missing annotations.

ANN101 and ANN102 are deprecated so we ignore them 
ANN401 (no Any type) ignored to be in sync with mypy config


libs/core/langchain_core/_api/deprecation.py
```diff
@@ -271,28 +271,40 @@ def finalize(wrapper: Callable[..., Any], new_doc: str) -> T:
...
                     self.__orig_fset = fset
                     self.__orig_fdel = fdel
 
-                def __get__(self, instance, owner=None):  # type: ignore[no-untyped-def]
+                def __get__(
+                    self, instance: Any, owner: Union[type, None] = None
+                ) -> Any:
                     if instance is not None or owner is not None:
                         emit_warning()
+                    if self.fget is None:
```
cbornet, line 291: There was probably a bug here.
Other possibility could be to call `super().__get__` ?

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2186,7 +2186,7 @@ class Foo(BaseModelV2):
         )
 
     @tool(args_schema=Foo)
-    def foo(x):  # type: ignore[no-untyped-def]
+    def foo(x: list[int]) -> list[int]:
```
eyurtsev, line 2189: Probably reasonable to not update this type signature -- this test is attempting to verify that all the information is getting picked up from the args_schema?

eyurtsev (approved):

libs/core/tests/unit_tests/test_tools.py
cbornet: Indeed!
Done in https://github.com/langchain-ai/langchain/pull/29271/commits/454b292ba46aff42779406d7c002589481f96f03

cbornet:
Hi @eyurtsev, do you think this can be merged ?

PR #29286, opened 2025-01-18
core: Add ruff rules for Pylint PLC (Convention) and PLE (Errors)

cbornet:
See https://docs.astral.sh/ruff/rules/#pylint-pl

### DIFF
--- a/libs/core/langchain_core/output_parsers/list.py
+++ b/libs/core/langchain_core/output_parsers/list.py
@@ -8,7 +8,6 @@ from abc import abstractmethod
 from collections import deque
 from io import StringIO
 from typing import TYPE_CHECKING, TypeVar, Union
-from typing import Optional as Optional
 
 from typing_extensions import override
 
--- a/libs/core/langchain_core/tools/__init__.py
+++ b/libs/core/langchain_core/tools/__init__.py
@@ -20,45 +20,51 @@ tool for the job.
 from __future__ import annotations
 
 from langchain_core.tools.base import (
-    FILTERED_ARGS as FILTERED_ARGS,
-)
-from langchain_core.tools.base import (
-    ArgsSchema as ArgsSchema,
-)
-from langchain_core.tools.base import (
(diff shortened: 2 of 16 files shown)


libs/core/langchain_core/tools/__init__.py
```diff
@@ -20,42 +20,49 @@
 from __future__ import annotations
 
 from langchain_core.tools.base import (
-    FILTERED_ARGS as FILTERED_ARGS,
-)
-from langchain_core.tools.base import (
-    BaseTool as BaseTool,
```
cbornet, line 26: Not sure why there were all these import renaming to the same name 🤔.
Is there a good reason for these ?

libs/core/langchain_core/output_parsers/list.py
```diff
@@ -6,7 +6,6 @@
 from collections import deque
 from collections.abc import AsyncIterator, Iterator
 from io import StringIO
-from typing import Optional as Optional
```
eyurtsev, line 9: These were intentionally due to pydantic forward references not being properly resolved. Do you know if the underlying issue is resolved? We'd potentially be seeing issues when running langchain or langchain community unit tests

cbornet: The tests pass, so I guess it's OK ?
Do you have the pydantic reference issue ?

cbornet:
Hi, could this be reviewed ?

libs/core/langchain_core/output_parsers/list.py
eyurtsev: got it -- if our community tests are passing as well, probably OK

libs/core/langchain_core/tools/__init__.py
```diff
@@ -20,45 +20,51 @@
 from __future__ import annotations
```
eyurtsev, line 20: Thank you!

eyurtsev (approved):
good to go

PR #29288, opened 2025-01-18
core: Add ruff rules for PLW (Pylint Warnings)

cbornet:
See https://docs.astral.sh/ruff/rules/#warning-w_1


libs/core/pyproject.toml
```diff
@@ -75,6 +75,7 @@ optional = true
 classmethod-decorators = [ "classmethod", "langchain_core.utils.pydantic.pre_init", "pydantic.field_validator", "pydantic.v1.root_validator",]
 
 [tool.ruff.lint.per-file-ignores]
+"langchain_core/utils/mustache.py" = [ "PLW0603",]
```
cbornet, line 78: Maybe this rule could be ignored completely ?

eyurtsev (approved):

libs/core/langchain_core/output_parsers/json.py
```diff
@@ -13,9 +13,9 @@
 from langchain_core.output_parsers.transform import BaseCumulativeTransformOutputParser
 from langchain_core.outputs import Generation
 from langchain_core.utils.json import (
-    parse_and_check_json_markdown,
+    parse_and_check_json_markdown,  # noqa: F401 backward compatibility
```
eyurtsev, line 16: hmm i don't think this is the correct way to specify that this is for backwards compatibility

eyurtsev (commented):
actually need to review this in more detail

libs/core/langchain_core/output_parsers/json.py
```diff
@@ -123,7 +123,9 @@ def _type(self) -> str:
         return "simple_json_output_parser"
 
 
-# For backwards compatibility
-SimpleJsonOutputParser = JsonOutputParser
-parse_partial_json = parse_partial_json
-parse_and_check_json_markdown = parse_and_check_json_markdown
+__all__ = [
```
eyurtsev, line 126: @cbornet i think __all__ or an explicit redclare of the import is better (`from blah import foo as foo`)

cbornet: Sure. I think SimpleJsonOutputParser should stay for backward compatibility ?

PR #29289, opened 2025-01-18
community(doc_loaders): allow any credential type in AzureAIDocumentI…

ianchi:
allow any credential type in AzureAIDocumentInteligence, not only `api_key`.
This allows to use any of the credentials types integrated with AD.

### DIFF
--- a/libs/community/langchain_community/document_loaders/doc_intelligence.py
+++ b/libs/community/langchain_community/document_loaders/doc_intelligence.py
@@ -29,7 +29,7 @@ class AzureAIDocumentIntelligenceLoader(BaseLoader):
         mode: str = "markdown",
         *,
         analysis_features: Optional[List[str]] = None,
-        azure_credentials: Optional["TokenCredential"] = None,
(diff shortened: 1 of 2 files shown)


libs/community/langchain_community/document_loaders/doc_intelligence.py
```diff
@@ -24,6 +29,7 @@ def __init__(
         mode: str = "markdown",
         *,
         analysis_features: Optional[List[str]] = None,
+        credentials: Optional["TokenCredential"] = None,
```
ccurme, line 32: Could we name this `azure_credential`, for consistency with what was done here: https://github.com/langchain-ai/langchain/pull/28873

ccurme (commented):
Thanks, looks good!

libs/community/langchain_community/document_loaders/doc_intelligence.py
ianchi: updated

ccurme (approved):

PR #29318, opened 2025-01-20
community: adds support for getting github releases for the configured repository

MacsDickinson:
**Description:** adds support for github tool to query github releases on the configure respository
**Issue:** N/A
**Dependencies:** N/A
**Twitter handle:** @macsdickinson

### DIFF
--- a/docs/docs/integrations/tools/github.ipynb
+++ b/docs/docs/integrations/tools/github.ipynb
@@ -204,6 +204,37 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Include release tools\n",
+    "\n",
+    "By default, the toolkit does not include release-related tools. You can include them by setting `include_release_tools=True` when initializing the toolkit:"
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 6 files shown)


ccurme (commented):
This is a somewhat breaking change. Consider users who are running agents that access tools in this toolkit. If you are adding tools, you are giving these agents new capabilities that users may not desire or be aware of. What do you think about gating these behind a flag? e.g., `include_release_tools: bool = False`.

We should also document them in [this page](https://python.langchain.com/docs/integrations/tools/github/) (can be modified [here](https://github.com/langchain-ai/langchain/blob/master/docs/docs/integrations/tools/github.ipynb)).

MacsDickinson:
added the flag which is defaulted to False. This can be used when instantiating the toolkit

```
github = GitHubAPIWrapper()
toolkit = GitHubToolkit.from_github_api_wrapper(github, include_release_tools = True)
```

Let me know if this isn't the preferred place to apply this, I'm not familiar with the convention

ccurme (approved):

PR #29333, opened 2025-01-21
Update rag.ipynb with missing pip dependency langchain_openai

adiguno:

- [ ] **PR title**: "docs: add missing pip dependency `langchain_openai` to `rag.ipynb"

    - **Description:** add missing pip dependency `langchain_openai`
    - **Twitter handle:** dian_tweets


ccurme (commented):
The pip install is already rendered if you select OpenAI chat models or embeddings: https://python.langchain.com/docs/tutorials/rag/

We should not hard-code it because it is only needed if these components are selected.

PR #29335, opened 2025-01-21
core: Add ruff rules PYI

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi


cbornet:
Hi, could this PR be reviewed ?

libs/core/pyproject.toml
```diff
@@ -103,3 +96,53 @@ classmethod-decorators = [ "classmethod", "langchain_core.utils.pydantic.pre_ini
 "tests/unit_tests/runnables/test_graph.py" = [ "E501",]
 "tests/**" = [ "S",]
 "scripts/**" = [ "INP", "S",]
+
+[tool.ruff.lint]
+ignore = [ "ANN401", "COM812", "FA100", "ISC001", "S110", "S112", "TC001", "TC002", "TC003", "UP007",]
+flake8-type-checking.runtime-evaluated-base-classes = ["pydantic.BaseModel","langchain_core.load.serializable.Serializable","langchain_core.runnables.base.RunnableSerializable"]
+flake8-annotations.allow-star-arg-any = true
+flake8-annotations.mypy-init-return = true
+select = [
```
eyurtsev, line 105: @cbornet adding new lines and alphabetical to avoid merge conflicts

libs/core/langchain_core/utils/pydantic.py
```diff
@@ -491,28 +486,30 @@ def _create_root_model_cached(
...
+        model_name,
         __config__=_SchemaConfig,
         **_remap_field_definitions(field_definitions),
     )
 
 
 def create_model(
-    __model_name: str,
-    __module_name: Optional[str] = None,
+    model_name: str,
```
eyurtsev, line 501: we don't want this change I think -- it's a public function and it may conflict with field_definitions

libs/core/langchain_core/utils/pydantic.py
```diff
@@ -491,28 +486,30 @@ def _create_root_model_cached(
 
 @lru_cache(maxsize=256)
 def _create_model_cached(
-    __model_name: str,
+    model_name: str,
```
eyurtsev, line 489: same here

cbornet: Isn't this one private ?

libs/core/langchain_core/utils/pydantic.py
cbornet: The double-underscore is a convention for positional arguments.
Although it's possible to use it non-positionally, it would be quite weird to do so.
The change replaces it with a runtime-checked positional argument thanks to the `/` in the args.
So it is equivalent in the intention and normal usage. But breaking for users who really used it incorrectly.
We can also note that this function is deprecated and `create_model_v2` should be used instead. 
I think it doesn't conflict with `field_definitions` as it concerns positional args, not kwargs.

eyurtsev: `field_definitions` may have a field called "model_name" I think?

eyurtsev: I was probably the last one to deal w/ the refactor here -- the concern was working around a bad namespacing API where the positional arguments could collide w/ content in field_definitions over which we don't have control

cbornet: That would be unfortunate but OK. I'll keep the double-underscore then.

eyurtsev: TIL

eyurtsev (approved):

PR #29370, opened 2025-01-23
core[patch]: fix API reference for draw_ascii

WrRan:
typo: no `draw` but `draw_ascii` and other things

now, it works:
image


ccurme (approved):
Thank you!

PR #29381, opened 2025-01-23
core: Add ruff rules PT (pytest)

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt

### DIFF
--- a/libs/core/tests/unit_tests/language_models/chat_models/test_base.py
+++ b/libs/core/tests/unit_tests/language_models/chat_models/test_base.py
@@ -99,6 +99,7 @@ async def test_async_batch_size(messages: list, messages_2: list) -> None:
         assert (cb.traced_runs[0].extra or {}).get("batch_size") == 1
 
 
+@pytest.mark.xfail(reason="This test is failing due to a bug in the testing code")
 async def test_stream_error_callback() -> None:
     message = "test"
 
--- a/libs/core/tests/unit_tests/prompts/test_chat.py
+++ b/libs/core/tests/unit_tests/prompts/test_chat.py
@@ -1,3 +1,4 @@
+import re
 import warnings
 from pathlib import Path
 from typing import Any, Union, cast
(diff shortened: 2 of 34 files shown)


libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -95,6 +95,7 @@ async def test_async_batch_size(messages: list, messages_2: list) -> None:
         assert (cb.traced_runs[0].extra or {}).get("batch_size") == 1
 
 
+@pytest.mark.xfail(reason="This test is failing due to a bug in the testing code")
```
cbornet, line 98: The linter found an issue in the test code here.
It was not seen before because `eval_response` was never executed.

libs/core/tests/unit_tests/language_models/llms/test_base.py
```diff
@@ -93,6 +93,7 @@ async def test_async_batch_size() -> None:
         assert (cb.traced_runs[0].extra or {}).get("batch_size") == 1
 
 
+@pytest.mark.xfail(reason="This test is failing due to a bug in the testing code")
```
cbornet, line 96: The linter found an issue in the test code here.
It was not seen before because eval_response was never executed.

libs/core/tests/unit_tests/prompts/test_chat.py
```diff
@@ -167,15 +168,14 @@ def test_create_system_message_prompt_list_template_partial_variables_not_null()
...
-        graph_analyst_template = SystemMessagePromptTemplate.from_template(
+    with pytest.raises(
+        ValueError, match="Partial variables are not supported for list of templates."
+    ):
+        _ = SystemMessagePromptTemplate.from_template(
             template=[graph_creator_content1, graph_creator_content2],
             input_variables=["variables"],
             partial_variables={"variables": "foo"},
         )
-        graph_analyst_template.format(variables="foo")
```
cbornet, line 176: This call was not executed.

libs/core/tests/unit_tests/language_models/llms/test_base.py
cbornet: This is fixed by #30228 .

eyurtsev (approved):

PR #29384, opened 2025-01-23
core: Add ruff rules RET

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-return-ret
All auto-fixes


eyurtsev (approved):
read through all changes. looks good

PR #29388, opened 2025-01-23
core: Add ruff rules TRY (tryceratops)

cbornet:
TRY004 ("use TypeError rather than ValueError") existing errors are marked as ignore to preserve backward compatibility.
LMK if you prefer to fix some of them.


efriis (approved):
seems reasonable

PR #29391, opened 2025-01-23
standard-tests: Fix ToolsIntegrationTests to correctly handle "content_and_artifact" tools

rectalogic:
**Description:**

The response from `tool.invoke()` is always a ToolMessage, with content and artifact fields, not a tuple.
The tuple is converted to a ToolMessage here
https://github.com/langchain-ai/langchain/blob/b6ae7ca91dd0f54f58475463804540ad25eeea7d/libs/core/langchain_core/tools/base.py#L726

**Issue:**

Currently `ToolsIntegrationTests` requires `invoke()` to return a tuple and so standard tests fail for "content_and_artifact" tools. This fixes that to check the returned ToolMessage.

This PR also adds a test that now passes.

### DIFF
--- a/libs/standard-tests/langchain_tests/integration_tests/tools.py
+++ b/libs/standard-tests/langchain_tests/integration_tests/tools.py
@@ -29,15 +29,10 @@ class ToolsIntegrationTests(ToolsTests):
         )
         result = tool.invoke(tool_call)
 
-        if tool.response_format == "content":
-            tool_message = result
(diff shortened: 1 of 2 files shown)


rectalogic:
@efriis could you take a look at this when you get a chance? Currently `ToolsIntegrationTests` don't seem usable for "content_and_artifact" tools.

ccurme (approved):

PR #29398, opened 2025-01-23
anthropic[patch]: always return content blocks if citations are generated

ccurme:
We currently return string (and therefore no content blocks / citations) if the response is of the form
```
[
    {"text": "a claim", "citations": [...]},
]
```

There are other cases where we do return citations as-is:
```
[
    {"text": "a claim", "citations": [...]},
    {"text": "some other text"},
    {"text": "another claim", "citations": [...]},
]
```
Here we update to return content blocks including citations in the first case as well.

### DIFF
--- a/libs/partners/anthropic/langchain_anthropic/chat_models.py
+++ b/libs/partners/anthropic/langchain_anthropic/chat_models.py
@@ -764,7 +764,11 @@ class ChatAnthropic(BaseChatModel):
         llm_output = {
             k: v for k, v in data_dict.items() if k not in ("content", "role", "type")
         }
-        if len(content) == 1 and content[0]["type"] == "text":
+        if (
+            len(content) == 1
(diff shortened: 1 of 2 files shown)


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -764,7 +764,11 @@ def _format_output(self, data: Any, **kwargs: Any) -> ChatResult:
         llm_output = {
             k: v for k, v in data_dict.items() if k not in ("content", "role", "type")
         }
-        if len(content) == 1 and content[0]["type"] == "text":
+        if (
+            len(content) == 1
+            and content[0]["type"] == "text"
+            and "citations" not in content[0]
```
efriis, line 770: any chance of empty citations?

efriis (approved):

PR #29424, opened 2025-01-25
core: Add ruff rule FBT003 (boolean-trap)

cbornet:
See https://docs.astral.sh/ruff/rules/boolean-positional-value-in-call/#boolean-positional-value-in-call-fbt003
This PR also fixes some FBT001/002 in private methods but does not enforce these rules globally atm.


eyurtsev (approved):
Looks good

PR #29427, opened 2025-01-26
community: Add goodfire chat model

keenanpepper:
- **Description:** Add a chat model implementation for Goodfire
- **Issue:** n/a
- **Dependencies:** goodfire package as an optional dependency
- **Twitter handle:** keenanpepper


ccurme (commented):
Hello, thanks for this!

This adds a new integration into `langchain-community`. We are recommending that these be implemented via external packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-goodfire`)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the model, and you have control over versioning (including of the `goodfire` dependency).

Let me know what you think!

keenanpepper:
Ah, thanks @ccurme I didn't know what the preferred structure was.

Issue #29431, opened 2025-01-27
Llama-3.2-3B-Instruct failed to use with HuggingfacePipeline because of setting a non-string value as the pad_token


tishizaki:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code:

```python
from langchain_community.llms import HuggingFacePipeline

hf = HuggingFacePipeline.from_model_id(
    model_id="meta-llama/Llama-3.2-3B-Instruct",
    task="text-generation",
    pipeline_kwargs={"max_new_tokens": 10},
)

from langchain_core.prompts import PromptTemplate

template = """Question: {question}

Answer: Let's think step by step."""
prompt = PromptTemplate.from_template(template)

chain = prompt | hf

question = "What is electroencephalography?"

print(chain.invoke({"question": question}))
```

### Error Message and Stack Trace (if applicable)

```
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00,  2.69it/s]
Traceback (most recent call last):
  File "/home/ishi/work/hf_pipeline_sample.py", line 4, in <module>
    hf = HuggingFacePipeline.from_model_id(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ishi/work/langchain/libs/community/langchain_community/llms/huggingface_pipeline.py", line 172, in from_model_id
    tokenizer.pad_token_id = model.config.eos_token_id
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ishi/work/.venv/lib/python3.12/site-packages/transformers/tokenization_utils_base.py", line 1077, in __setattr__
    raise ValueError(f"Cannot set a non-string value as the {key}")
ValueError: Cannot set a non-string value as the pad_token
```

### Description

I tried to do HuggingfacePipeline  of langchain_community with Llama-3.2-3B-Instruct, but the error occured.

I think that it's the same bug as transformers issue [34869](https://github.com/huggingface/transformers/issues/34869).

[...]

PR #29434, opened 2025-01-27
community: Fixed the procedure of initializing pad_token_id

tishizaki:
- **Description:** Add to check pad_token_id and eos_token_id of model config. It seems that this is the same bug as the HuggingFace TGI bug. In addition, the source code of libs/partners/huggingface/langchain_huggingface/llms/huggingface_pipeline.py also requires similar changes.
- **Issue:** #29431
- **Dependencies:** none
- **Twitter handle:** tell14


ccurme (approved):
Would welcome a follow-up adding this to `libs/partners/huggingface` (i.e., the `langchain-huggingface` library), with tests.

PR #29451, opened 2025-01-28
langchain: release 0.3.16

efriis:


arian-gogani:
the issue here is that callbacks fire after execution -- they tell you what happened, not what was authorized to happen. an agent that bypasses a callback still runs the tool; the callback just doesn't fire.

the primitive that actually enforces this: sign the intent before execution. a pre-execution receipt (action_ref = SHA-256 of tool_name + input_hash + agent_id + scope, Ed25519 signed) must exist before the tool runs. no signed pre-receipt = no execution. the receipt chain is the enforcement mechanism, not the logging mechanism.

this also gives you the third-party verifiability that callbacks don't -- a verifier with the public key and the receipt chain can confirm what happened without access to the callback logs.

nobulex implements this for LangChain: github.com/arian-gogani/nobulex (ComplianceCallbackHandler, LangChain RFC #35691)

PR #29494, opened 2025-01-30
Structured Output Support for Chat Deepseek

keenborder786:
I have added support for structured output for ChatDeepSeek as per their official documentation: https://api-docs.deepseek.com/guides/json_mode


keenborder786:
@ccurme

ccurme (commented):
ChatDeepSeek should already support structured output since it inherits from BaseChatOpenAI.
```python
from langchain_deepseek import ChatDeepSeek

class Joke(BaseModel):
    """Joke to tell user."""

llm = ChatDeepSeek(model="deepseek-chat").with_structured_output(Joke)
result = llm.invoke("Tell me a joke about cats.")
```

PR #29504, opened 2025-01-30
text-splitters: fix bug in TextSplitter's create_documents method 

DmitrySirakov:
Description:
Fixed a bug in the `TextSplitter` class within the `create_documents` method. When `_add_start_index` is set to `True`, the length of the chunk was measured using the `len()` function (which counts characters) instead of the user-defined` self.length_function()`. This inconsistency could lead to incorrect values in `metadata['start_index']`.

Solution:
Replaced `len(chunk)` with `self.length_function(chunk)` to ensure that the chunk length is measured consistently using the user-provided function, thereby maintaining the correctness of `metadata['start_index']`.

### DIFF
--- a/libs/text-splitters/langchain_text_splitters/base.py
+++ b/libs/text-splitters/langchain_text_splitters/base.py
@@ -82,7 +82,7 @@ class TextSplitter(BaseDocumentTransformer, ABC):
                     offset = index + previous_chunk_len - self._chunk_overlap
                     index = text.find(chunk, max(0, offset))
                     metadata["start_index"] = index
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this. I think the index is intended to be in characters as opposed to units of the custom length function:

- `index` is calculated with `text.find`, which is counting characters
- `self._chunk_overlap` also appears in the calculation, and this is documented as `"Overlap in characters between chunks"`.

Closing for now but let me know if I've misunderstood this or if you can add a test case.

Issue #29470, opened 2025-01-29
PyMuPDF changes from 0.3.15 mess with existing Vectorstores


KarlCJ:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
docs = [doc for doc in PyMuPDFLoader(file_path=file_path, extract_images=False).load()]

documents = SemanticChunker(
            breakpoint_threshold_type="gradient",
            embeddings=embeddings,
            breakpoint_threshold_amount=settings.chunking_breakpoint_threshold_amount,
            min_chunk_size=settings.min_chunk_size,
        ).split_documents([doc for doc in docs if doc.page_content])

index(
                    docs_source=documents,
                    vector_store=self.store,
                    record_manager=self.record_manager,
                    cleanup=None,
                )
```

### Error Message and Stack Trace (if applicable)

ERROR:langchain_weaviate.vectorstores:indexer-0:Failed to add object: None
Reason: WeaviateInsertManyAllFailedError("Every object failed during insertion. Here is the set of all errors: no such prop with name 'creationdate' found in class 'Default_index_openai_text_embedding_3_small' in the schema. Check your schema files for which properties in this class are available\nno such prop with name 'moddate' found in class 'Default_index_openai_text_embedding_3_small' in the schema. Check your schema files for which properties in this class are available")

### Description

With the changes from langchain_community 0.3.15 the metadata keys have changed from modDate and creationDate to moddate and creationdate.
This causes issues with already existing vectorstore schemas(in my case Weaviate). Error is below.
This causes the index() function from langchain.indexes to fail but not raise an exception, instead it return value describes that everything got indexed correctly which it did not. The exception is only visible in the log and cannot be catched.

For my project I fixed this by mapping the new metadata keys to the old ones because weaviate refuses to accept the new keys to an already existing schema.

reproducible by creating a vectorstore and indexing a pdf (loaded with PyMuPDFLoader) with langchain_community <= 0.3.14 and after updating to langchain_community >= 0.3.15 trying to index a pdf the same way.

### System Info

System Information
------------------
> OS:  Linux
[...]

PR #29526, opened 2025-01-31
community[minor]: 04 - Refactoring PDFMiner parser

pprados:
This is one part of a larger Pull Request (PR) that is too large to be submitted all at once. This specific part focuses on updating the PDFMiner parser.

For more details, see [PR 28970](https://github.com/langchain-ai/langchain/pull/28970).


pprados:
@eyurtsev
The next one ;-)

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -464,120 +465,335 @@ def extract_images_from_page(self, page: pypdf._page.PageObject) -> str:
...
+    def __init__(
+        self,
+        extract_images: bool = False,
+        *,
+        password: Optional[str] = None,
+        mode: Literal["single", "page"] = "single",
+        pages_delimiter: str = _DEFAULT_PAGES_DELIMITER,
+        images_parser: Optional[BaseImageBlobParser] = None,
+        images_inner_format: Literal["text", "markdown-img", "html-img"] = "text",
+        concatenate_pages: Optional[bool] = None,
```
eyurtsev, line 529: This is a breaking change w/ respect due to the change in the default. 

Could we restore to how it was, update unit tests w/ an exception for this class (i.e., specifically pass concatenate_pages=False)

---

For making breaking changes in the API, let's not do this as part of the large PRs since they're hard to catch, and we'll want to document them properly / figure out how to roll them out.

eyurtsev: It's OK if the content of the Document changes due to improvements (e.g., the content is extracted in a format that better matches the PDF). We just don't want to cause silent failures in user code; e.g., 

user assumed it would be a single page and they're code does something like extract only the first document emitted

eyurtsev:
need to resolve: https://github.com/langchain-ai/langchain/issues/29470 to make sure this and other loaders aren't affected

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: @eyurtsev 
What do you think?

eyurtsev: Ah got it. OK looks fine!

eyurtsev (approved):

Issue #29541, opened 2025-02-02
SQLDatabaseToolkit adds temperature=0 by default and breaks the execution when using 03-mini


aifa:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

In langchain_community/agent_toolkits/sql/toolkit.py, class  SQLDatabaseToolkit(BaseToolkit): line 48:     

Instantiate:
        .. code-block:: python

            from langchain_community.agent_toolkits.sql.toolkit import SQLDatabaseToolkit
            from langchain_community.utilities.sql_database import SQLDatabase
            from langchain_openai import ChatOpenAI

            db = SQLDatabase.from_uri("sqlite:///Chinook.db")
            llm = ChatOpenAI(temperature=0)

            toolkit = SQLDatabaseToolkit(db=db, llm=llm)

This instructs the LLM to generate code using temperature=0. When o3-mini is used ,then this causes:
 BadRequestError('Error code: 400 - {\'error\': {\'message\': "Unsupported parameter: \'temperature\' is not supported with this model.", 

I removed the temperature=0 from my local copy of the code and instruct it to not use temperature with o3-mini and the error stopped. 
It also works when I pass the instruction not to set temperature in the user/system prompt directly.

### Error Message and Stack Trace (if applicable)

BadRequestError('Error code: 400 - {\'error\': {\'message\': "Unsupported parameter: \'temperature\' is not supported with this model.", \'type\': \'invalid_request_error\', \'param\': \'temperature\', \'code\': \'unsupported_parameter\'}}')Traceback (most recent call last):

  File "/root/.cache/pypoetry/virtualenvs/gix-service-9TtSrW0h-py3.12/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 637, in generate
    self._generate_with_cache(

  File "/root/.cache/pypoetry/virtualenvs/gix-service-9TtSrW0h-py3.12/lib/python3.12/site-packages/langchain_core/language_models/chat_models.py", line 855, in _generate_with_cache
    result = self._generate(
             ^^^^^^^^^^^^^^^

[...]

PR #29546, opened 2025-02-02
Fix temperature parameter will not be consider for reasoning models

rawathemant246:
- **Description:** The OpenAI reasoning model does not support the `temperature` parameter. Attempting to set it results in a `BadRequestError`. This commit modifies the ChatOPENAI api  will not consider temperature parameter for reasoning models.
    - **Issue:** #29541
    - **Dependencies:** None

Additional guidelines:
- Optional dependencies are imported within a function.
- No dependencies were added to pyproject.toml files.
- This PR modifies only the `community` package.
- Changes are backwards compatible.


ccurme (commented):
Duplicated with https://github.com/langchain-ai/langchain/pull/29551

Issue #29555, opened 2025-02-03
Improve encoding handling in loader parent class to prevent garbled text across multiple loaders


engkimo:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from __future__ import annotations
import os
import tempfile
import argparse
import asyncio
from io import BytesIO
from dotenv import load_dotenv
from langchain_community.document_loaders import PDFMinerLoader
from langchain.text_splitter import CharacterTextSplitter
from langchain_openai import OpenAI
from langchain.prompts import PromptTemplate
from langchain_core.documents import Document

class BaseOpenAIClient:
    def __init__(self):
        self.api_key = os.getenv("OPENAI_API_KEY")
        if not self.api_key:
            raise EnvironmentError("OPENAI_API_KEY is not set")
        # Token and chunk settings
        self.max_completion_tokens = 600
        self.chunk_size = 100
        self.chunk_overlap = 10

    async def extract_text_from_file(self, file_buffer: BytesIO, file_extension: str) -> str:
        if not file_buffer or file_buffer.getbuffer().nbytes == 0:
            raise ValueError("Cannot read an empty file")
        with tempfile.NamedTemporaryFile(suffix=f".{file_extension}", delete=False) as temp_file:
            file_buffer.seek(0)
            temp_file.write(file_buffer.read())
            temp_file_path = temp_file.name
            print(f"Temporary file: {temp_file_path}")
        try:
            loader = PDFMinerLoader(temp_file_path)
            pages = loader.load()
            if not pages:
                raise ValueError("Failed to extract text from PDF")
            text_content = []
            for page in pages:
                if hasattr(page, 'page_content'):
                    text_content.append(page.page_content)
                else:
                    text_content.append(str(page))
            combined_text = " ".join(text_content)
            # Normalize text encoding to UTF-8, replacing problematic characters
            combined_text = combined_text.encode("utf-8", errors="replace").decode("utf-8")
            return combined_text
        finally:
            os.unlink(temp_file_path)

class CareerSummaryClient(BaseOpenAIClient):
    def __init__(self):
        super().__init__()
        self.llm = OpenAI(
            openai_api_key=self.api_key,
[...]

PR #29557, opened 2025-02-03
fix(pdf): Fix garbled text in PDF loaders

engkimo:
Hi team,

This PR implements the first part of the fix for the garbled text issue (#29555), starting with PDF loaders. This is part of a larger effort to address encoding issues across multiple document loaders.

スクリーンショット 2025-02-04 0 22 43

## Changes
- Add `PDFEncoding` enum for standardized encoding options
- Implement `PDFEncodingConfig` for centralized encoding management
- Enhance `BasePDFLoader` with encoding configuration support
- Update `PDFMinerParser` to handle encoding properly
- Add fallback mechanism for encoding errors

## Next Steps
This PR focuses on PDF loaders as the first implementation. Similar fixes will be needed for:
- Docx2txtLoader
- PyPDFLoader
- TextLoader
- UnstructuredExcelLoader
- UnstructuredWordDocumentLoader

## Notes
- No breaking changes to existing API
- Default encoding remains UTF-8
- Users can now specify custom encoding configurations when needed
- Fixes regression introduced in o3 release for PDF loaders

## Environment
- Tested on Python 3.13.1
- Compatible with:
  - langchain 0.3.17
  - langchain_community 0.3.16
  - langchain_core 0.3.33

Related issue: #29555

### DIFF
--- a/libs/community/langchain_community/document_loaders/parsers/pdf.py
+++ b/libs/community/langchain_community/document_loaders/parsers/pdf.py
@@ -21,6 +21,11 @@ from typing import (
     cast,
 )
 from urllib.parse import urlparse
+from enum import Enum
+from dataclasses import dataclass
+from abc import ABC
+from typing import Protocol
+from io import StringIO
 
 import numpy
 import numpy as np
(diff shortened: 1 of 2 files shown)


efriis:
hey there! is the issue that there _isn't_ a way to pass in an encoding to these loaders currently, or is the issue that there isn't an encoding abstraction that is compatible with these different loaders?

I would rather just make sure we're passing through kwargs you might want to use to configure these underlying libraries, and skip standardizing the PDF encoding definition because different PDF libraries might have other flags you might want to configure as well, and we're not aiming to be a full abstraction layer for all pdf parsing libraries' configuration

engkimo:
@efriis 
Hi! Thank you for the clarification. The encoding issues affect multiple loaders (PDFMinerLoader, Docx2txtLoader, PyPDFLoader, TextLoader, UnstructuredExcelLoader, UnstructuredWordDocumentLoader, etc.).

Would this approach work better?
1. Add **kwargs to the BaseLoader class __init__
2. Pass these kwargs through to each underlying document parsing library
3. Document the encoding options available for each loader type

This would:
- Maintain flexibility for each document library's unique features
- Allow consistent encoding configuration across different loader types
- Keep LangChain's role as a thin wrapper rather than a full abstraction layer

Does this align better with LangChain's philosophy while addressing the encoding issues across multiple loader types?

ccurme (commented):

PR #29559, opened 2025-02-03
core: allow for `tool_choice` in `with_structured_output` to be overriden by client wrappers

christianarty:
### Description

This change allows for wrappers around `BaseChatModel` to change the `tool_choice` parameter.  This will default to what is was set prior to this change, however, if client libraries want to change it, they can now do so.

## Issue
When using a custom wrapper with `BaseChatModel`, i encountered an issue where there was an incompatibility. 	The default implemantation of `with_structured_output`, hard coded an `any` key, when that key can be other values. 

With my wrapper, it uses the OpenAI spec, and in the spec [`any` is not a valid option for the `tool_choice` parameter](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice)

### Dependencies
None

### Twitter handle
@christian_arty

### Additional request
Could this change, if approved, be backported to 0.2.* (e.g. 0.2.44).

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -1193,6 +1193,7 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
         schema: Union[typing.Dict, type],  # noqa: UP006
         *,
         include_raw: bool = False,
+        tool_choice: Optional[str] = "any",
         **kwargs: Any,
     ) -> Runnable[LanguageModelInput, Union[typing.Dict, BaseModel]]:  # noqa: UP006
         """Model wrapper that returns outputs formatted to match the given schema.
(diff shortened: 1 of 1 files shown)


ccurme (commented):
I think we'd like to support `.bind_tools([schema], tool_choice="any")` across models. My suggestion in this case would be to override `bind_tools` and translate `"any"` to `"required"` or similar. Would this work?

christianarty:
> I think we'd like to support `.bind_tools([schema], tool_choice="any")` across models. My suggestion in this case would be to override `bind_tools` and translate `"any"` to `"required"` or similar. Would this work?

@ccurme This PR doesn't change that method call at all; if clients want to pass in a `tool_choice` or override the `.bind_tools` fn in their wrapper, they can. But if they don't, the functionality would be the same. Effectively, this extends the flexibility of this base class. This way, clients can maintain the default behavior or customize it as needed without adding extra complexity of overriding other classes. 

With your suggestion, we would have to ignore what was passed into the function and override in the function our own `tool_choice`, which in my opinion would be an area for error and unnecessary. Let me know your thoughts!

PR #29561, opened 2025-02-03
Community/Partner: Adding Azure community and partner user agent to better track usage in Python 

marlenezw:
- This pull request includes various changes to add a `user_agent` parameter to Azure OpenAI, Azure Search and Whisper in the Community and Partner packages. This helps in identifying the source of API requests so we can better track usage and help support the community better. I will also be adding the user_agent to the new `langchain-azure` repo as well. 

- No issue connected or  updated dependencies. 
- Utilises existing tests and docs


marlenezw:
@sinedied @santiagxf I need to run linting and tests but could you both have a look. Will add to the `langchain-azure` package as well.

marlenezw:
@santiagxf  I think there are certain classes for example the AzureOpenAI class within OpenAI that I think is useful to leave as is. My feeling is that we should also give people a depreciation timeline so they know to transition over. I'd also like to compare the numbers for a bit to see if the move makes any difference at all, so maybe a weeks or so of data could be helpful! What are your thoughts?

santiagxf:
I meant the ones on the Community package (langchain-community). Indeed the ones on partner packages should remain!

marlenezw:
That makes sense! Yes, for the community packages I definitely think we should remove the classes when we are ready to depreciate it completely. @santiagxf do you think we should set a final depreciation date and add this as a message in the code? I can create a seperate PR for it in the mean time.

marlenezw:
Hi @efriis or @ccurme, this PR is ready for review! I've run tests and things are working as expected.

efriis (approved):

PR #29591, opened 2025-02-04
anthropic[patch]: support citations in streaming

ccurme:


efriis (approved):

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -649,3 +649,23 @@ def test_citations() -> None:
...
+
+    streamed_citations = [
+        cast(dict, block)["citations"] for block in full.content if "citations" in block
+    ]
+    invoked_citations = [
+        cast(dict, block)["citations"]
+        for block in response.content
+        if "citations" in block
+    ]
+    assert streamed_citations == invoked_citations
```
efriis, line 671: are we certain these will come back identical? likely ends up flaky, even if we set temperature to 0.

might be worth just testing that the streamed ones return citations

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1254,6 +1258,18 @@ def _tools_in_params(params: dict) -> bool:
     )
 
 
+def _documents_in_params(params: dict) -> bool:
+    for message in params.get("messages", []):
+        for block in message.get("content", []):
```
jacoblee93, line 1263: I guess this will work but might be nice to explicitly check for a string

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1299,31 +1321,37 @@ def _make_message_chunk_from_anthropic_event(
...
-        if event.delta.type == "text_delta":
-            if coerce_content_to_string:
+        if event.delta.type in ("text_delta", "citations_delta"):
+            if coerce_content_to_string and hasattr(event.delta, "text"):
                 text = event.delta.text
                 message_chunk = AIMessageChunk(content=text)
             else:
                 content_block = event.delta.model_dump()
                 content_block["index"] = event.index
                 content_block["type"] = "text"
```
jacoblee93, line 1352: Why hardcode this?

PR #29595, opened 2025-02-05
feat(standard-tests): model param test

efriis:
features that depend on initialization with `model=`:
- standard interface documented as a ["standard parameter"](https://python.langchain.com/docs/concepts/chat_models/#standard-parameters)
- (optional/nice to have): standard model param for init_chat_model (alternatively could initialize elsewhere)

features that depend on accessing model by `.model`:
- default populating of `ls_params["ls_model_name"]`, which langsmith uses for token cost lookup etc (note currently there's a backup of `model_name` too but it's only those 2)

notes:
- I think it's ok to support this in addition to "more relevant" things like `deployment_name` on e.g. AzureChatOpenAI - this should still be configurable in addition to if you want to get token cost tracking


ccurme (commented):
Are there any impacts to serialization we need to be concerned about?

I think `load` if the model is saved with `"model_name"` should still work due to the alias.

efriis:
deserialization would still work. deserialization of new name with _old_ version of libraries would be breaking, and that's ok in my mind. happy to reverse alias/name in cases it's defined this way to keep serialization consistent, but feels like we'd want to do this at some point anyways

serialization will serialize the other name as mentioned

mdrxy:
@ccurme @eyurtsev how's this for 0.4?

PR #29609, opened 2025-02-05
community: Use Blockbuster to detect blocking calls in asyncio during tests

cbornet:
Same as https://github.com/langchain-ai/langchain/pull/29043 for langchain-community.

**Dependencies:**
- blockbuster (test)

**Twitter handle:** cbornet_


libs/community/langchain_community/utils/openai.py
```diff
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import functools
 from importlib.metadata import version
 
 from packaging.version import parse
 
 
+@functools.cache
```
cbornet, line 9: Caching `is_openai_v1` as it reads the package metadata file.

libs/community/tests/unit_tests/conftest.py
```diff
@@ -1,12 +1,28 @@
...
+from blockbuster import blockbuster_ctx
 from pytest import Config, Function, Parser
 
 
+@pytest.fixture(autouse=True)
+def blockbuster() -> Iterator[None]:
+    with blockbuster_ctx("langchain_community") as bb:
+        (
+            bb.functions["os.stat"]
+            .can_block_in("langchain_community/utils/openai.py", "is_openai_v1")
```
cbornet, line 17: Now cached so only the first call will block.

libs/community/tests/unit_tests/conftest.py
```diff
@@ -1,12 +1,28 @@
...
+            .can_block_in("httpx/_client.py", "_init_transport")
```
cbornet, line 18: Httpx has some blocking calls when initialising TLS async client.

efriis (approved):

PR #29627, opened 2025-02-06
community: add support for using GPUs with FastEmbedEmbeddings

vemonet:
- **Description:** add a `gpu: bool = False` field to the `FastEmbedEmbeddings` class which enables to use GPU (through ONNX CUDA provider) when generating embeddings with any fastembed model. It just requires the user to install a different dependency and we use a different provider when instantiating `fastembed.TextEmbedding`
- **Issue:** when generating embeddings for a really large amount of documents this drastically increase performance (honestly that is a must have in some situations, you can't just use CPU it is way too slow)
- **Dependencies:** no direct change to dependencies, but the users will need to install `fastembed-gpu` instead of `fastembed`, I made all the changes to the init function and docstrings to properly let the user know which dependency they should install depending on if they enabled `gpu` or not
 
cf. fastembed docs about GPU for more details: https://qdrant.github.io/fastembed/examples/FastEmbed_GPU/

I did not added test because it would require access to a GPU in the testing environment

### DIFF
--- a/libs/community/langchain_community/embeddings/fastembed.py
+++ b/libs/community/langchain_community/embeddings/fastembed.py
@@ -65,6 +65,14 @@ class FastEmbedEmbeddings(BaseModel, Embeddings):
     Defaults to `None`.
     """
 
+    gpu: bool = False
+    """Enable the use of GPU through CUDA. This requires to install `fastembed-gpu`
+    instead of `fastembed`. See https://qdrant.github.io/fastembed/examples/FastEmbed_GPU
+    for more details.
+    Defaults to False.
+    """
+
+
     model: Any = None  # : :meta private:
 
     model_config = ConfigDict(extra="allow", protected_namespaces=())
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/embeddings/fastembed.py
```diff
@@ -76,26 +83,30 @@ def validate_environment(cls, values: Dict) -> Dict:
...
         except ModuleNotFoundError:
             raise ImportError(
-                "Could not import 'fastembed' Python package. "
-                "Please install it with `pip install fastembed`."
+                f"Could not import '{pkg_to_import}' Python package. "
+                f"Please install it with `pip install {pkg_to_import}`."
             )
 
-        if importlib.metadata.version("fastembed") < MIN_VERSION:
+        if importlib.metadata.version(pkg_to_import) < MIN_VERSION:
```
ccurme, line 98: Assuming `fastembed` and `fastembed-gpu` versions are kept in sync? (They appear on the same minor version now.)

ccurme (approved):

libs/community/langchain_community/embeddings/fastembed.py
vemonet: @ccurme it seems like it is perfectly synced:
- https://pypi.org/project/fastembed/#history
- https://pypi.org/project/fastembed-gpu/#history

PR #29635, opened 2025-02-06
standard-tests[patch]: use `has_structured_output` property to engage structured output tests

ccurme:
Motivation: dedicated structured output features are becoming more common, such that integrations can support structured output without supporting tool calling.

Here we make two changes:

1. Update the `has_structured_output` method to default to True if a model supports tool calling (in addition to defaulting to True if `with_structured_output` is overridden).
2. Update structured output tests to engage if `has_structured_output` is True.

### DIFF
--- a/libs/standard-tests/langchain_tests/unit_tests/chat_models.py
+++ b/libs/standard-tests/langchain_tests/unit_tests/chat_models.py
@@ -133,7 +133,7 @@ class ChatModelTests(BaseStandardTests):
         return (
             self.chat_model_class.with_structured_output
             is not BaseChatModel.with_structured_output
-        )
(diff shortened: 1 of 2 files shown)


libs/standard-tests/langchain_tests/unit_tests/chat_models.py
```diff
@@ -133,7 +133,7 @@ def has_structured_output(self) -> bool:
         return (
             self.chat_model_class.with_structured_output
             is not BaseChatModel.with_structured_output
-        )
+        ) or self.has_tool_calling
```
efriis, line 136: nice

efriis (approved):
lgtm

Issue #29632, opened 2025-02-06
Extend support for OpenAI o3 style models in BaseChatOpenAI class


zachschillaci27:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

NA

### Description

The `BaseChatOpenAI` class of `langchain_openai` contains some hard-coded references to the `o1` model, such as the one below:
```
    @model_validator(mode="before")
    @classmethod
    def validate_temperature(cls, values: Dict[str, Any]) -> Any:
        """Currently o1 models only allow temperature=1."""
        model = values.get("model_name") or values.get("model") or ""
        if model.startswith("o1") and "temperature" not in values:
            values["temperature"] = 1
        return values
```
taken from https://github.com/langchain-ai/langchain/blob/master/libs/partners/openai/langchain_openai/chat_models/base.py#L558.

These need to be extended to support the `o3` series of models - namely `o3-mini` for now. Perhaps using `model.startswith("o")` is a good enough patch for now, but something more future-proof may want to be considered.

### System Info

I took the references directly from langchain's master branch on GitHub

ccurme:
`validate_temperature` was added in langchain-openai ~0.2, which set a default temperature of 0.7. So `o1` would break with the default value.

langchain-openai ~0.3 updated this default to None. I don't run into any issues using `o1`, `o1-preview`, `o1-mini`, or `o3-mini`.

What is the motivation for extending this?

zachschillaci27:
> `validate_temperature` was added in langchain-openai ~0.2, which set a default temperature of 0.7. So `o1` would break with the default value.
> 
> langchain-openai ~0.3 updated this default to None. I don't run into any issues using `o1`, `o1-preview`, `o1-mini`, or `o3-mini`.
> 
> What is the motivation for extending this?

Thanks for the quick response! I can better understand the context now.

I personally find it a bit confusing that the patch only handles `o1`. It's still the case that `temperature` is not a valid parameter for `o3-mini`, unless it's set to 1:
```
[...]

PR #29636, opened 2025-02-06
langchain : Extending support for OpenAI o series models from only o1 model

ashwantmanikoth:
**Description:**
This PR extends the `BaseChatOpenAI` model validation logic to support the `o3` series models (e.g., `o3-mini`). Previously, the validation was hardcoded for `o1` models only. Now supports all series that can start the name with 'o'.

This serves as a temporary workaround but can be configured in such a way to allow future model names in a set or a separate config file, making it more maintainable.  

**Issue:**  
  ID #29632

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -490,7 +490,7 @@ class BaseChatOpenAI(BaseChatModel):
     reasoning_effort: Optional[str] = None
     """Constrains effort on reasoning for reasoning models. 
     
-    o1 models only.
+    o series models only.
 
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this. I think there are a few issues:

1. `validate_temperature` was a hold-over from langchain-openai ~0.2, which had a default temperature of 0.7. langchain-openai ~0.3 updated this default to None, so o-series models no longer require temperature to be specified this way.

2. Importantly, `o1-preview`, `o1-mini`, and `o3-mini` all support streaming just fine, so we definitely should not disable streaming on them.

Also commented on the issue linked to this PR.

PR #29649, opened 2025-02-06
openai: Removed tool_calls from completion chunk after other chunks have already been sent.

marcammann:
- **Description:** Before sending a completion chunk at the end of an OpenAI stream, removing the tool_calls as those have already been sent as chunks.
- **Issue:** -
- **Dependencies:** -
- **Twitter handle:** -

@ccurme as mentioned in another PR

### DIFF
--- a/libs/partners/openai/tests/integration_tests/chat_models/test_base.py
+++ b/libs/partners/openai/tests/integration_tests/chat_models/test_base.py
@@ -1213,3 +1213,26 @@ def test_multi_party_conversation() -> None:
     ]
     response = llm.invoke(messages)
     assert "Bob" in response.content
+
+
+def test_structured_output_and_tools() -> None:
+    class ResponseFormat(BaseModel):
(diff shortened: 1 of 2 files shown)


libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -1213,3 +1213,26 @@ def test_multi_party_conversation() -> None:
...
     assert "Bob" in response.content
+
+
+def test_structured_output_and_tools() -> None:
+    class ResponseFormat(BaseModel):
+        response: str
+        explanation: str
+
+    llm = ChatOpenAI(model="gpt-4o-mini").bind_tools(
+        [GenerateUsername], strict=True, response_format=ResponseFormat
```
marcammann, line 1224: I didn't know you could do that! TIL, thanks for adding tests around this.

PR #29654, opened 2025-02-06
langchain: release 0.3.18

ccurme:


.github/workflows/_test_pydantic.yml
```diff
@@ -19,6 +19,7 @@ on:
 
 env:
   UV_FROZEN: "true"
+  UV_NO_SYNC: "true"
```
ccurme, line 22: We explicitly `uv sync` throughout our workflows. Sometimes we
```
uv sync
uv pip install <something else>
uv run ...
```
We need to make sure `uv run` does not overwrite the result of the previous pip install.

Example with min pydantic version tests:
Screenshot 2025-02-07 at 9 14 08 AM
^ This second installation is removed after setting `UV_NO_SYNC`.

- [Example run without UV_NO_SYNC](https://github.com/langchain-ai/langchain/actions/runs/13189400262/job/36819103222?pr=29654)
- [Example run with UV_NO_SYNC](https://github.com/langchain-ai/langchain/actions/runs/13201471563/job/36854360412?pr=29654)

1174-yu:
Ok update

PR #29691, opened 2025-02-08
feat(ollama): Switch Ollama default to json_schema

jacoblee93:
Function calling is unreliable for some models.

Should be accompanied by a minor bump.


ccurme (commented):
Closing as duplicated with https://github.com/langchain-ai/langchain/pull/30273

(Apologies as this one was first)

PR #29693, opened 2025-02-09
community: fix azure cosmosdb no sql

ianchi:
Fixes error when using custom field names

Custom fields are already aliased in the sql query, so they are no longer present in the response item. Reference the fields by the aliases.

Adds posibility to customize metadata fields.
Currently it only works if the collection has a single key with an object containing all metadata.
This adds more alternatives:

metadata_key: Field(s) to use for metadata. It can be:
                - "*": all fields in the document will go in metadata.
                - a string: field in the document that contains the metadata.
                    It must be of `object` data type (compatible with current behavior)
                    (it's properties will go in metadata).
                - a list of strings: fields in the document to add to metadata.
                    Use this if metadata is spread across multiple fields.
                - a dictionary: same as list but allows to alias.
                    A mapping of fields in the document to aliases to add as metadata
                    (e.g., {"field1": "alias1", "field2": "alias2"}).
                - None: no metadata will be added.


ccurme (commented):

PR #29699, opened 2025-02-09
community: add include_comment_forest to RedditSearch

hemati:
- **Description:**
    Introduces a new `include_comment_forest` boolean parameter in the `RedditSearchSchema` and `RedditSearchRun` tool. When set to `True`, this fetches and recursively parses the entire nested comment tree using `_parse_comment_forest`, and formats it using `_format_comment_forest`. Updated `run` and `results` methods to support displaying the full comment hierarchy.

  - **Issue:**
    (No specific issue to reference.)

  - **Dependencies:**
    None.


ccurme (commented):
Do you think it makes more sense to initialize the tool with `include_comment_forest`, instead of requiring the model to populate True/False?

Note that as-is, this PR would require all models using this tool to populate the new parameter. Some models will also start filling out True, will pull a huge comment history, and cause problems for maintainers. IMO this should be an opt-in feature.

hemati:
**Thanks for the thoughtful feedback!**

My original reason for having `include_comment_forest` as a parameter in each request was to let the model/agent decide **on a per-call basis** whether it needs the full comment tree. For example:

1. **Selective Deep-Dive**  
   - The agent might first retrieve only post metadata. If it finds something important, it can dynamically request the full comment forest (`include_comment_forest=True`) on a second call.
2. **Contextual User Requests**  
   - If a user specifically asks, “What are people saying about X in the comments?” the agent can enable `include_comment_forest=True` only when needed, instead of always pulling comments.
3. **Adaptive Workflows**  
   - The agent can decide *on the fly* based on resource limits or user instructions, rather than one static setting for all calls.

However, I understand your concern that this could lead to unintended large data pulls if an LLM or user call sets `include_comment_forest=True` by default. 

**Proposed Compromise:**

1. Keep `include_comment_forest` in the request schema, **defaulted to `False`**.  
2. Add a constructor parameter, for example `allow_comment_forest`, which **defaults to `False`**.  
3. If `allow_comment_forest` is `False`, the tool will ignore any `True` value for `include_comment_forest` in the call, effectively preventing deep comment pulls.  
4. If someone **needs** dynamic control, they explicitly set `allow_comment_forest=True` so the agent can opt in on a per-call basis.

This way:
- **Maintain flexibility**: Developers who want dynamic deep dives can enable them at runtime.
- **Protect performance**: If maintainers want to prevent accidental large data usage, they leave `allow_comment_forest=False`.

Would that address both the need for runtime flexibility *and* your concern about possibly large or unintended data pulls?

ccurme (commented):
Thanks for the update. This could work but it would be confusing for a model to specify `include_comment_forest` with no effect. You could dynamically select the schema. But I think ideally we get this out of langchain-community so we can test and version changes like this properly (there also appear to be no tests on this tool, which makes reviewing new features difficult).

I believe you contributed the Discord integration, would you be interested in doing something similar for Reddit?

hemati:
> Thanks for the update. This could work but it would be confusing for a model to specify `include_comment_forest` with no effect. You could dynamically select the schema. But I think ideally we get this out of langchain-community so we can test and version changes like this properly (there also appear to be no tests on this tool, which makes reviewing new features difficult).
> 
> I believe you contributed the Discord integration, would you be interested in doing something similar for Reddit?

Thanks for the feedback! Yes, I did contribute the Discord integration. I’d be happy to work on a similar, fully-tested and separately versioned Reddit integration if that’s the direction you prefer. 

Do you want to close this PR?

PR #29702, opened 2025-02-09
community: make elastic_search_bm25 retriever directly index documents, accept user-input analyzer for flexible full-text search

bluearrow98:
**Description:** Extend full-text search capability of elastic_search_bm25 retriever by accepting additional parameters such as analyzer type, client kwargs and possibility to directly index documents.

**docs**: Improved existing documentation.


bluearrow98:
Thanks @ccurme. Just did that

ccurme (commented):
Thanks for this. We should be using [langchain-elastic](https://github.com/langchain-ai/langchain-elastic). If this functionality can be accessed there, let's do that. Otherwise we should create an issue or PR in that repo.

PR #29708, opened 2025-02-10
community: change wikidata rest api version from v0 to v1

zzaebok:
**Description:**

According to the [wikidata documentation](https://www.wikidata.org/wiki/Wikidata_talk:REST_API), Wikibase REST API version 1 (stable) is released from November 11, 2024.
Their guide is to use the new v1 API and, it just requires replacing v0 in the routes with v1 in almost all cases.
So I replaced WIKIDATA_REST_API_URL from v0 to v1 for stable usage.


ccurme (approved):
Thank you!

PR #29709, opened 2025-02-10
community[minor]: 06 - Refactoring PDFPlumber

pprados:
Modify PDFPlumber to:
- separate into loader/parser components
- identify tables in PDFs
- process images in PDFs like other parsers
- support multi-threading


pprados:
@eyurtsev the next.
I use a special approach to handle capitalized properties. I hope this works for you.
If not, we can “double” the keys to maintain compatibility.

pprados:
@eyurtsev ping

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1362,106 +1365,571 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
                 _format_inner_image(blob, text_from_image, self.images_inner_format)
             )
             image.close()
         return _FORMAT_IMAGE_STR.format(image_text=_JOIN_IMAGES.join(str_images))
 
 
+# The legacy PDFPlumberParser use key with upper case.
+# This is not aligned with the new convention, which requires the key to be in
+# lower case.
+class _PDFPlumberParserMetadata(dict):
```
eyurtsev, line 1381: This feels a bit heavy handed as a solution?

Could we add an init parameter to `PDFPlumberParser` called "metadata_format" and have it act as a version? It could allow a "legacy" and "standard" versions?

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1362,106 +1365,571 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
+                                doc_metadata
+                                | {
+                                    "page": page.page_number - 1,
+                                }
+                            )
+                        ),
+                    )
+                else:
+                    contents.append(all_text)
+                # "tables_as_html": [self._convert_table_to_html(table)
```
eyurtsev, line 1631: commented out code

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1362,106 +1365,571 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
+        )
+
+        # Iterate over words. If a word is in a table,
+        # yield the accumulated text, and the table
+        # A the word is in a previously see table, ignore it
+        # Finish with the accumulated text
+        kwargs.update(
+            {
+                "keep_blank_chars": True,
+                # "use_text_flow": True,
```
eyurtsev, line 1693: There's a lot of commented out code in this implementation -- would you be able to remove it?

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1362,106 +1365,571 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
+        Returns:
+            The table content as a string in CSV format.
+        """
+        if not table:
+            return ""
+
+        output = ["\n\n"]
+
+        # skip first row in details if header is part of the table
+        # j = 0 if self.header.external else 1
```
eyurtsev, line 1860: commented out code?

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1362,106 +1365,571 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
+        clean = True
+
+        # iterate over detail rows
+        for row in table:
+            line = "<tr>"
+            for i, cell in enumerate(row):
+                # output None cells with empty string
+                cell = "" if cell is None else cell.replace("\n", " ")
+                if clean:  # remove sensitive syntax
+                    cell = html.escape(cell.replace("-", "&#45;"))
```
eyurtsev, line 1895: Add a doc-string about the presence of replacement for `-`. I don't think most developers will know the character set that's covered by html.escape. Is escaping the `-` necessary?

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1362,106 +1365,571 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
+                    )
+                else:
+                    logger.warning("Unknown PDF Filter!")
+
+        return images
+
+    def _extract_tables_bbox_from_page(
+        self,
+        page: pdfplumber.page.Page,
+    ) -> list[tuple]:
```
eyurtsev, line 1788: tuple[float,float,float,float] ? 

Could you document the mapping between the type and the coordinate system? different coordinate systems are used by different parsers

i.e., 

top, left, bottom, right
left, top, right, bottom
left, top, width, height

and position can be in pixels or sometimes in percentage w/ respect to the container

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1362,106 +1365,571 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
-                            "page": page.page_number - 1,
-                            "total_pages": len(doc.pages),
-                        },
-                        **{
-                            k: doc.metadata[k]
-                            for k in doc.metadata
-                            if type(doc.metadata[k]) in [str, int]
-                        },
+            doc = pdfplumber.open(file_path, password=self.password)  # open document
+            from pdfplumber.utils import geometry  # import WordExctractor, TextMap
```
eyurtsev, line 1564: commented out code (rest of import)

eyurtsev (approved):

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: The default strategy will be "legacy" or "standard"?

pprados: @eyurtsev 
Another solution can be to set upper and lower case keys.

eyurtsev: Maybe default strategy would be `None` which will resolve to `legacy` to avoid breaking user code.

This would allow emitting a warning users to alert them that they should specify a format either `legacy` or `standard`.

Having both upper and lower case keys doesn't feel like an API we'd want to end up with long-term. Having both keys in the metadata will be confusing (i.e., are both keys the same are they different why are there two etc.)

An alternative is also to not modify the metadata at all and handle this on the client side

pprados: I implement a replication of keys. The legacy keys and lower keys. It's compatible without breaking user code.

pprados: It's in the last version now.

pprados:
@eyurtsev all is fixed now.

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -748,7 +751,7 @@ def render(item: LTItem) -> None:
                                 blob = Blob.from_path(Path(tempdir) / filename)
                                 blob.metadata["source"] = "#"
                                 image_text = next(
-                                    self.images_parser.lazy_parse(blob)
+                                    self.images_parser.lazy_parse(blob)  # type: ignore
```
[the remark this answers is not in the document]
eyurtsev: OK looks like this is not resolved to the current PR but to a previous PR. OK to ignore for now. A fix may be potentially required later for the existing code path as the number of documents returned from the interface can be 0 or more than 1 document.

libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -1371,97 +1376,525 @@ def _extract_images_from_page(self, page: pypdfium2._helpers.page.PdfPage) -> st
...
-                            k: doc.metadata[k]
-                            for k in doc.metadata
-                            if type(doc.metadata[k]) in [str, int]
-                        },
-                    ),
+            doc = pdfplumber.open(file_path, password=self.password)  # open document
+            from pdfplumber.utils import geometry  # import WordExctractor, TextMap
+
+            contents = []
+            doc_metadata = doc.metadata | _purge_metadata(
```
eyurtsev, line 1525: Does this not end up doubling some metadata keys?

eyurtsev (commented):
Hi @pprados! Looks good. I'm not 100% sure i followed the metadata solution? Doesn't the current solution end up duplicating some keys?

Also there's some commented out code in the implementation would you mind doing a pass through the code to clean it up?

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: Double keys, no. Have similar keys, camel case for the historical version and lower for the new version, yes.

Why is this important?

A `RouterParser is about to arrive, allowing you to choose a parser based on document characteristics. Client code that retrieves documents and metadata shouldn't have to worry about knowing which parser is used for parsing, and add code accordingly. This creates lots of code branches, generates bugs, complicates TU, etc.

The aim of all my work is to do just that. To be able to change from one Parser to another, without any further modification to the code.

For historical reasons, some keys have to be kept as duplicates. It's a small price to pay to avoid numerous bugs.

For example, you can
```
 routes = [
        (“Microsoft”, {“producer”: “Microsoft”, “creator”: “Microsoft”},
        PyMuPDFParser()),
        (“LibreOffice”, {“producer”: “LibreOffice”, }, PDFPlumberParser()),
        (“Xdvipdfmx”, {“producer”: “xdvipdfmx.*”, “page1”: “Hello”}, PDFPlumberParser()),
        (“defautl”, {}, PyPDFium2Parser())
    ]
 loader = PDFRouterLoader(filename, routes=routes)
 docs=loader.load()
docs[0].metadata[“creationdate”] # No bug. Not  sometimes CreationDate
```

Translated with DeepL.com (free version)

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: it's private for PDFPlumber

libs/community/langchain_community/document_loaders/parsers/pdf.py
eyurtsev: > For historical reasons, some keys have to be kept as duplicates. It's a small price to pay to avoid numerous bugs.

The two options for handling the metadata are that we discussed in the context of this PR were:

1. Allow the user to choose the format in the init
2. Keep the legacy format and push the burden to the client

I was under the impression that we agreed on (1) as the solution?

If (2) is chosen as the solution then, the metadata standardization can done in the PDFRouterLoader without modifying this implementation or other implementations.

Any improvements that make parsing more robust or layout better are fantastic for the parsers.

Changes that are either breaking or make the existing API look strange (e.g., double keys in the metadata) are ones that we shouldn't make. While they may support some uses cases.

----

For code blocks like these:

```python
doc_metadata = doc.metadata | _purge_metadata(
    (
        doc.metadata
        | {
            "source": blob.source,
            "file_path": blob.source,
            "total_pages": len(doc.pages),
        }
    )
)
```

It's really helpful to include doc-strings that explain **why** the logic is what it is. This is not for users, but for future maintainers that will be looking at this and wonder why is doc.metadata so complex?

An implementation w/ the init could look like this:


```python
if metadata_format == 'legacy':
   metadata = ... # original implementation
elif metadata_format == "standard":
   metadata = _standardize_metadata
else:
   ...
```

eyurtsev (commented):
@pprados still outstanding comment w/ regards to metadata. I pushed a change to remove commented out code.

libs/community/langchain_community/document_loaders/parsers/pdf.py
pprados: @eyurtsev 
The default value for metadata_format must be "legacy" ?

That's what I don't like about this solution. 
Either, the user must explicitly indicate that he wants the new approach, and add this parameter forever, all future versions 

```
class XXXLoader(metadata_format:str="legacy", 
...):
loader=XXXLoader(metadata_format="standard", ...) # Don't forget this crazy parameter
```.
Or, by default, it's the new approach 
```
class XXXLoader(metadata_format:str="standard", 
...):
loader=XXXLoader( ...) # Break compatibility
...

Or, a mixture of both, so that the old code continues to work as long as the uppercase keys are not deprecated. The new code doesn't have to take the history into account forever.
```
loader=XXXLoader( ...) # Compatible with new and old code
```
What do you think?

Which solution do you prefer? What should I use as the default value for `metadata_format`?

pprados:
@eyurtsev 
The default value for metadata_format must be "legacy" ?

That's what I don't like about this solution. 
Either, the user must explicitly indicate that he wants the new approach, and add this parameter forever, all future versions 

```
class XXXLoader(metadata_format:str="legacy", 
...):
loader=XXXLoader(metadata_format="standard", ...) # Don't forget this crazy parameter
```.
Or, by default, it's the new approach 
```
class XXXLoader(metadata_format:str="standard", 
...):
loader=XXXLoader( ...) # Break compatibility
...

Or, a mixture of both, so that the old code continues to work as long as the uppercase keys are not deprecated. The new code doesn't have to take the history into account forever.
```
loader=XXXLoader( ...) # Compatible with new and old code
```
What do you think?

Which solution do you prefer? What should I use as the default value for `metadata_format`?

eyurtsev:
> The default value for metadata_format must be "legacy" ?
> That's what I don't like about this solution.
> Either, the user must explicitly indicate that he wants the new approach, and add this parameter forever, all future versions

I agree -- it's a problem developing in the community package right now as:

* Users rely on the code to not break between releases
* It's hard to deprecate functionality since one has to wait for a major release.

The simplest solution for this is using a standalone package instead of trying to update the community implementation. This allows evolving the implementation w/ appropriate semver semantics. This also allows the you (the developer) to move quickly.

Another solution is to introduce logic to mark the parameter as pending deprecation. We'd want the deprecation to go into effect in the next **minor** or **major** release rather than the next **patch** release. This solution works, but it's slow to make interface changes.

---

> Which solution do you prefer? What should I use as the default value for metadata_format?

`metadata_format="legacy"` for now, and add a pending warning that will kick off in the next release that will inform the user that the default will change.

But I in general, would prefer to not make changes in the metadata format or to only restrict the changes to addition of previously unincluded fields.

Document loaders in langchain currently do not standardize metadata. So even if this is fixed for some of the PDF loaders, it's not generally a feature of langchain document loaders right now.

pprados:
From the analysis we've made of all the parsers out there, there isn't one that's right for every situation. Some are very efficient for documents coming from Word, as they rely on the PDF format, others are better for PowerPoint documents, as they rely on an image analysis of the pages, and are able to find the reader order, understand diagrams, etc. It doesn't seem possible to have a solution that works well in all situations. It doesn't seem possible to have one solution that works well in all situations. Cost impacts can also be an important variable.

In my experience, every project tells me that it works well for one type of document, but not for another. Regardless of the parser used.

That's why I wanted to propose a simple, `GenericLoader` compatible approach, with a split between the `BlobLoader` and the `Parser`.

To be able to write in 20 lines what is usually written in 2000 lines.

```python
vector_store=...
record_manager=...
loader=GenericLoader(
    blob_loader=FileSystemBlobLoader(  # Ou CloudBlobLoader
        path="mydata/",
        glob="**/*",
        show_progress=True,
    ),
    blob_parser=MimeTypeBasedParser(
        handlers={
          "application/pdf": PDFRouterParser(...
          "application/vnd.openxmlformats-officedocument.wordprocessingml.document":
            MsWordParser(),
        },
        fallback_parser=TextParser(),
    )
)
index(
    loader.lazy_load(),
    record_manager,
    vector_store,
    batch_size=100,
)
```
Of course, it's possible to make an external component. The visibility of this approach will not be the same as if the solution is directly integrated into langchain-community. I'm afraid that, for several years to come, we'll be seeing wobbly projects that implement solutions outside langchain, out of ignorance or fear of complexity. Absolutely all the projects I come across don't use Loaders correctly. It's a real shame. I wanted to improve this situation.

My aim is really to simplify it. To be able to get the maximum value out of langchain, without running away from it when you can't figure out how to do it. I'll be presenting a [conference](https://www.lagrosseconf.com/talks-2025#philippe_prados) next Tuesday on this very subject. (sorry, it's in French).

For example, there's a new `langchain-unstructured` project. From what I understand, this project is intended to retrieve all loaders of type `UnstructuredXXXLoader`. Current implementations don't respect the Loader/parser split, don't accept a `Path` as a filename, and don't all have unit tests.

For example, there is a new `langchain-unstructured` project. From what I understand, this project is intended to retrieve all loaders of type `UnstructuredXXXLoader`. Current implementations don't respect the Loader/parser split, don't accept a `Path` as a filename, and don't all have unit tests.

I'm currently working on a migration of all these `UnstructuredXXXLoader` into the `langchain-unstructured` project, improving the Loader/Parser breakdown, adding all the tests, etc. I'm wondering if I'm on the right track.

Am I on the right track? I don't know. I probably should have talked about it before embarking on this work. It seems relevant to me, to be able to use `GenericLoader`, which isn't possible at the moment.

In the end, it's up to you or your team. Should I continue down this path? Should I stop here?

pprados:
@eyurtsev 
An important objective of the redesign of PDF parsers is to make them compatible with GenericLoader. Currently, this is not the case for the last 2.

pprados:
@eyurtsev 
To fix the [Issue 30454](https://github.com/langchain-ai/langchain/issues/30454) can you merge this PR?

Note: Current implemention of PDFPlumber has several limitations:
- It does not provide a parser
- Uses load()`` instead of lazy_load()`
- Does not handle tables
- Does not support image conversions.

pprados:
@ccurme or @baskaryan, can you review this PR?

ccurme (commented):
Closing as `langchain-community` has been moved to a standalone repo: https://github.com/langchain-ai/langchain-community

PR #29718, opened 2025-02-11
community: add support for Azure CosmosDB NoSQL query_constructors

ianchi:
Current metadata filter implementation in Azure CosmosDB NoSQL vectorstore uses custom objects, which are not native to CosmosDB nor to LangChain.
This PR adds the option in the vector store to filter with the more native "WHERE" sql clause as string.
It also adds a Translator class, so that in can be natively used with Structured Queries / Filter Expressions.

It adds another (more native) filtering syntax, but it is backwards compatible with current implementation.

### DIFF
--- a/libs/community/langchain_community/query_constructors/cosmosdb_no_sql.py
+++ b/libs/community/langchain_community/query_constructors/cosmosdb_no_sql.py
@@ -1,0 +1,89 @@
+from typing import Any, Dict, Tuple
+
+from langchain_core.structured_query import (
+    Comparator,
+    Comparison,
+    Operation,
+    Operator,
+    StructuredQuery,
+    Visitor,
+)
+
+SQL_COMPARATOR = {
+    Comparator.EQ: "=",
+    Comparator.NE: "!=",
+    Comparator.GT: ">",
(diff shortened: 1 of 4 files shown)


ccurme (commented):

ianchi:
Hi @ccurme, I see a lot of changes already there.
But I'll adjust this PR  to move it there and send a new PR later today.

Another question. This PR introduces a native way to express filter conditions in a native way to CosmosDB NoSQL, that is by a simple string, and by using a FilterExpression for the Langchain's standard objects.

The original implementation "invented" a custom Condition class, very similar conceptually to the langchain standard FilterDirective.

Wouldn't this change of repo be a good moment to deprecate such a custom object in favor of these two standard compliant alternatives? (FilterDirective and/or WHERE clause)?
I already see many breaking changes in the new AzureCosmosDBNoSqlVectorSearch class, so it should be the best moment to do it.

I'll include this change as a separate commit in the PR so that you can judge.

I've sent other PR with other fixes for this vector store, I'll move them all here. And there where more errors I encountered but have not yet send the PR, I'll move those directly to the new repo.

PR #29719, opened 2025-02-11
community(query_constructor): add mql_to_structured helper

ianchi:
Adds a helper function to transform a mql-like dict into a FilterDirective (and be later used with any translator integration).

This can be helpful in many ways:
+ to construct the FilterDirective in code from a much less verbose mql dict.
+ It can also reuse mql expressions that come from other sources, as mql is a popular standard
+ one source can also be from an llm as structured output directly as mql (as it is a well known format for many llm's)

### DIFF
--- a/libs/community/langchain_community/utils/mql.py
+++ b/libs/community/langchain_community/utils/mql.py
@@ -1,0 +1,160 @@
+from typing import Any, Dict, List, Union
+
+from langchain_core.structured_query import (
+    Comparator,
+    Comparison,
+    FilterDirective,
+    Operation,
+    Operator,
+)
+
+# Mapping of MongoDB comparison operators to LangChain comparators
+MQL_TO_LC_COMPARATOR = {
+    "$eq": Comparator.EQ,
+    "$ne": Comparator.NE,
+    "$gt": Comparator.GT,
+    "$gte": Comparator.GTE,
+    "$lt": Comparator.LT,
+    "$lte": Comparator.LTE,
+    "$in": Comparator.IN,
+    "$nin": Comparator.NIN,
+}
+
+MQL_LOGICAL_OPERATORS = {
+    "$and": Operator.AND,
+    "$or": Operator.OR,
+    "$not": Operator.NOT,
+}
+
+
+def _parse_condition(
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Should this live in the [langchain-mongodb](https://github.com/langchain-ai/langchain-mongodb) package?

ianchi:
I'm not sure, but would argue that it should not.
This is more about taking advantage of a well known way of expressing a filter expression as a json.

We are using this json format even if we don't use MongoDB at all (but SQL in fact).

ccurme (commented):
Is there anywhere in community now where we could use this to refactor something?

I'm not sure we should add this into community unless we have documented use cases, or are able to demonstrate a refactor it enables.

ianchi:
Hi @ccurme,

I'm not sure if a refactor, but more of an alternative.
It enables an alternativa approach to self-query. But for generating the filter directive, instead of relying on a custom syntax that must be teached to the llm in the prompt, and later parsed from text with an AST parser (`get_query_constructor_prompt` + `StructuredQueryOutputParser`) it relies on the previous llm knowledge on MQL syntax, and the fact that MQL is naturally structured and doesn't need to get parsed.

In some complex queries we found that the LLM does better with the more knwon MQL than custom syntax. (With the additional removal on the external dependency on Lark).

With the function in this PR the proces would be:

prompt + StructuredOutputParser (with the mql_filter dict) + mql_to_structured.

the two last steps could also be wrappend in an alternative StructuredQueryOutputParser.

The rest of the process is the same in both alternatives, using a translator with the structured query and a retriever.

Thinking it over, perhaps instead of in community, this should go somewhere in `langchain.chains.query_constructor`

ianchi:
Hi @ccurme, any thoughts on this?
Should I change the PR?

ccurme:
Thanks for the ping. I am concerned that if we merge as-is, it will not get used as there is minimal documentation and it's not imported elsewhere in the library. I feel like we need to demonstrate the value by incorporating into an existing abstraction (without changing its interface) or documenting a use-case. (If you do want to write some docs, please ping me with ideas so we can make sure we're documenting in a reasonable place.)

ccurme:
Closing due to inactivity, can re-open if needed!

PR #29720, opened 2025-02-11
community(vectorstore): fix CosmosDB NoSQL create_container

ianchi:
Currently the parameter "create_container" is not considered and the database/container are always created even
if `create_container` is set to False

This fixes the issue, and make policies parameters only needed for creation optional (if flag is set to False).


ccurme (commented):

Issue #29626, opened 2025-02-06
JSONDecodeError for deepseek-reasoner model using ChatDeepSeek and BaseChatOpenAI


HenryXiaoYang:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_deepseek import ChatDeepSeek

llm = ChatDeepSeek(
    api_key="sk-xxx",
    model="deepseek-reasoner",
)

resp = llm.invoke("How to plant apple in dessert?")

print(resp.content)
```

### Error Message and Stack Trace (if applicable)

```python
Traceback (most recent call last):
  File "/Users/keren/Library/Application Support/JetBrains/PyCharm2024.1/scratches/scratch.py", line 8, in <module>
    resp = llm.invoke("How to plant apple in dessert?")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/keren/PycharmProjects/XYBotV2/.venv/lib/python3.11/site-packages/langchain_core/language_models/chat_models.py", line 284, in invoke
    self.generate_prompt(
  File "/Users/keren/PycharmProjects/XYBotV2/.venv/lib/python3.11/site-packages/langchain_core/language_models/chat_models.py", line 860, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/keren/PycharmProjects/XYBotV2/.venv/lib/python3.11/site-packages/langchain_core/language_models/chat_models.py", line 690, in generate
    self._generate_with_cache(
  File "/Users/keren/PycharmProjects/XYBotV2/.venv/lib/python3.11/site-packages/langchain_core/language_models/chat_models.py", line 925, in _generate_with_cache
    result = self._generate(
             ^^^^^^^^^^^^^^^
[...]

PR #29758, opened 2025-02-12
[Exception Handling] DeepSeek JSONDecodeError

AhmedTammaa:
For Context please check #29626 

The Deepseek is using langchain_openai. The error happens that it show `json decode error`. 

I added a handler for this to give a more sensible error message which is DeepSeek API returned empty/invalid json. 

Reproducing the issue is a bit challenging as it is inconsistent, sometimes DeepSeek returns valid data and in other times it returns invalid data which triggers the JSON Decode Error. 

This PR is an exception handling, but not an ultimate fix for the issue.


ccurme (commented):
This is a problem with DeepSeek. I don't think we should handle this in langchain-openai (there are many providers who subclass BaseChatOpenAI, it's not tenable to account for these possibilities for all providers).

What do you think about implementing this on ChatDeepSeek-- e.g., overriding `_stream` and other methods and handling exceptions when you call `super()._stream`?

AhmedTammaa:
Aha, thank you @ccurme, I didn't know that it is used by others other than Openai and Deepseek. You are absolutely right about your concern. I will make the suggested changes and revert the BaseChatOpenAI changes.

AhmedTammaa:
@ccurme I have made the changes as suggested. However, I need an advice here: Should I raise the error as ValueError as we have it from the parent class or JSONDecodeError?

please note I did a small refactor in the function `_create_chat_result` the ruff formatter was moving the type ignore comment so it was triggering the linting checker.

libs/partners/deepseek/langchain_deepseek/chat_models.py
```diff
@@ -239,3 +245,37 @@ def _convert_chunk_to_generation_chunk(
...
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> Iterator[ChatGenerationChunk]:
+        try:
+            yield from super()._stream(
+                messages, stop=stop, run_manager=run_manager, **kwargs
+            )
+        except ValueError as e:
+            raise ValueError(
```
ccurme, line 261: Should we catch ValueError, or JSONDecodeError?

In this commit we were catching JSONDecodeError: https://github.com/langchain-ai/langchain/pull/29758/commits/4bd23d6eaa2736dfc188bec102680c7e4c6edaa5

I think we should raise the same error type as is emitted now (e.g., JSONDecodeError, if that's what the client raises).

ccurme (commented):
Thanks!

libs/partners/deepseek/langchain_deepseek/chat_models.py
AhmedTammaa: Perfect thank you very much. I was very perplexed here. I will update it

AhmedTammaa:
I handled the exception as JSONDecodeError as this the triggered exception.

ccurme (approved):

PR #29785, opened 2025-02-13
openai[patch]: update system role to developer for o-series models

ccurme:
Some o-series models will raise a 400 error for `"role": "system"` (`o1-mini` and `o1-preview` will raise, `o1` and `o3-mini` will not).

Here we update `ChatOpenAI` to update the role to `"developer"` for all model names matching `^o\d`.

We only make this change on the ChatOpenAI class (not BaseChatOpenAI).


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -2020,6 +2021,12 @@ def _get_request_payload(
         # in September 2024 release
         if "max_tokens" in payload:
             payload["max_completion_tokens"] = payload.pop("max_tokens")
+
+        # Mutate system message role to "developer" for o-series models
+        if self.model_name and re.match(r"^o\d", self.model_name):
+            for message in payload.get("messages", []):
+                if message["role"] == "system":
+                    message["role"] = "developer"
```
baskaryan, line 2029: should we put this logic in convert_message_to_dict, so all the message creation logic is in one place?

ccurme: `convert_message_to_dict` applies to BaseChatOpenAI, and my mental model is:

- ChatOpenAI stays up to date with the OpenAI API
- BaseChatOpenAI does its best to accommodate the multitude of "openai-compatible" APIs

`"developer"` roles seemed like a strictly OpenAI thing so it felt safer to limit to ChatOpenAI. But I don't feel strongly about this if we are restricting by `model_name`.

We'd also need to add the model name as a kwarg to `convert_message_to_dict`.

baskaryan: if we're adding model name as kwarg to convert_message_to_dict (makes sense to me) then couldn't the logic still be openai-specific?

ccurme: In the sense that model names that start with `o\d` are OpenAI models, yes.

ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain/actions/runs/13499663061

PR #29811, opened 2025-02-14
langchain_community: add error checking in HuggingFaceInferenceAPIEmbeddings JSON response

brian-ogrady:
When using a custom HuggingFace API Endpoint there can be an error in the JSON response. I propose added two checks when calling the `embed_documents` method of the `HuggingFaceInferenceAPIEmbeddings` class:

1. Add a check for a `JSONDecodeError`
2. Add a check within the JSON response for an error

While working with the `HuggingFaceInferenceAPIEmbeddings` class, I sometimes receive a nebulous error `ValueError: could not convert string to float: 'e'`
Screenshot 2025-02-13 at 5 04 36 PM

This results because the `embed_documents` method, which is called during the execution of the `add_documents` method of a vector store, can return a JSON response like the following: `{'error': 'Input validation error: "inputs" cannot be empty', 'error_type': 'Validation'}`

The initial `ValueError` raised from `add_documents` is misleading because it masks the underlying error returned in the JSON response. I propose surfacing this error by adding a check during `embed_documents` to return the JSON error in a Pythonic manner.

I have detailed [this request in the discussions](https://github.com/langchain-ai/langchain/discussions/29791#discussion-7962964).


brian-ogrady:
@efriis from a strategic perspective, do you see the value in keeping the `HuggingFaceInferenceAPIEmbeddings` around? There is another class, [HuggingFaceEndpointEmbeddings](https://github.com/langchain-ai/langchain/blob/master/libs/partners/huggingface/langchain_huggingface/embeddings/huggingface_endpoint.py), in `langchain-huggingface` that serves the same purpose but instead leverages the `huggingface-hub` inference client. 

If, strategically, you believe it worthwhile to keep this class around then my changes are worthwhile. If not, however, I can close out this PR w/o merging.

ccurme (commented):
Thanks for this. We do recommend using langchain-huggingface instead of the implementations in community. We should seek to deprecate this class, either pointing to an existing replacement in langchain-huggingface, or moving it over if there isn't equivalent functionality.

Created an issue here: https://github.com/langchain-ai/langchain/issues/29910

PR #29812, opened 2025-02-14
core[patch]: allow passing JSON schema as args_schema to tools

vbarda:


libs/core/langchain_core/tools/base.py
```diff
@@ -460,7 +469,7 @@ def tool_call_schema(self) -> type[BaseModel]:
 
     def get_input_schema(
         self, config: Optional[RunnableConfig] = None
-    ) -> type[BaseModel]:
+    ) -> type[ArgsSchema]:
```
vbarda, line 472: this is not going to work because Runnables require pydantic output here. looking into a solution

baskaryan (commented):
looks reasonable to me! think we should add lots of tests

libs/core/langchain_core/tools/base.py
```diff
@@ -605,7 +626,12 @@ async def _arun(self, *args: Any, **kwargs: Any) -> Any:
     def _to_args_and_kwargs(
         self, tool_input: Union[str, dict], tool_call_id: Optional[str]
     ) -> tuple[tuple, dict]:
-        if self.args_schema is not None and not get_fields(self.args_schema):
+        if (
+            self.args_schema is not None
+            and not isinstance(self.args_schema, dict)
```
eyurtsev, line 631: `not isinstance(self.args_schema, dict)` isn't logically correct. The conditional should be in terms of what `args_schema` **is** rather than what it *is not*.

e.g., 

`args_schema` is truthy, and it can be subclassed, and it's a subclass of one of basemodel or basemodelv1..

eyurtsev (commented):
Looks good! Could you update the branching logic so it's easy to reason about it and add a unit test?

libs/core/langchain_core/tools/base.py
vbarda: yup, this was a hacky solution to deal w/ issubclass checks but there is a better way, replaced it

eyurtsev (approved):

PR #29813, opened 2025-02-14
community: add JustextWebLoader for cleaned web content extraction

majrie:
**Description:** Adds a new JustextWebLoader to langchain-community that uses the justext library for extracting cleaned text from web pages. The loader removes boilerplate content and navigation elements while preserving main content, with support for multiple languages and configurable content splitting.

**Dependencies:** 
- justext==3.0.1
- beautifulsoup4==4.11.1

All dependencies are properly imported within functions and added to the optional web_justext group.


ccurme (commented):
Hello, thanks for this. This adds a net-new integration into langchain-community. We are recommending that new contributions be implemented via separate packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-justext`)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the loader, and you have control over versioning.

Let me know what you think!

Issue #29803, opened 2025-02-14
ChatMistralAI does not pass on Frequency Penalty Argument


Bernsai:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_mistralai import ChatMistralAI

ChatMistralAI(
            api_key=api_key,
            base_url=base_url,
            model=model_type.value,
            temperature=0.1,
            frequency_penalty=0.5,
        )

ChatMistralAI(
            api_key=api_key,
            base_url=base_url,
            model=model_type.value,
            temperature=0.1,
            model_kwargs={
                "frequency_penalty": 0.5,
            },
        )

ChatMistralAI(
            api_key=api_key,
            base_url=base_url,
            model=model_type.value,
            temperature=0.1,
            kwargs={
                "frequency_penalty": 0.5,
            },
        )
```

### Description

When providing frequency penalty to ChatMistralAI (langchain-mistralai==0.2.6) it is not passed on to Mistral API.

When checking the request, which is actually sent to Mistral API frequency_penalty is always null. Adapting that request to expose e.g. `frequency_penalty=0.5` and manually sending to Mistral API solves the issue.

It seems ChatMistralAI does not use frequency_penalty argument in any way, although it is supported by Mistral API.

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  #53-Ubuntu SMP PREEMPT_DYNAMIC Sat Jan 11 00:06:25 UTC 2025
> Python Version:  3.12.3 (main, Jan 17 2025, 18:03:48) [GCC 13.3.0]

Package Information
-------------------
> langchain_core: 0.3.35
> langchain: 0.3.18
> langchain_community: 0.3.17
> langsmith: 0.3.8
> langchain_elasticsearch: 0.3.2
> langchain_huggingface: 0.1.2
> langchain_mistralai: 0.2.6
> langchain_openai: 0.3.5
> langchain_text_splitters: 0.3.6

PR #29838, opened 2025-02-17
mistral[patch]: support model_kwargs

keenborder786:
- **Description:** Frequency_penalty added as a client parameter
- **Issue:** #29803


keenborder786:
@ccurme

ccurme (commented):
Thanks! Updated this to implement model_kwargs instead, so we handle things like `presence_penalty` and other params.

This is "breaking" in the sense that we will now error for anyone who passes erroneous parameters (e.g., unsupported parameters, misspellings, etc.). I'd categorize this as a bug fix.

ccurme (approved):

PR #29841, opened 2025-02-17
docs: Update gemini vector docs

pnikonowicz:
# Description

2 changes: 
1. removes get pass from the code example as it reads from stdio causing a freeze to occur
2. updates to the latest gemini model in the example


ccurme (approved):
Thanks for this. Leaving in getpass for now for consistency with the rest of the docs. It's pretty ubiquitous in the docs so if we want to remove it we should do everywhere.

Issue #29843, opened 2025-02-17
ChatFireworks doesn't retry on API errors (max_retries parameter is fake)


aimestereo:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

nothing custom, just copy/paste example from [ChatFireworks documentation](https://python.langchain.com/docs/integrations/chat/fireworks/) (notice specified parameter `max_retries `):

```python
import getpass
import os

if "FIREWORKS_API_KEY" not in os.environ:
    os.environ["FIREWORKS_API_KEY"] = getpass.getpass("Enter your Fireworks API key: ")
```

```python
from langchain_fireworks import ChatFireworks

llm = ChatFireworks(
    model="accounts/fireworks/models/llama-v3-70b-instruct",
    temperature=0,
    max_tokens=None,
    timeout=None,
    max_retries=2,
    # other params...
)
```

```python
messages = [
    (
        "system",
        "You are a helpful assistant that translates English to French. Translate the user sentence.",
    ),
    ("human", "I love programming."),
]
ai_msg = llm.invoke(messages)
print(ai_msg.content)
```

### Error Message and Stack Trace (if applicable)

`httpx.ReadError`:
```
email-writer-9   | Traceback (most recent call last):
email-writer-9   |   File "/usr/local/lib/python3.12/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
email-writer-9   |     yield
email-writer-9   |   File "/usr/local/lib/python3.12/site-packages/httpx/_transports/default.py", line 394, in handle_async_request
email-writer-9   |     resp = await self._pool.handle_async_request(req)
email-writer-9   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
email-writer-9   |   File "/usr/local/lib/python3.12/site-packages/httpcore/_async/connection_pool.py", line 256, in handle_async_request
email-writer-9   |     raise exc from None
email-writer-9   |   File "/usr/local/lib/python3.12/site-packages/httpcore/_async/connection_pool.py", line 236, in handle_async_request
email-writer-9   |     response = await connection.handle_async_request(
[...]

PR #29856, opened 2025-02-18
OpenAI client for `ChatFireworks`

keenborder786:
- **Description:** OpenAI Client for Firework 
- **Issue:**  #29843

### DIFF
--- a/libs/partners/fireworks/langchain_fireworks/chat_models.py
+++ b/libs/partners/fireworks/langchain_fireworks/chat_models.py
@@ -23,7 +23,7 @@ from typing import (
     cast,
 )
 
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Although you can access Fireworks with the openai sdk, I don't think the openai sdk is a drop-in replacement for the fireworks sdk.

Instead of migrating everything to openai, I'd guess that adding retries is a simpler solution to the issue, but haven't looked carefully. Mistral embeddings implements that, for example: https://github.com/langchain-ai/langchain/blob/d8bab89e6eb55f04c91050605482171c763d2c8f/libs/partners/mistralai/langchain_mistralai/embeddings.py#L226

PR #29881, opened 2025-02-19
community: lakeFS document loader - support object level user metadata

### LINKED ISSUE #8408: LangChain LakeFSLoader: load document metadata tags
**Problem**

Currently our [LakeFSLoader](https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/document_loaders/lakefs.py) loads document without metadata object tags and only adds the logical path tag as source key to Document. We ignore any user set object metadata. 

**So What?** 

This means that the LakeFSLoader  is not helpful the moment I need to use object metadata for embedding, retrieval etc. 
For example, very common case in RAG of [time weighted retriever](https://python.langchain.com/v0.1/docs/modules/data_connection/retrievers/time_weighted_vectorstore/). 

To use the object tags I would have to implement custom logic for this  listing all documents and getting the tags, bottom line it means I don't have a use in LakeFSLoader since I would have to load everything anyway myself. 

**Suggestion** 

When calling LakeFSLoader constructor in Langchain, add additional optional param to indicate id metadata tags should be added to the Document. 

```python
loader = LakeFSLoader(repo="...",ref="...",path="...", with_metadata=True) 
```

guy-har:
Description: lakeFS has object-level metadata that may be useful when loading documents. This PR adds an option to request metadata when initiating a lakeFS loader. By default, metadata is not requested.

Issue: No issue linked.

Dependencies: No new dependencies. The added tests mock the HTTP response, which now calls the load functionality. The lakeFS loader returns an `UnstructuredLakeFSLoader`, which is a subclass of `UnstructuredBaseLoader` and requires installing `unstructured` to run the tests.

### DIFF
--- a/libs/community/langchain_community/document_loaders/lakefs.py
+++ b/libs/community/langchain_community/document_loaders/lakefs.py
@@ -71,7 +71,6 @@ class LakeFSClient:
             for obj in results
         ]
 
-
     def is_presign_supported(self) -> bool:
         config_endpoint = self.__endpoint + "config"
         response = requests.get(config_endpoint, auth=self.__auth)
--- a/libs/community/langchain_community/document_loaders/lakefs.py
+++ b/libs/community/langchain_community/document_loaders/lakefs.py
@@ -1,7 +1,7 @@
 import os
 import tempfile
 import urllib.parse
-from typing import Any, List, Optional
+from typing import Any, List, Optional, Tuple
 from urllib.parse import urljoin
 
 import requests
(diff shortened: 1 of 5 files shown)


ccurme (commented):
Hello, thanks for this. Would you be interested in publishing an OSS integration package (e.g., `langchain-lakefs` or `langchain-treeverse`)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the model, and you have control over versioning.

Let me know what you think!

guy-har:
@ccurme thanks for the suggestion! 
Thanks for the suggestion! We’ll go ahead with publishing an OSS integration package as you recommended. 
In the meantime, all we need is to add object-level metadata support, so we'd like to submit a small PR. If it helps, we can make it even smaller by including only the absolute necessary changes.

Let me know what you think!

PR #29894, opened 2025-02-19
core[patch]: set version="v2" as default in astream_events

ccurme:


libs/core/tests/unit_tests/runnables/test_runnable_events_v2.py
```diff
@@ -1057,9 +1060,11 @@ def with_parameters_and_callbacks(x: int, y: str, callbacks: Callbacks) -> dict:
         """A tool that does nothing."""
         return {"x": x, "y": y}
 
+    kwargs = {"version": "v2"} if specify_version else {}
```
eyurtsev, line 1063: feels like an unnecessary complication for this specific unit test

The test seems to boil down to confirming that the default version is `v2` 

Perhaps a separate logical test for it would be good?

eyurtsev (approved):

PR #29912, opened 2025-02-21
core[patch]: raise error if a tool call has an empty tool call ID

vbarda:
Should address issues like https://github.com/langchain-ai/langgraph/issues/3538


vbarda:
Closing since this is a breaking change in favor of a more light-weight https://github.com/langchain-ai/langchain/pull/29921

PR #29914, opened 2025-02-21
community: Update SQLiteVec table trigger

Talostat:
**Issue**:  This trigger can only be used by the first table created. Cannot create additional triggers for other tables.

**fixed**:  Update the trigger name so that it can be used for new tables.

### DIFF
--- a/libs/community/langchain_community/vectorstores/sqlitevec.py
+++ b/libs/community/langchain_community/vectorstores/sqlitevec.py
@@ -95,7 +95,7 @@ class SQLiteVec(VectorStore):
         )
         self._connection.execute(
             f"""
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Thanks for this. Could you add a test?

Talostat:
The original trigger works correctly when the database is stored in memory (db_file=":memory:"). 
However, its behavior fails when the database is stored in local files (e.g., test.db), so that's what I did to update it.

I attempted to add a new test case to address this, but due to the creation of actual files during local file storage and my unfamiliarity with langchain file structure, I've refrained from including a concrete example in this context

ccurme (approved):

PR #29915, opened 2025-02-21
community/mlx_pipeline: fix crash at mlx call

jpdournel:
- **Description:** 
Since mlx_lm 0.20, all calls to mlx crash due to deprecation of the way parameters are passed to methods generate and generate_step.
Parameters top_p, temp, repetition_penalty and repetition_context_size are not passed directly to those method anymore but wrapped into "sampler" and "logit_processor".

- **Dependencies:** mlx_lm (optional)

-  **Tests:** 
I've had a new test to existing test file: tests/integration_tests/llms/test_mlx_pipeline.py

### DIFF
--- a/libs/community/langchain_community/chat_models/mlx.py
+++ b/libs/community/langchain_community/chat_models/mlx.py
@@ -160,6 +160,7 @@ class ChatMLX(BaseChatModel):
 
         try:
             import mlx.core as mx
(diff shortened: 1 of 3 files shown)


ccurme (approved):
Thanks! We should move this out of langchain-community so that we can version changes properly and manage the `mlx` dependency. Created an issue [here](https://github.com/langchain-ai/langchain/issues/29922).

Issue #29925, opened 2025-02-21
Only instantiate SDKs when necessary


ccurme:
### Privileged issue

- [x] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

We should prefer cached properties for clients (e.g., the `openai` sdk) to avoid excessive overhead.

E.g., BaseChatOpenAI will generate sync/async root clients + completions clients, even if a user is only interested in async completions.

PR #29926, opened 2025-02-21
openai[patch]: instantiate clients lazily

ccurme:
https://github.com/langchain-ai/langchain/issues/29925

Simplified version of the change below.

**Before**:
```python
from typing import Any, Optional

import openai
from pydantic import BaseModel, Field, SecretStr, model_validator
from typing_extensions import Self


class BaseChatOpenAI(BaseModel):
    client: Any = Field(default=None, exclude=True)
    async_client: Any = Field(default=None, exclude=True)
    my_special_api_key: Optional[SecretStr] = Field(alias="api_key")

    @model_validator(mode="after")
    def validate_environment(self) -> Self:
        client_params = {"api_key": self.my_special_api_key}
        if not self.client:
            self.client = openai.OpenAI(**client_params)  # type: ignore[arg-type]
        if not self.async_client:
            self.async_client = openai.AsyncOpenAI(**client_params)  # type: ignore[arg-type]
        return self
```

**After**:
```python
from typing import Any, Dict, Optional

import openai
from pydantic import BaseModel, Field, PrivateAttr, SecretStr, model_validator
from typing_extensions import Self

class BaseChatOpenAI(BaseModel):
    _client: Any = PrivateAttr(default=None)
    _async_client: Any = PrivateAttr(default=None)
    _client_params: Dict[str, Any] = PrivateAttr(default_factory=dict)
    my_special_api_key: Optional[SecretStr] = Field(alias="api_key")

    # Needed to support initializing the class with `client`
    def __init__(
        self, client: Any = None, async_client: Any = None, **kwargs: Any
    ) -> None:
        super().__init__(**kwargs)
        self._client = client
        self._async_client = async_client

    @model_validator(mode="after")
    def validate_environment(self) -> Self:
        self._client_params = {"api_key": self.my_special_api_key}
        return self

    @property
    def client(self) -> Any:
        if self._client is None:
            self._client = openai.OpenAI(**self._client_params)  # type: ignore[arg-type]
        return self._client

    @client.setter
    def client(self, value: Any) -> None:
        self._client = value

    @property
    def async_client(self) -> Any:
        if self._async_client is None:
            self._async_client = openai.AsyncOpenAI(**self._client_params)  # type: ignore[arg-type]
        return self._async_client

    @async_client.setter
    def async_client(self, value: Any) -> None:
        self._async_client = value
```

I'm not sure there is a way to do this that is not semver-breaking. One usage pattern I know breaks is subclasses of BaseChatOpenAI, most of which have unique names for their API keys and other attributes (e.g., `xai_api_key`, `together_api_key`, etc.). These subclasses have post init steps that look like:
```python
@model_validator(mode="after")
def validate_environment(self) -> Self:
    client_params = {"api_key": self.xai_api_key.get_secret_value()}
    if not self.client:  # <-- breaks
        self.client = openai.OpenAI(**client_params)
```
The check `if not self.client` breaks because we now try to instantiate the client in a property, and this requires that client params be accessible to the property.

Current plan is to update classes we control and explain how to update sub-classes in release notes. The update is simple but would be annoying.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -634,7 +634,7 @@ class BaseChatOpenAI(BaseChatModel):
         return self._http_async_client
 
     @property
-    def root_client(self) -> openai.OpenAI:
+    def root_client(self) -> Any:
         if self._root_client is None:
             sync_specific = {"http_client": self.http_client}
             self._root_client = openai.OpenAI(
@@ -648,7 +648,7 @@ class BaseChatOpenAI(BaseChatModel):
         self._root_client = value
 
     @property
-    def root_async_client(self) -> openai.AsyncOpenAI:
+    def root_async_client(self) -> Any:
         if self._root_async_client is None:
             async_specific = {"http_client": self.http_async_client}
             self._root_async_client = openai.AsyncOpenAI(
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -8,7 +8,6 @@ import logging
 import os
 import sys
 import warnings
-from functools import cached_property
 from io import BytesIO
 from math import ceil
 from operator import itemgetter
@@ -564,7 +563,7 @@ class BaseChatOpenAI(BaseChatModel):
 
         return self
 
-    @cached_property
+    @property
     def _http_client(self) -> Optional[httpx.Client]:
         """Optional httpx.Client. Only used for sync invocations.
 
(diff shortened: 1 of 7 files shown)


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -511,6 +528,24 @@ def validate_temperature(cls, values: Dict[str, Any]) -> Any:
...
         return values
 
+    def __init__(
+        self,
+        client: Optional[Any] = None,
+        async_client: Optional[Any] = None,
+        root_client: Optional[Any] = None,
+        async_root_client: Optional[Any] = None,
+        http_client: Optional[Any] = None,
+        http_async_client: Optional[Any] = None,
```
baskaryan, line 538: ooc are these untyped on purpose?

baskaryan (commented):
looks reasonable to me, assuming we re-ran the benchmarks and this fixed the issue?

ccurme:
Closing for now, need to benchmark performance gains.

PR #29939, opened 2025-02-23
core(mermaid): allow greater customization

ianchi:
Adds greater style customization by allowing a custom frontmatter config. This allows to set a `theme` and `look` or to adjust theme by setting `themeVariables`

Example:

```python

node_colors = NodeStyles(
    default="fill:#e2e2e2,line-height:1.2,stroke:#616161",
    first="fill:#cfeab8,fill-opacity:0",
    last="fill:#eac3b8",
)

frontmatter_config = {
    "config": {
        "theme": "neutral",
        "look": "handDrawn"
    }
}

graph.get_graph().draw_mermaid_png(node_colors=node_colors, frontmatter_config=frontmatter_config)
```

image


vbarda:
@ianchi thanks for the contribution! could you please separate the subgraph fix into a separate PR and add a test for it in `libs/core/tests/runnables/test_graph.py`?

ianchi:
Done. 
The fix has been moved to PR #29970

vbarda:
thanks! I merged the fix PR - feel free to update

ianchi:
hi @vbarda the PR has conflicts resolved

ianchi:
Hi @vbarda, I believe this is ready to be merged.
If you find anything else is needed please let me know.

vbarda:
@ianchi thanks! i might actually do a slightly different refactor, since looks like directives are actually deprecated now so might as well go w/ the new recommended format https://mermaid.js.org/config/directives.html

vbarda:
@ianchi i was looking more at mermaid docs and it looks like frontmatter is _really_ poorly documented. i think we can keep your original change! sorry for the false alarm

ianchi:
It is not well documented, but basically in the `config` property of the front matter you include the same as in directives.
The documentation for directive is valid for "config".

It is not a bad idea to use front matter, if you use the live playground it is what is generated, so it is easy to copy from there.

ianchi:
Hi @vbarda, updated tests.

vbarda:
Thank you @ianchi - made a few small changes and will merge!

PR #29963, opened 2025-02-24
core[patch]: pydantic 2.11 compat

### LINKED ISSUE #29951: Errors with upcoming 2.11 Pydantic release
### Checked other resources

### Example Code

NA

_No response_

### Description

Pydantic 2.11 is going to be released soon, and causes errors similar to the ones encountered with the 2.10 release.

This time, it is the [`ChatAnthropic`](https://github.com/langchain-ai/langchain/blob/8b511a3a784657fb4a5a34670986242c4046aa4f/libs/partners/anthropic/langchain_anthropic/chat_models.py#L286) that fails to resolve an annotation, and the culprit is the same as https://github.com/langchain-ai/langchain/issues/28284 -- as `ChatAnthropic` has `BaseLanguageModel` as one of its bases.

As I mentioned in [this comment](https://github.com/langchain-ai/langchain/issues/28284#issuecomment-2495954444):

> https://github.com/langchain-ai/langchain/pull/28297 fixed the issue, however there might be a more robust way to fix it. [...] instead of rebuilding every model making use of `BaseLanguageModel`, I would strongly recommend having `BaseLanguageModel` defined properly in the first place. It seems like moving `BaseCache` and `Callbacks` outside of the if `TYPE_CHECKING:` block does not cause any circular import issues.

Two options are available:
- Do the same as https://github.com/langchain-ai/langchain/pull/28297, extending the model rebuilds to the necessary models (I wouldn't recommend it, as per my initial comment).
- Investigate if moving `BaseCache` and `Callbacks` outside of the if `TYPE_CHECKING:` block works. If not, try to find a way to rebuild the `BaseLanguageModel` class as early as possible.

### System Info

Package Information
-------------------
> langchain_core: 0.3.37
> langsmith: 0.3.10
> langchain_anthropic: 0.3.7

Pydantic main (the latest alpha doesn't contain the internal refactor that led to the langchain errors).

ccurme:
Resolves https://github.com/langchain-ai/langchain/issues/29951

Was able to reproduce the issue with Anthropic installing from pydantic `main` and correct it with the fix recommended in the issue.

Thanks very much @Viicos for finding the bug and the detailed writeup!

### DIFF
--- a/libs/core/langchain_core/language_models/base.py
+++ b/libs/core/langchain_core/language_models/base.py
@@ -18,6 +18,8 @@ from pydantic import BaseModel, ConfigDict, Field, field_validator
 from typing_extensions import TypeAlias, TypedDict, override
 
 from langchain_core._api import deprecated
+from langchain_core.caches import BaseCache
+from langchain_core.callbacks import Callbacks
 from langchain_core.messages import (
     AnyMessage,
     BaseMessage,
@@ -29,8 +31,6 @@ from langchain_core.runnables import Runnable, RunnableSerializable
 from langchain_core.utils import get_pydantic_field_names
 
 if TYPE_CHECKING:
-    from langchain_core.caches import BaseCache
-    from langchain_core.callbacks import Callbacks
     from langchain_core.outputs import LLMResult
 
 


Viicos:
Great to see the imports can just be moved out of the type checking block

PR #29971, opened 2025-02-24
anthropic[patch]: support claude 3.7 sonnet

ccurme:


libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -244,6 +244,15 @@ def _format_messages(
                                     if k in ("type", "text", "cache_control")
                                 }
                             )
+                    elif block["type"] == "thinking":
```
baskaryan, line 247: any reason to handle redacted_thinking?

baskaryan (approved):
tangential but ooc what happens if you send thinking block without the signature?

libs/partners/anthropic/langchain_anthropic/chat_models.py
ccurme: Looking now

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -774,6 +794,13 @@ def _format_output(self, data: Any, **kwargs: Any) -> ChatResult:
                 and block["citations"] is None
             ):
                 block.pop("citations")
+            if (
+                isinstance(block, dict)
+                and block.get("type") == "thinking"
+                and "text" in block
```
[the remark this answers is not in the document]
ccurme: model_dumping the content generates a key `{"text": None}`, so I remove it here.

ccurme:
> tangential but ooc what happens if you send thinking block without the signature?

BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.1.content.0.thinking.signature: Field required'}}

baskaryan:
we should follow up with reasoning token counts if their api exposes those (guessing it does)

ccurme:
> we should follow up with reasoning token counts if their api exposes those (guessing it does)

I'm not seeing these carved out in the response:
```python
import anthropic 
 
client = anthropic.Anthropic()
response = client.messages.create(
    model="claude-3-7-sonnet-latest",
    max_tokens=5000,
    thinking={
        "type": "enabled",
        "budget_tokens": 2000,
    },
    messages=[{
        "role": "user",
        "content": "Hello",
    }]
)

response.usage
```
```
Usage(cache_creation_input_tokens=0, cache_read_input_tokens=0, input_tokens=36, output_tokens=128)
```

baskaryan:
> > we should follow up with reasoning token counts if their api exposes those (guessing it does)
> 
> I'm not seeing these carved out in the response:
> 
> ```python
> import anthropic 
>  
> client = anthropic.Anthropic()
> response = client.messages.create(
>     model="claude-3-7-sonnet-latest",
>     max_tokens=5000,
>     thinking={
>         "type": "enabled",
>         "budget_tokens": 2000,
>     },
>     messages=[{
>         "role": "user",
>         "content": "Hello",
>     }]
> )
> 
> response.usage
> ```
> 
> ```
> Usage(cache_creation_input_tokens=0, cache_read_input_tokens=0, input_tokens=36, output_tokens=128)
> ```

yea doesn't seem updated yet https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/usage.py

PR #29972, opened 2025-02-24
Transition prolog_tool doc to langgraph

apisani1:
@ccurme As suggested I transitioned the prolog_tool documentation to langgraph


ccurme (approved):
Thanks!

PR #29975, opened 2025-02-24
community: add Onnxruntime-genai support

omer-demir:
**Description:** 

As part of our project, we had chance to work with [Onnxruntime-Genai](https://onnxruntime.ai/docs/genai/) library which is optimized for GenAI models to infer with OnnxRuntime. As it was really performant in our project, we wanted to extend langchain for community, hence this PR.

**Dependencies:**

By this PR, we created `OnnxruntimeGenai` that requires `onnruntime-genai` package.

**Tests:**

Integration test introduced for onnxruntime-genai to work.

### DIFF
--- a/libs/community/langchain_community/llms/__init__.py
+++ b/libs/community/langchain_community/llms/__init__.py
@@ -434,6 +434,12 @@ def _import_azure_openai() -> Type[BaseLLM]:
     return AzureOpenAI
 
 
+def _import_onnxruntime_genai() -> Type[BaseLLM]:
+    from langchain_community.llms.onnxruntime_genai import OnnxruntimeGenAi
+
+    return OnnxruntimeGenAi
+
+
 def _import_openai() -> Type[BaseLLM]:
     from langchain_community.llms.openai import OpenAI
 
--- a/libs/community/langchain_community/llms/__init__.py
+++ b/libs/community/langchain_community/llms/__init__.py
@@ -422,6 +422,11 @@ def _import_ollama() -> Type[BaseLLM]:
     return Ollama
 
 
+def _import_onnxruntime_genai() -> Type[BaseLLM]:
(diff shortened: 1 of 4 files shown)


ccurme (commented):
Hello, thanks for this. This adds a net-new integration into langchain-community. We are recommending that new contributions be implemented via separate packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-onnxruntime-genai`)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the model, and you have control over versioning.

Let me know what you think!

omer-demir:
> Hello, thanks for this. This adds a net-new integration into langchain-community. We are recommending that new contributions be implemented via separate packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-onnxruntime-genai`)? We've written a walkthrough on this process here:
> 
> https://python.langchain.com/docs/contributing/how_to/integrations/
> 
> We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the model, and you have control over versioning.
> 
> Docs would continue to be maintained in the `langchain` repo.
> 
> Let me know what you think!

Great, is that also preferred way to include `ChatModel` there as well. As that will be part of our changes as well. What do you suggest for that? @ccurme

ccurme:
> > 100% use chat models instead of LLMs if you can.
> 
> @ccurme As a part of integration, we need to extend `BaseLLM` to create new `ChatModel`. Is it fine to have both changes? After all it is another inferencing method.

If you are talking about a change to langchain-core, can you first open an issue describing your proposal? Are you saying BaseChatModel will not suffice? I think it is unlikely that the correct thing to do is to add an abstraction to core for this case.

If you open an issue, feel free to tag me.

PR #29976, opened 2025-02-24
core[patch]: allow passing description to @tool decorator

vbarda:


libs/core/langchain_core/tools/convert.py
```diff
@@ -66,6 +70,7 @@ def tool(
     name_or_callable: Optional[Union[str, Callable]] = None,
     runnable: Optional[Runnable] = None,
     *args: Any,
+    description: Optional[str] = None,
```
eyurtsev, line 73: we'll need to document the new option so it appears in the API reference.


Also to confirm description isn't captured in the args schema at all?

vbarda: it is, i added a test for that

libs/core/langchain_core/tools/convert.py
```diff
@@ -83,6 +88,9 @@ def tool(
             converted to a tool. Must be provided as a positional argument.
         runnable: Optional runnable to convert to a tool. Must be provided as a
             positional argument.
+        description: Optional description for the tool.
```
eyurtsev, line 91: Could you specify the precedence rules?

What if description and arg schema and parse_docstring are all specified/used

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2468,3 +2468,33 @@ def foo(x: int) -> str:
...
+        return "hi"
+
+    @tool(description="description", args_schema=ArgsSchema)
+    def foo_args_schema_description(x: int) -> str:
+        return "hi"
+
+    assert foo.description == "Foo."
+    assert foo_description.description == "description"
+    assert foo_args_schema.description == "Bar."
+    assert foo_args_schema_description.description == "description"
```
eyurtsev, line 2500: could we validate the descriptino of the tool that's produced when we generate it's json schema?

eyurtsev (approved):

PR #29989, opened 2025-02-25
community: MariaDB vector store addition

rusher:
### New Feature

Since version 11.7.1, MariaDB support vector. This is a super fast implementation (see [some perf blog](https://smalldatum.blogspot.com/2025/01/evaluating-vector-indexes-in-mariadb.html)
The goal is to support MariaDB with langchain

Implementation can be done aither using this PR or using a specific detached project https://github.com/rusher/langchain-mariadb transfering ownership

- test added
- doc added
- lint and test validated


ccurme (commented):
Hello, thanks for this. This adds a net-new integration into langchain-community. We are recommending that new contributions be implemented via separate packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-mariadb`)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the vector store, and you have control over versioning.

Let me know what you think!

ccurme:
> @ccurme i've read those docs, and i might have miss something, but seem ok to me (test are ok and pass). I'm not sure what is missing then. Could you indicate what more you expect ?

Here are the changes I'd request:

1. Publish `langchain-mariadb` (I don't see an entry in [pypi](https://pypi.org/project/langchain-mariadb/), let me know if I've missed it).
2. Remove all changes to `libs/community/` and `.github/workflows/`. The implementation should live in the external repo.
3. Record the library in the langchain [package registry](https://github.com/langchain-ai/langchain/blob/master/libs/packages.yml).
4. Add a provider page for MariaDB (can be brief, see example [here](https://github.com/langchain-ai/langchain/blob/master/docs/docs/integrations/providers/jenkins.mdx)).

Happy to re-open the PR if this makes sense!

PR #30001, opened 2025-02-26
Add asynchronous generate interface

TheSongg:
- [ ] **PR title**: [langchain_community.llms.xinference]: Add asynchronous generate interface

- [ ] **PR message**: The asynchronous generate interface  support  stream data and non-stream data.
          
        chain = prompt | llm
        async for chunk in chain.astream(input=user_input):
            yield chunk

- [ ] **Add tests and docs**:

       from langchain_community.llms import Xinference

       llm = Xinference(
       server_url="http://0.0.0.0:9997",  # replace your xinference server url
       model_uid={model_uid}  # replace model_uid with the model UID return from launching the model
           stream = True
            )
       prompt = PromptTemplate(input=['country'], template="Q: where can we visit in the capital of {country}? A:")
       chain = prompt | llm
       async for chunk in chain.astream(input=user_input):
           yield chunk


ccurme (commented):
How is authentication handled here?

ccurme:
Does their client not support async calls?

TheSongg:
> Does their client not support async calls?

no，xinference client only support sync call. In addition, langchain_community.chat_models doesn't support xinference chat client.
https://github.com/xorbitsai/inference/blob/main/xinference/client/restful/restful_client.py
image

TheSongg:
> How is authentication handled here?

In fact, there was no authentication during initialization.
https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/llms/xinference.py
image

ccurme (approved):

PR #30004, opened 2025-02-26
fix(core): fix BaseTool when `args_schema` has pydantic validators that change the input keys

VMinB12:
**Description:**

We encountered a bug where we used a pydantic validator to patch broken llm inputs to our tools. The validator would fix the input dict, but the BaseTool would return an empty dict to the tool because our validator changed the dictionary keys. This PR changes one line of code to fix the issue and adds a test to improve the test coverage.

- **Issue:**

Concretely, the following code would fail before the PR:
```
from pydantic import BaseModel, model_validator


class B(BaseModel):
    b: int


class A(BaseModel):
    a: B

    @model_validator(mode="before")
    def a_wrapper(cls, data):
        if "a" not in data:
            return {"a": data}
        return data


if __name__ == "__main__":
    # Valid input passes
    print(A.model_validate({"a": {"b": 1}}))
    # Invalid input also passes because validator patches it
    print(A.model_validate({"b": 1}))


@tool(args_schema=A, infer_schema=False)
def answer(
    **kwargs,
):
    """Use this function to provide your final answer to the request."""
    return kwargs


if __name__ == "__main__":
    # Valid input passes
    assert answer.invoke({"a": {"b": 1}}) == {"a": B(b=1)}
    # Invalid input should pass, but throws assertion error
    res = answer.invoke({"b": 1})
    assert res == {"a": B(b=1)}, f"Failure: {res} of type {type(res)}"
    # AssertionError: Failure: {} of type <class 'dict'>
```

**Add tests and docs**:
I added a test for the above use case.

**Comments**:

Please note that I had to remove `*args: Any` from one of the tests. The previous code would filter out the `args` keyword. I don't see any realistic use case where one would have `*args` in their tool as this gives no meaningful instruction to the LLM on how to use the `*args`. Please let me know if you see any issues here. I don't see a clean and robust method that would have the new test passing while still filtering out the `*args` argument. Let me know if you see a better solution.


hugodendievelb12:
@ccurme

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -628,7 +628,7 @@ def test_named_tool_decorator_return_direct() -> None:
     """Test functionality when arguments and return direct are provided as input."""
 
     @tool("search", return_direct=True)
-    def search_api(query: str, *args: Any) -> str:
+    def search_api(query: str) -> str:
```
eyurtsev, line 631: No modifications to unrelated code should be made unless original code was wrong

```suggestion
    def search_api(query: str, *args: Any) -> str:
```

VMinB12: In fact, this change is necessary for the test to pass after the proposed change.
As I mentioned in the PR body, I see no clean way to have my new test and this old test pass without changes to the old test.

I opted for changing the existing test, because adding `*args` to an `@tool` decorated method is a strange thing to do. It means the LLM is instructed to add arbitrary additional `args` without any instructions on how the LLM should write these. I think it is inadvisable to write a tool in such a manner, and hence I think its acceptable for this PR to break that kind of usage of the @tool decorator.

Of course, having full backwards compatibility would be much preferred, so let me know if you see a way to get that.

VMinB12:
@eyurtsev I allow myself to retag you for this PR.

Let me also copy a reply of mine to an outdated comment you wrote:
"""
In fact, this change is necessary for the test to pass after the proposed change.
As I mentioned in the PR body, I see no clean way to have my new test and this old test pass without changes to the old test.

I opted for changing the existing test, because adding *args to an @tool decorated method is a strange thing to do. It means the LLM is instructed to add arbitrary additional args without any instructions on how the LLM should write these. I think it is inadvisable to write a tool in such a manner, and hence I think its acceptable for this PR to break that kind of usage of the @tool decorator.

Of course, having full backwards compatibility would be much preferred, so let me know if you see a way to get that.
"""

eyurtsev:
Closing in favor of: https://github.com/langchain-ai/langchain/pull/32925

PR #30048, opened 2025-02-28
anthropic, mistral: return `model_name` in response metadata

ccurme:
Took a "census" of models supported by init_chat_model-- of those that return model names in response metadata, these were the only two that had it keyed under `"model"` instead of `"model_name"`.


langchain-infra (approved):

lars20070:
Very nice. Thanks for the quick fix.

Issue #30049, opened 2025-02-28
Partially initialised variable ignored when concatenating ChatPromptTemplates.


thoffmann-artidis:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_core.prompts import ChatPromptTemplate

prompt = ChatPromptTemplate.from_messages([('system', 'Prompt {x} {y}')]).partial(x='1')
appendix = ChatPromptTemplate.from_messages([('system', 'Appendix {z}')])

(prompt + appendix).invoke({'y': '2', 'z': '3'})
```

### Error Message and Stack Trace (if applicable)

`KeyError: "Input to ChatPromptTemplate is missing variables {'x'}.  Expected: ['x', 'y', 'z'] Received: ['y', 'z']\nNote: if you intended {x} to be part of the string and not a variable, please escape it with double curly braces like: '{{x}}'.\nFor troubleshooting, visit: https://python.langchain.com/docs/troubleshooting/errors/INVALID_PROMPT_INPUT "`

### Description

Shouldn't the code snippet run without raising a key error?

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6020
> Python Version:  3.12.9 (main, Feb  4 2025, 14:38:38) [Clang 16.0.0 (clang-1600.0.26.6)]

Package Information
-------------------
> langchain_core: 0.3.40
> langchain: 0.3.19
> langchain_community: 0.3.15
> langsmith: 0.2.11
> langchain_openai: 0.3.1
> langchain_text_splitters: 0.3.6
> langgraph_sdk: 0.1.51

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp: 3.11.11
> aiohttp<4.0.0,>=3.8.3: Installed. No version info available.
> async-timeout<5.0.0,>=4.0.0;: Installed. No version info available.
> dataclasses-json: 0.6.7
> httpx: 0.28.1
> httpx-sse: 0.4.0
[...]

PR #30096, opened 2025-03-04
core[patch]: fix loss of partially initialized variables during prompt composition

Thommy257:
**Description:**
This PR addresses the loss of partially initialised variables when composing different prompts. I.e. it allows the following snippet to run:

```python

prompt = ChatPromptTemplate.from_messages([('system', 'Prompt {x} {y}')]).partial(x='1')
appendix = ChatPromptTemplate.from_messages([('system', 'Appendix {z}')])

(prompt + appendix).invoke({'y': '2', 'z': '3'})
```

Previously, this would have raised a `KeyError`, stating that variable `x` remains undefined.

**Issue**
References issue #30049

**Todo**


libs/core/langchain_core/prompts/chat.py
```diff
@@ -1041,28 +1042,30 @@ def __add__(self, other: Any) -> ChatPromptTemplate:
             Combined prompt template.
         """
         partials = {**self.partial_variables}
-        try:
+        with contextlib.suppress(AttributeError):
             partials = {**partials, **other.partial_variables}
```
Thommy257, line 1046: I was a bit lazy here. Feel free to suggest other solutions

Thommy257:
Let me know if I should add some unit tests.

libs/core/langchain_core/prompts/chat.py
```diff
@@ -1040,19 +1041,31 @@ def __add__(self, other: Any) -> ChatPromptTemplate:
         Returns:
             Combined prompt template.
         """
+        partials = {**self.partial_variables}
+        with contextlib.suppress(AttributeError):
```
Undertone0809, line 1045: or try to use the following code?
 ```
if hasattr(other, "partial_variables") and other.partial_variables:
        partials.update(other.partial_variables)
```

Comment (Undertone0809) [reply libs/core/langchain_core/prompts/chat.py:1045]: It looks like AttributeError might override other possible errors.

Comment (thoffmann-artidis) [reply libs/core/langchain_core/prompts/chat.py:1045]: Python's `hasattr` also just executes a try except block. I'd rather avoid it as it's not very explicit.

Comment (Thommy257) [reply libs/core/langchain_core/prompts/chat.py:1045]: Python's hasattr also just executes a try except block. I'd rather avoid it as it's not very explicit.

Comment (Thommy257) [reply libs/core/langchain_core/prompts/chat.py]: removed in https://github.com/langchain-ai/langchain/pull/30096/commits/d2746a49017023678ab7cede8f40cfa7770edefe

Comment (eyurtsev) [reply libs/core/langchain_core/prompts/chat.py:1045]: Re-writing to this version, as it's the more common variant for python developers

```python
if hasattr(other, "partial_variables") and other.partial_variables: 
   partials.update(other.partial_variables)
```

eyurtsev (approved):

eyurtsev:
Thanks for the patch!! @Thommy257  and for the reviews @Undertone0809

PR #30128, opened 2025-03-05
core, openai, standard-tests: improve OpenAI compatibility with Anthropic content blocks

ccurme:
- Support thinking blocks in core's `convert_to_openai_messages` (pass through instead of error)
- Ignore thinking blocks in ChatOpenAI (instead of error)
- Support Anthropic-style image blocks in ChatOpenAI

---

Standard integration tests include a `supports_anthropic_inputs` property which is currently enabled only for tests on `ChatAnthropic`. This test enforces compatibility with message histories of the form:
```
- system message
- human message
- AI message with tool calls specified only through `tool_use` content blocks
- human message containing `tool_result` and an additional `text` block
```
It additionally checks support for Anthropic-style image inputs if `supports_image_inputs` is enabled.

Here we change this test, such that if you enable `supports_anthropic_inputs`:
- You support AI messages with text and `tool_use` content blocks
- You support Anthropic-style image inputs (if `supports_image_inputs` is enabled)
- You support thinking content blocks.

That is, we add a test case for thinking content blocks, but we also remove the requirement of handling tool results within HumanMessages (motivated by existing agent abstractions, which should all return ToolMessage). We move that requirement to a ChatAnthropic-specific test.

### DIFF
--- a/libs/core/langchain_core/messages/utils.py
+++ b/libs/core/langchain_core/messages/utils.py
@@ -1191,6 +1191,8 @@ def convert_to_openai_messages(
                                 },
                             }
                         )
+                    elif block.get("type") == "thinking":
+                        content.append(block)
                     else:
                         err = (
                             f"Unrecognized content block at "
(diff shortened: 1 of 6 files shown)


ccurme:
Integration tests pass for [openai](https://github.com/langchain-ai/langchain/actions/runs/13701210305) and [anthropic](https://github.com/langchain-ai/langchain/actions/runs/13701212962).

Issue #30122, opened 2025-03-05
Granite 3.2 Thinking


lemassykoi:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use GitHub Discussions.
- [x] I added a clear and detailed title that summarizes the issue.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I included a self-contained, minimal example that demonstrates the issue INCLUDING all the relevant imports. The code run AS IS to reproduce the issue.

### Example Code

```python
from langgraph.prebuilt import create_react_agent

def build_ollama_llm(model:str = OLLAMA_CHAT_MODEL) -> ChatOllama:
    return ChatOllama(
        base_url    = OLLAMA_BASE_URL,
        model       = model,
        temperature = OLLAMA_TEMPERATURE,
        seed        = OLLAMA_SEED,
        format      = "",
        verbose     = OLLAMA_CHAT_FUNCTION_VERBOSE,
        disable_streaming = True,
        keep_alive  = 0
    )

def build_react_agent(model):
    logger.info('Creating ReAct Agent')
    return create_react_agent(
        model        = model,
        tools        = TOOLS_LIST,
        checkpointer = memory,
        name         = "ReAct_Agent_WIP",
        debug        = REACT_AGENT_VERBOSE
    )

llm = build_ollama_llm()
agent_supervisor = build_react_agent(llm)

messages_with_think = {
            "messages": [
                {
                    "role": "control",
                    "content": "thinking"
                },
                (
                    "user",
                    question,
                )
            ]
}

if "granite3.2" in OLLAMA_CHAT_MODEL:
    messages = messages_with_think

async for event in agent_supervisor.astream(
    input = messages,
    config = config,
    stream_mode = "values",
    debug = OPENAI_SUPERVISOR_VERBOSE
):
    logger.debug(event)
```

### Error Message and Stack Trace (if applicable)

```shell
Traceback (most recent call last):
  File "c:\react_WIP.py", line 945, in astream_interactive
    async for event in agent_supervisor.astream(
  File "c:\Users\user\AppData\Local\Programs\Python\Python312\Lib\site-packages\langgraph\pregel\__init__.py", line 2304, in astream  
    while loop.tick(input_keys=self.input_channels):
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #30147, opened 2025-03-07
 langchain_core/langchain_community: add ControlMessage to langchain_core and enable passing messages with 'control' role through ChatOllama

rylativity:
- **Description:** this creates a ControlMessage class in langchain_core for messages with the 'control' role, and allows passing messages with 'control' role through to ollama through ChatOllama. The update to langchain_ollama’s ChatOllama relies on the added ControlMessage class in langchain_core, so updates to both libs are included in this PR
- **Issue:** #30122 
- **Dependencies:** no new dependencies

- [x] **PR title**: 
- [x] **PR message**: 
- [x] **Lint and test**: format, lint, and test completed and passing in langchain_core and langchain_community.

### DIFF
--- a/libs/community/langchain_community/chat_models/ollama.py
+++ b/libs/community/langchain_community/chat_models/ollama.py
@@ -12,6 +12,7 @@ from langchain_core.messages import (
     AIMessageChunk,
     BaseMessage,
     ChatMessage,
+    ControlMessage,
     HumanMessage,
     SystemMessage,
 )
@@ -107,6 +108,8 @@ class ChatOllama(BaseChatModel, _OllamaCommon):
                 message_text = f"[INST] {message.content} [/INST]"
         elif isinstance(message, AIMessage):
             message_text = f"{message.content}"
+        elif isinstance(message, ControlMessage):
+            message_text = f"{message.content}"
         elif isinstance(message, SystemMessage):
             message_text = f"<<SYS>> {message.content} <</SYS>>"
         else:
@@ -130,6 +133,8 @@ class ChatOllama(BaseChatModel, _OllamaCommon):
                 role = "assistant"
             elif isinstance(message, SystemMessage):
                 role = "system"
+            elif isinstance(message, ControlMessage):
+                role = "control"
             else:
                 raise ValueError("Received unsupported message type for Ollama.")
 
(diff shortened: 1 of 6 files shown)


ccurme (commented):
Several issues with this:

1. It's unlikely we are going to add a new message type to core until multiple large providers adopt something similar.
2. We should be updating `langchain-ollama` in `libs/partners/ollama` instead of the community implementation, which is deprecated (you should be seeing warning messages if you are using it).

Are you able to use use [ChatMessage](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/chat.py) with a custom role for this instead?

rylativity:
@ccurme a ChatMessage with a custom role _can_ potentially work, but there is currently an incompatibility between [the type of the `role` attribute in the ChatMessage class (which is a str)](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/chat.py#L14) and [the type of the `role` variable in ChatOllama's `_convert_messages_to_ollama_messages` method (which is a Literal)](https://github.com/langchain-ai/langchain/blob/master/libs/partners/ollama/langchain_ollama/chat_models.py#L488.)

Also, ChatOllama's `_convert_messages_to_ollama_messages` does not currently have an `isinstance` check for ChatMessages.

So in order to go that route, 1) the `_convert_messages_to_ollama_messages` method would need to be updated to handle ChatMessages like so:
```
...
 elif isinstance(message, ChatMessage):
    role = message.role
...
```
and 2) either the ChatMessage's `role` attribute would need to be changed to a literal (probably not ideal because that takes away its flexibility), or the "role" variable's type in `_convert_messages_to_ollama_messages` would need to be updated to accept the `str` type role attribute of a ChatMessage instance:
```
...
role: Literal["user", "assistant", "system", "tool", "control"] | str
...
```

The ChatMessage can technically work after addressing item `1)` above, but the mypy typechecking will fail without implementing something like item`2)` above.

If you are ok with this proposed route @ccurme (open to any feedback or alternative suggestions), and if that would satisfy your requirements/use-case @lemassykoi, I am happy to make those changes and open a new PR.

ccurme:
Very much appreciate the detailed plan! Let's go for it, we can work around / document the issues with `role`.

rylativity:
New PR opened here - https://github.com/langchain-ai/langchain/pull/30191 - much cleaner. 

We can go ahead and close this in favor of #30191

PR #30152, opened 2025-03-07
community: PlaywrightURLLoader can take stored session file

rauberdaniel:
**Description:**
Implements an additional `browser_session` parameter on PlaywrightURLLoader which can be used to initialize the browser context by providing a stored playwright context.


eyurtsev (approved):

libs/community/langchain_community/document_loaders/url_playwright.py
```diff
@@ -172,7 +177,22 @@ def lazy_load(self) -> Iterator[Document]:
             browser = p.chromium.launch(headless=self.headless, proxy=self.proxy)
             for url in self.urls:
```
eyurtsev, line 178: * does the context needs to be opened up for every URL?
* Is it correct to have a single browser session associated with multiple URLs?
* What is the use case here?

eyurtsev:
@rauberdaniel could you comment on the use case? Why do you need access to the browser session?

eyurtsev (changes requested):
Requesting changes to make sure I understand the use case and whether it's correct to reload the browser session per url

libs/community/langchain_community/document_loaders/url_playwright.py
rauberdaniel: You're right, the session should probably not be loaded for every url but only once. Having a single session / context for multiple urls simply for retrieval should not be an issue in my opinion.

The use case is to provide a session file with cookies or local storage set in order to evaluate urls that require a login for example.

rauberdaniel:
@eyurtsev I have made some changes in order to only load the session once and use the same context for all provided urls. Also, I added a `page.close()` call after each evaluation. I suspect that missing this could have led to memory issues until now, as pages only get closed after all urls have been opened in a new page each.

eyurtsev (approved):

Issue #30145, opened 2025-03-07
Ollama tools argument type modification


pawelka:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```
import json
from langchain_ollama.chat_models import _parse_arguments_from_tool_call

raw_response = """{"model":"sample-model","message":{"role":"assistant","content":"","tool_calls":[{"function":{"name":"get_profile_details","arguments":{"xxxyyy":"12345678901234567890123456"}}}]},"done":false}"""
raw_tool_calls = json.loads(raw_response)['message']['tool_calls']
_parse_arguments_from_tool_call(raw_tool_calls[0])

# result: {'xxxyyy': 12345678901234567890123456}

#expected {'xxxyyy': '12345678901234567890123456'}
```

### Error Message and Stack Trace (if applicable)

Error is cause by pydantic that tool expect string but received int.

### Description

Change in ticket https://github.com/langchain-ai/langchain/pull/28225 cause that simple argument like string with only digits only are convert to int and because it's long string with digits part of value is lost.

After commenting changes from related ticket:
```
def _parse_arguments_from_tool_call(
    raw_tool_call: dict[str, Any],
) -> Optional[dict[str, Any]]:
    """Parse arguments by trying to parse any shallowly nested string-encoded JSON.

    Band-aid fix for issue in Ollama with inconsistent tool call argument structure.
    Should be removed/changed if fixed upstream.
    See https://github.com/ollama/ollama/issues/6155
    """
    if "function" not in raw_tool_call:
        return None
    arguments = raw_tool_call["function"]["arguments"]
    parsed_arguments = {}
    if isinstance(arguments, dict):
        for key, value in arguments.items():
            # if isinstance(value, str):
            #     parsed_arguments[key] = _parse_json_string(
            #         value, skip=True, raw_tool_call=raw_tool_call
            #     )
            # else:
            parsed_arguments[key] = value
    else:
        parsed_arguments = _parse_json_string(
            arguments, skip=False, raw_tool_call=raw_tool_call
        )
    return parsed_arguments


[...]

PR #30154, opened 2025-03-07
(Ollama) Fix String Value parsing in  _parse_arguments_from_tool_call

keenborder786:
- **Description:** Fix String Value parsing in  _parse_arguments_from_tool_call
- **Issue:** #30145

### DIFF
--- a/libs/partners/ollama/langchain_ollama/chat_models.py
+++ b/libs/partners/ollama/langchain_ollama/chat_models.py
@@ -132,7 +132,7 @@ def _parse_arguments_from_tool_call(
                     value, skip=True, raw_tool_call=raw_tool_call
                 )
                 if not isinstance(parsed_arguments[key], (dict, list)):
-                    parsed_arguments[key] = str(value)
+                    parsed_arguments[key] = str(value) if not value.isalnum() else value
             else:
                 parsed_arguments[key] = value
     else:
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Can you add a test demonstrating what this is fixing?

keenborder786:
@ccurme okay

keenborder786:
Done

keenborder786:
@ccurme

keenborder786:
@ccurme

libs/partners/ollama/langchain_ollama/chat_models.py
```diff
@@ -131,6 +131,8 @@ def _parse_arguments_from_tool_call(
                 parsed_arguments[key] = _parse_json_string(
                     value, skip=True, raw_tool_call=raw_tool_call
                 )
+                if not isinstance(parsed_arguments[key], (dict, list)):
+                    parsed_arguments[key] = str(value)
```
ccurme, line 135: This will cast all inputs to string, correct? What about tools that require other types (e.g., ints)?

ccurme (commented):
Posted on the issue asking for a minimal reproducible example of the underlying issue.

libs/partners/ollama/langchain_ollama/chat_models.py
ccurme: Should be good actually, I missed we're in a condition `if isinstance(value, str):`

ccurme (approved):

Issue #30067, opened 2025-03-02
print reasoning with deepseek model


laurafbec:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

The following code does not return DeepSeek reasoning:
`from langchain_core.prompts import ChatPromptTemplate
from dotenv import load_dotenv

load_dotenv()
import os

from langchain_deepseek import ChatDeepSeek
from langchain_openai import ChatOpenAI

#os.environ["DEEPSEEK_API_KEY"] = os.getenv("DEEPSEEK_API_KEY")

llm = ChatDeepSeek(
    model="deepseek-chat",
    temperature=0,
    max_tokens=None,
    timeout=None,
    max_retries=2,
    # other params...
)

llm = ChatOpenAI(
    #model = 'deepseek/deepseek-chat:free',
    #model = 'deepseek/deepseek-chat',
    model = 'deepseek/deepseek-r1:nitro',
    #model='deepseek/deepseek-r1-distill-qwen-32b',
    #model='deepseek/deepseek-r1-distill-llama-8b',
    base_url="https://openrouter.ai/api/v1",
    api_key=os.getenv("OPENROUTER_API_KEY"),
    model_kwargs={"extra_body": {"include_reasoning": True}}
)

messages = [
    (
        "system",
        "You are a helpful assistant that translates English to French. Translate the user sentence.",
    ),
    ("human", "I love programming."),
]
ai_msg = llm.invoke(messages)
print(ai_msg.content)

prompt = ChatPromptTemplate(
    [
        (
            "system",
            "You are a helpful assistant that translates {input_language} to {output_language}.",
        ),
        ("human", "{input}"),
    ]
)

chain = prompt | llm
result = chain.invoke(
    {
        "input_language": "English",
        "output_language": "German",
        "input": "I love programming.",
    }
)
print(result.content)
print(result)`

### Error Message and Stack Trace (if applicable)

I'm not able to print the reasoning content of deepseek model. Find attached the code I'm using.
Thanks in advance!

### Description

`from langchain_core.prompts import ChatPromptTemplate
from dotenv import load_dotenv

load_dotenv()
import os

from langchain_deepseek import ChatDeepSeek
[...]

PR #30186, opened 2025-03-09
Core , openai: add reasoning content to solve issues #30067

JasonHonKL:
The pull request addresses the issue of not having reasoning content when using ChatOpenAI with the DeepSeek reasoner. The issue is resolved by modifying the cover dictionary to the message, adding reasoning content to the content if it exist. There are no changes to other parts of the code, and it is believed that there will be no side effects from adding the reasoning content. 

#30067

### DIFF
--- a/libs/core/langchain_core/messages/ai.py
+++ b/libs/core/langchain_core/messages/ai.py
@@ -164,6 +164,11 @@ class AIMessage(BaseMessage):
     This is a standard representation of token usage that is consistent across models.
     """
 
+    """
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Hi, thanks for this. We should use `langchain-deepseek` to interact with Deepseek (it already is extending BaseChatOpenAI). See my comment on the issue [here](https://github.com/langchain-ai/langchain/issues/30067#issuecomment-2708842139) and let me know if there's an issue with `langchain-deepseek`.

PR #30187, opened 2025-03-09
docs: hide jsx in llm chain tutorial

hesreallyhim:
## **Description:** 
The Jupyter notebooks in the docs section are extremely useful and critical for widespread adoption of LangChain amongst new developers. However, because they are also converted to MDX and used to build the HTML for the Docusaurus site, they contain JSX code that degrades readability when opened in a "notebook" setting (local notebook server, google colab, etc.). For instance, here we see the website, with a nice React tab component for installation instructions (`pip` vs `conda`):

Screenshot 2025-03-07 at 2 07 15 PM

Now, here is the same notebook viewed in colab:

Screenshot 2025-03-07 at 2 08 41 PM

Note that the text following "To install LangChain run:" contains snippets of JSX code that is (i) confusing, (ii) bad for readability, (iii) potentially misleading for a novice developer, who might take it literally to mean that "to install LangChain I should run `import Tabs from...`" and then an ill-formed command which mixes the `pip` and `conda` installation instructions.

Ideally, we would like to have a system that presents a similar/equivalent UI when viewing the notebooks on the documentation site, or when interacting with them in a notebook setting - or, at a minimum, we should not present ill-formed JSX snippets to someone trying to execute the notebooks. As the documentation itself states, running the notebooks yourself is a great way to learn the tools. Therefore, these distracting and ill-formed snippets are contrary to that goal.

## **Fixes:**
* Comment out the JSX code inside the notebook `docs/tutorials/llm_chain` with a special directive `<!-- HIDE_IN_NB` (closed with `HIDE_IN_NB -->`). This makes the JSX code "invisible" when viewed in a notebook setting.
* Add a custom preprocessor that runs process_cell and just erases these comment strings. This makes sure they are rendered when converted to MDX.
* Minor tweak: Refactor some of the Markdown instructions into an executable codeblock for better experience when running as a notebook.
* Minor tweak: Optionally try to get the environment variables from a `.env` file in the repo so the user doesn't have to enter it every time. Depends on the user installing `python-dotenv` and adding their own `.env` file.
* Add an environment variable for "LANGSMITH_PROJECT" (default="default"), per the LangSmith docs, so a local user can target a specific project in their LangSmith account.

**NOTE:** If this PR is approved, and the maintainers agree with the general goal of aligning the notebook execution experience and the doc site UI, I would plan to implement this on the rest of the JSX snippets that are littered in the notebooks.

**NOTE:** I wasn't able to/don't know how to run the linkcheck Makefile commands.


hesreallyhim:
@ccurme would love some feedback on this and the general Issue [here](https://github.com/langchain-ai/langchain/issues/30217) regarding upgrading the quality of the notebooks as executable learning resources. i'm working through most of the notebooks now and can easily make small changes as I go if the team likes the direction

eyurtsev (approved):

eyurtsev:
@ccurme fyi merging this in. This might be a bit confusing for doc writers, but we'll err on the side of making it nicer for first time users. Maybe there's a better solution we can try later

PR #30194, opened 2025-03-10
community: expose Azure Doc Intelligence result

laurentiupiciu:
**Description**: This PR adds a `document_analysis_result` attribute to the `AzureAIDocumentIntelligenceParser` class that stores the result from Azure's document analysis service.
- Azure Document Intelligence is a paid service, with costs incurred per document analysis
- Storing the analysis result as an accessible attribute allows users to:
   1. Perform custom post-processing on the raw analysis data
   2. Cache results during development and testing to avoid redundant API calls and charges
   3. Implement more efficient workflows when working with the same document multiple times

### DIFF
--- a/libs/community/langchain_community/document_loaders/doc_intelligence.py
+++ b/libs/community/langchain_community/document_loaders/doc_intelligence.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Iterator, List, Optional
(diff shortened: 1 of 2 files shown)


eyurtsev (changes requested):
This introduces state / mutability into the parser.

Are you able to re-fetch the result using some identifier?

laurentiupiciu:
Hi @eyurtsev. Thanks for clarification and for suggestion. I'll open a new PR taking into consideration the proposed approach

PR #30216, opened 2025-03-11
core :add support for non-retryable error types in retry decorator

asimchoudhary:
- Description:  This PR introduces a new parameter, `non_retry_error_types`, to the `create_base_retry_decorator` function. This enhancement allows users to specify exceptions that should **never** be retried, even if they are subclasses of retryable exceptions. This prevents unnecessary retries for known invalid inputs, such as `InvalidArgument`, while still allowing retries for transient errors like `ResourceExhausted` and `ServiceUnavailable`.  

- Issue:  [Langchain_google: issue~762](https://github.com/langchain-ai/langchain-google/issues/762)
- Dependencies:  No new dependencies are introduced. The change builds on the existing `tenacity` retry mechanism.  

### DIFF
--- a/libs/core/langchain_core/language_models/llms.py
+++ b/libs/core/langchain_core/language_models/llms.py
@@ -76,6 +76,7 @@ def _log_error_once(msg: str) -> None:
 def create_base_retry_decorator(
     error_types: list[type[BaseException]],
     max_retries: int = 1,
+    non_retry_error_types: Optional[list[type[BaseException]]] = None,
     run_manager: Optional[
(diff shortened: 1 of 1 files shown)


vbarda:
@asimchoudhary thanks for the contribution! looks like it's already been addressed for the underlying issue based on this thread https://github.com/langchain-ai/langchain-google/issues/762, so going to close this one.

for the implementation (if you ever need one), i would also recommend using tenacity's `retry_unless_exception_type` instead of adding custom logic

PR #30225, opened 2025-03-11
community[tool]: added the 'topic' param to TavilySearchResults

e7217:
**description**
Hello,

It seems that some parameters were missing in the existing code, so I have updated the client code to include them.
The added parameter is topic.
image
I also noticed other missing parameters, such as `days`,`time_range`. If needed, I can add them in the next PR.

I would appreciate it if you could review this PR. Thank you!


e7217:
cc @ccurme @vbarda

ccurme (commented):
Thanks for this. We're moving the Tavily search tool into its own package [here](https://github.com/tavily-ai/langchain-tavily), would you mind opening a PR or issue there?

PR #30231, opened 2025-03-11
openai[patch]: support Responses API

ccurme:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -405,6 +406,22 @@ def _handle_openai_bad_request(e: openai.BadRequestError) -> None:
         raise
 
 
+def _is_builtin_tool(tool: dict) -> bool:
+    return set(tool.keys()) == {"type"}
```
baskaryan, line 410: dont builtins have other keys?

Screenshot 2025-03-11 at 12 15 28 PM

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -405,6 +406,22 @@ def _handle_openai_bad_request(e: openai.BadRequestError) -> None:
...
+
+
+def _transform_payload_for_responses(payload: dict) -> dict:
+    updated_payload = payload.copy()
+    if messages := updated_payload.pop("messages"):
+        last_user_message = next(
+            (m for m in reversed(messages) if m.get("role") == "user"), None
+        )
+        if last_user_message:
+            updated_payload["input"] = last_user_message["content"]
```
baskaryan, line 420: input can still be a list of messages no?

Screenshot 2025-03-11 at 12 16 34 PM

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -889,6 +916,35 @@ def _create_chat_result(
...
+                )
+                usage_metadata = _create_usage_metadata_responses(token_usage)
+                message = AIMessage(
+                    content=joined, id=output.id, usage_metadata=usage_metadata
+                )
+                if output.status:
+                    generation_info["status"] = output.status
+                gen = ChatGeneration(message=message, generation_info=generation_info)
+                generations.append(gen)
+        llm_output = {"model_name": response.model}
```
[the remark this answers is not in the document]
baskaryan: feel like id, metadata, model, status could all be nice in response_metadata?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -889,6 +916,35 @@ def _create_chat_result(
 
         return ChatResult(generations=generations, llm_output=llm_output)
 
+    def _create_chat_result_responses(
+        self, response: Response, generation_info: Optional[Dict] = None
+    ) -> ChatResult:
+        generations = []
+
+        if error := response.error:
+            raise ValueError(error)
```
baskaryan, line 925: wonder if we can raise a more useful error here. the error object has a code and an error

libs/partners/openai/langchain_openai/chat_models/base.py
baskaryan: also should probably check the content block types of all messages. eg they updated type: text -> type: input_text and type: image_url -> type: input_image

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -889,6 +916,35 @@ def _create_chat_result(
...
+
+        token_usage = response.usage.model_dump() if response.usage else {}
+        generation_info = {}
+        for output in response.output:
+            if output.type == "message":
+                joined = "".join(
```
baskaryan, line 931: should we just keep the content blocks so annotations are included? thats what we do for anthropic right

baskaryan: or we could check for annotations and join if there aren't any

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -405,6 +452,18 @@ def _handle_openai_bad_request(e: openai.BadRequestError) -> None:
...
 
 
+def _is_builtin_tool(tool: dict) -> bool:
+    return "type" in tool and tool["type"] != "function"
+
+
+def _transform_payload_for_responses(payload: dict) -> dict:
+    updated_payload = payload.copy()
+    if messages := updated_payload.pop("messages"):
+        updated_payload["input"] = messages
```
baskaryan, line 462: think we need to upload contents here

baskaryan: also think function tools look a lil different, should probably check for it. they unnested
from {"type": "function", "function": {"name": ...} to {"type": "function", "name": ...}

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -2818,3 +2748,162 @@ def _create_usage_metadata_responses(oai_token_usage: dict) -> UsageMetadata:
...
+                for tool_call in tool_calls:
+                    function_call = {
+                        "type": "function_call",
+                        "name": tool_call["name"],
+                        "arguments": tool_call["arguments"],
+                        "call_id": tool_call["id"],
+                        "id": function_call_ids[tool_call["id"]],
+                    }
+                    input_.append(function_call)
+            if msg.get("content"):
```
ccurme, line 2788: need type `"output_text"` on blocks here
```python
if msg.get("content"):
    if isinstance(msg["content"], list):
        for i, block in enumerate(msg["content"]):
            if "text" in block:
                msg["content"][i]["type"] = "output_text"
    input_.append(msg)
```

PR #30255, opened 2025-03-13
libs: remove redundant words

teslaedison:
remove redundant words


ccurme (commented):
These are grammatically correct.

"Verify that [that operator] is an operator"

"Assert that [that] is what is returned"

[that operator] and [that] are noun phrases.

PR #30262, opened 2025-03-13
community: support in-memory data (Blob.from_data) in all audio parsers

qonnop:
OpenAIWhisperParser, OpenAIWhisperParserLocal, YandexSTTParser do not handle in-memory audio data (loaded via Blob.from_data) correctly. They require Blob.path to be set and AudioSegment is always read from the file system. In-memory data is handled correctly only for FasterWhisperParser so far. I changed OpenAIWhisperParser, OpenAIWhisperParserLocal, YandexSTTParser accordingly to match FasterWhisperParser.
Thanks for reviewing the PR!

### DIFF
--- a/libs/community/langchain_community/document_loaders/parsers/audio.py
+++ b/libs/community/langchain_community/document_loaders/parsers/audio.py
@@ -281,12 +281,8 @@ class OpenAIWhisperParser(BaseBlobParser):
             raise ImportError(
                 "openai package not found, please install it with `pip install openai`"
             )
-        try:
-            from pydub import AudioSegment
-        except ImportError:
-            raise ImportError(
-                "pydub package not found, please install it with `pip install pydub`"
-            )
+
+        audio = _get_audio_from_blob(blob)
 
         if is_openai_v1():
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/document_loaders/parsers/audio.py
```diff
@@ -298,9 +298,16 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
             if self.base_url:
                 openai.api_base = self.base_url
 
-        # Audio file from disk
+        # get the audio
+        if isinstance(blob.data, bytes):
```
vbarda, line 302: could you factor this out into a utility to avoid the same check in several places?

vbarda (commented):
thank you for your contribution!

libs/community/langchain_community/document_loaders/parsers/audio.py
vbarda: no need for a separate file, what you added is exactly what i was thinking. thanks!

vbarda (approved):

PR #30265, opened 2025-03-13
openai[patch]: support structured output via Responses API

ccurme:
Also runs all standard tests using Responses API.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -881,7 +881,10 @@ class BaseChatOpenAI(BaseChatModel):
             generation_info = {"headers": dict(raw_response.headers)}
         elif self._use_responses_api(payload):
             response = self.root_client.responses.create(**payload)
-            return _construct_lc_result_from_responses_api(response)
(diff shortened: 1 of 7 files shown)


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -2837,17 +2845,30 @@ def _construct_responses_api_payload(
...
                 "Can specify at most one of 'response_format' or 'text', received both:"
-                f"\n{response_format=}\n{text=}"
+                f"\n{schema=}\n{text=}"
             )
         # chat api: {"type": "json_schema, "json_schema": {"schema": {...}, "name": "...", "description": "...", "strict": ...}}  # noqa: E501
         # responses api: {"type": "json_schema, "schema": {...}, "name": "...", "description": "...", "strict": ...}  # noqa: E501
-        if response_format["type"] == "json_schema":
-            payload["text"] = {"type": "json_schema", **response_format["json_schema"]}
+        if _is_pydantic_class(schema):
+            # Responses API does not take Pydantic directly
```
baskaryan, line 2858: i think it does if you call responses.parse(..., text_format=pydantic)

PR #30283, opened 2025-03-14
community[patch]: fix bilibili loader handling of multi-page content

homeffjy:
Previously the loader would only extract subtitles from the first page of multi-page videos.

### DIFF
--- a/libs/community/langchain_community/document_loaders/bilibili.py
+++ b/libs/community/langchain_community/document_loaders/bilibili.py
@@ -11,6 +11,7 @@ from langchain_community.document_loaders.base import BaseLoader
 # Pre-compile regular expressions for video ID extraction
 BV_PATTERN = re.compile(r"BV\w+")
 AV_PATTERN = re.compile(r"av[0-9]+")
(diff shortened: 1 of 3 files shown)


libs/community/langchain_community/document_loaders/bilibili.py
```diff
@@ -93,8 +94,17 @@ def _get_bilibili_subs_and_info(self, url: str) -> Tuple[str, dict]:
         if not self.credential:
             return "", video_info
 
+        cid = 0
+        page_match = PAGE_INDEX_PATTERN.search(url)
+        if page_match:
+            cid = video_info["pages"][int(page_match.group()) - 1][
```
ccurme, line 100: Do we need `.group(1)` here?

```python
import re

PAGE_INDEX_PATTERN = re.compile(r"p=(\d+)")

url = "https://www.bilibili.com/video/BV16b4y1R7wP/?p=5"
page_match = PAGE_INDEX_PATTERN.search(url)

int(page_match.group())
```
> ValueError: invalid literal for int() with base 10: 'p=5'

ccurme (approved):

PR #30297, opened 2025-03-15
core: allow pydantic model as input of template

ianchi:
currently if you use langgraph's `create_react_agent` with a pydantic schema and the input of the graph invocation is such a pydantic object, the agent calls the chain with the prompt+model with the pydantic object as input.

This object is not well handled by the prompt template and so the chain does not work as expected.

It is very confusing to allow a pydantic schema in the graph, but then be obliged to invoke the graph with a dict.

This PR adds handling of pydantic objects in the template.

### DIFF
--- a/libs/core/langchain_core/prompts/base.py
+++ b/libs/core/langchain_core/prompts/base.py
@@ -147,6 +147,9 @@ class BasePromptTemplate(
         )
 
     def _validate_input(self, inner_input: Any) -> dict:
+        if isinstance(inner_input, BaseModel):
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Hi @ianchi, thanks for this. I'm not sure we typically pass BaseModels as input into runnables (let me know if you see prominent examples of that anywhere). We are also trying to limit the surface area of Pydantic in LangChain. Going to close for now.

For `create_react_agent`, if possible I'd pass in a function into `prompt` as @vbarda recommends.

PR #30327, opened 2025-03-17
core[patch]: optimize trim_messages

vbarda:
Refactored w/ Claude

Up to 20x speedup! (with theoretical max improvement of `O(n / log n)`)


libs/core/langchain_core/messages/utils.py
```diff
@@ -1231,29 +1238,82 @@ def _first_max_tokens(
...
+    # idx now contains the maximum number of complete messages we can include
+    idx = left
+
+    # Handle partial message inclusion if requested
+    if partial_strategy and idx < len(messages):
+        excluded = messages[idx].model_copy(deep=True)
         included_partial = False
+
+        # Handle list content
         if isinstance(messages[idx].content, list):
```
baskaryan, line 1261:
```suggestion
        if isinstance(excluded.content, list):
```

libs/core/langchain_core/messages/utils.py
```diff
@@ -1231,29 +1238,82 @@ def _first_max_tokens(
...
-            excluded = messages[idx].model_copy(deep=True)
-            num_block = len(excluded.content)
+            # Store current messages to avoid repeated token counting on the same subset
+            base_messages = messages[:idx]
+
             if partial_strategy == "last":
+                # For last strategy, reverse content blocks to keep the last ones
                 excluded.content = list(reversed(excluded.content))
```
baskaryan, line 1267: save ourselves some deep copies:
```python
excluded_content = list(reversed(excluded.content))
num_blocks = len(excluded_content)

left = 0
right = num_blocks
while left < right:
  mid = (left + right + 1) // 2
  excluded.content = excluded_content[:mid]
  if token_counter(base_messages + [excluded]) <= max_tokens:
    left = mid
  else:
    right = mid - 1

if left > 0:
  excluded.content= list(reversed(excluded_content[:left])
  ...
```

baskaryan: same below

eyurtsev (approved):

PR #30332, opened 2025-03-17
refactor(core): drop redundant chat model trace content

baskaryan:


mdrxy:
Is this still relevant @baskaryan?

>  `# TODO: Update once langsmith moves to Pydantic V2 and we can swap run.dict for run.model_dump`

PR #30342, opened 2025-03-18
Core: Adding Azure AI to Supported Chat Models

marlenezw:
- **Description:**  I was testing out `init_chat` and saw that chat models can now be inferred. Azure OpenAI is currently only supported but we would like to add support for Azure AI which is a different package. This PR edits the `base.py` file to add the chat implementation. 
- I don't think this adds any additional dependencies 
- Will add a test and lint, but starting an initial draft PR. 

cc @santiagxf

### DIFF
--- a/libs/langchain/langchain/chat_models/base.py
+++ b/libs/langchain/langchain/chat_models/base.py
@@ -108,6 +108,7 @@ def init_chat_model(
             - 'openai'              -> langchain-openai
             - 'anthropic'           -> langchain-anthropic
             - 'azure_openai'        -> langchain-openai
+            - 'azure_ai             -> langchain-ai
             - 'google_vertexai'     -> langchain-google-vertexai
             - 'google_genai'        -> langchain-google-genai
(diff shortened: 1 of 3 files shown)


marlenezw:
@ccurme, besides editing the base.py file is there anything else we need to do to add Azure AI as a supported model?

ccurme:
> @ccurme, besides editing the base.py file is there anything else we need to do to add Azure AI as a supported model provider?

Hi @marlenezw, I just added it to package extras as well, so we can do `pip install "langchain[azure-ai]"` and it will pick up the dependency on `langchain-azure-ai`. Let me know if that looks good (e.g., the naming of the extra) and will merge.

marlenezw:
Awesome!  This looks good, feel free to merge! I will submit another PR later because we also provide some other MSFT specific models like Phi etc, and that could work for inferring from our package. I will just confirm which ones and add them in a separate PR

ccurme (approved):

PR #30343, opened 2025-03-18
langchain-tests: allow test_serdes for packages outside the default valid namespaces

mattf:
**Description:**

a third party package not listed in the default valid namespaces cannot pass test_serdes because the load() does not allow for extending the valid_namespaces.

test_serdes will fail with -
   ValueError: Invalid namespace: {'lc': 1, 'type': 'constructor', 'id': ['langchain_other', 'chat_models', 'ChatOther'], 'kwargs': {'model_name': '...', 'api_key': '...'}, 'name': 'ChatOther'}

this change has test_serdes automatically extend valid_namespaces based off the ChatModel under test's namespace.


mattf:
@ccurme ptal

ccurme (approved):

Issue #29679, opened 2025-02-07
Got AzureAIException when use open models hosted in Azure (serverless) through LiteLLM - Extra parameters ['n'] are not allowed...


Vivihung:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_community.chat_models import ChatLiteLLM
from langchain_core.messages import HumanMessage

chat = ChatLiteLLM(model="azure_ai/<model_deployment_name>")

messages = [
    HumanMessage(
        content="Translate this sentence from English to French. I love programming."
    )
]

# BadRequestError: litellm.BadRequestError: Azure_aiException - {"detail":"Extra parameters ['n'] are not allowed when extra-parameters is not set or set to be 'error'. Set extra-parameters to 'pass-through' to pass to the model."}
chat(messages)
```

### Error Message and Stack Trace (if applicable)

```
---------------------------------------------------------------------------
HTTPStatusError                           Traceback (most recent call last)
File c:\code\code-maestro\.venv\Lib\site-packages\litellm\llms\custom_httpx\llm_http_handler.py:107, in BaseLLMHTTPHandler._make_common_sync_call(self, sync_httpx_client, provider_config, api_base, headers, data, timeout, litellm_params, stream)
    [106](file:///C:/code/code-maestro/.venv/Lib/site-packages/litellm/llms/custom_httpx/llm_http_handler.py:106) try:
--> [107](file:///C:/code/code-maestro/.venv/Lib/site-packages/litellm/llms/custom_httpx/llm_http_handler.py:107)     response = sync_httpx_client.post(
    [108](file:///C:/code/code-maestro/.venv/Lib/site-packages/litellm/llms/custom_httpx/llm_http_handler.py:108)         url=api_base,
    [109](file:///C:/code/code-maestro/.venv/Lib/site-packages/litellm/llms/custom_httpx/llm_http_handler.py:109)         headers=headers,
    [110](file:///C:/code/code-maestro/.venv/Lib/site-packages/litellm/llms/custom_httpx/llm_http_handler.py:110)         data=json.dumps(data),
[...]

PR #30360, opened 2025-03-19
community: aligning ChatLiteLLM default parameters with litellm

florianchappaz:
**Description:**
Since `ChatLiteLLM` is forwarding most parameters to `litellm.completion(...)`, there is no reason to set other default values than the ones defined by `litellm`.

In the case of parameter 'n', it also provokes an issue when trying to call a serverless endpoint on Azure, as it is considered an extra parameter. So we need to keep it optional.

We can debate about backward compatibility of this change: in my opinion, there should not be big issues since from my experience, calling `litellm.completion()` without these parameters works fine.

**Issue:** 
- #29679 

**Dependencies:** None

### DIFF
--- a/libs/community/langchain_community/chat_models/litellm.py
+++ b/libs/community/langchain_community/chat_models/litellm.py
@@ -259,7 +259,7 @@ class ChatLiteLLM(BaseChatModel):
     organization: Optional[str] = None
     custom_llm_provider: Optional[str] = None
     request_timeout: Optional[Union[float, Tuple[float, float]]] = None
-    temperature: Optional[float] = 1
+    temperature: Optional[float] = None
     """Run inference with this temperature. Must be in the closed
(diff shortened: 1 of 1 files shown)


ccurme (approved):
Not backward-compatible (as you point out), but langchain-community does not strictly follow semver as it's not feasible given the number of integrations. Seems like cost is small compared to the improvement (e.g., for Azure).

We are recommending that community integrations be carved out into separate packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-litellm`)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the models, and you have control over versioning.

Let me know what you think!

Added an issue to track: https://github.com/langchain-ai/langchain/issues/30368

PR #30363, opened 2025-03-19
langchain: add ParrotFakeChatModel support to init_chat_model

milosz-l:
**Description:** This PR adds support for initializing ParrotFakeChatModel through the `init_chat_model` interface using the 'fake' model provider, making it easier to write tests that don't require external API credentials or calls. It's useful when building something provider-agnostic that relies heavily on `init_chat_model`.

Example code looks like this (I have added it to the docs):

```python

# ParrotFakeChatModel echoes back the last message
parrot = init_chat_model("fake:parrot")
result = parrot.invoke([HumanMessage(content="Hello")])
assert result.content == "Hello"
```

**Dependencies:** None - uses existing ParrotFakeChatModel from langchain_core

**Twitter handle:** I'm rather using [LinkedIn](https://www.linkedin.com/in/milosz-lopatto/), but if it's not supported, then my Twitter handle is [@miloszl_](https://x.com/miloszl_)


ccurme (commented):
Thanks, I think this is a reasonable idea but I hesitate because it's unclear that this is the most useful chat model to use for testing purposes.

See for example [GenericFakeChatModel](https://python.langchain.com/api_reference/core/language_models/langchain_core.language_models.fake_chat_models.GenericFakeChatModel.html). On the sidebar there you'll see five other chat models intended for testing.

1. Including multiple of them will pollute / crowd the options
2. If we choose Parrot now and folks want to use a different one, changing it is breaking
3. `init_chat_model` largely serves to manage dependencies / imports of integration packages, which is not necessary here since anyone using any langchain package can import `ParrotFakeChatModel` from `langchain-core`.

milosz-l:
Hi @ccurme! Thank you for your feedback!

When building workflows that use `init_chat_model` multiple times in different places **with separate configuration**, having fake model support would be valuable for testing existing code without making changes to the code that relies on the `init_chat_model` pattern. That's why I came up with this idea.

I understand your concerns. I actually started the implementation with support for different fake models, but then I decided, that for now it might be better make this proposed change as small as possible to avoid the pollution you mentioned.

I would propose reconsidering these options:
1. Using `GenericFakeChatModel` instead of `ParrotFakeChatModel` and making the interface look something like this:
```python
fake_model = init_chat_model("fake_provider:fake_model")
```
2. Adding other fake chat models as an option.

If you'd prefer to see the code before making the decision, please let me know and I can provide it. However, if there's no chance for support of fake models in `init_chat_model`, I'll understand it.

Thank you for the great work!

PR #30370, opened 2025-03-19
community: Azure Document Intelligence parser features not available fixed

LouisAuneau:

- **Description:** Azure Document Intelligence OCR solution has a *feature* parameter that enables some features such as high-resolution document analysis, key-value pairs extraction, ... In langchain parser, you could be provided as a `analysis_feature` parameter to the constructor that was passed on the `DocumentIntelligenceClient`.
However, according to the `DocumentIntelligenceClient` [API Reference](https://learn.microsoft.com/en-us/python/api/azure-ai-documentintelligence/azure.ai.documentintelligence.documentintelligenceclient?view=azure-python), this is not a valid constructor parameter. It was therefore remove and instead stored as a parser property that is used in the `begin_analyze_document`'s `features` parameter  (see [API Reference](https://learn.microsoft.com/en-us/python/api/azure-ai-formrecognizer/azure.ai.formrecognizer.documentanalysisclient?view=azure-python#azure-ai-formrecognizer-documentanalysisclient-begin-analyze-document)). 
I also removed the check for "Supported features" since all features are supported out-of-the-box. Also I did not check if the provided `str` actually corresponds to the Azure package enumeration of features, since the `ValueError` when creating the enumeration object is pretty explicit.
Last caveat, is that some features are not supported for some kind of documents. This is documented inside Microsoft documentation and exception are also explicit.
- **Issue:** N/A
- **Dependencies:** No
- **Twitter handle:** @Louis___A


libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
```diff
@@ -45,23 +45,6 @@ def __init__(
         if api_version is not None:
             kwargs["api_version"] = api_version
 
-        if analysis_features is not None:
-            _SUPPORTED_FEATURES = [
-                DocumentAnalysisFeature.OCR_HIGH_RESOLUTION,
-            ]
-
-            analysis_features = [
```
ccurme, line 53: below, we are passing `features=analysis_features` into `DocumentIntelligenceClient`. This is changing what features get passed in there, correct? Is that a breaking change for users?

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
```diff
@@ -45,23 +45,6 @@ def __init__(
         if api_version is not None:
             kwargs["api_version"] = api_version
 
-        if analysis_features is not None:
-            _SUPPORTED_FEATURES = [
-                DocumentAnalysisFeature.OCR_HIGH_RESOLUTION,
```
ccurme, line 50: Do we know why this warning was there in the first place? Is your view that it was completely erroneous? Seems suspicious.

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
```diff
@@ -97,6 +83,7 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:
                 body=file_obj,
                 content_type="application/octet-stream",
                 output_content_format="markdown" if self.mode == "markdown" else "text",
+                features=self.features,
```
ccurme, line 86: Do we need to pass these in on each invocation of the client, or can we pass them in as an init parameter, as was the case before?

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
LouisAuneau: I don't know the reason why the features were restricted in the first place to be honest. But with my implementation all features are available out-of-the-box. The only exception is that some features are not available for some types of documents, but this wasn't handle before either, and the service error returned is quite explicit.

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
LouisAuneau: As stated in the PR description, if you look at the class reference in Microsoft site ([here](https://learn.microsoft.com/en-us/python/api/azure-ai-documentintelligence/azure.ai.documentintelligence.documentintelligenceclient?view=azure-python)), there is no features init parameter. It didn't raised an error because the class accepts `kwargs`, but the features were not passed on the actual service invocation.

ccurme: Thanks, should we delete `features=analysis_features` below where we pass them into `DocumentIntelligenceClient`?

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
```diff
@@ -71,6 +54,9 @@ def __init__(
         )
         self.api_model = api_model
         self.mode = mode
+        self.features = [
+            DocumentAnalysisFeature(feature) for feature in analysis_features
```
ccurme, line 58: The default for `analysis_features` is None, will this raise an error on init for all users who aren't specifying `analysis_features`? Linter is complaining about this.

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
LouisAuneau: Oh sorry, I had forgotten to update the base branch. I did this, as well as fixing the linter issues.

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
LouisAuneau: Fixed !

libs/community/langchain_community/document_loaders/parsers/doc_intelligence.py
```diff
@@ -45,32 +45,19 @@ def __init__(
         if api_version is not None:
             kwargs["api_version"] = api_version
 
-        if analysis_features is not None:
-            _SUPPORTED_FEATURES = [
-                DocumentAnalysisFeature.OCR_HIGH_RESOLUTION,
```
Pathoomix, line 50: https://github.com/langchain-ai/langchain/actions/workflows/check_diffs.yml/badge.svg

Pathoomix, line 50: git clone https://github.com/googlecloudplatform/deploystack-nosql-client-server

Pathoomix, line 50: 51 -                     googlecloudplatformdeploystack-nosql-client-server

ccurme (approved):

PR #30373, opened 2025-03-19
core[patch]: add util for approximate token counting

vbarda:


ccurme (approved):

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+
+def count_tokens_approximately(
+    messages: list[BaseMessage],
+    token_length: int = DEFAULT_APPROXIMATE_TOKEN_LENGTH,
+    extra_tokens_per_message: int = DEFAULT_EXTRA_TOKENS_PER_MESSAGE,
+    include_name: bool = True,
+) -> int:
+    """Count the number of tokens in a list of messages using a simple approximation.
+
+    Calculates the combined number of characters in the message content, role and name,
```
eyurtsev, line 1442:
```suggestion
    The token count includes the message content, role and optionally message name. + add a sentence to eplain how to tool calls / tool messages are handled
```

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+    as well as tool calls in AI messages and tool call IDs in tool messages, when applicable,
+    and divides by the token length to get the number tokens.
+
+    Args:
+        messages: List of messages to count tokens for.
```
eyurtsev, line 1447:
```suggestion
        messages: List of messages to count
```

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+        token_length: Number of characters per token to use for the approximation.
```
eyurtsev, line 1448:
```suggestion
        chars_per_token: Approximate number of characters per token
```

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
     ]
+
+
+DEFAULT_APPROXIMATE_TOKEN_LENGTH = 4
+DEFAULT_EXTRA_TOKENS_PER_MESSAGE = 3
+
+
+def count_tokens_approximately(
+    messages: list[BaseMessage],
+    token_length: int = DEFAULT_APPROXIMATE_TOKEN_LENGTH,
```
eyurtsev, line 1436:
```suggestion
    token_length: float = DEFAULT_APPROXIMATE_TOKEN_LENGTH,
```

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+    extra_tokens_per_message: int = DEFAULT_EXTRA_TOKENS_PER_MESSAGE,
```
eyurtsev, line 1437: Probably float this one as well

```suggestion
    extra_tokens_per_message: float = DEFAULT_EXTRA_TOKENS_PER_MESSAGE,
```

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+        if message.name and include_name:
+            message_chars += len(message.name)
+
+        # NOTE: we're rounding up per message to ensure that the token counts
+        # are always consistent
+        token_count += math.ceil(message_chars / token_length)
+
+        # add extra tokens per message
+        token_count += extra_tokens_per_message
+
```
eyurtsev, line 1495: ceiling on the token count here

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
         for tool_call in tool_calls
     ]
+
+
+DEFAULT_APPROXIMATE_TOKEN_LENGTH = 4
+DEFAULT_EXTRA_TOKENS_PER_MESSAGE = 3
+
+
+def count_tokens_approximately(
+    messages: list[BaseMessage],
```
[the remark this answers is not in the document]
vbarda: yep, will add support

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+
+        role = _get_message_openai_role(message)
+        message_chars += len(role)
+
+        if message.name and include_name:
+            message_chars += len(message.name)
+
+        # NOTE: we're rounding up per message to ensure that the token counts
+        # are always consistent
+        token_count += math.ceil(message_chars / token_length)
```
eyurtsev, line 1491: Ah missed this ceil. Does this ceil need to be inside the for loop per message? (Is that to make sure that an algorithm on top for slicing things works correclty?

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,72 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
         }
         for tool_call in tool_calls
     ]
+
+
+DEFAULT_APPROXIMATE_TOKEN_LENGTH = 4
```
eyurtsev, line 1430: Why are there global constants for these? Perhaps inline the constant. The key thing is documenting the reason for the constant properly.

Ideally each constant gets a LINK to a reference / short explanation why it's necessary what it does

vbarda: mostly style habit, inline is fine
will add links / explanations

libs/core/langchain_core/messages/utils.py
vbarda: yea, this is done so that you don't get different token counts if you pass different subsets of the same message list, e.g. `sum(count_tokens([m]) for m in messages)` should be the same as `count_tokens(messages)` IMO

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,78 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+) -> int:
+    """Approximate the total number of tokens in messages.
+
+    The token count includes stringified message content, role, and (optionally) name.
+    - For AI messages, the token count also includes stringified tool calls.
+    - For tool messages, the token count also includes the tool call ID.
+
+    Args:
+        messages: List of messages to count tokens for.
+        chars_per_token: Number of characters per token to use for the approximation.
```
eyurtsev, line 1445: in the doc-string perhaps explain why the type is float rather than int

eyurtsev (approved):

libs/core/langchain_core/messages/utils.py
```diff
@@ -1424,3 +1425,78 @@ def _convert_to_openai_tool_calls(tool_calls: list[ToolCall]) -> list[dict]:
...
+
+    .. versionadded:: 0.3.46
+    """
+    token_count = 0.0
+    for message in convert_to_messages(messages):
+        message_chars = 0
+        if isinstance(message.content, str):
+            message_chars += len(message.content)
+
+        # TODO: add support for approximate counting for image blocks
```
eyurtsev, line 1470: Is there a simple rule by looking at the number of bytes for data uris?

PR #30375, opened 2025-03-19
langchain-tests: skip instead of passing image message tests

mattf:
**Description:** use skip for image message tests


ccurme (approved):
Thanks!

PR #30379, opened 2025-03-19
langchain-groq: Add response metadata when streaming

injeniero:
- **Description:** Add missing `model_name` and `system_fingerprint` metadata when streaming.


ccurme (commented):
Thanks for this.

Do we have guarantees that model name and system fingerprint are only emitted once during the stream? If not, when aggregating AIMessageChunks during a stream we will append them (e.g., we will get `"model_name": "<model name><model name><model name>..."`).

One option (best if we can do it): only add these to a chunk during a certain point in the stream (e.g., if there's a `"finish_reason"` or similar final chunk).

Another option: count chunks with these response metadata in [this test](https://github.com/langchain-ai/langchain/blob/aa5ac9279a9315ed6c7eb006d9c1c93532f3dd1f/libs/partners/groq/tests/integration_tests/test_chat_models.py#L85) and raise an error if there are multiple.

injeniero:
Groq only emits the total token usage in a stream response with a final chunk with just token usage information in `x_groq`. This is what I observed, and also partially [documented](https://console.groq.com/docs/api-reference#chat-create:~:text=Example%20code.-,stream_options,-object%20or%20null). They mention that property, but it is not present in their latest SDK, so they send it just by setting streaming.

I agree it would be better to use `finish_reason` instead as a more robust solution. I'm also able to use `self.model` now, which is how the model_name is being filled elsewhere.

ccurme (approved):

ccurme:
Thank you!

injeniero:
Hi @ccurme, when can we expect a new release of langchain_groq?

PR #30382, opened 2025-03-19
refactor(core): remove unused Context API

hinthornw:
Never documented and makes it easier to create memory leaks


nfcampos (approved):

eyurtsev:
I'm in favor, but are we sure no one is using it?

libs/core/langchain_core/beta/runnables/context.py
```diff
@@ -1,401 +0,0 @@
...
-        steps: The runnable steps.
-
-    Returns:
-        The patched runnable config.
-    """
-    return _config_with_context(config, steps, _setter, _getter, threading.Event)
-
-
-@beta()
-class ContextGet(RunnableSerializable):
```
baskaryan, line 157: out of abundance caution maybe we leave stup class with nice deprecation message on init?

baskaryan:
> I'm in favor, but are we sure no one is using it?

imo pretty safe to remove. esp because its beta. but i still would like a stub class + warning

eyurtsev:
OK let's remove and add a stub for a warning so it's clear that the code is gone

eyurtsev:
Porting to v1 branch

eyurtsev:
Taking over

PR #30395, opened 2025-03-20
langchain-tests: allow image specific params for image model tests

mattf:
**Description:** allow image specific model params for image tests

this is important for ChatModels where model="a" may support all the chat features but not image messages, and model="a-vision" supports image messages but not all other chat features

those ChatModel test classes can override `image_model_params` similar to `chat_model_params`

### DIFF
--- a/libs/standard-tests/langchain_tests/integration_tests/chat_models.py
+++ b/libs/standard-tests/langchain_tests/integration_tests/chat_models.py
@@ -294,6 +294,18 @@ class ChatModelIntegrationTests(ChatModelTests):
             def supports_image_inputs(self) -> bool:
                 return True
 
+        If the model that supports images requires special initialization, override
+        the image_model_params property. This may happen if you need to pass a different
+        model name.
+
+        Example:
+
+        .. code-block:: python
+
+            @property
+            def image_model_params(self) -> dict:
+                return {"model": "image-model-001"}
+
     .. dropdown:: supports_video_inputs
 
         Boolean property indicating whether the chat model supports image inputs.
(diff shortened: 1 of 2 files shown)


mattf:
@ccurme please take a look

ccurme (commented):
Thanks for this. I see the appeal of this but IMO we want to think through how to do this generally-- i.e., use different models for different features. This could apply to reasoning, other modalities like audio, forced tool calling, etc.

It's possible that the way to do it is to create `<feature>_model` fixtures and `<feature>_model_params` properties but this might get crowded.

We are considering some related work around enumerating model capabilities and will consider how to address this problem as well.

For now, can we do something like this:
```python
def test_my_image_model() -> None:

    class MySpecialTests(ChatModelIntegrationTests):
        @property
        def chat_model_class(self) -> Type[BaseChatModel]:
            return ChatXYZ

        @property
            return {"model": "image-model"}

        @property
            return True

    test_instance = MySpecialTests()
    model = test_instance.chat_model_class(**test_instance.chat_model_params)
    MySpecialTests().test_image_inputs(model)
```

mattf:
@ccurme thank you for the guidance and suggestion. while waiting for a general solution, i'll follow this pattern -

```
    @pytest.mark.xfail(reason="Model may not support images")
    def test_image_inputs(
        self, image_model: BaseChatModel
    ) -> None:
        super().test_image_inputs(image_model)
```

ccurme:
@mattf I might be misunderstanding but image tests should only run if you override the `supports_image_inputs` property to True. It defaults to False: https://github.com/langchain-ai/langchain/blob/a8ce63903df90e714a60be4ff7c156c1d736b8a8/libs/standard-tests/langchain_tests/unit_tests/chat_models.py#L157-L161

mattf:
@ccurme in my case the ChatModel supports both chat models as well as vlm models. the chat models don't support images and the vlm models don't support the full chat feature set. to properly test the ChatModel implementation i need to test chat functionality against a chat model and image functionality against a vlm model.

PR #30423, opened 2025-03-21
deepseek: temporarily bypass tests

ccurme:
Deepseek infra is not stable enough to get through integration tests during release.

Previous two attempts had two tests time out, they both pass locally.


Deezy33-creator (commented):
[devin adams](https://www.paypalobjects.com/marketing/web23/us/en/quantum-leap/smb/custom-checkout/product-scroll-01-advanced_size-all.png?quality=75&width=900&format=webp)

PR #30430, opened 2025-03-22
community：ChatLiteLLM support output reasoning content

Misakar-0v0:


libs/community/langchain_community/chat_models/litellm.py
```diff
@@ -134,6 +134,8 @@ def _convert_delta_to_message_chunk(
     content = _dict.get("content") or ""
     if _dict.get("function_call"):
         additional_kwargs = {"function_call": dict(_dict["function_call"])}
+    elif _dict.get("reasoning_content"):
+        additional_kwargs = {"reasoning_content": _dict["reasoning_content"]}
```
ccurme, line 138: should this be `elif`? i.e., is it possible for a response to include both function calls and reasoning content?

ccurme (approved):
Thanks for this, could you add a test [here](https://github.com/langchain-ai/langchain/blob/master/libs/community/tests/integration_tests/chat_models/test_litellm.py)? Would welcome that in a second PR.

Misakar-0v0:
> Thanks for this, could you add a test [here](https://github.com/langchain-ai/langchain/blob/master/libs/community/tests/integration_tests/chat_models/test_litellm.py)? Would welcome that in a second PR.

Thank you for your review, I will add the test.

Issue #29933, opened 2025-02-23
`langchain_community.vectorstores.DuckDB` requires pandas - add a helpful (error) message


DvGils:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
import duckdb
from langchain_community.vectorstores import DuckDB
from langchain_core.documents import Document
from langchain_openai import OpenAIEmbeddings

conn = duckdb.connect(
    "mydb.duckdb",
)

documents = [
    Document(page_content="This is a document!"),
    Document(page_content="This is another document!"),
]


database = DuckDB(connection=conn, embedding=OpenAIEmbeddings())

database.add_documents(documents)

# This works perfectly fine:
query = "SELECT text FROM embeddings"
document_text = conn.execute(query).fetchall()
print(document_text)

# Without pandas installed, this will throw silently (after some time) in Windows
# and will raise a Segmentation fault error in Linux:
retriever = database.as_retriever()
another_document = retriever.invoke("another document")
print(another_document)

# Same goes for `search` and other methods
another_document = database.search("another document", search_type="similarity")
print(another_document)
```

### Error Message and Stack Trace (if applicable)

For the life of me, I could not get any sort of message using Windows. Seems to raise silently and the program just dies. On Ubuntu, however, I did get an (unhelpful) error message: `Segmentation fault (core dumped)`

### Description

Using DuckDB vectorstore requires that `pandas` (or one of pandas' dependencies, not sure) is installed. It took me a long time to figure out why my program did not run because there was no message whatsoever. Running queries on the duckdb database (without langchain) worked perfectly fine.

Could you raise an error (or warning) when DuckDB is initialized without having pandas installed? FYI: I have not tried this for other vectorstores, they may have the same problem.

### System Info

Code does not work with the following dependencies, but works when pandas is installed (python 3.12.3):

aiohappyeyeballs==2.4.6
aiohttp==3.11.12
aiosignal==1.3.2
annotated-types==0.7.0
anyio==4.8.0
appdirs==1.4.4
asgiref==3.8.1
attrs==25.1.0
build==1.2.2.post1
CacheControl==0.14.2
certifi==2025.1.31
cffi==1.17.1
[...]

PR #30435, opened 2025-03-22
feature: added warning when duckdb is used as a vectorstore without pandas

dhruva71:
added warning when duckdb is used as a vectorstore without pandas being installed (currently used for similarity search result processing)

- [ ] **PR title**: "community: added warning when duckdb is used as a vectorstore without pandas"

    - **Description:** displays a warning when using duckdb as a vector store without pandas being installed, as it is used by the `similarity_search` function
    - **Issue:** #29933 
    - **Dependencies:** None


ccurme (approved):
Hi, thanks for this. I hesitate to raise an error on init that we weren't raising before, especially if other methods function without pandas (and take some precautions to work if pandas is not installed). So I changed this to a warning inside `similarity_search`.

This method is just using pandas to parse out the result. We should be able to use `.fetchall()` instead of `.fetchdf()` to remove the need for pandas altogether. Would welcome a PR to that effect if tested.

dhruva71:
On it. I'll test and update.

PR #30442, opened 2025-03-23
Fix typo in with_structured_output function of class BaseChatOpenAI

rishabh1024:
docs: Correct description of 'include_raw' parameter in Runnable outputs

**Description:** Corrected the description of the 'include_raw' parameter in Runnable outputs. Changed "Otherwise, if ``include_raw`` is False then Runnable outputs a dict." to "Otherwise, if ``include_raw`` is True then Runnable outputs a dict." to align with the correct information mentioned on the following line.

**Issue:** N/A

**Dependencies:** None

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -1501,7 +1501,7 @@ class BaseChatOpenAI(BaseChatModel):
         Returns:
             A Runnable that takes same inputs as a :class:`langchain_core.language_models.chat.BaseChatModel`.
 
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this. The existing description is correct.

- include_raw False with Pydantic schema --> Pydantic object
- include_raw False with non-Pydantic schema (e.g., schema is TypedDict or JSON Schema) --> dict output
- include_raw True --> dict output

rishabh1024:
Hey, Thanks for getting back! I misunderstood  the word "Otherwise". 
I assumed this is for the opposite case for include_raw=True.

PR #30444, opened 2025-03-23
community: Fix retrieval node objects

galshubeli:
- **Description:** This PR fixes conditions where FalkorDB queries contain FalkorDB node objects and converts them into a dictionary structure.
  - **Issue:** When the retrieval contains FalkorDB objects, the Q&A step cannot handle this context as natural language.
  - **Dependencies:** falkordb-py

### DIFF
--- a/libs/community/langchain_community/chains/graph_qa/falkordb.py
+++ b/libs/community/langchain_community/chains/graph_qa/falkordb.py
@@ -4,8 +4,8 @@ from __future__ import annotations
 
 import re
 from typing import Any, Dict, List, Optional
-from falkordb.node import Node
 
+from falkordb.node import Node
 from langchain.chains.base import Chain
 from langchain.chains.llm import LLMChain
 from langchain_core.callbacks import CallbackManagerForChainRun
--- a/libs/community/langchain_community/chains/graph_qa/falkordb.py
+++ b/libs/community/langchain_community/chains/graph_qa/falkordb.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import re
 from typing import Any, Dict, List, Optional
+from falkordb.node import Node
 
 from langchain.chains.base import Chain
 from langchain.chains.llm import LLMChain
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Can you add a test that demonstrates what is being fixed? I assume the string representation for Node is nonsensical? This is a change in the interface so we need to demonstrate that this is a bug fix (i.e., the existing behavior is not useful).

galshubeli:
> Can you add a test that demonstrates what is being fixed? I assume the string representation for Node is nonsensical? This is a change in the interface so we need to demonstrate that this is a bug fix (i.e., the existing behavior is not useful).



> Can you add a test that demonstrates what is being fixed? I assume the string representation for Node is nonsensical? This is a change in the interface so we need to demonstrate that this is a bug fix (i.e., the existing behavior is not useful).



> Can you add a test that demonstrates what is being fixed? I assume the string representation for Node is nonsensical? This is a change in the interface so we need to demonstrate that this is a bug fix (i.e., the existing behavior is not useful).

galshubeli:
> Can you add a test that demonstrates what is being fixed? I assume the string representation for Node is nonsensical? This is a change in the interface so we need to demonstrate that this is a bug fix (i.e., the existing behavior is not useful).

You're right. This is a bug fix for the string representation of nodes. In other graph providers, it is implemented in the graph file (see the image below). I believe the correct place for this is when the user needs the query for context. Do you agree?

Could you explain how to add a test for this?
image

ccurme:
@galshubeli For testing, we'd ideally add a unit test into a (new) file [here](https://github.com/langchain-ai/langchain/tree/master/libs/community/tests/unit_tests/chains). It should test what context is being supplied to a [(fake) chat model](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/language_models/fake_chat_models.py). It looks like we return the context in intermediate steps, so you should be able to access that in the test.

The test should fail for the implementation that's currently on `master`, and demonstrate the impact of the change in this PR.

ccurme (commented):
@galshubeli we've now moved `langchain-community` into a standalone repo. Would you mind opening your PR there, along with a unit test?

PR #30449, opened 2025-03-23
community: Add edge properties to the gremlin graph schema

ianmuge:
Description: Extend the gremlin graph schema to include the edge properties, grouped by its triples; i.e: `inVLabel` and `outVLabel`. This should give more context when crafting queries to run against a gremlin graph db

### DIFF
--- a/libs/community/langchain_community/graphs/gremlin_graph.py
+++ b/libs/community/langchain_community/graphs/gremlin_graph.py
@@ -20,6 +20,8 @@ class GremlinGraph(GraphStore):
     traversal_source (str): The traversal source to use for queries. Defaults to 'g'.
     message_serializer (Optional[Any]): The message serializer to use for requests.
                                         Defaults to serializer.GraphSONSerializersV2d0()
+    include_edge_properties (bool): Whether to include edge properties in
+                                    the gremlin graph schema. Defaults to False
     *Security note*: Make sure that the database connection uses credentials
         that are narrowly-scoped to only include necessary permissions.
         Failure to do so may result in data corruption or loss, since the calling
@@ -46,6 +48,7 @@ class GremlinGraph(GraphStore):
         password: Optional[str] = None,
         traversal_source: str = "g",
         message_serializer: Optional[Any] = None,
+        include_edge_properties: bool = False,
     ) -> None:
         """Create a new Gremlin graph wrapper instance."""
         try:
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/graphs/gremlin_graph.py
```diff
@@ -91,11 +91,22 @@ def refresh_schema(self) -> None:
...
         )
+        edge_properties = (
+            self.client.submit(
+                "g.E().group().by(label)"
+                ".by(project('inVLabel', 'outVLabel','properties')"
+                ".by(inV().label()).by(outV().label()).by(properties().key().dedup()"
+                ".fold()).dedup().fold())"
+            )
+            .all()
+            .result()[0]
```
ccurme, line 102: Are there any possible circumstances in which the line above for `vertex_properties` executes fine, but this raises an error?

libs/community/langchain_community/graphs/gremlin_graph.py
```diff
@@ -105,6 +116,9 @@ def refresh_schema(self) -> None:
                 "Edge labes are the following:",
                 ",".join(edge_schema),
                 f"Vertices have following properties:\n{vertex_properties}",
+                "Edges have the following properties, grouped by label and ",
+                "the distinct inV and outV labels:\n",
+                edge_properties,
```
ccurme, line 121: How will this addition impact existing applications using `GremlinGraph`?

libs/community/langchain_community/graphs/gremlin_graph.py
ianmuge: No, there are no cases I can imagine whereby this can throw errors where the previous steps would have passed

libs/community/langchain_community/graphs/gremlin_graph.py
ianmuge: The addition should complement the context we are attempting to generate and pass on to the llm. It would allow us to query the graph based on edge properties on top of the already available node properties

ccurme: Should we allow users to opt-in to this (e.g., `include_edge_properties: bool = False`) so that we don't change prompts out from under them?

ianmuge: That makes sense, I have modified it as requested: https://github.com/langchain-ai/langchain/pull/30449/commits/6da0571a5ba8fbe4adb0cd9441c421ba45d6b8b4

ccurme (approved):

PR #30474, opened 2025-03-25
community: add Mistral AI callback for token usage counting

hrouis:
**Description:** Add Mistal AI callback for token usage counting based on the Mistral AI pricing page ( in USD)

### DIFF
--- a/libs/community/langchain_community/callbacks/manager.py
+++ b/libs/community/langchain_community/callbacks/manager.py
@@ -13,6 +13,7 @@ from langchain_core.tracers.context import register_configure_hook
 from langchain_community.callbacks.bedrock_anthropic_callback import (
     BedrockAnthropicTokenUsageCallbackHandler,
 )
+from langchain_community.callbacks.mistralai_callback import MistralAiCallbackHandler
 from langchain_community.callbacks.openai_info import OpenAICallbackHandler
 from langchain_community.callbacks.tracers.comet import CometTracer
 from langchain_community.callbacks.tracers.wandb import WandbTracer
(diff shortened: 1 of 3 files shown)


eyurtsev:
Adding @ccurme. We'll likely want to break this apart into a generic callback manager + some metadata about token costs. the generic call back manager will be in `core`. costs will be somewhere else. no changes should be in community

ccurme (commented):
@hrouis I'm adding a callback manager for token usage to langchain-core here: https://github.com/langchain-ai/langchain/pull/30481

It will largely eliminate the need for custom callback managers for individual chat models. I encourage you to try it out in case you have feedback.

The one missing piece would be cost. I'm somewhat ambivalent about hard-coding USD conversions in the library for these reasons:

- We need to keep them up to date. If pricing is updated or new models are added, users could get inaccurate cost information.
- The currency is ambiguous here and I suspect Mistral has a large portion of its user-base that transacts in Euro.

Let me know what you think. If you can use the token counter from core, it should be straightforward to maintain the pricing information locally and do the multiplication.

ccurme:
Hi @hrouis,

Thanks very much for putting this together to share with the rest of the community!

Would appreciate any feedback you have on the recently-released callback manager-- see docs [here](https://python.langchain.com/docs/how_to/chat_token_usage_tracking/#using-callbacks). It will work with Mistral but be sure to update langchain-core and langchain-mistralai.

hrouis:
Works as expected, i am able to retrieve the tokens and compute the cost.

``` python annotate
 usage_metadata = self._callback.usage_metadata
 usage: UsageMetadata = usage_metadata.get(self._config.model)
 input_tokens = usage.get("input_tokens", 0)
 output_tokens = usage.get("output_tokens", 0)
 response += "\n cost: " + f"{get_mistral_ai_token_cost(input_tokens,
                                                            output_tokens,
                                                            self._config.model)}"
```
 output : 

> {'mistral-large-latest': {'input_tokens': 2616, 'output_tokens': 7, 'total_tokens': 2623}}
cost: 0.005274000000000001

PR #30475, opened 2025-03-25
community: add HANA dialect to SQLDatabase

Byunk:
This PR includes support for HANA dialect in SQLDatabase, which is a wrapper class for SQLAlchemy.

Currently, it is unable to set schema name when using HANA DB with Langchain. And, it does not show any message to user so that it makes hard for user to figure out why the SQL does not work as expected.

Here is the reference document for HANA DB to set schema for the session.

- [SET SCHEMA Statement (Session Management)](https://help.sap.com/docs/SAP_HANA_PLATFORM/4fe29514fd584807ac9f2a04f6754767/20fd550375191014b886a338afb4cd5f.html)


libs/community/langchain_community/utilities/sql_database.py
```diff
@@ -465,6 +465,11 @@ def _execute(
                         (self._schema,),
                         execution_options=execution_options,
                     )
+                elif self.dialect == "hana":
+                    connection.exec_driver_sql(
+                        f"SET SCHEMA {self._schema}",
```
ccurme, line 470: Can we use a parameterized query, instead of f-strings, similar to some of the above (`bigquery`, `trino`, `postgresql`, etc)?

eyurtsev: In this case, could you do something like

```python
sanitized_schema = _sanitize(self._schema) # no-op if valid, otherwise raises an 
```

And use the sanitized_schema name in exec_driver_sql

Will verify that the schema name consists of only: letters, digits, and underscores.

eyurtsev (changes requested):
We can merge as soon as we do some sanitization

libs/community/langchain_community/utilities/sql_database.py
Byunk: I added a `sanitize_schema` function and applied it inside of `exec_driver_sql`.

eyurtsev (approved):

PR #30477, opened 2025-03-25
langchain[patch]: propagate kwargs in EnsembleRetriever

ccurme:


libs/langchain/langchain/retrievers/ensemble.py
```diff
@@ -112,7 +112,9 @@ def invoke(
             **kwargs,
         )
         try:
-            result = self.rank_fusion(input, run_manager=run_manager, config=config)
+            result = self.rank_fusion(
+                input, run_manager=run_manager, config=config, **kwargs
```
ccurme, line 116: need to look into this, this may not be safe-- for instance, above we do `kwargs.get("verbose")`. if someone passes `verbose`, it will now be propagated to the retriever.

PR #30481, opened 2025-03-25
core[patch]: add token counting callback handler

ccurme:
Stripped-down version of [OpenAICallbackHandler](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/callbacks/openai_info.py) that just tracks `AIMessage.usage_metadata`.

```python
from langchain_core.callbacks import get_usage_metadata_callback

    return "It's sunny."

tools = [get_weather]
agent = create_react_agent("openai:gpt-4o-mini", tools)

with get_usage_metadata_callback() as cb:
    result = await agent.ainvoke({"messages": "What's the weather in Boston?"})
    print(cb.usage_metadata)
```

### DIFF
--- a/libs/core/langchain_core/callbacks/usage.py
+++ b/libs/core/langchain_core/callbacks/usage.py
@@ -30,11 +30,10 @@ class UsageMetadataCallbackHandler(BaseCallbackHandler):
     .. versionadded:: 0.3.49
     """
 
-    usage_metadata: Optional[UsageMetadata] = None
-
-    def __init__(self) -> None:
+    def __init__(self, usage_metadata: Optional[UsageMetadata] = None) -> None:
         super().__init__()
         self._lock = threading.Lock()
+        self.usage_metadata = usage_metadata
 
     def __repr__(self) -> str:
         return str(self.usage_metadata)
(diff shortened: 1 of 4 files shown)


eyurtsev (commented):
implementation looks good, needs more api reference documentation

libs/core/langchain_core/callbacks/usage.py
```diff
@@ -0,0 +1,100 @@
...
+        .. code-block:: python
+
+            from langchain.chat_models import init_chat_model
+            from langchain_core.callbacks import UsageMetadataCallbackHandler
+
+            llm = init_chat_model(model="openai:gpt-4o-mini")
+
+            callback = UsageMetadataCallbackHandler()
+            results = llm.batch(["Hello", "Goodbye"], config={"callbacks": [callback]})
+            print(callback)
```
eyurtsev, line 28: printing the callback isn't that useful I think.

instead

`print(callback.usage_metadata)` and consider showing example output

eyurtsev (approved):

libs/core/langchain_core/callbacks/usage.py
ccurme: the callback does implement `__repr__`, so it renders OK but I think it's better to do `callback.usage_metadata` as you suggest to show people how to access the underlying dict.

PR #30483, opened 2025-03-25
core[patch]: Remove old accidental commit

eyurtsev:
Remove commented out file that was accidentally added


hinthornw (approved):
Consider it a learning experience

PR #30497, opened 2025-03-26
standard-tests[patch]: require model_name in response_metadata if returns_usage_metadata

ccurme:
We are implementing a token-counting callback handler in `langchain-core` that is intended to work with all chat models supporting usage metadata. The callback will aggregate usage metadata by model. This requires responses to include the model name in its metadata.

To support this, if a model `returns_usage_metadata`, we check that it includes a string model name in its `response_metadata` in the `"model_name"` key.

More context: https://github.com/langchain-ai/langchain/pull/30487


ccurme:
cc @mattf

Issue #30441, opened 2025-03-23
TypeError: Object of type AsyncCallbackManagerForToolRun is not JSON serializable


sei-li-miidas:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
# -*- coding: utf-8 -*-
"""langgraph-agents-with-openai.ipynb

Automatically generated by Colab.

Original file is located at
    https://colab.research.google.com/drive/1S-hf5LsZciW0eTcu3hNUJ0yiQ-6DX_3P
"""

pip install python-dotenv langgraph langchain_core langchain_openai langgraph-checkpoint langgraph-checkpoint-sqlite asyncio

import os
import sys
import json
import re
import pprint
from dotenv import load_dotenv

import warnings

warnings.filterwarnings("ignore")
import logging

# Set basic configs
log_level = os.environ.get("LOG_LEVEL", "INFO").strip().upper()
logging.basicConfig(format="[%(asctime)s] p%(process)s {%(filename)s:%(lineno)d} %(levelname)s - %(message)s")
logger = logging.getLogger(__name__)
logger.setLevel(log_level)

pprint.PrettyPrinter(indent=2, width=100)

_ = load_dotenv(".env")

from langchain_openai import ChatOpenAI

model = ChatOpenAI(
                model="gpt-4o",
                temperature=1,
                top_p=0.999,
                # model_kwargs={"top_k": 250},
            )

from langchain_core.tools import tool
from langchain_core.tools import StructuredTool
import asyncio
from typing import Any

def save_user_name(**arguments: dict[str, Any],) -> str:
    """Save user name

    Args:
        user_name: user name
    """
    print(arguments)
    user_name = arguments.get("user_name")

    print(f"User name saved: {user_name}")
    return f"User name saved: {user_name}"

async def sleep(
        **arguments: dict[str, Any],
    ) -> tuple[str | list[Any] | None]:
    """Sleep for a while

    Args:
        seconds: How long to sleep
    """
    seconds = int(arguments["seconds"])

    print(f"Sleep: {seconds} seconds")
    await asyncio.sleep(seconds)
    return "good"

my_tool = [
    StructuredTool(
        name="save_user_name",
[...]

PR #30510, opened 2025-03-27
core[patch]: prevent langchain_core.tools.base.BaseTool.ainvoke mutating the original input parameters

xzq-xu:
## Description

Issue: Serialization error in #30441
Root cause: When passing external reference objects to ainvoke, the method adds a non-serializable run_manager property during execution. This property shouldn't be propagated externally.


xzq-xu:
The original trigger scenario can be viewed in [issue #30441](https://github.com/langchain-ai/langchain/issues/30441). The example I provided is a simplified version. As long as an external reference is passed to tool.ainvoke, this situation will occur. Once run_manager is added to the original object, whether it is to continue asking questions LLM (JSON serialization) or store it in checkpointer (msgpack), an error will occur.

eyurtsev (approved):

eyurtsev:
@xzq-xu thank you! I moved the fix to a different location in the code and updated the unit tests to remove everything that wasn't strictly necessary.

xzq-xu:
[issue #30441](https://github.com/langchain-ai/langchain/issues/30441)  This issue does not seem to be automatically closed. You may need to operate it manually.

PR #30520, opened 2025-03-27
core[patch]: stop deleting records with "scoped_full" when doc is empty

KeiichiHirobe:
Fix a bug that causes `scoped_full` in index to delete records when there are no input docs.

### DIFF
--- a/libs/core/tests/unit_tests/indexing/test_indexing.py
+++ b/libs/core/tests/unit_tests/indexing/test_indexing.py
@@ -822,6 +822,158 @@ async def test_ascoped_full_fails_with_bad_source_ids(
         )
 
 
+def test_index_empty_doc_scoped_full(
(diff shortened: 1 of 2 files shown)


libs/core/tests/unit_tests/indexing/test_indexing.py
```diff
@@ -822,6 +822,158 @@ async def test_ascoped_full_fails_with_bad_source_ids(
...
+            "num_added": 4,
+            "num_deleted": 0,
+            "num_skipped": 0,
+            "num_updated": 0,
+        }
+
+    with patch.object(
+        record_manager, "get_time", return_value=datetime(2021, 1, 2).timestamp()
+    ):
+        assert index(
```
KeiichiHirobe, line 869: I've confirmed that this test failed with `{'num_deleted': 4} != {'num_deleted': 0}` without this PR change.

eyurtsev (approved):
Thank you

PR #30525, opened 2025-03-27
Docs: Add sambanova cloud embeddings docs

jhpiedrahitao:
- **Description:**  Add samba nova cloud embeddings docs, only samabastudio embeddings were supported, now in the latest release of langchan_sambanova sambanova cloud embeddings is also available


eyurtsev:
@jhpiedrahitao what's the difference between the two? I read through both docs and have no idea what the difference is or which one I should be using. The same will be true of LangChain users.

eyurtsev:
Why isn't this implemented a single client that takes a different URL? (Why are two clients necessary?)

jhpiedrahitao:
Hi @eyurtsev, there are two different SambaNova products, SambaNova cloud is the public access one, SambaStudio is an enterprise platform, we have one for each because, we have some differences in the APIs for cloud (full openai compatible) and for sambastudio (open ai + own api structure), this also applies to LLMs, and chat models (which right now are separated), and now we added embeddings to cloud

eyurtsev (approved):

PR #30532, opened 2025-03-28
community: add feat AsyncOpenAICallbackHandler for asynchronous OpenAI tracking

KhanhD1nh:
- **Title**: "community: addAsyncOpenAICallbackHandler for asynchronous OpenAI tracking"

### DIFF
--- a/libs/community/langchain_community/callbacks/__init__.py
+++ b/libs/community/langchain_community/callbacks/__init__.py
@@ -59,7 +59,6 @@ if TYPE_CHECKING:
     )
     from langchain_community.callbacks.openai_info import (
         OpenAICallbackHandler,
-        AsyncOpenAICallbackHandler,
     )
     from langchain_community.callbacks.promptlayer_callback import (
         PromptLayerCallbackHandler,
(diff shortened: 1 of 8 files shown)


eyurtsev (changes requested):
We do not want an openai specific implementation.

https://github.com/langchain-ai/langchain/pull/30534

@ccurme recently added support into core for a generic implementation, if you'd like to contribute consider contributing an async callback handler directly to core.

We want to separate the cost calculation from tallying up the tokens

eyurtsev:
See https://python.langchain.com/docs/how_to/chat_token_usage_tracking/#using-callbacks for usage w/ new functionality in langchain_core

ccurme (commented):
Thanks for this @KhanhDinh03. As mentioned above we recently added a callback manager into `langchain-core`:
- [Docs](https://python.langchain.com/docs/how_to/chat_token_usage_tracking/#using-callbacks)
- [Implementation](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/callbacks/usage.py)

There should be minimal overhead if running this sync callback manager in an async context (it will context-switch to a new thread, but we are just adding token counts), but I'm happy to discuss or review if you want to add an async version.

PR #30534, opened 2025-03-28
community: Add google genai callback

taeminlee:
## Description
This PR adds a callback handler for Google GenAI (Gemini) models to track token usage and costs when using these models with LangChain.

## Features
- Tracks total tokens, prompt tokens, completion tokens, and cached tokens
- Calculates costs based on current pricing for all Gemini models
- Supports token caching cost calculations

## Usage example
```python
from langchain_community.callbacks import GoogleGenAICallbackHandler

# Create the callback handler
callback = GoogleGenAICallbackHandler()

# Use it with a Google GenAI model
llm = ChatGoogleGenerativeAI(
    model="gemini-2.0-flash", 
    callbacks=[callback]
)
response = llm.invoke("Hello, how are you?")

# View token usage and cost information
print(callback)
```


eyurtsev (changes requested):
We want to generalize the callback handlers so we don't end up with a callback handler per chat model

a-holm:
**Potential Token Mapping Bug:** In `google_genai_callback.py`'s `on_llm_end` method, it appears `prompt_tokens` might be incorrectly assigned from `usage_metadata["output_tokens"]` and `completion_tokens` from `usage_metadata["input_tokens"]`. Could you please double-check the actual structure of `usage_metadata` provided by the `langchain-google-genai` integration and ensure the mapping is correct (usually prompt=input, completion=output)? This impacts both the stored token counts and the calculated cost.

taeminlee:
> @taeminlee we're in the process of generalizing the implementation so there's a single callback implementation across all model providers.
> 
> Please see this callback handler for mode details:
> 
> https://python.langchain.com/docs/how_to/chat_token_usage_tracking/#using-callbacks
> 
> For now you can apply a mapping from model name to cost client side

 It's great to hear that you're working on generalizing the implementation to create a single callback implementation across all model providers.

I'll examine the callback handler in detail through the LangChain documentation link you provided. The interim solution of applying a model name to cost mapping on the client side seems very practical as well.

> **Potential Token Mapping Bug:** In `google_genai_callback.py`'s `on_llm_end` method, it appears `prompt_tokens` might be incorrectly assigned from `usage_metadata["output_tokens"]` and `completion_tokens` from `usage_metadata["input_tokens"]`. Could you please double-check the actual structure of `usage_metadata` provided by the `langchain-google-genai` integration and ensure the mapping is correct (usually prompt=input, completion=output)? This impacts both the stored token counts and the calculated cost.

I've fixed the issue where prompt_tokens and completion_tokens were incorrectly assigned. Thank you for pointing this out. However, as you suggested, implementing a general method and running my PR implementation client-side seems to be the right approach.

I'll close this PR. Thank you to everyone who provided feedback. I look forward to contributing again in the future.

PR #30543, opened 2025-03-28
pytest: specify event loop scope

eyurtsev:
Specify default event loop scope


eyurtsev:
cc @cbornet assuming this is fine w/ regards to blockbuster

PR #30549, opened 2025-03-29
community: Add custom sitemap URL parameter to GitbookLoader

andrasfe:
## Description
This PR adds a new `sitemap_url` parameter to the `GitbookLoader` class that allows users to specify a custom sitemap URL when loading content from a GitBook site. This is particularly useful for GitBook sites that use non-standard sitemap file names like `sitemap-pages.xml` instead of the default `sitemap.xml`.
The standard `GitbookLoader` assumes that the sitemap is located at `/sitemap.xml`, but some GitBook instances (including GitBook's own documentation) use different paths for their sitemaps. This parameter makes the loader more flexible and helps users extract content from a wider range of GitBook sites.
## Issue
Fixes bug [30473](https://github.com/langchain-ai/langchain-community/issues/297) where the `GitbookLoader` would fail to find pages on GitBook sites that use custom sitemap URLs.
## Dependencies
No new dependencies required.
*I've added*:
* Unit tests to verify the parameter works correctly
* Integration tests to confirm the parameter is properly used with real GitBook sites
* Updated docstrings with parameter documentation
The changes are fully backward compatible, as the parameter is optional with a sensible default.


libs/community/tests/unit_tests/document_loaders/test_gitbook.py
```diff
@@ -0,0 +1,193 @@
+import unittest
+from typing import Any
+from unittest.mock import MagicMock, patch
+
+from bs4 import BeautifulSoup
+
+from langchain_community.document_loaders.gitbook import GitbookLoader
+
+
+class TestGitbookLoader(unittest.TestCase):
```
eyurtsev, line 10: The repo uses pytest style tests rather than unittest.TestCase tests, would you mind updating?

libs/community/langchain_community/document_loaders/gitbook.py
```diff
@@ -21,6 +21,7 @@ def __init__(
         content_selector: str = "main",
         continue_on_failure: bool = False,
         show_progress: bool = True,
+        sitemap_url: Optional[str] = None,
```
eyurtsev, line 24:
```suggestion
        *,
        sitemap_url: Optional[str] = None,
```

eyurtsev (commented):
Looks good two minor things to resolve and we can merge!

libs/community/tests/unit_tests/document_loaders/test_gitbook.py
andrasfe: Noted and fixed.

libs/community/langchain_community/document_loaders/gitbook.py
andrasfe: Added change as suggested.

andrasfe: Made all the necessary changes. Thank you for the prompt turnaround!

libs/community/tests/unit_tests/document_loaders/test_gitbook.py
```diff
@@ -0,0 +1,204 @@
...
+
+    soup_info = loader.scrape()
+    paths = loader._get_paths(soup_info)
+
+    # Check that paths were extracted correctly
+    assert len(paths) == 3
+    assert paths == ["/page1", "/page2", "/page3"]
+
+
+@patch("requests.get")
```
eyurtsev, line 120: Thanks!

@andrasfe I don't know if you're familiar with the responses library, but it's generally a **much** cleaner way to patch requests.get / patch / post if you're comparing it to using the built in patch methods.

Going to merge as is (not a blocker). But if you have time to re-write w/ responses will be appreciated. It'll make the code much less brittle!

---

I've observed that the usage of `@patch` w/ strings specifying the patched code introduces more technical debt sometimes than it solves. patch object is slightly better since it's easier to catch namespacing changes, but for network requests `responses` is better in virtually all cases

eyurtsev (approved):
thanks!

Issue #30531, opened 2025-03-27
RunnableBinding with_config does not propagate config factories


mahirshah:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python

from typing import Optional, Any, Dict, List
from uuid import UUID

from langchain_core.callbacks import BaseCallbackHandler
from langchain_core.runnables import RunnableLambda


# prints:
# handler chain_start
# listener chain_start
def test_with_config_then_with_listeners():
    chain = RunnableLambda(lambda x: x * 2).with_config(callbacks=[Handler()]).with_listeners(on_start=lambda run: print("listener chain_start"))
    chain.invoke(1)


# prints:
# handler chain_start
def test_with_listeners_then_with_config():
    chain = RunnableLambda(lambda x: x * 2).with_listeners(on_start=lambda run: print("listener chain_start")).with_config(callbacks=[Handler()])
    chain.invoke(1)


class Handler(BaseCallbackHandler):
    def on_chain_start(
        self,
        serialized: Dict[str, Any],
        inputs: Dict[str, Any],
        *,
        run_id: UUID,
        parent_run_id: Optional[UUID] = None,
        tags: Optional[List[str]] = None,
        metadata: Optional[Dict[str, Any]] = None,
        **kwargs: Any,
    ) -> Any:
        print("handler chain_start")

```

### Description

When trying to attach a config via `with_config` to a runnable binding (such as one created when calling `with_listeners`), the config factories are not propagated to the new runnable binding. Thus, the new runnable binding loses any config factories that were attached to the original runnable binding and the original configs are never called upon invocation.

I imagine the most common scenario this can happen in is the one that is shown in the test case attached: a runnable binding is created with config factories via `with_listeners` and then `with_config` is used to attach a config to this runnable binding. The new runnable binding now only has the config attached via `with_config` and does not call the `with_listeners` callback upon invocation.

[...]

PR #30551, opened 2025-03-30
[Core]: Passing config_factories in  `with_config` for RunnableBinding

keenborder786:
- **Description:**  We were not passing `config_factories` to `RunnableBinding`  method `with_config` which was causing the parent `config_factories` to be lost as shown in the issue but this has been fixed
- **Issue**:  #30531

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -5670,6 +5670,7 @@ class RunnableBinding(RunnableBindingBase[Input, Output]):
             bound=self.bound,
             kwargs=self.kwargs,
             config=cast("RunnableConfig", {**self.config, **(config or {}), **kwargs}),
+            config_factories=self.config_factories,
             custom_input_type=self.custom_input_type,
             custom_output_type=self.custom_output_type,
         )


libs/core/langchain_core/runnables/base.py
```diff
@@ -5670,6 +5670,7 @@ def with_config(
             bound=self.bound,
             kwargs=self.kwargs,
             config=cast("RunnableConfig", {**self.config, **(config or {}), **kwargs}),
+            config_factories=self.config_factories,
```
mahirshah, line 5673: I think this issue happens during other `with_*` methods as well since config_factories do not get passed down -- can we make this change for all related methods as well?

mahirshah (changes requested):
Can we add unit tests for this change?

libs/core/langchain_core/runnables/base.py
keenborder786: Sure

eyurtsev:
@keenborder786 could you add unit tests?

mahirshah:
> * **Description:**  We were not passing `config_factories` to `RunnableBinding`  method `with_config` which was causing the parent `config_factories` to be lost as shown in the issue but this has been fixed
> * **Issue**:  [RunnableBinding with_config does not propagate config factories #30531](https://github.com/langchain-ai/langchain/issues/30531)

Sorry to co-opt this change, but I need this fix in soon as it is affecting app behavior. Added unit tests and additional propagation in this PR: #30603

eyurtsev:
Continuing with #30603

PR #30567, opened 2025-03-31
langchain-openai(partners): Append additional keyword parameters or metadata to the message

houxiyao:
Provides a flexible way to store supplementary information about a message, which can contain custom data that is not part of the standard message structure, allowing additional context or metadata to be carried in the message object.

Can store additional information according to different APIs.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -304,7 +304,7 @@ def _convert_message_to_dict(message: BaseMessage) -> dict:
 
 
 def _convert_delta_to_message_chunk(
-    _dict: Mapping[str, Any], default_class: Type[BaseMessageChunk]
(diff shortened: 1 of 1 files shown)


ccurme (commented):
@houxiyao can you share how you are using this?

houxiyao:
@eyurtsev @ccurme

ccurme (commented):
Hi @houxiyao,

Thanks for the submission. It's a good thought to improve extensibility. For your particular issue I'm wondering if [langchain-qwq](https://pypi.org/project/langchain-qwq/) will work for you. Failing that, I'd recommend using `langchain-deepseek`, which is basically `ChatOpenAI` but it extracts the `reasoning_content` field as you do here.

The reasons this is complicated:
- In streaming case, we need to know if the field is repeated with each chunk (in which case we should just take it once), or if the chunks represent deltas.
- When passing an AIMessage back into the model (as part of a chat history), we need to know how to structure the fields in `additional_kwargs` back into the model payload.

PR #30572, opened 2025-03-31
community[major]: DynamoDBChatMessageHistory bulk add messages, raise errors

wuodar:
This PR addresses two key issues:

- **Prevent history errors from failing silently**: Previously, errors in message history were only logged and not raised, which can lead to inconsistent state and downstream failures (e.g., ValidationError from Bedrock due to malformed message history). This change ensures that such errors are raised explicitly, making them easier to detect and debug.
(Side note: I’m using AWS Lambda Powertools Logger but hadn’t configured it properly with the standard Python logger—my bad. If the error had been raised, I would’ve seen it in the logs 😄) This is a **BREAKING CHANGE**

- **Add messages in bulk instead of iteratively**: This introduces a custom add_messages method to add all messages at once. The previous approach failed silently when individual messages were too large, resulting in partial history updates and inconsistent state. With this change, either all messages are added successfully, or none are—helping avoid obscure history-related errors from Bedrock.


eyurtsev (commented):
This PR is a breaking change in dynamodb implementation, but i think it makes sense otherwise it's masking errors causing user code to fail silently

PR #30581, opened 2025-03-31
openai[patch]: support structured output and tools

baskaryan:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1636,13 +1637,16 @@ def with_structured_output(
...
-            llm = self.bind(
+            bind_kwargs = dict(
                 response_format=response_format,
                 ls_structured_output_format={
                     "kwargs": {"method": method, "strict": strict},
                     "schema": convert_to_openai_tool(schema),
                 },
             )
+            if tools:
+                bind_kwargs["tools"] = [convert_to_openai_tool(t) for t in tools]
```
ccurme, line 1648:
```suggestion
                bind_kwargs["tools"] = [convert_to_openai_tool(t, strict=strict) for t in tools]
```

ccurme (commented):
Why implement `tools` on `with_structured_output`, as opposed to `response_format` on `bind_tools`? I don't think the parsers will give us an AIMessage with tool calls.

```python

class ResponseFormat(BaseModel):
    response: str


    """Get weather at a location."""
    pass

structured_llm = llm.with_structured_output(
    ResponseFormat,
    tools=[get_weather],
    strict=True,
)

structured_llm.invoke("What's the weather in Boston?")
```
We raise an error in [_oai_structured_outputs_parser](https://github.com/langchain-ai/langchain/blob/ecff055096bc137bc10d7933d71016e2af56c06d/libs/partners/openai/langchain_openai/chat_models/base.py#L2771) because there's no parsed response. We'd need to change the behavior there right?

baskaryan:
> Why implement `tools` on `with_structured_output`, as opposed to `response_format` on `bind_tools`? I don't think the parsers will give us an AIMessage with tool calls.
> 
> ```python
> from langchain.chat_models import init_chat_model
> from pydantic import BaseModel
> 
> class ResponseFormat(BaseModel):
>     response: str
> 
> 
> def get_weather(location: str) -> str:
>     """Get weather at a location."""
>     pass
> 
> llm = init_chat_model("openai:gpt-4o-mini")
> 
> structured_llm = llm.with_structured_output(
>     ResponseFormat,
>     tools=[get_weather],
>     strict=True,
> )
> 
> structured_llm.invoke("What's the weather in Boston?")
> ```
> 
> We raise an error in [_oai_structured_outputs_parser](https://github.com/langchain-ai/langchain/blob/ecff055096bc137bc10d7933d71016e2af56c06d/libs/partners/openai/langchain_openai/chat_models/base.py#L2771) because there's no parsed response. We'd need to change the behavior there right?

imagining this would only be used with include_raw=True, since what you really want back is the message. could add a new output mode for that but didn't want to do that preemptively

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -2770,14 +2824,16 @@ def _convert_to_openai_response_format(
...
+) -> Optional[PydanticBaseModel]:
     if parsed := ai_msg.additional_kwargs.get("parsed"):
         if isinstance(parsed, dict):
             return schema(**parsed)
         else:
             return parsed
     elif ai_msg.additional_kwargs.get("refusal"):
         raise OpenAIRefusalError(ai_msg.additional_kwargs["refusal"])
+    elif ai_msg.tool_calls:
+        return None
```
ccurme, line 2836: @baskaryan this change is only to support streaming. The fallback to `parser_none` may not be working properly with `include_raw=True` in streaming, although it works fine with invoke.

I also think it is cleaner to return None here, otherwise we always include a `ValueError` in the `"parsing_error"` key.

If desired we could also only engage this behavior if `tools` are passed into `with_structured_output`, though I think functionally that makes no difference right now.

ccurme (approved):

PR #30599, opened 2025-04-01
core: Add ruff rules S110 and S112

cbornet:


libs/core/langchain_core/sys_info.py
```diff
@@ -22,11 +22,7 @@ def _get_sub_deps(packages: Sequence[str]) -> list[str]:
             continue
 
         for req in required:
-            try:
-                cleaned_req = req.split(" ")[0]
-            except Exception:  # In case parsing of requirement spec fails
-                continue
-
+            cleaned_req = req.split(" ")[0]
```
cbornet, line 25: I don't see how this can raise.

eyurtsev (approved):

PR #30612, opened 2025-04-02
community: YDB Vector Store

vgvoleg:
This PR adds support of YDB as a vector store

[YDB](https://ydb.tech/) is a versatile open-source distributed SQL database. It supports [vector search](https://ydb.tech/docs/en/yql/reference/udf/list/knn) which means it can be used as a vector store with langchain.

### DIFF
--- a/docs/docs/integrations/vectorstores/ydb.ipynb
+++ b/docs/docs/integrations/vectorstores/ydb.ipynb
@@ -1,0 +1,472 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "683953b3",
+   "metadata": {},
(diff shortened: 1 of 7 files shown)


ccurme (commented):
Hello, thanks for this. This adds a net-new integration into langchain-community. We are recommending that new contributions be implemented via separate packages. Would you be interested in publishing an OSS integration package (e.g., `langchain-ydb`)? We've written a walkthrough on this process here:

We are encouraging contributors of LangChain integrations to go this route. This way we don't have to be in the loop for reviews, you're able to properly integration test the vector store, and you have control over versioning.

Let me know what you think!

PR #30617, opened 2025-04-02
chore: adds langchain_gigachat support to init_chat_model and init_embeddings

Rai220:
This PR adds support for the langchain_gigachat integration to the init_chat_model and init_embeddings utility functions. This allows users to initialize GigaChat models in a unified way, consistent with other supported LLM providers.


libs/langchain/langchain/chat_models/base.py
```diff
@@ -412,6 +414,11 @@ def _init_chat_model_helper(
         from langchain_huggingface import ChatHuggingFace
 
         return ChatHuggingFace(model_id=model, **kwargs)
+    elif model_provider == "gigachat":
```
eyurtsev, line 417: Why do the integrations docs have `verify_ssl_certs=False`?

https://python.langchain.com/docs/integrations/chat/gigachat/

Rai220: Hello!

The official GigaChat API requires the installation of a digital certificate issued by the Russian Ministry of Digital Development in order to work over HTTPS, due to local legal restrictions. The langchain_gigachat library allows you to bypass this requirement by using an optional flag. This flag is disabled by default, but we've decided to highlight it in the documentation since most non-corporate users are unlikely to install the certificate.

PR #30620, opened 2025-04-02
community[patch]: Fix empty producer in PDF Parsers

pprados:
If a pdf file doesn't have a “producer” in metadata, it generates an exception.

Fix this problem.


libs/community/langchain_community/document_loaders/parsers/pdf.py
```diff
@@ -723,7 +723,8 @@ def lazy_parse(self, blob: Blob) -> Iterator[Document]:  # type: ignore[valid-ty
             pages = PDFPage.get_pages(pdf_file_obj, password=self.password or "")
             rsrcmgr = PDFResourceManager()
             doc_metadata = _purge_metadata(
-                self._get_metadata(pdf_file_obj, password=self.password or "")
+                {"producer": "PDFMiner", "creator": "PDFMiner", "creationdate": ""}
```
eyurtsev, line 726: why do we have both a creator and a producer?

eyurtsev (approved):

Issue #29277, opened 2025-01-17
docs: Replace `initialize_agent` use with langgraph.prebuilt.create_react_agent


efriis:
### Privileged issue

- [x] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Currently there's a bunch of tools and other integrations that use `initialize_agent` in their docs page (search langchain/docs for "initialize_agent")

This function has been deprecated since 0.1 and should be replaced with basic langgraph usage as a demo

Deprecated function: https://python.langchain.com/api_reference/langchain/agents/langchain.agents.initialize.initialize_agent.html

Example page using create_react_agent in an example: https://python.langchain.com/docs/integrations/tools/gmail/#use-within-an-agent

This issue doesn't need to be resolved by a single PR and can be tackled incrementally! Just tag this issue for tracking purposes :)

aybdee:
I want to try working on this

rawathemant246:
Hii, @efriis  Happy sunday hope you are on your two wheeler today, can i start contributing on the documentation changes ?

Thejaswi05:
@efriis do you expect the changes in the documentation for the tools and integrations using `initialize_agent'  listed at the bottom of the page you shared? 

Such as: 
[AINetwork Toolkit](https://python.langchain.com/docs/integrations/tools/ainetwork/)

[AWS Lambda](https://python.langchain.com/docs/integrations/tools/awslambda/)

turboslapper:
Hi everyone,  

I’ve partially updated the documentation to replace deprecated references to `initialize_agent` with `langgraph.prebuilt.create_react_agent`. Specifically:  

- I addressed the instances for **page 1 of 3** in the search:  
  **`repo:langchain-ai/langchain path:/^docs\// initialize_agent`**.  
- I plan to work on pages 2 and 3 shortly to complete the updates.  

Additionally, I noticed a recurring pattern in examples, such as:  
```python
agent.run("what is google's stock")
```

This doesn't inherently print the output of the agent.run command, and I considered adding something like:

```python
response = agent.run("what is google's stock")
print(response)
```

to improve clarity for users. However, as this is my first commit, I didn’t want to overstep the scope without further guidance. Let me know if you’d like me to include such improvements!

Looking forward to completing the remaining pages soon.

AffanShaikhsurab:
If no one is working on this issue , can I work on this ?

rthummaluru:
[...]

PR #30630, opened 2025-04-03
docs: Fix deprecated initialize agent in memorize.ipynb

CtrlMj:
- **Description:** Replace the deprecated method of creating agents by:

    1. Remove the imports and code for the initialize_agent function
    2. Add the correct import and code for create_react_agent function from langgraph  

  - **Issue:** #29277
  - **Dependencies:** N/A
  - **Twitter handle:** N/A

### DIFF
--- a/docs/docs/integrations/tools/memorize.ipynb
+++ b/docs/docs/integrations/tools/memorize.ipynb
@@ -115,17 +115,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 20,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
-    "agent = initialize_agent(\n",
-    "    tools,\n",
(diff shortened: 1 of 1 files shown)


ccurme (commented):
`GradientLLM` will raise an error when you pass into `create_react_agent` because it is not a chat model and does not implement `bind_tools`.

Will re-open if you run the notebook to demonstrate that it works!

CtrlMj:
@ccurme should we replace the LLMs with their corresponding chat models when doing this exercise?

ccurme:
As far as I know `create_react_agent` will not function with a legacy [text-completion LLM](https://python.langchain.com/docs/concepts/text_llms/), whereas AgentExecutor can.

The most important thing is that the guides run without error after the update (especially if they run without error before the update).

PR #30656, opened 2025-04-03
core: Add ruff rules PGH (except PGH003)

cbornet:
Add ruff rules PGH: https://docs.astral.sh/ruff/rules/#pygrep-hooks-pgh
Except PGH003 which will be dealt in a dedicated PR.


eyurtsev (approved):

libs/core/pyproject.toml
```diff
@@ -101,7 +101,8 @@ ignore = [
     "ERA",
     "FBT001",
     "FBT002",
-    "PGH",
+    "FIX",
```
eyurtsev, line 104:
```suggestion
```

PR #30664, opened 2025-04-04
community: Add support for custom runtimes to Riza tools

andrewmbenton:
**Description:**
Adds support for Riza custom runtimes to the two Riza code interpreter tools, allowing users to run LLM-generated code that depends on libraries outside stdlib.
**Issue:** N/A
**Dependencies:** None
**Twitter handle:** @rizaio

### DIFF
--- a/libs/community/langchain_community/tools/riza/command.py
+++ b/libs/community/langchain_community/tools/riza/command.py
@@ -66,8 +66,11 @@ class ExecPython(BaseTool):  # type: ignore[override, override]
     handle_tool_error: bool = True
 
     client: Any = None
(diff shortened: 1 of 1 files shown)


libs/community/langchain_community/tools/riza/command.py
```diff
@@ -66,8 +66,11 @@ class ExecPython(BaseTool):  # type: ignore[override, override]
     handle_tool_error: bool = True
 
     client: Any = None
+    runtime_revision_id: Optional[str] = None
```
ccurme, line 69: Might be worth documenting how to use this parameter.

ccurme (approved):

PR #30665, opened 2025-04-04
Langchain Community tool description improvements.

ivanmilevtues:
# Tool description updated for Langchain Community ArXiv and PubMed tool
- **Description**: this PR contains changes in the description of multiple tools in Langchain.
- **Issue**: The issue with the descriptions is that they are incomplete. Being incomplete hinders their utility to the models and leads to runtime errors. The newly introduced descriptions reduce runtime errors without narrowing down the current functionality.
- **Dependencies**: no new dependencies are required for this change
- **Twitter handle**: @ivan_milev21

I have previously submitted PR on the same topic for just two tools. Maybe this change will get more attention as it fixes multiple tools and improves the robustness of the tooling leading to overall better agent experience.


ccurme (commented):
Thanks @ivanmilevtues for submitting this.

There's a problem with this update to community: when we change tool descriptions, users who might be running them in production will find that their prompts change out from under them. This can lead to unpredictable changes in behavior. There needs to be an obvious or proven benefit to the change-- via integration tests or benchmarks or similar. Otherwise, a small, generic improvement to a tool description might not be worth the cost.

Because langchain-community houses hundreds of integrations, it has some drawbacks that work against us here:
- We cannot version it properly (e.g., with semver)
- It is impractical to run an integration test suite.

It looks like your original motivation was to improve the Pubmed tool. That tool might be best implemented as a `StructuredTool` with distinct parameters that can be populated by modern tool-calling models. See [this guide](https://python.langchain.com/docs/how_to/custom_tools/) for implementing custom tools.

To overcome the limitations of langchain-community, we are recommending that new integrations be implemented via separate packages. If you are interested in making this improvement via a separate package (e.g., `langchain-pubmed`), we've written up a guide walking through the process:

Docs would continue to live in the langchain repo, and the library would be featured in our [integrations page](https://python.langchain.com/docs/integrations/providers/).

PR #30675, opened 2025-04-04
core: Specify code when ignoring type issue (ruff PGH003)

cbornet:
See https://docs.astral.sh/ruff/rules/blanket-type-ignore/


eyurtsev (approved):
OMG thank you!

eyurtsev (approved):

PR #30681, opened 2025-04-04
community: Add recursive sitemap support to GitbookLoader with concurrent processing

### LINKED ISSUE #30629: Bug: GitbookLoader fails to process nested sitemaps
### Checked other resources

### Example Code

The following code returns 0.

```python
from langchain_community.document_loaders import GitbookLoader

loader = GitbookLoader(
    web_page="https://platform-docs.opentargets.org/",
    load_all_paths=True,
    sitemap_url="https://platform-docs.opentargets.org/sitemap.xml",
)
docs = loader.load()
print(len(docs))  # Returns 0 instead of expected documents
```
Expected Behavior
The loader should process the sitemap index file, follow links to child sitemaps, and extract URLs to content pages, returning all available documents.
Actual Behavior
The loader processes only the top-level sitemap file, fails to recognize it as a sitemap index, and returns 0 documents.
Root Cause
The code doesn't distinguish between sitemap index files (<sitemapindex>) and regular sitemap files (<urlset>)
It isn't using the proper XML parser for sitemap files
There is no recursive processing for nested sitemaps

code above returns 0.

### Description

Trying to load the gitbook document. The GitbookLoader fails to extract documents when the target site uses nested sitemaps (a sitemap index pointing to child sitemaps). When encountering a sitemap index file, the loader attempts to process it as a regular content page rather than recursively exploring the referenced sitemaps, resulting in zero documents returned.

PS: this is a continuation of issue [30473](https://github.com/langchain-ai/langchain-community/issues/297). Prior PR solved for the example but not for the recursive sitemap issue.

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  langchain-ai/langchain#59-Ubuntu SMP PREEMPT_DYNAMIC Sat Mar 15 17:40:59 UTC 2025
> Python Version:  3.11.11 (main, Dec 11 2024, 16:28:39) [GCC 11.2.0]

Package Information
-------------------
> langchain_core: 0.3.49
> langchain: 0.3.22
> langchain_community: 0.3.20
> langsmith: 0.3.22
> langchain_anthropic: 0.3.10
> langchain_openai: 0.3.11
> langchain_text_splitters: 0.3.7

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp<4.0.0,>=3.8.3: Installed. No version info available.
> anthropic<1,>=0.49.0: Installed. No version info available.
> async-timeout<5.0.0,>=4.0.0;: Installed. No version info available.
> dataclasses-json<0.7,>=0.5.7: Installed. No version info available.
> httpx: 0.28.1
> httpx-sse<1.0.0,>=0.4.0: Installed. No version info available.
> jsonpatch<2.0,>=1.33: Installed. No version info available.
> langchain-anthropic;: Installed. No version info available.
> langchain-aws;: Installed. No version info available.
> langchain-azure-ai;: Installed. No version info available.
> langchain-cohere;: Installed. No version info available.
> langchain-community;: Installed. No version info available.
> langchain-core<1.0.0,>=0.3.45: Installed. No version info available.
> langchain-core<1.0.0,>=0.3.49: Installed. No version info available.
> langchain-deepseek;: Installed. No version info available.
> langchain-fireworks;: Installed. No version info available.
> langchain-google-genai;: Installed. No version info available.
> langchain-google-vertexai;: Installed. No version info available.
> langchain-groq;: Installed. No version info available.
> langchain-huggingface;: Installed. No version info available.
> langchain-mistralai;: Installed. No version info available.
> langchain-ollama;: Installed. No version info available.
> langchain-openai;: Installed. No version info available.
> langchain-text-splitters<1.0.0,>=0.3.7: Installed. No version info available.
> langchain-together;: Installed. No version info available.
> langchain-xai;: Installed. No version info available.
> langchain<1.0.0,>=0.3.21: Installed. No version info available.
> langsmith-pyo3: Installed. No version info available.
> langsmith<0.4,>=0.1.125: Installed. No version info available.
> langsmith<0.4,>=0.1.17: Installed. No version info available.
> numpy<3,>=1.26.2: Installed. No version info available.
> openai-agents: Installed. No version info available.
> openai<2.0.0,>=1.68.2: Installed. No version info available.
> opentelemetry-api: Installed. No version info available.
> opentelemetry-exporter-otlp-proto-http: Installed. No version info available.
> opentelemetry-sdk: Installed. No version info available.
> orjson: 3.10.16
> packaging: 24.2
> packaging<25,>=23.2: Installed. No version info available.
> pydantic: 2.11.1
> pydantic-settings<3.0.0,>=2.4.0: Installed. No version info available.
> pydantic<3.0.0,>=2.5.2;: Installed. No version info available.
> pydantic<3.0.0,>=2.7.4: Installed. No version info available.
> pydantic<3.0.0,>=2.7.4;: Installed. No version info available.
> pytest: 8.3.5
> PyYAML>=5.3: Installed. No version info available.
> requests: 2.32.3
> requests-toolbelt: 1.0.0
> requests<3,>=2: Installed. No version info available.
> rich: Installed. No version info available.
> SQLAlchemy<3,>=1.4: Installed. No version info available.
> tenacity!=8.4.0,<10,>=8.1.0: Installed. No version info available.
> tenacity!=8.4.0,<10.0.0,>=8.1.0: Installed. No version info available.
> tiktoken<1,>=0.7: Installed. No version info available.
> typing-extensions>=4.7: Installed. No version info available.
> zstandard: 0.23.0

andrasfe:
## Description:
Enhanced `GitbookLoader` to support recursive sitemap structures and asynchronous processing. The loader now recursively processes sitemap index files, following links to child sitemaps, and extracts all URLs to content pages. Also added async processing.

## Issue:
Fixes #30629 - GitbookLoader fails to process nested sitemaps

## Dependencies:
None added

### DIFF
--- a/libs/community/langchain_community/document_loaders/gitbook.py
+++ b/libs/community/langchain_community/document_loaders/gitbook.py
@@ -1,7 +1,8 @@
-import asyncio
+import warnings
 from typing import Any, AsyncIterator, Iterator, List, Optional, Set
 from urllib.parse import urljoin, urlparse
 
+from bs4 import BeautifulSoup
 from langchain_core.documents import Document
 
 from langchain_community.document_loaders.web_base import WebBaseLoader
@@ -11,9 +12,7 @@ class GitbookLoader(WebBaseLoader):
     """Load `GitBook` data.
 
     1. load from either a single page, or
-    2. load all (relative) paths in the navbar.
-
-    Supports both synchronous and asynchronous loading.
+    2. load all (relative) paths in the sitemap, handling nested sitemap indexes.
     """
 
     def __init__(
@@ -24,7 +23,6 @@ class GitbookLoader(WebBaseLoader):
         content_selector: str = "main",
         continue_on_failure: bool = False,
         show_progress: bool = True,
-        max_concurrent_requests: int = 10,
         *,
         sitemap_url: Optional[str] = None,
     ):
@@ -44,8 +42,6 @@ class GitbookLoader(WebBaseLoader):
                 exception. Setting this to True makes the loader more robust, but also
                 may result in missing data. Default: False
             show_progress: whether to show a progress bar while loading. Default: True
-            max_concurrent_requests: Maximum number of concurrent requests when
-                using async methods. Default: 10
             sitemap_url: Custom sitemap URL to use when load_all_paths is True.
                 Defaults to "{base_url}/sitemap.xml".
         """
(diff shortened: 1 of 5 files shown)


libs/community/langchain_community/document_loaders/gitbook.py
```diff
@@ -11,6 +12,8 @@ class GitbookLoader(WebBaseLoader):
 
     1. load from either a single page, or
     2. load all (relative) paths in the navbar.
+
```
eyurtsev, line 15: @andrasfe The solution should be leaning into composition rather than inheritance. We shouldn't be overriding a lot of the implementation.

eyurtsev:
Could we speed up web loader instead of making changes in gitbook loader?

libs/community/langchain_community/document_loaders/gitbook.py
andrasfe: Ah, my bad. Completely overlooked the WebBaseLoader functionality. I did the necessary changes, albeit sitemap processing requires custom code and hence the base class methods have been used where applicable only. Also, noticed an issue with the base class (WebBaseLoader) aload() method calls asyncio.run(self.fetch_all(urls)) inside its implementation. So alazy_load() is the one implemented in GitBookLoader.

andrasfe:
> Could we speed up web loader instead of making changes in gitbook loader?

You right - any optimizations should be performed in WebBaseLoader. Nevertheless, this is a more involved task as it has lots of dependencies, so perhaps this could be another ticket.

andrasfe:
@eyurtsev , can you please review when you got a chance? Thank you!

eyurtsev:
@andrasfe I'll review this tomorrow more carefully. 

But if there are no changes here that optimize specifically for gitbookloader, then either changes will need to be made in the parent abstraction or the implementation needs to be refactored before we can merge the code.

eyurtsev:
For context, could you provide a quick explanation of what this does that was not do-able with the web based loader in terms of concurrent processing?

andrasfe:
> For context, could you provide a quick explanation of what this does that was not do-able with the web based loader in terms of concurrent processing?

The main issue I addressed, as outlined in the ticket I opened, was that the original implementation could not handle hierarchical sitemaps—clearly a bug. I introduced the optimization without realizing that WebBaseLoader already supports asynchronous loading. While there’s  room for further optimization in this class, as you pointed out, that’s not the primary focus of this ticket or fix. Thank you!!

libs/community/pyproject.toml
```diff
@@ -56,7 +56,7 @@ test = [
     "toml>=0.10.2",
 ]
 codespell = ["codespell<3.0.0,>=2.2.0"]
-test_integration = ["pytest-vcr<2.0.0,>=1.0.2", "vcrpy<7,>=6"]
+test_integration = ["pytest-vcr<2.0.0,>=1.0.2", "vcrpy<7,>=6", "lxml"]
```
eyurtsev, line 59: You can remove `lxml` from here 

Instead use in the pytest marker in the unit tests

```
@pytest.mark.requires("lxml")
```

libs/community/langchain_community/document_loaders/gitbook.py
```diff
@@ -62,23 +64,153 @@ def __init__(
...
+                    warnings.warn(
+                        f"Skipping already processed sitemap URL: {sitemap_url}"
+                    )
+                    continue
+
+                processed_urls.add(sitemap_url)
+                try:
+                    # We need to temporarily set the web_paths to the sitemap URL
+                    original_web_paths = self.web_paths
+                    self.web_paths = [sitemap_url]
```
eyurtsev, line 104: The mutation in place makes this code not thread safe?

libs/community/tests/unit_tests/document_loaders/test_gitbook.py
```diff
@@ -90,115 +67,309 @@ def test_lazy_load_with_custom_sitemap(
...
-@patch("langchain_community.document_loaders.gitbook.GitbookLoader.scrape")
-def test_get_paths_extraction(
-    mock_scrape: MagicMock, mock_soups: Tuple[BeautifulSoup, BeautifulSoup]
-) -> None:
-    # Test that _get_paths correctly extracts paths from sitemap
-    mock_sitemap_soup, _ = mock_soups
-    mock_scrape.return_value = mock_sitemap_soup
+    sitemap_changelog_pages_content = """
+    <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+        <url><loc>https://example.com/changelog/update1</loc></url>
```
eyurtsev, line 111: What happens if the URL points to https://localhost:8000/foo?

libs/community/langchain_community/document_loaders/gitbook.py
```diff
@@ -10,7 +12,7 @@ class GitbookLoader(WebBaseLoader):
     """Load `GitBook` data.
 
     1. load from either a single page, or
-    2. load all (relative) paths in the navbar.
+    2. load all (relative) paths in the sitemap, handling nested sitemap indexes.
```
eyurtsev, line 15: The implementation seems too complex to me.

Why does GitbookLoader not inherit from DocumentLoader?

It can implement lazy_load and alazy_load and use WebBaseLoader as it needs to fetch content?

eyurtsev (commented):
Is it possible to make 3 changes to this:

1) Stop inherit from WebBaseLoader and re-use WebBaseLoader to simplify implementation?
2) Make sure that this does not introduce a CVE unless there's something already built in that can filter urls that are not in allowed domains.
3) remove changes to pyproject.toml and uv.lock and replace with pytest.marker.requires

libs/community/pyproject.toml
```diff
@@ -7,8 +7,8 @@ authors = []
 license = { text = "MIT" }
 requires-python = "<4.0,>=3.9"
 dependencies = [
-    "langchain-core<1.0.0,>=0.3.45",
-    "langchain<1.0.0,>=0.3.21",
+    "langchain-core<1.0.0,>=0.3.51",
+    "langchain<1.0.0,>=0.3.23",
```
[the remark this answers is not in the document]
andrasfe: Done. while I made changes to pyproject.toml and uv.sync earlier, I reverted them to match upstream in my latest commit.

libs/community/pyproject.toml
andrasfe: Done.

libs/community/tests/unit_tests/document_loaders/test_gitbook.py
andrasfe: Added unit test to validate change that addresses the CVE concerns: added an optional parameter, `allowed_domains` that whitelists accepted domain. This required in turn a modification to work with absolute url-s as opposed to relative ones. Added a unit test to validate the functionality.

libs/community/langchain_community/document_loaders/gitbook.py
andrasfe: addressed this by creating a private method that creates a new WebLoader, `_create_web_loader()` per thread - hence reusing the same instance iteratively. I feel this is a cleaner approach than using thread synchronization.

libs/community/pyproject.toml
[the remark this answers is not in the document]
andrasfe: 1. Implementation now extends BalseLoader and implements lazy_lad and alazy_load in a thread-safe manner. Code is simpler, indeed.
2. As mentioned above, added optional init argument/member allowed_domains that ignores with warning any non-whitelisted domain. This required change to work strictly with absolute paths, given the possibility for existence of sitemaps pointing to external domains. If this argument is None, default behavior is to accept only documents from the domain of the constructor-passed web_page.
3. mad those changes to revert the 2 files and remove dependency on lxml. I updated these files from the latest in master origin, hence the changes are not mine.

andrasfe:
All changes implemented, as requested. thank you!

libs/community/pyproject.toml
andrasfe: Remove dependency for unit tests and added it for integration.

andrasfe:
Hi @eyurtsev , when you got a chance.... thank you :)

libs/community/langchain_community/document_loaders/gitbook.py
```diff
@@ -42,43 +46,304 @@ def __init__(
...
+
+            if not new_urls:
+                return []
+
+            # Update the web_paths of the loader to fetch all sitemaps at once
+            original_web_paths = web_loader.web_paths
+            web_loader.web_paths = new_urls
+
+            # Use the same WebBaseLoader's ascrape_all for efficient parallel fetching
+            soups = await web_loader.ascrape_all(new_urls, parser="xml")
```
eyurtsev, line 238: This is likely using the built-in `xml` library in python which is not a safe default.

https://docs.python.org/3/library/xml.html

can we specify `lxml` by default?
`lxml`

libs/community/tests/unit_tests/document_loaders/test_gitbook.py
```diff
@@ -1,11 +1,15 @@
-from typing import Any, Tuple
+from typing import Any, List, Tuple
 from unittest.mock import MagicMock, patch
 
 import pytest
 from bs4 import BeautifulSoup
 
 from langchain_community.document_loaders.gitbook import GitbookLoader
 
+# These tests mock the XML parsing functionality to avoid requiring lxml.
```
eyurtsev, line 9: `pytest.mark.requires(["lxml"])` allows testing with LXML. 

from community root folder `git grep requires` for usage examples

eyurtsev (commented):
@andrasfe thank you for the updated implementation and apologies for the delay I just came back from vacation yesterday.

The implementation looks good overall, i flagged a few places that still need to be adjusted for security purposes.

We're in the process of moving community to a separate repository:

Would you be able to address the issue and re-open the PR there?

libs/community/langchain_community/document_loaders/gitbook.py
```diff
@@ -42,43 +46,304 @@ def __init__(
...
+        Args:
+            url: The URL to check
+
+        Returns:
+            bool: True if the domain is allowed or if no allowed_domains set is defined
+        """
+        if self.allowed_domains is None:
+            return True
+
+        netloc = urlparse(url).netloc
```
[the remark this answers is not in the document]
andrasfe: existing implementation uses  `netloc = urlparse(url).netloc`, which extracts the full network location (like www.something.x or www.something.xy). If allowed_domains = {"www.something.x"}, the check "www.something.xy" in {"www.something.x"} will correctly return False. Added fix to validate the scheme, which was indeed not there. Also updated unit tests to validate functionality.

libs/community/langchain_community/document_loaders/gitbook.py
```diff
@@ -42,43 +46,304 @@ def __init__(
...
+
+            # Update the web_paths of the loader to fetch all sitemaps at once
+            original_web_paths = web_loader.web_paths
+            web_loader.web_paths = new_urls
+
+            # Use the same WebBaseLoader's ascrape_all for efficient parallel fetching
+            soups = await web_loader.ascrape_all(new_urls, parser="xml")
+
+            # Restore original web_paths
+            web_loader.web_paths = original_web_paths
```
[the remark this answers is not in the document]
andrasfe: the suggestion to avoid in-place mutation by creating a web loader would indeed make the code cleaner and less likely for future changes to impact its functionality, it would, however, introduce a slight performance overhead, especially for some sites that have hundreds or even thousand nested pages. I left it this way for now but if you feel we should change it, will be happy to do so.

libs/community/langchain_community/document_loaders/gitbook.py
andrasfe: fixed.

libs/community/tests/unit_tests/document_loaders/test_gitbook.py
andrasfe: got it and fixed it.

andrasfe:
Thank you @eyurtsev and no problem at all. I saw you had no activity during this time frame - know you were off. I completed the changes ad provided my 2 cents for some, but I'm open to further suggestions. I will open a PR against the new repo once I get confirmation from you that we are good to go.

ccurme:
Closing for now as we've already moved the library.

andrasfe:
re-created PR in the new repo: https://github.com/langchain-ai/langchain-community/pull/13

PR #30683, opened 2025-04-05
community: support usage_metadata for litellm streaming calls

ikbenale:
Support "usage_metadata" for LiteLLM streaming calls.

This is a follow-up to https://github.com/langchain-ai/langchain/pull/30625, which tackled non-streaming calls.

### DIFF
--- a/libs/community/langchain_community/chat_models/litellm.py
+++ b/libs/community/langchain_community/chat_models/litellm.py
@@ -452,6 +452,7 @@ class ChatLiteLLM(BaseChatModel):
         params = {**params, **kwargs, "stream": True}
 
         default_chunk_class = AIMessageChunk
+        added_model_name = False
         for chunk in self.completion_with_retry(
             messages=message_dicts, run_manager=run_manager, **params
         ):
@@ -460,7 +461,15 @@ class ChatLiteLLM(BaseChatModel):
             if len(chunk["choices"]) == 0:
                 continue
             delta = chunk["choices"][0]["delta"]
+            usage = chunk.get("usage", {})
             chunk = _convert_delta_to_message_chunk(delta, default_chunk_class)
+            if isinstance(chunk, AIMessageChunk):
+                if not added_model_name:
+                    chunk.response_metadata = {
+                        "model_name": self.model_name or self.model
+                    }
+                    added_model_name = True
+                chunk.usage_metadata = _create_usage_metadata(usage)
             default_chunk_class = chunk.__class__
             cg_chunk = ChatGenerationChunk(message=chunk)
             if run_manager:
@@ -478,6 +487,7 @@ class ChatLiteLLM(BaseChatModel):
         params = {**params, **kwargs, "stream": True}
 
         default_chunk_class = AIMessageChunk
+        added_model_name = False
         async for chunk in await acompletion_with_retry(
             self, messages=message_dicts, run_manager=run_manager, **params
         ):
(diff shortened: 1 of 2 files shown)


ikbenale:
@ccurme would you mind reviewing this PR?

libs/community/langchain_community/chat_models/litellm.py
```diff
@@ -460,7 +460,11 @@ def _stream(
             if len(chunk["choices"]) == 0:
                 continue
             delta = chunk["choices"][0]["delta"]
+            usage = chunk.get("usage", {})
             chunk = _convert_delta_to_message_chunk(delta, default_chunk_class)
+            if isinstance(chunk, AIMessageChunk):
+                chunk.response_metadata = {"model_name": self.model_name or self.model}
+                chunk.usage_metadata = _create_usage_metadata(usage)
```
ccurme, line 467: If you inspect an aggregated response (e.g., in the [test](https://github.com/langchain-ai/langchain/blob/afd457d8e177020898319da588c3a042e001c129/libs/standard-tests/langchain_tests/integration_tests/chat_models.py#L938) we are enabling) you will see the problem with this:
> AIMessageChunk(content="Whispers of the leaves,  \nAutumn's breath in gold and red,  \nNature's soft embrace.  \n\nMoonlight on still pond,  \nRipples dance in silver nights,  \nDreams reflected clear.", additional_kwargs={}, response_metadata={'model_name': 'gpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-minigpt-4o-mini'}, id='run-1308cc2a-ef6f-4e32-9b50-b4e19dddac7f', usage_metadata={'input_tokens': 20, 'output_tokens': 44, 'total_tokens': 64})

1. Model name should be populated only once during the stream
2. If we are checking for token usage in every chunk, we should verify or test that only one chunk contains usage metadata (or that each chunk returns incremental token usage).

For model name, if that's returned during the stream, we can take it off the response. Otherwise it's simple enough to do something like this: https://github.com/langchain-ai/langchain-aws/pull/440/files

ccurme (changes requested):

libs/community/langchain_community/chat_models/litellm.py
ikbenale: > Model name should be populated only once during the stream

Good catch. Fixed by implementing a solution similar to https://github.com/langchain-ai/langchain-aws/pull/440/files. I pushed a new version of the commit.

> If we are checking for token usage in every chunk, we should verify or test that only one chunk contains usage metadata (or that each chunk returns incremental token usage).

I see that is already part of the integration test: https://github.com/langchain-ai/langchain/blob/9cfe6bcacd5abc774496f22561297a572d022a36/libs/standard-tests/langchain_tests/integration_tests/chat_models.py#L986-L995. Would that be sufficient? 

> For model name, if that's returned during the stream, we can take it off the response.

I do see a `model` field in the stream chunk, but pre-existing code gets the `model` from `self.model_name` or `self.model`, see https://github.com/langchain-ai/langchain/blob/884125e129c88b04f4cfa4d85e9ba1fc2fba949e/libs/community/langchain_community/chat_models/litellm.py#L427-L430, so I was trying to be consistent with that. What do you think?

ikbenale:
@ccurme addressed your feedback, let me know of any additional changes you'd like.

ccurme (approved):

libs/community/langchain_community/chat_models/litellm.py
ccurme: Looks good, thanks. Agreed on consistency with regard to using `self.model_name`.

PR #30684, opened 2025-04-05
Feat: add oauth2 support for Jira toolkit

pol-defont-reaulx:
**Description:** add support for oauth2 in Jira tool by adding the possibility to pass a dictionary with oauth parameters. I also adapted the documentation to show this new behavior

### DIFF
--- a/libs/community/langchain_community/utilities/jira.py
+++ b/libs/community/langchain_community/utilities/jira.py
@@ -35,8 +35,8 @@ class JiraAPIWrapper(BaseModel):
     - with OAuth2, you need to provide the JIRA_OAUTH2 environment variable or
         argument as a dict having as fields "client_id" and "token" which is
         a dict containing at least "access_token" and "token_type".
-    ex: JIRA_OAUTH2='{"client_id": "your_client_id", "token": {"access_token": "your_access_token",
-        "token_type": "bearer"}}'
+    ex: JIRA_OAUTH2='{"client_id": "your_client_id", "token":
+        {"access_token": "your_access_token","token_type": "bearer"}}'
     """
 
     jira: Any = None  #: :meta private:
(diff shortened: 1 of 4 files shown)


libs/community/langchain_community/utilities/jira.py
```diff
@@ -14,6 +14,7 @@ class JiraAPIWrapper(BaseModel):
     confluence: Any = None
     jira_username: Optional[str] = None
     jira_api_token: Optional[str] = None
+    jira_oauth_dict: Optional[Union[Dict[str, Union[str, Dict[str, str]]], str]] = None
```
eyurtsev, line 17: 1. `dict` isn't a good thing to put in this variable name. Perhaps `jira_oauth2` to indicate it's just proxying the `oauth2` variable?
2. Could you document inline what values are accepted? And explicit type annotation for the dict using a TypedDict (`typing_extensions import TypedDict`) would help!
3. Now that there's more than one option to authenticate, it would be good to document this as part of the wrapper -- you could add an authentication example in the JiraAPIWrapper and inline a doc-string for both variables

eyurtsev (commented):
Looks good! happy to merge after:

1. update variable name
2. adding documentation

libs/community/langchain_community/utilities/jira.py
pol-defont-reaulx: Is it what you had in mind?

libs/community/langchain_community/utilities/jira.py
```diff
@@ -1,19 +1,51 @@
...
+class JiraOauth2Token(TypedDict):
+    """Jira OAuth2 token."""
+
+    access_token: str
+    """Jira OAuth2 access token."""
+    token_type: str
+    """Jira OAuth2 token type ('bearer' or other)."""
+
+
+class JiraOauth2(TypedDict):
```
eyurtsev, line 19: typing_extensions.TypedDict might be needed w/ pydantic and python < 3.12

eyurtsev (approved):

eyurtsev (approved):

PR #30696, opened 2025-04-06
core: Add ruff rules PLR

cbornet:
Add ruff rules [PLR](https://docs.astral.sh/ruff/rules/#refactor-plr)
Except PLR09xxx and PLR2004.


libs/core/langchain_core/callbacks/file.py
```diff
@@ -53,11 +53,10 @@ def on_chain_start(
         """
         if "name" in kwargs:
             name = kwargs["name"]
+        elif serialized:
```
eyurtsev, line 56: this is one of the few refactors i never liked: personally I find this easier to read while being more difficult to understand

eyurtsev (approved):

libs/core/langchain_core/callbacks/file.py
cbornet: We could exclude this rule ?

PR #30737, opened 2025-04-09
core: Improve mypy config

cbornet:
* Cleanup mypy config
* Add mypy `strict` rules except `disallow_any_generics`, `warn_return_any` and `strict_equality` (TODO)
* Add mypy `strict_byte` rule
* Add mypy support for PEP702 `@deprecated` decorator
* Bump mypy version to 1.15


libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -738,20 +736,17 @@ def test_tool_outputs() -> None:
     assert not response.tool_calls
 
 
-@pytest.mark.parametrize("use_extension_typed_dict", [True, False])
-@pytest.mark.parametrize("use_extension_annotated", [True, False])
+@pytest.mark.parametrize("typed_dict", [ExtensionsTypedDict, TypingTypedDict])
```
eyurtsev, line 739: (not important) -- any chance you know whether this will render nicely when looking at pytest output?

eyurtsev (approved):
Looks great! we can merge

libs/core/tests/unit_tests/utils/test_function_calling.py
cbornet: Not very nicely, no.
I’ll add tags to make it pretty.

PR #30757, opened 2025-04-09
partners/openai: OpenAIEmbeddings not respecting chunk_size argument

aubford:
When calling `embed_documents` and providing a `chunk_size` argument, that argument is ignored when `OpenAIEmbeddings` is instantiated with its default configuration (where `check_embedding_ctx_length=True`).

`_get_len_safe_embeddings` specifies a `chunk_size` parameter but it's not being passed through in `embed_documents`, which is its only caller. This appears to be an oversight, especially given that the `_get_len_safe_embeddings` docstring states it should respect "the set embedding context length and chunk size."

Developers typically expect method parameters to take effect (also, take precedence) when explicitly provided, especially when instantiating using defaults. I was confused as to why my API calls were being rejected regardless of the chunk size I provided.

This bug also exists in langchain_community package.  I can add that to this PR if requested otherwise I will create a new one once this passes.


aubford:
Hi @ccurme could you take a look at this when you get a chance?

aubford:
Hi @ccurme @eyurtsev.  Did this fall through the cracks or is there an issue with my PR I need to address?

ccurme (approved):
Thank you!

PR #30778, opened 2025-04-10
fix(ollama): Remove redundant message from response_metadata

jacoblee93:
Fixes trace deserialization


eyurtsev (approved):
any chance there's an issue that can be referenced?

PR #30792, opened 2025-04-11
Galaxia integration

rrozanski-smabbler:
- [ ] **PR title**: "docs: adding Smabbler's Galaxia integration"

- [ ] **PR message**:  **Twitter handle:** @Galaxia_graph

I'm adding docs here + added the package to the packages.yml. I didn't add a unit test, because this integration is just a thin wrapper on top of our API. There isn't much left to test if you mock it away.


ccurme (commented):
@rrozanski-smabbler could you make sure all headings required in the [template](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/retrievers.ipynb) are represented here?

They should be at the appropriate level (e.g., `## Setup` instead of `# Setup`) and include all headers in the template.

We can link to the Github repo instead of the API reference if you aren't building API reference docs yet.

rrozanski-smabbler:
@ccurme thanks for the feedback!

I have updated the .ipynb file and renamed both files to `galaxia-retriever...`, because this seems to be the expected name. The only section missing is the "API reference" - should I add the github repo link there as well while keeping this section name? This could be confusing.

I see there is this error in the workflow:
> [ERROR] Error: Unable to build website for locale en.
> (...)
>   [cause]: Error: Docusaurus found broken links!

is this something I can fix on my end somehow?

ccurme:
Hi @rrozanski-smabbler, please revert the change in file name for the provider page (i.e., `galaxia-retriever.mdx` -> `galaxia.mdx`). I had pushed an update to fix this in https://github.com/langchain-ai/langchain/pull/30792/commits/6239b342b5982bcc9fd32d31679d0922a096d231. You can also revert that commit but my sense is `Galaxia` is a better name for the provider page.

Regarding the API ref: let's keep the section name for now to keep the templates uniform. I'll revisit enforcing that in the template.

rrozanski-smabbler:
OK, thanks:
- I've changed the provider's doc name
- and added the API section to integration doc, with a github link

ccurme (approved):

PR #30803, opened 2025-04-13
langchain-openai: support custom fallback content for tool-only messages

louisgthier:
Some APIs (notably Gemini, when accessed via the OpenAI-compatible endpoint) require **every message** to have a non-null, non-empty string as `content`. However, messages that contain only tool calls (e.g., `ToolMessage`) often have `content=None`, leading to a `400 Bad Request` with:

```bash
Error code: 400 - [{‘error’: {‘code’: 400, ‘message’: ‘Expected string or list of content parts, got: null’, ‘status’: ‘INVALID_ARGUMENT’}}]
```

This PR adds support for specifying a fallback string (e.g. `" "` or `"N/A"`) to populate the content field of tool-only messages, ensuring compatibility with stricter API implementations.

**Issue:**  
N/A — discovered while integrating LangChain with Gemini API.

**Dependencies:**  
None

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -10,7 +10,6 @@ import re
 import ssl
 import sys
 import warnings
-from collections.abc import AsyncIterator, Iterator, Mapping, Sequence
 from functools import partial
 from io import BytesIO
 from json import JSONDecodeError
@@ -227,7 +226,7 @@ def _format_message_content(content: Any) -> Any:
     return formatted_content
 
 
-def _convert_message_to_dict(message: BaseMessage) -> dict:
+def _convert_message_to_dict(message: BaseMessage, content_null_value: Optional[str] = None) -> dict:
     """Convert a LangChain message to a dictionary.
 
     Args:
@@ -267,7 +266,7 @@ def _convert_message_to_dict(message: BaseMessage) -> dict:
             pass
         # If tool calls present, content null value should be None not empty string.
         if "function_call" in message_dict or "tool_calls" in message_dict:
-            message_dict["content"] = message_dict["content"] or None
+            message_dict["content"] = message_dict["content"] or content_null_value
 
         if "audio" in message.additional_kwargs:
             # openai doesn't support passing the data back - only the id
@@ -542,6 +541,9 @@ class BaseChatOpenAI(BaseChatModel):
 
     .. versionadded:: 0.3.9
     """
+    
+    content_null_value: Optional[str] = None
+    """The value to use for content when it is null. Should be set to some non-empty string for Gemini API for example."""
 
     model_config = ConfigDict(populate_by_name=True)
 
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Tool calling is supported for Gemini models via the `langchain-google-genai` and `langchain-google-vertexai` packages, depending on how you are authenticating. See docs [here](https://python.langchain.com/docs/integrations/providers/google/). Will these work for your use-case?

Many providers offer a chat-completions API such that you can use the OpenAI SDK to interact with them. But they often feature small differences like this. If the dedicated Google packages will work for you, that's preferable to attempting to support more providers via ChatOpenAI.

louisgthier:
I gave it a try, and the issue seems to be present in both `langchain-google-genai` and `langchain-openai`. Tool calling generally works fine, except in cases where the LLM returns only tool calls without any content—that’s when it breaks.

Since I frequently switch between models, and most providers support the OpenAI SDK, I prefer sticking with that—it works out of the box and avoids these edge cases. I get that keeping things simple is important, and adding parameters everywhere isn’t ideal.

Maybe when the provider is detected to be Google, we could just make sure the content field is never left empty?

It's not a huge issue anyway—I ran into it using the built-in LangGraph React Agent, and I worked around it by adjusting the messages using a "prompt" pre-hook.

ccurme:
Thanks @louisgthier. Are you able to provide a minimal reproducible example? This would help us figure out the best solution.

louisgthier:
> Thanks @louisgthier. Are you able to provide a minimal reproducible example? This would help us figure out the best solution.

```python
from langchain_core.messages import HumanMessage, SystemMessage
from langgraph.checkpoint.memory import MemorySaver
from typing import List, Any
import os

# 1. Define a simple tool
def add_numbers(a: int, b: int) -> int:
    """Add two numbers and return the result."""
    return a + b
add_numbers = tool(add_numbers)

# 2. Set up the Gemini model
model_name = "gemini-2.0-flash"
api_key = os.getenv("GEMINI_API_KEY")

llm = ChatGoogleGenerativeAI(
    model=model_name,
    google_api_key=api_key,
)
llm = llm.bind_tools([add_numbers])

# 3. Define a minimal state class
class State(dict):
    messages: List[Any]
    remaining_steps: int

# 4. Create the agent
graph = create_react_agent(
    llm,
    [add_numbers],
    state_schema=State,
    checkpointer=MemorySaver()
)

def main():
    # 5. Prepare the initial state
    system = SystemMessage(content="You are a helpful assistant. Use tools if needed.")
    user = HumanMessage(content="Please use the add_numbers tool to add 2 and 3.")
    state = State(messages=[system, user], remaining_steps=5)  # 5 is arbitrary, just needs to be present

    # 6. Run the agent
    result = graph.invoke(state, config={"configurable": {"thread_id": "test-thread"}})
    print("Final messages:")
    for msg in result['messages']:
        print(f"{type(msg).__name__}: {getattr(msg, 'content', msg)}")

if __name__ == "__main__":
    main()
```

When running this simple script, I get:

```
langchain_google_genai.chat_models.ChatGoogleGenerativeAIError: Invalid argument provided to Gemini: 400 * GenerateContentRequest.contents: contents is not specified
```

louisgthier:
Please note I have the same result with ChatOpenAI

ccurme:
Closing as author is unresponsive, but please ping me if we should re-open.

PR #30808, opened 2025-04-13
feat(langchain): improve performance `split_list_of_docs()`

ianchi:
Currently `split_list_of_docs` has duration of O(n2), this PR adds an option to optimize it to O(n)

For example in a list of 500 very small docs (imagine comments in posts) that actually fits the token length,  it calls 125250 times (instead of just 500) docs in the length function.
This makes that something that should take less than a fraction of second ends up taking more than a minute.

To make it non-breaking, this PR leaves the default behaviour backwards compatible, but adds a parameter that allows the optimization, assuming that the length function is accumulative over the docs (which I belive is the more common case).


ianchi:
Hi, any comments about this?

libs/langchain/langchain/chains/combine_documents/reduce.py
```diff
@@ -43,9 +58,15 @@ def split_list_of_docs(
     """
     new_result_doc_list = []
     _sub_result_docs = []
+    _num_tokens = length_func([], **kwargs) if acum_length else 0
```
eyurtsev, line 61: why does it not start from `0`

eyurtsev (changes requested):
1) I don't think this code is used much, so don't know how important it is for optimization.
2) I'm fine w/ the assumption that the length works independently on docs by default. I suggest removing the acum_length variable, and simplifying the internal logic

libs/langchain/langchain/chains/combine_documents/reduce.py
ianchi: To give the oportunity of a custom init value. It is subtle, but you could use it to remove the length of the initial sepparator.

-sep + (sep + doc1) + (sep + doc2) + .... + (sep + docn)

the `length_func` could account for the document tokens plus some constant separator, and remove the initial separator for the empty array

ianchi:
Hi @eyurtsev, the code is used in some example documentation, and it is where I took it from, and with it came a terrible performance degradation.

I made it optional with `acum_length=False` to keep the exact same old behaviour.
With point 2, you refer to change it to always behave in the accumulative fashion, no?

libs/langchain/langchain/chains/combine_documents/reduce.py
eyurtsev: While it's mathematically correct, it's not that useful in practice and introduces unnecessary complexity.

This code should err on being approximately correct, but fast to run and simple to understand.

libs/langchain/langchain/chains/combine_documents/reduce.py
```diff
@@ -27,14 +27,29 @@ async def __call__(self, docs: list[Document], **kwargs: Any) -> str:
 
 
 def split_list_of_docs(
-    docs: list[Document], length_func: Callable, token_max: int, **kwargs: Any
+    docs: list[Document],
+    length_func: Callable,
+    token_max: int,
+    acum_length: bool = False,
```
eyurtsev, line 33: Let's swap the implementation to be more efficient and remove complexity from the interface

```suggestion
```

mdrxy:
@ianchi awaiting changes

eyurtsev:
Closing due to no changes from the author

PR #30809, opened 2025-04-13
partners: ChatAnthropic supports urls

Konsti-s:
**Description:**
partners-anthropic: ChatAnthropic supports b64 and urls in the part[image_url][url] message variable

**Issue**:
ChatAnthropic right now only supports b64 encoded images in the part[image_url][url] message variable. This PR enables ChatAnthropic to also accept image urls in said variable and makes it compatible with OpenAI messages to make model switching easier.


ccurme (approved):
Thanks for this. This week we released a standard format for multimodal content blocks and extended Anthropic's support to URLs in the process. See docs [here](https://python.langchain.com/docs/how_to/multimodal_inputs/#images-from-a-url). But this is nice to have!

PR #30810, opened 2025-04-13
community: have ChatLlamaCpp handle "auto" and "any" for tool_choice

clinty:
Description: This adds "auto" and "any" to ChatLlamaCpp's tool_choice options, and handles them both by selecting the first tool in the list
Issue: trustcall assumes support for "any" at https://github.com/AIXerum/trustcall/blob/main/trustcall/_base.py#L717


libs/community/langchain_community/chat_models/llamacpp.py
```diff
@@ -362,6 +363,8 @@ def bind_tools(
                         f"Tool choice {tool_choice=} was specified, but the only "
                         f"provided tools were {tool_names}."
                     )
+            elif isinstance(tool_choice, str) and tool_choice in ['any', 'auto']:
```
ccurme, line 366: Don't `"auto"` and `"any"` have different meanings? Doesn't `"auto"` typically mean the model can elect not to call a tool? See for example [OpenAI](https://platform.openai.com/docs/guides/function-calling?api-mode=chat#tool-choice) and [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#forcing-tool-use) docs.

libs/community/langchain_community/chat_models/llamacpp.py
```diff
@@ -362,6 +363,8 @@ def bind_tools(
...
+                tool_choice = formatted_tools[0]
```
ccurme, line 367: IMO, if `tools` is length one, then we should translate `"any"` as you do here, but if there are multiple tools, we should raise an error.

clinty:
@ccurme what about this?  I suppose it does the wrong thing if there's a tool called "auto".

ccurme (commented):
I think `"auto"` should have the same behavior as `None` for tool choice, right? Should we just assign None in that case?

For `"any"`, my thought is to do something like this:
```python
if tool_choice == "any":
    if len(formatted_tools) == 1:
        tool_choice = formatted_tools[0]
    else:
        raise ValueError(
            "tool_choice `'any'` only supported if one tool is provided."
        )
```

This change will need some tests as well.

ccurme (commented):
This looks good, can you add a test [here](https://github.com/langchain-ai/langchain/blob/master/libs/community/tests/integration_tests/chat_models/test_llamacpp.py)?

ccurme (commented):
@clinty we've now moved `langchain-community` into a standalone repo. Would you mind opening your PR there with a test? Feel free to tag me.

Issue #30764, opened 2025-04-10
Anthropic does not accept only System Messages


mathislindner:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

from langchain_core.messages import SystemMessage, HumanMessage, AIMessage
from langchain.chat_models import init_chat_model

LLM = init_chat_model("claude-3-5-haiku-20241022", model_provider = "anthropic")

prompt = SystemMessage(test")

print(LLM.invoke([prompt]))

### Error Message and Stack Trace (if applicable)

Traceback (most recent call last):
  File "test_script.py", line 15, in <module>
    LLM.invoke([prompt])
  File ".../site-packages/langchain_core/language_models/chat_models.py", line 331, in invoke
    self.generate_prompt(
  File ".../site-packages/langchain_core/language_models/chat_models.py", line 894, in generate_prompt
    return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/langchain_core/language_models/chat_models.py", line 719, in generate
    self._generate_with_cache(
  File ".../site-packages/langchain_core/language_models/chat_models.py", line 960, in _generate_with_cache
    result = self._generate(
             ^^^^^^^^^^^^^^^
  File ".../site-packages/langchain_anthropic/chat_models.py", line 1102, in _generate
    data = self._client.messages.create(**payload)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/anthropic/_utils/_utils.py", line 275, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../site-packages/anthropic/resources/messages/messages.py", line 953, in create
    return self._post(
           ^^^^^^^^^^^
  File ".../site-packages/anthropic/_base_client.py", line 1336, in post
[...]

PR #30822, opened 2025-04-14
anthropic: emit informative error message if there are only system messages in a prompt

mathislindner:
**PR message**: Not sure if I put the check at the right spot, but I thought throwing the error before the loop made sense to me.
**Description:** Checks if there are only system messages using AnthropicChat model and throws an error if it's the case. Check Issue for more details
**Issue:** #30764


ccurme (commented):
Instead of blocking this ourselves, could we catch the error from Anthropic and add whatever clarifications we want to the error message? This way if Anthropic adds support we will not be blocking.

Here's an example: https://github.com/langchain-ai/langchain/blob/3a64c7195faa90538fbb3aee019dba6d66dab81a/libs/partners/openai/langchain_openai/chat_models/base.py#L383

Will need to handle for all invocation modes (sync/async invoke and stream)

mathislindner:
okay, I just thought it would be wiser to do this before sending a call to Anthropic that would get us a bad request, but yes it does makes sense to expect them to change it down the line :)

mathislindner:
```
from langchain_core.messages import SystemMessage, HumanMessage, AIMessage
import asyncio

LLM = init_chat_model("claude-3-5-haiku-20241022", model_provider = "anthropic")
prompt = SystemMessage("test")


#test _generate
def test_generate():
    try:
        _ = LLM._generate([prompt])
    except ValueError as e:
        assert str(e) == "Received only system message(s). "
        
def test_stream():
    try:
        stream = LLM._stream([prompt])
        for _ in stream:
            pass
    except ValueError as e:
        assert str(e) == "Received only system message(s). "

async def test_agenerate():
    try:
        _ = await LLM._agenerate([prompt])
    except ValueError as e:
        assert str(e) == "Received only system message(s). "

async def test_astream():
    try:
        stream = LLM._astream([prompt])
        async for _ in stream:
            pass
    except ValueError as e:
        assert str(e) == "Received only system message(s). "
        
if __name__ == "__main__":
    test_generate()
    test_stream()
    asyncio.run(test_agenerate())
    asyncio.run(test_astream())
```

throws the errors, wasn't sure if you wanted tests but this works now

clementsicard (approved):

ccurme (approved):

PR #30841, opened 2025-04-15
docs: Register langchain-singlestore integration

volodymyr-memsql:
I created and published `langchain-singlestoe` integration package that should replace SingleStoreDB community implementation.


volodymyr-memsql:
@ccurme, could you please help me with landing this one? And also #30846, which should come after the current one.

ccurme (approved):

PR #30846, opened 2025-04-15
community: Add deprecation decorator to SingleStore community integrations

volodymyr-memsql:
SingleStore integration now has its package `langchain-singlestore', so the community implementation will no longer be maintained.

Added `deprecated` decorator to `SingleStoreDBChatMessageHistory`, `SingleStoreDBSemanticCache`, and `SingleStoreDB` classes in the community package.

**Dependencies:** https://github.com/langchain-ai/langchain/pull/30841


ccurme (approved):
Thanks. I think it will be confusing to have two SingleStore integrations in the docs:
Screenshot 2025-04-18 at 12 30 21 PM

I've deleted the docs for the community version and added redirects to the LangChain v0.2 versions. Let me know if you disagree.

ccurme (approved):

PR #30847, opened 2025-04-15
[community] Propose PDFRouterParser and Loader

pprados:
# Load PDFs using different parsers based on the metadata of the PDF or the body of the first page.

The routes are defined as a list of tuples, where each tuple contains the name, a dictionary of metadata and regex pattern and the parser to use. The special key "page1" is to search in the first page with a regexp. Use the route in the correct order, as the first matching route is used.a Add a default route ("default", {}, parser) at the end to catch all PDFs. This code is similar to `MimeTypeBasedParser`, but on the content of the PDF file.

Sample:
```python
    from langchain_community.document_loaders import PyPDFLoader
    from langchain_community.document_loaders.parsers.pdf import PyMuPDFParser
    from langchain_community.document_loaders.parsers.pdf import PyPDFium2Parser
    from langchain_community.document_loaders.parsers import PDFPlumberParser
    routes = [
        ("Microsoft", {"producer": "Microsoft", "creator": "Microsoft"},
        PyMuPDFParser()),
        ("LibreOffice", {"producer": "LibreOffice", }, PDFPlumberParser()),
        ("Xdvipdfmx", {"producer": "xdvipdfmx.*", "page1":"Hello"}, PDFPlumberParser()),
        ("defautl", {}, PyPDFium2Parser())
    ]
    loader = PDFRouterLoader(filename, routes)
    loader.load()
```

### DIFF
--- a/libs/community/extended_testing_deps.txt
+++ b/libs/community/extended_testing_deps.txt
@@ -59,7 +59,7 @@ openapi-pydantic>=0.3.2,<0.4
 oracle-ads>=2.9.1,<3
 oracledb>=2.2.0,<3
 pandas>=2.0.1,<3
-pdfminer-six==20231228
+pdfminer-six==20250327
 pdfplumber>=0.11
 pgvector>=0.1.6,<0.2
 playwright>=1.48.0,<2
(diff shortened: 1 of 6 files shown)


ccurme (commented):

Issue #30851, opened 2025-04-15
`_get_response` from `base.py` does not return generated file id when running as an agent


georgi-stefanov:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```python

from langchain_community.agents.openai_assistant import OpenAIAssistantV2Runnable

openai_assistant = OpenAIAssistantV2Runnable(
            assistant_id="your_assistant_id", as_agent=True
        )

output = openai_assistant.invoke(
                {
                    "content": "Generate a .txt file containing a poem",
                    "thread_id": "your_thread_id",
                }
            )

print(output)

```

in libs/langchain/langchain/agents/openai_assistant/base.py _get_response() is called by invoke

```python

    def _get_response(self, run: Any) -> Any:
        # TODO: Pagination

        if run.status == "completed":
            import openai

            major_version = int(openai.version.VERSION.split(".")[0])
            minor_version = int(openai.version.VERSION.split(".")[1])
            version_gte_1_14 = (major_version > 1) or (
                major_version == 1 and minor_version >= 14
            )

            messages = self.client.beta.threads.messages.list(
                run.thread_id, order="asc"
            )
            new_messages = [msg for msg in messages if msg.run_id == run.id]
            if not self.as_agent:
                return new_messages
            answer: Any = [
                msg_content for msg in new_messages for msg_content in msg.content
            ]
            if all(
                (
                    isinstance(content, openai.types.beta.threads.TextContentBlock)
                    if version_gte_1_14
                    else isinstance(
                        content, openai.types.beta.threads.MessageContentText
                    )
                )
                for content in answer
            ):
                answer = "\n".join(content.text.value for content in answer)
            return OpenAIAssistantFinish(
                return_values={
                    "output": answer,
                    "thread_id": run.thread_id,
                    "run_id": run.id,
                },
                log="",
                run_id=run.id,
                thread_id=run.thread_id,
            )
        elif run.status == "requires_action":
            if not self.as_agent:
                return run.required_action.submit_tool_outputs.tool_calls
            actions = []
            for tool_call in run.required_action.submit_tool_outputs.tool_calls:
[...]

PR #30853, opened 2025-04-15
langchain: return attachments in _get_response

georgi-stefanov:
This is a PR to return the message attachments in _get_response, as when files are generated these attachments are not returned thus generated files cannot be retrieved

Fixes issue: https://github.com/langchain-ai/langchain/issues/30851


libs/langchain/langchain/agents/openai_assistant/base.py
```diff
@@ -584,6 +584,9 @@ def _get_response(self, run: Any) -> Any:
             answer: Any = [
                 msg_content for msg in new_messages for msg_content in msg.content
             ]
+            attachments = [
+                attachment for msg in new_messages for attachment in msg.attachments
```
ccurme, line 588: does `msg` always have an attachments attribute, even when there are no attachments?

ccurme (approved):

PR #30861, opened 2025-04-15
langchain: infer Perplexity provider for sonar model prefix

milosz-l:
**Description:** This PR adds provider inference logic to `init_chat_model` for Perplexity models that use the "sonar..." prefix (`sonar`, `sonar-pro`, `sonar-reasoning`, `sonar-reasoning-pro` or `sonar-deep-research`).

This allows users to initialize these models by simply passing the model name, without needing to explicitly set `model_provider="perplexity"`.

The docstring for `init_chat_model` has also been updated to reflect this new inference rule.


ccurme (approved):
Thanks!

PR #30869, opened 2025-04-16
langchain[patch]: Allow async indexing code to work for vectorstores that only defined sync delete

shkarupa-alex:
`aindex` function should check not only `adelete` method, but `delete` method too

**PR title**: "core: fix async indexing issue with adelete/delete checking"
**PR message**: Currently `langchain.indexes.aindex` checks if vector store has overrided adelete method. But due to `adelete` default implementation store can have just `delete` overrided to make `adelete` working.


shkarupa-alex:
Failed test doesn't apply to the modified code

shkarupa-alex:
Can someone check this PR please?

eyurtsev:
Thank you @shkarupa-alex

PR #30872, opened 2025-04-16
langchain_openai: clean duplicate code for openai embedding.

zhanluxianshen:
The `_chunk_size` has not changed by method `self._tokenize`, So i think these is duplicate code.


ccurme (approved):

zhanluxianshen:
thanks.

PR #30888, opened 2025-04-16
langchain-openai: Fix openapi _resize case.

zhanluxianshen:
If ```width``` or ```height``` set =2049 , the case logic is wrong.
It will downgrade to `768` size..

I think it is wrong code for that condition.

### DIFF
--- a/libs/partners/openai/tests/unit_tests/chat_models/test_base.py
+++ b/libs/partners/openai/tests/unit_tests/chat_models/test_base.py
@@ -52,6 +52,7 @@ from langchain_openai.chat_models.base import (
     _create_usage_metadata,
     _format_message_content,
     _oai_structured_outputs_parser,
+    _resize,
 )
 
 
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Could you add a unit test that fails on `master`, but passes on this branch? Thanks.

zhanluxianshen:
> Could you add a unit test that fails on `master`, but passes on this branch? Thanks.
Yes , how about add a test like that?

libs/partners/openai/tests/unit_tests/chat_models/test_base.py
```diff
@@ -1022,6 +1023,44 @@ def test_structured_outputs_parser() -> None:
     assert result == parsed_response
 
 
+def test_image_resize() -> None:
+    ret1 = _resize(2049, 2049)
+    assert ret1[0] == 2048
+    assert ret1[1] == 2048
```
ccurme, line 1029: All of these tests pass on master, except for this condition, which enforces that (2049, 2049) -> (2048, 2048). Currently, this gets resized to (768, 768).

However, that is what OpenAI's pricing calculator says should happen:

Screenshot 2025-04-27 at 3 23 08 PM

See "How is pricing calculated for images" under FAQ [here](https://openai.com/api/pricing/).

zhanluxianshen: ah, thanks for the explain..

PR #30912, opened 2025-04-17
core: Fix some private member accesses

cbornet:
See https://github.com/langchain-ai/langchain/pull/30666

### DIFF
--- a/libs/core/langchain_core/tracers/langchain.py
+++ b/libs/core/langchain_core/tracers/langchain.py
@@ -51,7 +51,7 @@ def log_error_once(method: str, exception: Exception) -> None:
 def wait_for_all_tracers() -> None:
     """Wait for all tracers to finish."""
     if rt._CLIENT is not None:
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/tracers/langchain.py
```diff
@@ -51,7 +51,7 @@ def log_error_once(method: str, exception: Exception) -> None:
 def wait_for_all_tracers() -> None:
     """Wait for all tracers to finish."""
     if rt._CLIENT is not None:
-        rt._CLIENT.flush()
+        get_client().flush()
```
eyurtsev, line 54: Let's use a ruff ignore here -- it's already accessing `rt._CLIENT`

eyurtsev, line 54:
```suggestion
        rt._CLIENT.flush()
```

eyurtsev (approved):

PR #30965, opened 2025-04-22
fireworks[minor]: remove default model and temperature

ccurme:
`mixtral-8x-7b-instruct` was recently retired from Fireworks Serverless.

Here we remove the default model altogether, so that the model must be explicitly specified on init:
```python
ChatFireworks(model="accounts/fireworks/models/llama-v3p1-70b-instruct")  # for example
```

We also set a null default for `temperature`, which previously defaulted to 0.0. This parameter will no longer be included in request payloads unless it is explicitly provided.

### DIFF
--- a/libs/partners/fireworks/tests/integration_tests/test_chat_models.py
+++ b/libs/partners/fireworks/tests/integration_tests/test_chat_models.py
@@ -13,54 +13,13 @@ from typing_extensions import TypedDict
 
 from langchain_fireworks import ChatFireworks
 
-
-def test_chat_fireworks_call() -> None:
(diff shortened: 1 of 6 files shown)


libs/partners/fireworks/tests/integration_tests/test_chat_models.py
```diff
@@ -13,54 +13,13 @@
 
 from langchain_fireworks import ChatFireworks
 
-
-def test_chat_fireworks_call() -> None:
```
ccurme, line 17: Deleted some tests that are redundant with standard tests.

Issue #26701, opened 2024-09-20
Anthropic's prompt caching in langchain does not work with ChatPromptTemplate.


raajChit:
### URL

https://python.langchain.com/docs/how_to/llm_caching/

### Checklist

- [X] I added a very descriptive title to this issue.
- [X] I included a link to the documentation page I am referring to (if applicable).

### Issue with current documentation:

I have not found any documentation for prompt caching in the langchain documentation. There seems to be only one post on twitter regarding prompt caching in langchain. I am trying to implement prompt caching in my rag system. I am using history aware retriever. 

I have instantiated the model like this:

llm_claude = ChatAnthropic(
    model="claude-3-5-sonnet-20240620",
    temperature=0.1,
    extra_headers={"anthropic-beta": "prompt-caching-2024-07-31"}
)

And using the ChatPromptTemplate like this:

contextualize_q_prompt = ChatPromptTemplate.from_messages(
                [
                    ("system", contextualize_q_system_prompt),
                    ("human", "{input}"),
                ]
            )
            
I am not able to find a way to include prompt caching with this. 
I tried making the prompt like this, but still doesnt work.

prompt = ChatPromptTemplate.from_messages([
        SystemMessage(content=contextualize_q_system_prompt, additional_kwargs={"cache_control": {"type": "ephemeral"}}),
        HumanMessage(content= "{input}")
    ])

Please help me with how I should enable prompt caching in langchain.

### Idea or request for content:

Langchain documentation should be updated with how to use prompt caching with different prompt templates. And especially with a RAG system.

drorm:
This is a major issue. Without caching it's just too slow and too expensive. I tried:
- Adding additional_kwargs={"cache_control": {"type": "ephemeral"}
- Adding {"cache_control": {"type": "ephemeral"} to the message
and caching doesn't happen.

The only thing that does get cached is the system prompt.

kornicameister:
@raajChit  @drorm @eyurtsev ; I'm struggling with this one too. I have 80k token prompt that has some placeholders there. The moment I use `SystemMessage`, I get caching to work but I'm loosing the tool filling in the placeholders. If I do replace that with `SystemMessagePromptTemplate`, placeholders receive values but caching is not working.

It is not possible by design to cache that because it is not static prompt??

DonghaeSuh:
I've posted a discussion on #29747 to enable the feature mentioned in this issue.

Nilesh3105:
[...]

PR #30967, opened 2025-04-22
core[patch]: update dict prompt template

baskaryan:
Align with JS changes made in https://github.com/langchain-ai/langchainjs/pull/8043


libs/core/langchain_core/prompts/dict.py
```diff
@@ -0,0 +1,137 @@
...
+from langchain_core.load import dumpd
+from langchain_core.prompts.string import (
+    DEFAULT_FORMATTER_MAPPING,
+    get_template_variables,
+)
+from langchain_core.runnables import RunnableConfig, RunnableSerializable
+from langchain_core.runnables.config import ensure_config
+
+
+class DictPromptTemplate(RunnableSerializable[dict, dict]):
```
baskaryan, line 16: note: no longer inherits from BaseMessagePromptTemplate since thats not really how its used

libs/community/tests/unit_tests/load/test_serializable.py
baskaryan, line 1: afaict this was just never moved to core and isn't needed anymore?

ccurme (approved):
This should close https://github.com/langchain-ai/langchain/issues/26701

PR #30968, opened 2025-04-22
core, openai: support standard multi-modal blocks in convert_to_openai_messages

ccurme:


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -61,7 +61,7 @@
     ToolCall,
     ToolMessage,
     ToolMessageChunk,
-    convert_to_openai_image_block,
+    convert_to_openai_data_block,
```
jacoblee93, line 64: Won't this require a core version bump?

ccurme: Yes, openai release will be blocked by core release + bumping min version.

Issue #30970, opened 2025-04-22
`Union[Objects]`-typed args raising invalid_function_parameters error in strict mode of OpenAI function calling / structured output


shengbo-ma:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

Example of Failure
1. Pydantic Schema output using function calling method.
```py
class NestedA(BaseModel):
    foo: str

class NestedB(BaseModel):
    bar: int

class NestedC(BaseModel):
    baz: bool

class MySchema(BaseModel):
    my_arg: Union[NestedA, NestedB, NestedC]


prompt = "..."
llm=ChatOpenAI(...)
chain = prompt | llm.with_structured_output(
    MySchema, strict=True, method="function_calling"
)
chain.invoke("...")

```

### Error Message and Stack Trace (if applicable)

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid schema for function 'MySchema': In context=('properties', 'my_arg', 'items', 'anyOf', '0'), 'additionalProperties' is required to be supplied and to be false.", 'type': 'invalid_request_error', 'param': 'tools[0].function.parameters', 'code': 'invalid_function_parameters'}}
```

### Description

[OpenAI Function Calling Strict Mode support `anyOf`](https://platform.openai.com/docs/guides/structured-outputs?api-mode=responses#supported-types), which is `Union` in python typing.
However, it fails in LangChain when there are args of type union of objects. 

### System Info

argcomplete==3.3.0
click==8.1.7
pipx==1.5.0
platformdirs==4.2.1
userpath==1.9.2
uv==0.6.16

shengbo-ma:
I sent and PR to fix this: https://github.com/langchain-ai/langchain/pull/30971

PR #30971, opened 2025-04-22
core: support `Union` type args in strict mode of OpenAI function calling / structured output

shengbo-ma:
**Issue:**[ #309070](https://github.com/langchain-ai/langchain/issues/30970)

**Cause**
Arg type in python code
```
arg: Union[SubSchema1, SubSchema2]
``` 
is translated to `anyOf` in **json schema**
```
"anyOf" : [{sub schema 1 ...}, {sub schema 1 ...}]
```
The value of anyOf is a list sub schemas. 
The bug is caused since the sub schemas inside `anyOf` list is not taken care of.
The location where the issue happens is `convert_to_openai_function` function -> `_recursive_set_additional_properties_false` function, that recursively adds `"additionalProperties": false` to json schema which is [required by OpenAI's strict function calling](https://platform.openai.com/docs/guides/structured-outputs?api-mode=responses#additionalproperties-false-must-always-be-set-in-objects)

**Solution:**
This PR fixes this issue by iterating each sub schema inside `anyOf` list.
A unit test is added.

**Twitter handle:** shengboma 


shengbo-ma:
@eyurtsev Would you mind review this PR? Should be a straight forward fix.

eyurtsev (approved):

eyurtsev (approved):
Thank you!

PR #30980, opened 2025-04-23
docs: replace deprecated initialize_agent with create_react_agent in NASA toolkit example

Jisolp:
**Description:**
This PR updates the NASA toolkit agent example in the documentation by replacing the deprecated `initialize_agent` with the recommended `create_react_agent` method from `langgraph.prebuilt`. 

**Issue:**
Part of #29277

**Dependencies:**
None

### DIFF
--- a/docs/docs/integrations/tools/nasa.ipynb
+++ b/docs/docs/integrations/tools/nasa.ipynb
@@ -39,7 +39,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from langchain.agents import AgentType, initialize_agent\n",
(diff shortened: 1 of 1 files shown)


ccurme (commented):
I suspect this code is introducing errors:

- The [legacy](https://python.langchain.com/docs/concepts/text_llms/) `OpenAI` LLM does not support `bind_tools`, which is required for `create_react_agent`.
- Order of args (tools, llm) in `create_react_agent` is incorrect.
- Input format (a dict with key `"input"`) is incorrect.

PR #30984, opened 2025-04-23
feat(core): Autogenerate filenames for when converting file content blocks to OpenAI format

jacoblee93:
CC @ccurme


libs/core/langchain_core/messages/utils.py
```diff
@@ -1236,6 +1237,32 @@ def convert_to_openai_messages(
...
+                    if block.get("source_type") != "base64":
+                        err = (
+                            f"Unrecognized content block at "
+                            f"messages[{i}].content[{j}] has 'type': "
+                            f"'file', but has a 'source_type' of {block.get('source_type')}."
+                            f"Only 'base64' is supported for 'source_type'."
+                            f"Full content block:\n\n{block}"
+                        )
+                        raise ValueError(err)
+                    filename = block.get("filename", _generate_default_filename(block.get("mime_type")))
```
jacoblee93, line 1250: I think this is ok here because you are explicitly calling out that you want this converted to OpenAI format?

ccurme (approved):

PR #30985, opened 2025-04-23
rfc: support image generation in ChatOpenAI via client.images

ccurme:
OpenAI released an image generation model `gpt-image-1` today. See [docs](https://platform.openai.com/docs/guides/image-generation?image-generation-model=gpt-image-1).

It features two methods:
- `client.images.generate` generates an image from a string prompt
- `client.images.edit` generates an image from a string prompt and a list of images

Here we hack this into a chat model interface, so that you can edit images in a conversational setting.

Example:
```python

llm = init_chat_model("openai:gpt-image-1")

message_1 = "Make an image of a giraffe."
response_1 = llm.invoke([message_1])

# response_1 is:
# AIMessage(
#     content=[
#         {
#             "type": "image",
#             "source_type": "base64",
#             "data": "<base64 image data>",
#             "mime_type": "image/png",
#         },
#     ],
#     usage_metadata={"input_tokens": 14, "output_tokens": 6240, ...},
# )

message_2 = "Invert the colors in the image."
response_2 = llm.invoke([message_1, response_1, message_2])

# response_2 formatted similarly
```

Can inspect images locally with
```python
import base64

with open("/path/to/image.png", "wb") as f:
    f.write(base64.b64decode(response_1.content[0]["data"]))
```

Pros: maybe nice to support in a standard chat model interface. Easy to incorporate into a chatbot (e.g., using LangGraph persistence or similar).

Cons: abuse of `client.images`, will likely run into nonsensical usage patterns.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -32,6 +32,7 @@ from urllib.parse import urlparse
 import certifi
 import openai
 import tiktoken
+from langchain_core._api.beta_decorator import warn_beta
 from langchain_core._api.deprecation import deprecated
 from langchain_core.callbacks import (
     AsyncCallbackManagerForLLMRun,
@@ -416,6 +417,7 @@ def _handle_openai_bad_request(e: openai.BadRequestError) -> None:
 
 
 _MessageContent = Union[str, list[Union[str, dict]]]
+WARNED_IMAGE_GEN_BETA = False
 
 
 def _get_image_bytes_from_content(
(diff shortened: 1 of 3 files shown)


ccurme:
Closing for now, will prioritize support through Responses API when that's available.

eric-burel:
@ccurme isn't chat model support still relevant in the context of generating images through a conversation, eg when "gpt-image-1" is called indirectly via "gpt-4o"? Or would it work the same with the Response API? In my (beginner) understanding, Response API is tailored for programmatic use cases, but chat is still useful for a semi-automated workflow.

PR #30987, opened 2025-04-24
community: add indexname to other functions in opensearch

bhemati:
- [x] **PR title**: "community: add indexname to other functions in opensearch"

- [x] **PR message**:
    - **Description:**  add ability to over-ride index-name if provided in the kwargs of sub-functions. When used in WSGI application it's crucial to be able to dynamically change parameters.

- [ ] **Add tests and docs**:


libs/community/langchain_community/vectorstores/opensearch_vector_search.py
```diff
@@ -480,6 +480,8 @@ def __add(
         bulk_size = bulk_size if bulk_size is not None else self.bulk_size
         _validate_embeddings_and_bulk_size(len(embeddings), bulk_size)
         index_name = kwargs.get("index_name", self.index_name)
+        if self.index_name is None:
+            raise ValueError("index_name must be provided.")
```
ccurme, line 484: `self.index_name` is typed as str. Does it make sense to raise ValueError on `__init__` instead of distributing them into individual methods like this?

ccurme (approved):

libs/community/langchain_community/vectorstores/opensearch_vector_search.py
bhemati: At first, I considered making it optional in __init__ since it doesn't really make sense to have to provide an index_name on connection level, but unfortunately, since it's the second parameter, that was impossible.
But raise ValueError in __init__ is also a valid point. I will keep it in mind for the next PR.

ccurme: Does raising ValueError in `__add` and `__aadd` when index_name is null help you solve that problem?

bhemati: I put the check everywhere before I realized it's impossible to make it optional in __main__. Thinking about it it's not necessary anymore sice it's impossible for index_name to not be set to a string before we reach there, my bad.

Issue #30879, opened 2025-04-16
`langchain-chroma`: Query filters are restrictive


hesreallyhim:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

chromadb and langchain-chroma allow for querying with metadata filtering. In most instances in the langchain-chroma codebase, the type of `filter` is `dict[str, str]`. This is extremely restricted compared to the actual filter (= "where") syntax in the chroma query methods, which allow queries like:
```python
{
  "$and": [{"page": {"$gt": 5}}, {"keyword": {"$in": ["foo", "bar"]}}]
}
```
The langchain-chroma types only allow for the most trivial query with one string comparison on one key. However, because these are just hints, the regular chromadb filter syntax still works, so there is just a big mis-match between the type hints and the implementation.

I would like to rectify this, but I would like to know from a maintainer or chroma contributor or whomever - is the "filter" syntax intended to be so restrictive for some reason? Is there any reason to not go ahead and just fix the types so that query methods will continue to work as normal but you won't get type warnings if you try anything beyond {"string": "string"}?

I already posted about this problem, but haven't received a reply, and would like to submit a fix, I'll close down one of these issues but would love if someone could weigh in. Otherwise I will go ahead and assume that the types are just incorrect, and submit a fix for the type hints, which will not affect the functionality anyway and will make the function type much more clear and accurate.

Thanks!

P.S. I don't think the `SelfQueryRetriever` is the answer.

### Error Message and Stack Trace (if applicable)

mypy error/pylance warning on valid filter clauses due to overly restrictive type hints.

### Description

see above

### System Info

can replicate in any chroma notebook on langchain doc site

mdrxy:
Looks to be addressed? If not, let me know and I can reopen

Issue #30507, opened 2025-03-27
ChromaDB/Docs: `similarity_search_*` filter type hints are incorrect and API docs are incorrect


hesreallyhim:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_chroma import Chroma
from langchain_openai import OpenAIEmbeddings
from langchain_core.documents import Document

chroma = Chroma(embedding_function=OpenAIEmbeddings())
docs = [
    Document(page_content="Hello world", metadata={"author": "john", "topic": "chroma"}),
    Document(page_content="Hello world 2", metadata={"author": "jack", "topic": "chroma"}),
]
chroma.add_documents(docs)
# THIS FILTER DOES NOT RAISE TYPE WARNING: `{"a": "b", "c": "d"}` is shorthand for `{"$and": [{"a": {"$eq": "b"}}, {"c": {"$eq": d"}}]}`
results1 = chroma.similarity_search(
    "Hello world",
    k=4,
    filter={"author": "john", "topic": "chroma"},
)
# THIS FILTER RAISES TYPE WARNING:
results2 = chroma.similarity_search(
    "Hello world",
    k=4,
    # THIS CONDITION IS FROM [CHROMA DOCS](https://github.com/chroma-core/chroma/blob/main/examples/basic_functionality/where_filtering.ipynb)
    filter={"$and": [{"category": "chroma"}, {"$or": [{"author": "john"}, {"author": "jack"}]}]}
)
```

### Error Message and Stack Trace (if applicable)

Pylance error/warning:

```console
Argument of type "dict[str, list[dict[str, str] | dict[str, list[dict[str, str]]]]]" cannot be assigned to parameter "filter" of type "Dict[str, str] | None" in function "similarity_search"
  Type "dict[str, list[dict[str, str] | dict[str, list[dict[str, str]]]]]" is not assignable to type "Dict[str, str] | None"
    "dict[str, list[dict[str, str] | dict[str, list[dict[str, str]]]]]" is not assignable to "Dict[str, str]"
      Type parameter "_VT@dict" is invariant, but "list[dict[str, str] | dict[str, list[dict[str, str]]]]" is not the same as "str"
[...]

PR #31058, opened 2025-04-28
DOCS: `partners/chroma`: Fix documentation around `chroma` query filter syntax

hesreallyhim:

  - Where "package" is whichever of langchain, community, core, etc. is being modified. Use "docs: ..." for purely docs changes, "infra: ..." for CI changes.

**Description**:
* Starting to put together some PR's to fix the typing around `langchain-chroma` `filter` and `where_document` query filtering, as mentioned:

https://github.com/langchain-ai/langchain/issues/30879
https://github.com/langchain-ai/langchain/issues/30507

The typing of `dict[str, str]` is on the one hand too restrictive (marks valid filter expressions as ill-typed) and also too permissive (allows illegal filter expressions). That's not what this PR addresses though. This PR just removes from the documentation some examples of filters that are illegal, and also syntactically incorrect: (a) dictionaries with keys like `$contains` but the key is missing quotation marks; (b) dictionaries with multiple entries - this is illegal in Chroma filter syntax and will raise an exception. (`{"foo": "bar", "qux": "baz"}`). Filter dictionaries in Chroma must have one and one key only. Again this is just the documentation issue, which is the lowest hanging fruit. I also think we need to update the types for `filter` and `where_document` to be (at the very least `dict[str, Any]`), or, since we have access to Chroma's types, they should be `Where` and `WhereDocument` types. This has a wider blast radius though, so I'm starting small.

This PR does not fix the issues mentioned above, it's just starting to get the ball rolling, and cleaning up the documentation.

Additional guidelines:

### DIFF
--- a/.github/scripts/check_diff.py
+++ b/.github/scripts/check_diff.py
@@ -38,8 +38,8 @@ IGNORED_PARTNERS = [
 ]
 
 PY_312_MAX_PACKAGES = [
-    "libs/partners/huggingface",  # https://github.com/pytorch/pytorch/issues/130249
     "libs/partners/voyageai",
+    "libs/partners/chroma", # https://github.com/chroma-core/chroma/issues/4382
 ]
 
 
--- a/docs/docs/integrations/tools/bash.ipynb
+++ b/docs/docs/integrations/tools/bash.ipynb
@@ -26,7 +26,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "id": "f7b3767b",
    "metadata": {
     "tags": []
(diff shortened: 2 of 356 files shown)


ccurme (approved):

hesreallyhim:
Thanks @ccurme .... I was going to follow up on this by changing the types for the `chroma` `filter` from `dict[str, str]` to the actual types, which are available through `chromadb` (`Where` and `WhereDocument`) because `dict[str, str]` is really not accurate.... what I found though, is that my IDE still couldn't validate whether a `filter` expression was correct or not, because `Where` is a nested/recursive type. One alternative is to change the type to `dict[str, Any]`, since it seems to me like it's better to be overly permissive than overly restrictive in this case, but at the end of the day I think it's not possible to give them the right type without other strategies of validation

PR #31080, opened 2025-04-30
core, openai[patch]: prefer provider-assigned IDs when aggregating message chunks

ccurme:
When aggregating AIMessageChunks in a stream, core prefers the leftmost non-null ID. This is problematic because:
- Core assigns IDs when they are null to `f"run-{run_manager.run_id}"`
- The desired meaningful ID might not be available until midway through the stream, as is the case for the OpenAI Responses API.

For the OpenAI Responses API, we assign message IDs to the top-level `AIMessage.id`. This works in `.(a)invoke`, but during `.(a)stream` the IDs get overwritten by the defaults assigned in langchain-core. These IDs [must](https://community.openai.com/t/how-to-solve-badrequesterror-400-item-rs-of-type-reasoning-was-provided-without-its-required-following-item-error-in-responses-api/1151686/9) be available on the AIMessage object to support passing reasoning items back to the API (e.g., if not using OpenAI's `previous_response_id` feature). We could add them elsewhere, but seeing as we've already made the decision to store them in `.id` during `.(a)invoke`, addressing the issue in core lets us fix the problem with no interface changes.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -3095,6 +3095,52 @@ def _pop_summary_index_from_reasoning(reasoning: dict) -> dict:
     return new_reasoning
 
 
+def _explode_on_id(msg: dict[str, Any]) -> list[dict[str, Any]]:
+    """
+    Split a message dict into a list of message dicts, hoisting any `id`
+    that appears on a content block up to the outer level.
+
+    Consecutive content blocks that share the same non-None `id`
+    end up in the same outer block.
+    A content block **without** an `id` always goes into its *own* block.
+
+    Nothing else about the structure is assumed – every key except
+    `content` is copied unchanged into each resulting block.
+
+    The function avoids deepcopying large structures; only the small
+    pieces that actually change are copied.
+    """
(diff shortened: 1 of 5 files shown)


libs/partners/openai/tests/integration_tests/chat_models/test_responses_api.py
```diff
@@ -270,7 +270,7 @@ def multiply(x: int, y: int) -> int:
         """return x * y"""
         return x * y
 
-    llm = ChatOpenAI(model=MODEL_NAME)
+    llm = ChatOpenAI(model=MODEL_NAME, use_responses_api=True)
```
ccurme, line 273: unrelated fix here

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -3123,6 +3123,7 @@ def _construct_responses_api_input(messages: Sequence[BaseMessage]) -> list:
             reasoning_items = []
             if reasoning := lc_msg.additional_kwargs.get("reasoning"):
                 reasoning_items.append(_pop_summary_index_from_reasoning(reasoning))
+            input_.extend(reasoning_items)
```
ccurme, line 3126: This logic is already tested in integration tests

PR #31082, opened 2025-04-30
anthropic: Allow kwargs to pass through when counting tokens

bengladwell:
- **Description:** `ChatAnthropic.get_num_tokens_from_messages` does not currently receive `kwargs` and pass those on to `self._client.beta.messages.count_tokens`. This is a problem if you need to pass specific options to `count_tokens`, such as the `thinking` option. This PR fixes that.
- **Issue:** N/A
- **Dependencies:** None
- **Twitter handle:** @bengladwell


ccurme (approved):
Thank you!

Some other conveniences we could consider:
- pull known fields off the chat model instance (e.g., `self.thinking`)
- automatically enable thinking if there are thinking blocks in the messages (though I think we'd need to come up with a token budget somewhere)

PR #31117, opened 2025-05-05
core: enable true caching support for arbitrary message stacks in langgraph

cosminacho:
**Description**: This PR introduces improvements to enable robust caching when using LangGraph or any arbitrary stack of messages. It addresses two core challenges:

**1. Stable Message IDs:**
For effective caching, message IDs must remain consistent across runs. Previously, these IDs were often generated using UUIDs or other random methods, causing cache misses. This fix ensures IDs are deterministically set before making an LLM request—especially important when using complex message stacks involving AIMessage, ToolMessage, etc.

**2. Cache Serialization Bug Fix:**
A subtle deserialization issue in cache stores (e.g., SQLite) caused cache mismatches due to key order changes in the tool_calls field of AIMessage. This fix enforces a deterministic key order during serialization, ensuring consistent cache lookups.

Issue: N/A (internal reliability improvement)

Dependencies: None

Twitter handle: N/A

PS: This PR is intended to highlight the current caching issues and demonstrate a straightforward fix. I’m happy to continue refining it and implement a more robust solution based on your feedback and suggestions.

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -46,7 +46,7 @@ from langchain_core.language_models.base import (
     LangSmithParams,
     LanguageModelInput,
 )
-from langchain_core.load import dumpd, dumps
+from langchain_core.load import dumpd, dumps, loads
 from langchain_core.messages import (
     AIMessage,
     AnyMessage,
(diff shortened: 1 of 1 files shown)


eyurtsev (changes requested):
High level both issues make sense as something we'd like to resolve.

The implementation does not look like it's in the right place in this case.

1. we don't want to add a lodas(dumps( round trip in the caching code. There must be some other fix for key determinism
2. mutating message.id attribute in the current location does not look correct. The ID should be generated elsewhere, but preserved when written to the cache.

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -1100,6 +1102,7 @@ async def _agenerate_with_cache(
                 **result.generations[0].message.response_metadata,
             }
         if check_cache and llm_cache:
+            result.generations = loads(dumps(result.generations))
```
eyurtsev, line 1105: Wouldn't the problem be in the caching implementation? Mutating .generations here does not seem like it can be a correct solution

eyurtsev:
Closing the PR for now -- if you' like you can open an issue and a PR with unit tests that demonstrate the failures. We can then try to identify the appropriate place to resolve the problems

PR #31140, opened 2025-05-07
fix(core): Permit OpenAI style blocks to be passed into convert_to_openai_messages

jacoblee93:
Should effectively be a noop, just shouldn't throw

CC @madams0013


ccurme (approved):
Thanks!

PR #31146, opened 2025-05-07
partners: (langchain-openai) total_tokens should not add 'Nonetype' t…

zhurou603:
partners: (langchain-openai) total_tokens should not add 'Nonetype' t…

# PR Description

## Description
Fixed an issue in `langchain-openai` where `total_tokens` was incorrectly adding `None` to an integer, causing a TypeError. The fix ensures proper type checking before adding token counts.

## Issue
Fixes the TypeError traceback shown in the image where `'NoneType'` cannot be added to an integer.

## Dependencies
None

## Twitter handle
None
image

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -2924,9 +2924,9 @@ class OpenAIRefusalError(Exception):
 
 
 def _create_usage_metadata(oai_token_usage: dict) -> UsageMetadata:
-    input_tokens = oai_token_usage.get("prompt_tokens", 0)
-    output_tokens = oai_token_usage.get("completion_tokens", 0)
-    total_tokens = oai_token_usage.get("total_tokens", input_tokens + output_tokens)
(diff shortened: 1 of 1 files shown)


zhurou603:
Hope this can be merged

zhurou603:
cc @ccurme

ccurme (approved):
These are typed as `int` in the SDK, I'm curious what API you're using?

https://github.com/openai/openai-python/blob/67997a4ec1ebcdf8e740afb0d0b2e37897657bde/src/openai/types/completion_usage.py#L40

zhurou603:
> These are typed as `int` in the SDK, I'm curious what API you're using?
> 
> https://github.com/openai/openai-python/blob/67997a4ec1ebcdf8e740afb0d0b2e37897657bde/src/openai/types/completion_usage.py#L40

image

Qwen-max API sometimes will return CompletionUsage as a NoneType instead of int, because sometimes they cached same query to reduce the consumption of token. In this case, use "get" to get a value of dictionary will get a NoneType value instead of return a non-exist error.

PR #31147, opened 2025-05-07
docs: Pinecone Rerank example notebook

simjak:
Created an example notebook of how to use Pinecone Reranking service
cc @jamescalam


ccurme (commented):
I'd recommend linking out to this from `docs/docs/integrations/providers/pinecone` for discoverability.

ccurme (approved):

Issue #31051, opened 2025-04-28
tool args descriptions inference cannot handle PEP 563 stringized annotations


iyanging:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```Python
from __future__ import annotations

from typing import Annotated

from langchain_core.tools import tool


@tool
def search_internet(
    query_fragments: Annotated[
        list[str],
        "A list of query fragments",
    ],
) -> list[str]:
    """Search the Internet and retrieve relevant result items."""
    return []


print(search_internet.args)
```

### Description

If I write `from __future__ import annotations` at the head of py file, the annotation will be string during runtime.

`langchain_core.tools.base._infer_arg_descriptions()` and `_get_annotation_description()` cannot correctly handle un-evaluated annotation string.

The output of MRE is:
```
❌ {'query_fragments': {'items': {'type': 'string'}, 'title': 'Query Fragments', 'type': 'array'}}
```
rather than:
```
✅ {'query_fragments': {'description': 'A list of query fragments', 'items': {'type': 'string'}, 'title': 'Query Fragments', 'type': 'array'}}
```

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.5.0: Tue Apr 15 21:34:34 PDT 2025; root:xnu-11417.120.105.501.1~2/RELEASE_ARM64_T6000
> Python Version:  3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 16.0.0 (clang-1600.0.26.6)]

Package Information
-------------------
> langchain_core: 0.3.56
> langchain: 0.3.24
> langchain_community: 0.3.22
> langsmith: 0.3.37
> langchain_deepseek: 0.1.3
> langchain_ollama: 0.3.2
> langchain_openai: 0.3.14
> langchain_text_splitters: 0.3.8

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp<4.0.0,>=3.8.3: Installed. No version info available.
> async-timeout<5.0.0,>=4.0.0;: Installed. No version info available.
[...]

PR #31180, opened 2025-05-11
[tools]: Annotationg Type Casting Fix in _infer_arg_descriptions

keenborder786:
- **Description:** In `_infer_arg_descriptions`, the annotations dict is getting a string type rather than actual typing object which is causing a failure for _is_annotated_type and therefore the description is not coming up.
  - **Issue:** #31051


keenborder786:
@efriis @ccurme

PR #31206, opened 2025-05-13
langchain-core[patch]: `Incremental` record manager deletion should be batched

loperamos:
**Description:** Before this commit, if one record is batched in more than 32k rows for sqlite3 >= 3.32 or more than 999 rows for sqlite3 < 3.31, the `record_manager.delete_keys()` will fail, as we are creating a query with too many variables.

This commit ensures that we are batching the delete operation leveraging the `cleanup_batch_size` as it is already done for `full` cleanup.

Added unit tests for incremental mode as well on different deleting batch size.


eyurtsev (approved):

eyurtsev:
@loperamos thanks for the patch!

PR #31209, opened 2025-05-13
ollama: Add separate kwargs parameter for async client

AlexeyBond:
**Description**:

Add a `async_client_kwargs` field to ollama chat/llm/embeddings adapters that is passed to async httpx client constructor.

**Motivation:**

In my use-case:
- chat/embedding model adapters may be created frequently, sometimes to be called just once or to never be called at all
- they may be used in bots sunc and async mode (not known at the moment they are created)

So, I want to keep a static transport instance maintaining connection pool, so model adapters can be created and destroyed freely. But that doesn't work when both sync and async functions are in use as I can only pass one transport instance for both sync and async client, while transport types must be different for them. So I can't make both sync and async calls use shared transport with current model adapter interfaces.

In this PR I add a separate `async_client_kwargs` that gets passed to async client constructor, so it will be possible to pass a separate transport instance. For sake of backwards compatibility, it is merged with `client_kwargs`, so nothing changes when it is not set.

I am unable to run linter right now, but the changes look ok.

### DIFF
--- a/libs/partners/ollama/langchain_ollama/chat_models.py
+++ b/libs/partners/ollama/langchain_ollama/chat_models.py
@@ -436,14 +436,22 @@ class ChatOllama(BaseChatModel):
     """Base url the model is hosted under."""
 
     client_kwargs: Optional[dict] = {}
-    """Additional kwargs to pass to the httpx Client. 
-    For a full list of the params, see [this link](https://www.python-httpx.org/api/#client)
+    """Additional kwargs to pass to the httpx clients. 
+    These arguments are passed to both synchronous and async clients.
+    Use sync_client_kwargs and async_client_kwargs to pass different arguments
+    to synchronous and asynchronous clients.
     """
 
     async_client_kwargs: Optional[dict] = {}
     """Additional kwargs to merge with client_kwargs before
     passing to the httpx AsyncClient.
     For a full list of the params, see [this link](https://www.python-httpx.org/api/#asyncclient)
+    """
+
+    sync_client_kwargs: Optional[dict] = {}
+    """Additional kwargs to merge with client_kwargs before
+    passing to the httpx Client.
+    For a full list of the params, see [this link](https://www.python-httpx.org/api/#client)
     """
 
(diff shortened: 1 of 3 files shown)


libs/partners/ollama/langchain_ollama/chat_models.py
```diff
@@ -436,8 +436,13 @@ class Multiply(BaseModel):
     """Base url the model is hosted under."""
 
     client_kwargs: Optional[dict] = {}
-    """Additional kwargs to pass to the httpx Client.
-    For a full list of the params, see [this link](https://pydoc.dev/httpx/latest/httpx.Client.html)
```
AlexeyBond, line 440: Documentation link here seems broken, so I've changed it to a valid one.

ccurme: Thank you!

ccurme (commented):
It seems a little unintuitive that we should merge these vs. just maintaining separate (potentially redundant) sets of kwargs.

For instance, if there's a key in `client_kwargs`, it's not possible to initialize the async client without this key, right?

Not sure how realistic this is but let me know what you think.

AlexeyBond:
> It seems a little unintuitive that we should merge these vs. just maintaining separate (potentially redundant) sets of kwargs.
> 
> For instance, if there's a key in `client_kwargs`, it's not possible to initialize the async client without this key, right?
> 
> Not sure how realistic this is but let me know what you think.

I don't think it is a significant issue, at least for now - as most of the parameters (except for `transport`) of sync and async clients are the same and will have the same values. But I have added another field - `sync_client_kwargs` that will be merged to synchronous client parameters. So it will be possible to specify some parameters for sync client only.

I think `client_args` field should be kept and merged into parameters of both sync and async client. Both to keep backwards compatibility and for some convenience - as in most cases the additional parameters for sync and async clients are the same.

ccurme (approved):
Thank you!

ccurme:
Just released.

PR #31219, opened 2025-05-13
docs: update how_to docs to reflect new loaders interface and functionalities

MedlockM:
- **Description:** Updates two notebooks in the how_to documentation to reflect new loader interfaces and functionalities.
- **Issue:** Some how_to notebooks were still using loader interfaces from previous versions of LangChain and did not demonstrate the latest loader functionalities (e.g., extracting images with `ImageBlobParser`, extracting tables in specific output formats, parsing documents using Vision-Language Models with `ZeroxPDFLoader`, and using `CloudBlobLoader` in the `GenericLoader`, etc.).
- **Dependencies:** `py-zerox`
- **Twitter handle:** @MarcMedlock2


ccurme (approved):

MedlockM:
@ccurme 
Could you please justify your changes? Especially concerning the pdf guide. For example, why not keep the part explaining the new method for parsing with a vision model using Zerox? The way it's currently presented (convert pdf to an image then send the image to a chat object) is "old-fashioned", isn't it?

Likewise, the fact that you can set up an ImageBlobParser directly in a loader to extract tables or images with different strategies is a new feature that should be highlighted, don't you think?

PR #31222, opened 2025-05-13
text-splitters: fix text splitter start index

### LINKED ISSUE #29884: TokenTextSplitter start indices are sometimes -1
### Checked other resources

### Example Code

To reproduce:
```python
from langchain_text_splitters import TokenTextSplitter

text = """
"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
"""
chunker = TokenTextSplitter(add_start_index=True, chunk_size=10, chunk_overlap=5)
chunker.create_documents([text])
> the result contains examples with -1 as start-index
```

_No response_

### Description

There is a problem with start indices because of how `TokenTextSplitter` inherits from `TextSplitter`. The `create_documents` method presumes that `self._chunk_overlap` contains a number expressing the number of characters, however with a tokenizer this becomes the number of *tokens* and thus the correct piece of text might not be found:

```python
    def create_documents(
        self, texts: List[str], metadatas: Optional[List[dict]] = None
    ) -> List[Document]:
        """Create documents from a list of texts."""
        _metadatas = metadatas or [{}] * len(texts)
        documents = []
        for i, text in enumerate(texts):
            index = 0
            previous_chunk_len = 0
            for chunk in self.split_text(text):
                metadata = copy.deepcopy(_metadatas[i])
                if self._add_start_index:
                    offset = index + previous_chunk_len - self._chunk_overlap
                    index = text.find(chunk, max(0, offset))
                    metadata["start_index"] = index
                    previous_chunk_len = len(chunk)
                new_doc = Document(page_content=chunk, metadata=metadata)
                documents.append(new_doc)
        return documents
```

Possible solution could be to change the loop to something like this, it will make lookup slightly slower. It might also point to the wrong piece of text if there are lots of duplications ...

```python
    def create_documents(
        self, texts: List[str], metadatas: Optional[List[dict]] = None
    ) -> List[Document]:
        """Create documents from a list of texts."""
        _metadatas = metadatas or [{}] * len(texts)
        documents = []
        for i, text in enumerate(texts):
            index = -1  # modified
            previous_chunk_len = 0
            for chunk in self.split_text(text):
                metadata = copy.deepcopy(_metadatas[i])
                if self._add_start_index:
                    index = text.find(chunk, max(0, index + 1), min(len(text), index + 1 + previous_chunk_len + len(chunk))  # modified
                    metadata["start_index"] = index
                    previous_chunk_len = len(chunk)
                new_doc = Document(page_content=chunk, metadata=metadata)
                documents.append(new_doc)
        return documents
```

### System Info

langchain version : '0.3.19'

aowen87:
**Description:**
I've updated the base text splitter so that it correctly captures the `start_index`. Without this fix, there are many cases of `start_index` receiving -1. This was a result of the chunk overlap being in number of _tokens_, not characters. 

The search for a matching text would often begin at an index that was past the true starting point of the chunk. In the best case scenario, we would end up with -1 as a `start_index`. In the worst case scenario, we would end up with an incorrect `start_index` due to finding another instance of that chunk later in the text. 

**Issue:** 
Fixes #29884

**Dependencies:**
None


libs/text-splitters/tests/integration_tests/test_text_splitter.py
```diff
@@ -118,3 +119,83 @@ def test_sentence_transformers_multiple_tokens(sentence_transformers: Any) -> No
         - splitter.maximum_tokens_per_chunk
     )
     assert expected == actual
+
+
+@pytest.mark.requires("langchain")
+def test_text_splitter_from_tiktoken_start_index() -> None:
+    from langchain.docstore.document import Document
```
aowen87, line 126: @ccurme I haven't yet come across other tests that import from a different langchain package like this, so I'm not sure how I should go about this. Can you point me to an example as a reference point if you know of any?

aowen87: Maybe it would be better to move these tests to the langchain package since they rely on both `langchain.docstore.document` and `langchain_text_splitters`.

aowen87: Scratch that. I just noticed that `langchain-core` is already a test dependency. Sorry for the noise!

aowen87:
Unfortunately, this approach of finding the index offset from the tokenized/encoded chunk doesn't work because the tokenizers and encoders often don't preserve the original text (removing whitespace, for example). I don't see an easy way to fix this without a heavy rewrite.

PR #31226, opened 2025-05-14
langchain_openai: Use `convert_to_openai_messages` internally in `langchain-openai`

edwardzjl:
- **Description:** This Pull Request refactors the langchain_openai package by replacing the three internal usages of the private helper function `_convert_message_to_dict` with calls to the public utility function `langchain_core.messages.utils.convert_to_openai_messages`.

- **Issue:** This PR is not for a issue but rather a discussion: #31225

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -62,6 +62,7 @@ from langchain_core.messages import (
     ToolMessage,
     ToolMessageChunk,
     convert_to_openai_data_block,
+    convert_to_openai_messages,
     is_data_content_block,
 )
 from langchain_core.messages.ai import (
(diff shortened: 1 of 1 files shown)


edwardzjl:
I noticed that `langchain_core.messages.utils.convert_to_openai_messages` will automatically convert `None` content to an empty string (`""`):

https://github.com/langchain-ai/langchain/blob/d4f77a8c8fae9a6a33e55d572ee9e034c762eeb0/libs/core/langchain_core/messages/utils.py#L1028-L1029

However, `langchain_openai.chat_models.base._convert_message_to_dict` preserves the content as-is:

https://github.com/langchain-ai/langchain/blob/d4f77a8c8fae9a6a33e55d572ee9e034c762eeb0/libs/partners/openai/langchain_openai/chat_models/base.py#L228-L229

Since both functions are responsible for converting LangChain messages to OpenAI format, I'm wondering which behavior is preferred. Should they be consistent?

ccurme:
There are some subtle differences between these. The langchain-core util is primarily used in the context of LangSmith to convert messages from other providers to a common format.

Issue #31047, opened 2025-04-28
Latest chroma (server) release in incompatible with langchain-chroma


andreivladmatei:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

n/a

### Description

Chroma released a new version based on rust (chromadb 1.x) which requires the client to be on the same version.

Using the current langchain-chroma package (for chroma in client mode) will get you this error:

error described [here](https://github.com/chroma-core/chroma/issues/4386#issuecomment-2833775701)

### System Info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.4.0: Fri Apr 11 18:33:47 PDT 2025; root:xnu-11417.101.15~117/RELEASE_ARM64_T6000
> Python Version:  3.11.9 (v3.11.9:de54cf5be3, Apr  2 2024, 07:12:50) [Clang 13.0.0 (clang-1300.0.29.30)]

Package Information
-------------------
> langchain_core: 0.3.56
> langchain: 0.3.22
> langchain_community: 0.3.20
> langsmith: 0.3.37
> langchain_chroma: 0.2.3
> langchain_nomic: 0.1.4
> langchain_openai: 0.3.11
> langchain_text_splitters: 0.3.7
> langgraph_sdk: 0.1.60
Other Dependencies
------------------
> aiohttp<4.0.0,>=3.8.3: Installed. No version info available.
> async-timeout<5.0.0,>=4.0.0;: Installed. No version info available.
> chromadb!=0.5.10,!=0.5.11,!=0.5.12,!=0.5.4,!=0.5.5,!=0.5.7,!=0.5.9,<0.7.0,>=0.4.0: Installed. No version info available.
> dataclasses-json<0.7,>=0.5.7: Installed. No version info available.
> httpx: 0.28.1
> httpx-sse<1.0.0,>=0.4.0: Installed. No version info available.
> jsonpatch<2.0,>=1.33: Installed. No version info available.
> langchain-anthropic;: Installed. No version info available.
> langchain-aws;: Installed. No version info available.
> langchain-azure-ai;: Installed. No version info available.
[...]

PR #31251, opened 2025-05-15
partners: update deps for langchain-chroma

jairad26:
Updates dependencies to Chroma to integrate the major release of Chroma with improved performance, and to fix issues users have been seeing using the latest chroma docker image with langchain-chroma
https://github.com/langchain-ai/langchain/issues/31047#issuecomment-2850790841
Updates chromadb dependency to >=1.0.9

This also removes the dependency of chroma-hnswlib, meaning it can run against python 3.13 runners for tests as well.

Tested this by pulling the latest Chroma docker image, running langchain-chroma using client mode
```
httpClient = chromadb.HttpClient(host="localhost", port=8000)

vector_store = Chroma(
    client=httpClient,
    collection_name="test",
    embedding_function=embeddings,
)
```


jairad26:
Ran the tests and fixed the errors. I've tested the integration, and there are no breaking changes for the LangChain integration, so a minor version bump should be sufficient (0.2.4).

ccurme (approved):

PR #31268, opened 2025-05-17
huggingface[major]: Reduce disk footprint by 95% by making large dependencies optional

Simon-Stone:
**Description:** 
`langchain_huggingface` has a very large installation size of around 600 MB (on a Mac with Python 3.11). This is due to its dependency on `sentence-transformers`, which in turn depends on `torch`, which is 320 MB all by itself. Similarly, the depedency on `transformers` adds another set of heavy dependencies. With those dependencies removed, the installation of `langchain_huggingface` only takes up ~26 MB. This is only 5 % of the full installation! 

These libraries are not necessary to use `langchain_huggingface`'s API wrapper classes, only for local inferences/embeddings. All import statements for those two libraries already have import guards in place (try/catch with a helpful "please install x" message).

This PR therefore moves those two libraries to an optional dependency group `full`. So a `pip install langchain_huggingface` will only install the lightweight version, and a `pip install "langchain_huggingface[full]"` will install all dependencies.

I know this may break existing code, because `sentence-transformers` and `transformers` are now no longer installed by default. Given that users will see helpful error messages when that happens, and the major impact of this small change, I hope that you will still consider this PR.

**Dependencies:** No new dependencies, but new optional grouping.


Simon-Stone:
@ccurme Is this change something you might consider?

Simon-Stone:
Tagging for visibility:

@baskaryan, @eyurtsev, @ccurme, @vbarda, @hwchase17

I think this could be a major improvement for anyone who is using `langchain_huggingface` with remote endpoints only.

eyurtsev:
Yes. I think this makes sense. We usually wrap optional dependencies with a try except and re-raise with explicit instructions to the end user (e.g., `pip install ...`)

If you're able to make this change, we can release the integration.

We'll release with a properly updated version to mark that this is a breaking change

Simon-Stone:
I believe those guards are already in place everywhere they are needed.

eyurtsev (approved):

joegoldbeck:
This is incredible! Was just looking to solve this – big pain point indeed.

Do folks on this thread know when the release will hit PyPI?

ccurme:
> This is incredible! Was just looking to solve this – big pain point indeed.
> 
> Do folks on this thread know when the release will hit PyPI?

Just released!

PR #31298, opened 2025-05-21
langchain-huggingface: fixed reading of `self.model_id` before it is set in `_resolve_model_id()`

Haroon96:
- [ ] **PR title**: "langchain-huggingface: fixed reading of `self.model_id` before it is set in `_resolve_model_id()`"

- [ ] **PR message**: 
    - **Description:** While resolving the `model_id` when HuggingFacePipeline is used, the code loads an AutoTokenizer using the value of `self.model_id` before it is set to `self.llm.model_id`. The value of `self.model_id` is currently `None` at this point and raises an exception. I changed the code to load the tokenizer value from `self.llm.model_id` instead.


TheAnkurGoswami:
Issue (https://github.com/langchain-ai/langchain/issues/31357) resolved here https://github.com/langchain-ai/langchain/pull/31358 .

PR #31304, opened 2025-05-21
openai[patch]: support built-in code interpreter and remote MCP tools

ccurme:


libs/core/langchain_core/utils/function_calling.py
```diff
@@ -553,10 +553,16 @@ def convert_to_openai_tool(
 
         Return OpenAI Responses API-style tools unchanged. This includes
         any dict with "type" in "file_search", "function", "computer_use_preview",
-        "web_search_preview".
+        "web_search_preview", "code_interpreter", or "mcp".
```
eyurtsev, line 556: This should be 

.. versionchanged:: 0.3.60

docs/docs/integrations/chat/openai.ipynb
```diff
@@ -915,6 +915,174 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "d8c82895-5011-4062-a1bb-278ec91321e9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tool_outputs = response.additional_kwargs[\"tool_outputs\"]\n",
+    "assert len(tool_outputs) == 1\n",
+    "container_id = tool_outputs[0][\"container_id\"]\n",
```
eyurtsev, line 972: Maybe highlight this one as well? I might be just tired, but I didn't notice this line when I was skimming through the code snippet looking directly at the highlighted line below

docs/docs/integrations/chat/openai.ipynb
```diff
@@ -915,6 +915,174 @@
...
+    "Responses may then include blocks with type `\"mcp_approval_request\"`.\n",
+    "\n",
+    "To submit approvals for an approval request, structure it into a content block in an input message:\n",
+    "\n",
+    "```python\n",
+    "approval_message = {\n",
+    "    \"role\": \"user\",\n",
+    "    \"content\": [\n",
+    "        {\n",
+    "            \"type\": \"mcp_approval_response\",\n",
```
eyurtsev, line 1067: Do we want to pass it directly in openai's current format or do we want to try to generalize?

There are a lot of ways to represent this part

{
   "type": "review",
   "approve": True,
}

{
   "type": "review",
   "action": "approved",

}

---

Historically choice has always been:

1. try to support syntax as closely as possible to provider
2. do not generalize from n=1

cc @sydney-runkle worth for you to take a look at the representation as well

eyurtsev (approved):
Looks good! Main thing is whether bearer tokens will be sent as part of traces if so, we'll need to prevent that from happening

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1176,12 +1188,22 @@ def _get_invocation_params(
...
+        params = {
             "model": self.model_name,
             **super()._get_invocation_params(stop=stop),
             **self._default_params,
             **kwargs,
         }
+        # Redact headers from built-in remote MCP tool invocations
+        if (tools := params.get("tools")) and isinstance(tools, list):
+            params["tools"] = [
+                ({**tool, "headers": "**REDACTED**"} if "headers" in tool else tool)
```
eyurtsev, line 1200: Looks good. The main thing would be if it's possible to make this redaction more prominent (e.g., pull something into a standalone function), so that when one reads this code it's very easy to see that redaction is taking place

eyurtsev (approved):

Issue #31309, opened 2025-05-22
core: RunnablePick may not return a dict if keys is a string


cbornet:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```python
def test_runnablepick() -> None:
    chain = RunnablePick("foo")
    chain.invoke({"foo": "bar"})
    assert chain.invoke({"foo": "bar"}) == "bar"
```

### Description

According to its typing and documentation `RunnnablePick` should return a dict:
```python
class RunnablePick(RunnableSerializable[dict[str, Any], dict[str, Any]]):
```
but if its `keys` param is a string and not a list, it returns the value of the key, not a dict filtered on the key.

To fix, we can either:

1.  Change the signature to

```python
class RunnablePick(RunnableSerializable[dict[str, Any], Any):
```

2.  Or change the behavior to always return a dict. But that would be a breaking change...

### System Info
<details>

```sh

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.1.0: Thu Oct 10 21:06:57 PDT 2024; root:xnu-11215.41.3~3/RELEASE_ARM64_T6041
> Python Version:  3.13.2 (main, Feb 18 2025, 17:10:43) [Clang 16.0.0 (clang-1600.0.26.6)]

Package Information
-------------------
> langchain_core: 0.3.47
> langchain: 0.3.21
> langchain_community: 0.3.20
> langsmith: 0.3.18
> langchain_astradb: 0.5.3
> langchain_text_splitters: 0.3.7

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp<4.0.0,>=3.8.3: Installed. No version info available.
> astrapy: 1.5.2
> async-timeout<5.0.0,>=4.0.0;: Installed. No version info available.
> dataclasses-json<0.7,>=0.5.7: Installed. No version info available.
> httpx: 0.28.1
> httpx-sse<1.0.0,>=0.4.0: Installed. No version info available.
> jsonpatch<2.0,>=1.33: Installed. No version info available.
> langchain-anthropic;: Installed. No version info available.
[...]

PR #31321, opened 2025-05-22
fix(core): `RunnablePick` may not return a dict if keys is a string 

Swastik-Swarup-Dash:
Change made From:
```python
class RunnablePick(RunnableSerializable[dict[str, Any], dict[str, Any]]):
```
To:
```python
class RunnablePick(RunnableSerializable[dict[str, Any], Any]):
```
As suggested by @cbornet 

Fixes ##31309


ccurme (commented):
Could you add a unit test to demonstrate the change? The test should fail prior to the change and pass after.

cbornet:
I think we should first discuss in https://github.com/langchain-ai/langchain/issues/31309 what should be done.

PR #31396, opened 2025-05-28
langchain-core: Add image_generation tool to list of known openai tools

eyurtsev:
Add image generation tool to the list of well known tools. This is needed for changes in the ChatOpenAI client. TODO: Some of this logic needs to be moved from core directly into the client as changes in core should not be required to add a new tool to the openai chat client.

### DIFF
--- a/libs/core/langchain_core/utils/function_calling.py
+++ b/libs/core/langchain_core/utils/function_calling.py
@@ -515,7 +515,7 @@ _WellKnownOpenAITools = (
     "computer_use_preview",
     "code_interpreter",
     "mcp",
-    "image_generation"
+    "image_generation",
 )
 
 
--- a/libs/core/langchain_core/utils/function_calling.py
+++ b/libs/core/langchain_core/utils/function_calling.py
@@ -506,16 +506,17 @@ def convert_to_openai_function(
     return oai_function
 
 
-# List of well known tools supported by OpenAI's chat models or respones API.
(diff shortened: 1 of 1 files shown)


libs/core/langchain_core/utils/function_calling.py
```diff
@@ -558,15 +571,13 @@ def convert_to_openai_tool(
...
 
         Added support for OpenAI's built-in code interpreter and remote MCP tools.
+
+    .. versionchanged:: 0.3.63
+
+        Added support for OpenAI's image_generation built-in tool.
     """
     if isinstance(tool, dict):
-        if tool.get("type") in (
-            "function",
```
ccurme, line 564: I believe the check for `"function"` is for native OpenAI format, we should include that as well correct?

libs/core/langchain_core/utils/function_calling.py
```diff
@@ -558,15 +571,13 @@ def convert_to_openai_tool(
     .. versionchanged:: 0.3.61
 
         Added support for OpenAI's built-in code interpreter and remote MCP tools.
+
+    .. versionchanged:: 0.3.63
+
+        Added support for OpenAI's image_generation built-in tool.
```
eyurtsev, line 577:
```suggestion
        Added support for OpenAI's image generation built-in tool.
```

ccurme (approved):

PR #31413, opened 2025-05-29
langchain : set tool_choice="required" for Azure OpenAI compatibility and to make sure final response is always in required  format for with_structured_output

samueljohnsiby:
**Description:**  
For with_structured_output , i have modified tool_choice to required becuase  it has to be called to make sure response is in right format. Current value of "any" was causing issue with Azure Open AI.

**Issue:**  
When using with_structured_output with Azure  Open AI , i am getting an error that "any" value for tool choice  is not supported.

**Dependencies:**  
None.

**Twitter handle:**  
N/A

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -1460,7 +1460,7 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
 
         llm = self.bind_tools(
             [schema],
-            tool_choice="any",
+            tool_choice="required",
             ls_structured_output_format={
                 "kwargs": {"method": "function_calling"},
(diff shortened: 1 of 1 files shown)


ccurme (commented):
AzureChatOpenAI does not rely on BaseChatModel.with_structured_output, which is what you are modifying here. In its implementation, it catches `"any"` and translates to `"required"`: https://github.com/langchain-ai/langchain/blob/cc6df95e583c6f11b036a0e5a180278170272c3f/libs/partners/openai/langchain_openai/chat_models/base.py#L1498-L1501

This change would also break any chat models that do rely on the BaseChatModel implementation, assuming they work with `"any"`.

Would be curious to see code to reproduce the issue you are running into.

PR #31418, opened 2025-05-29
feat(groq): add TranscriptionGroq for Whisper transcription support

Krishnaugale353:
## Description

This PR adds support for audio transcription using Groq's Whisper models via the new `TranscriptionGroq` class.

### Highlights:
- `TranscriptionGroq` class to interact with `https://api.groq.com/openai/v1/audio/transcriptions`.
- Unit and integration tests covering success, error cases, and real transcription.
- Integration exposed via `langchain_groq.__init__`.

## Usage

```python
from langchain_groq import TranscriptionGroq

transcriber = TranscriptionGroq()
text = transcriber.transcribe("my_audio.mp3")


ccurme (commented):
Hello, thanks for this. I'm not sure this fits with LangChain, it doesn't implement any existing LangChain component or adhere to a standard interface that I know of. It also reads specifically from the filesystem and I'm not sure it makes sense for APIs to have that constraint.

PR #31424, opened 2025-05-29
openai[minor]: add image generation to responses api

eyurtsev:
Does not support partial images during generation at the moment. Before doing that I'd like to figure out how to specify the aggregation logic without requiring changes in core.


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -118,6 +118,15 @@
 
 _FUNCTION_CALL_IDS_MAP_KEY = "__openai_function_call_ids__"
 
+WellKnownTools = (
```
ccurme, line 121: should we import this from core

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -3050,6 +3053,11 @@ def _construct_responses_api_payload(
                 new_tools.append({"type": "function", **tool["function"]})
             else:
                 new_tools.append(tool)
+
+            if tool["type"] == "image_generation" and "partial_images" in tool:
+                raise NotImplementedError(
+                    "Partial image generation is not yet supported via the LangChain ChatOpenAI client. "
```
ccurme: mild preference for dropping `partial_images` and emitting a warning.

alternatively, could update the error message here to make it very obvious what to do (drop `partial_images` from the `image_generation` tool definition)

libs/partners/openai/langchain_openai/chat_models/base.py
eyurtsev: Not at the moment. We should do something in the reverse direction likely -- core should not know anything specific about the responses API. We should resolve in a separate refactor to figure out how to decouple core completely from openai specifics.

libs/partners/openai/langchain_openai/chat_models/base.py
eyurtsev: Updated the doc-string

ccurme (approved):

PR #31425, opened 2025-05-29
standard-tests: migrate to pytest-recording

ccurme:


libs/standard-tests/langchain_tests/conftest.py
```diff
@@ -1,25 +1,85 @@
...
+        # If cassette path is already Path this is a no-op
+        cassette_path = Path(cassette_path)
+        if not cassette_path.is_file():
+            raise CassetteNotFoundError(
+                f"Cassette file {cassette_path} does not exist."
+            )
+        with cassette_path.open(mode="rb") as f:
+            data = f.read()
+        deser = serializer.deserialize(data)
+        return deser["requests"], deser["responses"]
```
ccurme, line 61: @eyurtsev I updated the return type of `load_cassette` following https://github.com/kevin1024/vcrpy/blob/19bd4e012c8fd6970fd7f2af3cc60aed1e5f1ab5/vcr/cassette.py#L361

ccurme (commented):
Looks like request / response body are loaded as bytes, not sure if that's intended / desired

Screenshot 2025-05-30 at 11 24 51 AM

ccurme:
@eyurtsev FYI the new serializer is noticeably slower, I don't think it's a problem so will ignore the regressions flagged by codspeed.

libs/partners/openai/tests/conftest.py
```diff
@@ -13,36 +13,36 @@
...
     config["before_record_request"] = remove_request_headers
     config["before_record_response"] = remove_response_headers
     config["serializer"] = "yaml.gz"
     config["path_transformer"] = VCR.ensure_suffix(".yaml.gz")
-
     return config
 
 
 @pytest.fixture
 def vcr(vcr_config: dict) -> VCR:
```
ccurme, line 43: @eyurtsev blocking issue:

if we don't define the `vcr` fixture like this, it gets value None for some reason in `langchain-tests`:
```
cd libs/partners/anthropic
rm tests/cassettes/*
ANTHROPIC_API_KEY=... uv run python -m pytest --capture=no tests/integration_tests/test_standard.py::TestAnthropicStandard::test_stream_time
```
however, implementing the `vcr` fixture here might overwrite pytest-recording's fixture, so if you decorate a test with `@pytest.mark.vcr` no cassette is generated.

I suspect we should remove the `vcr` fixture implementation here and somehow help the test in langchain-tests pick up the fixture.

ccurme: Believe this is fixed in https://github.com/langchain-ai/langchain/pull/31425/commits/3aa729d2e3986d1c6ecbd4bb16a9278a75c433c8

PR #31428, opened 2025-05-29
refactor: trace configurable as metadata

baskaryan:


mdrxy:
Can I get some context on this @baskaryan? Triaging old PRs and want to see if this is something we should address for 0.4.

Issue #31429, opened 2025-05-30
module 'defusedxml' has no attribute 'ElementTree'


yosheng:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

here is demo code
```python
from langchain_openai import ChatOpenAI
from langchain_core.prompts import PromptTemplate
from langchain_core.output_parsers import XMLOutputParser

model = ChatOpenAI(model="gpt-4o", temperature=0)

actor_query = "Generate a simplified list of Ang Lee's films, sorted by most recent in descending order"
parser = XMLOutputParser(tags=["movies", "actor", "film", "name", "genre"])
prompt = PromptTemplate(
    template="Answer \n{format_instructions}\n{query}\n",
    input_variables=["query"],
    partial_variables={"format_instructions": parser.get_format_instructions()},
)

chain = prompt | model | parser
for s in chain.stream({"query": actor_query}):
    print(s)
```

### Error Message and Stack Trace (if applicable)

Traceback (most recent call last):
  File "D:\Workspace\ai-learning\day7-demo\output-parser\xml_output_parser_stream_demo.py", line 16, in <module>
    for s in chain.stream({"query": actor_query}):
             ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Workspace\ai-learning\venv\Lib\site-packages\langchain_core\runnables\base.py", line 3438, in stream
    yield from self.transform(iter([input]), config, **kwargs)
  File "D:\Workspace\ai-learning\venv\Lib\site-packages\langchain_core\runnables\base.py", line 3424, in transform
    yield from self._transform_stream_with_config(
    ...<4 lines>...
    )
  File "D:\Workspace\ai-learning\venv\Lib\site-packages\langchain_core\runnables\base.py", line 2215, in _transform_stream_with_config
    chunk: Output = context.run(next, iterator)
                    ~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "D:\Workspace\ai-learning\venv\Lib\site-packages\langchain_core\runnables\base.py", line 3386, in _transform
    yield from final_pipeline
[...]

PR #31431, opened 2025-05-30
fix: module 'defusedxml' has no attribute 'ElementTree' (#31429)

yosheng:


eyurtsev:
@cbornet do you mind taking a quick look?

libs/core/langchain_core/output_parsers/xml.py
```diff
@@ -207,7 +207,7 @@ def parse(self, text: str) -> dict[str, Union[str, list[Any]]]:
         # likely if you're reading this you can move them to the top of the file
         if self.parser == "defusedxml":
             try:
-                from defusedxml import ElementTree
+                import defusedxml.ElementTree as ElementTree
```
eyurtsev, line 210: @yosheng could you confirm which version of defusedxml you're on? Could we confirm that this is a false positive from the ruff linter vs. some other bug?

yosheng: I have tried versions 0.8.0rc2, 0.7.1, and 0.6.0, but the problem still persists.

libs/core/langchain_core/output_parsers/xml.py
```diff
@@ -51,15 +51,15 @@ def __init__(self, parser: Literal["defusedxml", "xml"]) -> None:
         """
         if parser == "defusedxml":
             try:
-                import defusedxml  # type: ignore[import-untyped]
+                from defusedxml.ElementTree import XMLParser as DefusedXMLParser
```
cbornet: We can just use XMLParser without renaming
```suggestion
                from defusedxml.ElementTree import XMLParser

File: libs/core/langchain_core/output_parsers/xml.py [hunk 3]
```diff
@@ -51,15 +51,15 @@ def __init__(self, parser: Literal["defusedxml", "xml"]) -> None:
...
             except ImportError as e:
                 msg = (
                     "defusedxml is not installed. "
                     "Please install it to use the defusedxml parser."
                     "You can install it with `pip install defusedxml` "
                 )
                 raise ImportError(msg) from e
-            _parser = defusedxml.ElementTree.DefusedXMLParser(target=TreeBuilder())
+            _parser = DefusedXMLParser(target=TreeBuilder())
```
cbornet, line 62:
```suggestion
            _parser = XMLParser(target=TreeBuilder())
```

libs/core/langchain_core/output_parsers/xml.py
cbornet, line 210: Why this change ?

PR #31465, opened 2025-06-02
docs: fix grammar issues in airbyte_shopify.ipynb and airbyte_stripe.ipynb

MichaelLi65535:
fix grammar issue


eyurtsev:
@MichaelLi65535 would you be able to update the PR summary in the PRs? It can even be completely empty. The goal is to make sure that git log commits don't have the template in them in case folks are relying on git log to review history

PR #31468, opened 2025-06-02
feat(perplexity): expose `search_results` in chat model

inacionery:
Description
The Perplexity chat model already returns a search_results field, but LangChain dropped it when mapping Perplexity responses to additional_kwargs.
This patch adds "search_results" to the allowed attribute lists in both _stream and _generate, so downstream code can access it just like images, citations, or related_questions.

Dependencies
None. The change is purely internal; no new imports or optional dependencies required.

https://community.perplexity.ai/t/new-feature-search-results-field-with-richer-metadata/398

### DIFF
--- a/docs/docs/integrations/chat/perplexity.ipynb
+++ b/docs/docs/integrations/chat/perplexity.ipynb
@@ -248,7 +248,7 @@
     "### Accessing the search results metadata\n",
     "\n",
     "Perplexity often provides a list of the web pages it consulted (“search_results”).\n",
-    "You don’t need to pass any special parameter — the list is placed in\n",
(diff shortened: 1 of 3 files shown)


mdrxy:
Hi @inacionery, when I run the code in your notebook I don't see `'search_results'` returned - only `'citations'`. Could you update and add integration tests for this?

inacionery:
Hi @mdrxy,

The notebook in this PR already includes the search_results section. Could you confirm you’re running it against your local langchain-perplexity code and that it prints a list under response.additional_kwargs["search_results"] without error?

Thanks!

mdrxy (approved):
LGTM

PR #31361, opened 2025-05-26
docs: replace deprecated `initialize_agent` with `create_react_agent` and non-deprecated functions

tk1475:
**Description:**  
This PR updates approximately 4' occurences of the deprecated `initialize_agent` function in LangChain documentation and examples, replacing it with the recommended `create_react_agent` and  pattern. It also refactors related examples to align with current best practices.

**Issue:**  
Partially Fixes #29277

**Dependencies:**  
None

**X handle:**  
@TK1475

PR #31492, opened 2025-06-04
Revert "docs: replace deprecated `initialize_agent` with `create_react_agent` and non-deprecated functions"

eyurtsev:
Reverts langchain-ai/langchain#31361


eyurtsev:
cc @tk1475 -- focus on changing a single doc rather than 4 if you want to make the changes since they still have errors in them.

tk1475:
@eyurtsev, how would you want me to test out the code, as it seemed to be running fine (without the use of an API key) for me? I'll stick to a singular change if that's ideal. Could you help me a little in this regard?

PR #31518, opened 2025-06-06
openai[patch]: pass through with_structured_output kwargs

baskaryan:
Support 
```python


class ResponseSchema(BaseModel):
    response: str


    """Get weather"""
    pass

structured_llm = llm.with_structured_output(
    ResponseSchema,
    tools=[get_weather],
    strict=True,
    include_raw=True,
    tool_choice="required",
    parallel_tool_calls=False,
)

structured_llm.invoke("whats up?")
```

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/azure.py
+++ b/libs/partners/openai/langchain_openai/chat_models/azure.py
@@ -854,7 +854,7 @@ class AzureChatOpenAI(BaseChatOpenAI):
                             "parsed": None,
                         }
 
-                kwargs: Additional keyword args are passed through to the model.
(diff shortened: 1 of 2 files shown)


libs/partners/openai/langchain_openai/chat_models/azure.py
```diff
@@ -880,6 +880,12 @@ def get_weather(location: str) -> str:
 
             ``method`` default changed from "function_calling" to "json_schema".
 
+        .. versionchanged:: 0.3.12
```
ccurme, line 883: thank you

ccurme (approved):

PR #31533, opened 2025-06-08
feat(huggingface): batching in huggingface pipeline

TheAnkurGoswami:
This PR introduces true batching support for Hugging Face pipelines.
The previous implementation simulated batching by launching multiple parallel calls using ContextThreadPoolExecutor - a pattern more suited for API-based workloads [reference code](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/runnables/base.py#L763-L797).

With this update, batches are now processed natively through Hugging Face models/pipelines running on local GPUs, enabling more efficient and performant execution.


TheAnkurGoswami:
@baskaryan @ccurme @hwchase17

libs/partners/huggingface/langchain_huggingface/chat_models/huggingface.py
```diff
@@ -749,10 +783,10 @@ def _to_chatml_format(self, message: BaseMessage) -> dict:
     @staticmethod
     def _to_chat_result(llm_result: LLMResult) -> ChatResult:
         chat_generations = []
-
-        for g in llm_result.generations[0]:
+        for g in llm_result.generations:
             chat_generation = ChatGeneration(
-                message=AIMessage(content=g.text), generation_info=g.generation_info
```
ccurme, line 984: How does this change impact `invoke`?

TheAnkurGoswami: This won't affect `invoke`.
```python
llm_result = LLMResult(
    generations=[
        [Generation(text=...)]    # Output result 1
        # if batch is called, we get more output results corresponding to their resp. inputs 
        [Generation(text=...)]    # Output result 2
        ...
    ]
)
```

For `invoke`, we pass one input only & get only one output result.

TheAnkurGoswami:
@mdrxy @ccurme

TheAnkurGoswami:
Hi @ccurme @mdrxy, following up on this PR - any updates needed, or are we good to merge?

libs/partners/huggingface/langchain_huggingface/chat_models/huggingface.py
```diff
@@ -714,6 +719,34 @@ async def _astream(
...
+            return []
+
+        if _is_huggingface_pipeline(self.llm):
+            prompt_messages = [
+                self._convert_input(input_).to_messages() for input_ in inputs
+            ]
+
+            llm_inputs = list(map(self._to_chat_prompt, prompt_messages))
+
+            llm_results = self.llm._generate(prompts=llm_inputs)
```
ccurme, line 970: Can we delegate to `.invoke`, as in the [base implementation](https://github.com/langchain-ai/langchain/blob/10dfeea110e3160b119e7c0fee048cc7a9c3d241/libs/core/langchain_core/runnables/base.py#L871) for `batch`?

The problem with this is that we lose callbacks, LangSmith tracing, propagation of configs, caching, and all other features that are handled in langchain-core.

ccurme (commented):
Thanks for this, sorry for the delay on review.

Left a comment. Also wondering if it makes sense to override `abatch`.

libs/partners/huggingface/langchain_huggingface/chat_models/huggingface.py
TheAnkurGoswami: I'm not sure if that would be possible as the `.invoke` method here calls the model(endpoint in most cases) in parallel. The above mentioned features are maintaining the Input(Request)-Output pair.

The `.batch` functionality that's been implemented in this PR is for in-memory model.
So we will be having one config, callback, etc for a batch. If this approach sounds feasible to you, I can look into it.

TheAnkurGoswami: @ccurme

Issue #31565, opened 2025-06-11
`GenerationChunk` addition raises `TypeError` when `generation_info` contains duplicate keys with primitive types


Shibayan003:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```python
from langchain_core.outputs import GenerationChunk

chunk1 = GenerationChunk(text="", generation_info={"len": 0})
chunk2 = GenerationChunk(text="Non-empty text", generation_info={"len": 14})
result = chunk1 + chunk2  # Raises TypeError
```
### Error Message and Stack Trace (if applicable)

TypeError: Additional kwargs key len already exists in left dict and value has unsupported type <class 'int'>.

### Description

When I tried adding two GenerationChunk instances that have the same key in their generation_info, it raised a TypeError. I expected GenerationChunk addition to handle this properly, since GenerationChunk often contains metadata like len, index, etc., which can have duplicate keys.

### System Info

System Information
------------------
> OS:  Linux
> OS Version:  #148-Ubuntu SMP Fri Mar 14 19:05:48 UTC 2025
> Python Version:  3.10.16 (main, Dec 11 2024, 16:24:50) [GCC 11.2.0]

Package Information
-------------------
> langchain_core: 0.3.28
> langchain: 0.3.13
> langchain_community: 0.3.13
> langsmith: 0.1.137
> langchain_anthropic: 0.3.1
> langchain_aws: 0.2.2
> langchain_chroma: 0.2.0
> langchain_experimental: 0.3.2
> langchain_fireworks: 0.2.6
> langchain_google_vertexai: 2.0.5
> langchain_groq: 0.2.2
> langchain_mistralai: 0.2.4
> langchain_openai: 0.2.14
> langchain_text_splitters: 0.3.4
> langchain_together: 0.2.0
> langchain_unstructured: 0.1.5
> langgraph_sdk: 0.1.34

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp: 3.10.10
> anthropic: 0.40.0
> anthropic[vertexai]: Installed. No version info available.
> async-timeout: 4.0.3
[...]

PR #31572, opened 2025-06-11
core[patch]: Int Combine when Merging Dicts

keenborder786:
- **Description:** Combining the Int Types by adding them which makes the most sense.
- **Issue:**  #31565


rawathemant246 (approved):

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev

eyurtsev:
@ccurme assigning to you since you might be more up to date.. Most changes to these code paths in the past were wrong, if that's the case, we should probably add a documentation string to explain what the correct behavior is w/ respect to ints/floats

keenborder786:
@ccurme

ccurme (commented):
This should be OK.

Raising here has caught many bugs in the past, so summing instead will likely let bugs through, but that on its own isn't a reason to raise.

N.B. AIMessages have a field [usage_metadata](https://python.langchain.com/api_reference/core/messages/langchain_core.messages.ai.AIMessage.html#langchain_core.messages.ai.AIMessage.usage_metadata) that sums as expected. @Shibayan003
I'm curious what data you are storing that you'd like to sum across chunks this way.

PR #31591, opened 2025-06-12
core: In LangChainTracer, send only the first token event

nfcampos:
- only the first one is used for analytics


agola11 (approved):

libs/core/langchain_core/tracers/langchain.py
```diff
@@ -235,6 +236,11 @@ def _llm_run_with_token_event(
         parent_run_id: Optional[UUID] = None,
     ) -> Run:
         """Append token event to LLM run and return the run."""
+        run_id_str = str(run_id)
+        if run_id_str not in self.run_has_token_event_map:
+            self.run_has_token_event_map[run_id_str] = True
```
hinthornw, line 241: Should we pop this in _end_trace?

PR #31594, opened 2025-06-12
bump langsmith to allow 0.4

EugeneJinXin:
Langsmith 0.4 is launched so bump it up across OSS: langchain and langchain-core. Will have separate langsmith-doc announcement for that

### DIFF
--- a/libs/langchain/pyproject.toml
+++ b/libs/langchain/pyproject.toml
@@ -9,7 +9,7 @@ requires-python = ">=3.9"
 dependencies = [
     "langchain-core<1.0.0,>=0.3.58",
     "langchain-text-splitters<1.0.0,>=0.3.8",
-    "langsmith<0.4,>=0.1.17",
+    "langsmith<0.5,>=0.1.17",
     "pydantic<3.0.0,>=2.7.4",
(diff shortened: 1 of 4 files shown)


libs/langchain/pyproject.toml
```diff
@@ -76,7 +76,7 @@ test = [
 ]
 codespell = ["codespell<3.0.0,>=2.2.0"]
 test_integration = [
-    "pytest-vcr<2.0.0,>=1.0.2",
```
EugeneJinXin, line 79: About this unexpected change:  


After bumping langsmith i hit this Integration Test [error](https://github.com/langchain-ai/langchain/actions/runs/15622944899/job/44011626693?pr=31594) in ci/cd.

My best-effort attempt to fix is remove pytest-vcr and use vcrpy directly after looking online.

This is a conflict between pytest-recording and pytest-vcr: INTERNALERROR> RuntimeError: `pytest-recording` is incompatible with `pytest-vcr`. Please, uninstall `pytest-vcr` in order to use `pytest-recording`.

ccurme (approved):

PR #31608, opened 2025-06-14
fix: fixing missing Docstring Bug if no Docstring is provided in BaseModel class

### LINKED ISSUE #31606: bug[core]: fix tool descriptions when BaseModels have no docstring
### Privileged issue

- [x] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

When Pydantic models are cast to BaseTool and are missing a docstring, we use the docstring for BaseModel itself:

```python


@tool
class MyTool(BaseModel):
    foo: str


MyTool.description
```
> '!!! abstract "Usage Documentation"\n    [Models](../concepts/models.md)\n\nA base class for creating Pydantic models.\n\nAttributes:\n    __class_vars__: The names of the class variables defined on the model.\n    __private_attributes__: Metadata about...'

We should use an empty string in this case.

keenborder786:
- **Description:** Ensure that the tool description is an empty string when creating a Structured Tool from a Pydantic class in case no description is provided
- **Issue:** Fixes #31606


libs/core/langchain_core/tools/structured.py
```diff
@@ -16,6 +16,7 @@
 )
 
 from pydantic import Field, SkipValidation
+from pydantic._internal._model_construction import ModelMetaclass
```
eyurtsev: We need to do this without reaching into `pydantic._internal` which is a private API for pydantic. They reserve the right to change it at any point, and that would break things in langchain

The simplest way to achieve this is by comparing the doc-string we get to the default doc-string for BaseModel.

keenborder786: ok @eyurtsev

keenborder786:
@eyurtsev please review

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev please let me know if there is still an issue?

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev

PR #31619, opened 2025-06-15
openai[patch]:  Improve error message when response type is malformed

LKChenLK:
**Description:** Currently the key, 'choices', is assumed to be present in the response dict. However not all OpenAI-compatible servers return the response correctly. This modification helps identify the malformed response instead of obfuscating it.
 
**Issue:** A malformed response that does not contain 'choices' would be obfuscated with `TypeError: 'NoneType' object is not iterable`, making it hard to troubleshoot. 

**Dependencies**: None

**Twitter handle:** None


ccurme (approved):
Thanks for this, I see you marked this WIP but this looks great to me, merging but please ping if you had other intentions and feel free to follow up!

LKChenLK:
Thanks for the review and comments! Just wondering, is there a reason for not raising the error message with the entire `response_dict` received ? IME it'd make it easier for troubleshooting / more detailed error handling downstream instead of receiving just the keys.

ccurme:
> Thanks for the review and comments! Just wondering, is there a reason for not raising the error message with the entire `response_dict` received ? IME it'd make it easier for troubleshooting / more detailed error handling downstream instead of receiving just the keys.

Mostly in case the values had sensitive information, or large payloads.

PR #31628, opened 2025-06-16
text-splitters[patch]: update uv.lock

ccurme:


ccurme:
https://github.com/langchain-ai/langchain/pull/31626 beat me to it

PR #31639, opened 2025-06-17
core: Improve `RunnableWithMessageHistory` init arg types

cbornet:
`Runnable`'s `Input` is contravariant so we need to enumerate all possible inputs and it's not possible to put them in a `Union`.
Also, it's better to only require a runnable that accepts`list[BaseMessage]` instead of a broader `Sequence[BaseMessage]` as internally the runnable is only called with a list.

### DIFF
--- a/libs/core/langchain_core/runnables/history.py
+++ b/libs/core/langchain_core/runnables/history.py
@@ -241,7 +241,11 @@ class RunnableWithMessageHistory(RunnableBindingBase):
         self,
         runnable: Union[
             Runnable[
-                Union[MessagesOrDictWithMessages],
+                list[BaseMessage],
+                Union[str, BaseMessage, MessagesOrDictWithMessages],
+            ],
+            Runnable[
+                dict[str, Any],
                 Union[str, BaseMessage, MessagesOrDictWithMessages],
             ],
             LanguageModelLike,
@@ -258,7 +262,7 @@ class RunnableWithMessageHistory(RunnableBindingBase):
 
         Args:
             runnable: The base Runnable to be wrapped. Must take as input one of:
-                1. A sequence of BaseMessages
+                1. A list of BaseMessages
                 2. A dict with one key for all messages
                 3. A dict with one key for the current input string/message(s) and
                     a separate key for historical messages. If the input key points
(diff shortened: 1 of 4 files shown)


eyurtsev (approved):

libs/core/langchain_core/runnables/history.py
```diff
@@ -241,7 +241,11 @@ def __init__(
         self,
         runnable: Union[
             Runnable[
-                Union[MessagesOrDictWithMessages],
+                list[BaseMessage],
```
eyurtsev, line 244: I think this supported Sequence (which is both tuple and list)

eyurtsev: https://github.com/langchain-ai/langchain/blob/29e17fbd6bcba110b4004a3bbd4697fb3cc400eb/libs/core/langchain_core/runnables/history.py#L37-L37

cbornet: Counter-intuitively, this would not be nice to users as it means their runnable must be able to handle a `Sequence` whereas RunnableWithMessageHistory always calls it with a `list` (it gets the result of `def _enter_history(self, value: Any, config: RunnableConfig) -> list[BaseMessage]:`). So for the open-close principle, using `list` is better than `Sequence` here.

cbornet: For instance, if we set `Sequence` here, then `def _fake_llm(messages: list[BaseMessage]) -> list[BaseMessage]:` in `test_ignore_session_id` must be changed to `def _fake_llm(messages: Sequence[BaseMessage]) -> list[BaseMessage]:`

libs/core/tests/unit_tests/runnables/test_history.py
```diff
@@ -876,8 +876,8 @@ def test_get_output_messages_with_value_error() -> None:
         with_history.bound.invoke([HumanMessage(content="hello")], config)
 
     illegal_int_message = 123
-    runnable = _RunnableLambdaWithRaiseError(lambda _: illegal_int_message)
-    with_history = RunnableWithMessageHistory(runnable, get_session_history)
+    runnable2 = _RunnableLambdaWithRaiseError(lambda _: illegal_int_message)
+    with_history = RunnableWithMessageHistory(runnable2, get_session_history)  # type: ignore[arg-type]
```
cbornet, line 880: this test is precisely to check the wrong runnable output type at runtime. The typing now also detects the problem at mypy check time.

eyurtsev (approved):

PR #31649, opened 2025-06-18
core[patch]: Add additional hashing options to indexing API, warn on SHA-1

eyurtsev:
Add additional hashing options to the indexing API, warn on SHA-1

Requires:

- Bumping langchain-core version
- bumping min langchain-core in langchain


libs/core/langchain_core/indexing/api.py
```diff
@@ -291,6 +317,21 @@ def index(
         force_update: Force update documents even if they are present in the
             record manager. Useful if you are re-indexing with updated embeddings.
             Default is False.
+        key_encoder: Hashing algorithm to use for hashing the document content and
+            metadata. Default is "sha1".
+            Other options include "blake2b", "sha256", and "sha512".
+            .. versionadded:: 0.3.??
```
eyurtsev:
```suggestion
            .. versionadded:: 0.3.66
```

libs/langchain/tests/unit_tests/indexes/test_hashed_document.py
```diff
@@ -1,50 +0,0 @@
-import pytest
```
eyurtsev, line 1: Deleted since it was duplicated with test_hashed_document.py in core

libs/core/langchain_core/indexing/api.py
```diff
@@ -168,21 +124,103 @@ def _get_source_id_assigner(
...
+        return hashlib.sha512(text.encode("utf-8")).hexdigest()
+    msg = f"Unsupported hashing algorithm: {algorithm}"
+    raise ValueError(msg)
+
+
+def _get_document_with_hash(
+    document: Document,
+    *,
+    key_encoder: Callable[[Document], str]
+    | Literal["sha1", "sha256", "sha512", "blake2b"],
```
ccurme: (here and below) Is this python 3.9 compatible? Not sure why CI isn't picking it up.

libs/core/langchain_core/indexing/api.py
```diff
@@ -168,21 +124,103 @@ def _get_source_id_assigner(
...
+    """
+    metadata: dict[str, Any] = dict(document.metadata or {})
+
+    if callable(key_encoder):
+        # If key_encoder is a callable, we use it to generate the hash.
+        hash_ = key_encoder(document)
+    else:
+        # The hashes are calculated separate for the content and the metadata.
+        content_hash = _calculate_hash(document.page_content, algorithm=key_encoder)
+        serialized_meta = json.dumps(metadata, sort_keys=True)
```
ccurme: Is `metadata` always json serializable in this case? should we catch `TypeError`? looks like earlier we were catching `Exception`.

libs/core/langchain_core/indexing/api.py
```diff
@@ -291,6 +331,21 @@ def index(
         force_update: Force update documents even if they are present in the
             record manager. Useful if you are re-indexing with updated embeddings.
             Default is False.
+        key_encoder: Hashing algorithm to use for hashing the document content and
+            metadata. Default is "sha1".
+            Other options include "blake2b", "sha256", and "sha512".
+            .. versionadded:: 0.3.66
```
ccurme, line 347: You may need some whitespace for it to render properly:
```suggestion

            .. versionadded:: 0.3.66

```
Screenshot 2025-06-20 at 11 29 59 AM

ccurme (approved):

libs/core/langchain_core/indexing/api.py
eyurtsev: good catch -- i'll update to 3.10

PR #31655, opened 2025-06-18
[Doc] Improve api doc for DeepSeek

ZhangShenao:
- Add param in api doc
- Fix word spelling

### DIFF
--- a/libs/partners/deepseek/langchain_deepseek/chat_models.py
+++ b/libs/partners/deepseek/langchain_deepseek/chat_models.py
@@ -354,7 +354,7 @@ class ChatDeepSeek(BaseChatOpenAI):
                 response will be returned. If an error occurs during output parsing it
                 will be caught and returned as well. The final output is always a dict
                 with keys "raw", "parsed", and "parsing_error".
-                
+
             strict:
                 Unsupported: whether to enable strict schema adherence when generating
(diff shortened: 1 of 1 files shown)


libs/partners/deepseek/langchain_deepseek/chat_models.py
```diff
@@ -355,6 +355,11 @@ def with_structured_output(
                 will be caught and returned as well. The final output is always a dict
                 with keys "raw", "parsed", and "parsing_error".
 
+            strict:
+                Unsupported: whether to enable strict schema adherence when generating
+                the output. This parameter is included for compatibility with other
+                chat models, but is currently ignored.
```
ccurme: Is it ignored? It will impact the payload that's sent to the `openai` client (e.g., if `strict` is True, the tool schema will have `'strict': True`.

Do you know that DeepSeek is ignoring this parameter server-side? I suspect that it is, because it's not documented in their docs, but I'm not sure that would motivate me to document here that they explicitly ignore it.

ZhangShenao: I have confirmed that DeepSeek's server does not support this parameter `strict`, but I think you are right, and there is no need to ignore it. I have modified the PR, please review.

ccurme (approved):

PR #31686, opened 2025-06-20
openai[patch]: allow specification of output format for Responses API

ccurme:
Add an `output_version` attribute to BaseChatOpenAI. Motivation is to allow users to opt-in to breaking changes in AIMessage formats.

`"v0"` is the default and corresponds to current format. We intend to introduce standard types for reasoning, citations, and other AIMessage content with `"v1"`. At that point we will add the attribute to BaseChatModel.

Here we implement `"responses/v1"` to allow users to opt-in to the change described in https://github.com/langchain-ai/langchain/pull/31587 — this is a breaking change that is necessary to support some features (e.g., remote MCP tool use under zero data retention contexts).


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -305,6 +305,20 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
     - If False (default), will always use streaming case if available.
     """
 
+    output_version: str = "v0"
```
eyurtsev: Can we remove this until there's more than one version?

sydney-runkle (commented):
This makes a lot of sense - thanks for the clean docs, etc.

A few follow up questions: 
* Can we make the default `v1` when we bump version to v1?
* I presume `v1` will be using a new form of standard output?

A higher level question: isn't our job to support a uniform output format across all models? In that sense, shouldn't we not support `responses` format specifically? That seems quite tied to openai... perhaps this just necessitates the shift towards, for lack of a better word, `stdout` 😉

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -615,6 +615,22 @@ class BaseChatOpenAI(BaseChatModel):
...
+    This field is used to roll-out new output formats for chat model AIMessages
+    in a backwards-compatible way.
+
+    Supported values:
+
+    - ``"v0"``: AIMessage format as of langchain-openai 0.3.x.
+    - ``"responses/v1"``: Formats Responses API output
+      items into AIMessage content blocks.
+
+      .. versionadded:: 0.3.25
```
hntrl: Is it fair to say that `v0` is the deprecated output version? Do we want to make a stronger indication of that if that's the case?

ccurme: Added some language here, thanks.

PR #31711, opened 2025-06-23
core: adding reasoning_content to AIMessage

MichalTesnar:
- **Description:** This PR adds reasoning_content to the AI Message. This is returned by the OpenAI API used also by DeepSeek https://api-docs.deepseek.com/guides/reasoning_model
- **Issue:** Fixes #31326 
- **Dependencies:**: None

Note from the author: I made minimal amount of changes I have found fit. I hope you like the proposal, there seem to be many people asking for it. I understand that it might need some more work. Please let me know what I should specifically do about (1) testing -- adding the attribute to more tests and (2) docs -- what exactly you would like to see mentioned.


ccurme (commented):
Thanks for this. I commented on the issue, but you should be able to just use `langchain-deepseek` out of the box for this. Please let me know if it's not the case.

We are designing a standard interface for reasoning output but it will not be exposed as a string attribute on AIMessage. The reason is that in some settings (notably the OpenAI Responses API) the order of the reasoning items within the output sequence is crucial, and we lose that if we separate it into a separate attribute.

MichalTesnar:
Thanks for the response, one quick question.

> The reason is that in some settings (notably the OpenAI Responses API) the order of the reasoning items within the output sequence is crucial, and we lose that if we separate it into a separate attribute.

Can you explain what exactly do you mean? Aren't reasoning items just a list of strings that always just precede the final answer of the LLM?
@ccurme

ccurme:
> Can you explain what exactly do you mean? Aren't reasoning items just a list of strings that always just precede the final answer of the LLM? @ccurme

For OpenAI specifically, [reasoning items](https://platform.openai.com/docs/guides/reasoning?api-mode=responses) are objects that are interleaved into their responses. There might be reasoning, calls to internal tools, and more reasoning before a final response. The reasoning items don't need to have human-readable text at all; they are primarily references to data that OpenAI stores server-side and does not expose to the client (even in zero data retention scenarios, the reasoning content is [encrypted](https://platform.openai.com/docs/guides/reasoning?api-mode=responses#encrypted-reasoning-items)).

PR #31716, opened 2025-06-24
docs: replace deprecated llama 3 model with sonar in ChatPerplexity example

w3bdesign:
**Description:** Updates ChatPerplexity documentation to replace deprecated llama 3 model reference with the current sonar model in the API key example code block.

**Issue:** N/A (maintenance update for deprecated model)

**Dependencies:** No new dependencies required


0x090909:
the perpelxychat class is deprecated itself when using it with langflow the max token parameter is simply ignored

PR #31756, opened 2025-06-26
openai[fix]: Correctly set usage metadata for OpenAI Responses API

MrJaeger:
**Description:**
Fixes the `_create_usage_metadata_responses` function to match the actual shape of the `usage` object found on an OpenAI Response object (see reference [here](https://platform.openai.com/docs/api-reference/responses/object#responses/object-usage))


ccurme (approved):
Thanks very much for this!

PR #31759, opened 2025-06-26
docs: enhance docstring for `disable_streaming` parameter in BaseChatModel

### LINKED ISSUE #31758: docs: Improve `disable_streaming` docstring
At first glance, it might be unclear why the `disable_streaming` param exists for chat models. I propose expanding the docstring or adding a dedicated page somewhere else in the docs to illuminate why someone might set this.

Inspiring comment *"Would a bit of documentation here be helpful to explain WHY one might not be able to stream?"*

_Originally posted by @eyurtsev in https://github.com/langchain-ai/langchain/pull/25070#discussion_r1707254859_

mdrxy:
Resolves #31758

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -301,25 +301,14 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
 
     - If True, will always bypass streaming case.
     - If ``'tool_calling'``, will bypass streaming case only when the model is called
-      with a ``tools`` keyword argument.
+      with a ``tools`` keyword argument. In other words, LangChain will automatically
+      switch to non-streaming behavior (``invoke()``) only when the tools argument is
+      provided. This offers the best of both worlds.
     - If False (default), will always use streaming case if available.
 
-    Useful when integrating with systems or models that are not designed to handle
-    streaming data. For instance, in situations where a model's feature, such as tool
-    calling, does not support a token-by-token stream of the final answer. By setting
-    ``disable_streaming``, the ``stream()`` method can still be called, but it will
-    behave like ``invoke()``, returning the full response in a single chunk. This
-    prevents errors and allows for a uniform method call regardless of the underlying
-    model's capabilities.
-
-    Second, if an LLM decides to use a tool but doesn't support streaming tool calls, it
-    will generate a complete tool call request within a single chunk. If you are using
-    the ``stream()`` method, this can lead to unexpected behavior or errors if the
-    application is expecting a continuous stream of text.
-
-    By setting ``disable_streaming='tool_calling'``, you instruct LangChain to
-    automatically switch to a non-streaming (``invoke()``) behavior only when the tools
-    argument is provided. This offers the best of both worlds.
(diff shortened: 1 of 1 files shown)


sydney-runkle (approved):
Nice, drive by docs improvements are super helpful.

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -300,9 +300,27 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
...
+    ``disable_streaming``, the ``stream()`` method can still be called, but it will
+    behave like ``invoke()``, returning the full response in a single chunk. This
+    prevents errors and allows for a uniform method call regardless of the underlying
+    model's capabilities.
+
+    Second, when a LLM decides to use a tool, it doesn't stream the thought process of
+    the tool call token by token in the final answer. Instead, it generates a complete
+    tool call request. If you are using the ``stream()`` method, this can lead to
+    unexpected behavior or errors if the application is expecting a continuous stream of
+    text.
```
ccurme: This sounds like a general statement about LLMs, I would clarify that this is an example (e.g., "If an LLM doesn't support streaming tool calls"). Most LLMs we integrate with do support [streaming tool calls](https://python.langchain.com/docs/how_to/tool_streaming/).

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -312,11 +312,10 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
...
 
-    Second, when a LLM decides to use a tool, it doesn't stream the thought process of
-    the tool call token by token in the final answer. Instead, it generates a complete
-    tool call request. If you are using the ``stream()`` method, this can lead to
-    unexpected behavior or errors if the application is expecting a continuous stream of
-    text.
+    Second, if an LLM decides to use a tool but doesn't support streaming tool calls, it
+    will generate a complete tool call request within a single chunk. If you are using
+    the ``stream()`` method, this can lead to unexpected behavior or errors if the
+    application is expecting a continuous stream of text.
```
mdrxy: @ccurme lmk if this sounds better

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -300,9 +300,26 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
     defer to ``invoke()``/``ainvoke()``.
 
     - If True, will always bypass streaming case.
-    - If "tool_calling", will bypass streaming case only when the model is called
+    - If ``'tool_calling'``, will bypass streaming case only when the model is called
       with a ``tools`` keyword argument.
     - If False (default), will always use streaming case if available.
+
+    Useful when integrating with systems or models that are not designed to handle
```
eyurtsev: I think the explanation needs to be paraphrased a bit. The LangChain abstraction supports streaming in all of these cases, but there might be some issue with the implementation (e.g., maybe the specific **model instance**  does not support streaming yet, even though a different model instance from the same provider does).

----

The main reason for this flag is that code might be written using .stream() and a user may want to swap out a given model for another model for which the implementation does not support streaming properly.

PR #31796, opened 2025-07-01
langchain-fireworks[patch]: Add ruff bandit rules to linter

mdrxy:
- Add ruff bandit rules
- Address a s113 error


libs/partners/fireworks/langchain_fireworks/llms.py
```diff
@@ -148,7 +148,9 @@ def _call(
 
         # filter None values to not pass them to the http payload
         payload = {k: v for k, v in payload.items() if v is not None}
-        response = requests.post(url=self.base_url, json=payload, headers=headers)
+        response = requests.post(
+            url=self.base_url, json=payload, headers=headers, timeout=5
```
eyurtsev: This is a functional change that can break a user's application if it takes the provider more than 5 seconds to respond.

Let's push this up to ~20-60 seconds default, and expose as a parameter available via init

eyurtsev: Async path should have a matching param

PR #31813, opened 2025-07-01
core[path]: Use context manager for FileCallbackHandler

eyurtsev:
Recommend using context manager for FileCallbackHandler to avoid opening too many file descriptors

### DIFF
--- a/libs/core/langchain_core/callbacks/file.py
+++ b/libs/core/langchain_core/callbacks/file.py
@@ -5,8 +5,9 @@ from __future__ import annotations
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional, TextIO, cast
 
-from typing_extensions import override
+from typing_extensions import Self, override
 
+from langchain_core._api import warn_deprecated
 from langchain_core.callbacks import BaseCallbackHandler
 from langchain_core.utils.input import print_text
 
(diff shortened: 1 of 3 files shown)


libs/core/langchain_core/callbacks/file.py
```diff
@@ -5,87 +5,194 @@
...
         Args:
-            filename: The filename to write to.
-            mode: The mode to open the file in. Defaults to "a".
-            color: The color to use for the text. Defaults to None.
+            filename: Path to the output file.
+            mode: File open mode (e.g., ``'w'``, ``'a'``, ``'x'``). Defaults to ``'a'``.
+            color: Default text color for output. Defaults to ``None``.
         """
-        self.file = cast("TextIO", Path(filename).open(mode, encoding="utf-8"))  # noqa: SIM115
+        self.filename = filename
```
mdrxy, line 68: Should this be validated? E.g. sanitized to ensure it does not contain directory traversal characters (`..`, `/`, `\`). Or to restrict file writing to a hardcoded base directory.

eyurtsev: Not in this case.. you could argue that perhaps filename should be called path, but there's no reason to expect that the path is not a relative or an absolute path

mdrxy (approved):

PR #31818, opened 2025-07-01
langchain[patch]: Add bandit rules

mdrxy:
Integrate Bandit for security analysis, suppress warnings for specific issues, and address potential vulnerabilities such as hardcoded passwords and SQL injection risks. Adjust documentation and formatting for clarity.

### DIFF
--- a/libs/langchain/langchain/embeddings/cache.py
+++ b/libs/langchain/langchain/embeddings/cache.py
@@ -35,7 +35,7 @@ def _sha1_hash_to_uuid(text: str) -> uuid.UUID:
     but new applications should swap this out for a stronger hash function like
     xxHash, BLAKE2 or SHA-256, which are collision-resistant.
     """
-    sha1_hex = hashlib.sha1(text.encode("utf-8")).hexdigest()  # noqa: S324
+    sha1_hex = hashlib.sha1(text.encode("utf-8"), usedforsecurity=False).hexdigest()
     # Embed the hex string in `uuid5` to obtain a valid UUID.
     return uuid.uuid5(NAMESPACE_UUID, sha1_hex)
 
--- a/libs/langchain/langchain/embeddings/cache.py
+++ b/libs/langchain/langchain/embeddings/cache.py
@@ -26,14 +26,14 @@ NAMESPACE_UUID = uuid.UUID(int=1985)
 
 
 def _sha1_hash_to_uuid(text: str) -> uuid.UUID:
-    """Return a UUID derived from *text* using SHA‑1 (deterministic).
+    """Return a UUID derived from *text* using SHA-1 (deterministic).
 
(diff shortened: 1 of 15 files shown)


libs/langchain/langchain/memory/entity.py
```diff
@@ -319,62 +329,55 @@ def __init__(
...
 
     def delete(self, key: str) -> None:
-        query = f"""
-            DELETE FROM {self.full_table_name}
-            WHERE key = ?
-        """
-        with self.conn:
-            self.conn.execute(query, (key,))
+        """Deletes a key-value pair, safely quoting the table name."""
+        query = f'DELETE FROM "{self.full_table_name}" WHERE key = ?'  # noqa: S608
```
eyurtsev, line 371: For the places where we're using `# noqa: S608` right before we want to add a comment explaining why we're doing that so it doesn't look like we're just ignoring a security rule and so that other contributors don't cargo-cult

libs/langchain/tests/mock_servers/robot/server.py
```diff
@@ -201,5 +201,6 @@ def custom_openapi() -> dict[str, Any]:
 # This lets us prevent the "servers" configuration from being overwritten in
 # the auto-generated OpenAPI schema
 app.openapi = custom_openapi
+
 if __name__ == "__main__":
-    uvicorn.run(app, host="0.0.0.0", port=PORT)
+    uvicorn.run(app, host="0.0.0.0", port=PORT)  # noqa: S104
```
eyurtsev: Change 0.0.0.0 to 127.0.0.1 / localhost so requests on the local network don't make it through (even if it's testing).

Also this file is probably not used by anything and we could delete it in a bit

libs/langchain/langchain/embeddings/cache.py
```diff
@@ -26,16 +26,16 @@
...
+    Deterministic and fast, **but not collision-resistant**.
 
     A malicious attacker could try to create two different texts that hash to the same
     UUID. This may not necessarily be an issue in the context of caching embeddings,
     but new applications should swap this out for a stronger hash function like
-    xxHash, BLAKE2 or SHA‑256, which are collision-resistant.
+    xxHash, BLAKE2 or SHA-256, which are collision-resistant.
     """
-    sha1_hex = hashlib.sha1(text.encode("utf-8")).hexdigest()
+    sha1_hex = hashlib.sha1(text.encode("utf-8")).hexdigest()  # noqa: S324
```
eyurtsev: There's an arg we can pass to sha1 that says it's not used for cryptographic purposes -- better than the nosqa: S324

mdrxy:
Think we're ready here @eyurtsev

eyurtsev (approved):

PR #31825, opened 2025-07-02
fix(core): add type key when tracing in some cases

ccurme:
LangSmith assumes all message content blocks have a `type` field. Some provider-specific formats— e.g., Bedrock Converse's `cachePoint`— don't follow this constraint.

Here we catch one class of cases, where the dict has a single key that isn't `type`. We transform
```python
{
    "cachePoint": {"type": "default"},
}
```
to
```python
{
    "type": "cachePoint",
    "cachePoint": {"type": "default"},
}
```
just for tracing purposes, leaving chat model payloads unaltered.

Original trace: https://smith.langchain.com/public/2ac49a26-3012-4821-a876-147c184dd1b6/r
Updated trace: https://smith.langchain.com/public/0a523bcb-fca5-492d-ae9e-0c777f3bf870/r


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -125,20 +125,35 @@ def _format_for_tracing(messages: list[BaseMessage]) -> list[BaseMessage]:
...
-                ):
-                    if message_to_trace is message:
-                        message_to_trace = message.model_copy()
-                        # Also shallow-copy content
-                        message_to_trace.content = list(message_to_trace.content)
-
-                    message_to_trace.content[idx] = (  # type: ignore[index]  # mypy confused by .model_copy
-                        convert_to_openai_image_block(block)
-                    )
+                if isinstance(block, dict):
```
eyurtsev, line 129: A little bit more inline documentation about what's going on would be helpful

Issue #31808, opened 2025-07-01
bug in tool schema generation - missing field description for fields with pydantic model type


navalnica:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

code snippet to reproduce error:

```python
import json
import os

import langchain
from dotenv import load_dotenv
from langchain_openai import AzureChatOpenAI
from pydantic import BaseModel, Field

print(f'langchain version: {langchain.__version__}')
print()

load_dotenv('.env')
LLM_HOST = os.environ['LLM_HOST']
LLM_API_KEY = os.environ['LLM_API_KEY']


class Foo(BaseModel):
    value: str = Field(default='a', description='Foo.value field description')


class Bar(BaseModel):
    """Bar docsting"""

    value: str = Field(default='b', description='Bar.value field description')


class ToolInputs(BaseModel):
    foo: Foo = Field(description='foo field description')
    bar: Bar = Field(description='bar field description')


tool_inputs = ToolInputs(foo=Foo(), bar=Bar())

print(' tool_inputs.model_json_schema() '.center(70, '='))
print()
print('here we produce a schema directly from a pydantic model instance. it works as expected:')
print('in the produced schema we see that both `foo` and `bar` fields have field descriptions.')
print('`Bar` model also has a docstring')
print()
print(json.dumps(tool_inputs.model_json_schema(), indent=2))
print()


def tool_a(inputs: ToolInputs):
    """
    Tool A description
    """
    return f'Tool A called with inputs: {inputs}'


llm = AzureChatOpenAI(
    model='gpt-4o-2024-11-20',
    temperature=0.0,
    azure_endpoint=LLM_HOST,
    api_key=LLM_API_KEY,
    api_version='2024-08-01-preview',
)
llm_with_tools = llm.bind_tools([tool_a])

print(' llm_with_tools.kwargs '.center(70, '='))
print()
print('here we access tool schema through a langchain LLM instance. it does not work as expected:')
print('in the produced schema we see that the `foo` field does not have any description,')
print('while `bar` field has a description corresponding to a Bar model docstring')
print()
[...]

PR #31826, opened 2025-07-02
fix(core): Pydantic nested model field descriptions in tool schema generation

VedankPurohit:
## Description
This Pull Request fixes a bug in LangChain's tool schema generation logic where descriptions for fields containing nested Pydantic models were not correctly handled. This issue, reported in #31808, led to inconsistent or missing descriptions in the tool specifications provided to Large Language Models (LLMs).

### Problem Statement (from #31808):
When a Pydantic model is used to define the arguments for a LangChain tool, the description provided for a field containing a nested Pydantic model is not handled correctly.
The behavior was inconsistent:
- If the nested Pydantic model had a `__doc__` string, that docstring was used as the description, overwriting the description from the parent field.
- If the nested Pydantic model did not have a `__doc__` string, the description from the parent field was completely ignored, and the field ended up with no description at all in the final tool schema.
The desired behavior is that if a nested model does not have a docstring, the system should fall back to using the description provided in the parent model's `Field`.

## Changes Made
The core of the fix is in the `_get_schema_from_model` utility function within `libs/core/langchain_core/utils/function_calling.py`. This function has been modified to implement the desired fallback logic:
1.  **Prioritize Nested Model Docstring:** If a nested Pydantic model has a `__doc__` string, its content is used as the description for that field in the schema. This maintains existing behavior where a docstring explicitly defines the model's purpose.
2.  **Fallback to Parent Field Description:** If a nested Pydantic model does *not* have a `__doc__` string, the `description` provided in the parent `Field` definition is now correctly used as a fallback. This ensures that descriptions are not lost when nested models do not have their own docstrings.
3.  **Recursive Handling:** The logic for resolving nested model definitions (`$defs` for Pydantic v2 and `definitions` for Pydantic v1) has been updated to correctly pass the parent `Field`'s description during recursion. This ensures the fallback mechanism works for arbitrarily deep nesting of Pydantic models.

This PR directly addresses and resolves the issue described in #31808, ensuring accurate and consistent tool schema generation for Pydantic models with nested fields.

### DIFF
--- a/libs/core/tests/unit_tests/utils/test_function_calling.py
+++ b/libs/core/tests/unit_tests/utils/test_function_calling.py
@@ -1209,3 +1209,63 @@ def test_convert_to_openai_function_json_schema_missing_title_includes_schema() 
     }
     with pytest.raises(ValueError, match="my_field"):
         convert_to_openai_function(schema_without_title)
+
+
+def test_nested_model_field_description_fallback() -> None:
+    """Test that field descriptions propagate to nested models without docstrings."""
+
+    class NestedModel(BaseModel):
+        value: str
+
+    class ParentModel(BaseModel):
+        """Parent model."""
+
+        nested: NestedModel = Field(description="A nested model field")
+
+    result = convert_to_openai_function(ParentModel)
+    # Assert nested field has the description from the parent field
+    assert (
+        result["parameters"]["properties"]["nested"].get("description")
+        == "A nested model field"
+    )
+
+
+def test_nested_model_preserves_own_docstring() -> None:
+    """Test that nested models with docstrings keep their own descriptions."""
+
+    class NestedModel(BaseModel):
+        """Nested model's own docstring."""
+
+        value: str
+
+    class ParentModel(BaseModel):
+        """Parent model."""
+
+        nested: NestedModel = Field(description="Parent's field description")
+
+    result = convert_to_openai_function(ParentModel)
+    # Nested model should keep its own docstring, not use field description
+    assert (
+        result["parameters"]["properties"]["nested"].get("description")
+        == "Nested model's own docstring."
+    )
+
+
+def test_nested_model_list_not_processed() -> None:
+    """Test that List[NestedModel] fields are not processed for description fallback."""
+
+    class NestedModel(BaseModel):
+        value: str
+
+    class ParentModel(BaseModel):
+        """Parent model."""
+
+        nested_list: list[NestedModel] = Field(description="A list of nested models")
+
+    result = convert_to_openai_function(ParentModel)
+    # The list field should not have description applied to items
+    # This test ensures container types don't trigger the fallback logic
+    assert "properties" in result["parameters"]
+    # The nested_list field should exist but not have a description propagated
+    # to its items (the field itself may have description from array handling)
+    assert "nested_list" in result["parameters"]["properties"]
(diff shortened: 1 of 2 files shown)


mdrxy:
@VedankPurohit can you resolve these as appropriate and ensure testing passes in CI?

VedankPurohit:
> @VedankPurohit can you resolve these as appropriate and ensure testing passes in CI?

As I didn't see any messages in this pr I assumed the Approch itself was somehow flawed and I got busy thats why I couldn't fix this errors.

Do you think the Approch is correct? If yes then I could try and resolve this issue and then hopefully we can get this changes merged

Tanmay-24:
We should be good to merge this now!!

Tanmay-24:
@mdrxy Can you check the PR now?

mdrxy:
The code accesses `schema["properties"][field_name]` without verifying these keys exist. This could cause exceptions in edge cases.

Please add guards:

```python
if (
    nested_schema.get("description")
    and "properties" in schema
    and field_name in schema["properties"]
):
    schema["properties"][field_name]["description"] = nested_schema["description"]
```

The Pydantic v1 path is missing the `hasattr` check that exists in the v2 path. For generic types like `list[Model]`, `field.type_` won't have `__name__`:

```python
if (
    hasattr(field.type_, "__name__")  # Add this
    and field.type_.__name__ in schema["definitions"]
    ...
```

Per project guidelines, every bugfix must be covered by unit tests. Please add tests for:
- The `field_description` fallback behavior (nested model without docstring uses parent field description)
- Nested models with their own docstrings (should use docstring, not field description)
- Edge cases like `list[NestedModel]` being correctly skipped

VedankPurohit:
@mdrxy Thanks for the review! I've addressed all three points in commit 945863c18a:

Added guards for dictionary access - Both Pydantic v1 and v2 paths now check "properties" in schema and field_name in schema["properties"] before accessing

Added hasattr check for Pydantic v1 - Added hasattr(field.type_, "__name__") to match the v2 path

Added unit tests covering:

test_nested_model_field_description_fallback - nested model without docstring uses parent field description
test_nested_model_preserves_own_docstring - nested model with docstring keeps its own description
test_nested_model_list_not_processed - list[NestedModel] is correctly skipped

libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -1209,3 +1209,63 @@ def test_convert_to_openai_function_json_schema_missing_title_includes_schema()
...
+        """Parent model."""
+
+        nested: NestedModel = Field(description="Parent's field description")
+
+    result = convert_to_openai_function(ParentModel)
+    # Nested model should keep its own docstring, not use field description
+    assert (
+        result["parameters"]["properties"]["nested"].get("description")
+        == "Nested model's own docstring."
+    )
```
mdrxy, line 1251: @sydney-runkle can you confirm that this is intended behavior?

mdrxy:
Thanks for the follow-up. I checked this against current `master`, and the original issue no longer reproduces.

Current `master` already preserves the parent field’s explicit `Field(description=...)` when dereferencing nested Pydantic models, including cases where the nested model has its own docstring. That also matches the expected precedence discussed in #31808: explicit field descriptions should take priority over model docstrings.

Given that, I don’t think this PR is still needed in its current form. It’s also currently conflicting with `master`, and one of the proposed expectations — using the nested model docstring instead of the parent field description — appears to differ from the desired behavior.

I’m going to close this as superseded by the current implementation. Thanks again for working on it.

PR #31858, opened 2025-07-03
text-splitters: Ruff autofixes

cbornet:
Auto-fixes from ruff with rule `ALL`


mdrxy (approved):
LGTM, but why AbstractSet?

cbornet:
This is rule PYI025: https://docs.astral.sh/ruff/rules/unaliased-collections-abc-set-import

PR #31859, opened 2025-07-03
langchain[patch]: harden xml parser for xmloutput agent

eyurtsev:
Harden the default implementation of the XML parser for the agent


mdrxy (approved):
I think it looks good, but for an extreme approach see #31890

mdrxy (approved):

PR #30950, opened 2025-04-21
feat(docs): add PGVectorStore

dishaprakash:
Thank you for contributing to LangChain!

-  **Adding documentation for PGVectorStore**: 
docs: Adding documentation for the new PGVectorStore as a part of langchain-postgres

-  **Add docs**: The notebook for PGVectorStore is now added to the directory `docs/docs/integrations`.
As a part of this change, we've also updated the VectorStore features table and VectorStoreTabs

PR #31860, opened 2025-07-03
docs: migrate from PGVector to PGVectorStore

### LINKED ISSUE #31824: [DOCUMENTATION] PGVector is deprecated since 0.0.14, migrate the tutorial and other langchain docs to PGVectorStore
The langchain-postgres team has declared that they are migrating into `PGVectorStore` and `PGVector` is currently deprecated: 

https://github.com/langchain-ai/langchain-postgres/blob/18b1bcdb75ed152da717e3d624e1ed822d17d60f/README.md?plain=1#L28-L31

Please update the docs in order to use the most up-to-date version. I started setting up my project with 0.0.14 by following the outdated documentation and now I need to migrate to the new class.

https://github.com/langchain-ai/langchain/blob/eb122945832eae9b9df7c70ccd8d51fcd7a1899b/docs/docs/integrations/vectorstores/pgvector.ipynb#L123

itz-Amethyst:
Description:

This Pull Request address a migration-related issue mentioned in this issue #31824 using the deprecated `PGVector` class. The example in the Jupyter notebook has been updated to use `PGVectorStore`, which is the recommended replacement after the deprecation of `PGVector`.

Changes Made:

-  Updated the vector store tables to be compatible with the synchronous version of the `PGVectorStore` engine.

-  Replaced integer IDs with `UUIDs` for document identifiers to prevent conflicts with the table’s default ID schema.

-  Revised examples in the notebook file to match with the migrated version.

Fixes #31824


mdrxy:
Is this different from #30950?

itz-Amethyst:
> Is this different from #30950?

I've checked the PR you mentioned, and here are the key differences I observed:

To have access and perform searches on metadata fields of documents, you must declare metadata columns when initializing the vectorstore table. Otherwise, you can't apply any set of filters defined in the notebook (or generally). If we take the "location" column as an instance, you’ll get an error like:
```bash
ProgrammingError: (sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class 'asyncpg.exceptions.UndefinedColumnError'>: column "category" does not exist [SQL: SELECT "langchain_id", "content", "embedding", "langchain_metadata", cosine_distance("embedding", $1) as distance FROM "public"."vectorstore" WHERE category = ANY($2) ORDER BY "embedding" <=> $1 LIMIT $3; ]
```
A proof to that can be found here:
https://github.com/langchain-ai/langchain/blob/c4dd4a92791255466892adb44d904a87211a4709/docs/docs/integrations/vectorstores/pgvectorstore.ipynb#L611

If you look at the line, the author tried to filter a document based on *content* field in *metadata*, but in fact the filter query misinterpreted it with the actual content column that `PGVectorStore` creates, which is the `Document.page_content`.

As I said earlier, metadata fields are defined in the Document but not stored in the `pgvector table`, and that's why if you try to search on any other columns in the metadata while not defining them, you'll get error.

To my best knowledge, to enable metadata-based filtering in PGVectorStore, the fields must be declared in table columns.

There are also some other minor changes, but this is the most important one.

itz-Amethyst:
I realized I forgot to update the documentation to match with the migration to `PGVectorStore`, 
If you're okay with the current updates, I’ll go ahead and sync the documentation.

eyurtsev:
official docs here: https://github.com/langchain-ai/langchain/pull/30950

PR #31869, opened 2025-07-04
standard-tests: Add ruff rules PGH

cbornet:
See https://docs.astral.sh/ruff/rules/#pygrep-hooks-pgh


mdrxy:
+1, was going to suggest this

PR #31873, opened 2025-07-04
langchain: Add ruff rules EM

cbornet:
All auto-fixes.
See https://docs.astral.sh/ruff/rules/#flake8-errmsg-em


libs/langchain/langchain/agents/tool_calling_agent/base.py
```diff
@@ -90,11 +90,13 @@ def magic_function(input: int) -> int:
         prompt.input_variables + list(prompt.partial_variables)
     )
     if missing_vars:
-        raise ValueError(f"Prompt missing required variables: {missing_vars}")
+        msg = f"Prompt missing required variables: {missing_vars}"
+        raise ValueError(msg)
 
     if not hasattr(llm, "bind_tools"):
+        msg = "This function requires a .bind_tools method be implemented on the LLM."
```
mdrxy: Nit, we should use `bind_tools()` instead of `.bind_tools`

mdrxy (approved):
LGTM

PR #31874, opened 2025-07-04
langchain: Add ruff rule RUF

cbornet:
All auto-fixes
See https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf


libs/langchain/langchain/memory/entity.py
```diff
@@ -348,7 +348,7 @@ def _create_table_if_not_exists(self) -> None:
     def get(self, key: str, default: Optional[str] = None) -> Optional[str]:
         """Retrieves a value, safely quoting the table name."""
         # `?` placeholder is used for the value to prevent SQL injection
-        # noqa since we validate for malicious table/session names in `__init__`
+        # Ignore S608 since we validate for malicious table/session names in `__init__`
```
cbornet, line 351: Interestingly, ruff was seeing this comment as a `# noqa` command.

PR #31875, opened 2025-07-04
langchain: Add ruff rule RET

cbornet:
All auto-fixes
See https://docs.astral.sh/ruff/rules/#flake8-return-ret


mdrxy (approved):
LGTM

cbornet:
Lots of conflicts. Let me rebase.

PR #31877, opened 2025-07-05
core: Ruff preview fixes

cbornet:
Auto-fixes from `uv run ruff check --fix --unsafe-fixes --preview`


libs/core/langchain_core/language_models/llms.py
```diff
@@ -119,7 +119,7 @@ def _before_sleep(retry_state: RetryCallState) -> None:
     # 4 seconds, then up to 10 seconds, then 10 seconds afterwards
     retry_instance: retry_base = retry_if_exception_type(error_types[0])
     for error in error_types[1:]:
-        retry_instance = retry_instance | retry_if_exception_type(error)
+        retry_instance |= retry_if_exception_type(error)
```
eyurtsev, line 122: Is this an auto fix as well?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -544,7 +544,7 @@ def from_template_file(
         Returns:
             A new instance of this class.
         """
-        template = Path(template_file).read_text()
+        template = Path(template_file).read_text(encoding="utf-8")
```
eyurtsev: This is a breaking change we can roll this out as part of 0.4 cc @ccurme / @sydney-runkle 

Ideally change default encoding to utf-8 in every instance

libs/core/langchain_core/prompts/chat.py
```diff
@@ -1053,9 +1053,9 @@ def validate_input_variables(cls, values: dict) -> Any:
                 if message.variable_name not in input_types:
                     input_types[message.variable_name] = list[AnyMessage]
         if "partial_variables" in values:
-            input_vars = input_vars - set(values["partial_variables"])
+            input_vars -= set(values["partial_variables"])
```
eyurtsev, line 1057: Is this an auto fix as well? Why the change?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -1338,11 +1338,11 @@ def _create_template_from_message_type(
     Raises:
         ValueError: If unexpected message type.
     """
-    if message_type in ("human", "user"):
+    if message_type in {"human", "user"}:
         message: BaseMessagePromptTemplate = HumanMessagePromptTemplate.from_template(
             template, template_format=template_format
         )
-    elif message_type in ("ai", "assistant"):
+    elif message_type in {"ai", "assistant"}:
```
eyurtsev, line 1346: It probably doesn't matter much but do we know whether this is overall faster or slower? It requires both creating the data structure (unless it gets cached by the interpreter) and checking for membership.

libs/core/langchain_core/language_models/llms.py
cbornet: Yes, it’s an auto-fix

libs/core/langchain_core/prompts/chat.py
cbornet: Yes it’s an auto-fix

libs/core/langchain_core/language_models/llms.py
cbornet: An "unsafe" one since the first syntax creates a new object each time whereas `|=` may mutate the initial object.
Here it's OK to mutate `retry_instance`.

libs/core/langchain_core/prompts/chat.py
cbornet: This is rule PLR6201 : https://docs.astral.sh/ruff/rules/literal-membership/
See also https://docs.python.org/3/whatsnew/3.2.html#optimizations
> Python’s peephole optimizer now recognizes patterns such x in {1, 2, 3} as being a test for membership in a set of constants. The optimizer recasts the [set](https://docs.python.org/3/library/stdtypes.html#set) as a [frozenset](https://docs.python.org/3/library/stdtypes.html#frozenset) and stores the pre-built constant.
Now that the speed penalty is gone, it is practical to start writing membership tests using set-notation. This style is both semantically clear and operationally fast

PR #31879, opened 2025-07-05
langchain: Add ruff rules C4

cbornet:
All auto-fixes
See https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4


mdrxy (approved):
LGTM

PR #31885, opened 2025-07-06
langchain: Add ruff rules FBT

cbornet:
* Fixed for private functions and in tests
* Added noqa escapes for public functions
See https://docs.astral.sh/ruff/rules/#flake8-boolean-trap-fbt


libs/langchain/langchain/chains/openai_functions/openapi.py
```diff
@@ -23,14 +23,6 @@
     from openapi_pydantic import Parameter
 
 
-def _get_description(o: Any, prefer_short: bool) -> Optional[str]:
```
mdrxy, line 26: is this unused? why remove?

mdrxy:
Why are we adding these rules only to add `noqa: FBT001,FBT002` for each occurrence?

cbornet:
We add `noqa` for public functions/methods because we can't change them in a backward compatible way.
For private functions, we fix by renaming.
This rule will protect from introducing new FBTs in public functions in the future.

libs/langchain/langchain/chains/openai_functions/openapi.py
mdrxy: Looks like it

mdrxy (approved):
LGTM

PR #31908, opened 2025-07-08
langchain: Add ruff rules B

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-bugbear-b

### DIFF
--- a/libs/langchain/langchain/__init__.py
+++ b/libs/langchain/langchain/__init__.py
@@ -29,10 +29,12 @@ def _warn_on_import(name: str, replacement: Optional[str] = None) -> None:
         warnings.warn(
             f"Importing {name} from langchain root module is no longer supported. "
             f"Please use {replacement} instead.",
+            stacklevel=3,
         )
     else:
         warnings.warn(
             f"Importing {name} from langchain root module is no longer supported.",
+            stacklevel=3,
         )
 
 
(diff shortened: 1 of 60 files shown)


libs/langchain/langchain/agents/openai_functions_agent/base.py
```diff
@@ -213,9 +215,7 @@ def return_stopped_response(
     @classmethod
     def create_prompt(
         cls,
-        system_message: Optional[SystemMessage] = SystemMessage(
-            content="You are a helpful AI assistant.",
-        ),
+        system_message: Optional[SystemMessage] = _NOT_SET,  # type: ignore[assignment]
```
cbornet, line 218: Method args must be immutable: https://docs.astral.sh/ruff/rules/function-call-in-default-argument/

libs/langchain/langchain/__init__.py
```diff
@@ -29,10 +29,12 @@ def _warn_on_import(name: str, replacement: Optional[str] = None) -> None:
         warnings.warn(
             f"Importing {name} from langchain root module is no longer supported. "
             f"Please use {replacement} instead.",
+            stacklevel=3,
```
cbornet, line 32: I adjusted the stacklevels so thewarnings appear correctly on the first user facing function.

libs/langchain/langchain/chains/retrieval_qa/base.py
```diff
@@ -320,15 +319,6 @@ class VectorDBQA(BaseRetrievalQA):
     search_kwargs: dict[str, Any] = Field(default_factory=dict)
     """Extra search args."""
 
-    @model_validator(mode="before")
-    @classmethod
-    def raise_deprecation(cls, values: dict) -> Any:
```
cbornet, line 325: removed this redundant deprecation since `VectorDBQA` is already decorated by `@deprecated`

mdrxy (approved):

PR #31911, opened 2025-07-08
langchain: Use pytest.raises and pytest.fail to handle exceptions in tests

cbornet:


libs/langchain/tests/unit_tests/output_parsers/test_datetime_parser.py
```diff
@@ -39,11 +42,5 @@ def test_datetime_output_parser_parse() -> None:
...
     # Test invalid input
-    try:
-        sleep(0.001)
-        datestr = date.strftime(parser.format)
-        result = parser.parse(datestr)
-        assert result == date
-        assert False, "Should have raised AssertionError"
-    except AssertionError:
-        pass
+    with pytest.raises(OutputParserException):
```
cbornet, line 44: I believe the previous test was not correct.

mdrxy (approved):
LGTM

PR #31917, opened 2025-07-08
ruff: add bugbear across packages

mdrxy:
WIP, other packages will get in next PRs


mdrxy:
@cbornet I think the merge might have messed up some of the stack levels, can you let me know if any need to be changed back?

cbornet:
> @cbornet I think the merge might have messed up some of the stack levels, can you let me know if any need to be changed back?

Yes they were all overriden.
Can you set back all the values that were different than `2` ?

PR #31926, opened 2025-07-08
multiple: standard outputs

ccurme:


libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -7,6 +7,93 @@
...
+    title: NotRequired[str]
+    """Source title."""
+
+    cited_text: NotRequired[str]
+    """Text from the source that is being cited."""
+
+    start_index: NotRequired[int]
+    """Start index of the response text for which the annotation applies."""
+
+    end_index: NotRequired[int]
```
eyurtsev, line 46: Perhaps change end_index into limit_index and use a convention where it's non inclusive?

`end` is often ambiguous enough that it causes off by 1 inconsistencies in implementation.

The doc-string should also say that this is non-inclusive.

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -7,6 +7,93 @@
...
+    """End index of the response text for which the annotation applies."""
+
+
+class NonStandardAnnotation(TypedDict, total=False):
+    """Provider-specific annotation format."""
+
+    type: Literal["non_standard_annotation"]
+    """Type of the content block."""
+    value: dict[str, Any]
```
eyurtsev, line 55: Is it guaranteed to be a dict?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -7,6 +7,93 @@
...
+    """Provider-specific annotation data."""
+
+
+class TextContentBlock(TypedDict, total=False):
```
eyurtsev: This is using both `total=False` and `NotRequired` which is not correct I think

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -7,6 +7,93 @@
...
+    """Content block for text output."""
+
+    type: Literal["text"]
+    """Type of the content block."""
+    text: str
+    """Block text."""
```
eyurtsev, line 65: doc-string paraphrase

libs/core/langchain_core/messages/content_blocks.py
eyurtsev: Is `Required` available via typing_extensions?

https://docs.python.org/3/library/typing.html#typing.Required

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -68,6 +155,28 @@ class IDContentBlock(TypedDict):
 _DataContentBlockAdapter: TypeAdapter[DataContentBlock] = TypeAdapter(DataContentBlock)
 
 
+# Non-standard
+class NonStandardContentBlock(TypedDict, total=False):
+    """Content block provider-specific data.
```
eyurtsev, line 160:
```suggestion
    """Content block representing provider-specific data.
```

libs/partners/openai/langchain_openai/chat_models/_compat.py
```diff
@@ -248,3 +273,279 @@ def _convert_from_v03_ai_message(message: AIMessage) -> AIMessage:
         },
         deep=False,
     )
+
+
+# v1 / Chat Completions
+def _convert_to_v1_from_chat_completions(message: AIMessage) -> AIMessage:
```
eyurtsev, line 279: If the code is doing mutation in place could we make a convention where we remove the return value?

I think it's standard for python to have no return to signify mutation in place

eyurtsev (commented):
Quick passthrough

ccurme:
Messed up history, closing in favor of https://github.com/langchain-ai/langchain/pull/31964

PR #31930, opened 2025-07-09
chore(langchain): select ALL rules with exclusions

cbornet:


libs/langchain/langchain/agents/openai_functions_multi_agent/base.py
```diff
@@ -72,7 +72,6 @@ def _parse_ai_message(message: BaseMessage) -> Union[list[AgentAction], AgentFin
                 del _tool_input["action_name"]
             function_name = tool_schema["action_name"]
 
-            # HACK HACK HACK:
```
mdrxy, line 75: We should keep these

cbornet: What are they for ?
Can we change to something that would not trigger the  FIX rule ?

mdrxy: Sure; I'm open to another string that wouldn't trigger

mdrxy (changes requested):

libs/langchain/langchain/agents/openai_functions_multi_agent/base.py
cbornet: done

mdrxy (approved):
LGTM

PR #31935, opened 2025-07-09
text-splitters: Add ruff rules FBT

cbornet:
See [flake8-boolean-trap (FBT)](https://docs.astral.sh/ruff/rules/#flake8-boolean-trap-fbt)


mdrxy (approved):
LGTM

PR #31967, opened 2025-07-10
feat: standard outputs

ccurme:


libs/core/langchain_core/messages/base.py
```diff
@@ -23,7 +24,7 @@ class BaseMessage(Serializable):
     Messages are the inputs and outputs of ChatModels.
     """
 
-    content: Union[str, list[Union[str, dict]]]
+    content: Union[str, list[Union[str, ContentBlock, dict]]]
```
mdrxy: this is now outdated right?

libs/partners/openai/tests/integration_tests/chat_models/test_responses_api.py
```diff
@@ -141,13 +146,15 @@ async def test_web_search_async() -> None:
         assert tool_output["type"] == "web_search_call"
 
 
-@pytest.mark.flaky(retries=3, delay=1)
-def test_function_calling() -> None:
+@pytest.mark.default_cassette("test_function_calling.yaml.gz")
+@pytest.mark.vcr
+@pytest.mark.parametrize("output_version", ["v0", "responses/v1", "v1"])
```
mdrxy, line 151: did you mean to remove `responses/v1`?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -3913,21 +3987,34 @@ def _advance(output_idx: int, sub_idx: Optional[int] = None) -> None:
...
-            {
-                # langchain-core uses the `index` key to aggregate text blocks.
-                "summary": [
-                    {"index": chunk.summary_index, "type": "summary_text", "text": ""}
-                ],
-                "index": current_index,
-                "type": "reasoning",
-            }
-        )
+        if output_version in ("v0", "responses/v1"):
```
mdrxy, line 3995: same here, `'v1'` instead of ``responses/*``?

libs/core/langchain_core/messages/ai.py
```diff
@@ -196,6 +197,79 @@ def lc_attributes(self) -> dict:
...
+
+        .. code-block:: python
+
+            from langchain.chat_models import init_chat_model
+
+            llm = init_chat_model("...", output_version="v1")
+
+        otherwise, does best-effort parsing to standard types.
+        """
+        blocks: list[types.ContentBlock] = []
```
eyurtsev: I think you can short circuit the logic in some cases to make it easier to think about things


```
if isinstance(self.content, str):
  return {
     "type": "text", 
     "text": self.content
  }

# Then we know that self content has to be a list, so no need for an if statement
for item in self.content:
  ....

```

ccurme: we still want to add blocks from tool calls for string content, though.

libs/core/langchain_core/messages/ai.py
```diff
@@ -196,6 +197,79 @@ def lc_attributes(self) -> dict:
...
+
+        elif isinstance(self.content, list):
+            for item in self.content:
+                if isinstance(item, str):
+                    blocks.append({"type": "text", "text": item})
+
+                elif isinstance(item, dict):
+                    item_type = item.get("type")
+                    if item_type == "text":
+                        blocks.append(cast("types.TextContentBlock", item))
```
eyurtsev: The casts aren't needed here since they don't affect anything on the user's end.

The user only sees ```def beta_content(self) -> list[types.ContentBlock]:``` this type signature.

The logic here only needs to reshape the actual data to match the types.ContentBLock convention

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -68,6 +155,28 @@ class IDContentBlock(TypedDict):
 _DataContentBlockAdapter: TypeAdapter[DataContentBlock] = TypeAdapter(DataContentBlock)
 
 
+# Non-standard
+class NonStandardContentBlock(TypedDict, total=False):
```
eyurtsev: Why is total=False? Do we not expect type to be present?

libs/partners/openai/langchain_openai/chat_models/_compat.py
```diff
@@ -252,3 +260,290 @@ def _convert_from_v03_ai_message(message: AIMessage) -> AIMessage:
...
+        return new_ann
+
+    elif annotation_type == "non_standard_annotation":
+        return annotation["value"]
+
+    else:
+        return dict(annotation)
+
+
+def _implode_reasoning_blocks(blocks: list[dict[str, Any]]) -> Iterable[dict[str, Any]]:
```
eyurtsev, line 516: Perhaps avoid `i` and `n`? It's not looking like conventional python code. `i` and `n` are more typical of c code

i -> idx, `n` -> `num_blocks` or `num_total` or `total`

libs/core/langchain_core/messages/ai.py
eyurtsev: then perhaps do something like:


```python
content = [self.content] if isistance(self.content, str) else self.content
```

it should still have the effect of eliminating an if statement and denesting by one level

ccurme: thanks!

libs/core/langchain_core/messages/content_blocks.py
ccurme: forgot to remove this one, updated.

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -311,6 +311,18 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
     does not properly support streaming.
     """
 
+    output_version: str = "v0"
```
mdrxy, line 350: see https://github.com/langchain-ai/langchain/pull/31977

libs/partners/openai/tests/integration_tests/chat_models/test_responses_api.py
ccurme: responses/v1 remains a valid option for ChatOpenAI. IMO we allow use of this field for provider packages to opt-into their own breaking changes. responses/v1 was added for that purpose. This is why the field is typed as a `str` in core instead of a literal.

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -4,10 +4,97 @@
...
+    """Tool call ID."""
+
+
+# Reasoning
+class ReasoningContentBlock(TypedDict):
+    """Content block for reasoning output."""
+
+    type: Literal["reasoning"]
+    """Type of the content block."""
+    reasoning: NotRequired[str]
```
mdrxy, line 413: Why is this optional?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -4,10 +4,97 @@
...
+    text: str
+    """Block text."""
+    annotations: NotRequired[
+        list[Union[UrlCitation, DocumentCitation, NonStandardAnnotation]]
+    ]
+    """Citations and other annotations."""
+
+
+# Tool calls
+class ToolCallContentBlock(TypedDict):
```
mdrxy: Aren't we missing `name`?

mdrxy: Or is name going under a tool call in `tool_calls`?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -1,110 +1,785 @@
...
-class URLContentBlock(BaseDataContentBlock):
-    """Content block for data from a URL."""
+    url: NotRequired[str]
+    """URL of the image."""
 
-    type: Literal["image", "audio", "file"]
+    base64: NotRequired[str]
+    """Data as a base64 string."""
+
+    index: NotRequired[int]
```
mdrxy: will others need `index`?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -1,110 +1,785 @@
...
+        "image",
+        "video",
+        "file",
+        "text-plain",
+    ) and any(
+        key in block
+        for key in (
+            "url",
+            "base64",
+            "file_id",
```
mdrxy, line 754: each block has `file_id`, should this be removed so that this is a set of only non-required keys? if so we should consider adding `title` and `context`

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -1,110 +1,785 @@
...
+            "text",
+            "source_type",  # backwards compatibility
+        )
+    )
 
 
 def convert_to_openai_image_block(content_block: dict[str, Any]) -> dict:
     """Convert image content block to format expected by OpenAI Chat Completions API."""
-    if content_block["source_type"] == "url":
+    if "url" in content_block:
```
mdrxy: should this also check the input `content_block` for `type = "image"`?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -117,36 +792,38 @@ def convert_to_openai_data_block(block: dict) -> dict:
         formatted_block = convert_to_openai_image_block(block)
 
     elif block["type"] == "file":
-        if block["source_type"] == "base64":
-            file = {"file_data": f"data:{block['mime_type']};base64,{block['data']}"}
+        if "base64" in block or block.get("source_type") == "base64":
```
mdrxy, line 792: same here, check for `type in set(DataContentBlock.type)`?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -1,110 +1,785 @@
...
+
+        This represents a request to call the tool named "foo" with arguments {"a": 1}
+        and an identifier of "123".
+    """
+
+    name: str
+    """The name of the tool to be called."""
+    args: dict[str, Any]
+    """The arguments to the tool call."""
+    id: Optional[str]
```
mdrxy, line 210: should we be using `Optional` or `NotRequired`?

libs/core/langchain_core/messages/v1.py
```diff
@@ -0,0 +1,566 @@
...
+        tool_call_id: ID of the tool call this message responds to.
+        content: The result content from tool execution.
+        artifact: Optional app-side payload not intended for the model.
+        status: Execution status ("success" or "error").
+        type: Message type identifier, always "tool".
+    """
+
+    id: str
+    tool_call_id: str
+    content: list[dict[str, Any]]
```
mdrxy, line 540: why isn't this `list[types.ContentBlock]`?

libs/core/langchain_core/messages/v1.py
```diff
@@ -0,0 +1,566 @@
...
+    ):
+        """Initialize a system message.
+
+        Args:
+            content: System instructions as string or list of content blocks.
+            id: Optional unique identifier for the message.
+        """
+        self.id = _ensure_id(id)
+        if isinstance(content, str):
+            self.content = [{"type": "text", "text": content}]
```
mdrxy, line 511: for each case where we have `self.content = [{"type": "text", "text": content}]`, could we use the approach where we construct the TypedDict and then convert to dict? e.g.:

```suggestion
        if isinstance(content, str):
            content_typed = TextContentBlock(type="text", text=content)
            self.content = [content_typed.asdict()]
```

libs/core/langchain_core/messages/v1.py
```diff
@@ -0,0 +1,566 @@
...
+        self._invalid_tool_calls: list[types.InvalidToolCall] = []
+
+    @property
+    def text(self) -> Optional[str]:
+        """Extract all text content from the AI message as a string."""
+        text_blocks = [block for block in self.content if block["type"] == "text"]
+        if text_blocks:
+            return "".join(block["text"] for block in text_blocks)
+        return None
+
```
mdrxy, line 144: I think a property for `reasoning` is missing here, no?

mdrxy: It's present for `Chunk`

libs/core/langchain_core/messages/v1.py
```diff
@@ -0,0 +1,566 @@
...
+            content: Message content as string or list of content blocks.
+            id: Optional unique identifier for the message.
+        """
+        self.id = _ensure_id(id)
+        if isinstance(content, str):
+            self.content = [{"type": "text", "text": content}]
+        else:
+            self.content = content
+
+    def text(self) -> str:
```
mdrxy, line 472: doesn't this need `@property`?

libs/core/langchain_core/messages/v1.py
```diff
@@ -0,0 +1,566 @@
...
+            content: System instructions as string or list of content blocks.
+            id: Optional unique identifier for the message.
+        """
+        self.id = _ensure_id(id)
+        if isinstance(content, str):
+            self.content = [{"type": "text", "text": content}]
+        else:
+            self.content = content
+
+    def text(self) -> str:
```
mdrxy, line 515: same here, `@property`?

libs/core/langchain_core/messages/utils.py
```diff
@@ -293,6 +299,45 @@ def _create_message_from_message_type(
     return message
 
 
+def _convert_from_v1_message(message: MessageV1) -> BaseMessage:
```
mdrxy, line 377: is this still a WIP? the v0 AIMessage will have `content` with dicts in the shape of the new content blocks - is there a plan to convert back to the non-std-output (or is that even possible?) don't think so, but wanted to ask for clarity

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -220,6 +222,23 @@ def _format_ls_structured_output(ls_structured_output_format: Optional[dict]) ->
     return ls_structured_output_format_dict
 
 
+def _convert_to_v1(message: AIMessage) -> AIMessageV1:
+    """Best-effort conversion of a V0 AIMessage to V1."""
+    if isinstance(message.content, str):
+        content: list[types.ContentBlock] = []
+        if message.content:
+            content = [{"type": "text", "text": message.content}]
```
mdrxy, line 230: perhaps the same `asdict` conversion here?

PR #31968, opened 2025-07-11
fix: support dict input with 'messages' key in convert_to_messages

Krish0909:
## Description
Fixes issue where `convert_to_messages()` fails when passed a dict with 'messages' key, which is the format used by LangGraph agent invocations.

## Changes Made
- Add support for `{"messages": [...]}` input format in `convert_to_messages()`
- Add test coverage for the new functionality

## Testing
- [x] Manual testing confirms fix works
- [x] Test case added and passes
- [x] Original issue #31800 is resolved

Fixes #31800


ccurme (commented):
I think this change is unrelated to the [original issue](https://github.com/langchain-ai/langchain/issues/31800).

PR #31977, opened 2025-07-11
docs: improve `output_version` description

mdrxy:


ccurme (commented):
IMO we keep support for customization in provider packages. I think this change as-is would break typing in langchain-openai. Also, why `Optional`? What's denoted by None vs. v0?

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -311,14 +311,19 @@ class BaseChatModel(BaseLanguageModel[BaseMessage], ABC):
     does not properly support streaming.
     """
 
-    output_version: str = "v0"
+    output_version: Optional[Literal["v0", "v1"]] = Field(default="v0")
```
eyurtsev: Don't use optional as post initialization this field should never be a None. Theres no need to use Field in this case

sydney-runkle (approved):

PR #32001, opened 2025-07-13
feat(core): enhance string input parsing to handle JSON-like inputs

ChrisCoder9000:
core: Add JSON detection and parsing for string tool inputs that appear to be JSON objects. 
This improves user experience by allowing tools to accept JSON data passed as strings, 
while maintaining backward compatibility with existing string input behavior. The enhancement 
includes proper schema validation for both Pydantic v1 and v2 models, handles injected 
tool call IDs, and gracefully falls back to original behavior when JSON parsing fails.

Twitter: christiannonis


mdrxy:
Could you please expand on the intended use case/illuminate with some examples? Also, please add tests and update to pass CI checks such as linting.

mdrxy:
Closing unless @ChrisCoder9000 has follow up to my previous comment

PR #32002, opened 2025-07-13
feat(core): Adding a fake LLM with tool calling capabilities for easy integration with LangGraph

sijil-salim:
Added a new Fake LLM in lines of FakeListChatModel to return tool calls so that it is directly compatible with LangGraph. This fake LLM can be used for testing LangGraph workflows


sijil-salim:
@eyurtsev Hi, I pushed a change for adding a fake llm with tool building capabilities yesterday. However, the deployment got skipped and got a message from verbal with links which upon opening shows Vercel error page. Can you please help me with what might have gone wrong and how I can rectify it

eyurtsev (changes requested):
Will review a bit more carefully later, but we don't want another implementation. Instead one of the existing implementations should be enhanced properly

ccurme (commented):
Thanks for this. Closing because I believe we now have an implementation in https://github.com/langchain-ai/langchain/blob/master/libs/langchain_v1/tests/unit_tests/agents/model.py.

Issue #32003, opened 2025-07-13
Langchain Core Unnecessarily Pins Packaging


ntjohnson1:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

Error on installation:

```bash
uv pip install "langchain_core>0.3.0" "packaging==25.0.0"
```

### Error Message and Stack Trace (if applicable)

 No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of langchain-core are available:
--snip big printout since nothing is compatible--
And because you require langchain-core>0.3.0 and packaging==25.0.0, we can conclude that your requirements are unsatisfiable.

### Description

The primary `langchain` package puts the packaging dependency inside of its testing deps since that is the only place it appears to be used directly.

This came up when trying to install a variety of things so unclear what in the ecosystem is exactly forcing 25.0 but I still don't think langchain actually requires the pin to run.

### System Info

python -m langchain_core.sys_info

System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041
> Python Version:  3.12.11 (main, Jun 29 2025, 16:16:55) [Clang 20.1.4 ]

Package Information
-------------------
> langchain_core: 0.3.68b0
> langchain: 0.3.26
> langchain_community: 0.3.23
> langsmith: 0.3.45
> langchain_openai: 0.3.27
> langchain_text_splitters: 0.3.9

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp<4.0.0,>=3.8.3: Installed. No version info available.
> async-timeout<5.0.0,>=4.0.0;: Installed. No version info available.
> dataclasses-json<0.7,>=0.5.7: Installed. No version info available.
> httpx: 0.28.1
> httpx-sse<1.0.0,>=0.4.0: Installed. No version info available.
> jsonpatch<2.0,>=1.33: Installed. No version info available.
> langchain-anthropic;: Installed. No version info available.
> langchain-aws;: Installed. No version info available.
[...]

PR #32004, opened 2025-07-13
fix(core): unblock packaging dependency

ntjohnson1:
**Description:**
* Moves packaging to a test only dependency to avoid install collisions conceptually
* Bumps packaging to <26 (since tests and things run with 25), could alternatively unpin since its not clear from the original commit exactly why the upperbound was required
* Deduplicates the version specifier to match other projects in the monorepo (because I hit this when double checking things locally)
  * This ended up touching a few more things than expected and I can separate out if for any reason it is more controversial than it appears

**Issue:** #32003


libs/core/langchain_core/version.py
```diff
@@ -1,3 +1,3 @@
 """langchain-core version information and utilities."""
 
-VERSION = "0.3.68"
+__version__ = "0.3.68"
```
eyurtsev: Why has this change been made?

eyurtsev: This is a breaking API change -- we don't want to make this

ntjohnson1: It allowed the pyproject to determine the version automatically so that it didn't have to be tracked in two places but retained the import performance requirements. Not critical to the packaging pinning issues that was the motivation for the PR, so I reverted.

When I was testing it locally I changed the version in the pyproject but the imported version didn't change which was unexpected so I tried to centralize.

eyurtsev: OK we'll likely make this fix separately

eyurtsev: cc @sydney-runkle

eyurtsev:
Could you allow maintainers to update this PR?

libs/core/pyproject.toml
```diff
@@ -39,6 +38,7 @@ dev = [
     "grandalf<1.0,>=0.8",
 ]
 test = [
+    "packaging<26,>=23.2",
```
eyurtsev, line 41: It should not be moved into the `test` group. It's an explicit dependency of langchain_core

eyurtsev:
I'll take over and make a separate PR to unblock

PR #32017, opened 2025-07-14
fix(core): properly handle empty or missing arguments for `ToolCallChunk`

### LINKED ISSUE #32016: Failed to used non-argument tool call or MCP in stream mode
### Checked other resources

- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```python
from langgraph.graph import StateGraph, START, END
from langchain_core.runnables import RunnableConfig
from langgraph.types import Command

from langchain_core.messages import HumanMessage, AIMessage
from typing import List, Dict, Any
from typing import Annotated, Literal
from langgraph.graph.message import add_messages
import asyncio


class State(TypedDict):
    messages: Annotated[list, add_messages]


def init_llm():
    return ChatOpenAI(
        base_url="http://127.0.0.1:8000/v1", # mock local llm server
        model="qwen3-32b", # let's assume this is a valid model name
        api_key="none",
    )

def get_book_list() -> List[Dict[str, Any]]:
    """
    Returns a list of available books.

    Returns:
        List[Dict[str, Any]]: A list of books, each containing the book name, author.
    """
    # mock
    return [
        {"author": "abc", "name": "abc"},
        {"author": "def", "name": "def"},
        {"author": "ghi", "name": "ghi"},
        {"author": "jkl", "name": "jkl"},
        {"author": "mnq", "name": "mnq"}
    ]

async def agent_node(
    state: State, config: RunnableConfig
) -> Command[Literal["__end__"]]:
    # mcp_servers = config["configurable"]["mcp_servers"]
    # client = MultiServerMCPClient(mcp_servers)
    # tools = await client.get_tools()

    agent = create_react_agent(
        name="info-retrieval-agent",
        model=init_llm(),
        tools=[get_book_list],
        prompt="{CURRENT_TIME}\n\n You are a helpful e-book library assistant, you can answer the readers' question, such as `the number of books you have`.",
    )

    message = state.get("messages")[-1].content
    # e.g. how many books do you have?
    agent_input = {"messages": [HumanMessage(message)]}
    # raise error here, invalid tool
    result = await agent.ainvoke(input=agent_input, config={"recursion_limit": 5})

    return Command(
        update={"messages": [AIMessage(result["messages"][-1].content)]}, goto="__end__"
    )


async def main():
    builder = StateGraph(State)
    builder.add_node("agent", agent_node)
    builder.add_edge(START, "agent")
    builder.add_edge("agent", END)
    compiled_graph = builder.compile()
    user_input = {
        "messages": "how many books do you have",
    }

    # it will treat non-argument tool -> invalid tool which is not expected
    async for _, _, event_data in compiled_graph.astream(
        user_input,
        config={
            # "mcp_servers": {
            #     "mock": {
            #         "url": "http://127.0.0.1:3000/sse",
            #         "transport": "sse",
            #         "args": [],
            #         "command": [],
            #         "env": [],
            #     }
            # }
        },
        stream_mode=["messages", "updates"],
        subgraphs=True,
    ):
        pass

    # works fine
    # result = await compiled_graph.ainvoke(user_input)


if __name__ == "__main__":
    asyncio.run(main())
```

_No response_

### Description

* I’m using LangChain/LangGraph to build an agent application.
* I’ve created some MCP tools to enhance the agent’s capabilities.
* However, LangChain fails to handle tools with empty arguments properly: it incorrectly marks them as invalid, and no error message is generated.
* The bug only occurs in stream mode.

### System Info

System Information

> OS: Linux
> OS Version: Ubuntu
> Python Version: 3.12.11

Package Iinformation

> langchain_core: 0.3.68
> langchain: 0.3.26
> langchain_community: 0.3.27
> langsmith: 0.4.4
> langgraph_sdk: 0.1.72

amznero:
Description: Streaming tool call chunks: if `chunk["args"] == None`, the tool call is incorrectly marked as an `InvalidToolCall`.

Fixes #32016


mdrxy:
Passing this off to @eyurtsev for final review/approval

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2609,15 +2609,15 @@ async def async_no_op(foo: int) -> str:
         "type": "tool_call",
     }
 
-    assert tool.invoke(tool_call["args"]) == "good"
+    assert tool.invoke(tool_call.get("args") or {}) == "good"
```
cbornet:
```suggestion
    assert tool.invoke(tool_call.get("args", {})) == "good"
```

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2609,15 +2609,15 @@ async def async_no_op(foo: int) -> str:
...
+    assert tool.invoke(tool_call.get("args", {})) == "good"
     assert tool_call == {
         "name": "sample_tool",
         "args": {"foo": 2},
         "id": "call_0_82c17db8-95df-452f-a4c2-03f809022134",
         "type": "tool_call",
     }
 
-    assert await tool.ainvoke(tool_call["args"]) == "good"
+    assert tool.invoke(tool_call.get("args") or {}) == "good"
```
cbornet:
```suggestion
    assert tool.invoke(tool_call.get("args", {})) == "good"
```

libs/core/tests/unit_tests/test_tools.py
```diff
@@ -2657,7 +2657,7 @@ async def async_no_op(foo: int) -> str:
         "type": "tool_call",
     }
 
-    assert tool.invoke(tool_call["args"]) == "good"
+    assert tool.invoke(tool_call.get("args") or {}) == "good"
```
cbornet:
```suggestion
    assert tool.invoke(tool_call.get("args", {})) == "good"
```

libs/core/langchain_core/messages/tool.py
```diff
@@ -195,7 +195,7 @@ class ToolCall(TypedDict):
 
     name: str
     """The name of the tool to be called."""
-    args: dict[str, Any]
+    args: Optional[dict[str, Any]]
```
eyurtsev, line 198: Why is this change necessary? 

An empty dict represents no args currently. I'd rather not add a second way to represent the same information

mdrxy: Yes, an empty dict should represent a **completed** tool call that has no arguments. 

`None` is used exclusively to represent the in-progress state where arguments have not yet been received during streaming.

Using an empty dictionary for both the "no arguments" state and the "arguments are coming" state makes them indistinguishable. We wouldn't know how to interpret the message, e.g. two scenarios to illustrate this:

1. Final State (No Arguments): The model finishes a tool call and intends for it to have zero args. It correctly sends `args={}`. We'd see this and execute the tool immediately.
2. In-Transit State (Streaming): The model starts a tool call and is about to stream the arguments. If it used `{}` as a temporary placeholder, we'd see `args={}`

If we went back to `args_ = parse_partial_json(chunk["args"]) if chunk["args"] != "" else {}`, it doesn't handle the case where `chunk["args"]` is literally `None`. When a streaming chunk arrives with `args=None`, `chunk["args"] != ""` evaluates to `True` (since `None != ""`). The code then tries to execute `parse_partial_json(None)`, which results in a TypeError

Even if it mapped `None` to `{}` we still lose the ability to distinguish between the two different scenarios.

ccurme: > If it used {} as a temporary placeholder, we'd see args={}

As I understand we'd never see `{}` as a temporary placeholder. We concatenate string args when summing AIMessageChunks; there's no mechanism to revise earlier chunks as you'd need to do with a placeholder (let me know if I misunderstood that).

If a tool call has no args, when streaming you either stream an empty string or the string `"{}"` (Anthropic does the former, OpenAI and Gemini do the latter, for example). These correctly get parsed to an empty dict here: https://github.com/langchain-ai/langchain/blob/f08dfb6f4916c1eeb1743501958d2672896ef281/libs/core/langchain_core/messages/ai.py#L355

ccurme (commented):
Thanks for this. Left a comment on the original issue. If there is an issue, IMO we should address it in the underlying integration; I think the `args` types for tool call and tool call chunks are OK.

Closing for now but can re-open if needed.

PR #32021, opened 2025-07-14
feat(langchain): add ruff rules DTZ

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-datetimez-dtz

### DIFF
--- a/libs/langchain/langchain/output_parsers/datetime.py
+++ b/libs/langchain/langchain/output_parsers/datetime.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from langchain_core.exceptions import OutputParserException
 from langchain_core.output_parsers import BaseOutputParser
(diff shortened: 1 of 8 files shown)


libs/langchain/langchain/output_parsers/datetime.py
```diff
@@ -48,7 +48,7 @@ def get_format_instructions(self) -> str:
     def parse(self, response: str) -> datetime:
         """Parse a string into a datetime object."""
         try:
-            return datetime.strptime(response.strip(), self.format)
+            return datetime.strptime(response.strip(), self.format)  # noqa: DTZ007
```
cbornet, line 51: Ideally, the default format of `DatetimeOutputParser` would be `%Y-%m-%dT%H:%M:%S.%f%z` and we would ensure the parse method returns a tz-aware datetime.
But changing that would not be backward compatible.

libs/langchain/pyproject.toml
```diff
@@ -216,6 +218,11 @@ pyupgrade.keep-runtime-typing = true
 "langchain/chains/constitutional_ai/principles.py" = [
     "E501", # Line too long
 ]
+"**/retrievers/*time_weighted_retriever.py" = [
```
cbornet, line 220: Fixing TimeWeightedVectorStoreRetriever to use tz-aware datetimes would not be backward compatible.

eyurtsev (approved):

PR #32029, opened 2025-07-14
feat(langchain): add ruff rules G

cbornet:
https://docs.astral.sh/ruff/rules/#flake8-logging-format-g

### DIFF
--- a/libs/langchain/langchain/smith/evaluation/runner_utils.py
+++ b/libs/langchain/langchain/smith/evaluation/runner_utils.py
@@ -461,8 +461,9 @@ def _determine_input_key(
         input_key = config.input_key
         if run_inputs and input_key not in run_inputs:
             logger.warning(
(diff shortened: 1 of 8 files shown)


libs/langchain/langchain/smith/evaluation/runner_utils.py
```diff
@@ -461,16 +461,18 @@ def _determine_input_key(
         input_key = config.input_key
         if run_inputs and input_key not in run_inputs:
             logger.warning(
-                f"Input key {input_key} not in chain's specified"
-                f" input keys {run_inputs}. Evaluation behavior may be undefined.",
+                "Input key {input_key} not in chain's specified input keys %s. "
```
mdrxy: `{input_key}` missed?

cbornet: Indeed, good catch

mdrxy (approved):

PR #32040, opened 2025-07-15
test: Add failing test for `BaseCallbackManager.merge`

Shibayan003:
This pull request introduces a failing unit test to reproduce the bug reported in issue #32028. 
The test asserts the expected behavior: `BaseCallbackManager.merge()` should combine `handlers` and `inheritable_handlers` independently, without mixing them. This test will fail on the current codebase and is intended to guide the fix and prevent future regressions.


eyurtsev:
Ran lint

PR #32041, opened 2025-07-15
fix(core): correctly resolve list indices in JSON Pointer refs and improve docs

anmol-aidora:
- **Description:**
This PR fixes incorrect handling of numeric path tokens in internal JSON-Schema $ref resolution. Numeric tokens are now treated as zero-based list indices whenever the current node is a list, eliminating false-negative look-ups and spurious KeyErrors. The helper also raises an explicit “index out of range” error for invalid indices, validates that every reference starts with #, and returns a deep copy to prevent unintended mutation. A comprehensive doc-string and inline comments clarify pointer semantics, control flow, and error cases.

- **Issue:** #32042

- **Dependencies:**: None

### DIFF
--- a/libs/core/langchain_core/utils/json_schema.py
+++ b/libs/core/langchain_core/utils/json_schema.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, Optional
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
+    from typing import Dict, Any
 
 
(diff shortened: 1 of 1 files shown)


eyurtsev (changes requested):
Would you mind adding a unit test that shows the case that breaks?

anmol-aidora:
> Would you mind adding a unit test that shows the case that breaks?

will do

mdrxy:
I believe this is fixed by https://github.com/langchain-ai/langchain/commit/fb5da8384e636db91b0313cab3cc21b661889684

PR #32047, opened 2025-07-15
feat(langchain): add ruff rules TRY

cbornet:
See https://docs.astral.sh/ruff/rules/#tryceratops-try

* TRY004 (replace by TypeError) in main code is escaped with `noqa` to not break backward compatibility. The rule is still interesting for new code.
* TRY301 ignored at the moment. This one is quite hard to fix and I'm not sure it's very interesting to activate it.


libs/langchain/langchain/chains/router/llm_router.py
```diff
@@ -178,10 +178,10 @@ def parse(self, text: str) -> dict[str, Any]:
             parsed = parse_and_check_json_markdown(text, expected_keys)
             if not isinstance(parsed["destination"], str):
                 msg = "Expected 'destination' to be a string."
-                raise ValueError(msg)
+                raise TypeError(msg)
```
cbornet, line 181: I think here we can replace with `TypeError` without breaking compatibility since it's wrapped later in a `OutputParserException `

PR #32073, opened 2025-07-16
fix(core): better error message for empty var names

isahers1:
Previously, we hit an index out of range error with empty variable names (accessing tag[0]), now we through a slightly nicer error


mdrxy:
re:

```
FAILED tests/unit_tests/prompts/test_structured.py::test_structured_prompt_template_empty_vars - langchain_core.utils.mustache.ChevronError: empty tag at line 1
```

If this is the expected behavior, we'll need to mark it as [xfail](https://blog.ganssle.io/articles/2021/11/pytest-xfail.html)

mdrxy (approved):
LGTM

PR #32075, opened 2025-07-16
chore: copilot instructions

mdrxy:
https://docs.github.com/en/copilot/how-tos/custom-instructions/adding-repository-custom-instructions-for-github-copilot

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,47 +1,151 @@
-* Default to asynchronous methods (ainvoke, abatch, astream) where applicable.
-* Prioritize LangChain Expression Language (LCEL) for creating chains. Use the pipe operator (|) to connect components.
-* Generate modular imports from specific packages like langchain_core, langchain_community, and langchain_openai. Do not import from the top-level langchain package.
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,47 @@
+* Default to asynchronous methods (ainvoke, abatch, astream) where applicable.
```
eyurtsev: IMO:

let's avoid using copilot as a replacement for a linter. (e.g., no need to lint the PR title -- the linter is already doing this.)

Here's a list of rules I generated for community:

https://github.com/langchain-ai/langchain-community/blob/main/.github/copilot-instructions.md

----

I don't know to what extent co-pilot can do this but review process is:

1. For PRs: Do we want this feature/patch etc? (big picture?) What are some other possible designs this? Is a better design? What are limitations of this design etc.
2. If high level design looks good etc, we should now inspect the implementation: is the implementation good, is it unit tested, does it produce any breaking changes, is it documented well. 
3. Are there any security issues (that won't be caught by a linter) etc.

PR #32098, opened 2025-07-18
chore(langchain): add ruff rules N

cbornet:
See https://docs.astral.sh/ruff/rules/#pep8-naming-n


libs/langchain/langchain/agents/agent_toolkits/csv/__init__.py
```diff
@@ -1,27 +1,20 @@
...
 from typing import Any
 
-from langchain_core._api.path import as_import_path
-
 
 def __getattr__(name: str) -> Any:
     """Get attr name."""
 
     if name == "create_csv_agent":
-        # Get directory of langchain package
```
eyurtsev, line 11: i think there was an edge case here which is why i did it this way rather

libs/langchain/tests/unit_tests/stubs.py
```diff
@@ -20,28 +20,28 @@ def __eq__(self, other: object) -> bool:
 # subclassed strings.
 
 
-def _AnyIdDocument(**kwargs: Any) -> Document:
+def any_id_document(**kwargs: Any) -> Document:
```
eyurtsev: I'd prefer to keep these as they were as they stand in for replacement classes

libs/langchain/langchain/agents/agent_toolkits/csv/__init__.py
cbornet: This change replaces the dynamic determination of the package names with a static one.
I don't see how it could be wrong.
It even fixes the target path as the dynamic way resolves to `langchain_experimental.agents.agent_toolkits.csv. create_csv_agent` which doesn't exist.

libs/langchain/tests/unit_tests/stubs.py
cbornet: How about using real classes then ?
Something like
```python
class _AnyIDMixin(BaseModel):
    def __eq__(self, other):
        if isinstance(other, BaseModel):
            dump = self.model_dump()
            dump.pop("id")
            other_dump = other.model_dump()
            other_dump.pop("id")
            return dump == other_dump
        return False

class AnyIdAiMessage(AIMessage, _AnyIDMixin):
    pass
    
class AnyIdAiMessageChunk(AIMessageChunk, _AnyIDMixin):
    pass

class AnyIdHumanMessage(HumanMessage, _AnyIDMixin):
    pass
```

cbornet: I opened https://github.com/langchain-ai/langchain/pull/32209 .
LMK what you think.

cbornet: done.

libs/langchain/langchain/retrievers/self_query/base.py
```diff
@@ -97,7 +97,7 @@ def _get_builtin_translator(vectorstore: VectorStore) -> Visitor:
         Pinecone as CommunityPinecone,
     )
 
-    BUILTIN_TRANSLATORS: dict[type[VectorStore], type[Visitor]] = {
+    builtin_translators: dict[type[VectorStore], type[Visitor]] = {
```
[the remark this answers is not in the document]
cbornet: Ruff rule N disagrees.

PR #32099, opened 2025-07-18
fix(docs): handle collection not found error on RAG tutorial when qdrant is selected as vectorStore

YasienDwieb:
In [Rag Part 1 Tutorial](https://python.langchain.com/docs/tutorials/rag/), when QDrant vector store is selected, the sample code does not work
It fails with error  `ValueError: Collection test not found`

So, this fix is creating that collection and ensuring its dimension size is matching the selection the embedding size of the selected LLM Model


YasienDwieb:
**Before**
Screenshot from 2025-07-19 14-06-14

**After**
Screenshot from 2025-07-19 14-05-48

mdrxy (approved):

PR #32110, opened 2025-07-19
chore(langchain): add ruff rules ARG

cbornet:
See https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg

### DIFF
--- a/libs/langchain/langchain/agents/agent.py
+++ b/libs/langchain/langchain/agents/agent.py
@@ -116,8 +116,8 @@ class BaseSingleActionAgent(BaseModel):
     def return_stopped_response(
         self,
         early_stopping_method: str,
-        intermediate_steps: list[tuple[AgentAction, str]],
-        **kwargs: Any,
+        intermediate_steps: list[tuple[AgentAction, str]],  # noqa: ARG002
+        **_: Any,
     ) -> AgentFinish:
         """Return response when agent has been stopped due to max iterations.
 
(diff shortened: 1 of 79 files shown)


libs/langchain/langchain/agents/agent.py
```diff
@@ -749,18 +751,6 @@ def return_values(self) -> list[str]:
         """Return values of the agent."""
         return ["output"]
 
-    def _fix_text(self, text: str) -> str:
```
cbornet, line 753: This method is private and doesn't seem to be used anywhere.
OK to remove it ?

libs/langchain/langchain/smith/evaluation/progress.py
```diff
@@ -18,7 +19,6 @@ def __init__(
         total: int,
         ncols: int = 50,
         end_with: str = "\n",
-        **kwargs: Any,
```
mdrxy, line 22: Why?

cbornet: ProgressBarCallback doesn't use kwargs.
It seems it was there by mistake and nobody would call it with args other than the named args of the signature.
In theory it would be a breaking change, but in practice I think it's OK.
LMK if you prefer to keep the kwargs.

mdrxy: Deferring to @eyurtsev for decision

mdrxy (approved):

PR #32166, opened 2025-07-22
feat(langchain): v1 WIP

eyurtsev:
This PR adds scaffolding for langchain 1.0 entry package.

Most contents have been removed. 

Currently remaining entrypoints for:

* chat models
* embedding models
* memory -> trimming messages, filtering messages and counting tokens [we may remove this]
* prompts -> we may remove some prompts
* storage: primarily to support cache backed embeddings, may remove the kv store
* tools -> report tool primitives

Things to be added:

* Selected agent implementations
* Selected workflows
* Common primitives: messages, Document
* Primitives for type hinting: BaseChatModel, BaseEmbeddings
* Selected retrievers
* Selected text splitters

Things to be removed:

* Globals needs to be removed (needs an update in langchain core)

Todos: 

* TBD indexing api (requires sqlalchemy which we don't want as a dependency)
* Be explicit about public/private interfaces (e.g., likely rename chat_models.base.py to something more internal)
* Remove dockerfiles
* Update module doc-strings and README.md

### DIFF
--- a/libs/langchain_v1/.dockerignore
+++ b/libs/langchain_v1/.dockerignore
@@ -1,0 +1,6 @@
+.venv
+.github
+.git
+.mypy_cache
+.pytest_cache
+Dockerfile
--- a/.github/scripts/check_diff.py
+++ b/.github/scripts/check_diff.py
@@ -16,6 +16,7 @@ LANGCHAIN_DIRS = [
     "libs/core",
     "libs/text-splitters",
     "libs/langchain",
+    "libs/langchain_v1",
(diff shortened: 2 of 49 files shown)


libs/langchain_v1/.dockerignore
mdrxy, line 1: Is this meant for the dev container?

eyurtsev: No idea why it's there I'll remove do a quick follow up to clean up everything we don't need

PR #32168, opened 2025-07-22
docs: add documentation about how to use extended thinking with ChatBedrockConverse

jasonkang14:
- Added how to use [extended thinking](https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-extended-thinking.html) with `ChatBedrockConverse`
- Closes #31535


ccurme (approved):
Thanks!

PR #32208, opened 2025-07-23
feat: new message formats (v0.4)

ccurme:


libs/core/pyproject.toml
```diff
@@ -105,6 +106,7 @@ unfixable = ["PLW1510",]
 
 flake8-annotations.allow-star-arg-any = true
 flake8-annotations.mypy-init-return = true
+flake8-builtins.ignorelist = ["id"]
```
eyurtsev: We should probably add:

* `input` (Runnable interface)
* `type` (schema)

We're aiming to be cross-language compatible, and would prefer not to change intuitive names b/c of shadowing of python built ins.

libs/core/pyproject.toml
```diff
@@ -86,6 +86,7 @@ ignore = [
     "FIX002", # Line contains TODO
     "ISC001", # Messes with the formatter
     "PERF203", # Rarely useful
+    "PLC0414",  # Enable re-export
```
eyurtsev, line 89: OK change. The rule doesn't make sense as re-exporting with an alias is actually the recommended way to re-export

libs/core/langchain_core/messages/v1.py
```diff
@@ -0,0 +1,538 @@
...
+    """
+
+    id: Optional[str] = None
+
+    lc_version: str = "v1"
+    """Encoding version for the message."""
+
+    content: list[types.ContentBlock] = field(default_factory=list)
+    usage_metadata: Optional[UsageMetadata] = None
+    response_metadata: dict = field(default_factory=dict)
```
eyurtsev, line 94: Could we document what the purpose of this field is. What kind of information gets stored there.

Specifically, who is responsible for setting this:

1. Is this raw data from the provider
2. Is this standardized data by langchain
3. Is a user allowed to mutate this

libs/core/pyproject.toml
cbornet: Why not just use `noqa: A002` when needed ?

libs/core/pyproject.toml
```diff
@@ -105,6 +105,7 @@ unfixable = ["PLW1510",]
 
 flake8-annotations.allow-star-arg-any = true
 flake8-annotations.mypy-init-return = true
+flake8-builtins.ignorelist = ["id"]
```
cbornet: Why not just use `noqa: A002` when needed ?

PR #32253, opened 2025-07-26
feat(core): support for adding `PromptTemplate`s with formats other than `f-string`

### LINKED ISSUE #32151: Support for adding `PromptTemplate`s of Jinja2 template format
Hi there!

Currently adding `PromptTemplate`s of  Jinja2 template format is not supported:

https://github.com/langchain-ai/langchain/blob/64261449b837c937f7152b1966b5ec1cf6b745de/libs/core/langchain_core/prompts/prompt.py#L136-L145

while it would be quite useful for the users employing Jinja2 as the template format. If there's no major limitation to it, I'd be happy to contribute a PR to add support.

sadra-barikbin:
Allow adding`PromptTemplate`s with formats other than `f-string`. Fixes #32151

### DIFF
--- a/libs/core/langchain_core/prompts/prompt.py
+++ b/libs/core/langchain_core/prompts/prompt.py
@@ -153,8 +153,8 @@ class PromptTemplate(StringPromptTemplate):
         # Allow for easy combining
         if isinstance(other, PromptTemplate):
             if self.template_format != other.template_format:
-                msg = "Cannot add two templates of different formats"
-                raise ValueError(msg) 
+                msg = "Cannot add templates of different formats"
+                raise ValueError(msg)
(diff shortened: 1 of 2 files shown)


eyurtsev:
@sadra-barikbin overall looks good. Could you offer a few sentences on why you need to add 2 jinja2 prompts rather than just assemble a single prompt from the get-go?

sadra-barikbin:
@eyurtsev , Sure. I was to make improvement in a prompt builder class with many private methods each one returning a part of the total prompt, all added up together finally. Changing the way prompt was built to having a huge template file containing the whole prompt was considered a sweeping change.

eyurtsev (approved):

PR #32275, opened 2025-07-28
feat(core): Add `bind_tools` on `FakeChatModel` for development purpose

amosctlee:
Implements the bind_tools method for FakeListChatModel, enabling tool binding and invocation in development


ccurme (commented):
Thanks for this. `llm.bind_tools(...).invoke(...)` should return an AIMessage with `.tool_calls`; it should not return a dict or invoke the tool.

In most tests (e.g., with GenericFakeChatModel below) you define the sequence of messages that will be generated. So you can already include AIMessage with tool calls in this sequence.

Under the hood, `bind_tools` just calls `llm.bind(...)` with a provider-specific tool definition. If you want to test the behavior of the result of `llm.bind_tools`, you should be able to call `llm.bind(foo="bar")`.

Let me know if I misunderstood the goal.

PR #32299, opened 2025-07-29
fix(text-splitters): override `create_document` in `TokenTextSplitter`

### LINKED ISSUE #29884: TokenTextSplitter start indices are sometimes -1

ZacharyGonzalez:
Description: Overrode create_documents in TokenTextSplitter to avoid incompatibility with the base inherited TextSplitter method. Provides starting index's for token based text. 

I've added two test cases, one to cover the original issue and an one with repeating text to check for skipping indexes
Issue: Fixes #29884


ZacharyGonzalez:
Some things to note: tokenizer.decode is lossy so while the current implementation provides indexes for the TokenTextSplitter, it still has -1 as a fallback point since it is using find().

mdrxy:
Closing so that a new approach can be taken

PR #32312, opened 2025-07-29
docs(langchain): update pip command on the RAG example

IanOliv:
remove unnecessary character on the lib installation


mdrxy:
In Jupyter notebooks, putting a `%` in front of a command like `pip` install turns it into a magic command, ensuring that the package is installed in the same environment that the Jupyter kernel is using

PR #32316, opened 2025-07-29
feat(core): content block factories + ids + docs + tests

mdrxy:
## Benefits

  1. **Type Safety**: Compile-time validation of required fields and proper type setting
  2. **Less Boilerplate**: No need to manually set the `type` field or generate IDs
  3. **Input Validation**: Runtime validation prevents common errors (e.g., base64 without MIME type)
  4. **Consistent Patterns**: Standardized creation patterns across all block types
  5. **Better Developer Experience**: Cleaner, more intuitive API than manual TypedDict construction. Also follows similar other patterns (e.g. `create_react_agent`, `init_chat_model`


libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -397,8 +402,12 @@ class CodeInterpreterResult(TypedDict):
 
 
 # --- Reasoning ---
```
eyurtsev: Unrelated to change but could remove some of the intermediate comments since they don't add any extra information

```suggestion
```

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -13,31 +13,32 @@
 ``NonStandardContentBlock``, which allows for provider-specific data to be included
 without losing the benefits of type checking and validation.
 
-Furthermore, provider-specific fields *within* a standard block will be allowed as extra
-keys on the TypedDict per `PEP 728 <https://peps.python.org/pep-0728/>`__. This allows
-for flexibility in the data structure while maintaining a consistent interface.
+Furthermore, provider-specific fields *within* a standard block are fully supported
```
eyurtsev: Thank you for the doc-string here

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -828,3 +825,460 @@ def convert_to_openai_data_block(block: dict) -> dict:
...
+    text: str,
+    *,
+    url: Optional[str] = None,
+    base64: Optional[str] = None,
+    file_id: Optional[str] = None,
+    title: Optional[str] = None,
+    context: Optional[str] = None,
+    id: Optional[str] = None,
+    index: Optional[int] = None,
+) -> PlainTextContentBlock:
```
eyurtsev, line 1231: Are we set on the name PlainTextContentBlock vs. TextContentBlock?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -828,3 +825,460 @@ def convert_to_openai_data_block(block: dict) -> dict:
...
+
+def create_file_block(
+    *,
+    url: Optional[str] = None,
+    base64: Optional[str] = None,
+    file_id: Optional[str] = None,
+    mime_type: Optional[str] = None,
+    id: Optional[str] = None,
+    index: Optional[int] = None,
+) -> FileContentBlock:
```
eyurtsev, line 1174: 1. should these accommodate kwargs? (as extras?)

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -828,3 +825,460 @@ def convert_to_openai_data_block(block: dict) -> dict:
         raise ValueError(error_msg)
 
     return formatted_block
+
+
+# Factories
```
eyurtsev: What's the goal of the factories? Is it primarily to add additional run time validation?

eyurtsev: closed after rereading PR summary

libs/core/langchain_core/messages/content_blocks.py
mdrxy: yes since there would be a conflict with the already defined `TextContentBlock` - plaintext to signify plaintext file

libs/core/langchain_core/messages/content_blocks.py
mdrxy: Each block can take extra keys per `extra_items=Any` and as such those fields are typed as `Any` - this handles appropriately?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -185,8 +230,8 @@ class TextContentBlock(TypedDict):
     """Index of block in aggregate response. Used during streaming."""
 
 
-# --- Tool calls ---
-class ToolCall(TypedDict):
+class ToolCall(TypedDict):  # type: ignore[call-arg]
+    # class ToolCall(TypedDict, extra_items=Any):  # type: ignore[call-arg]
```
eyurtsev: commented out code? (same above) is this intentional?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -256,33 +298,62 @@ class ToolCallChunk(TypedDict):
         ).tool_call_chunks == [ToolCallChunk(name='foo', args='{"a":1}', index=0)]
     """
 
+    # TODO: Consider making fields NotRequired[str] in the future.
+
+    type: NotRequired[Literal["tool_call_chunk"]]
```
eyurtsev, line 321: How come this is not required?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -256,33 +298,62 @@ class ToolCallChunk(TypedDict):
...
+
+# Note: These are not standard tool calls, but rather provider-specific built-in tools.
 # Web search
 class WebSearchCall(TypedDict):
-    """Content block for a built-in web search tool call."""
+    """Built-in web search tool call."""
 
     type: Literal["web_search_call"]
-    """Type of the content block."""
+    """Type of the content block. Used for serialization."""
```
eyurtsev: serialization  -> discrimination

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -54,17 +66,28 @@
...
 
 **Example Usage**
 
 .. code-block:: python
 
+    from langchain_core.messages.content_blocks import create_text_block, create_image_block
+
+    # RECOMMENDED: Use factory functions for creating content blocks
+    multimodal_message: AIMessage = [
+        create_text_block("What is shown in this image?"),
```
ccurme: I think we can use these in our own code but IMO our docs should generally showcase construction with dicts, e.g.,
```python
HumanMessage([{"type": "text", "text": "..."}])
```

We can note that these are available as a convenience.

Relatedly, our docs should showcase what usage patterns we prefer— so if we showcase dicts, I'm not sure the API ref for this module should **recommend** using the factory functions (vs. noting they're available and touting their benefits).

libs/core/langchain_core/messages/v1.py
```diff
@@ -203,7 +210,45 @@ class AIMessageChunk(AIMessage):
...
+    """
+
+    id: Optional[str] = None
+    """Unique identifier for the message chunk.
+
+    If the provider assigns a meaningful ID, it should be used here. Otherwise, a
+    LangChain-generated ID will be used.
+    """
+
+    lc_version: str = "v1"
```
eyurtsev: this attribute name sounds like the version of `langchain` the package rather than the representation version. Likely fine since it's matching the `version` attribute name in chat models

libs/core/tests/unit_tests/messages/test_content_block_factories.py
```diff
@@ -0,0 +1,568 @@
...
+        citation = create_citation()
+        assert isinstance(citation, dict)
+        assert citation["type"] == "citation"
+
+        non_standard_block = create_non_standard_block({"data": "value"})
+        assert isinstance(non_standard_block, dict)
+        assert non_standard_block["type"] == "non_standard"
+
+
+# TODO: add when PEP 728 is fully supported
```
eyurtsev: Should we create an issue?

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -92,16 +114,25 @@ class Citation(TypedDict):
         ``start/end`` indices refer to the **response text**,
         not the source text. This means that the indices are relative to the model's
         response, not the original document (as specified in the ``url``).
+
+    .. note::
+        ``create_citation`` is the preferred way to create a ``Citation`` as it:
+
+        * Automatically generate unique IDs when not provided
```
ccurme: Do we require IDs or not? I think they should remain NotRequired, so that users can construct valid content blocks without any imports. If they're not required, I'm not sure we should rely on them for anything important.

libs/core/langchain_core/messages/content_blocks.py
mdrxy: Yes as this will be brought back in when `extra_items` is supported - can remove for now if needed

libs/core/langchain_core/messages/content_blocks.py
mdrxy: @eyurtsev via @ccurme it would break things now. v1 seems like a good time to do it

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -92,16 +114,25 @@ class Citation(TypedDict):
...
+        * Validate required arguments at creation time
```
ccurme: IDEs should flag missing keys for you already, no?
Screenshot 2025-07-30 at 12 42 44 PM

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -158,21 +189,35 @@ class NonStandardAnnotation(TypedDict):
 Annotation = Union[Citation, NonStandardAnnotation]
 
 
-class TextContentBlock(TypedDict):
-    """Content block for text output.
+class TextContentBlock(TypedDict):  # type: ignore[call-arg]
+    # class TextContentBlock(TypedDict, extra_items=Any):  # type: ignore[call-arg]
```
ccurme: ooc why is this commented out?

libs/core/langchain_core/messages/v1.py
mdrxy: There won't be a `version` attribute name anymore in chat models IIRC

Perhaps we keep `lc_version` matching the MAJOR version of the package itself? Unsure on this

libs/core/langchain_core/messages/content_blocks.py
ccurme: these are the tool call classes used in existing messages, IMO if we change them we should understand the scope of the impact and do it in a separate PR.

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -828,3 +958,457 @@ def convert_to_openai_data_block(block: dict) -> dict:
...
+        The ``id`` is generated automatically if not provided, using a UUID4 format
+        prefixed with ``'lc_'`` to indicate it is a LangChain-generated ID.
+
+    """
+    if not any([url, base64, file_id]):
+        msg = "Must provide one of: url, base64, or file_id"
+        raise ValueError(msg)
+
+    if base64 and not mime_type:
+        msg = "mime_type is required when using base64 data"
```
ccurme, line 1037: nit / not blocking but is this true?

libs/core/langchain_core/messages/content_blocks.py
mdrxy: see [this](https://github.com/langchain-ai/langchain/pull/32316#discussion_r2243292841) - was getting runtime issues in python 3.10.16

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -828,3 +958,457 @@ def convert_to_openai_data_block(block: dict) -> dict:
...
+        block["title"] = title
+    if context is not None:
+        block["context"] = context
+    if index is not None:
+        block["index"] = index
+
+    return block
+
+
+def create_tool_call(
```
ccurme, line 1275: We will at some point want to consolidate with the existing function here: https://github.com/langchain-ai/langchain/blob/a9e52ca605622be0724766efe9f15a44e1d2b11e/libs/core/langchain_core/messages/tool.py#L211-L224

only difference appears to be we auto-assign IDs here. I'd do in a separate PR in the future.

libs/core/langchain_core/messages/v1.py
```diff
@@ -203,7 +210,45 @@ class AIMessageChunk(AIMessage):
         usage_metadata: Optional metadata about token usage and costs.
     """
 
-    type: Literal["ai_chunk"] = "ai_chunk"  # type: ignore[assignment]
+    type: Literal["ai_chunk"] = "ai_chunk"
+    """Used for serialization."""
+
+    name: Optional[str] = None
```
ccurme: why are we re-implementing these but keeping the inheritance

libs/core/langchain_core/messages/content_blocks.py
mdrxy: I think keeping them `NotRequired` is fine so that things don't fail during construction, but if using a factory should auto-gen one I think since there appears no downside to doing so - see [this thread](https://langchain.slack.com/archives/C04K6GUJ7NW/p1753206368251329)

We shouldn't rely on them for important stuff, so maybe a note in the docs to indicate this? LMK

libs/core/langchain_core/messages/v1.py
mdrxy: IDK, you brought it over IIRC

libs/core/langchain_core/messages/content_blocks.py
mdrxy: I don't think strictly but it a strong best practice

Base64 without MIME type works fine in closed systems but is less than ideal in scenarios where the consumer needs to know how to handle the content (can't infer from the data string alone)

libs/core/langchain_core/messages/content_blocks.py
mdrxy: Yes, however for instances like multimodal where one of `url` `base64` or `file_id` is required this is useful since neither are individually required and we have no other way of enforcing that at least one is present

libs/core/tests/unit_tests/test_messages.py
```diff
@@ -294,30 +304,48 @@ def test_message_chunks_v1() -> None:
...
     )  # LangChain-assigned run ID
     meaningful_id = AIMessageChunkV1(
         content=[], id="msg_def456"
     )  # provider-assigned ID
 
-    assert (null_id + default_id).id == "run-abc123"
-    assert (default_id + null_id).id == "run-abc123"
+    # Provider-assigned IDs (non-run-*) have highest precedence
+    # Among provider-assigned IDs, the first one in the candidates list wins
+    assert (auto_id + meaningful_id).id == auto_id.id  # auto_id is first
```
ccurme: is this an issue? a stream might have:
```
chunk(id=None) + chunk(id=None) + chunk(id="msg_123")
```
if we auto-assign an ID, are we going to lose the meaningful ID?

can we treat `lc_` prefix the same as the existing `run-` prefix?

libs/core/langchain_core/messages/v1.py
ccurme: originally yes, but this code is green now because it is undoing this commit from yesterday: https://github.com/langchain-ai/langchain/pull/32311/commits/607ae56152681a8c239c32b7802dff4a25ec52e7

sydney-runkle:
> Also follows similar other patterns (e.g. create_react_agent, init_chat_model

What does this mean?

.github/workflows/codspeed.yml
```diff
@@ -20,6 +20,7 @@ jobs:
   codspeed:
     name: 'Benchmark'
     runs-on: ubuntu-latest
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'codspeed-ignore') }}
```
sydney-runkle, line 23: Are we using?

libs/core/langchain_core/messages/v1.py
```diff
@@ -467,7 +490,14 @@ class HumanMessage:
     """
 
     id: str
+    """Used for serialization.
+
+    If the provider assigns a meaningful ID, it should be used here. Otherwise, a
+    LangChain-generated ID will be used.
+    """
+
     content: list[types.ContentBlock]
```
sydney-runkle, line 514: Let's make sure all fields have docstrings

libs/core/pyproject.toml
```diff
@@ -13,7 +13,7 @@ dependencies = [
     "PyYAML>=5.3",
     "typing-extensions>=4.7",
     "packaging>=23.2",
-    "pydantic>=2.7.4",
+    "pydantic==2.11.7",
```
sydney-runkle: We can't pin pydantic

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -13,31 +13,46 @@
...
+        # Add required fields
         "type": "text",
         "text": "Hello, world!",
-        "extra_field": "This is allowed",
-        "another_field": 42,  # Any type is allowed
+        # Additional fields not specified in the TypedDict
+        # These are valid with PEP 728 and are typed as Any
+        "openai_metadata": {"model": "gpt-4", "temperature": 0.7},
+        "anthropic_usage": {"input_tokens": 10, "output_tokens": 20},
+        "custom_field": "any value",
```
sydney-runkle, line 47: Hmm why not for now do a `extra: dict[str, Any]` or something similar, given that PEP 728 is a ways away?

sydney-runkle: Happy to chat sync about this tomorrow

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -92,16 +132,24 @@ class Citation(TypedDict):
...
         not the source text. This means that the indices are relative to the model's
         response, not the original document (as specified in the ``url``).
+
+    .. note::
+        ``create_citation`` may also be used as a factory to create a ``Citation``.
+        Benefits include:
+
+        * Automatic ID generation (when not provided)
+        * Required arguments strictly validated at creation time
+
```
sydney-runkle, line 142: Not sure if we want to advertise these as public? We can, but wouldn't they only really be for provider sdks?

sydney-runkle (commented):
Really good foundational work @mdrxy, thanks for digging in deep here. Some follow up qs!

.github/workflows/codspeed.yml
mdrxy: It was useful to have the option when I was debugging a CI/local discrepancy

libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -830,3 +937,457 @@ def convert_to_openai_data_block(block: dict) -> dict:
...
+    )
+
+    if index is not None:
+        block["index"] = index
+
+    return block
+
+
+def create_reasoning_block(
+    reasoning: str,
```
ccurme: Unless you opt-in to reasoning summaries and do organization verification, openai reasoning blocks are:
```
{"type": "reasoning", "id": "rs_..."}
```
no reasoning string.

libs/core/langchain_core/messages/v1.py
```diff
@@ -41,13 +49,15 @@ class ResponseMetadata(TypedDict, total=False):
...
     response headers, service tiers, log probabilities, system fingerprints, etc.
 
-    Extra keys are permitted from what is typed here.
+    Extra keys are permitted from what is typed here (via `total=False`), allowing
+    for provider-specific metadata to be included without breaking the type
+    definition.
     """
 
-    model_provider: str
+    model_provider: NotRequired[str]
```
ccurme: I like having these required, we already check for `model_name` in standard tests for usage metadata and `model_provider` can be trivially added.

libs/core/tests/unit_tests/test_messages.py
```diff
@@ -332,7 +360,7 @@ def test_chat_message_chunks() -> None:
     ):
         ChatMessageChunk(role="User", content="I am") + ChatMessageChunk(
             role="Assistant", content=" indeed."
-        )
+        )  # type: ignore[reportUnusedExpression, unused-ignore]
```
ccurme, line 363: nit but if CI doesn't have a problem with it, feels weird to add a type ignore and then ignore the unused ignore.

not sure if it's possible to make your env consistent with CI.

ccurme (approved):

libs/core/tests/unit_tests/test_messages.py
mdrxy: It fails both with CI and local

PR #32325, opened 2025-07-30
chore(text-splitters): select ALL rules with exclusions

cbornet:


libs/text-splitters/pyproject.toml
```diff
@@ -60,11 +60,45 @@ ignore_missing_imports = "True"
 [tool.ruff]
 target-version = "py39"
 
+
 [tool.ruff.lint]
-select = ["E", "F", "FBT", "I", "UP", "PGH003", "T201", "D"]
-ignore = ["D100",]
+select = [ "ALL",]
```
mdrxy, line 70: I'm still against selecting `ALL` without scoping `fixable`:
* New ruff versions may introduce rules that automatically "fix" code in unintended ways
* Auto-fixes for complex rules might introduce bugs while "improving" code or making it less readable

So perhaps we keep `ALL` under `select` but manually opt new rules into `fixable` (not selecting `ALL`)

mdrxy: Other wise this PR looks good

cbornet: Ruff doesn't introduce new rules between patch versions (https://docs.astral.sh/ruff/versioning/) and we pin the minor version (currently 0.12).
So there's no difference with a long list of rules.
I can put the full list but I think using `ALL` makes the conf simpler.

libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -61,6 +62,25 @@ class ElementType(TypedDict):
     metadata: dict[str, str]
 
 
+def _find_all_strings(
```
cbornet, line 69: Unfortunately, `BeautifulSoup` doesn't define overloads for `Tag.find_all`.
So doing the type resolution ourselves.

PR #32327, opened 2025-07-30
docs(ollama): update `ollama.mdx` with note about basic auth support

btelles:
This is related to https://github.com/langchain-ai/langchain/issues/25055

Although langchain now allows users to specify headers, the client itself does not gracefully handle a standard URL of the form https://user:password@ollama.mysite.com:11343

Issue #25055, opened 2024-08-05
Unable to set authenticiation (headers or auth) like I used to do in the community ollama integration


turbopape:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

from langchain_ollama import ChatOllama
from langchain_ollama.embeddings import  OllamaEmbeddings

llama3_1 = ChatOllama(

   headers=headers,
   base_url="some_host",

    model="llama3.1",
    temperature=0.001,
)
from langchain.prompts import ChatPromptTemplate
chat_prompt = ChatPromptTemplate.from_messages([
    ("system", "you are a helpful search assistant"),
])
chain = chat_prompt | llama3_1
chain.invoke({})

# embeddings

embeddings = (
    OllamaEmbeddings(
    headers=headers,
    base_url="my_public_host",
    model="mxbai-embed-large",
)
)

### Error Message and Stack Trace (if applicable)

# for chat
Error 401
# for embeddings

ValidationError: 2 validation errors for OllamaEmbeddings
base_url
  extra fields not permitted (type=value_error.extra)
headers
  extra fields not permitted (type=value_error.extra)

### Description

After I migrated to the new ChatOllama module (langchain-chatollama), I am unable to set headers or auth.
I am hosting ollama behind ngrok publicly, and I need to authenticate the calls.
When using the langchain_community ChatOllama integration, I was able to set those.
This seems to be like base_url, which was added in the latest version. 
If you know of any env var that I can use to fix this (like OLLAMA_HOST) to set auth headers, I'd be very thankful :)

### System Info

langchain==0.2.12
langchain-community==0.2.11
langchain-core==0.2.28
langchain-ollama==0.1.1
langchain-openai==0.1.19
langchain-postgres==0.0.9
langchain-text-splitters==0.2.2
langchainhub==0.1.20

platform: mac
python: 3.12.3

isahers1:
@turbopape Thank you for bringing this up! Does this PR fix your issues: https://github.com/langchain-ai/langchain/pull/25078? (feel free to edit as needed)

turbopape:
[...]

PR #32328, opened 2025-07-30
feat(ollama): add basic auth support

mdrxy:
support for URL authentication in the format `https://user:password@host:port` for all LangChain Ollama clients.

Related to #32327 and #25055


btelles (approved):
Thanks! So nice to see you all on top of it! LGTM! Only nit is that there's some repetition when we update the headers after parsing auth parameters...a helper could have reduced that, but not a big deal.

PR #32358, opened 2025-07-31
feat(core): move tool call chunks to content (v1)

ccurme:
Tool call chunks must be represented in content so that we don't lose their sequencing among content blocks. Currently they are an attribute on AIMessageChunk.

Here we do the following:
- Store tool call chunks on `AIMessageChunk.content` (for v1 message chunks) and change `.tool_call_chunks` to a getter property.
- Add a bool attribute `is_last_chunk` to AIMessageChunk. When this chunk is encountered during a sum, we parse the tool call chunks in content and replace them with tool calls.

AIMessageChunk.tool_calls works as before, but does not update content.

Example:
```python
from langchain_openai.chat_models.base_v1 import ChatOpenAI

llm = ChatOpenAI(
    model="o4-mini",
    reasoning={"effort": "low", "summary": "auto"},
)

    pass

llm_with_tools = llm.bind_tools([get_weather])

full = None
chunks = []
for chunk in llm_with_tools.stream(
    "Can you look up the weather now in the city that had the "
    "third highest global population in the year 2000?"
):
    if full.tool_call_chunks:
        print(full.tool_call_chunks)

full.content
```
```
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '{"', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '{"location', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '{"location":"', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '{"location":"Mexico', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '{"location":"Mexico City', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '{"location":"Mexico City"}', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]
[{'type': 'tool_call_chunk', 'name': 'get_weather', 'args': '{"location":"Mexico City"}', 'id': 'call_GthiNOvoATgg0iwmYAb9dEtS', 'index': 2}]


[{'id': 'rs_688bd80e450881a09763e84da41f461704a12ed7e9ff865d',
  'type': 'reasoning',
  'index': 0,
  'reasoning': "**Determining city rankings for weather lookup**\n\nThe user is asking about..."},
 {'type': 'reasoning',
  'reasoning': '**Finalizing city selection for weather request**\n\nI’m working through the rankings...',
  'id': 'rs_688bd80e450881a09763e84da41f461704a12ed7e9ff865d',
  'index': 1},
 {'name': 'get_weather',
  'args': {'location': 'São Paulo'},
  'id': 'call_xYacfsBDaNOpCX2okVpUadDi',
  'type': 'tool_call'}]
```

### DIFF
--- a/libs/core/langchain_core/callbacks/manager.py
+++ b/libs/core/langchain_core/callbacks/manager.py
@@ -279,7 +279,7 @@ def _convert_llm_events(
             generations=[
                 [
                     ChatGeneration(
-                        text=args_list[0].text,
+                        text=args_list[0].text or "",
                         message=convert_from_v1_message(args_list[0]),
                     )
                 ]
--- a/libs/core/langchain_core/callbacks/manager.py
+++ b/libs/core/langchain_core/callbacks/manager.py
@@ -279,7 +279,7 @@ def _convert_llm_events(
             generations=[
                 [
                     ChatGeneration(
-                        text=args_list[0].text or "",
+                        text=args_list[0].text,
                         message=convert_from_v1_message(args_list[0]),
                     )
                 ]
--- a/libs/core/langchain_core/messages/content_blocks.py
+++ b/libs/core/langchain_core/messages/content_blocks.py
@@ -865,17 +865,22 @@ def _extract_typedict_type_values(union_type: Any) -> set[str]:
 
 
 KNOWN_BLOCK_TYPES = {
-    bt for bt in get_args(ContentBlock) for bt in get_args(bt.__annotations__["type"])
-}
-# Normalize any Literal[...] types in KNOWN_BLOCK_TYPES to their string values.
-# This ensures all entries are plain strings, not Literal objects.
-KNOWN_BLOCK_TYPES = {
-    t
-    if isinstance(t, str)
-    else t.__args__[0]
-    if hasattr(t, "__args__") and len(t.__args__) == 1
-    else t
-    for t in KNOWN_BLOCK_TYPES
+    "text",
+    "text-plain",
+    "tool_call",
+    "invalid_tool_call",
+    "tool_call_chunk",
+    "reasoning",
+    "non_standard",
+    "image",
+    "audio",
+    "file",
+    "video",
+    "code_interpreter_call",
+    "code_interpreter_output",
+    "code_interpreter_result",
+    "web_search_call",
+    "web_search_result",
 }
 
 
--- a/libs/core/langchain_core/messages/content_blocks.py
+++ b/libs/core/langchain_core/messages/content_blocks.py
@@ -840,6 +840,7 @@ ContentBlock = Union[
     TextContentBlock,
     ToolCall,
     InvalidToolCall,
+    ToolCallChunk,
     ReasoningContentBlock,
     NonStandardContentBlock,
     DataContentBlock,
@@ -862,6 +863,16 @@ def _extract_typedict_type_values(union_type: Any) -> set[str]:
 
 KNOWN_BLOCK_TYPES = {
     bt for bt in get_args(ContentBlock) for bt in get_args(bt.__annotations__["type"])
+}
+# Normalize any Literal[...] types in KNOWN_BLOCK_TYPES to their string values.
+# This ensures all entries are plain strings, not Literal objects.
+KNOWN_BLOCK_TYPES = {
+    t
+    if isinstance(t, str)
+    else t.__args__[0]
+    if hasattr(t, "__args__") and len(t.__args__) == 1
+    else t
+    for t in KNOWN_BLOCK_TYPES
 }
 
 
--- a/libs/core/langchain_core/messages/v1.py
+++ b/libs/core/langchain_core/messages/v1.py
@@ -159,10 +159,22 @@ class AIMessage:
                 if "id" in tool_call and tool_call["id"] in content_tool_calls:
                     continue
                 self.content.append(tool_call)
+        if invalid_tool_calls:
+            content_tool_calls = {
+                block["id"]
+                for block in self.content
+                if block["type"] == "invalid_tool_call" and "id" in block
+            }
+            for tool_call in invalid_tool_calls:
+                if "id" in tool_call and tool_call["id"] in content_tool_calls:
+                    continue
+                self.content.append(tool_call)
         self._tool_calls = [
             block for block in self.content if block["type"] == "tool_call"
         ]
-        self.invalid_tool_calls = invalid_tool_calls or []
+        self._invalid_tool_calls = [
+            block for block in self.content if block["type"] == "invalid_tool_call"
+        ]
 
     @property
     def text(self) -> Optional[str]:
@@ -173,19 +185,27 @@ class AIMessage:
         return None
 
     @property
-    def tool_calls(self) -> list[types.ToolCall]:  # update once we fix branch
+    def tool_calls(self) -> list[types.ToolCall]:
         """Get the tool calls made by the AI."""
-        if self._tool_calls:
-            return self._tool_calls
-        tool_calls = [block for block in self.content if block["type"] == "tool_call"]
-        if tool_calls:
-            self._tool_calls = tool_calls
-        return [block for block in self.content if block["type"] == "tool_call"]
+        if not self._tool_calls:
+            self._tool_calls = [
+                block for block in self.content if block["type"] == "tool_call"
+            ]
+        return self._tool_calls
 
     @tool_calls.setter
     def tool_calls(self, value: list[types.ToolCall]) -> None:
         """Set the tool calls for the AI message."""
         self._tool_calls = value
+
+    @property
+    def invalid_tool_calls(self) -> list[types.InvalidToolCall]:
+        """Get the invalid tool calls made by the AI."""
+        if not self._invalid_tool_calls:
+            self._invalid_tool_calls = [
+                block for block in self.content if block["type"] == "invalid_tool_call"
+            ]
+        return self._invalid_tool_calls
 
 
 @dataclass
@@ -205,8 +225,6 @@ class AIMessageChunk(AIMessage):
     """
 
     type: Literal["ai_chunk"] = "ai_chunk"  # type: ignore[assignment]
-
-    tool_call_chunks: list[types.ToolCallChunk] = field(init=False)
 
     def __init__(
         self,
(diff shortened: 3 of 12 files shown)


libs/core/langchain_core/messages/content_blocks.py
```diff
@@ -866,6 +867,16 @@ def _extract_typedict_type_values(union_type: Any) -> set[str]:
 KNOWN_BLOCK_TYPES = {
     bt for bt in get_args(ContentBlock) for bt in get_args(bt.__annotations__["type"])
 }
+# Normalize any Literal[...] types in KNOWN_BLOCK_TYPES to their string values.
+# This ensures all entries are plain strings, not Literal objects.
+KNOWN_BLOCK_TYPES = {
```
eyurtsev: Is this responsible for getting the names of the block types?

If so, can we make the code as dumb as possible and use a set of string literals?

If not, then what is this code doing?

libs/core/langchain_core/callbacks/manager.py
```diff
@@ -279,7 +279,7 @@ def _convert_llm_events(
             generations=[
                 [
                     ChatGeneration(
-                        text=args_list[0].text,
+                        text=args_list[0].text or "",
```
eyurtsev: The code looks like it's re-defining args_list[0] and re-assigning it?

Could we use a new variable name?

---

If args_list[0] is a Message. Can `text` be restricted to never be an Optional? i.e., is it useful to distinguish between empty string vs. None in user code?

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -654,3 +658,93 @@ def test_normalize_messages_edge_cases() -> None:
...
     assert messages == _normalize_messages(messages)
+
+
+def test_streaming_v1() -> None:
+    chunks = [
+        AIMessageChunkV1(
+            [
+                {
+                    "type": "reasoning",
+                    "reasoning": "Let's call a tool.",
```
eyurtsev, line 669: Is there a `is_first_chunk` that mirrors `is_last_chunk`? 


Should we use `chunk_pos: NotRequired[Literal['first', 'last']]`

libs/core/langchain_core/messages/v1.py
```diff
@@ -190,20 +202,28 @@ def text(self) -> Optional[str]:
...
-    def tool_calls(self) -> list[types.ToolCall]:  # update once we fix branch
+    def tool_calls(self) -> list[types.ToolCall]:
         """Get the tool calls made by the AI."""
-        if self._tool_calls:
-            return self._tool_calls
-        tool_calls = [block for block in self.content if block["type"] == "tool_call"]
-        if tool_calls:
-            self._tool_calls = tool_calls
-        return [block for block in self.content if block["type"] == "tool_call"]
+        if not self._tool_calls:
```
eyurtsev, line 205: Does this need to be memoized? It's a very short list traversal

Is this due to how aggregation on message chunks works? If so, we should add an explanation here about why this is happening

libs/core/langchain_core/messages/v1.py
```diff
@@ -190,20 +202,28 @@ def text(self) -> Optional[str]:
...
     @tool_calls.setter
     def tool_calls(self, value: list[types.ToolCall]) -> None:
         """Set the tool calls for the AI message."""
         self._tool_calls = value
 
+    @property
+    def invalid_tool_calls(self) -> list[types.InvalidToolCall]:
+        """Get the invalid tool calls made by the AI."""
+        if not self._invalid_tool_calls:
+            self._invalid_tool_calls = [
```
eyurtsev, line 220: I don't thin kthere's a need to memoize it?

libs/core/langchain_core/messages/v1.py
```diff
@@ -228,24 +248,18 @@ class AIMessageChunk(AIMessage):
...
         content: Union[str, list[types.ContentBlock]],
+        *,
         id: Optional[str] = None,
         name: Optional[str] = None,
         lc_version: str = "v1",
         response_metadata: Optional[ResponseMetadata] = None,
         usage_metadata: Optional[UsageMetadata] = None,
         tool_call_chunks: Optional[list[types.ToolCallChunk]] = None,
         parsed: Optional[Union[dict[str, Any], BaseModel]] = None,
+        is_last_chunk: bool = False,
```
eyurtsev: Do we need to know the beginning of a stream?

libs/core/tests/unit_tests/output_parsers/test_base_parsers.py
```diff
@@ -116,4 +116,4 @@ def parse_result(
     model_v1 = GenericFakeChatModelV1(message_chunks=["hello", " ", "world"])
     chain_v1 = model_v1 | StrInvertCase()
     chunks = list(chain_v1.stream(""))
-    assert chunks == ["HELLO", " ", "WORLD"]
+    assert chunks == ["HELLO", " ", "WORLD", ""]
```
eyurtsev, line 119: why is there an extra `""` message

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
hntrl: do we need an `is_first_chunk`? as I understand it, "last chunk" is just the arbitrary point in time where we collapse tool call chunks into tool calls

libs/core/langchain_core/messages/v1.py
```diff
@@ -269,119 +284,75 @@ def __init__(
...
-                        create_tool_call(
-                            name=chunk.get("name") or "",
-                            args=args_,
-                            id=chunk.get("id", ""),
-                        )
-                    )
-                else:
-                    add_chunk_to_invalid_tool_calls(chunk)
-            except Exception:
-                add_chunk_to_invalid_tool_calls(chunk)
```
hntrl, line 346: is this the only condition we have for a tool call chunk to be considered invalid? is if we couldn't cast it to a partial json?

libs/core/langchain_core/messages/v1.py
```diff
@@ -396,49 +367,76 @@ def __add__(self, other: Any) -> "AIMessageChunk":
...
         return left
-    content = merge_content(
-        cast("list[str | dict[Any, Any]]", left.content),
-        *(cast("list[str | dict[Any, Any]]", o.content) for o in others),
+    content = cast(
+        "list[types.ContentBlock]",
+        merge_content(
+            cast("list[str | dict[Any, Any]]", left.content),
+            *(cast("list[str | dict[Any, Any]]", o.content) for o in others),
+        ),
```
hntrl, line 432: python types seem fun

libs/core/tests/unit_tests/output_parsers/test_base_parsers.py
ccurme: the new `AIMessageChunk([], is_last_chunk=True)` is parsed by `StrInvertCase` parser as a string.

hntrl (commented):
Is there any case that a message chunk can be casted to a message before the last chunk is received? That's the only thing that stands out that might make for awkward behavior

sidebar from this: it would be handy if there was a way we could directly concat a chunk to a message. Story i'm thinking of is frontend where I only need to concern myself with rendering an `AIMessage` (even if it's incomplete) that way I don't need to have a conditional case to render both a chunk + a message object. Relevant for python in the spirit of a standard data model/ usability of python with js frontend hooks.

libs/core/langchain_core/messages/v1.py
ccurme: it's to support the setter below it, lmk if there's a better way to do.

libs/core/langchain_core/messages/content_blocks.py
ccurme: it is getting the names of the block types. there's a test [here](https://github.com/langchain-ai/langchain/pull/32358/files#diff-dd757275bbf1398b2be91668a3af949f0dfd7b74eb72af8bdff2b08884ae4fb7R1371) that checks against an expected list of strings.

if we hard-code the list, we're liable to add a new type to the `ContentBlock` union and forget to update this set. but open to hard-coding, or lmk if I misunderstood you.

TBH I forget why we added this and am OK deleting it, it doesn't appear used. but feels helpful :)

libs/core/langchain_core/callbacks/manager.py
ccurme: None distinguishes between `[{"type": "text", "text": ""}]` and a list with no `text` blocks. idk if that is ever a useful distinction to make. If ergonomics of `.text` always being str are worth it I can update.

libs/core/langchain_core/messages/content_blocks.py
eyurtsev: Imo we should put this logic in the unit test and in the prod path put the string literals -- i.e. same logic but flipped

libs/core/langchain_core/callbacks/manager.py
ccurme: changed this to always return str

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
ccurme: updated this to `chunk_position: Optional[Literal["last"]] = None` for now.

libs/core/langchain_core/messages/v1.py
ccurme: I believe so, yes. InvalidToolCall just captures invalid json args.

ccurme:
> Is there any case that a message chunk can be casted to a message before the last chunk is received? That's the only thing that stands out that might make for awkward behavior
> 
> sidebar from this: it would be handy if there was a way we could directly concat a chunk to a message. Story i'm thinking of is frontend where I only need to concern myself with rendering an `AIMessage` (even if it's incomplete) that way I don't need to have a conditional case to render both a chunk + a message object. Relevant for python in the spirit of a standard data model/ usability of python with js frontend hooks.

@hntrl for now I am not casting to an AIMessage, just parsing tool_call_chunks into tool_calls. AIMessageChunk currently inherits AIMessage, so I didn't see much value and did see risk as we cannot continue adding after we convert to AIMessage..

PR #32372, opened 2025-08-03
fix(core): add missing type hints for `@chain` decorator with `RunnableConfig`

hqkqn32:
Closes #32223

---

`@chain` is a convenience wrapper around `RunnableLambda`, but its type annotations did not include all callable signatures that `RunnableLambda` already supports at runtime. As a result, users could pass a `config`-aware callable directly to `RunnableLambda` without type errors, while the equivalent `@chain` usage produced a false-positive mypy error.

This updates `chain()`'s overloads to match the supported `RunnableLambda` callable forms, including functions that accept `config`, `run_manager`, both parameters together, async variants, and functions that return a `Runnable`.

The user story is: a developer should be able to write a small configurable runnable with `@chain` and access `RunnableConfig` inside the decorated function, just as they can when constructing `RunnableLambda` directly.

This appears to have been missed when the underlying `RunnableLambda` callable support and typing were added. There is no evidence that `@chain` was intended to intentionally exclude these supported signatures; `chain()` simply delegates to `RunnableLambda(func)`.

Tests were added to cover the supported `@chain` callable shapes and to verify existing behavior continues to work.


hqkqn32:
is there any problem? @eyurtsev

libs/core/tests/unit_tests/runnables/test_chain_decorator.py
```diff
@@ -0,0 +1,42 @@
...
+
+from langchain_core.messages import AIMessage
+from langchain_core.prompt_values import ChatPromptValue
+from langchain_core.runnables import RunnableConfig, chain
+
+
+class TestChainDecorator:
+    """Test @chain decorator with RunnableConfig parameter."""
+
+    def test_chain_with_config_parameter(self) -> None:
```
cbornet: I would move this test to `test_runnable.py`

mdrxy:
Closing this for now because the change is typing-only and the proposed overloads need more work before they are safe to merge.

The runtime behavior already works through `RunnableLambda`: `config` and `run_manager` are injected via `accepts_config` / `accepts_run_manager`, and `chain()` delegates directly to `RunnableLambda(func)`. The remaining question is whether `chain()` exposes that behavior accurately to static type checkers.

The current version has a few blockers:

- Some overloads are shadowed by earlier generic `Callable[[Input], Output]` overloads. In particular, functions returning `Runnable[...]` and async functions accepting `config` reveal incorrect output types under mypy.
- The PR does not add type-level tests, so the core fix is not guarded. The runtime tests would still pass if the relevant overloads were removed or misordered.
- The new two- and three-argument callable overloads imply positional support for arbitrary parameter names, but runtime injection depends on the exact keyword names `config` and `run_manager`.

This still looks like a real issue and likely an omission in `@chain`'s typing rather than an intentional exclusion, but the fix should be revisited with:

- overload ordering that mirrors `RunnableLambda.__init__`,
- keyword-name-aware callable protocols for `config` / `run_manager`, and
- explicit typing tests using `assert_type` / mypy coverage for the advertised cases.

Thanks for the contribution — this is worth revisiting with those typing guarantees in place.

PR #32390, opened 2025-08-04
chore(core): bump mypy version to 1.17

cbornet:


libs/core/langchain_core/runnables/fallbacks.py
```diff
@@ -447,7 +447,7 @@ async def abatch(
         if not return_exceptions and sorted_handled_exceptions:
             raise sorted_handled_exceptions[0][1]
         to_return.update(handled_exceptions)
-        return [output for _, output in sorted(to_return.items())]  # type: ignore[misc]
+        return [cast("Output", output) for _, output in sorted(to_return.items())]
```
cbornet, line 450: The signature of `abatch` is `Output` but if `return_exceptions` is true, it is actually `Output | Exception`.
Should this be fixed ?

cbornet, line 450: The signature of `abatch` is `Output` but if `return_exceptions` is true, it is actually `Output | Exception`.
Should this be fixed ?

PR #32396, opened 2025-08-04
feat(groq): add support for `json_schema`

janzheng:
**Description:** Added support for Groq's new `json_mode` for constrained outputs / structured outputs; these ONLY pertain to certain models, but improves consistency dramatically. Docs here: https://console.groq.com/docs/structured-outputs. Also replaced the default Gemma2 test with a `llama-3.3-70b-versatile` test since Gemma2 is deprecated. Also bumped the `groq-sdk` to `0.30.0`

**Dependencies:** groq python sdk: https://pypi.org/project/groq/

**Tests and Docs:** No new tests have been added, but `test_chat_models.py` default model has been updated to `moonshotai/kimi-k2-instruct`. Tests are passing.


ccurme (approved):

ccurme:
Thanks for this. LangChain convention is to use `method="json_schema"` for dedicated structured output features. I've made that update and added support for TypedDict schemas.

PR #32397, opened 2025-08-04
feat(core): support returning v1 ToolMessage in tools

ccurme:
N.B. we may not need this in create_react_agent as we can infer output type from input messages.

### DIFF
--- a/libs/core/tests/unit_tests/test_tools.py
+++ b/libs/core/tests/unit_tests/test_tools.py
@@ -1381,40 +1381,30 @@ def test_tool_annotated_descriptions() -> None:
     }
 
 
-def test_tool_call_input_tool_message_output() -> None:
+@pytest.mark.parametrize("output_version", ["v0", "v1"])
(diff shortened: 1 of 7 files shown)


libs/core/tests/unit_tests/test_tools.py
```diff
@@ -1379,17 +1381,28 @@ def foo(
...
         "id": "123",
         "type": "tool_call",
     }
-    tool = _MockStructuredTool()
-    expected = ToolMessage(
-        "1 True {'img': 'base64string...'}", tool_call_id="123", name="structured_api"
-    )
+    tool = _MockStructuredTool(output_version=output_version)
+    if output_version == "v0":
+        expected: Union[ToolMessage, ToolMessageV1] = ToolMessage(
```
mdrxy, line 1394: why Union?

ccurme: otherwise mypy complains since I'm re-using the `expected` variable

PR #32401, opened 2025-08-05
fix(standard-tests): Add possibility to skip retriever tests that use the k arg

cbornet:
It's usual for retrievers to support a `k` arg in their constructor or in their `invoke` method but it is not a requirement as per the API, the usage of retrievers in  the code, and the docs.
So the tests in `RetrieversIntegrationTests` that relate to the `k` arg should be optional.


cbornet:
I found that the test for non overrides (`test_no_overrides_DO_NOT_OVERRIDE`) allowed to override with `pytest.mark.xfail`. I'll do another PR to hint to this solution instead in the pydoc.

PR #32408, opened 2025-08-05
fix: Support no-args tools by defaulting args to empty dict

jitokim:
Description:  
This PR ensures that tool calls without explicitly provided `args` will default to an empty dictionary (`{}`), allowing tools with no parameters (e.g. `def foo() -> str`) to be registered and invoked without validation errors. This change improves compatibility with agent frameworks that may omit the `args` field when generating tool calls.

Issue:  
See [langgraph#5722](https://github.com/langchain-ai/langgraph/issues/5722) –  
LangGraph currently emits tool calls without `args`, which leads to validation errors  
when tools with no parameters are invoked. This PR ensures compatibility by defaulting  
`args` to `{}` when missing.

Dependencies:  
None


eyurtsev (changes requested):
Hi @jitokim we should figure out what's generating the incorrect tool_call["args"] rather than try to patch all the places in the code that behave correctly

If I'm not mistaken all of these places behave correctly with respect to the schema of ToolCall. And the schema of ToolCall is correct -- it can represent parameterless calls with an empty dictionary.

jitokim:
Hi @eyurtsev .
Thanks for the review — I agree that it’s better to fix the root cause rather than applying broad exception handling.

I’ll create a minimal reproducible example to trace where the incorrect tool_call["args"] is being generated, and update the PR accordingly.

Appreciate the guidance 😃

jitokim:
Hi @eyurtsev .
I rolled back the previous code and made a new commit.
This is the result of running the reproduction code from [langgraph#5722](https://github.com/langchain-ai/langgraph/issues/5722)

```
import os
from langgraph.graph import StateGraph, START, END, MessagesState
from langgraph.prebuilt import ToolNode, tools_condition


llm = ChatOpenAI(
    model="gpt-4o",
    api_key=os.environ["OPENAI_API_KEY"],
    streaming=True
)

@tool
def current_user_info() -> str:
    """get current user info"""

    return '{"name": "test", "age": 18}'

tools = [current_user_info]


# Define the two nodes we will cycle between
async def call_model(state: MessagesState):
    """Call the LLM powering our "agent"
    """

    model = llm.bind_tools(tools)
    # Get the model's response
    response = await model.ainvoke([*state['messages']])
    return {"messages": [response]}



builder = StateGraph(MessagesState)

builder.add_node(call_model)
builder.add_node("tools", ToolNode(tools))
builder.add_edge(START, "call_model")
builder.add_conditional_edges(
    "call_model",
    tools_condition,
    {"tools": "tools", END: END}
)
builder.add_edge("tools", "call_model")

graph = builder.compile(name="Test Agent")


async def main():
    """Main function to test the reproduce scenario."""

    print("🔍 Testing tool calls with no arguments...")
    print("This should demonstrate the fix for None args in streaming scenarios.")
    print("=" * 60)

    # Create a message that should trigger the tool call
    initial_message = HumanMessage(content="Please get my current user info")

    try:
        # Run the graph
        result = await graph.ainvoke({"messages": [initial_message]})

        print("✅ Graph execution completed successfully!")
        print(f"Final messages count: {len(result['messages'])}")

        # Print the conversation
        for i, msg in enumerate(result['messages']):
            print(f"\n📝 Message {i+1} ({type(msg).__name__}):")
            if hasattr(msg, 'content'):
                print(f"   Content: {msg.content}")
            if hasattr(msg, 'tool_calls') and msg.tool_calls:
                print(f"   Tool calls: {msg.tool_calls}")
            if hasattr(msg, 'tool_call_id'):
                print(f"   Tool call ID: {msg.tool_call_id}")

    except Exception as e:
        print(f"❌ Error during execution: {e}")
        print(f"Error type: {type(e).__name__}")
        import traceback
        traceback.print_exc()


if __name__ == "__main__":
    import asyncio
    asyncio.run(main())

```

jitokim:
Hi @mdrxy , I've added a commit that addresses the root cause of the issue.  
Whenever you have time, I'd appreciate it if you could review the PR. 🙇  
Please let me know if there's anything else that needs to be updated!

overfiter:
I encountered the same problem

PR #32414, opened 2025-08-05
feat(groq): implement smart buffering for empty content chunks during streaming

mdrxy:
Smart buffering to handle empty content chunks during streaming operations. This addresses issues with certain Groq models (like `openai/gpt-oss-20b`) that return multiple empty chunks while still reporting token usage.

### Problem
- Models like `openai/gpt-oss-20b` generate multiple empty content chunks (`["''", "''"]`) during streaming
- These empty chunks create noise in the streaming output while preserving important metadata
- Existing streaming logic yielded every chunk, including consecutive empty ones

### Solution
- Buffer empty chunks and accumulate their metadata instead of yielding immediately
- Combine `generation_info`, `usage_metadata`, and other important chunk data
- Still yield buffered chunks when `finish_reason` is present
- Working models continue to function normally


mdrxy:
Will re-open if it becomes motivated by real use

PR #32418, opened 2025-08-05
release: 0.4

ccurme:


docs/api_reference/create_api_rst.py
```diff
@@ -217,7 +217,11 @@ def _load_package_modules(
         # Get the full namespace of the module
         namespace = str(relative_module_name).replace(".py", "").replace("/", ".")
         # Keep only the top level namespace
-        top_namespace = namespace.split(".")[0]
+        # (but make special exception for content_blocks and messages.v1)
+        if namespace == "messages.content_blocks" or namespace == "messages.v1":
```
mdrxy: This likely needs to updated after namespace refactor from `messages.v1` to `v1.messages`

libs/partners/openai/pyproject.toml
```diff
@@ -56,6 +56,8 @@ langchain-tests = { path = "../../standard-tests", editable = true }
 
 [tool.mypy]
 disallow_untyped_defs = "True"
+disable_error_code = ["typeddict-unknown-key"]
```
mdrxy, line 59: Is this better than per-line ignores?

PR #32443, opened 2025-08-07
feat(standard-tests): add a property to set the name of the parameter for the number of results to return

cbornet:
Not all retrievers use `k` as param name to set the number of results to return. Even in LangChain itself. Eg: https://github.com/langchain-ai/langchain/blob/bc4251b9e0074faf852c9b4e184ba681a594b03e/libs/core/langchain_core/indexing/in_memory.py#L31

So it's helpful to be able to change it for a given retriever.
The change also adds hints to disable the tests if the retriever doesn't support setting the param in the constructor or in the invoke method (for instance, the `InMemoryDocumentIndex` in the link supports in the constructor but not in the invoke method).

This change is backward compatible.


mdrxy (approved):
Thank you!

PR #32451, opened 2025-08-07
fix(openai): don't crash get_num_tokens_from_messages on gpt-5

Twixes:
- **Description:** The GPT-5 family just went live, but `model.get_num_tokens_from_messages()` crashes with those models (`gpt-5`, `gpt-5-mini`, `gpt-5-mini`) with `get_num_tokens_from_messages() is not presently implemented for model ...`. It seems we can assume the same encoding mechanism as the GPT-4 series. There's no _verified_ info from OpenAI on this that I've found, but it's better to be slightly off here, than not support the models at all. It's an issue we're seeing in https://github.com/PostHog/posthog.
  - **Issue:** PRs > issues
  - **Dependencies:** none
  - **Twitter handle:** @MatlokaM


ccurme (approved):

ccurme:
Thank you!

PR #32455, opened 2025-08-07
feat(openai): `minimal` and `verbosity`

mdrxy:


libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -418,7 +418,7 @@ class MakeASandwich(BaseModel):
 
 
 def test_tool_use() -> None:
-    llm = ChatOpenAI(model="gpt-4-turbo", temperature=0)
+    llm = ChatOpenAI(model="gpt-5", temperature=0)
```
ccurme: do the tests pass with gpt-5-nano?

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -875,7 +871,7 @@ def test_audio_output_modality() -> None:
 
 def test_audio_input_modality() -> None:
     llm = ChatOpenAI(
-        model="gpt-4o-audio-preview",
+        model="gpt-5",
```
ccurme: does this work?

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -1041,8 +1041,32 @@ def test_o1_stream_default_works() -> None:
...
 
+@pytest.mark.parametrize("use_responses_api", [False, True])
+@pytest.mark.parametrize("use_max_completion_tokens", [True, False])
+def test_minimal_reasoning_effort(
+    use_max_completion_tokens: bool, use_responses_api: bool
+) -> None:
+    if use_max_completion_tokens:
+        kwargs: dict = {"max_completion_tokens": MAX_TOKEN_COUNT}
+    else:
+        kwargs = {"max_tokens": MAX_TOKEN_COUNT}
```
ccurme: what does this have to do with minimal reasoning effort?

maybe subjective but I'd just make a unit test that instantiates `ChatOpenAI` and calls `_get_request_payload`, making 4 calls to reasoning models for this test seems heavy.

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -1041,8 +1041,32 @@ def test_o1_stream_default_works() -> None:
...
+    response = ChatOpenAI(
+        model="gpt-5",
+        reasoning_effort="minimal",
+        use_responses_api=use_responses_api,
```
ccurme:
```suggestion
        output_version="responses/v1",
```
(shouldn't make a difference here but we should just use this everywhere for the responses API)

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1178,6 +1210,13 @@ def _get_request_payload(
             kwargs["stop"] = stop
 
         payload = {**self._default_params, **kwargs}
+
+        if self.model_name.startswith("gpt-5"):
+            # gpt-5 models don't support these parameters
```
ccurme: (here and above) I think it's OK to let openai raise BadRequestError or whatever it is instead of silently dropping these.

I don't think cataloging what models maintain what features and warning / dropping parameters is worthwhile— it's also not forwards compatible if support is added later.

Temperature may be an exception since it's in the LS playground and is common. Actually ChatOpenAI had a default temperature of 0.7 in langchain-openai < 0.3, so `o1` broke out of the box, that's why we had the fix here.

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -527,6 +527,17 @@ class BaseChatOpenAI(BaseChatModel):
 
     .. versionadded:: 0.3.24
 
+    """
+    verbosity: Optional[str] = None
+    """Controls the verbosity level of responses for reasoning models. For use with the
+    Responses API.
```
ccurme, line 533: does this work in the Responses API?

ccurme (approved):
Looks good although not sure we should be dropping params (audio, modalities, etc). BadRequestError or whatever is raised might be preferable.

libs/partners/openai/langchain_openai/chat_models/base.py
mdrxy: works with some reasoning models (o1) in regular Chat API but gets rejected by Responses API

will add TODO test and xfail for now

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -2366,7 +2394,11 @@ class GetPopulation(BaseModel):
 
             from langchain_openai import ChatOpenAI
 
-            llm = ChatOpenAI(model="gpt-4.1-mini", use_responses_api=True)
+            llm = ChatOpenAI(
+                model="gpt-4.1-mini",
+                use_responses_api=True,
+                output_version="responses/v1",
```
ccurme, line 2400: we don't need both of these, just `output_version="responses/v1"`

libs/partners/openai/langchain_openai/chat_models/base.py
ccurme: docstring should probably not specify "For use with the Responses API" in the meantime.

PR #32456, opened 2025-08-07
chore: formatting across codebase

mdrxy:
To prevent polluting future PRs


ccurme (commented):
Why not do this on `master`, then merge in? This makes it much harder to review the v0.4 branch.

PR #32461, opened 2025-08-08
fix(anthropic): incorrect `input_tokens` counting while streaming

adityamohta:
TLDR: The lanchain_anthropic client is reporting incorrect input_tokens usage when used with any tools/mcp-servers while streaming, it also doesn't report any input_cache_read/input_cache_creation token usage at all. This PR fixes it.

## Details

Currently the input_tokens are counted in the `message_start` event, which is inaccurate. 

We can never count input tokens accurately at the start of the anthropic response. This is true when we use any tools/mcp, which increases the input_tokens count right after a tool call.

Additionally, we should only emit token usage at the end of the streaming anyway, because Anthropic doesn't charge for failed requests. So, streaming user input's token usage at the start of message doesn't make sense.

From Anthropic docs: https://support.anthropic.com/en/articles/8977456-how-do-i-pay-for-my-api-usage
> Failed requests are not charged, and you will only be billed for successful API calls and completed tasks.

---

Therefore, this PR, fixes the logic, and only emit the input_tokens details with "message_delta" events. 
I have also modified it to use the existing `_create_usage_metadata` method (it is a well tested function, [test case here](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/tests/unit_tests/test_chat_models.py#L983)), which will also emit the input cached read/write tokens correctly (they were not being emitted at all before)

### DIFF
--- a/libs/partners/anthropic/langchain_anthropic/chat_models.py
+++ b/libs/partners/anthropic/langchain_anthropic/chat_models.py
@@ -2065,20 +2065,12 @@ def _make_message_chunk_from_anthropic_event(
     message_chunk: Optional[AIMessageChunk] = None
     # See https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/streaming/_messages.py  # noqa: E501
     if event.type == "message_start" and stream_usage:
-        usage_metadata = _create_usage_metadata(event.message.usage)
-        # We pick up a cumulative count of output_tokens at the end of the stream,
-        # so here we zero out to avoid double counting.
-        usage_metadata["total_tokens"] = (
-            usage_metadata["total_tokens"] - usage_metadata["output_tokens"]
-        )
-        usage_metadata["output_tokens"] = 0
         if hasattr(event.message, "model"):
             response_metadata = {"model_name": event.message.model}
         else:
             response_metadata = {}
         message_chunk = AIMessageChunk(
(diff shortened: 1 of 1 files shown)


adityamohta:
Hi @mdrxy 

I have recently started using Langchain and absolutely loving it! Thank you for building & maintaining it! 

I am running into this issue, which is currently forcing me to run a forked version in one of our service. It would be great if it could be fixed upstream.

I completely understand how valuable your time is, so if there’s anything I can do to help make the review process easier or faster, please let me know. Thanks again for all your hard work! 🙇‍♂️

mdrxy:
Investigating.

> it also doesn't report any input_cache_read/input_cache_creation token usage at all

Regarding this, are you setting `cache_control={"type": "ephemeral"}` on your final message?

For example,

```python

llm = ChatAnthropic(model="claude-3-7-sonnet-20250219")

messages = [
    {
        "role": "system",
        "content": [
            {
                "type": "text",
                "text": "Below is some long context:",
            },
            {
                "type": "text",
                "text": f"{long_text}",
            },
        ],
    },
    {
        "role": "user",
        "content": "What's that about?",
    },
]

response = llm.invoke(messages)
response.usage_metadata["input_token_details"]
```

```python
{'cache_read': 0, 'cache_creation': 1458}
```

We're also making it such that you can provide `cache_control` at invocation time in the next `langchain-anthropic` release.

PR #32479, opened 2025-08-09
fix(core): fix PydanticOutputParser's `get_format_instructions` for v1 models

cbornet:


libs/core/tests/unit_tests/output_parsers/test_pydantic_parser.py
```diff
@@ -185,6 +183,14 @@ class SampleModel(BaseModel):
     }
 
 
+@pytest.mark.parametrize("pydantic_object", [ForecastV2, ForecastV1])
+def test_format_instructions(pydantic_object: TypeBaseModel) -> None:
```
cbornet, line 196: This test would fail if you removed the fix in `get_format_instructions`

libs/core/langchain_core/output_parsers/pydantic.py
mishra-krishna, line 1: This looks like a good fix. The new test case is also a great addition to ensure this doesn't break in the future.

I'm assuming that the `_get_schema` method is defined elsewhere and that it correctly handles both Pydantic v1 and v2 models. Can you confirm this?

mishra-krishna (commented):
This looks like a good fix. The new test case is also a great addition to ensure this doesn't break in the future.

I'm assuming that the `_get_schema` method is defined elsewhere and that it correctly handles both Pydantic v1 and v2 models. Can you confirm this?

cbornet:
> I'm assuming that the _get_schema method is defined elsewhere and that it correctly handles both Pydantic v1 and v2 models. Can you confirm this?

Please don't use AI without enough context and check the code by yourself. This is just noise.
We already have Copilot for AI reviews.

cbornet:
Hello, could this be reveiwed ?
This fix is needed to fix another issue with typing where `PydanticBaseModel` is `BaseModel` where it should be `BaseModel | BaseModelV1`.
It would probably be good to fix it before v1 release.
@mdrxy @eyurtsev

cbornet:
Pinging again as v1 release is close!
Can this have some eyes ?

ccurme (commented):
@cbornet is it correct to say that this update is non-breaking, as we are extending support of a method to pydantic v1 models?

I don't see any harm to this, but also don't see any urgency as we are no longer claiming full support for pydantic v1.

cbornet:
Yes that's non breaking.
It's actually fixing something that would raise an exception if used with a v1 model.

ccurme (approved):

PR #32480, opened 2025-08-09
fix(core): resolve for including "default" field value when deref is needed

hugehoo:
- **Description:** this pr fixes the json schema to have default field value by adding additional props in `_dereference_refs_helper` method.
- **Issue:** the issue #32232 it fixes

### Example
this code is based on issue `#32232`
``` python
import json
from enum import Enum

class Status(Enum):
    PENDING = "pending"
    COMPLETED = "completed"
    ERROR = "error"


@tool
def a_test_tool(status: Status = Status.PENDING) -> str:
    """
    A test tool that accepts a status and returns a string representation of it.
    """
    return f"Status is: {status.value}"


print(json.dumps(convert_to_openai_tool(a_test_tool), indent=2))
```

if the default Status is Pending, now the result has default props in json result.
``` json
{
  "type": "function",
  "function": {
    "name": "a_test_tool",
    "description": "A test tool that accepts a status and returns a string representation of it.",
    "parameters": {
      "properties": {
        "status": {
          "enum": [
            "pending",
            "completed",
            "error"
          ],
          "type": "string",
          "default": "pending"
        }
      },
      "type": "object"
    }
  }
}

```


libs/core/langchain_core/utils/json_schema.py
mishra-krishna, line 1: The fix is correct, but the implementation can be slightly improved for efficiency. Instead of creating a new dictionary `additional_props`, you can directly iterate over the items of `obj` and update `target`. This avoids creating an intermediate dictionary.

```python
# Suggested change:
for key, value in obj.items():
    if key != "$ref":
        target[key] = value
```

mishra-krishna (commented):
The fix is correct, but the implementation can be slightly improved for efficiency. Instead of creating a new dictionary `additional_props`, you can directly iterate over the items of `obj` and update `target`. This avoids creating an intermediate dictionary.

```python
# Suggested change:
for key, value in obj.items():
    if key != "$ref":
        target[key] = value
```

libs/core/langchain_core/utils/json_schema.py
hugehoo: thx for the review, i just updated as you suggested

hugehoo:
@mishra-krishna hello, i just wonder this pr can be reviewed :)  thx

eyurtsev:
Hi @hugehoo thanks for the contribution. I'm going to take over the issue and will close this PR. Generally guidelines for PRs are to include unit tests that verify the fix.

hugehoo:
@eyurtsev so if i added test code, this pr wouldn't be closed?

PR #32487, opened 2025-08-10
fix(core): fix support of Pydantic v1 models in BaseTool.args

cbornet:


libs/core/tests/unit_tests/test_tools.py
```diff
@@ -1861,6 +1861,11 @@ def _run(self, *args: Any, **kwargs: Any) -> str:
         name="some_tool", description="some description", args_schema=pydantic_model
     )
 
+    assert tool.args == {
```
cbornet, line 1907: Without the fix in `BaseTool.args` this fails

libs/core/langchain_core/tools/structured.py
```diff
@@ -67,16 +67,6 @@ async def ainvoke(
 
     # --- Tool ---
 
-    @property
-    def args(self) -> dict:
```
cbornet, line 71: Same as parent class.

libs/core/langchain_core/tools/base.py
mishra-krishna, line 1: Good catch! This fix is important for maintaining compatibility with Pydantic v1. The code simplification in the subclasses is also a nice improvement. Thanks for the contribution!

mishra-krishna (approved):
Good catch! This fix is important for maintaining compatibility with Pydantic v1. The code simplification in the subclasses is also a nice improvement. Thanks for the contribution!

libs/core/langchain_core/tools/base.py
```diff
@@ -543,6 +543,8 @@ def args(self) -> dict:
         """
         if isinstance(self.args_schema, dict):
             json_schema = self.args_schema
+        elif self.args_schema and issubclass(self.args_schema, BaseModelV1):
```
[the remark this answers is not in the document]
cbornet: BaseModelV1 is imported.

eyurtsev:
Thank you @cbornet

PR #32497, opened 2025-08-11
fix(core): fix beta decorator for properties

cbornet:


libs/core/tests/unit_tests/_api/test_beta_decorator.py
```diff
@@ -91,8 +91,8 @@ def beta_staticmethod() -> str:
         """Original doc."""
         return "This is a beta staticmethod."
 
+    @beta()  # type: ignore[prop-decorator]
```
cbornet, line 94: order of decorators matters.
We also need to silence `prop-decorator`. See https://github.com/python/mypy/issues/14461 . This should be OK since `beta` preserves the type of the property.

mdrxy (approved):

PR #32518, opened 2025-08-12
fix(anthropic): streaming token counting to defer input tokens until completion

mdrxy:
Supersedes #32461

Fixed incorrect input token reporting during streaming when tools are used. Previously, input tokens were counted at `message_start` before tool execution, leading to inaccurate counts. Now input tokens are properly deferred until `message_delta` (completion), aligning with Anthropic's billing model and SDK expectations.

**Before Fix:**
- Streaming with tools: Input tokens = 0 ❌
- Non-streaming with tools: Input tokens = 472 ✅

**After Fix:**
- Streaming with tools: Input tokens = 472 ✅
- Non-streaming with tools: Input tokens = 472 ✅

Aligns with Anthropic's SDK expectations. The SDK handles input token updates in `message_delta` events:

```python
# https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/streaming/_messages.py
if event.usage.input_tokens is not None:
      current_snapshot.usage.input_tokens = event.usage.input_tokens
```


mdrxy:
Unclear how this may affect LangSmith tracing

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -2270,11 +2314,30 @@ def _make_message_chunk_from_anthropic_event(
...
+                else None,
+            )
+            usage_metadata = _create_usage_metadata(combined_usage)
+        else:
+            # Fallback to just output tokens if no stored usage
+            usage_metadata = UsageMetadata(
+                input_tokens=0,
+                output_tokens=event.usage.output_tokens,
+                total_tokens=event.usage.output_tokens,
+            )
```
adityamohta: I read [here](https://docs.anthropic.com/en/docs/build-with-claude/streaming#event-types) that Anthropic sends cumulative token usage in `message_delta` events:
> The token counts shown in the usage field of the message_delta event are cumulative.

---

I created a [new change](https://github.com/langchain-ai/langchain/pull/32525), with a simpler solution which just always emit the usage received in the event as is to the consumer.

However, I am not aware of inner working of how langsmith reads these traces, so, in case it doesn't expect cumulative logs, then what you have here might be better.

I would only suggest one small change here is, replacing:

on line 2321
```python
                input_tokens=getattr(stored_input_usage, "input_tokens", 0),
```
with 
```python
                input_tokens=event.usage.input_tokens or getattr(stored_input_usage, "input_tokens", 0),
```

and 
on line 2337

```python
                input_tokens=0,
```
with 
```python
                input_tokens=event.usage.input_tokens or 0,
```

Because the input token count can get increased due to mcp tool calling, leading to `message_delta` events returning updated cumulative `input_tokens` count.

adityamohta (commented):
Thank you so much for taking out time to look into the issue @mdrxy I really appreciate it. 

I understand your approach, I have left a small comment, let me know if that doesn't make any sense 🙇‍♂️

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -70,6 +70,20 @@ class AnthropicTool(TypedDict):
     cache_control: NotRequired[dict[str, str]]
 
 
+class _CombinedUsage(BaseModel):
```
ccurme, line 73: Why create an extra BaseModel on each streaming event?

ccurme (commented):
Is this doing something different from reading the input tokens from the end of the stream, instead of the beginning?

ccurme:
This appears to break streaming out of the box with latest anthropic sdk

PR #32519, opened 2025-08-12
fix(core): Filter empty content blocks from formatted prompts

jacoblee93:


jacoblee93:
Failures seem unrelated

PR #32525, opened 2025-08-13
fix(anthropic): incorrect `input_tokens` counting while streaming

adityamohta:
Supersedes: https://github.com/langchain-ai/langchain/pull/32461

From docs: 

https://docs.anthropic.com/en/docs/build-with-claude/streaming#event-types
> The token counts shown in the usage field of the message_delta event are cumulative.

I realised my mistake in the previous PR, that the token counting is actually a cumulative sum. Therefore, langchain should emit the input_tokens count as is, if they exist in the event. 

And the responsibility can lie with the API consumer, to only pick up the max available value for any given token type.

### DIFF
--- a/libs/partners/anthropic/langchain_anthropic/chat_models.py
+++ b/libs/partners/anthropic/langchain_anthropic/chat_models.py
@@ -2065,12 +2065,20 @@ def _make_message_chunk_from_anthropic_event(
     message_chunk: Optional[AIMessageChunk] = None
     # See https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/streaming/_messages.py  # noqa: E501
     if event.type == "message_start" and stream_usage:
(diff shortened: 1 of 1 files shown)


mdrxy:
Thanks for the simpler approach! However, I'm closing this in favor of #32518 because this doesn't appear to handle the case where `message_delta` events have `input_tokens: null` during tool calls - which would result in 0 input tokens being reported instead of the correct count from `message_start`.

Happy to reopen if I've missed something - thanks again

adityamohta:
> Thanks for the simpler approach! However, I'm closing this in favor of #32518 because this doesn't appear to handle the case where `message_delta` events have `input_tokens: null` during tool calls - which would result in 0 input tokens being reported instead of the correct count from `message_start`.
> 
> Happy to reopen if I've missed something - thanks again

Yeah, makes sense! Thank you so much for all the time @mdrxy , I really appreciate it! 🙇‍♂️

PR #32529, opened 2025-08-13
chore(langchain): add mypy `warn_unreachable` setting

cbornet:
See https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_unreachable

### DIFF
--- a/libs/langchain/langchain/storage/_lc_store.py
+++ b/libs/langchain/langchain/storage/_lc_store.py
@@ -1,6 +1,6 @@
 """Create a key-value store for any langchain serializable object."""
 
-from typing import Callable, Optional
+from typing import Any, Callable, Optional
 
 from langchain_core.documents import Document
 from langchain_core.load import Serializable, dumps, loads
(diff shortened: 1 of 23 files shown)


mdrxy:
can you outline more what you mean by unreachable?

cbornet:
> can you outline more what you mean by unreachable?

I updated the title and description to be more clear.

libs/langchain/langchain/storage/_lc_store.py
```diff
@@ -14,7 +14,7 @@ def _dump_as_bytes(obj: Serializable) -> bytes:
     return dumps(obj).encode("utf-8")
 
 
-def _dump_document_as_bytes(obj: Document) -> bytes:
+def _dump_document_as_bytes(obj: Any) -> bytes:
```
mdrxy, line 17: Given that this will raise `TypeError` if not a `Document` it feels like the hint should remain even though not _technically_ correct - thoughts? Will this raise a mypy error?

cbornet: Yes I agree.
It will raise `unreachable` but we can put an ignore.

PR #32538, opened 2025-08-13
feat(core): enhance OpenAI data block conversion and extend support for file and audio types

mdrxy:


libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -568,30 +568,39 @@ def test_extend_support_to_openai_multimodal_formats() -> None:
         },
         {
             "type": "file",
-            "source_type": "base64",
-            "data": "<base64 string>",
```
ccurme, line 572: these cases should still pass

PR #32541, opened 2025-08-13
feat(core): allow overriding `ls_model_name` from kwargs

6matt:
Currently a runtime `model` override is not reflected in the LangSmith params
- `temperature` and `max_tokens` are already read from kwargs

Example to illustrate why this is needed:
```python
# User wants to compare responses from different models
llm = ChatAnthropic(model="claude-3-5-haiku-latest")

# Override at runtime for cost optimization
cheap_response = llm.invoke(messages,
model="claude-3-haiku-20240307")

# Override for complex tasks requiring better reasoning
complex_response = llm.invoke(messages,
model="claude-3-5-sonnet-20241022")

# or

def smart_llm_call(messages, complexity_score):
    llm = ChatOpenAI(model="gpt-4o-mini")  # Default cheaper model

    if complexity_score > 0.8:
        # Override for complex queries
            return llm.invoke(messages, model="gpt-4o")
      else:
            return llm.invoke(messages)  # Use default
```


mdrxy (approved):

eyurtsev:
@ccurme could you give this another look. Mason reviewed already and it looks fine to me from a quick glance, just not sure if there's anything else that could break

ccurme (approved):

PR #32560, opened 2025-08-15
chore(core): fix some mypy `warn_unreachable` issues

cbornet:
Found by setting `warn_unreachable: true` in mypy.


libs/core/langchain_core/utils/json_schema.py
```diff
@@ -3,23 +3,23 @@
...
         msg = (
             "ref paths are expected to be URI fragments, meaning they should start "
             "with #."
         )
         raise ValueError(msg)
-    out = schema
+    out: Union[list, dict] = schema
     for component in components[1:]:
-        if component in out:
+        if isinstance(out, dict) and component in out:
```
cbornet, line 22: A fix here: if out is a list that contains component, there will be an exception at the next line.

libs/core/langchain_core/utils/json_schema.py
```diff
@@ -3,23 +3,26 @@
...
         msg = (
             "ref paths are expected to be URI fragments, meaning they should start "
             "with #."
         )
         raise ValueError(msg)
-    out = schema
+    out: Union[list, dict] = schema
     for component in components[1:]:
         if component in out:
+            if isinstance(out, list):
```
cbornet, line 23: if `out` was a list, this was failing when calling `out[component]`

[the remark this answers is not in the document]
cbornet: Even if this use-case should not happen in practice, we need this for type checking.

mdrxy (approved):

Issue #31949, opened 2025-07-10
`set_llm_cache` does not apply when using `create_react_agent` from LangGraph prebuilt agents


valenvivaldi:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

``` python
from langchain.globals import set_llm_cache
from langgraph.prebuilt import create_react_agent
from langchain_community.cache import SQLiteCache
from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
from langchain_experimental.graph_transformers.llm import system_prompt
from langchain_openai import ChatOpenAI
import langchain
import time

SYSTEM_PROMPT_PLACEHOLDER = "{system_prompt}"

# Set up the cache
set_llm_cache(SQLiteCache(database_path="aaaaaacache.db"))

# Set up the LLM
llm = ChatOpenAI(model="gpt-4.1", temperature=0)

# Test with a simple prompt
prompt = "What is the capital of France?"
start_time = time.time()
response = llm.invoke(prompt)
print(f"First call: {response.content}")
print(f"Time: {time.time() - start_time} seconds")

start_time = time.time()
response = llm.invoke(prompt)
print(f"Second call: {response.content}")
print(f"Time: {time.time() - start_time} seconds")

# Tests with a ReAct agent
start_time = time.time()

agent = create_react_agent(llm, tools=[], prompt="You are a friendly agent")
response = agent.invoke({"messages": [{"role": "user", "content": "What is the capital of France?"}]})

print(f"Agent response: {response.get('messages')[1].content}")
print(f"Time: {time.time() - start_time} seconds")

start_time = time.time()
response = agent.invoke({"messages": [{"role": "user", "content": "What is the capital of France?"}]})
print(f"Agent response: {response.get('messages')[1].content}")
print(f"Time: {time.time() - start_time} seconds")

```

### Description

Description:
LangChain provides a function called set_llm_cache, which enables response caching for LLM provider calls. This works correctly when using the model directly — for example, calling ChatOpenAI().invoke() on the same prompt multiple times results in a cache hit after the first call.

[...]

PR #32572, opened 2025-08-15
refactor(core): data block handling, normalize message formats, strip IDs from messages

mdrxy:
> [!WARNING]
> **BREAKING:** Simplifies message normalization to single consistent path, requiring partner package updates

**Key Changes:**
- Consistent multimodal handling:
  - OpenAI `image_url` blocks pass through unchanged (broad compatibility)
  - OpenAI `input_audio` and `file` blocks convert to v1 standard equivalents
  - Legacy v0 multimodal blocks convert to v1 standard
  - Everything else passes through unchanged
- Partner packages must update content block parsing logic

**Partner Updates**

`output_version` affects how messages are serialized into `.content`. `_normalize_messages()` will now upgrade v0 content to v1, so, all partners now receive v1 format input regardless of `output_version`.

Migration:
- Partner packages must update to handle v1 input content blocks
- `output_version` still controls serialization format of responses (unchanged)


libs/core/langchain_core/language_models/_utils.py
```diff
@@ -57,43 +89,204 @@ def _parse_data_uri(uri: str) -> Optional[dict]:
...
+            base64=block["file"]["file_data"],
+            mime_type=mime_type,
+            filename=filename,
+            **all_extras,
+        )
+
+    # Escape hatch for non-standard content blocks
+    return create_non_standard_block(
+        value=block,
+    )
```
ccurme: should we just pass through here? docstring says "If parsing fails, passes block through unchanged."

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -57,43 +89,204 @@ def _parse_data_uri(uri: str) -> Optional[dict]:
...
-def _convert_openai_format_to_data_block(block: dict) -> dict:
-    """Convert OpenAI image content block to standard data content block.
+def _convert_openai_format_to_data_block(block: dict) -> ContentBlock:
+    """Convert OpenAI image/audio/file content block to respective v1 multimodal block.
+
+    We expect that the incoming block is verified to be in OpenAI Chat Completions
+    format.
 
-    If parsing fails, pass-through.
+    If parsing fails, passes block through unchanged.
```
ccurme: should return type not be `ContentBlock` in this case?

or have this or a helper return ContentBlock, then something else unpacks non-standard blocks?

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -57,43 +89,204 @@ def _parse_data_uri(uri: str) -> Optional[dict]:
...
-                parsed["filename"] = filename
-            return parsed
+                result["filename"] = filename
+            return result
         return block
 
     if block["type"] == "input_audio":
-        data = block["input_audio"].get("data")
-        audio_format = block["input_audio"].get("format")
+        audio = block.get("audio", {})
```
ccurme: why is this changing from `"input_audio"` to `"audio"`?

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -106,41 +299,270 @@ def _convert_openai_format_to_data_block(block: dict) -> dict:
...
+                    "dict",
+                    create_text_block(
+                        text=message.content,
+                    ),
+                ),
+            ]
+
+        elif isinstance(message.content, list):
             for idx, block in enumerate(message.content):
+                if isinstance(block, str):
```
ccurme: not sure it's necessary to convert all string content to text blocks here.

IMO scope of normalize should be unchanged and just convert to v1 instead of v0.
- image_url -> image_url
- input_audio -> v1 audio block instead of v0
- file -> v1 file block instead of v0

these would be untyped dicts (they are going in `.content` after all).

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -106,41 +299,270 @@ def _convert_openai_format_to_data_block(block: dict) -> dict:
...
+                    # unless `all` is True, in which case we also conver images
+                    if convert_all or block["type"] != "image_url":
+                        formatted_message.content[idx] = (  # type: ignore[call-overload, index]
+                            _convert_openai_format_to_data_block(block)
+                        )
+                    else:
+                        # If `all` is False, we pass through images unchanged
+                        formatted_message.content[idx] = block  # type: ignore[index]
+
+                # Convert LangChain v0 to v1 standard content blocks
```
ccurme: What is the difference between the logic here and the logic in `_convert_legacy_v0_content_block_to_v1` below?

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -167,6 +168,57 @@ def _format_for_tracing(messages: list[BaseMessage]) -> list[BaseMessage]:
     return messages_to_trace
 
 
+def _normalize_messages_for_cache(messages: list[BaseMessage]) -> list[BaseMessage]:
```
ccurme: (nit) would call this `_strip_ids_from_messages` or similar

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -167,6 +168,57 @@ def _format_for_tracing(messages: list[BaseMessage]) -> list[BaseMessage]:
...
+
+    for message in messages:
+        # Create a copy of the message
+        normalized_message = message.model_copy(deep=True)
+
+        # Remove auto-generated ID from message level
+        if (
+            hasattr(normalized_message, "id")
+            and normalized_message.id
+            and normalized_message.id.startswith(LC_AUTO_PREFIX)
```
ccurme: wondering if we should broaden this to resolve https://github.com/langchain-ai/langchain/issues/31949

libs/core/langchain_core/messages/base.py
```diff
@@ -129,19 +134,54 @@ def content_blocks(self) -> list[types.ContentBlock]:
...
+                    blocks.append(
+                        cast(
+                            "types.ContentBlock",
+                            {"type": "non_standard", "value": item},
+                        )
+                    )
+                    continue
                 if item_type not in types.KNOWN_BLOCK_TYPES:
+                    # Check if this is a single-key block that was normalized
+                    # (e.g., {"type": "cachePoint", "cachePoint": ...})
```
ccurme: the single-key normalization should only apply for tracing, it shouldn't impact any message content outside of that

what I'd expect to be in `.content` is the original `{"cachePoint": {"type": "default"}`, and this should become
`{"type": "non_standard": "value": {"cachePoint": {"type": "default"}}`.

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -106,41 +299,270 @@ def _convert_openai_format_to_data_block(block: dict) -> dict:
...
+                            id=block["id"],
+                        )
+                    elif block["source_type"] == "text" and block["type"] == "file":
+                        formatted_message.content[idx] = create_plaintext_block(  # type: ignore[call-overload, index]
+                            # In v0, URL points to the text file content
+                            text=block["url"],
+                        )
+                    continue
+
+                # Handle blocks with no type key but exactly one key
```
ccurme: this logic should only exist in `_format_for_tracing`.

that function and `_normalize_messages` are similar but different:
- `_format_for_tracing`: all v0, v1 images and PDFs should be traced in oai chat completions format. if they're missing, "type" keys should be added where possible.
- `_normalize_messages`: given a format for audio and pdf that chat models support (should be v1 IMO), cast oai chat completions audio and files to this format

libs/core/langchain_core/messages/base.py
```diff
@@ -129,19 +134,54 @@ def content_blocks(self) -> list[types.ContentBlock]:
...
             if isinstance(item, str):
+                # TODO: use factory here to include auto-ID generation?
                 blocks.append({"type": "text", "text": item})
             elif isinstance(item, dict):
                 item_type = item.get("type")
+                if item_type in {
+                    "image_url",
+                    "input_audio",
+                    "file",
+                } and _is_openai_data_block(item):
```
ccurme: can we refactor this logic into helper functions:
```python
def _convert_to_v1_from_chat_completions_input(content: list[types.ContentBlock]) -> list[types.ContentBlock]:

def _convert_to_v1_from_v0_multimodal_input(content: list[types.ContentBlock]) -> list[types.ContentBlock]:
```
so that we can do something like:
```python
@property
def content_blocks(self) -> list[types.ContentBlock]:
    blocks: list[types.ContentBlock] = []

    # First pass converting to standard blocks
    content = (
        [self.content]
        if isinstance(self.content, str) and self.content
        else self.content
    )
    for item in content:
        if isinstance(item, str):
            blocks.append({"type": "text", "text": item})
        elif isinstance(item, dict):
            item_type = item.get("type")
            if item_type not in types.KNOWN_BLOCK_TYPES:
                blocks.append({"type": "non_standard", "value": item})
            else:
                blocks.append(cast("types.ContentBlock", item))
        else:
            pass

    # Subsequent passes attempt to unpack non-standard blocks
    blocks = _convert_to_v1_from_v0_multimodal_input(blocks)
    blocks = _convert_to_v1_from_chat_completions_input(blocks)
    blocks = _convert_to_v1_from_anthropic_input(blocks)
    blocks = _convert_to_v1_from_gemini_input(blocks)
    ...

    return blocks
```
open to other ways of organizing this.

libs/core/tests/unit_tests/fake/test_fake_chat_model.py
```diff
@@ -212,12 +212,43 @@ async def on_llm_new_token(
 
 
 def test_chat_model_inputs() -> None:
-    fake = ParrotFakeChatModel()
+    fake = ParrotFakeChatModel(output_version="v1")
+
+    result = fake.invoke("hello")
+    # Remove the `id` from content blocks for comparison
```
ccurme: let's not normalize all string content to text blocks IMO.

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -461,19 +462,90 @@ def test_trace_images_in_openai_format() -> None:
...
             "source_type": "url",
             "url": "https://example.com/image.png",
         }
     ]
+    # But .content_blocks should perform v1 transformation
+    assert len(response.content_blocks) == 1
+    content_block = response.content_blocks[0]
+    if isinstance(content_block, dict) and "id" in content_block:
+        # Remove auto-generated id for comparison
+        content_without_id = {k: v for k, v in content_block.items() if k != "id"}
```
ccurme: helper to compare lists of content blocks agnostic of ID might be nice for tests

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -518,125 +589,219 @@ def test_trace_content_blocks_with_no_type_key() -> None:
...
-        {
-            "role": "user",
-            "content": [
-                {"type": "text", "text": "Hello"},
-                {
-                    "type": "image_url",
-                    "image_url": {"url": "https://example.com/image.png"},
+    """Test normalizing OpenAI audio, image, and file inputs to v1."""
+    # Audio and file only (chat model default)
+    messages = AIMessage(
```
ccurme: why change to AIMessage? these are typically (always?) input types

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -518,125 +589,219 @@ def test_trace_content_blocks_with_no_type_key() -> None:
...
+                "type": "image_url",
+                "image_url": {"url": "https://example.com/image.png"},
+            },
+            {  # image-base64
+                "type": "image_url",
+                "image_url": {"url": "data:image/jpeg;base64,/9j/4AAQSkZJRg..."},
+            },
+            {  # audio-base64
+                "type": "input_audio",
+                "audio": {
```
ccurme: why is input_audio changing to audio?

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -518,125 +589,219 @@ def test_trace_content_blocks_with_no_type_key() -> None:
...
+            {"type": "text", "text": "Hello"},
+            {  # audio-base64
+                "type": "input_audio",
+                "audio": {
+                    "format": "wav",
+                    "data": "data:audio/wav;base64,<base64 string>",
                 },
-                {
-                    "type": "image_url",
-                    "image_url": {"url": "data:image/jpeg;base64,/9j/4AAQSkZJRg..."},
```
ccurme: it looks like we are losing coverage on the condition that image_urls pass-through— this is important to keep

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -646,18 +811,49 @@ def test_normalize_messages_edge_cases() -> None:
...
+                    "base64": "base64encodedaudiodata",
+                    "mime_type": "audio/wav",
+                },
+                {
+                    "type": "file",
+                    "id": "file_123",
+                },
+                {
+                    "type": "reasoning",
+                    "text": "Let me think about this...",
```
ccurme: (nit) no content blocks that I know of (standard or provider-specific) has `"type": "reasoning", "text": "..."`

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -646,18 +811,49 @@ def test_normalize_messages_edge_cases() -> None:
                     "file_data": "uri",
                     "filename": "file-name",
                 },
+            ]
+        )
+    ]
+    assert expected_messages == _normalize_messages(input_messages)
```
ccurme: should we keep the test as `messages == _normalize_messages(messages)` for clarity?

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -701,33 +903,20 @@ async def test_output_version_ainvoke(monkeypatch: Any) -> None:
 def test_output_version_stream(monkeypatch: Any) -> None:
     messages = [AIMessage("foo bar")]
 
-    llm = GenericFakeChatModel(messages=iter(messages), output_version="v1")
```
ccurme, line 704: can we limit out-of-scope changes in PRs to ease review?

libs/core/tests/unit_tests/language_models/chat_models/test_rate_limiting.py
```diff
@@ -211,11 +213,16 @@ def test_rate_limit_skips_cache() -> None:
...
+        '"content": "foo"'
+        if output_version == "v0"
+        else '"content": [{"type": "text", "text": "foo"}]'
+    )
     assert list(cache._cache) == [
         (
-            '[{"lc": 1, "type": "constructor", "id": ["langchain", "schema", '
-            '"messages", '
-            '"HumanMessage"], "kwargs": {"content": "foo", "type": "human"}}]',
+            f'[{{"lc": 1, "type": "constructor", "id": ["langchain", "schema", '
```
ccurme: what's happening here?

ccurme (commented):
Let's assume that we will update all chat models that accept v0 mulltimodal blocks to take v1.

I'm not sure there's a way around this— if a user sets `output_version = "v1"`, then we will normalize to v1 types anyway, right? the purpose of `output_version` IMO is to control what is serialized, nothing should break if it's flipped.

So I think it's OK to not have a separate `_normalize_messages_v0` path. `_normalize_messages` should
- pass through image_url blocks
- update chat completions file and audio blocks to v1
- update v0 image, audio, file blocks to v1
- pass through everything else unchanged

libs/core/langchain_core/language_models/_utils.py
mdrxy: good catch

libs/core/langchain_core/language_models/chat_models.py
mdrxy: #32622

libs/core/langchain_core/messages/base.py
```diff
@@ -18,6 +23,94 @@
     from langchain_core.prompts.chat import ChatPromptTemplate
 
 
+def _convert_to_v1_from_chat_completions_input(
```
ccurme: can put this in block_translators/openai

libs/core/langchain_core/messages/base.py
```diff
@@ -134,18 +229,25 @@ def content_blocks(self) -> list[types.ContentBlock]:
...
             elif isinstance(item, dict):
                 item_type = item.get("type")
-                if item_type not in types.KNOWN_BLOCK_TYPES:
-                    msg = (
-                        f"Non-standard content block type '{item_type}'. Ensure "
-                        "the model supports `output_version='v1'` or higher and "
-                        "that this attribute is set on initialization."
+                # Check for v0 format blocks first
+                if item_type in {"image", "audio", "file"} and "source_type" in item:
+                    blocks.append(_convert_legacy_v0_content_block_to_v1(item))
```
ccurme: can we consolidate this logic into `_convert_legacy_v0_content_block_to_v1`?

the typing is wonky because below we cast to `ContentBlock` just based on the names of types.

but this is also true of anthropic: their image type is `image`, so typing-wise we're saying that's a standard block, but when we get to `_convert_to_v1_from_anthropic_input` we are looking for them in addition to `non_standard`.

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -57,90 +97,354 @@ def _parse_data_uri(uri: str) -> Optional[dict]:
...
-                        formatted_message.content = list(formatted_message.content)
-
-                    formatted_message.content[idx] = (  # type: ignore[index]  # mypy confused by .model_copy
-                        _convert_openai_format_to_data_block(block)
-                    )
+                    formatted_message = _ensure_message_copy(message, formatted_message)
+
+                    # Convert OpenAI audio/file block to LC v1 std content
+                    # unless `all` is True, in which case we also conver images
+                    if convert_all or block["type"] != "image_url":
```
ccurme: do we ever have `convert_all` True?

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -57,90 +97,354 @@ def _parse_data_uri(uri: str) -> Optional[dict]:
...
+                # Pass through blocks that look like they have v1 format unchanged
+                elif isinstance(block, dict) and block.get("type") in KNOWN_BLOCK_TYPES:
+                    # No conversion needed for v1 blocks, so no need to copy
+                    pass
+
+        # If we didn't modify the message, skip creating a new instance
+        # e.g. passing through content that is just `str`
+        if formatted_message is message:
+            formatted_messages.append(message)
+            continue
```
ccurme: given that we `.append` below, do we need this?

ccurme (commented):
Thanks for this, will look more tomorrow morning.

libs/core/langchain_core/language_models/_utils.py
mdrxy: Not yet, but gives a path forward in the future if needed

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -57,90 +96,348 @@ def _parse_data_uri(uri: str) -> Optional[dict]:
...
+    if (block["type"] == "image_url") and isinstance(
+        block["image_url"].get("url"), str
+    ):
+        known_keys = {"type", "image_url"}
+        extras = _extract_extras(block, known_keys)
+
+        image_url_known_keys = {"url"}
+        image_url_extras = _extract_extras(block["image_url"], image_url_known_keys)
+
+        all_extras = {**extras}
```
ccurme: (nit) why copy?

libs/core/langchain_core/language_models/_utils.py
```diff
@@ -57,90 +96,348 @@ def _parse_data_uri(uri: str) -> Optional[dict]:
...
+                    formatted_message = _ensure_message_copy(message, formatted_message)
+
+                    converted_block = _convert_legacy_v0_content_block_to_v1(block)
+                    _update_content_block(formatted_message, idx, converted_block)
+                    continue
+
+                # Pass through blocks that look like they have v1 format unchanged
+                elif isinstance(block, dict) and block.get("type") in KNOWN_BLOCK_TYPES:
+                    # No conversion needed for v1 blocks, so no need to copy
+                    pass
```
ccurme:
```suggestion
```
why have this condition?

libs/core/langchain_core/messages/base.py
```diff
@@ -18,6 +24,45 @@
     from langchain_core.prompts.chat import ChatPromptTemplate
 
 
+def _convert_to_v1_from_v0_multimodal_input(
```
ccurme: (nit) would put this in block translators

libs/core/langchain_core/messages/block_translators/openai.py
```diff
@@ -221,10 +278,11 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
                 ] = None
                 call_id = block.get("call_id", "")
                 if (
-                    isinstance(message, AIMessageChunk)
-                    and len(message.tool_call_chunks) == 1
+                    hasattr(message, "tool_call_chunks")
+                    and len(getattr(message, "tool_call_chunks", [])) == 1
```
ccurme: I protest

libs/core/langchain_core/language_models/fake_chat_models.py
```diff
@@ -363,7 +363,44 @@ def _generate(
         **kwargs: Any,
     ) -> ChatResult:
         """Top Level call."""
-        return ChatResult(generations=[ChatGeneration(message=messages[-1])])
+        # Get the last message to parrot
+        last_message = messages[-1]
+
+        # ParrotFakeChatModel is a special case that needs to return content
+        # in the format expected by output_version. Since input messages are
+        # normalized to v1 format, we need to convert back for v0 output.
```
ccurme: why is this useful to do?

ParrotFakeChatModel is useful for testing what makes it to the chat model. so it's useful for testing tracing because we can see what gets traced versus what goes into `_generate`.

if you want to test output_version, should we just use GenericFakeChatModel or similar?

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -167,6 +167,119 @@ def _format_for_tracing(messages: list[BaseMessage]) -> list[BaseMessage]:
     return messages_to_trace
 
 
+def _strip_ids_from_messages(
```
ccurme: can we do this in a separate PR? I can delete this function and the only thing that fails are the new tests in `test_strip_ids`.

if it's breaking, we can PR against the 1.0 branch.

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -461,19 +493,86 @@ def test_trace_images_in_openai_format() -> None:
...
-    # Test no mutation
+    # With v0 output_version, .content should preserve the original v0 format
     assert response.content == [
         {
             "type": "image",
             "source_type": "url",
             "url": "https://example.com/image.png",
         }
     ]
+    # But .content_blocks should perform v1 transformation
```
ccurme: this test can go in tests/unit_tests/messages/block_translators/test_openai.py

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -461,19 +493,86 @@ def test_trace_images_in_openai_format() -> None:
...
+        expected_content = {
+            "type": "image",
+            "url": "https://example.com/image.png",
+        }
+        assert content_without_id == expected_content
+    else:
+        assert content_block == {
+            "type": "image",
+            "url": "https://example.com/image.png",
+        }
```
ccurme: my sense is that the changes to ParrotFakeChatModel above stem from hijacking this test to test content block translation in addition to tracing.

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -508,156 +607,247 @@ def test_trace_content_blocks_with_no_type_key() -> None:
...
+                "extras": {"filename": "draconomicon.pdf"},
+            },
+            {  # ...
+                "type": "file",
+                "file_id": "<file id>",
+            },
+        ]
+    )
+
+    normalized_content = _normalize_messages([messages], convert_all=True)
```
ccurme: I would not add `convert_all` until we use it somewhere. right now it costs complexity for no benefit.

https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -679,18 +869,24 @@ def test_output_version_invoke(monkeypatch: Any) -> None:
...
+
+    # v1
     llm = GenericFakeChatModel(messages=iter(messages), output_version="v1")
     response = await llm.ainvoke("hello")
     assert response.content == [{"type": "text", "text": "hello"}]
     assert response.response_metadata["output_version"] == "v1"
 
-    llm = GenericFakeChatModel(messages=iter(messages))
-    response = await llm.ainvoke("hello")
-    assert response.content == "hello"
```
ccurme, line 692: (nit) ordering was deliberate here

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -701,33 +897,20 @@ async def test_output_version_ainvoke(monkeypatch: Any) -> None:
...
-
+    # v0
     llm = GenericFakeChatModel(messages=iter(messages))
     full = None
     for chunk in llm.stream("hello"):
         assert isinstance(chunk, AIMessageChunk)
         assert isinstance(chunk.content, str)
         assert chunk.content
-        full = chunk if full is None else full + chunk
+        full = chunk if full is None else full + chunk  # type: ignore[assignment]
```
ccurme: why is this type ignore now necessary?

unclear to me what improvements we're making to the tests here.

libs/core/langchain_core/language_models/_utils.py
mdrxy: ?

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
mdrxy: `tests/unit_tests/language_models/chat_models/test_base.py:895: error: Incompatible types in assignment (expression has type "BaseMessageChunk", variable has type "AIMessageChunk | None") [assignment]`

Issue #32562, opened 2025-08-15
`invalid_tool_calls` when using stream


uncle-yura:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

It is impossible to reproduce the bug with just a code example.

### Description

Changes from [PR](https://github.com/langchain-ai/langchain/pull/31587) added a function [_advance](https://github.com/langchain-ai/langchain/blob/4656f727da238c1fe24f95698842b94d579844c2/libs/partners/openai/langchain_openai/chat_models/base.py#L3995) to chunk processing that changes the index variable. This creates a bug that corrupts arguments for custom functions that are called after a document search.

The sequence of events to reproduce the bug:
1. Start a chat with the Responses API and the document search function
2. The model uses document search
3. The model calls the custom function and starts passing arguments:
`{"type": "response.function_call_arguments.delta", "output_index": 1, "raw": {"delta": " battery", "item_id": "..., "output_index": 1, "sequence_number": 165, "type": "response.function_call_arguments.delta", "obfuscation": "..."}
{"type": "response.function_call_arguments.delta", "output_index": 1, "raw": {"delta": " supply", "item_id": "....", "output_index": 1, "sequence_number": 166, "type": "response.function_call_arguments.delta", "obfuscation": "..."}
{"type": "response.function_call_arguments.delta", "output_index": 1, "raw": {"delta": " chain", "item_id": "....", "output_index": 1, "sequence_number": 167, "type": "response.function_call_arguments.delta", "obfuscation": "..."}`
4. At this point, the chunk index is 1
`{"name": null, "args": "-loop", "id": null, "index": 1, "type": "tool_call_chunk"}
{"name": null, "args": " supply", "id": null, "index": 1, "type": "tool_call_chunk"}
{"name": null, "args": " chain", "id": null, "index": 1, "type": "tool_call_chunk"}`
5. The model adds an annotation:
[...]

PR #32581, opened 2025-08-16
fix(langchain): prevent index desync causing `invalid_tool_calls`

AyhamJo7:
This fixes issue #32562 where response.output_text.annotation.added events were advancing the sub-index, causing function_call_arguments.delta chunks to get different indices and leading to invalid_tool_calls in OpenAI responses.

**Root Cause:**
The annotation event handler called _advance(output_index, content_index) with sub_idx, while function call argument deltas called _advance(output_index) without sub_idx. This caused current_index to increment mid-tool-call when annotations were interleaved with function arguments.

**Fix:**
Change annotation event to call _advance(output_index) without sub_idx, treating annotations as metadata on existing content rather than new blocks.

**Tests:**
- Added regression test test_tool_call_args_with_annotation_index_consistency
- Verifies tool call chunks maintain consistent indices
- Confirms invalid_tool_calls remains empty
- Ensures complete tool call argument reconstruction


libs/partners/openai/tests/unit_tests/chat_models/test_base.py
```diff
@@ -2734,3 +2734,116 @@ def test_gpt_5_temperature(use_responses_api: bool) -> None:
...
+            "response.function_call_arguments.delta", output_index=1, delta='{"param'
+        ),
+        # Second function argument delta
+        MockChunk(
+            "response.function_call_arguments.delta", output_index=1, delta='": "valu'
+        ),
+        # Annotation event - this advances sub-index in buggy version
+        MockChunk(
+            "response.output_text.annotation.added",
+            output_index=1,
```
ccurme, line 2785: shouldn't `output_index` be different, here?

annotations are not part of [function call items](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_function_tool_call.py).

ccurme (dismissed):
Thanks for this.

Is it actually the case that we **never** want to increment the index for annotations? They never represent a new content block, right?

So I think it's fine to no longer check the inner index, but I'm confused about the context in which this bug shows up— the test case shows function calls interleaved with annotations in a stream. Does this happen? Annotations aren't part of function calls, so are there message items we're omitting from the stream?

I've added a test directly constructing a Responses stream in case that is easier to work with.

ccurme (commented):
Can you clarify in what context you are seeing this?

Are you using OpenAI, or a proxy service?

This change assumes that we will never begin a text block with an annotation. This may or may not be the case— it doesn't seem typical. But the situation it's fixing does not make sense to me: in what context would we have annotations generated in the middle of streaming function call args?

uncle-yura:
@ccurme the bug can be seen if you run this example, but the model does not send annotations 100% of the time.
```
import operator
from typing import Annotated, TypedDict

from langgraph.graph import END, StateGraph
from langgraph.prebuilt import ToolNode


class AgentState(TypedDict):
    messages: Annotated[list, operator.add]


@tool
def show_results(text: str, return_direct: bool) -> int:
    """Displays results to the user"""
    return text


model = ChatOpenAI(
    temperature=1,
    model="gpt-4.1-2025-04-14",
    use_responses_api=True,
    streaming=True,
)
tools = [show_results]
tool_executor = ToolNode(tools)
model = model.bind_tools(
    [
        *tools,
        {
            "type": "file_search",
            "vector_store_ids": ["vs_...."],
        },
    ]
)


def should_continue(state):
    messages = state["messages"]
    last_message = messages[-1]

    if not last_message.tool_calls:
        return "end"
    else:
        arguments = last_message.tool_calls[0]["args"]
        if arguments.get("return_direct", False):
            return "final"
        else:
            return "continue"


def call_model(state):
    messages = state["messages"]
    response = model.invoke(messages)
    return {"messages": [response]}


workflow = StateGraph(AgentState)
workflow.add_node("agent", call_model)
workflow.add_node("action", tool_executor)
workflow.add_node("final", tool_executor)
workflow.set_entry_point("agent")
workflow.add_conditional_edges(
    "agent",
    should_continue,
    {
        "continue": "action",
        "final": "final",
        "end": END,
    },
)
workflow.add_edge("action", "agent")
workflow.add_edge("final", END)

app = workflow.compile()

message = {
    "role": "user",
    "content": [
        {
            "type": "text",
            "text": "Analyze this file and suggest the best camera for a novice. Also add output text annotations. Return this result directly by setting return_direct = True.",
        }
    ],
}

inputs = {"messages": [message]}

for mode, chunk in app.stream(inputs, stream_mode=["messages", "values"]):
    print(mode, str(chunk))
    print("\n")
```

Output:
```
messages (AIMessageChunk(content=[], additional_kwargs={'__openai_function_call_ids__': {}}, response_metadata={}, id='run--...', invalid_tool_calls=[{'name': None, 'args': '**', 'id': None, 'error': None, 'type': 'invalid_tool_call'}], tool_call_chunks=[{'name': None, 'args': '**', 'id': None, 'index': 1, 'type': 'tool_call_chunk'}]), {'langgraph_step': 1, 'langgraph_node': 'agent', 'langgraph_triggers': ('branch:to:agent',), 'langgraph_path': ('__pregel_pull', 'agent'), 'langgraph_checkpoint_ns': 'agent:..., 'checkpoint_ns': 'agent:...', 'ls_provider': 'openai', 'ls_model_name': 'gpt-4.1-2025-04-14', 'ls_model_type': 'chat', 'ls_temperature': 1.0})

ResponseOutputTextAnnotationAddedEvent(annotation={'type': 'file_citation', 'file_id': 'file-...', 'filename': 'Cameras.docx', 'index': 1600}, annotation_index=0, content_index=0, item_id='fc_...', output_index=1, sequence_number=375, type='response.output_text.annotation.added')

ResponseFunctionCallArgumentsDeltaEvent(delta='","', item_id='fc_..., output_index=1, sequence_number=376, type='response.function_call_arguments.delta', obfuscation='...')

messages (AIMessageChunk(content=[{'annotations': [{'type': 'file_citation', 'file_id': 'file-...', 'filename': 'Cameras.docx', 'index': 1600}], 'index': 2}], additional_kwargs={}, response_metadata={}, id='run--...'), {'langgraph_step': 1, 'langgraph_node': 'agent', 'langgraph_triggers': ('branch:to:agent',), 'langgraph_path': ('__pregel_pull', 'agent'), 'langgraph_checkpoint_ns': 'agent:...', 'checkpoint_ns': 'agent:...', 'ls_provider': 'openai', 'ls_model_name': 'gpt-4.1-2025-04-14', 'ls_model_type': 'chat', 'ls_temperature': 1.0})

messages (AIMessageChunk(content=[], additional_kwargs={'__openai_function_call_ids__': {}}, response_metadata={}, id='run--...', invalid_tool_calls=[{'name': None, 'args': '","', 'id': None, 'error': None, 'type': 'invalid_tool_call'}], tool_call_chunks=[{'name': None, 'args': '","', 'id': None, 'index': 2, 'type': 'tool_call_chunk'}]), {'langgraph_step': 1, 'langgraph_node': 'agent', 'langgraph_triggers': ('branch:to:agent',), 'langgraph_path': ('__pregel_pull', 'agent'), 'langgraph_checkpoint_ns': 'agent:...', 'checkpoint_ns': 'agent:..., 'ls_provider': 'openai', 'ls_model_name': 'gpt-4.1-2025-04-14', 'ls_model_type': 'chat', 'ls_temperature': 1.0})
```

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #5583, opened 2023-06-01
Batched length functions for text splitters

bhperry:
# Batched length functions for text splitters

Allows text splitters to take advantage of batched length functions instead of only calculating length on a single split at a time. This is particularly noticeable when using tokenized length functions, and even more so with huggingface "fast" tokenizers which have major performance improvements with batched input.

Additionally, by keeping track of the lengths of splits in the RecursiveCharacterTextSplitter and passing them into merge function it reduces the number of times that the same text needs to be re-tokenized.

Experimentally I've found that this speeds up text splitting around 5x when using a tokenized length function (of course depends heavily on the text being split, number of splits required, etc.)

@hwchase17

PR #32582, opened 2025-08-17
feat(text-splitters): add support for batched length function in `TextSplitter`

MartinRabaud:
PS: This PR is based on the work of @bhperry. Around 90% of the code comes from https://github.com/langchain-ai/langchain/pull/5583.
That PR was closed due to conflicts, and I found it easier to restart from an up to date branch.
All I did was make a few changes so it could pass the formatter and linter, updated the doc and integration tests.

If there’s a way to give him nearly full credit for this work within this PR, please let me know.

**Description**

Enable batching optimization when using token-based splitters.

Currently, each text split is tokenized individually, which makes the splitter very slow on large documents, especially compared to how fast HuggingFace tokenizers can be with batching.

**Tests**

Added a unit test (`test_batched_recursive_character_text_splitter`) verifying that `batched_length_functions` introduces no regressions and outperforms `simple_length_function`.

Changed the integration test `test_huggingface_tokenizer` so it uses a base tokenizer, verifying it still works as intended.

Added the integration test `test_huggingface_tokenizer_fast` to verify that the use of the fast tokenizer and batching works as intended.

**Dependencies**

No new dependencies added or changed.

### DIFF
--- a/libs/text-splitters/langchain_text_splitters/base.py
+++ b/libs/text-splitters/langchain_text_splitters/base.py
@@ -31,6 +31,7 @@ except ImportError:
 
 try:
     from transformers.tokenization_utils_base import PreTrainedTokenizerBase
+    from transformers.tokenization_utils_fast import PreTrainedTokenizerFast
 
     _HAS_TRANSFORMERS = True
 except ImportError:
@@ -50,6 +51,8 @@ class TextSplitter(BaseDocumentTransformer, ABC):
         chunk_overlap: int = 200,
         length_function: Callable[[str], int] = len,
         keep_separator: bool | Literal["start", "end"] = False,  # noqa: FBT001,FBT002
+        batched_length_function: Literal[False]  # noqa: FBT002
+        | Callable[[list[str]], list[int]] = False,
         add_start_index: bool = False,  # noqa: FBT001,FBT002
         strip_whitespace: bool = True,  # noqa: FBT001,FBT002
     ) -> None:
@@ -59,6 +62,7 @@ class TextSplitter(BaseDocumentTransformer, ABC):
             chunk_size: Maximum size of chunks to return
             chunk_overlap: Overlap in characters between chunks
             length_function: Function that measures the length of given chunks
+            batched_length_function: Function that measures chunk lengths in batches
             keep_separator: Whether to keep the separator and where to place it
                 in each corresponding chunk `(True='start')`
             add_start_index: If `True`, includes chunk's start index in metadata
@@ -85,6 +89,7 @@ class TextSplitter(BaseDocumentTransformer, ABC):
         self._chunk_size = chunk_size
         self._chunk_overlap = chunk_overlap
         self._length_function = length_function
+        self._batched_length_function = batched_length_function
         self._keep_separator = keep_separator
         self._add_start_index = add_start_index
         self._strip_whitespace = strip_whitespace
@@ -149,16 +154,32 @@ class TextSplitter(BaseDocumentTransformer, ABC):
             text = text.strip()
         return text or None
 
-    def _merge_splits(self, splits: Iterable[str], separator: str) -> list[str]:
+    def _iter_text_lengths(self, texts: Iterable[str]) -> Iterable[int]:
+        if self._batched_length_function is not False:
+            if not isinstance(texts, list):
+                texts = list(texts)
+            lengths = self._batched_length_function(texts)
+            yield from lengths
+        else:
+            for text in texts:
+                yield self._length_function(text)
+
+    def _merge_splits(
+        self,
+        splits: Iterable[str],
+        separator: str,
+        lengths: Literal[False] | Iterable[int] = False,  # noqa: FBT002
+    ) -> list[str]:
         # We now want to combine these smaller pieces into medium size
         # chunks to send to the LLM.
-        separator_len = self._length_function(separator)
+        separator_len = self._text_length(separator)
 
         docs = []
         current_doc: list[str] = []
+        current_doc_lengths: list[int] = []
         total = 0
-        for d in splits:
-            len_ = self._length_function(d)
+        split_lengths = lengths or self._iter_text_lengths(splits)
+        for d, len_ in zip(splits, split_lengths, strict=True):
             if (
                 total + len_ + (separator_len if len(current_doc) > 0 else 0)
                 > self._chunk_size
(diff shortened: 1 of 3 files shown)


MartinRabaud:
One test is failing, but I'm not sure what it is exactly. Should I do something?

MartinRabaud:
Also, I compared the speed for both batched and non batch version on large text, the difference is very much noticeable:

```
Text length: 1000059
Token indices sequence length is longer than the specified maximum sequence length for this model (318242 > 131072). Running this sequence through the model will result in indexing errors
Token length: 318242
Average time taken to split text without batched: 4.0355388879776
Average time taken to split text with batched: 2.8372336626052856
```

The test code for reference:

```py
from transformers import AutoTokenizer
import time
from langchain_text_splitters.character import RecursiveCharacterTextSplitter

text_path_file = "./wikipedia.txt"
text = ""
with open(text_path_file, "r") as file:
    for line in file:
        text += line
        if len(text) > 1000000:
            break
print("Text length:", len(text))


model_name = "Qwen/Qwen3-0.6B"
tokenizer = AutoTokenizer.from_pretrained(model_name)
tokenizer(text)
print("Token length:", len(tokenizer(text)["input_ids"]))

times = []
splitter = RecursiveCharacterTextSplitter.from_huggingface_tokenizer(
    tokenizer, chunk_size=500, separators=["\n\n", "\n", " "], batched=False)
for i in range(20):
    time_start = time.time()
    chunks = splitter.split_text(text)
    time_end = time.time()
    times.append(time_end - time_start)
print("Average time taken to split text without batched:", sum(times) / len(times))

times = []
splitter = RecursiveCharacterTextSplitter.from_huggingface_tokenizer(
    tokenizer, chunk_size=500, separators=["\n\n", "\n", " "], batched=True)
for i in range(20):
    time_start = time.time()
    chunks = splitter.split_text(text)
    time_end = time.time()
    times.append(time_end - time_start)

print("Average time taken to split text with batched:", sum(times) / len(times))

```

mdrxy:
@MartinRabaud would you mind rebasing against master to resolve conflicts?

MartinRabaud:
Hi @mdrxy, I did the rebase, resolved the conflicts and complied with the new linter rules.

Notes:

- I replaced `Optional` with `Literal[False]`, as it seems to be the standard on the project now (at least on text-splitters)
- I had to add `# type: ignore[no-any-return]` for the return of the tokenizer function. It seems that Transformers package didn't type this return, but it is indicated [in their docs](https://huggingface.co/docs/transformers/main_classes/tokenizer) that they do when you add `return_length=True`

As for the performance, I still see a net improvement:

```
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
Text length: 1000091
Token indices sequence length is longer than the specified maximum sequence length for this model (294508 > 131072). Running this sequence through the model will result in indexing errors
Token length: 294508
Average time taken to split text without batched: 5.861177968978882
Average time taken to split text with batched: 3.8749688267707825
```
See the test code [above](https://github.com/langchain-ai/langchain/pull/32582#issuecomment-3207572313)

libs/text-splitters/langchain_text_splitters/base.py
```diff
@@ -50,6 +51,8 @@ def __init__(
         chunk_overlap: int = 200,
         length_function: Callable[[str], int] = len,
         keep_separator: bool | Literal["start", "end"] = False,  # noqa: FBT001,FBT002
+        batched_length_function: Literal[False]  # noqa: FBT002
+        | Callable[[list[str]], list[int]] = False,
```
mdrxy: what if instead it was `Callable[[list[str]], list[int]] | None = None`?

libs/text-splitters/langchain_text_splitters/base.py
```diff
@@ -149,16 +154,32 @@ def _join_docs(self, docs: list[str], separator: str) -> str | None:
             text = text.strip()
         return text or None
 
-    def _merge_splits(self, splits: Iterable[str], separator: str) -> list[str]:
+    def _iter_text_lengths(self, texts: Iterable[str]) -> Iterable[int]:
+        if self._batched_length_function is not False:
```
mdrxy: following the above comment, this would become:
```suggestion
        if self._batched_length_function:
```

libs/text-splitters/langchain_text_splitters/base.py
```diff
@@ -149,16 +154,32 @@ def _join_docs(self, docs: list[str], separator: str) -> str | None:
...
+            yield from lengths
+        else:
+            for text in texts:
+                yield self._length_function(text)
+
+    def _merge_splits(
+        self,
+        splits: Iterable[str],
+        separator: str,
+        lengths: Literal[False] | Iterable[int] = False,  # noqa: FBT002
```
mdrxy: similar here

libs/text-splitters/tests/integration_tests/test_text_splitter.py
```diff
@@ -23,14 +24,24 @@ def test_huggingface_type_check() -> None:
 
 def test_huggingface_tokenizer() -> None:
     """Test text splitter that uses a HuggingFace tokenizer."""
-    tokenizer = GPT2TokenizerFast.from_pretrained("gpt2")
+    tokenizer = GPT2Tokenizer.from_pretrained("gpt2")
```
mdrxy, line 27: what's motivating this change?

libs/text-splitters/langchain_text_splitters/base.py
MartinRabaud: Yeah, that's far better. I guess I got confused by the optional being rejected by the linter and the line above using `Literal[False]` as the default value.

libs/text-splitters/tests/integration_tests/test_text_splitter.py
MartinRabaud: Since those lines:

```py
        if batched is False and isinstance(tokenizer, PreTrainedTokenizerFast):
            # Take advantage of huggingface's fast tokenizers
            # by batching tokenization of splits
            batched = True
```

change the behavior of the splitter when PreTrainedTokenizerFast is used, I wanted to make sure this test still uses the current behavior while adding a new test for the PreTrainedTokenizerFast for the new behavior to ensure no regressions both ways.

libs/text-splitters/langchain_text_splitters/base.py
MartinRabaud: Fixed

libs/text-splitters/langchain_text_splitters/base.py
MartinRabaud: Fixed

mdrxy:
Thanks for the contribution and for rebasing/responding to review feedback. I'm going to close this for now since the PR has gone stale and still needs some design/correctness follow-up before it would be mergeable. If you're still interested in pursuing this optimization, please open an issue describing the use case and proposed API/behavior so we can discuss the right shape before another implementation PR.

PR #32597, opened 2025-08-18
refactor(core): prefixes

mdrxy:
re: #32589 cc: @ccurme
- Rename namespace: `messages.content_blocks` -> `messages.content`
- Prefixes and ID logic are now in `messages.common` instead of `AIMessage` since the logic is shared between messages and message content. Did this instead of `utils` due to circular import problems that were hairy


libs/core/langchain_core/messages/common.py
```diff
@@ -0,0 +1,31 @@
+"""Common constants and utilities for LangChain messages."""
```
ccurme: Why not put this in messages/base.py?

PR #32602, opened 2025-08-19
feat(core): implement dynamic translator registration for model providers

mdrxy:
Extensible registry system for translating AI message content blocks from various model providers. Refactors the way provider-specific content is handled, moving from hardcoded logic to a plugin-like architecture.

### DIFF
--- a/libs/core/tests/unit_tests/messages/block_translators/test_openai.py
+++ b/libs/core/tests/unit_tests/messages/block_translators/test_openai.py
@@ -3,9 +3,6 @@ from typing import Optional
 from langchain_core.messages import AIMessage, AIMessageChunk
 from langchain_core.messages import content as types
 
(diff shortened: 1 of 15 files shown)


libs/core/tests/unit_tests/messages/block_translators/test_openai.py
```diff
@@ -3,6 +3,9 @@
 from langchain_core.messages import AIMessage, AIMessageChunk
 from langchain_core.messages import content as types
 
+# Import statements to trigger registration of OpenAI block translators
+from langchain_core.messages.block_translators import openai  # noqa: F401
```
ccurme: Can we register OpenAI (and other major providers) by default so we don't need to trigger?

ccurme (commented):
Looks great, thank you!

PR #32606, opened 2025-08-19
refactor: convert message content inside `BaseChatModel`

ccurme:
```python
ChatModel(model="...", output_version="v1").invoke("...")  # <-- .content is standard content blocks

ChatModel(model="...").invoke("...")  # <-- .content is unchanged / v0 (e.g., a string)

os.environ["LC_OUTPUT_VERSION"] = "v1"
ChatModel(model="...").invoke("...")  # <-- .content is standard content blocks
```


ccurme:
Codspeed regression comes from merging in changes from wip-v1.0— regenerated the cassette for streaming because the model used there is now deprecated. Streaming ended up being longer (e.g., in number of chunks) which gets interpreted as a regression.

PR #32611, opened 2025-08-19
chore(text-splitters): add mypy pydantic plugin

cbornet:


libs/text-splitters/pyproject.toml
```diff
@@ -52,8 +52,8 @@ test_integration = [
 langchain-core = { path = "../core", editable = true }
 
 [tool.mypy]
+plugins = ["pydantic.mypy"]
 strict = true
-strict_bytes = true
```
cbornet, line 56: `strict_bytes` is part of `strict`

mdrxy (approved):

PR #32614, opened 2025-08-20
fix(openai): Chat Message `Annotations` defaults to `[ ]` if not list or None

shahrukh-shaik:
**Description:** if annotation is not present or None, it shall default to empty list [ ]. Similar to line https://github.com/shahrukh-shaik/langchain/blob/09e70a346013f08304af608e6f4a3d9ed60d9aa4/libs/partners/openai/langchain_openai/chat_models/base.py#L3703

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -3875,8 +3875,8 @@ def _construct_lc_result_from_responses_api(
                         "annotations": [
                             annotation.model_dump()
                             for annotation in content.annotations
-                        ] 
-                        if isinstance(content.annotations, list) 
+                        ]
+                        if isinstance(content.annotations, list)
                         else [],
                         "id": output.id,
                     }
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -3875,7 +3875,7 @@ def _construct_lc_result_from_responses_api(
                         "annotations": [
                             annotation.model_dump()
                             for annotation in content.annotations
-                        ],
+                        ] if isinstance(content.annotations, list) else [] ,
                         "id": output.id,
                     }
                     content_blocks.append(block)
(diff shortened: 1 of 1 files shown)


shahrukh-shaik:
@ccurme  - Can you please review this pull request. This is the case to handle when annotations can be None or not in expected format. Please do suggest changes if any. 

Due to this error, we are seeing issue related to endpoints which does not set any annotations.

mdrxy:
This looks similar to #32557, am I missing anything?

shahrukh-shaik:
> This looks similar to #32557, am I missing anything?

yes, there are cases where annotations is literal `None`. The above PR seems to be adding when msg content is str and this one adds when the msg content is of type list.

@mdrxy

keyuchen21:
Hi @shahrukh-shaik! I noticed there's a formatting issue causing the CI to fail. The problem is trailing whitespace in `libs/partners/openai/langchain_openai/chat_models/base.py` on lines 3878-3879.

Here's the fix needed:

```diff
                        "annotations": [
                            annotation.model_dump()
                            for annotation in content.annotations
-                        ] 
-                        if isinstance(content.annotations, list) 
+                        ]
+                        if isinstance(content.annotations, list)
                        else [],
```

You need to remove the trailing spaces after `]` on line 3878 and after `list)` on line 3879. This should fix the ruff formatting error and make the CI pass.

shahrukh-shaik:
Thank you. I have removed the trailing whitespaces. All checks passed now.

ccurme (approved):

ccurme:
@shahrukh-shaik thanks for this.

Can you share what context you're running into this? Can I reproduce it with the OpenAI API?

shahrukh-shaik:
> @shahrukh-shaik thanks for this.
> 
> Can you share what context you're running into this? Can I reproduce it with the OpenAI API?

This is happening when an LLM agent is invoked from OpenAPI compatible endpoint created using MLFlow3 Responses Agent.

Ref: https://mlflow.org/docs/latest/genai/serving/responses-agent/

PR #32623, opened 2025-08-20
feat(anthropic): v1 support

ccurme:


libs/core/langchain_core/utils/_merge.py
```diff
@@ -116,11 +116,35 @@ def merge_lists(left: Optional[list], *others: Optional[list]) -> Optional[list]
                     if to_merge:
                         # TODO: Remove this once merge_dict is updated with special
                         # handling for 'type'.
-                        new_e = (
-                            {k: v for k, v in e.items() if k != "type"}
-                            if "type" in e
-                            else e
-                        )
+                        if (left_type := merged[to_merge[0]].get("type")) and (
```
ccurme, line 119: This logic is due to how server-side tool calls are streamed. They use a generic `input_json_delta` type— same as vanilla tool calls. So there's a challenge to distinguish the correct type in the case that we want to lazy-load them from an isolated AIMessageChunk.

What I've done for now is allow them to be non-standard blocks, but update the merging logic so that if we sum `standard + non_standard`, we unpack the `value` from the non-standard block and adopt the type of the standard block.

Example stream for reference:
```
[{'text': 'Searching', 'type': 'text', 'index': 0}]
[{'text': ' the web', 'type': 'text', 'index': 0}]
[{'text': '...', 'type': 'text', 'index': 0}]
[{'id': 'srvtoolu_016FULxR9WgfTakbU7YaD6Ay', 'input': {}, 'name': 'web_search', 'type': 'server_tool_use', 'index': 1}]
[{'partial_json': '', 'type': 'input_json_delta', 'index': 1}]
[{'partial_json': '{"qu', 'type': 'input_json_delta', 'index': 1}]
[{'partial_json': 'er', 'type': 'input_json_delta', 'index': 1}]
[{'partial_json': 'y": "query', 'type': 'input_json_delta', 'index': 1}]
[{'partial_json': 'tokens ', 'type': 'input_json_delta', 'index': 1}]
[{'partial_json': 'streamed ', 'type': 'input_json_delta', 'index': 1}]
[{'partial_json': 'here"}', 'type': 'input_json_delta', 'index': 1}]
```

ccurme: Open to alternatives here— one option is to change the type from `input_json_delta` to something like `web_search_query_delta` or other meaningful type.

This seems better but is technically breaking to anyone expecting `input_json_delta` in this case.

libs/core/langchain_core/messages/block_translators/langchain_v0.py
```diff
@@ -21,26 +21,20 @@ def _convert_v0_multimodal_input_to_v1(
...
-            # Check if this looks like v0 format
-            if (
-                value.get("type") in {"image", "audio", "file"}
-                and "source_type" in value
-            ):
-                converted_block = _convert_legacy_v0_content_block_to_v1(value)
-                converted_blocks.append(cast("types.ContentBlock", converted_block))
-            else:
-                converted_blocks.append(block)
+    unpacked_blocks: list[dict[str, Any]] = [
```
ccurme, line 24: this was ignoring openai blocks with `type` name collisions (file and audio)

PR #32624, opened 2025-08-20
fix(openai): allow temperature parameter for gpt-5 chat models

TheCrazyLex:
**Description:**
- GPT-5 chat models like "gpt-5-chat-latest" support the temperature parameter
- For those models, we should not force-remove the parameter


TheCrazyLex:
@ccurme Can you check? :)

ccurme (approved):

ccurme:
Thank you!

Issue #32489, opened 2025-08-10
AzureChatOpenAI creates a new `httpx` client on each instantiation


sbhrwlr:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

Code Example

``` python
import asyncio
import httpx
from langchain_openai import AzureChatOpenAI

# Track httpx client creations
original_init = httpx.AsyncClient.__init__
counter = {"count": 0}

def counting_init(self, *args, **kwargs):
    counter["count"] += 1
    print(f"New AsyncClient created: {counter['count']}")
    original_init(self, *args, **kwargs)

httpx.AsyncClient.__init__ = counting_init

def create_llm():
    return AzureChatOpenAI(
        azure_endpoint="https://<REDACTED-ENDPOINT>/",
        deployment_name="<REDACTED-DEPLOYMENT-NAME>",
        temperature=0.7
    )

async def main():
    llms = [create_llm() for _ in range(5)]
    print(f"Total AsyncClient instances: {counter['count']}")

asyncio.run(main())
```

### Description

langchain-core==0.3.65
langchain-openai==0.3.23

### System Info

backports.tarfile==1.2.0
certifi==2025.8.3
charset-normalizer==3.4.2
docutils==0.22
id==1.5.0
idna==3.10
importlib_metadata==8.7.0
jaraco.classes==3.4.0
jaraco.context==6.0.1
jaraco.functools==4.2.1
keyring==25.6.0
markdown-it-py==3.0.0
mdurl==0.1.2
more-itertools==10.7.0
nh3==0.3.0
packaging==25.0
Pygments==2.19.2
readme_renderer==44.0
requests==2.32.4
requests-toolbelt==1.0.0
rfc3986==2.0.0
rich==14.1.0
twine==6.1.0
urllib3==2.5.0
zipp==3.23.0
<details>
<summary>Agent Context</summary>

[...]

PR #32630, opened 2025-08-20
fix(openai): reuse HTTP clients in `AzureChatOpenAI`

keyuchen21:
Similar to #32531

## Summary

Fixes #32489

AzureChatOpenAI was creating new `httpx.AsyncClient/Client` instances on each instantiation instead of reusing cached clients like the base ChatOpenAI class does. This PR implements client caching to improve performance and reduce memory overhead.

## Problem

As reported in the issue, each `AzureChatOpenAI()` instantiation creates a new HTTP client even when configurations are identical, leading to:
- Unnecessary memory overhead
- Potential resource exhaustion with many instances
- Inconsistent behavior compared to base ChatOpenAI

## Solution

This PR modifies the Azure client initialization to use cached HTTP clients when not explicitly provided:
- Uses `_get_default_httpx_client()` for sync clients
- Uses `_get_default_async_httpx_client()` for async clients
- Only creates new clients when explicitly provided via `http_client` or `http_async_client` parameters

## Changes

1. **azure.py** - Updated client initialization logic:
   - Import cached client helper functions from `_client_utils`
   - Check if `http_client`/`http_async_client` are None before using cached versions
   - Maintain backward compatibility for explicitly provided clients

2. **test_azure.py** - Added unit test:
   - Verifies multiple AzureChatOpenAI instances share the same HTTP client
   - Tests both sync and async client caching

## Testing

- Added unit test `test_http_client_reuse()` to verify client caching works correctly
- Existing tests continue to pass, ensuring backward compatibility
- Manual testing shows reduced memory usage with multiple instances

## Impact

This is a performance improvement with no breaking changes. Users who explicitly provide HTTP clients will see no change in behavior.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/azure.py
+++ b/libs/partners/openai/langchain_openai/chat_models/azure.py
@@ -676,7 +676,7 @@ class AzureChatOpenAI(BaseChatOpenAI):
                 }
             else:
                 sync_specific = {"http_client": self.http_client}
-            self.root_client = openai.AzureOpenAI(**client_params, **sync_specific)  # type: ignore[arg-type]
+            self.root_client = openai.AzureOpenAI(**client_params, **sync_specific)  # type: ignore[call-overload]
             self.client = self.root_client.chat.completions
         if not self.async_client:
             # Use cached async client if not explicitly provided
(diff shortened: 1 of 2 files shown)


ccurme (commented):
I don't think we can import `_get_default_httpx_client` as-is, unfortunately.

That function hard-codes some logic from the OpenAI sdk: https://github.com/langchain-ai/langchain/blob/e0aaaccb616f02a6022feb1a9c4bc50fd27a9d9a/libs/partners/openai/langchain_openai/chat_models/_client_utils.py#L48-L50

this logic is invalid for the Azure SDK. We need to look into the internals of the Azure SDK and make sure there are zero changes for all existing langchain usage patterns that pass through the Azure SDK.

ccurme:
Closing this as I believe it is no longer needed following Azure OpenAI's v1 API, which can be accessed via ChatOpenAI: https://docs.langchain.com/oss/python/integrations/chat/openai#using-with-azure-openai

PR #32631, opened 2025-08-21
fix(openai): handle None annotations in chat messages

keyuchen21:
## Summary
This PR ensures that annotations in OpenAI chat messages default to an empty list when they are not present or are None.

## Problem
Some endpoints don't set annotations, which can cause issues when processing chat messages.

## Solution
Added a check to verify if `content.annotations` is a list before trying to iterate over it. If it's not a list (e.g., None), it defaults to an empty list.

## Related
- Addresses the same issue as #32614 but with proper formatting that passes linting checks
- Closes #32614 (supersedes it with a clean implementation)

## Testing
- Formatting verified with ruff
- All linting checks pass


ccurme:
PR is empty following changes from earlier PRs. Thank you for the fix!

Issue #32511, opened 2025-08-12
[bug]`dereference_refs` seems to lead to infinite recursion


filwaline:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```python
import os

from langchain_mcp_adapters.client import MultiServerMCPClient
from langchain_xai import ChatXAI


async def get_mcp_tools():
    client = MultiServerMCPClient(
        {
            "hasura": {
                "url": "http://127.0.0.1:5000/mcp",
                "transport": "streamable_http",
            }
        }
    )
    tools = await client.get_tools()
    print(f"Tools: {[tool.name for tool in tools]}")
    return tools


def create_agent():
    model = ChatXAI(
        xai_api_key=os.getenv("xai_api_key"), model="grok-4-latest", temperature=0.2
    )
    print("llm initialized")
    return model


async def main():
    agent = create_agent()
    print(f"Agent: {agent.name}")
    tools = await get_mcp_tools()
    agent = agent.bind_tools(tools)
    print(f"Agent with tools: {agent.name}")

    response = await agent.ainvoke(
        ("user", "What is the health status of the pigs?"),
    )

    print(response)


if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

Apollo-mcp-server run as an external service, can not provide example code.

### Error Message and Stack Trace (if applicable)

``` 
 File "/Users/rusty/Documents/susteynable/chatfarm/.venv/lib/python3.12/site-packages/langchain_core/utils/json_schema.py", line 58, in _dereference_refs_helper
    obj_out[k] = _dereference_refs_helper(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rusty/Documents/susteynable/chatfarm/.venv/lib/python3.12/site-packages/langchain_core/utils/json_schema.py", line 52, in _dereference_refs_helper
    full_ref = _dereference_refs_helper(
[...]

PR #32643, opened 2025-08-21
test(core): add regression tests for `dereference_refs` cycles and all…

Mdzub7:
Context
- Adds two regression tests for JSON Schema dereferencing:
  1) Three-node mixed-$ref cycle (A→B→C→A) ensuring termination and sibling metadata retention.
  2) allOf + $ref composition to ensure dereferencing preserves sibling parts.

Why
- #32511 reports infinite recursion with mixed $ref objects.
- #32578 introduces the fix; these tests broaden coverage (multi-node cycle + allOf).

Notes
- Unit-level tests; fast; no external services.
- Verified locally: both tests pass on #32578.

Refs
- Issue: #32511
- Fix PR: #32578


eyurtsev:
Added more unit tests in the PR: https://github.com/langchain-ai/langchain/pull/32578

PR #32649, opened 2025-08-22
fix(openai): Fix Azure OpenAI Responses API model field issue

JunHyungKang:
**Description:** Fixes Azure OpenAI Responses API calls failing due to incorrect `model` field value. 
The Responses API requires Azure deployment name in the `model` field, but the current implementation passes OpenAI model name (e.g., "gpt-4o"), causing API failures. 
This fix adds `_get_request_payload` override in `AzureChatOpenAI` to use `deployment_name` for Responses API while maintaining existing behavior for Chat Completions API.

**Issue:** N/A (discovered during development and testing)

**Dependencies:** None

**Twitter handle:** N/A

**Add tests and docs:** N/A (this is a bug fix for existing Azure OpenAI integration, not a new integration)

**Lint and test:** ✅ Ran `make format`, `make lint` and `make test` - all passing


libs/partners/openai/langchain_openai/chat_models/azure.py
```diff
@@ -754,6 +754,22 @@ def _create_chat_result(
...
+        *,
+        stop: Optional[list[str]] = None,
+        **kwargs: Any,
+    ) -> dict:
+        """Get the request payload, using deployment name for Azure Responses API."""
+        payload = super()._get_request_payload(input_, stop=stop, **kwargs)
+
+        # For Azure Responses API, use deployment name instead of model name
+        if self._use_responses_api(payload) and self.deployment_name:
+            payload["model"] = self.deployment_name
```
ccurme, line 773: Do you know if this is always safe to do? Any chance Azure is using the model server-side separate from the deployment name?

This appears to work if I use a deployment with a custom name (the model that ran is returned in the response metadata). Any docs or other information that would confirm that we can essentially drop `model` completely from AzureChatOpenAI usage would be helpful.

ccurme (commented):
Thanks for this.

libs/partners/openai/langchain_openai/chat_models/azure.py
JunHyungKang: @ccurme 
Thanks for the thoughtful review and for testing it out!

You raised a great point about whether this is always a safe change. I looked into the official Azure documentation, and it confirms that using the **deployment name** in the `model` parameter is the intended usage.

In the official guide for this API call (https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/responses?tabs=python-key#generate-a-text-response)), the provided code example is:

```python
response = client.responses.create(   
  model="gpt-4.1-nano", # Replace with your model deployment name 
  input="This is a test.",
)
```

The key piece of evidence is the comment directly from Microsoft: **`# Replace with your model deployment name`**.

Hope this clarifies the reasoning behind the change!

JunHyungKang: Just found a description in the Azure OpenAI REST API reference documentation.
(https://learn.microsoft.com/en-us/azure/ai-foundry/openai/reference-preview-latest)
image

ccurme: Thanks very much for following up.

ccurme (approved):
Thanks for this. I made an update such that if `model` is passed explicitly, we don't overwrite it. So what we effectively do here is add support for the case where deployment is passed and model isn't.

PR #32652, opened 2025-08-22
chore(infra): optimize codspeed benchmarking

mdrxy:
- Replace static package matrix with intelligent file change detection and benchmarks any package with changes
- Only run on a package if it had files change


ccurme (commented):
I think we had issues with this along the lines of "if a benchmark doesn't run it is assumed deleted"

PR #32655, opened 2025-08-22
fix(openai): Always add raw response object to OpenAI client errors for invoke

jacoblee93:


jacoblee93:
Side note - should we be using `with_streaming_response` here? https://github.com/langchain-ai/langchain/pull/32655/files#diff-197e34e7a297cef90049a95d035073870fb484896bab67f800a95fe190ab02bfR1089

https://github.com/openai/openai-python?tab=readme-ov-file#with_streaming_response

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1117,6 +1118,10 @@ def _stream(
                     yield generation_chunk
         except openai.BadRequestError as e:
             _handle_openai_bad_request(e)
+        except Exception as e:
```
ccurme: (nit) it looks like we won't enter this condition for BadRequestError

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1149,23 +1154,37 @@ def _generate(
                     "specified."
                 )
             payload.pop("stream")
+            raw_response = None
             try:
-                response = self.root_client.beta.chat.completions.parse(**payload)
+                raw_response = (
+                    self.root_client.beta.chat.completions.with_raw_response.parse(
```
ccurme: I think we no longer need `beta` here (we had a todo to update this).

the warning above might also be no longer accurate, given that we're using `with_raw_response` here. this has to do with them putting `.parse` out of beta.

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -1117,6 +1118,10 @@ def _stream(
                     yield generation_chunk
         except openai.BadRequestError as e:
             _handle_openai_bad_request(e)
+        except Exception as e:
+            if response is not None:
+                e.response = response  # type: ignore[attr-defined]
```
ccurme: does this end up tracing correctly? is there any handling we need to do in core?

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
```diff
@@ -1155,3 +1155,106 @@ def test_prompt_cache_key_usage_methods_integration() -> None:
...
+    except Exception as e:
+        assert e.response is not None  # type: ignore[attr-defined]
+    else:
+        assert False
+
+
+# VCR can't handle parameterized tests
+@pytest.mark.vcr()
+async def test_schema_parsing_failures_responses_api_async() -> None:
+    class BadModel(BaseModel):
```
ccurme: is this already defined at the top of the module

libs/partners/openai/langchain_openai/chat_models/base.py
jacoblee93: `_handle_openai_bad_request(e)` will also throw though no?

libs/partners/openai/langchain_openai/chat_models/base.py
jacoblee93: Yeah I think we need your other PR, will merge in and verify 100% though

libs/partners/openai/langchain_openai/chat_models/base.py
jacoblee93: Sure can update that here too

libs/partners/openai/tests/integration_tests/chat_models/test_base.py
jacoblee93: Ah shoot

libs/partners/openai/langchain_openai/chat_models/base.py
jacoblee93: Having said that there's a lot of them, should I remove them all? Maybe better done in another PR?

libs/partners/openai/langchain_openai/chat_models/base.py
jacoblee93: Updated to append raw `http_response` since that seems to make it work

libs/partners/openai/langchain_openai/chat_models/base.py
jacoblee93: Removed for this call and the async variant

ccurme (approved):

PR #32664, opened 2025-08-24
feat(core): parse `tool_call_chunks` in content in aggregated stream

ccurme:
Analogous to https://github.com/langchain-ai/langchain/pull/32358, which was implemented on the dataclass message types.

Features:
- When lazily parsing `.content_blocks` on a message that has been aggregated from a stream, it will include parsed tool_calls instead of tool_call_chunks.
- Similarly, when streaming standard content (`output_version="v1"`), on aggregation message content will contain a parsed tool call.

```python

    pass

llm_with_tools = llm.bind_tools([get_weather])
full = None
for chunk in llm_with_tools.stream("What's the weather in Boston?"):
    print(full.content_blocks)
```
```
[{'type': 'tool_call_chunk', 'id': 'call_MyfVd1XoUOeNJlqEE2KYqohK', 'name': 'get_weather', 'args': '', 'index': 0}]
[{'type': 'tool_call_chunk', 'id': 'call_MyfVd1XoUOeNJlqEE2KYqohK', 'name': 'get_weather', 'args': '{"', 'index': 0}]
[{'type': 'tool_call_chunk', 'id': 'call_MyfVd1XoUOeNJlqEE2KYqohK', 'name': 'get_weather', 'args': '{"location', 'index': 0}]
[{'type': 'tool_call_chunk', 'id': 'call_MyfVd1XoUOeNJlqEE2KYqohK', 'name': 'get_weather', 'args': '{"location":"', 'index': 0}]
[{'type': 'tool_call_chunk', 'id': 'call_MyfVd1XoUOeNJlqEE2KYqohK', 'name': 'get_weather', 'args': '{"location":"Boston', 'index': 0}]
[{'type': 'tool_call_chunk', 'id': 'call_MyfVd1XoUOeNJlqEE2KYqohK', 'name': 'get_weather', 'args': '{"location":"Boston"}', 'index': 0}]
[{'name': 'get_weather', 'args': {'location': 'Boston'}, 'id': 'call_MyfVd1XoUOeNJlqEE2KYqohK', 'type': 'tool_call'}]
^ Previously this last chunk was still tool_call_chunk
```

To do this, we add an optional attribute `chunk_position` to AIMessageChunk, which just takes value `"last"` at the last chunk. This triggers changing tool_call_chunks -> tool_calls.

Alternatives considered:
- Capture both the first and the last chunk via an attribute `chunk_span`, taking values `("first",)` on the first chunk, `("last",)` on the last chunk, and aggregating to `("first", "last")`. Require `("first", "last")` instead of just `"last"`. This is more robust to the case where tool_call_chunks are included in the final chunk— having `chunk_position="last"` will erroneously put tool calls in content in that isolated chunk. But afaik this never happens, and is avoided by using the empty chunk in langchain-core, so didn't feel the added complexity was warranted.
- Detect when the tool_call_chunk is finished by attempting to parse the json: tool_call_chunk args contain valid json substrings— e.g., OpenAI usually streams out `'":"'`, which parses to valid json.
- Detect when the tool_call_chunk is finished by comparing with tool_calls of the same ID:
  - if name is populated and json-parsed args match tool call, assume finished
  - note that json substrings can contain valid dict json substrings (e.g., `{'a': '` | `{'b': 3}` | `}`) but checking for populated name might be sufficient as an extra check. This solution is less disruptive but hacky.

N.B. Integrations can avoid any extra chunks by setting `chunk_position="last"` on the final chunk that contains usage / other metadata. If they don't, BaseChatModel will yield a final empty chunk with this value.


libs/core/langchain_core/language_models/fake_chat_models.py
```diff
@@ -112,7 +112,12 @@ def _stream(
             ):
                 raise FakeListChatModelError
 
-            yield ChatGenerationChunk(message=AIMessageChunk(content=c))
+            chunk_position: Optional[Literal["last"]] = (
+                "last" if i_c == len(response) - 1 else None
+            )
+            yield ChatGenerationChunk(
+                message=AIMessageChunk(content=c, chunk_position=chunk_position)
+            )
```
ccurme, line 120: chat models have the option of including `chunk_position="last"` themselves, so that there is no extra chunk produced in BaseChatModel. this change is otherwise not necessary.

libs/core/langchain_core/messages/block_translators/anthropic.py
```diff
@@ -226,29 +226,37 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
-                    not isinstance(message, AIMessageChunk)
-                    and len(message.tool_calls) == 1
-                ):
-                    tool_call_block = message.tool_calls[0]
+                else:
+                    tool_call_block: Optional[types.ToolCall] = None
+                    # Non-streaming or gathered chunk
+                    if len(message.tool_calls) == 1:
+                        tool_call_block = message.tool_calls[0]
+                    elif call_id := block.get("id"):
```
ccurme, line 248: unrelated improvement here

PR #32668, opened 2025-08-25
feat(core): (v1) restore separate type for `AIMessage.tool_calls`

ccurme:
`"type"` needs to be a required key for content blocks, including tool calls in content blocks, for type discrimination.

This has not a required key for the tool calls in `AIMessage.tool_calls`. If we require that content ToolCall and `.tool_calls` ToolCall both require `"type"`, we introduce type errors in user code:
```python
msg.tool_calls = [{"name": "foo", "args": {"a": "b"}, "id": "abc-123"}]
```
Any pydantic classes expecting tool calls would also raise validation errors if `"type"` is not provided— I realized this issue when this happened to some test classes in langgraph.

For now, we restore the old type, so that for both AIMessage.tool_calls and AIMessage.tool_call_chunks we have separate types that do not require the `"type"` key.

### DIFF
--- a/libs/core/langchain_core/messages/ai.py
+++ b/libs/core/langchain_core/messages/ai.py
@@ -255,9 +255,9 @@ class AIMessage(BaseMessage):
                         "args": tool_call["args"],
                     }
                     if "index" in tool_call:
-                        tool_call_block["index"] = tool_call["index"]
+                        tool_call_block["index"] = tool_call["index"]  # type: ignore[typeddict-item]
                     if "extras" in tool_call:
-                        tool_call_block["extras"] = tool_call["extras"]
+                        tool_call_block["extras"] = tool_call["extras"]  # type: ignore[typeddict-item]
                     blocks.append(tool_call_block)
 
         return blocks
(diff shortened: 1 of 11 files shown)


libs/core/langchain_core/messages/ai.py
```diff
@@ -255,9 +255,9 @@ def content_blocks(self) -> list[types.ContentBlock]:
                         "args": tool_call["args"],
                     }
                     if "index" in tool_call:
-                        tool_call_block["index"] = tool_call["index"]
+                        tool_call_block["index"] = tool_call["index"]  # type: ignore[typeddict-item]
```
mdrxy, line 258: ewwwwwwwwwwwww

libs/core/tests/unit_tests/messages/test_ai.py
```diff
@@ -203,6 +203,16 @@ def test_add_ai_message_chunks_usage() -> None:
...
 
 
+def test_init_tool_calls() -> None:
+    # Test we add "type" key on init
+    msg = AIMessage("", tool_calls=[{"name": "foo", "args": {"a": "b"}, "id": "abc"}])
+    assert len(msg.tool_calls) == 1
+    assert msg.tool_calls[0]["type"] == "tool_call"
+
+    # Test we can assign without adding type key
+    msg.tool_calls = [{"name": "bar", "args": {"c": "d"}, "id": "def"}]
```
mdrxy, line 213: assignment doesn't add a type key right?

PR #32676, opened 2025-08-26
fix(perplexity): preserve citations in structured output

keyuchen21:
## Description
This PR fixes an issue where citations from Perplexity API responses were being lost when using the `.with_structured_output()` method.

Fixes #32382

## Problem
When using structured output with ChatPerplexity, the citations that are normally available in `additional_kwargs` were not being preserved in the parsed structured output.

## Solution
Added a helper function `_add_citations_to_parsed_output()` that:
- Captures citations from the raw LLM response
- Adds them to the parsed structured output when applicable
- Works with both dict outputs and Pydantic models
- Handles both Pydantic v1 and v2 field detection

## Testing
Added comprehensive test coverage with three test cases:
- ✅ Test for Pydantic models with citations field
- ✅ Test for dict output preserving citations  
- ✅ Test for models without citations field (ensures no errors)

All tests pass, linting passes, and type checking passes.

## Example Usage
```python
from langchain_perplexity import ChatPerplexity

class OutputWithCitations(BaseModel):
    result: str
    citations: list[str] = Field(default_factory=list)

llm = ChatPerplexity(model="sonar-pro")
structured_llm = llm.with_structured_output(OutputWithCitations)
response = structured_llm.invoke("What is the capital of France?")
# response.citations now contains the citation URLs
```

### DIFF
--- a/libs/partners/perplexity/langchain_perplexity/chat_models.py
+++ b/libs/partners/perplexity/langchain_perplexity/chat_models.py
@@ -49,6 +49,38 @@ def _is_pydantic_class(obj: Any) -> bool:
     return isinstance(obj, type) and is_basemodel_subclass(obj)
 
 
+def _add_citations_to_parsed_output(parsed: Any, result: Any) -> Any:
+    """Add citations to parsed output if available.
+
+    Args:
+        parsed: The parsed output from the parser.
+        result: The raw result from the LLM.
+
+    Returns:
+        The parsed output with citations added if available.
+    """
+    if hasattr(result, "additional_kwargs"):
+        citations = result.additional_kwargs.get("citations")
+        if citations:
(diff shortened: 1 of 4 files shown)


keyuchen21:
@mdrxy any update?

ccurme (commented):
Hello, thanks for submitting this. I think the cleanest solution is to set [include_raw=True](https://python.langchain.com/docs/how_to/structured_output/#advanced-raw-outputs) when calling `with_structured_output`. This gives you access to both the underlying AIMessage and the parsed representation.

I think it is preferable to adding keys to the parsed representation that are not in the underlying schema. There are other fields you might want to access when using `with_structured_output`— token usage, reasoning, etc. The recommended approach has been to use `include_raw=True` instead of adding all these keys to the parsed representation.

PR #32694, opened 2025-08-26
feat(core): `output_version` on invoke / model call

mdrxy:
Q: If `output_version` is not set, `response_metadata` won't have an `output_version`, even though it defaults to `'v0'`. Do we want this?


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -1683,11 +1904,6 @@ class AnswerWithJustification(BaseModel):
             msg = f"Received unsupported arguments {kwargs}"
             raise ValueError(msg)
 
-        from langchain_core.output_parsers.openai_tools import (
-            JsonOutputKeyToolsParser,
-            PydanticToolsParser,
-        )
```
ccurme: Did we dig into why this was here? I'm assuming it was to resolve a circular import.

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -424,9 +428,35 @@ def invoke(
...
+
+        effective_output_version = (
+            output_version if output_version is not None else self.output_version
+        )
+        kwargs["_output_version"] = effective_output_version or "v0"
+
+        # Whether the user explicitly set an output_version for either model or call
+        kwargs["_output_version_explicit"] = (
+            output_version is not None or self.output_version is not None
+        )
```
ccurme, line 458: instead of determining this here, can we just propagate
```python
kwargs["_output_version"] = output_version
```
and then at call sites, do
```python
output_version = kwargs.pop("_output_version", None) or self.output_version
message = _update_message_content_to_blocks(message, output_version)
```
then update `_update_message_content_to_blocks` (maybe rename to `process_output_version` or similar) to do something like
```python
def _update_message_content_to_blocks(message: T, output_version: str) -> T:
    if output_version is None:
        return message
    elif output_version == "v1":
        return message.model_copy(
            update={
                "content": message.content_blocks,
                "response_metadata": {
                    **message.response_metadata,
                    "output_version": output_version,
                },
            }
        )
    else:
        # v0
        return message.model_copy(
            update={
                "response_metadata": {
                    **message.response_metadata,
                    "output_version": output_version,
                },
            }
        )
```

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -562,11 +649,27 @@ def stream(
                 input_messages = _normalize_messages(messages)
                 run_id = "-".join((LC_ID_PREFIX, str(run_manager.run_id)))
                 yielded = False
-                for chunk in self._stream(input_messages, stop=stop, **kwargs):
+
+                filtered_kwargs = {
+                    k: v
+                    for k, v in kwargs.items()
+                    if k not in ("_output_version", "_output_version_explicit")
+                }
```
ccurme, line 657: (nit) see above comment on just popping `_output_version` from kwargs and processing

libs/core/langchain_core/utils/_merge.py
```diff
@@ -44,6 +44,16 @@ def merge_dicts(left: dict[str, Any], *others: dict[str, Any]) -> dict[str, Any]
...
+                if right_k == "output_version":
+                    if merged[right_k] == right_v:
+                        continue
+                    msg = (
+                        "Unable to merge. Two different values seen for "
+                        f"'output_version': {merged[right_k]} and {right_v}. "
+                        "'output_version' should have the same value across "
+                        "all chunks in a generation."
+                    )
+                    raise ValueError(msg)
```
ccurme, line 56: BaseChatModel is inserting the output version, right? how would we hit this ValueError?

If a user receives this ValueError, are they just stuck? What can they do?

I think in most other cases in the merging logic we just accept the left value.

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -244,19 +253,21 @@ def _llm_type(self) -> str:
     model = ModelWithSyncStream()
     chunks = list(model.stream("anything"))
     assert chunks == [
+        _any_id_ai_message_chunk(content="a"),
         _any_id_ai_message_chunk(
-            content="a",
+            content="b",
+            chunk_position="last",
```
ccurme, line 259: why change formatting

libs/core/tests/unit_tests/language_models/chat_models/test_output_version.py
```diff
@@ -0,0 +1,366 @@
...
+from typing_extensions import override
+
+from langchain_core.callbacks import CallbackManagerForLLMRun
+from langchain_core.language_models import BaseChatModel
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+from langchain_core.messages import AIMessage, AIMessageChunk, BaseMessage, HumanMessage
+from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
+
+
+class OutputVersionTrackingChatModel(GenericFakeChatModel):
```
ccurme, line 18: can we not use GenericFakeChatModel on its own?

libs/core/tests/unit_tests/language_models/chat_models/test_rate_limiting.py
```diff
@@ -216,7 +216,8 @@ def test_rate_limit_skips_cache() -> None:
             '[{"lc": 1, "type": "constructor", "id": ["langchain", "schema", '
             '"messages", "HumanMessage"], "kwargs": {"content": "foo", '
             '"type": "human"}}]',
-            "[('_type', 'generic-fake-chat-model'), ('stop', None)]",
+            "[('_output_version', 'v0'), ('_output_version_explicit', False), ('_type',"
+            " 'generic-fake-chat-model'), ('stop', None)]",
```
ccurme, line 221: will this break caches

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1596,8 +1596,11 @@ def _stream(
         run_manager: Optional[CallbackManagerForLLMRun] = None,
         *,
         stream_usage: Optional[bool] = None,
+        output_version: Optional[str] = None,
         **kwargs: Any,
     ) -> Iterator[ChatGenerationChunk]:
+        # Note: output_version accepted for interface consistency; format conversion
+        # handled by core
```
ccurme, line 1606: are we introducing type errors into all chat models?

could we just add to signature of top-level `.invoke`, `.stream`, etc. instead of `_generate`, `_stream` and propagate through kwargs?

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -4057,6 +4122,9 @@ def _construct_lc_result_from_responses_api(
     )
     if output_version == "v0":
         message = _convert_to_v03_ai_message(message)
+    elif output_version == "v1":
+        # Use content_blocks property which handles v1 conversion via block_translators
+        message = message.model_copy(update={"content": message.content_blocks})
```
ccurme, line 4122: isn't this done entirely in core? we went to some lengths so that we wouldn't have to touch the chat model implementations.

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -967,18 +967,26 @@ def _stream_responses(
...
+        *,
+        output_version: Optional[str] = None,
         **kwargs: Any,
     ) -> Iterator[ChatGenerationChunk]:
         kwargs["stream"] = True
+        effective_output_version = (
+            output_version
+            if output_version is not None
+            else (self.output_version or "v0")
+        )
```
ccurme, line 976: is this logic needed? ideally we should need zero or minimal changes to child chat model classes.

libs/core/langchain_core/language_models/chat_models.py
mdrxy: Possibly, or just an erroneous placement (since we don't enforce placement with ruff). Moving didn't break anything 🙏

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
mdrxy: twas auto

PR #32707, opened 2025-08-27
!fix(core): `_should_stream` respects `streaming` param

mdrxy:
> [!IMPORTANT]
> This is a breaking change

`_should_stream()` in `BaseChatModel` wasn't checking the instance-level streaming parameter, leading to streaming being enabled even when explicitly disabled

- Before: `ChatOpenAI().stream()` always attempted streaming (ignored `streaming=False`)
- After: `ChatOpenAI().stream()` respects `streaming=False` and falls back to `invoke()`

Any existing code doing:
```python
model = ChatOpenAI()  # streaming=False by default
for chunk in model.stream(messages):  # This now calls invoke() instead of streaming!
    process(chunk)
```


libs/partners/openai/tests/unit_tests/chat_models/test_base.py
```diff
@@ -350,7 +350,7 @@ def mock_glm4_completion() -> list:
 
 async def test_glm4_astream(mock_glm4_completion: list) -> None:
     llm_name = "glm-4"
-    llm = ChatOpenAI(model=llm_name, stream_usage=True)
+    llm = ChatOpenAI(model=llm_name, stream_usage=True, streaming=True)
```
ccurme, line 353: why add these?

mdrxy: The unit tests needed updating because they were relying on the old behavior where the streaming parameter was ignored. Now they need to set `streaming=True` to test streaming. This is breaking - I'll update description to reflect.

ccurme (changes requested):
What is motivating this change?

I think the underlying issue is that `streaming: bool = False` on BaseChatOpenAI has no sentinel value— ideally it should be `streaming: bool | None = None`.

I think intended usage is that
```python
llm = ChatOpenAI(model="...")
llm.stream(...)
```
streams without you having to explicitly specify `streaming=True` in the constructor. Otherwise this is (very) breaking and it's unclear what the benefit is.

ccurme:
Can re-open if needed.

PR #32745, opened 2025-08-29
feat(openai): make `tiktoken_enabled` and `embedding_ctx_length` configurable

brucezo:


ccurme (commented):
Can you elaborate on the motivation here? Preferably with a PR summary and/or a test demonstrating desired behavior. From what I can tell there is functionally no change here. Closing for now.

PR #32757, opened 2025-08-30
fix(qdrant): allow `as_retriever` to work without embeddings in SPARSE mode

starchou6:
Description:
Fixes a bug where QdrantVectorStore.as_retriever() raised a ValueError if embedding=None, even when retrieval_mode=RetrievalMode.SPARSE. This fix relaxes the check in the .embeddings property to allow sparse-only configurations.

Fixes #32751
Dependencies: None


ccurme (approved):
Thank you!

Issue #24803, opened 2024-07-30
Standardize LLM Docs


efriis:
### Privileged issue

- [X] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

## Issue

To make our LLM integrations as easy to use as possible we need to make sure the docs for them are thorough and standardized. There are two parts to this: updating the llm docstrings and updating the actual integration docs.

This needs to be done for each LLM integration, ideally with one PR per LLM.

Related to broader issues #21983 and #22005.

## Docstrings

Each LLM class docstring should have the sections shown in the [Appendix](#appendix) below. The sections should have input and output code blocks when relevant.

To build a preview of the API docs for the package you're working on run (from root of repo):

```bash
make api_docs_clean; make api_docs_quick_preview API_PKG=openai
```

where `API_PKG=` should be the parent directory that houses the edited package (e.g. community, openai, anthropic, huggingface, together, mistralai, groq, fireworks, etc.). This should be quite fast for all the partner packages.

## Doc pages

Each LLM [docs page](https://python.langchain.com/v0.2/docs/integrations/llms/) should follow [this template](https://github.com/langchain-ai/langchain/blob/master/libs/cli/langchain_cli/integration_template/docs/llms.ipynb). 

- [ ] TODO(Erick): populate a complete example

You can use the `langchain-cli` to quickly get started with a new chat model integration docs page (run from root of repo):

```bash
poetry run pip install -e libs/cli
poetry run langchain-cli integration create-doc --name "foo-bar" --name-class FooBar --component-type LLM --destination-dir ./docs/docs/integrations/llms/
```

where `--name` is the integration package name without the "langchain-" prefix and `--name-class` is the class name without the "LLM" prefix. This will create a template doc with some autopopulated fields at docs/docs/integrations/llms/foo_bar.ipynb.

To build a preview of the docs you can run (from root):

```bash
make docs_clean
make docs_build
cd docs/build/output-new
yarn
yarn start
```

## Appendix

Expected sections for the LLM class docstring.

```python
    """__ModuleName__ completion model integration.

    # TODO: Replace with relevant packages, env vars.
    Setup:
        Install ``__package_name__`` and set environment variable ``__MODULE_NAME___API_KEY``.

        .. code-block:: bash

            pip install -U __package_name__
            export __MODULE_NAME___API_KEY="your-api-key"

    # TODO: Populate with relevant params.
    Key init args — completion params:
        model: str
[...]

PR #32758, opened 2025-08-30
docs: standardize `OllamaLLM` and `BaseOpenAI` docstrings

raviraj-441:
- Add comprehensive docstring following LangChain standards
- Include Setup, Key init args, Instantiate, Invoke, Stream, and Async sections
- Provide detailed parameter descriptions and code examples
- Fix linting issues for code formatting compliance

Contributes to #24803

### DIFF
--- a/libs/partners/ollama/langchain_ollama/llms.py
+++ b/libs/partners/ollama/langchain_ollama/llms.py
@@ -24,15 +24,93 @@ from ._utils import validate_model
 
 
 class OllamaLLM(BaseLLM):
-    """OllamaLLM large language models.
+    """Ollama large language models.
 
-    Example:
+    Setup:
+        Install ``langchain-ollama`` and install/run the Ollama server locally:
+
+        .. code-block:: bash
+
+            pip install -U langchain-ollama
+            # Visit https://ollama.com/download to download and install Ollama
+            # Then start the server: ollama serve
+
+        Download a model to use:
+
+        .. code-block:: bash
+
+            ollama pull llama3.1
+
+    Key init args — generation params:
+        model: str
+            Name of the Ollama model to use (e.g. "llama3.1", "llama2").
+        temperature: Optional[float]
+            Sampling temperature. Higher values make output more creative.
+        num_predict: Optional[int]
+            Maximum number of tokens to predict.
+        top_k: Optional[int]
+            Limits the next token selection to the K most probable tokens.
+        top_p: Optional[float]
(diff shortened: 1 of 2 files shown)


libs/partners/ollama/langchain_ollama/llms.py
```diff
@@ -24,15 +24,93 @@
...
+    """Ollama large language models.
+
+    Setup:
+        Install ``langchain-ollama`` and install/run the Ollama server locally:
+
+        .. code-block:: bash
+
+            pip install -U langchain-ollama
+            # Visit https://ollama.com/download to download and install Ollama
+            # Then start the server: ollama serve
```
mdrxy:
```suggestion
```

I think this only applies to [Ollama on Linux](https://github.com/ollama/ollama/blob/main/docs/linux.md#manual-install) - IIRC, at least on macOS, this is handled automatically. Let me know your thoughts

libs/partners/ollama/langchain_ollama/llms.py
```diff
@@ -24,15 +24,93 @@
...
+
+        Download a model to use:
+
+        .. code-block:: bash
+
+            ollama pull llama3.1
+
+    Key init args — generation params:
+        model: str
+            Name of the Ollama model to use (e.g. "llama3.1", "llama2").
```
mdrxy:
```suggestion
            Name of the Ollama model to use (e.g. ``'llama4'``).
```

Tend to prefer newer models. Also formatting.

mdrxy: (unless there's a compat issue with `llama4` I'm not aware of)

libs/partners/ollama/langchain_ollama/llms.py
raviraj-441: You're correct, on macOS and Windows, the server starts automatically. I wasn't sure about Linux, so I mentioned `ollama serve` just in case. I'll leave it as is unless you think it's better to remove it entirely.

libs/partners/ollama/langchain_ollama/llms.py
raviraj-441: I'll update the model example to 'llama4' and adjust the formatting with backticks as you suggested.

libs/partners/ollama/langchain_ollama/llms.py
mdrxy:
```suggestion
            # (Linux users): start the server with ``ollama serve``
```

Issue #32747, opened 2025-08-29
langchain-text-splitters depends on pip which seems to be needed only for development purposes


kopczynski-9livesdata:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

pyproject.toml file:

```toml
[project]
name="langchain_reproduction"
version = "0.1.0"
requires-python = ">=3.8"
dependencies = [
    "langchain-text-splitters==0.3.10",
]
```

environment prep:
```bash
tkopczynski@dotdata ~/t/langchain-reproduction> python -m venv .venv
tkopczynski@dotdata ~/t/langchain-reproduction> source .venv/bin/activate.fish
(.venv) tkopczynski@dotdata ~/t/langchain-reproduction [0|2]> uv pip install -r pyproject.toml
Resolved 27 packages in 736ms
Prepared 13 packages in 321ms
Uninstalled 1 package in 19ms
Installed 27 packages in 8ms
 + annotated-types==0.7.0
 + anyio==4.10.0
 + certifi==2025.8.3
 + charset-normalizer==3.4.3
 + h11==0.16.0
 + httpcore==1.0.9
 + httpx==0.28.1
 + idna==3.10
 + jsonpatch==1.33
 + jsonpointer==3.0.0
 + langchain-core==0.3.75
 + langchain-text-splitters==0.3.10
 + langsmith==0.4.20
 + orjson==3.11.3
 + packaging==25.0
 - pip==24.0
 + pip==25.2
 + pydantic==2.11.7
 + pydantic-core==2.33.2
 + pyyaml==6.0.2
 + requests==2.32.5
 + requests-toolbelt==1.0.0
 + sniffio==1.3.1
 + tenacity==9.1.2
 + typing-extensions==4.15.0
 + typing-inspection==0.4.1
 + urllib3==2.5.0
 + zstandard==0.24.0
```

### Error Message and Stack Trace (if applicable)

[...]

PR #32769, opened 2025-08-31
infra(text-splitters): remove pip

keenborder786:
- **Description:** Pip package removed from core dependencies
- **Issue:** #32747


mdrxy:
We wouldn't ever go backwards on a version; it's always forward

PR #32784, opened 2025-09-02
chore(core): add utf-8 encoding to `Path` `read_text`/`write_text`

cbornet:
## Summary

This PR standardizes all text file I/O to use `UTF-8`. This eliminates OS-specific defaults (e.g. Windows `cp1252`) and ensures consistent, Unicode-safe behavior across platforms.

## Breaking changes

Users on systems with a default encoding which is not utf-8 may see decoding errors from the following code paths:

* langchain_core.vectorstores.in_memroy.InMemoryVectorStore.load
* langchain_core.prompts.loading.load_prompt
* `from_template` in AIMessagePromptTemplate, HumanMessagePromptTemplate, SystemMessagePromptTemplate

## Migration

Change the encoding of files that are encoded with a non utf-8 encoding to utf-8.

### DIFF
--- a/libs/cli/scripts/generate_migrations.py
+++ b/libs/cli/scripts/generate_migrations.py
@@ -72,7 +72,7 @@ def generic(
     else:
         dumped = dump_migrations_as_grit(name, migrations)
 
-    Path(output).write_text(dumped)
+    Path(output).write_text(dumped, encoding="utf-8")
 
 
 def handle_partner(pkg: str, output: str | None = None) -> None:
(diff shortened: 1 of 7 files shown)


mdrxy:
https://docs.astral.sh/ruff/rules/unspecified-encoding/

Should ensure consistency

Also, windows doesn't use utf-8 by default so we may need to investigate remediation for those users

cbornet:
NB: currently missing encoding in `Path.read`/`write` is not detected by ruff rule PLW1514 but it will be in the future (already detected with `--preview` flag)

eyurtsev (approved):

PR #32813, opened 2025-09-04
refactor(langchain): refactor optional imports logic

cbornet:
* Use `importlib` to load dynamically the classes
* Removes missing package warnings

### DIFF
--- a/libs/langchain_v1/langchain/chat_models/base.py
+++ b/libs/langchain_v1/langchain/chat_models/base.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import functools
 import importlib
 import warnings
+from collections.abc import Callable
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -20,7 +21,7 @@ from langchain_core.runnables import Runnable, RunnableConfig, ensure_config
 from typing_extensions import override
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncIterator, Callable, Iterator, Sequence
+    from collections.abc import AsyncIterator, Iterator, Sequence
     from types import ModuleType
 
     from langchain_core.runnables.schema import StreamEvent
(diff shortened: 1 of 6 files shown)


libs/langchain_v1/langchain/chat_models/base.py
```diff
@@ -28,6 +28,47 @@
     from pydantic import BaseModel
 
 
+_SUPPORTED_PROVIDERS = {
+    "openai": ("langchain_openai", "ChatOpenAI"),
```
eyurtsev: I think we don't want to load things on `__init__` -- we want to have as fast of an import of langchain / langchain.chat_models as possible. Users would ideally only pay an import cost for the things they use

Refactor OK, but let's make it lazy and memoize appropriately?

cbornet: Ruff [rule PLC0415](https://docs.astral.sh/ruff/rules/import-outside-top-level/) recommends to do the imports at the module loading.
I think the `langchain_*` modules are quite fast to import, aren't they ?

eyurtsev: it's been a significant problem which is why we've swapped to lazy imports in many places. I think we saw import time grow to ~0.5 seconds, which is way too slow

eyurtsev: So we want as little of the code loaded up front, so that end of the day only the parts of the code that the user needs are loaded

cbornet: Ok. I can switch to a cached lazy load then.
Note that the big load time may come from one package alone. It would be interesting to identify it.

cbornet: done.

eyurtsev: > Ok. I can switch to a cached lazy load then. Note that the big load time may come from one package alone. It would be interesting to identify it.

That's very possible, but another issue is loading many small packages; e.g., we might be looking at a few ms per package. With 10 integrations, we've added 10-30 ms to init time.

cbornet: I’ve switched to lazy loading. Can you check?

cbornet:
Rebased to include latest changes.

cbornet:
Hi @mdrxy , would you have time to take a look ?

libs/langchain_v1/langchain/chat_models/base.py
```diff
@@ -13,13 +21,90 @@
...
     from collections.abc import AsyncIterator, Callable, Iterator, Sequence
+    from types import ModuleType
 
     from langchain_core.runnables.schema import StreamEvent
     from langchain_core.tools import BaseTool
     from langchain_core.tracers import RunLog, RunLogPatch
     from pydantic import BaseModel
 
 
+def _call(cls: type[BaseChatModel], **kwargs: Any) -> BaseChatModel:
```
cbornet, line 33: This could be replaced by `operator.call` in Python 3.11+

PR #32829, opened 2025-09-05
fix(core): fix `BaseMessage` `.text()` filtering out blocks with text

olieydt:
- **Description:** BaseMessage .text() filters for blocks of type "text" when content is a list of dict. Concatenated messages can result in type != "text" but containing a text prop. This results in calling .text() on the message but receiving "". To address this, only consider if a text prop is present and a str.

### DIFF
--- a/libs/core/langchain_core/messages/base.py
+++ b/libs/core/langchain_core/messages/base.py
@@ -98,11 +98,13 @@ class BaseMessage(Serializable):
             return self.content
 
         # must be a list
+        # if a block is a dict, it could be a merged type block
+        # check only if the text prop exists
         blocks = [
             block
             for block in self.content
             if isinstance(block, str)
-            or (block.get("type") == "text" and isinstance(block.get("text"), str))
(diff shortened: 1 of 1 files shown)


mdrxy:
You may want to take a look at the new implementations in our `wip-v1.0` branch - this will soon be outdated/may no longer be needed

ccurme (commented):
@olieydt can you say more about the context for this? What models are returning blocks with `"text"` keys that are not `"type": "text"`?

I think semantically it makes sense for `.text()` to only filter to blocks matching a certain type. It's easy to imagine situations where this is desirable:
```python
[
    {"type": "internal_thought_process": "text": "<LLM inner thoughts...>"},
    {"type": "text", "text": "Here is the answer..."},
]
```
^ in that example we'd likely just want `.text()` to return the output message text.

ccurme:
Closing for now— the failing tests here indicate that this is deliberate behavior. I'd suggest opening an issue and giving additional context so we can design a proper solution (potentially as part of langchain 1.0).

PR #32868, opened 2025-09-09
chore(langchain): add ruff rule BLE

cbornet:


libs/langchain/tests/unit_tests/chains/query_constructor/test_parser.py
```diff
@@ -143,27 +143,10 @@ def test_parse_date_value(x: str) -> None:
             '"2021-12-31T23:59:59Z"',
             {"datetime": "2021-12-31T23:59:59Z", "type": "datetime"},
         ),
-        (
-            '"invalid-datetime"',
```
cbornet, line 147: This test is useless, as it's testing a regular string, which we already have in another test.

libs/langchain/tests/unit_tests/conftest.py
```diff
@@ -110,11 +110,7 @@ def test_something():
                 # If we haven't yet checked whether the pkg is installed
                 # let's check it and store the result.
                 if pkg not in required_pkgs_info:
-                    try:
-                        installed = util.find_spec(pkg) is not None
```
cbornet, line 114: This was added [here](https://github.com/langchain-ai/langchain/pull/14373#pullrequestreview-1770960480).
But is it really useful ?

libs/langchain/tests/unit_tests/conftest.py
```diff
@@ -110,11 +110,7 @@ def test_something():
...
-                    except Exception:
```
cbornet, line 115: This was added [here](https://github.com/langchain-ai/langchain/pull/14373#pullrequestreview-1770960480).
But is it really useful ?

mdrxy (approved):

PR #32876, opened 2025-09-10
feat(openai): (v1) support pdfs passed via url in standard format

ccurme:
This requires a small refactor.

Currently we pass all standard inputs through `convert_to_openai_data_block` in langchain-core, which returns chat completions format. BaseChatOpenAI then translates to Responses format if needed.

PDFs specified via URLs are [not supported](https://platform.openai.com/docs/guides/pdf-files?api-mode=chat#file-urls) by chat completions, so ideally `convert_to_openai_data_block` continues to raise an informative ValueError there.

Simplest solution IMO is to enable `convert_to_openai_data_block` to return either Chat Completions or Responses format, and do the translation there instead of in BaseChatOpenAI.

BaseChatOpenAI continues to accept chat completions format directly on Responses path.

### DIFF
--- a/libs/core/langchain_core/messages/block_translators/openai.py
+++ b/libs/core/langchain_core/messages/block_translators/openai.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import json
 import warnings
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Optional, Union, cast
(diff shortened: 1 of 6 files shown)


ccurme:
Need to add back more descriptive warning / error if users pass `{"type": "file", "url": "..."}` to chat completions.

PR #32878, opened 2025-09-10
fix(core): honor `enabled=false` in nested tracing (#31986)

hinthornw:


mdrxy:
Don't think this is needed since we'll merge master into the branch

PR #26287, opened 2024-09-10
core[patch]: fix empty OpenAI tools when strict=True

baskaryan:
Fix #26232

PR #32879, opened 2025-09-10
fix(core): force overwrite additionalProperties to False in strict mode

zhongyu09:
# Description
This PR fixes a bug in _recursive_set_additional_properties_false used in function_calling.convert_to_openai_function.

Previously, schemas with "additionalProperties=True" were not correctly overridden when strict validation was expected, which could lead to invalid OpenAI function schemas.

The updated implementation ensures that:
- Any schema with "additionalProperties" already set will now be forced to False under strict mode.
- Recursive traversal of properties, items, and anyOf is preserved.
- Function signature remains unchanged for backward compatibility.

# Issue
When using tool calling in OpenAI structured output strict mode (strict=True), 400: "Invalid schema for response_format XXXXX 'additionalProperties' is required to be supplied and to be false" error raises for the parameter that contains dict type. OpenAI requires additionalProperties to be set to False.
Some PRs try to resolved the issue.
- PR #25169 introduced _recursive_set_additional_properties_false to recursively set additionalProperties=False.
- PR #26287 fixed handling of empty parameter tools for OpenAI function generation.
- PR #30971 added support for Union type arguments in strict mode of OpenAI function calling / structured output.

Despite these improvements, since Pydantic 2.11, it will always add `additionalProperties: True` for arbitrary dictionary schemas dict or Any (https://pydantic.dev/articles/pydantic-v2-11-release#changes). Schemas that already had additionalProperties=True in such cases were not being overridden, which this PR addresses to ensure strict mode behaves correctly in all cases.

# Dependencies
No Changes

### DIFF
--- a/libs/core/langchain_core/utils/function_calling.py
+++ b/libs/core/langchain_core/utils/function_calling.py
@@ -834,9 +834,11 @@ def _recursive_set_additional_properties_false(
         # Check if 'required' is a key at the current level or if the schema is empty,
         # in which case additionalProperties still needs to be specified.
         # If additionalProperties already be set to True, we need override it to False
-        if "required" in schema or (
-            "properties" in schema and not schema["properties"]
-        ) or "additionalProperties" in schema:
+        if (
+            "required" in schema
+            or ("properties" in schema and not schema["properties"])
+            or "additionalProperties" in schema
+        ):
             schema["additionalProperties"] = False
 
         # Recursively check 'properties' and 'items' if they exist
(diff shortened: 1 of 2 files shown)


eyurtsev (approved):

eyurtsev:
Thank you @zhongyu09

zhongyu09:
Hi @eyurtsev, I also create a PR https://github.com/langchain-ai/langchain/pull/32930 to automatic set required to include all properties in strict mode. I am not sure we need another parameter to control such behavior.

PR #32888, opened 2025-09-10
feat(core)!: raise exception if `type` value is different in dicts to merge

cbornet:
* Solves a TODO that was planned for v0.3 but still hasn't been done. See https://github.com/langchain-ai/langchain/pull/25858
* Removes commented code (ruff rule ERA)

### DIFF
--- a/libs/core/tests/unit_tests/messages/test_ai.py
+++ b/libs/core/tests/unit_tests/messages/test_ai.py
@@ -434,7 +434,7 @@ def test_content_blocks() -> None:
             {
                 "type": "non_standard",
                 "index": 0,
-                "value": {"type": "input_json_delta", "partial_json": "a"},
+                "value": {"type": "non_standard_tool", "partial_json": "a"},
             }
         ]
     )
@@ -443,7 +443,7 @@ def test_content_blocks() -> None:
             {
                 "type": "non_standard",
                 "index": 0,
-                "value": {"type": "input_json_delta", "partial_json": "b"},
+                "value": {"type": "non_standard_tool", "partial_json": "b"},
             }
         ]
     )
--- a/libs/core/langchain_core/utils/_merge.py
+++ b/libs/core/langchain_core/utils/_merge.py
@@ -132,22 +132,10 @@ def merge_lists(left: list | None, *others: list | None) -> list | None:
                         ):
                             if left_type != "non_standard":
                                 # standard + non_standard
-                                new_e: dict[str, Any] = {
-                                    "extras": {
-                                        k: v
-                                        for k, v in e["value"].items()
-                                        if k != "type"
-                                    }
-                                }
+                                new_e: dict[str, Any] = {"extras": e["value"]}
                             else:
                                 # non_standard + non_standard
-                                new_e = {
-                                    "value": {
-                                        k: v
-                                        for k, v in e["value"].items()
-                                        if k != "type"
-                                    }
-                                }
+                                new_e = {"value": e["value"]}
                                 if "index" in e:
                                     new_e["index"] = e["index"]
                         else:
--- a/libs/core/langchain_core/utils/_merge.py
+++ b/libs/core/langchain_core/utils/_merge.py
@@ -20,6 +20,8 @@ def merge_dicts(left: dict[str, Any], *others: dict[str, Any]) -> dict[str, Any]
     Raises:
         TypeError: If the key exists in both dictionaries but has a different type.
         TypeError: If the value has an unsupported type.
+        ValueError: If the two dictionaries have different values for the special key
+            'type'.
 
     Example:
         If `left = {"function_call": {"arguments": None}}` and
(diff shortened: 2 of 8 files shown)


cbornet:
cc @baskaryan as you introduced the TODO.

cbornet:
Some tests fail:
```python
assert AIMessageChunk(
        content=[{"index": 0, "text": "I am", "type": "text_block"}]
    ) + AIMessageChunk(
        content=[{"index": 0, "text": " indeed.", "type": "text_block_delta"}]
    ) == AIMessageChunk(
        content=[{"index": 0, "text": "I am indeed.", "type": "text_block"}]
    ), (
        "Concatenating when both content arrays are dicts with the same index "
        "and different types should merge without updating type"
    )
```
It seems the `merge_dicts` should now allow to merge with different types ?

So should we:

1. remove the TODOs and related code (basically revert #25858) ?
2. **or** modify the AIMessageChunk tests ?
3. **or** have an optional param in `merge_dicts` to allow merging dicts with different types ?
4. other ?

cbornet:
> So should we:
> remove the TODOs and related code (basically revert https://github.com/langchain-ai/langchain/pull/25858) ?
> or modify the AIMessageChunk tests ?
> or have an optional param in merge_dicts to allow merging dicts with different types ?
> other ?

@eyurtsev @mdrxy can I have your opinion on this ?

mdrxy:
@cbornet sorry, haven't been able to look at this closely. In what scenarios are mismatched keys attempted to merge? There may be some case where it is needed, and if so, I think a hard-coded conditional that permits them would make the most sense.

cbornet:
> In what scenarios are mismatched keys attempted to merge?

For instance in the example above, adding 2 `AIMessageChunk` uses `merge_dicts` on the content.
And the contents have different types `text_block` and `text_block_delta`.
I'll see if I can introduce a param.

cbornet:
I had a further look.
The failing test comes from this PR: https://github.com/langchain-ai/langchain/pull/22750 which was introduced before the commented code.
So it seems at least `langchain-anthropic` will try to concatenate types `text` and `text_delta`. See https://github.com/langchain-ai/langchain/pull/22687#issue-2341198011 which was the reason for the special handling of `type`:

> Anthropic events look like this:
> ```
> RawContentBlockStartEvent(content_block=TextBlock(text='', type='text'), index=0, type='content_block_start')
> RawContentBlockDeltaEvent(delta=TextDelta(text='<thinking>\nThe', type='text_delta'), index=0, type='content_block_delta')
> RawContentBlockDeltaEvent(delta=TextDelta(text=' provide', type='text_delta'), index=0, type='content_block_delta')
> ...
> RawContentBlockStartEvent(content_block=ToolUseBlock(id='toolu_01GJ6x2ddcMG3psDNNe4eDqb', input={}, name='get_weather', type='tool_use'), index=1, type='content_block_start')
> RawContentBlockDeltaEvent(delta=InputJsonDelta(partial_json='', type='input_json_delta'), index=1, type='content_block_delta')
> ```

Was the intention to fix `langchain-anthropic` to only add chunks with same types ? (note that `text` and `text_delta` types come from Anthropic itself, not `langchain-anthropic`.)
I lack context. Maybe @baskaryan you remember ?

Anyway, given all this, it seems to me merge_dicts should always accept different types or always reject it.

cbornet:
And actually, #22687 handled the case for `content_block_delta` by forcing `type` to `text` : `content_block["type"] = "text" `

https://github.com/langchain-ai/langchain/blob/f82de1a8d70268e901e9a23ba6ebee574a4428c2/libs/partners/anthropic/langchain_anthropic/chat_models.py#L2454-L2469

But it's still not OK for `tool_use`+`input_json_delta`:

https://github.com/langchain-ai/langchain/blob/f82de1a8d70268e901e9a23ba6ebee574a4428c2/libs/partners/anthropic/langchain_anthropic/chat_models.py#L2482-L2488

(`thinking_delta` and `signature_delta` are correctly changed to `thinking`)

cbornet:
cc @jacoblee93 as you are the author of https://github.com/langchain-ai/langchain/pull/22687

cbornet:
I added a fix to map `input_json_delta` to `tool_use` and fixed the tests. PTAL.

cbornet:
Translation was also done in langchain-aws for Bedrock: https://github.com/langchain-ai/langchain-aws/pull/119/files#diff-7b555324692cbefa47fc582ade684c4faa5d083023137e2839f37aff0401aa4eR128-R139

libs/core/tests/unit_tests/messages/test_ai.py
```diff
@@ -437,7 +437,7 @@ def test_content_blocks() -> None:
             {
                 "type": "non_standard",
                 "index": 0,
-                "value": {"type": "input_json_delta", "partial_json": "a"},
+                "value": {"type": "non_standard_tool", "partial_json": "a"},
```
cbornet, line 437: with the change, non-standard tools must at least map to same types in streamed chunks.
Is it OK ?

jacoblee93:
Thanks @cbornet! Adding @mdrxy here too.

cbornet:
Can this have some eyes ?
As it's breaking, it would be nice to be done before v1 release!

cbornet:
I rebased on master since v1 was released.
Do you want to do this ? or should we remove the TODO ?

PR #32922, opened 2025-09-12
feat(core): ChatPromptTemplate.to_messages

ccurme:
For convenience when working with prompt hub but not using langchain chat models.

### DIFF
--- a/libs/core/langchain_core/prompts/chat.py
+++ b/libs/core/langchain_core/prompts/chat.py
@@ -1243,6 +1243,16 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
                 raise ValueError(msg)  # noqa:TRY004
         return result
 
+    def to_messages(self) -> list[BaseMessage]:
+        """Convert template into a list of messages.
+
+        This method fills in the template variables with placeholder values.
+
+        Returns:
+            List of messages.
+        """
+        return self.format_messages(**{k: "{" + k + "}" for k in self.input_variables})
+
     def partial(self, **kwargs: Any) -> ChatPromptTemplate:
         """Get a new ChatPromptTemplate with some input variables already filled in.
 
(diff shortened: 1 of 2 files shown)


libs/core/langchain_core/prompts/chat.py
```diff
@@ -1243,6 +1243,16 @@ async def aformat_messages(self, **kwargs: Any) -> list[BaseMessage]:
...
 
+    def to_messages(self) -> list[BaseMessage]:
+        """Convert template into a list of messages.
+
+        This method fills in the template variables with placeholder values.
+
+        Returns:
+            List of messages.
+        """
+        return self.format_messages(**{k: "{" + k + "}" for k in self.input_variables})
```
eyurtsev, line 1254: Could you confirm that `k` cannot have `{` and `}` in it? What character set is it limited to? (Almost maybe a better name than `k`?)

On the face of it, the `"{" + k + "}"` part of the code looks like a perfect place for injecting malicious data (would be good to add an inline comment to explain what this does/why it's OK

libs/core/tests/unit_tests/prompts/test_chat.py
```diff
@@ -1236,3 +1236,46 @@ def test_dict_message_prompt_template_errors_on_jinja2() -> None:
...
+                        "url": "{url}",
+                    },
+                ],
+            },
+        ]
+    )
+    result = prompt.to_messages()
+    expected = [
+        SystemMessage("Describe the image provided."),
+        HumanMessage(
```
eyurtsev, line 1272: Does this not need to be `HumanMessagePromptTemplate`? I thought we don't support variables in `HumanMessage` ?

libs/core/langchain_core/prompts/chat.py
```diff
@@ -1243,6 +1243,16 @@ async def aformat_messages(self, **kwargs: Any) -> list[BaseMessage]:
                 raise ValueError(msg)  # noqa:TRY004
         return result
 
+    def to_messages(self) -> list[BaseMessage]:
+        """Convert template into a list of messages.
+
+        This method fills in the template variables with placeholder values.
```
eyurtsev, line 1249: Why is it useful to have placeholder values?

PR #32934, opened 2025-09-14
feat(core): export `env_var_is_set`

adityamulik:
Creating a PR to expose env_var_is_set method from langchain core utils which is useful for validating if a env variable is set or not.


ccurme (commented):
Thanks for this. This is a public function but I'm not sure it's needed in the `langchain_core.utils` namespace. It is just
```python
return env_var in os.environ and os.environ[env_var] not in {
    "",
    "0",
    "false",
    "False",
}
```
are you ok continuing to just import from `langchain_core.utils.env`?

PR #32947, opened 2025-09-15
fix(core): resolve `trim_messages` incompatibility with DeepSeek-Chat

hispiderman0528:
Fixed the incompatibility issue of the function from langchain_core.messages import trim_messages with the DeepSeek-Chat large model.


mdrxy (changes requested):
Could you list a source for reference? And give a code snippet of a use case that is broken without this?

ccurme:
Closing as author is unresponsive but can re-open if needed!

PR #32973, opened 2025-09-16
chore(langchain): (v1) drop cap on python version

ccurme:


ccurme:
Should be done on master and merged in.

Issue #32981, opened 2025-09-16
Reasoning tokens not passing through from OpenRouter to `ChatOpenAI`


sinanuozdemir:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

Run this code and inspect the response and see that it has no reasoning tokens even though anthropic returns them.

```python
import os
from langchain_openai import ChatOpenAI
from langchain_core.messages import HumanMessage


chat = ChatOpenAI(
    model="anthropic/claude-sonnet-4",
    api_key=os.environ.get("OPENROUTER_API_KEY"),
    openai_api_base="https://openrouter.ai/api/v1"
)

message = HumanMessage(content="Hello! What's the capital of France now and what was it in 1937?")


response = chat.invoke(
    [message],
    extra_body={'reasoning': {'max_tokens': 10}}
    )
print(f"📥 Response: {response.content}")
```

### Description

I'm using Anthropic Sonnet 4 through OpenRouter on Langchain (for agentic work) and noticed that the reasoning tokens aren't getting passed from OpenRouter to Langchain

### System Info

Obtaining file:///Users/sinanozdemir/Fun/langchain/libs/core
  Installing build dependencies ...   Checking if build backend supports build_editable ...   Getting requirements to build editable ...   Preparing editable metadata (pyproject.toml) ... Requirement already satisfied: langsmith>=0.3.45 in [./.venv/lib/python3.12/site-packages](https://file+.vscode-resource.vscode-cdn.net/Users/sinanozdemir/Fun/langchain/.venv/lib/python3.12/site-packages) (from langchain-core==0.3.76) (0.4.28)
[...]

PR #32982, opened 2025-09-16
fix(openai): add support for reasoning content in OpenAI chat responses from OpenRouter

sinanuozdemir:
Description: I added `reasoning_content` to `additional_kwargs` when the reasoning field is present in the response message. This enables support for reasoning content from providers like OpenRouter that include this field in their responses.
Issue: #32981 
Dependencies: N/A

P.S. this is my first time ever attempting to contribute to an open source package so feel free to rip me to shreds.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -934,6 +934,7 @@ class BaseChatOpenAI(BaseChatModel):
             if (reasoning_content := delta.get("reasoning_content")) is not None:
                 message_chunk.additional_kwargs["reasoning_content"] = reasoning_content
             elif (reasoning := delta.get("reasoning")) is not None:
+                # Handle use via OpenRouter
                 message_chunk.additional_kwargs["reasoning_content"] = reasoning
 
         generation_chunk = ChatGenerationChunk(
@@ -1309,7 +1310,9 @@ class BaseChatOpenAI(BaseChatModel):
                     "reasoning_content"
                 ] = message.reasoning
             elif hasattr(message, "reasoning_content"):
-                generations[0].message.additional_kwargs["reasoning_content"] = message.reasoning_content
+                generations[0].message.additional_kwargs[
+                    "reasoning_content"
+                ] = message.reasoning_content
 
         return ChatResult(generations=generations, llm_output=llm_output)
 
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Hi @sinanuozdemir, thanks for this.

You should be able to resolve your issue immediately by using `ChatDeepSeek` (I see from the originating issue you have `langchain-deepseek` already installed). That model basically subclasses BaseChatOpenAI and extends it with this feature. Let me know if that doesn't work for you.

There is a lot of confusion around this, and it's arguably weird to force use of ChatDeepSeek for some chat completions use cases, so I'm not opposed to going with this PR. If you want to add unit tests and extend to the streaming case, I think we could push it through.

MaciejoG:
Came across this PR while researching why using the `ChatOpenAI` with a model from Anthropic does not output any reasoning tokens (for my usecase, I cannot use the `ChatAnthropic` class).
Slightly modifying the changes from this PR (instead of `reasoning` look for `reasoning_content`) yields a positive result!

```
if hasattr(message, "reasoning_content"):
                generations[0].message.additional_kwargs[
                    "reasoning_content"
                ] = message.reasoning_content
```

Output contains reasoning_content
```
content='The capital of France is Paris.' additional_kwargs={'refusal': None, 'reasoning_content': "This is a straightforward factual question. The capital of France is Paris. This is basic geography knowledge that I'm confident about."}
```

@sinanuozdemir are you going to continue working on this PR? Would be great to make it functional for `reasoning_content` too!

sinanuozdemir:
I'm going to try to take this over the finish line in the next few days @ccurme and @MaciejoG! Thank you both for the tips.

sinanuozdemir:
@ccurme I think I'm ready for another review!

Saryazdi-Saman:
I came across this PR while running into the same issue. I had been using the OpenAI SDK directly for Claude via OpenRouter, since I couldn’t find documentation that ChatDeepSeek could be used in that context.

> "There is a lot of confusion around this, and it's arguably weird to force use of ChatDeepSeek for some chat completions use cases … If you want to add unit tests and extend to the streaming case, I think we could push it through."

@ccurme I agree with your point about developer confusion — requiring ChatDeepSeek for some chat completion use cases isn’t intuitive, especially since OpenRouter's docs direct to ChatOpenAI for Langchain implementation.

What do you think about the idea of inheriting from BaseChatOpenAI and creating a dedicated ChatOpenRouter wrapper? That might make the integration cleaner and reduce confusion. I’d be happy to contribute code for this (in TS and possibly Python) if that direction aligns with the project.

mdrxy:
@Saryazdi-Saman yep

We will be releasing a dedicated integration for OpenRouter (`ChatOpenRouter`) as they have diverged from the chat completions standard. See #34328 for tracking

Issue #32637, opened 2025-08-21
`ChatPromptTemplate` `save()` not implemented


zzqDeco:
### Checked other resources

- [x] This is a bug, not a usage question. For questions, please use the LangChain Forum (https://forum.langchain.com/).
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

#13794
```python
     def save(self, file_path: Union[Path, str]) -> None: 
         """Save prompt to file. 
  
         Args: 
             file_path: path to file. 
         """ 
         raise NotImplementedError() 
```

### Description

ChatPromptTemplate save method not implemented

### System Info

ChatPromptTemplate save method not implemented!Looking at ur fxxking lib!

ccurme:
Could we please address the failing test in https://github.com/langchain-ai/langchain/pull/33631?

UnkownHunter:
Can this issue be assigned to me ?.. The save function in the ChatPromptTemplate still not implemented... I want to implement if assigned to me.

yaowubarbara:
I'd like to work on this issue.

ccurme:
These routines are old, I'd suggest using `prompt.model_dump` if all you need is serialization, or [see the reference here](https://reference.langchain.com/python/langchain_core/load/).

PR #32997, opened 2025-09-18
fix(core): Enable save method for `ChatPromptTemplate` by removing `NotImplementedError` override

qizwiz:
This PR fixes Issue #32637 where ChatPromptTemplate's save method was raising NotImplementedError instead of inheriting the working implementation from BasePromptTemplate.

## Issue

The ChatPromptTemplate class was overriding the parent class's working `save` method with a stub that just raised `NotImplementedError`. This prevented users from saving ChatPromptTemplate instances to files, even though the parent class BasePromptTemplate has a fully functional save implementation.

## Root Cause

In `src/transformers/processing_utils.py` line 1303, there was an override:
```python
def save(self, file_path: Union[Path, str]) -> None:
    """Save prompt to file.

    Args:
        file_path: path to file.
    """
    raise NotImplementedError
```

## Solution

The fix removes this override, allowing ChatPromptTemplate to inherit the working `save` method from BasePromptTemplate. This enables users to save ChatPromptTemplate instances to JSON/YAML files as expected.

## Testing

Added a test case that verifies:
1. ChatPromptTemplate.save() no longer raises NotImplementedError
2. The saved file contains valid JSON with the expected structure
3. The _type field is correctly set to 'chat'

## Impact

This is a backward-compatible bug fix that restores expected functionality for ChatPromptTemplate serialization. Users can now save and load ChatPromptTemplate instances as documented, which is important for persisting prompt configurations.

Fixes #32637

### DIFF
--- a/libs/core/langchain_core/prompts/chat.py
+++ b/libs/core/langchain_core/prompts/chat.py
@@ -1300,13 +1300,6 @@ class ChatPromptTemplate(BaseChatPromptTemplate):
         """Name of prompt type. Used for serialization."""
         return "chat"
 
-    def save(self, file_path: Union[Path, str]) -> None:
-        """Save prompt to file.
-
-        Args:
-            file_path: path to file.
-        """
-        raise NotImplementedError
 
     @override
     def pretty_repr(self, html: bool = False) -> str:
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Closing as duplicated with https://github.com/langchain-ai/langchain/pull/33631.

PR #33012, opened 2025-09-18
fix(core): accept fenced JSON in tool call arguments

TokuiNico:
### Summary
Accept Markdown-fenced JSON in tool call arguments when using with_structured_output. This aligns the structured-output path with the robustness already present in PydanticOutputParser, while keeping strictness for incomplete payloads.

### Changes
- Use parse_json_markdown in parse_tool_call for the non-partial path (no partial fill). 
- Keep strict json.loads under the hood; malformed/incomplete JSON still raises.
- Add unit tests covering:
  - Normal JSON
  - Fenced JSON
  - Incomplete JSON (expected to error)

### Rationale
- Models may wrap JSON in ```json ... ``` fences. PydanticOutputParser already tolerates this via parse_json_markdown; with_structured_output should behave similarly so users get consistent parsing across pathways.
- We deliberately avoid parse_partial_json here to prevent accidentally accepting truncated tool arguments.

### Testing
- All core unit tests pass locally. The new tests specifically verify:
  - Normal JSON parses successfully
  - Fenced JSON parses successfully
  - Incomplete JSON raises OutputParserException

### Backwards Compatibility
- No API changes. For already-valid JSON arguments, behavior is unchanged. This only increases robustness for fenced JSON content.


ccurme (commented):
I'm not sure we want to do this, `parse_json_markdown` is using regex under the hood to parse patterns that were generated in legacy (pre tool-calling) agents.

Would you mind opening an issue explaining in what contexts you've seen this? I'm curious about how often / in what contexts models will populate markdown in tool call arguments.

PR #33021, opened 2025-09-19
feat: (v1) server tool call and result types

ccurme:
- Replaces dedicated types for `WebSearchCall`, `WebSearchResult`, `CodeInterpreterCall`, `CodeInterpreterResult` with monolithic `ServerToolCall` and `ServerToolResult`
- Implements `ServerToolCallChunk` for streaming partial arguments
- Full support on Anthropic and OpenAI


libs/core/langchain_core/messages/block_translators/anthropic.py
```diff
@@ -270,159 +270,102 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
-                            web_search_call["extras"][key] = value
-                    yield web_search_call
-
-                elif block.get("name") == "code_execution":
-                    code_interpreter_call: types.CodeInterpreterCall = {
-                        "type": "code_interpreter_call"
+                else:
+                    server_tool_call_chunk: types.ServerToolCallChunk = {
+                        "type": "server_tool_call_chunk",
+                        "args": block.get("partial_json", ""),
```
ccurme, line 287: It does seem risky to infer that `input_json_delta` types, if there are no accompanying tool calls on the AIMessageChunk, are server tool call chunks. But I haven't found a counterexample to this.

libs/core/langchain_core/messages/block_translators/anthropic.py
```diff
@@ -270,159 +270,149 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
+                    known_fields = {"type", "name", "input", "id", "index"}
+                    _populate_extras(server_tool_call, block, known_fields)
+                    if "index" in block:
+                        server_tool_call["index"] = block["index"]
+
+                    yield server_tool_call
 
-                code_interpreter_output: types.CodeInterpreterOutput = {
-                    "type": "code_interpreter_output"
+            elif block_type in (
```
ccurme: should we replace this with `block_type.endswith("_tool_result")`, to catch new block types?

libs/core/langchain_core/messages/block_translators/anthropic.py
```diff
@@ -270,159 +270,149 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
+                    }
+                    if "index" in block:
+                        server_tool_call_chunk["index"] = block["index"]
+                    yield server_tool_call_chunk
 
-                    if "id" in block:
-                        web_search_call["id"] = block["id"]
+            elif block_type == "server_tool_use":
+                if block.get("name") == "code_execution":
+                    server_tool_use_name = "code_interpreter"
```
hntrl, line 295: how come this is needed?

libs/core/langchain_core/messages/block_translators/openai.py
```diff
@@ -713,21 +713,27 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
-                if (
-                    "action" in block
-                    and isinstance(block["action"], dict)
-                    and block["action"].get("type") == "search"
-                    and "query" in block["action"]
-                ):
-                    web_search_call["query"] = block["action"]["query"]
+
+                sources: Optional[dict[str, Any]] = None
+                if "action" in block and isinstance(block["action"], dict):
```
hntrl, line 726: this might just be my inability to read python code, but what are we doing with this sources dict?

libs/core/langchain_core/messages/block_translators/openai.py
```diff
@@ -736,55 +742,227 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
+                if file_search_output := block.get("results"):
+                    file_search_result["output"] = file_search_output
+
+                status = block.get("status")
+                if status == "failed":
+                    file_search_result["status"] = "error"
+                elif status == "completed":
+                    file_search_result["status"] = "success"
+                elif status:
+                    file_search_result["extras"] = {"status": status}
```
hntrl, line 801: nit: is this important to have? can we just be smart about populating status on openAI's end?

libs/core/langchain_core/messages/block_translators/openai.py
ccurme: the openai type has `action`, where one of the allowed values is a search:
```python
class ActionSearch(BaseModel):
    query: str
    """The search query."""

    type: Literal["search"]
    """The action type."""

    sources: Optional[List[ActionSearchSource]] = None
    """The sources used in the search."""
```
here I take out the `sources` and put them on the tool output, below. then the rest of the fields (query, type) I put in `args`.

libs/core/langchain_core/messages/block_translators/openai.py
ccurme: status here can be
```
"in_progress", "searching", "completed", "incomplete", "failed"
```
I'm not sure in what contexts some of these would show up. but how else would we reconstruct the original value if they don't hitch a ride on the block somewhere? I might not be following.

libs/core/langchain_core/messages/block_translators/anthropic.py
ccurme: it's not, but makes it consistent with openai. I think with just this change the values for `name` and `args` for both web search and code interpreter are consistent for openai and anthropic.

libs/core/langchain_core/messages/block_translators/anthropic.py
hntrl: Would it be radical to say that we should make `server_tool_call_chunk` a first class citizen inside of `message.tool_call_chunks`? E.g. either making that field be a union of `ToolCallChunk | ServerToolCallChunk`, or adding a `type` field to ToolCallChunk

libs/core/langchain_core/messages/block_translators/anthropic.py
hntrl: I'm fine to do this, but I should say I can see some confusion coming up if the way we tell people to use the tool is by specifying `code_execution`, but the corresponding content block takes on the form of `code_interpreter`

Screenshot 2025-09-19 at 7 41 35 PM

hntrl: Also side Q: should we be telling people the way to consume these tool outputs is by using `.content_blocks` in docs? (I guess this is a broader discussion if we want to default to `.content_blocks` instead of `.content` in docs)

libs/core/langchain_core/messages/block_translators/anthropic.py
hntrl: yeah I think we should

libs/core/langchain_core/messages/block_translators/openai.py
hntrl: Does it make more sense to add a `.artifact` property on `ServerToolResult`, shove the raw model output there, and then use that instead of the server tool call object when we format it? I see the need for this, but I am partial to limiting the amount of magic we do when it comes to server tools

libs/core/langchain_core/messages/block_translators/anthropic.py
ccurme: what's the motivation? I was concerned about `input_json_delta` blocks that are neither tool calls nor server tool use.

libs/core/langchain_core/messages/block_translators/anthropic.py
ccurme: that's fair. my bias is to define standard types for web search + code interp tools and accept those as well in `bind_tools`. so then it would be consistent.

in docs updates I've been using `content_blocks` unless it's mostly non-standard (after the move to ServerToolCall there's very few non-standard).

libs/core/langchain_core/messages/block_translators/openai.py
ccurme: The only thing we're standardizing at the moment on `ServerToolResult` is `tool_call_id` and `status`. Most of the data is in `output`, and it can be pretty large (e.g., web search with Anthropic can return encrypted content for many sources). So I'm reluctant to duplicate it in `artifact`.

mdrxy (approved):

PR #33025, opened 2025-09-20
fix(core): Resolve `aadd_texts() got multiple values for keyword argument 'ids'`

qizwiz:
This PR fixes Issue #32283 where calling aadd_documents() with an ids parameter would cause a ValidationError due to parameter duplication.

## Issue

When users call:
```python
await vector_store.aadd_documents(documents, ids=["custom_id_1", "custom_id_2"])
```

They get the error:
```
TypeError: aadd_texts() got multiple values for keyword argument 'ids'
```

## Root Cause

The issue is in `VectorStore.aadd_documents()` where the method was passing the `ids` parameter both positionally and in `**kwargs` to `aadd_texts()`, causing Python to raise the duplication error.

## Solution

The fix extracts the `ids` parameter from `kwargs` before passing it to `aadd_texts()`, preventing the duplication while maintaining backward compatibility.

## Testing

Added a comprehensive test case that verifies:
1. `aadd_documents` works correctly with custom ids parameter
2. `aadd_documents` still works without ids parameter (backward compatibility)
3. The fix prevents the TypeError from occurring

## Impact

This is a backward-compatible bug fix that resolves a critical issue preventing users from specifying custom IDs when using async document addition with vector stores.

Fixes #32283


libs/core/langchain_core/vectorstores/base.py
```diff
@@ -307,7 +319,9 @@ async def aadd_documents(
 
             texts = [doc.page_content for doc in documents]
             metadatas = [doc.metadata for doc in documents]
-            return await self.aadd_texts(texts, metadatas, **kwargs)
+            # Extract ids from kwargs to avoid duplication with positional parameter
```
eyurtsev, line 288: Hi @qizwiz thanks for the PR. I don't quite see why this is fixing the issue. Could you add a unit test that reproduces the problem?

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #33045, opened 2025-09-21
feat(core): add `'approximate'` alias in place of `count_tokens_approximately`

Davda-James:
### Description: 
earlier we have to use like below:
```python
from langchain_core.messages import trim_messages
from langchain_core.messages.utils import count_tokens_approximately

trim_messages(..., token_counter=count_tokens_approximately)
```
Now can be used as like this also
```python
from langchain_core.messages import trim_messages

trim_messages(..., token_counter="approximate")
```
- [x] **Added tests**
- [x] **Lint and test**:  Run this as I made change in langchain/core,  uv run --group test pytest tests/unit_tests/messages/test_utils.py -v
image


libs/core/langchain_core/messages/utils.py
```diff
@@ -688,6 +688,7 @@ def trim_messages(
     *,
     max_tokens: int,
     token_counter: Union[
+        str,
```
eyurtsev: Let's use a Literal instead of str to enumerate the pre-built options.

Could we use "approximate" instead of "approx" so it's self-explanatory?

Davda-James: Yeah sure let me add it.

Davda-James: @eyurtsev have changed it and amend it in same commit. Changed the alias, test function name, comments and function name in code from approx to approximate .Changed the title, description, usage of PR as well.

eyurtsev: Thank you great!!

eyurtsev:
@Davda-James could you undo the changes in the lock file? And we can merge then?

Davda-James:
@eyurtsev reverted the lock file changes in new commit.

Davda-James:
Hey  @eyurtsev any further changes needed or anything to add ?

PR #33062, opened 2025-09-23
chore(langchain_v1): remove union return type in init_embeddings

eyurtsev:
Fix the return type of init_embeddings


sydney-runkle (approved):
Wonderful!

Issue #33129, opened 2025-09-26
`create_agent()` fails with `RunnableWithFallbacks` when using middleware


bart0401:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```python
from langchain.agents import create_agent
from langchain.agents.middleware import SummarizationMiddleware
from langchain_openai import ChatOpenAI
from langchain.tools import tool

# Setup models
main_model = ChatOpenAI(model="gpt-4")
fallback_model = ChatOpenAI(model="gpt-3.5-turbo")

# Create a model with fallback - this returns RunnableWithFallbacks
model_with_fallback = main_model.with_fallbacks([fallback_model])

# Setup middleware for conversation summarization
middleware = [
    SummarizationMiddleware(
        model=main_model,
        max_tokens_before_summary=10000,
        messages_to_keep=20
    )
]

# Define a simple tool for the agent
@tool
def get_weather(city: str) -> str:
    """Get weather information for a city."""
    return f"The weather in {city} is sunny."

# This fails with AssertionError
agent = create_agent(
    model=model_with_fallback,  # RunnableWithFallbacks type
    prompt="You are a helpful assistant that can provide weather information.",
    tools=[get_weather],
    middleware=middleware  # When middleware exists, assertion fails
)

# Run the agent
result = agent.invoke({"input": "What's the weather in New York?"})
print(result)
```

### Error Message and Stack Trace (if applicable)

```
Traceback (most recent call last):
  File "test_agent.py", line 24, in <module>
    agent = create_agent(
  File "/path/to/langchain/agents/react_agent.py", line 1145, in create_agent
    assert isinstance(model, str | BaseChatModel)
AssertionError
```

The assertion error occurs at line 1145 in `react_agent.py`:
```python
if middleware:
    assert isinstance(model, str | BaseChatModel)  # This fails for RunnableWithFallbacks
```

### Description

**What I'm trying to do:**
[...]

PR #33130, opened 2025-09-26
fix(langchain_v1): Allow RunnableWithFallbacks with middleware in create_agent()

Swastik-Swarup-Dash:
#33129 
Changed from
# Old
```assert isinstance(model, str BaseChatModel)  # noqa: S101```

# New
```assert isinstance(model, (str, BaseChatModel, RunnableWithFallbacks))  # noqa: S101```


eyurtsev:
Thanks for the PR! Let's discuss in the issue

Issue #33153, opened 2025-09-29
The Tool error handling is not working if the Agent created with middleware


ngockhanh5110:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

Working code:

```python
from langchain_aws import ChatBedrockConverse
from langchain.agents import ToolNode, create_agent
from langchain_core.tools import tool
import random


@tool
def fetch_user_data(user_id: str) -> str:
    """Fetch user data from database."""
    if random.random() > 0.7:
        raise ConnectionError("Database connection timeout")
    return f"User {user_id}: John Doe, john@example.com, Active"

@tool
def process_transaction(amount: float, user_id: str) -> str:
    """Process a financial transaction."""
    if amount > 10000:
        raise ValueError(f"Amount {amount} exceeds maximum limit of 10000")
    return f"Processed ${amount} for user {user_id}"

def handle_errors(e: Exception) -> str:
    if isinstance(e, ConnectionError):
        return "The database is currently overloaded, but it is safe to retry. Please try again with the same parameters."
    elif isinstance(e, ValueError):
        return f"Error: {e}. Try to process the transaction in smaller amounts."
    return f"Error: {e}. Please try again."

tool_node = ToolNode(
    tools=[fetch_user_data, process_transaction],
    handle_tool_errors=handle_errors
)

agent = create_agent(
    model=ChatBedrockConverse(model="global.anthropic.claude-sonnet-4-20250514-v1:0", region_name="us-west-2"),
    tools=tool_node,
    prompt="You are a financial assistant."
)

agent.invoke({
    "messages": [{"role": "user", "content": "Process a payment of 15000 dollars for user123. Generate a receipt email and address it to the user."}]
})
```

Not Working:

```python
...
agent = create_agent(
   ...
   middlewares=SomeMiddleware()
   ...
)
...
```

### Description

[...]

PR #33221, opened 2025-10-02
fix(langchain_v1): ToolNode initialization uses tool attributes in create_agent for middleware

keenborder786:
- **Description:** The additional attributes of tools are not being passed when it is an instance of ToolNode, which causes buggy behavior for clients creating agents with middleware.
- **Issue:** #33153


keenborder786:
@sydney-runkle

PR #33261, opened 2025-10-03
fix(core): fix string content when streaming output_version="v1"

ccurme:


libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -936,6 +988,41 @@ def test_output_version_stream(monkeypatch: Any) -> None:
     assert isinstance(full_v1, AIMessageChunk)
     assert full_v1.response_metadata["output_version"] == "v1"
 
+    assert full_v1.content == [{"type": "text", "text": "foo bar", "index": 0}]
```
eyurtsev, line 991: Did we need a test case where it's just regular text blocks -- no reasoning content?

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
```diff
@@ -907,6 +909,56 @@ async def test_output_version_ainvoke(monkeypatch: Any) -> None:
     assert response.response_metadata["output_version"] == "v1"
 
 
+class _AnotherFakeChatModel(BaseChatModel):
```
mdrxy, line 912: why?

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
mdrxy: i think the reasoning content is additive (it doesn't change underlying text case)

mdrxy (approved):
Should we add this to standard tests?

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
ccurme: added for good measure

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
ccurme: afaik there's no fake chat model that will stream content blocks.

mdrxy: not high prio but would it be worth adding to an existing? we have so many fake models 😢

ccurme:
> Should we add this to standard tests?

Will add!

libs/core/tests/unit_tests/language_models/chat_models/test_base.py
ccurme: yeah, that's why I added a private class here, because I was hesitant to add yet another. if we battle test / use this more / do a broader cleanup where we replace some of the older ones, I'm down to consolidate.

PR #33267, opened 2025-10-04
feat(langchain_v1): Implement Context Editing Middleware 

nfcampos:
Brings functionality similar to Anthropic's context editing to all chat models
https://docs.claude.com/en/docs/build-with-claude/context-editing


libs/langchain_v1/langchain/agents/middleware/context_editing.py
```diff
@@ -0,0 +1,249 @@
...
+    ) -> int:
+        """Apply an edit to the message list, returning the new token count."""
+        ...
+
+
+@dataclass(slots=True)
+class ClearToolUsesEdit(ContextEdit):
+    """Configuration for clearing tool outputs when token limits are exceeded."""
+
+    trigger_tokens: int = 100000
```
eyurtsev:
```suggestion
    trigger_tokens: int = 100_000
```

libs/langchain_v1/langchain/agents/middleware/context_editing.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+
+    edits: list[ContextEdit]
+    token_count_method: Literal["approx", "model"]
+
+    def __init__(
+        self,
+        *,
+        edits: Iterable[ContextEdit] | None = None,
+        token_count_method: Literal["approx", "model"] = "approx",  # noqa: S107
```
eyurtsev: 1. OK if we remove `model`? I think in practice it's **much** slower than approx and likely still has issues handling token counts correctly in some cases. 
2. "approx" -> "approximate" ? (for non native speakers?)

libs/langchain_v1/langchain/agents/middleware/context_editing.py
```diff
@@ -0,0 +1,249 @@
...
+DEFAULT_TOOL_PLACEHOLDER = "[cleared]"
+
+
+TokenCounter = Callable[
+    [Sequence[BaseMessage]],
+    int,
+]
+
+
+class ContextEdit(Protocol):
```
eyurtsev, line 39: nice i like that it's lifted to a protocol

libs/langchain_v1/langchain/agents/middleware/context_editing.py
nfcampos: I think that depends on the model, right? Eg. for openai seems to be just a python routine so as fast as approximate? I did make approx default on purpose, as I think that's more than good enough for most use cases

eyurtsev: openai delegates to tiktoken -- likely a few orders of magnitude difference in speed

for accuracy, I don't trust the current implementations of get_num_tokens_from_message to work correctly across models anyway..so don't really see a reason to expose this as an option at this stage.

IMO we remove this for now, and add after checking in with integrations team.

sydney-runkle (approved):
Cool, I like the extensibility of a protocol!

PR #33271, opened 2025-10-04
feat(langchain_v1): Implement PIIMiddleware

nfcampos:
- supports 6 well-known PII types (email, credit_card, ip, mac_address, url)
- 4 handling strategies (block, redact, mask, hash)
- supports custom PII types with detector functions or regex
- the built-in types were chosen because they are common, and detection can be reliably implemented with stdlib

### DIFF
--- a/libs/langchain_v1/langchain/agents/middleware/pii.py
+++ b/libs/langchain_v1/langchain/agents/middleware/pii.py
@@ -1,0 +1,753 @@
+"""PII detection and handling middleware for agents."""
+
+from __future__ import annotations
+
+import hashlib
+import ipaddress
+import re
+from typing import TYPE_CHECKING, Any, Literal
+from urllib.parse import urlparse
+
+from langchain_core.messages import AIMessage, AnyMessage, HumanMessage, ToolMessage
+from typing_extensions import TypedDict
+
+from langchain.agents.middleware.types import AgentMiddleware, AgentState, hook_config
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from langgraph.runtime import Runtime
+
+
+class PIIMatch(TypedDict):
+    """Represents a detected PII match in text."""
+
+    type: str
+    """The type of PII detected (e.g., 'email', 'ssn', 'credit_card')."""
+    value: str
+    """The actual matched text."""
+    start: int
+    """Starting position of the match in the text."""
+    end: int
+    """Ending position of the match in the text."""
+
+
+class PIIDetectionError(Exception):
+    """Exception raised when PII is detected and strategy is 'block'."""
+
+    def __init__(self, pii_type: str, matches: list[PIIMatch]) -> None:
+        """Initialize the exception with PII detection information.
+
+        Args:
+            pii_type: The type of PII that was detected.
+            matches: List of PII matches found.
+        """
+        self.pii_type = pii_type
+        self.matches = matches
+        count = len(matches)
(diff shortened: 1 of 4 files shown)


libs/langchain_v1/langchain/agents/middleware/pii.py
```diff
@@ -0,0 +1,753 @@
...
+    Args:
+        content: Text content to scan.
+
+    Returns:
+        List of detected SSN matches.
+    """
+    matches = []
+
+    # Format with dashes: 123-45-6789
+    pattern_dashed = r"\b\d{3}-\d{2}-\d{4}\b"
```
eyurtsev: * This will be brittle due to `-` not always being present
* Would prefer to remove also as its too us centric and there are other types of IDs.

libs/langchain_v1/langchain/agents/middleware/pii.py
```diff
@@ -0,0 +1,753 @@
...
+        ```
+    """
+
+    def __init__(
+        self,
+        pii_type: Literal["email", "ssn", "credit_card", "ip", "mac_address", "url"] | str,
+        *,
+        strategy: Literal["block", "redact", "mask", "hash"] = "redact",
+        detector: Callable[[str], list[PIIMatch]] | str | None = None,
+        apply_to_input: bool = True,
```
eyurtsev, line 482: Great! separation of input and output

libs/langchain_v1/langchain/agents/middleware/pii.py
nfcampos: Removed!

libs/langchain_v1/langchain/agents/middleware/pii.py
```diff
@@ -0,0 +1,696 @@
...
+        - ``mask``: Partially mask PII (e.g., ``****-****-****-1234`` for credit card)
+        - ``hash``: Replace PII with deterministic hash (e.g., ``<email_hash:a1b2c3d4>``)
+
+    Strategy Selection Guide:
+
+        ========  ===================  =======================================
+        Strategy  Preserves Identity?  Best For
+        ========  ===================  =======================================
+        `block`   N/A                  Avoid PII completely
+        `redact`  No                   General compliance, log sanitization
```
eyurtsev, line 438: @nfcampos I'll need to document that this is not enough for PII compliance likely as users will also need to make sure that they can sanitize anything associated with their own logging set up (kidna obvious, but maybe not to everyone) + langgraph checkpointer?

eyurtsev (approved):
Looks great!

sydney-runkle (commented):
Pretty different from the PII detection we were porting from JS, though seems much more functional + I like the addition of input and output stuff.

I thought we wanted to avoid having builtin patterns due to likelihood of changes to patterns, etc?

> the built-in types were chosen because they are common, and detection can be reliably implemented with stdlib

I do think these are pretty common cases. We might want other ones like:

* SSN
* Phone numbers

eyurtsev:
> I thought we wanted to avoid having builtin patterns due to likelihood of changes to patterns, etc?

Yeah my main concern is about localization specific patterns (phone number and to some extent SSN falls into this category -- so I actually would prefer not to add them)

libs/langchain_v1/langchain/agents/middleware/pii.py
nfcampos: Yes you should say this middleware exists to avoid sending PII to LLM, not to avoid having it in your own infra, which obvs can't be done from inside this agent

PR #33273, opened 2025-10-04
fix(langchain_v1): dynamic response format

eyurtsev:
* Preserve Auto type for the response format. cc @sydney-runkle Creating an extra type was the nicest devx I could find for this (makes it easy to do isinstance(thingy, AutoStrategy)

Remaining issue to address:
* Going to sort out why we're including tools in the tool node


eyurtsev:
@sydney-runkle I'm going to merge this one, but we can update the API afterwards if you find a nicer devx

PR #33297, opened 2025-10-06
chore(langchain_v1): remove text splitters from langchain v1 namespace

eyurtsev:
Removing text splitters for now for a lighter dependency. We may re-introduce


sydney-runkle (approved):
we should update lockfiles accordingly, LGTM otherwise

PR #33353, opened 2025-10-08
feat(langchain_v1): add on_tool_call decorator

eyurtsev:
Add on tool call decorator


eyurtsev:
Closed in favor of wrap_tool_call decorator (just name change)

PR #33357, opened 2025-10-08
feat(langchain_v1): tool emulator

eyurtsev:
This is tool emulation middleware. The idea is to help test out an agent that may have some tools that either take a long time to run or are expensive to set up. This could allow simulating the behavior a bit.


libs/langchain_v1/langchain/agents/middleware/__init__.py
```diff
@@ -37,6 +38,7 @@
     "ClearToolUsesEdit",
     "ContextEditingMiddleware",
     "HumanInTheLoopMiddleware",
+    "LLMToolEmulator",
```
eyurtsev:
```suggestion
```

libs/langchain_v1/langchain/agents/middleware/__init__.py
```diff
@@ -37,6 +38,7 @@
...
     "InterruptOnConfig",
```
eyurtsev, line 41:
```suggestion
    "InterruptOnConfig",
    "LLMToolEmulator",
```

PR #33364, opened 2025-10-09
chore(langchain_v1): update on_tool_call to regular callbacks

eyurtsev:
Refactor tool call middleware from generator-based to handler-based pattern

Simplifies on_tool_call middleware by replacing the complex generator protocol with a straightforward handler pattern. Instead of yielding requests and receiving results via .send(),
handlers now receive an execute callable that can be invoked multiple times for retry logic.

Before vs. After

Before (Generator):
```python
    def on_tool_call(self, request, state, runtime):
        for attempt in range(3):
            response = yield request  # Yield request, receive result via .send()
            if is_valid(response) or attempt == 2:
                return  # Final result is last value sent to generator
```

After (Handler):

```python
    def on_tool_call(self, request, handler):
        for attempt in range(3):
            result = handler(request)  # Direct function call
            if is_valid(result):
                return result
        return result
```

Follow up after this PR:

* Rename the interceptor to wrap_tool_call
* Fix the async path for the ToolNode


libs/langchain_v1/langchain/tools/tool_node.py
```diff
@@ -916,107 +873,44 @@ async def _arun_one(  # noqa: PLR0912
...
         try:
-            yielded = next(gen)
-        except StopIteration:
-            # Handler ended immediately without yielding
-            msg = (
-                "on_tool_call handler must yield at least once. "
-                "The final result is the last value sent to the handler."
+            result = self._on_tool_call(tool_request, execute)  # type: ignore[arg-type]
+            # If result is a coroutine, await it (though handler should be sync)
+            return await result if hasattr(result, "__await__") else result
```
eyurtsev, line 901: We'll need to update this

PR #33365, opened 2025-10-09
chore(langchain_v1): add runtime and context to model request

eyurtsev:
Add runtime and context to ModelRequest to make the API more convenient


eyurtsev:
cc @sydney-runkle I applied a bunch of type ignores in the types.py file right now, will need your help at some point to undo those. Going to continue with the refactor for now

PR #33371, opened 2025-10-09
style: remove more `Optional` syntax

mdrxy:


libs/core/langchain_core/callbacks/base.py
```diff
@@ -255,8 +255,8 @@ def on_llm_start(
             prompts (list[str]): The prompts.
             run_id (UUID): The run ID. This is the ID of the current run.
             parent_run_id (UUID): The parent run ID. This is the ID of the parent run.
-            tags (Optional[list[str]]): The tags.
-            metadata (Optional[dict[str, Any]]): The metadata.
+            tags (list[str] | None): The tags.
+            metadata (dict[str, Any] | None): The metadata.
```
cbornet, line 259: Since there are type annotations on the method, having them in the docstring is redundant and error prone.
And leads to additional refactoring work like here...
I think it would be even better to remove them from the docstring altogether ?

mdrxy: yeah that's the goal - remove all types from docstrings

wasn't too consistent during this pass, but it's the longterm goal!

cbornet: I opened https://github.com/langchain-ai/langchain/pull/33388 for langchain-core

PR #33377, opened 2025-10-09
chore(langchain): delete `CohereRerank`

mdrxy:


ccurme (commented):
We elected to keep this in langchain-classic.

PR #33388, opened 2025-10-09
chore(core): remove arg types from docstrings

cbornet:
* Remove types args
* Remove types from Returns
* Remove types from Yield
* Replace `kwargs` by `**kwargs` when needed


mdrxy (approved):
LGTM. We can also remove `Defaults to `None`.` or `Defaults to x` in all cases where the default is defined in the signature, but that can come in a future PR.

PR #33398, opened 2025-10-09
chore(langchain_v1): update handler signature for wrap tool call

eyurtsev:
We'll have the handler accept only the inner ToolCall instead of the full request


eyurtsev:
After discussion keeping current interface. This version seems to be a bit more awkward

PR #33406, opened 2025-10-09
chore(text-splitters): remove arg types from docstrings

cbornet:


libs/text-splitters/langchain_text_splitters/html.py
```diff
@@ -532,31 +532,34 @@ class HTMLSemanticPreservingSplitter(BaseDocumentTransformer):
...
 
     Example:
-        ```python
-        from langchain_text_splitters.html import HTMLSemanticPreservingSplitter
 
-        def custom_iframe_extractor(iframe_tag):
-            ```
-            Custom handler function to extract the 'src' attribute from an <iframe> tag.
-            Converts the iframe to a Markdown-like link: [iframe:<src>](src).
+        .. code-block:: python
```
mdrxy: We're moving to markdown-style code blocks instead of the Sphinx `.. code-block::`

mdrxy (changes requested):

libs/text-splitters/langchain_text_splitters/html.py
cbornet: Rolled-back.

cbornet: > We're moving to markdown-style code blocks instead of the Sphinx .. code-block::

One of the bad aspect is that I think PyCharm doesn't support markdown docstrings 😞 ...

cbornet: Also I think reST is still the official markdown for Python : https://peps.python.org/pep-0287/

cbornet: > One of the bad aspect is that I think PyCharm doesn't support markdown docstrings 😞 ...

... but works fine with VSCode

mdrxy: Unfortunate. Maybe a plugin for PyCharm would support it?

For better or worse, we're fairly optimized to VSCode due to its widespread usage.

mdrxy (approved):

PR #33420, opened 2025-10-10
feat(langchain_v1): add async implementation for wrap_tool_call

eyurtsev:
Add async implementation. No automatic delegation to sync at the moment.


eyurtsev:
cc async implementation for wrap_tool_call

PR #33432, opened 2025-10-10
chore(core): Added a warning in case of nested variables

keenborder786:
- **Description:** Simple Warning in case nested variables are passed to `mustache_template_vars` to inform the user since we are only taking the top-level variable
- **Issue:** #31976


keenborder786:
@mdrxy @ccurme

eyurtsev:
@keenborder786 This does not seem like a good thing to do right? A user being confused about how the API behaves doesn't mean that we should be emitting run time warnings

keenborder786:
@eyurtsev then we need to update the documentation to inform the user since it is not clear

PR #33434, opened 2025-10-10
feat(core): Added 2 new Stringify functions with unit tests

Samama-Intellixcore:
Description: Adds two useful string utility functions to langchain_core.utils.strings:

truncate_with_ellipsis - Truncates text to max length with customizable ellipsis
normalize_whitespace - Normalizes multiple spaces/tabs to single spaces, optionally preserving newlines

Both functions include units tests and are exported in the public API. These utilities will be valuable for document preprocessing, text cleaning, and data formatting throughout the LangChain ecosystem.


eyurtsev (changes requested):
Hi @Samama-Intellixcore thank you for the contribution, but there's no reason for this library to ship the helper functions

PR #33448, opened 2025-10-12
fix(langchain): handle unrecognized FinishReason enum gracefully

bcmaymonegalvao:
### **Description**

This **PR** fixes an unhandled exception that occurs when the `Gemini API` returns a `FinishReason` enum value that is not recognized by the current version of the `google.generativeai` library.

Previously, this caused the `ChatGoogleGenerativeAI.with_structured_output()` method to raise:

```python
AttributeError: 'int' object has no attribute 'name'
```

This fix adds a safeguard that gracefully handles unknown enum values by defaulting to a string representation, rather than raising an exception. As a result, structured output parsing remains stable even when the Gemini API introduces new or experimental FinishReason codes.

A dedicated unit test has been added for `ChatGoogleGemini` to ensure graceful handling of unrecognized FinishReason enum values. The test uses a fully mocked `Gemini API` response to simulate an unknown finish reason **(12)** and verifies that the model safely returns `"UNKNOWN(12)"` without raising an exception.

### **Rationale**

The Gemini API recently started returning a new `FinishReason`  code **(12)** that is not yet covered in the existing enum definitions. This change prevents downstream crashes by:

- Handling unknown FinishReason codes defensively.
- Maintaining backwards compatibility with all existing models and integrations.
- Allowing users to inspect the raw finish reason safely if needed.

### **How to Reproduce (Before Fix)**

```python
from typing import Literal

class RouteDecision(BaseModel):
    intent: Literal["action", "conversational"]
    reasoning: str

llm = ChatGoogleGenerativeAI(model="gemini-1.5-flash", temperature=0)
structured = llm.with_structured_output(RouteDecision)
structured.invoke({"user_query": "What is the weather in San Francisco?"})
```

#### **Result (Before):**
```python
AttributeError: 'int' object has no attribute 'name'
```

#### **Result (After):**
```python
RouteDecision(intent='conversational', reasoning='...')  # handled gracefully
```

### **Changes**

**1.** Added defensive handling for unknown `FinishReason` values in `google_gemini.py`. 
**2.** Added `fix-handle-unknown-finish-reason.rst` changelog entry.

### **Tested with**

- `langchain-core==0.3.79`
- `langchain-google-genai==2.0.10`
- `Python 3.13.0 (Windows 11)`

### **Notes**

- This patch ensures stability across evolving `Gemini API` updates.
- No breaking changes were introduced.
- One unrelated test failed in `libs/core` (`test_runnable_lambda_context_config`).
This is a known intermittent issue with the FakeTracer callback when run under pytest-xdist (see langchain-ai/langchain#31284 and langchain-ai/langchain#32077), and is not related to this change. All community and partner tests pass successfully.

### **Fixes:**
 _Unrecognized FinishReason enum value: 12 from Gemini API causes crash in ChatGoogleGenerativeAI_ [#33444](https://github.com/langchain-ai/langchain-google/issues/1268)

### DIFF
--- a/.github/scripts/check_diff.py
+++ b/.github/scripts/check_diff.py
@@ -33,23 +33,13 @@ LANGCHAIN_DIRS = [
 ]
 
 # When set to True, we are ignoring core dependents
-# in order to be able to get CI to pass for each individual
-# package that depends on core
-# e.g. if you touch core, we don't then add textsplitters/etc to CI
 IGNORE_CORE_DEPENDENTS = False
 
-# ignored partners are removed from dependents
-# but still run if directly edited
+# ignored partners are removed from dependents, but still run if directly edited
 IGNORED_PARTNERS = [
-    # remove huggingface from dependents because of CI instability
-    # specifically in huggingface jobs
-    # https://github.com/langchain-ai/langchain/issues/25558
     "huggingface",
-    # prompty exhibiting issues with numpy for Python 3.13
-    # https://github.com/langchain-ai/langchain/actions/runs/12651104685/job/35251034969?pr=29065
     "prompty",
 ]
-
 
 def all_package_dirs() -> Set[str]:
     return {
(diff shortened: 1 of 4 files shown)


ccurme (commented):
ChatGoogleGenerativeAI is maintained in https://github.com/langchain-ai/langchain-google.

PR #33460, opened 2025-10-13
chore(langchain): fix CI

ccurme:


ccurme:
numexpr numpy 1.26 compatibility was temporarily broken, fixed in numexpr==2.14.1

PR #33461, opened 2025-10-13
fix(core): support for Python 3.14

cbornet:
* Fix detection of support of context in `asyncio.create_task`
* Fix: in Python 3.14 `asyncio.get_event_loop()` raises an exception if there's no running loop
* Bump pydantic to version 2.12
* Skips tests with pydantic v1 models as they are not supported with Python 3.14
* Run core tests with Python 3.14 in CI.


libs/core/langchain_core/runnables/utils.py
```diff
@@ -141,7 +142,7 @@ def coro_with_context(
     Returns:
         The coroutine with the context.
     """
-    if asyncio_accepts_context():
+    if sys.version_info >= (3, 11):
```
cbornet: The signature we get in py314 is `asyncio.create_task(coro, **kwargs)` so we can't check the presence of `context`.
There's very little chance that support for context in `create_task` will ever be removed so I think we can just do a version check here.

libs/core/langchain_core/tracers/log_stream.py
```diff
@@ -264,7 +264,10 @@ def __init__(
         self.exclude_types = exclude_types
         self.exclude_tags = exclude_tags
 
-        loop = asyncio.get_event_loop()
+        try:
+            loop = asyncio.get_event_loop()
```
cbornet, line 268: `asyncio.get_event_loop` raises an exception in Python  3.14 if there's no running loop (was a warning in previous versions)

libs/core/langchain_core/runnables/utils.py
eyurtsev: Is it possible to fix the implementation of the function: asyncio_accepts_context instead of not using it?

libs/core/tests/unit_tests/output_parsers/test_pydantic_parser.py
```diff
@@ -22,15 +23,23 @@ class ForecastV2(pydantic.BaseModel):
...
+_FORECAST_MODELS_TYPES = type[ForecastV2]
+
+if sys.version_info < (3, 14):
+
+    class ForecastV1(V1BaseModel):
+        temperature: int
+        f_or_c: Literal["F", "C"]
+        forecast: str
+
+    _FORECAST_MODELS_TYPES |= type[ForecastV1]
```
eyurtsev: `_FORECAST_MODELS_TYPES` -- which type is this? something looks off. If it's meant to be a set or a list could we use the dedicated APIs for adding an element instead of relying on the operator shorthand (`|=`) to maximize for readability?

libs/core/tests/unit_tests/utils/test_pydantic.py
```diff
@@ -139,6 +141,9 @@ class Foo(BaseModel):
     assert fields == {"x": Foo.model_fields["x"]}
 
 
+@pytest.mark.skipif(
+    sys.version_info >= (3, 14), reason="Pydantic v1 not supported with Python 3.14+"
```
eyurtsev: Nit: could be helpful to paraphrase the reason a bit (wording is confusing w/ respect to whether this is pydantic 1 or pydantic 2's v1 namespace)


You could say `The pydantic.v1 namespace is not supported with python 3.14+` or something like that (to make it clear that it's specifically v1 within pydantic 2)

eyurtsev (commented):
Looks very reasonable to me left a few nit comments

libs/core/tests/unit_tests/utils/test_pydantic.py
cbornet: Sure. Good idea!

libs/core/tests/unit_tests/output_parsers/test_pydantic_parser.py
cbornet: `_FORECAST_MODELS_TYPES` is a `typing.Union` of types. Used for type checking. I don’t think there is an API to do this. But there’s maybe a simpler way to write it. I’ll have a look.

libs/core/langchain_core/runnables/utils.py
cbornet: Sure.

cbornet: done.

libs/core/tests/unit_tests/output_parsers/test_pydantic_parser.py
cbornet: I tried to simplify. PTAL.

libs/core/tests/unit_tests/utils/test_pydantic.py
cbornet: done

cbornet:
@eyurtsev in the end, I added the upgrade to pydantic 2.12 to this PR.
This way it can be validated in CI with Python 3.14.

eyurtsev (approved):

mdrxy (approved):

sydney-runkle (approved):

Issue #30200, opened 2025-03-10
chunk_overlap Not Working When Using RecursiveCharacterTextSplitter


sallahbaksh:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
def __init__(self, separators: List[str], chunk_size: int = 4000, chunk_overlap: int = 200):
        headers_to_split_on = [("#", "Header 1"), ("##", "Header 2"), ("###", "Header 3"), ("####", "Header 4")]
        self.separators = separators
        self.chunk_size = chunk_size
        self.chunk_overlap = chunk_overlap
        # Use LangChain's MarkdownHeaderTextSplitter to split on headers (levels 1-4)
        self.header_splitter = MarkdownHeaderTextSplitter(headers_to_split_on=headers_to_split_on, strip_headers=False)
        self.text_splitter = RecursiveCharacterTextSplitter(
            separators=self.separators,
            chunk_size=self.chunk_size,
            chunk_overlap=self.chunk_overlap,
            length_function=len
        )
```

### Description

I'm trying to split text from an md file. I first, use the MarkdownHeaderTextSplitter to split on headers and then I use the RecursiveCharacterTextSplitter. I want to have a chunk overlap of 200, but when I specify the overlap and use split_text, there is no overlap actually occurring.

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.26100
> Python Version:  3.12.8 | packaged by conda-forge | (main, Dec  5 2024, 14:06:27) [MSC v.1942 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 0.3.23
> langchain: 0.3.10
> langsmith: 0.1.142
> langchain_openai: 0.2.8
> langchain_text_splitters: 0.3.2

Optional packages not installed
-------------------------------
> langserve

Other Dependencies
------------------
> aiohttp: 3.10.11
> async-timeout: Installed. No version info available.
> httpx: 0.28.1
> jsonpatch: 1.33
> numpy: 1.26.4
> openai: 1.59.5
> orjson: 3.10.11
> packaging: 24.2
> pydantic: 2.10.6
> PyYAML: 6.0.2
> requests: 2.32.3
[...]

PR #33462, opened 2025-10-13
docs(text-splitters): troubleshooting when chunk_overlap seems not to apply

sachinML:
Refs #30200 

- Use `split_documents(docs)` after header-based splitting (preserves per-section metadata; overlap applied per document).
- Overlap appears only when a single section exceeds `chunk_size`.
- Overlap does not cross section/document boundaries.
- Consider `strip_headers=True` to avoid a tiny header-only chunk; keep `""` as a fallback separator if text lacks newlines/spaces.

### DIFF
--- a/libs/text-splitters/README.md
+++ b/libs/text-splitters/README.md
@@ -21,6 +21,17 @@ LangChain Text Splitters contains utilities for splitting into chunks a wide var
 
 For full documentation, see the [API reference](https://reference.langchain.com/python/langchain).
 
+## 🛠️ Troubleshooting: `chunk_overlap` seems not to apply
+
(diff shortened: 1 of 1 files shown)


sachinML:
CI is green; CodeQL is pending. This is a docs-only change. Could a maintainer please trigger the CodeQL scan so this can be merged?

ccurme (commented):
Thanks for this. I think this is not the right level documentation for this information, can we add to the dedicated docs site [here](https://docs.langchain.com/oss/python/integrations/splitters/markdown_header_metadata_splitter)?

You can submit a PR updating the source [here](https://github.com/langchain-ai/docs/blob/main/src/oss/python/integrations/splitters/markdown_header_metadata_splitter.mdx).

sachinML:
Thanks @ccurme, I’ve opened a PR to the docs site, updating the MarkdownHeaderTextSplitter page with the troubleshooting note:
https://github.com/langchain-ai/docs/pull/1061.

If preferred, I can close or revert the README change here and keep the guidance centralized in the docs site.

PR #33465, opened 2025-10-13
feat(langchain_v1): add override to model request and tool call request

eyurtsev:
Add override to model request and tool call request


sydney-runkle (approved):
This implementation looks good. I think there are 2 confusing things here that we should be sure to document well.

1. It's a bit unintuitive that state + runtime can't be overridden, but all of the other fields can on. Should we make more of an effort to make these readonly?
2. It's a bit unintuitive that tool call is nested, but model call stuff kwargs are flat. Definitely more intuitive to patch when things are flat. Should we flatten tool call kwargs?

Issue #27260, opened 2024-10-10
KeyError '*' in parse_result when using `with_structured_output` on ChatAnthropic


srtab:
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.
- [X] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_anthropic import ChatAnthropic
from pydantic import BaseModel, ConfigDict, Field
from langchain_core.messages import SystemMessage, HumanMessage

class RequestAssessmentResponse(BaseModel):
    """
    Respond to the reviewer's feedback with an assessment of the requested changes.
    """

    model_config = ConfigDict(title="request_assessment")

    request_for_changes: bool = Field(description="Set to True if the reviewer requested changes; otherwise, False.")
    justification: str = Field(description="Justify why you think it's a change request.")


model = ChatAnthropic(model="claude-3-5-sonnet-20240620").with_structured_output(RequestAssessmentResponse, method="json_schema")
model.invoke([SystemMessage("Your goal is to determine whether the comment is a direct request for changes or not"), HumanMessage("Change the name of the file foo.py.")])
```

### Error Message and Stack Trace (if applicable)

```
KeyError                                  Traceback (most recent call last)
Cell In[1], line 17
     13     justification: str = Field(description="Justify why you think it's a change request.")
     16 model = ChatAnthropic(model="claude-3-5-sonnet-20240620").with_structured_output(RequestAssessmentResponse, method="json_schema")
---> 17 model.invoke([SystemMessage("Your goal is to determine whether the comment is a direct request for changes or not"), HumanMessage("Change the name of the file foo.py.")])

File ~/.venv/lib/python3.12/site-packages/langchain_core/runnables/base.py:3024, in RunnableSequence.invoke(self, input, config, **kwargs)
   3022             input = context.run(step.invoke, input, config, **kwargs)
   3023         else:
-> 3024             input = context.run(step.invoke, input, config)
   3025 # finish the root run
   3026 except BaseException as e:

File ~/.venv/lib/python3.12/site-packages/langchain_core/output_parsers/base.py:193, in BaseOutputParser.invoke(self, input, config, **kwargs)
    186 def invoke(
    187     self,
    188     input: Union[str, BaseMessage],
[...]

PR #33479, opened 2025-10-14
fix(core): Fix tool name check in name_dict for PydanticToolsParser

keenborder786:
- **Description:** The root cause of this issue is that when a user defines `model_config` in a `BaseModel`, the `{"type": <tool_name>}` value is derived from the title specified in `model_config` when the results are parsed [here](https://vscode.dev/github/keenborder786/langchain/blob/fix/tool_name_dict/libs/core/langchain_core/output_parsers/openai_tools.py#L199). However, [tool.__name__](https://vscode.dev/github/keenborder786/langchain/blob/fix/tool_name_dict/libs/core/langchain_core/output_parsers/openai_tools.py#L331) uses the class name (in uppercase) of the `BaseModel`, resulting in a `KeyError` when a custom title is provided in `model_config`.
 

 The Best Solution will be to use the title provided in `model_config` attribute if provided one since that is what `type` will be parsed to, if not then use `tool.__name__`. But need to make sure that this works only for Pydantic V2.

  - **Issue:** #27260


keenborder786:
@eyurtsev

eyurtsev (commented):
@keenborder786 awesome! could you add a unit test?

keenborder786:
@eyurtsev Added Unit Tests

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev is there any change you would like me to do or is it looking good now?

keenborder786:
@eyurtsev

keenborder786:
@eyurtsev this is ready to be merged, or there is something else I need to change?

PR #33491, opened 2025-10-15
feat(mistralai): remove tenacity retries for embeddings

noeliecarbo:
MistralAI (and embeddings in particular) is the only package that uses tenacity retries.

This PR removes them, offering the following benefits:
- consistency with other packages
- users can more easily customize retries according to their needs.


eyurtsev (dismissed):
@noeliecarbo thanks for the contribution!

This is a breaking change, but it's not clear to me what users gain from it. What kind of customization would they want to do that's meaningful in practice from what they can already do?

noeliecarbo:
> @noeliecarbo thanks for the contribution!
> 
> This is a breaking change, but it's not clear to me what users gain from it. What kind of customization would they want to do that's meaningful in practice from what they can already do?

@eyurtsev Thanks for the feedback! From my perspective, I see the following benefits:
- For my use case, I need to catch specific HTTP 400 errors (too many inputs). It seems that the official MAX_TOKENS of the Mistral API is inaccurate and varies somewhat inexplicably.
- Users may need different retry behaviors than those provided by `retry_if_exception_type` and `wait_fixed` (tenacity provides many more options).

eyurtsev:
what happens if you set max_retries = 0?

noeliecarbo:
@eyurtsev if I set max_retries=0, then no retries are performed. However, a tenacity.RetryError is still returned, which makes it more difficult to catch specific errors and debug the code in general

ccurme (commented):
Should we enable users to opt-out of the retries for now? e.g., make `max_retries` and `wait_time` `int | None` with existing int defaults.

noeliecarbo:
I like the idea and added a commit

ccurme (approved):

Issue #30703, opened 2025-04-07
`on_llm_new_token` gets list of dicts instead of expected string


maver1ck:
(Rewritten by @mdrxy)

TL;DR: more and more models have list content in responses. We need to consider how to update callbacks to handle this.

Repro:

```python
from langchain_core.callbacks import StdOutCallbackHandler
from langchain_anthropic import ChatAnthropic
from langchain_core.tools import tool


class TokenPrinter(StdOutCallbackHandler):
    """Custom callback handler to inspect token types during streaming."""

    def on_llm_new_token(self, token: str, **kwargs) -> None:
        print(f"TOKEN: {token}, type: {type(token)}", flush=True)


@tool
def add_numbers(a: int, b: int) -> int:
    """Add two numbers together.

    Args:
        a: First number to add.
        b: Second number to add.
    """
    return a + b


llm = ChatAnthropic(model="claude-3-7-sonnet-20250219")
llm_with_tools = llm.bind_tools([add_numbers])

# Test with streaming and the custom callback
print("--- Streaming with tool binding ---")
for chunk in llm_with_tools.stream(
    "What is 2 + 3?",
    config={"callbacks": [TokenPrinter()]}
):
    pass

print("\n--- Done ---")
```

antonyscerri:
Any progress with this issue? It impacts AWS Bedrock converse clients also. Anywhere that model responses are including a list in the content field and not a simple string can cause this. Even though on_llm_new_token is declared as accepting a string the providers are simply passing msg.content which can be the lists as described above, rather than maybe using msg.text().

ccurme:
I think it's possible that the type itself is wrong, coercing to string will likely have a substantial impact which needs to be understood and scoped.

It would be helpful to hear broader context— if the type were expanded to [str | list[str | dict]](https://github.com/langchain-ai/langchain/blob/9f470d297f0177b562c42a8a2f60a6f36ea48324/libs/core/langchain_core/messages/base.py#L98), would this resolve issues on your end?

PR #33498, opened 2025-10-15
fix(core): `on_llm_new_token` correct casting for .content to string

keenborder786:
- **Description:** The `on_llm_new_token` function should only accept values of type `string`. However, the `.content` attribute can sometimes be a list, so it’s unclear why it wasn’t being cast to a string. This PR addresses that by adding the necessary type casting.
- **Issue:** #30703


keenborder786:
@eyurtsev

libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -531,9 +531,13 @@ def stream(
...
                             if "index" not in block:
                                 block["index"] = index
-                    run_manager.on_llm_new_token(
-                        cast("str", chunk.message.content), chunk=chunk
-                    )
+
+                    if isinstance(chunk.message.content, list):
+                        run_manager.on_llm_new_token(
+                            json.dumps(chunk.message.content), chunk=chunk
+                        )
```
ccurme, line 538: do we want to json.dumps the content or just call `chunk.message.text` as the original issue proposed?

my bias is to do `chunk.message.text`, which will represent tokens of text content. it will miss tokens from reasoning, tool calls, and other content, but I don't think `on_llm_new_token` was designed to support these well. so my thought is to go with well-defined behavior for now.

eyurtsev:
This requires very careful design review. It's possible that the type is wrong and the callback handler actually needs to support all content not just text. We need to check how things are wired

keenborder786:
@eyurtsev right now it only supports text, do you mean we need to update `on_llm_new_token ` itself to support all type of objects?

ccurme:
Yeah, I think @eyurtsev's point was that given how often this is used it's likely that coercing to string will break things, and that the original type is actually not accurate. But we'd need to scope the change. Closing for now to continue discussion on the original issue.

PR #33523, opened 2025-10-16
refactor: replace deprecated type hints from List to built-in modern list for consistency

Srijan-D:
Description: This change updates type hints across multiple files to use the built-in `list` type instead of `List` from the `typing` module, aligning with modern Python and replacing the deprecated imports.

### DIFF
--- a/.github/scripts/check_diff.py
+++ b/.github/scripts/check_diff.py
@@ -19,7 +19,7 @@ import os
 import sys
 from collections import defaultdict
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


eyurtsev (changes requested):
Thanks! @Srijan-D we generally leave refactors like this for core maintainers. They're very fast to do with automatic linters.

Contributions that would be better would be bug fixes, issues, questions about docs, fixes the main docs etc

Srijan-D:
ah cool!

PR #33532, opened 2025-10-16
feat(openai): add callable support for openai_api_key parameter

marlenezw:
# Description

This PR adds support for passing a callable function as the `openai_api_key` parameter in `ChatOpenAI`, `OpenAI`, and `OpenAIEmbeddings` classes. This enhancement enables dynamic API key retrieval at runtime, which is needed for Azure OpenAI's new v1 API with Microsoft Entra ID (formerly Azure AD) token-based authentication.

## Context: Azure OpenAI v1 API Evolution
As outlined in this [Azure OpenAI API version lifecycle documentation](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle?tabs=python#next-generation-api-1), Azure OpenAI introduced a v1 API (Generally Available as of August 2025) that brings several improvements. The most relevant here:

- **Unified OpenAI client**: Use the standard `OpenAI()` client instead of Azure-specific clients
- **Native token-based authentication**: The v1 API adds automatic token refresh support directly to the `OpenAI()` client, removing the dependency on separate Azure clients. 

## Key Use Case: Microsoft Entra ID Authentication
The v1 API documentation specifically shows that Microsoft Entra ID authentication now works by passing a token provider callable directly to the api_key parameter:

```python 
from openai import OpenAI
from azure.identity import DefaultAzureCredential, get_bearer_token_provider

token_provider = get_bearer_token_provider(
    DefaultAzureCredential(), "https://cognitiveservices.azure.com/.default"
)

client = OpenAI(  
    base_url="https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1/",  
    api_key=token_provider  # Callable that handles automatic token refresh
)
```

This PR enables LangChain to support this pattern. Previously, LangChain only accepted static strings or `[SecretStr]` parameter, which prevented users from being able to use an Azure credential. 

## Changes Made

Updated the code in the OpenAI Base, LLMs and Emeddings files to contain this code: 

```python
api_key_value = None
if self.openai_api_key is not None:
    if isinstance(self.openai_api_key, SecretStr):
        api_key_value = self.openai_api_key.get_secret_value()
    elif callable(self.openai_api_key):
        api_key_value = self.openai_api_key()
```

I also added an environmental check to see if the user is using a key from Azure OpenAI instead of one from OpenAI 

```python
openai_api_key: SecretStr | None | Callable[[], str] = Field(
    alias="api_key",
    default_factory=secret_from_env(
        ["OPENAI_API_KEY", "AZURE_OPENAI_API_KEY"], 
        default=None
    ),
)
```

Added tests to the unit tests secret testing file. 

Enhancement to support Azure OpenAI v1 API authentication patterns

Dependencies
None - uses existing dependencies


marlenezw:
Tests passing! cc @ccurme and @mdrxy. This PR addresses the issue we discussed in Slack to add in support for a callable in OpenAI for the new v1 API. Note this is the same in OpenAI client code here https://github.com/openai/openai-python/blob/ebf32212f7bf5bec6b24cc2276ac0d9a28dd63bb/src/openai/_client.py#L102

libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -465,8 +449,11 @@ class BaseChatOpenAI(BaseChatModel):
     """What sampling temperature to use."""
     model_kwargs: dict[str, Any] = Field(default_factory=dict)
     """Holds any model parameters valid for `create` call not explicitly specified."""
-    openai_api_key: SecretStr | None = Field(
-        alias="api_key", default_factory=secret_from_env("OPENAI_API_KEY", default=None)
+    openai_api_key: SecretStr | None | Callable[[], str] = Field(
+        alias="api_key",
+        default_factory=secret_from_env(
+            ["OPENAI_API_KEY", "AZURE_OPENAI_API_KEY"], default=None
```
ccurme: Can we implement this separately on AzureChatOpenAI? I don't think we want ChatOpenAI to start looking for API keys in AZURE_OPENAI_API_KEY.

ccurme (commented):
Thank you for this!

libs/partners/openai/langchain_openai/chat_models/base.py
marlenezw: Sounds good! Let me remove that!

marlenezw: resolved @ccurme if you want to take another look!

marlenezw:
Hi @ccurme, would be awesome to get this merged before the v1 announcement if possible! We'd like to add these updates about EntraID to the MSFT Learn docs

libs/partners/openai/langchain_openai/embeddings/base.py
```diff
@@ -292,10 +293,17 @@ def validate_environment(self) -> Self:
...
             )
             raise ValueError(msg)
+
+        # Resolve API key from SecretStr or Callable
+        api_key_value = None
+        if self.openai_api_key is not None:
+            if isinstance(self.openai_api_key, SecretStr):
+                api_key_value = self.openai_api_key.get_secret_value()
+            elif callable(self.openai_api_key):
+                api_key_value = self.openai_api_key()
```
ccurme: @marlenezw do we actually want to pass in the callable itself here (instead of calling it)

marlenezw: I think we want to pass the callable, but let me run some tests on my changes with the token to confirm!

marlenezw: Hi @ccurme This works as expected when I run the following code with the updates! 

```python
from azure.identity import DefaultAzureCredential, get_bearer_token_provider

token_provider = get_bearer_token_provider(
    DefaultAzureCredential(), "https://cognitiveservices.azure.com/.default"
)

llm = ChatOpenAI(
  model = "gpt-5-mini",
  base_url = "https://{your_resource_name}.openai.azure.com/openai/v1/",
  api_key = token_provider
)


# Example invocation
messages = [
("system", "You are a helpful assistant."),
("human", "Translate 'I love programming' to French.")
]
response = llm.invoke(messages)
print(response.content)
```
I get the response I was expecting

marlenezw: One thing i did notice is that I need to make the same changes for `AzureChatOpenAI`, `AzureOpenAI` and the Azure embeddings. Would you like me to add the changes in this PR or create a second one for these updates? Changes will be exactly the same but with the additional files.

ccurme:
I think there's a configuration issue with codspeed, same errors are coming up on an unrelated PR that just adds a unit test: https://github.com/langchain-ai/langchain/pull/33134/checks?check_run_id=53206813137

marlenezw:
No worries, feel free to add the changes when you've confirmed. I'll check back in tomorrow

ccurme:
@marlenezw I'll get this across, it increased in scope somewhat to support the [async case](https://github.com/openai/openai-python/blob/dff16b5e9964bf85157eb41181255ed5dde8dda4/src/openai/_client.py#L453) but I think it's mostly done. Will come back tomorrow.

eyurtsev (approved):

ccurme (approved):
Thanks very much @marlenezw. I've made some updates to support the async case, since the AsyncOpenAI client requires an async callable (otherwise `await model.ainvoke("...")` errors out).

Let me know if you see any issues, I'll follow-up with a release.

marlenezw:
Awesome! Thank you  @ccurme!!! I'll update the docs!

Issue #32066, opened 2025-07-16
docstrings shouldn't inherit from parents


mdrxy:
> ```python
> from langchain_core.tools import tool
> from pydantic import BaseModel
> 
> 
> class MyTool(BaseModel):
>     """Parent Tool"""
>     foo: str
> 
> @tool
> class ChildTool(MyTool):
>     bar: str
> ```
> 
> For 0.4 release we could consider updating the behavior so that doc-strings are not inherited from their parents. It's fairly odd that doc-strings are inherited for BaseModels as in general the descriptions need to be different 

 _Originally posted by @eyurtsev in [#31606](https://github.com/langchain-ai/langchain/issues/31606#issuecomment-3009548334)_

chandrani121189:
@mdrxy  If you add a docstring to the base class does it not override?

R-Fischer47:
Hi @mdrxy & @eyurtsev, I'm a new contributor and I'd like to work on this issue. Can I be assigned?

mdrxy:
@veeceey @tysoncung @mrutunjay-kinagi why did you each independently open PRs for the same issue, when there is already an open PR? Although we appreciate the initiative, this is added noise for maintainers and it would generally be better to discuss first. Closing them for now.

ayeshamallick6514-aye:
Hi, I'd like to work on this issue. Is it currently available for contribution?

harshavardhan194:
Hi! I'm interested in working on this issue if it's still available. If no one is actively implementing a fix, I'd be happy to investigate and submit a PR.

zm-michelle:
I think I found the issue. The docstring is inherited from inspect.getdoc() in tools/base.py (_infer_arg_descriptions), rather than the tool's class own .__doc__ or pydantic.   This walks the MRO and returns the parent's docstring when the object's own __doc__ is None.

I have a two-part fix if the desired behavior is keeping the class docstring empty when it's empty, rather than inheriting from the parent. I have a fix I implemented and tested on my side, where in tools/base.py a function gets the Tool's own docstring and then in tools/structured.py I change one line so a class-based tool with no docstring gets an empty description instead of raising an error. Can I get it assigned ?

tlysanhuo:
I looked into this and would like to take it if maintainers agree with the approach.

[...]

PR #33540, opened 2025-10-16
fix(core): prevent inherited docstrings for class-based tools    

R-Fischer47:
**Description**

This PR prevents class-based tools from inheriting the docstring of their parent class. When a child tool class does not have its own docstring, it was incorrectly using the parent's, leading to an inaccurate description for the tool.

This holistic fix modifies two files:
- `langchain_core/tools/base.py`: The `_infer_arg_descriptions` function now correctly identifies an inherited docstring and returns `None` to signal that it should be ignored.
- `langchain_core/tools/structured.py`: The `StructuredTool.from_function` method now correctly handles this `None` signal. It creates the tool with an empty string `""` for its description, while still correctly raising an error for regular functions that are missing a docstring.

This ensures that tools have accurate descriptions and that the developer experience is intuitive.

**Issue**

Fixes #32066

**Dependencies**

N/A


R-Fischer47:
Hi @sydney-runkle and @eyurtsev, the merge is now failing due to tests added after I suggested this PR that have already been merged. Should I work on resolving those? Or is there anything else I can do to move this forward?

libs/core/langchain_core/tools/structured.py
```diff
@@ -215,20 +216,22 @@ def add(a: int, b: int) -> int:
...
+            if inspect.isclass(source_function) and is_basemodel_subclass(
+                source_function
+            ):
+                description_ = ""
+            else:
+                msg = "Function must have a docstring if description not provided."
+                raise ValueError(msg)
+
         if description is None:
-            # Only apply if using the function's docstring
```
mdrxy, line 227: Why are we removing comments?

libs/core/langchain_core/tools/base.py
```diff
@@ -189,25 +189,26 @@ def _infer_arg_descriptions(
     *,
     parse_docstring: bool = False,
     error_on_invalid_docstring: bool = False,
-) -> tuple[str, dict]:
-    """Infer argument descriptions from function docstring and annotations.
-
-    Args:
```
mdrxy, line 195: Why are we removing sections from the docstring?

sydney-runkle:
Closing as stale

PR #33541, opened 2025-10-16
fix(exa)!: rename 'type' parameter to 'search_type' to avoid shadowing Python built-in

gjeltep:
**Description:**
Fixes a code quality issue where the `type` parameter in Exa tools was shadowing Python's reserved `type` keyword. The parameter has been renamed to `search_type`.

**Changes:**
- Renamed `type` parameter to `search_type` in `ExaSearchResults._run()` method
- Updated `ExaSearchRetriever` class to use `search_type` attribute
- Updated integration test to use the new parameter name

**Issue:**
Addresses TODO comment in `libs/partners/exa/langchain_exa/tools.py:122`

**Dependencies:**
None

**Note:**
This is a breaking change for users directly passing `type=` to these methods. They will need to update to `search_type=`.

### DIFF
--- a/libs/partners/exa/langchain_exa/retrievers.py
+++ b/libs/partners/exa/langchain_exa/retrievers.py
@@ -55,7 +55,7 @@ class ExaSearchRetriever(BaseRetriever):
     """The end date for when the document was published (in YYYY-MM-DD format)."""
     use_autoprompt: bool | None = None
     """Whether to use autoprompt for the search."""
-    type: str = "neural"
+    search_type: str = "neural"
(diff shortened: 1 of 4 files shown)


mdrxy (approved):

eyurtsev (changes requested):
This is a breaking change in an API -- not something that we can do for a library :)

eyurtsev:
The way I'd think about this:

it doesn't matter that it's shadowing python's built-in. python allows shadowing built-ins. 

we should try avoiding shadowing python builtints when we can, especially if it results in confusing code.

But we shouldn't change code that's working after the fact to appease a linter as it's going break user's actual code for no practical benefit.

PR #33544, opened 2025-10-16
fix(deepseek): inject `model_provider` in `response_metadata`

mdrxy:
& slight tests rfc


ccurme (approved):
Augmenting the existing reasoning test with a check for content_blocks would be nice.

PR #33582, opened 2025-10-17
chore(langchain-classic): remove unused duckdb dependency

cbornet:
* The dependency is not used.
* It takes a long time to build in Python 3.14 as there are no prebuilt binaries yet. This slows down CI a lot.


mdrxy (approved):
ty

PR #33595, opened 2025-10-18
[DO NOT MERGE]: investigate issue with Pydantic and Python 3.14

cbornet:
I didn't manage to reproduce locally yet.


cbornet:
I eventually managed to reproduce.
Interestingly, it's reproduced with Python 3.14.0 but not with Python 3.14.0rc2 🤔

PR #33601, opened 2025-10-20
style: more refs polishing

mdrxy:


mdrxy:
race condition in CI check

PR #33618, opened 2025-10-20
fix(core): applied secrets_map in load() to plain string values

npage902:
**Description:** Fixes the bug in the `load()` function where secret placeholders in plain dicts were not replaced, even if they match a key in `secrets_map`, and adds a test case.

Example:
```py
obj = {"api_key": "__SECRET_API_KEY__"}
secret_key = "secret_key_1234"
secrets_map = {"__SECRET_API_KEY__": secret_key}
result = load(obj, secrets_map=secrets_map)
```
Before this change, printing `api_key` in `result` would output `"__SECRET_API_KEY__"`. Now, it will properly output `"secret_key_1234"`.

**Issue:** Fixes #31804 

**Dependencies:** None

`make format`, `make lint`, and `make test` have all passed on my machine.

### DIFF
--- a/libs/core/langchain_core/load/load.py
+++ b/libs/core/langchain_core/load/load.py
@@ -265,6 +265,8 @@ def load(
             return reviver(loaded_obj)
         if isinstance(obj, list):
             return [_load(o) for o in obj]
+        if isinstance(obj, str) and obj in reviver.secrets_map:
+            return reviver.secrets_map[obj]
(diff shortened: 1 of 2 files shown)


npage902:
Things seem to be scuffed on my end in regards to makefile testing. I'll need to look into this, but I'll just close this up for now. Apologies for any inconvenience.

PR #33648, opened 2025-10-23
fix(mistralai): resolve duplicate tool calls when converting to mistral chat message

Romi45:
### Description

A bug inside the mistralai.chat_models._convert_message_to_mistral_chat_message function duplicates tools_calls when converting an AIMessage to a Mistral assistant message. The bug is caused by a for loop structure that iterates over the full list of tool_calls for every tool_call in the list. 

This means that the final Mistral assistant message will contain N*N tool_calls for an N number of tool_calls in the original AIMessage array. If the Mistral model calls one tool then the code will not raise, but any time it calls at least two tools, agent.invoke() will raise with the following error : 
```bash
Error response 400 while fetching https://api.mistral.ai/v1/chat/completions: {"object":"error","message":"Duplicate tool call id in assistant message","type":"invalid_request_message_order","param":null,"code":"3230"}
```

The fix is a simple modification of the looping logic to make sure each tool_call only gets appended once to prevent duplication. Here is the faulty part of the function **before** the fix : 
```python
tool_calls = []
if message.tool_calls or message.invalid_tool_calls:
    for tool_call in message.tool_calls:
        tool_calls.extend(
            [
                _format_tool_call_for_mistral(tool_call)
                for tool_call in message.tool_calls
            ]
        )
    for invalid_tool_call in message.invalid_tool_calls:
        tool_calls.extend(
            _format_invalid_tool_call_for_mistral(invalid_tool_call)
            for invalid_tool_call in message.invalid_tool_calls
        )
```

And here is the same part **after** the fix : 
```python
tool_calls = []
if message.tool_calls or message.invalid_tool_calls:
    if message.tool_calls:
        # Add each tool call exactly once.
        tool_calls.extend(
            _format_tool_call_for_mistral(tc) for tc in message.tool_calls
        )
    if message.invalid_tool_calls:
        # Add each invalid tool call exactly once.
        tool_calls.extend(
            _format_invalid_tool_call_for_mistral(itc)
            for itc in message.invalid_tool_calls
        )
```

I could not find any issue mentioning this specific bug on the github so no issue linked. Lit and test pass correctly. No additional dependencies.

### DIFF
--- a/libs/partners/mistralai/langchain_mistralai/chat_models.py
+++ b/libs/partners/mistralai/langchain_mistralai/chat_models.py
@@ -364,16 +364,14 @@ def _convert_message_to_mistral_chat_message(
         return {"role": "user", "content": message.content}
     if isinstance(message, AIMessage):
         message_dict: dict[str, Any] = {"role": "assistant"}
-        tool_calls = []
+        tool_calls: list = []
         if message.tool_calls or message.invalid_tool_calls:
-            for tool_call in message.tool_calls:
+            if message.tool_calls:
                 tool_calls.extend(
-                    [
-                        _format_tool_call_for_mistral(tool_call)
-                        for tool_call in message.tool_calls
-                    ]
+                    _format_tool_call_for_mistral(tool_call)
+                    for tool_call in message.tool_calls
                 )
-            for invalid_tool_call in message.invalid_tool_calls:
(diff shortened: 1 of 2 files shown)


ccurme (approved):
Great catch thanks very much for this.

ccurme:
Released in latest langchain-mistralai.

PR #33650, opened 2025-10-23
feat(langchain): add CockroachDB vector store alias using PGVector (#33642)

himelanoleuca-blip:
**Description:** This PR adds a CockroachDB vector store wrapper that inherits from the existing `PGVector` vector store to leverage CockroachDB's pgvector-compatible API. It enables LangChain users to utilize CockroachDB's vector search without additional code changes.

**Issue:** Closes #33642 by implementing the requested CockroachDB support based on the existing pgvector implementation.

**Dependencies:** No new dependencies are introduced; this wrapper reuses the `langchain_community` PGVector backend.

### DIFF
--- a/libs/langchain/langchain_classic/vectorstores/cockroachdb.py
+++ b/libs/langchain/langchain_classic/vectorstores/cockroachdb.py
@@ -1,0 +1,16 @@
+"""CockroachDB vector store for LangChain.
+
+This class uses the pgvector-compatible API provided by CockroachDB.
+It inherits from the PGVector vector store to leverage existing functionality.
+"""
+
+from __future__ import annotations
+
+from langchain_community.vectorstores.pgvector import PGVector
+
+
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Hello, thanks for this. There are some issues:
- It's unclear that this would solve any problems (are we just renaming PGVector?)
- We use the [deprecated](https://github.com/langchain-ai/langchain-community/blob/77b245d87497b4ed793b3e71fcc8c6df22ffa149/libs/community/langchain_community/vectorstores/pgvector.py#L221) version of PGVector in langchain-community instead of the implementation in langchain-postgres (docs [here](https://docs.langchain.com/oss/python/integrations/vectorstores/pgvector))
- If a new integration is needed (and it looks like it isn't if this PR is complete), it should be implemented as a separate package following our [contributing docs](https://docs.langchain.com/oss/python/contributing/integrations-langchain).

PR #33670, opened 2025-10-25
feat(perplexity): Created Dedicated Output Parser to Support Reasoning Model Output for perplexity

keenborder786:
- **Description:** Add the perplexity Output Parser to support reasoning model output to remove <think> tags which was invalidating the JSON output.
- **Issue:** #32465


keenborder786:
@mdrxy

keenborder786:
@ccurme

libs/partners/perplexity/langchain_perplexity/output_parsers.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Args:
+        text: The input text that may contain think tags.
+    """
+
+    def remove_think_tags(text: str) -> str:
+        """Remove content between <think> and </think> tags more safely."""
+        result = []
+        i = 0
+        while i < len(text):
```
ccurme: We shouldn't need to loop through the text characters here.

Perplexity provides a reference implementation [here](https://github.com/perplexityai/api-discussion/blob/main/utils/extract_json_reasoning_models.py), can we adapt this?

keenborder786: @ccurme done

ccurme (approved):

PR #33671, opened 2025-10-25
refactor(anthropic): remove unnecessary url check

arun-geek:
if "url" in annotation: in Line 15 , already ensures "url" is key in annotation , so no need to check again to set "url" key in out object


mdrxy (approved):
TY!

PR #33673, opened 2025-10-26
refactor(core): Improve performance and code simplicity

arun-geek:
1. Currently , Removes tag one by one by passing it in as list argument to remove_tags
2. Can create and pass the list at once instead of having it to create a list of just on object and passing it downstream remove_tags


ccurme (commented):
Behavior between these differs in edge cases. If this change is motivated by a material performance issue, could you open an issue describing it? Open to re-opening the PR if it's solving a problem.

PR #33674, opened 2025-10-26
refactor(core): Minor refactor for code readability

arun-geek:
1. Avoid enumerate for first object check
2.  String is immutable and concating repeatedly can have performance issues if used at scale


libs/core/langchain_core/utils/strings.py
```diff
@@ -30,10 +30,7 @@ def stringify_dict(data: dict) -> str:
     Returns:
         The stringified dictionary.
     """
-    text = ""
-    for key, value in data.items():
-        text += key + ": " + stringify_value(value) + "\n"
-    return text
+    return "\n".join(key + ": " + stringify_value(value) for key, value in data.items())
```
ccurme: this is changing behavior to drop trailing newlines, which could break caches and is arguably not desired.

libs/core/langchain_core/outputs/llm_result.py
```diff
@@ -67,9 +67,9 @@ def flatten(self) -> list[LLMResult]:
                 Generation.
         """
         llm_results = []
-        for i, gen_list in enumerate(self.generations):
+        for gen_list in self.generations:
```
ccurme: I see no compelling reason to make this change, conditioning on `i == 0` is clear and readable with no performance issues.

ccurme (approved):

PR #33698, opened 2025-10-28
fix(core): fix Pydantic v1 support in tools/runnable

cbornet:
`BaseTool.args_schema` is documented as accepting a Pydantic v1 model, but several code paths assumed v2 and raised when handed a v1 schema (e.g. an `AttributeError` from calling `model_json_schema()`/`model_fields` on a v1 model). This affected anyone using a v1 `args_schema`, and anyone composing runnables whose input/output schema is a v1 model.

This PR makes the tool/runnable schema-derivation code version-agnostic.

## Type contract

`TypeBaseModel` (and `PydanticBaseModel`) now include `pydantic.v1.BaseModel`, so the type honestly reflects what tools and runnables already accept at runtime. The public schema accessors (`Runnable.get_input_schema`/`get_output_schema` and the `input_schema`/`output_schema` properties) return `TypeBaseModel`.

## Version-agnostic helpers

Added to `langchain_core.utils.pydantic`, each dispatching on the model's Pydantic version so callers don't have to:

- `model_json_schema(model)` — JSON schema for either version.
- `model_validate(model, obj)` — validation for either version.
- `get_fields(model)` — field map for either version (existing helper, now used consistently).

Internally, direct `.model_json_schema()` / `.model_fields` calls are replaced with these helpers (or with `get_input_jsonschema()` / `get_output_jsonschema()`).

## Behavior change worth a close look

When deriving a schema from a v1 model (in `RunnableParallel`, `RunnableAssign`, and `RunnableSequence` output schemas), a **required** v1 field is now correctly carried over as required. Previously the v1 path read the field's `default` — which is `None` for a required v1 field — and silently turned required fields into optional/nullable ones; `default_factory` fields were dropped entirely. The new `_get_schema_field_definition` helper translates a v1 `ModelField` faithfully (required → `...`, factory preserved) and dispatches explicitly on the field type.


cbornet:
Hi, could this get some eyes ?
It's on the path to activate `warn_unreachable` on `core`.

PR #33710, opened 2025-10-29
refactor: use reduce and avoid slicing

arun-geek:
Slicing in python creates a new list. There is a minor performance/memory gain.
Reduce can be used to do the same without having to create a new list from slicing and iterating.


libs/partners/anthropic/langchain_anthropic/_compat.py
eyurtsev, line 25: Why is this change here?

libs/langchain_v1/langchain/agents/factory.py
```diff
@@ -158,9 +159,7 @@ def inner_handler(req: ModelRequest) -> ModelResponse:
         return composed
 
     # Compose right-to-left: outer(inner(innermost(handler)))
-    result = handlers[-1]
-    for handler in reversed(handlers[:-1]):
-        result = compose_two(handler, result)
+    result = reduce(lambda handler1, handler2: compose_two(handler2, handler1), reversed(handlers))
```
eyurtsev, line 162: @arun-geek we typically have ~1-10 handlers... have you tried benchmarking to confirm that:

1. The change has the effect you expect?
2. That it's actually a bottleneck? i.e., any improvements here may have no practical effect.

arun-geek: @eyurtsev 
its minor performance , but mostly for readability , do you think its counterproductive ?

PR #33735, opened 2025-10-29
fix(langchain_v1): removed unsed functions in tool_call_limit middleware

christian-bromann:
These functions seem unused and can be removed.


sydney-runkle (approved):
Nice catch, thanks!

PR #33745, opened 2025-10-30
fix(core): add re-entrancy guard to `RootListenersTracer` to prevent recursive listener invocations

riunyfir:
PR Message
- Description:  
  This change introduces a contextvars-based re-entrancy guard in RootListenersTracer and AsyncRootListenersTracer to prevent listener callbacks (on_start, on_end, on_error) from invoking nested listeners. When multiple custom PII middlewares are stacked, listener implementations may indirectly trigger another chain/graph run and reattach listeners, forming a loop that eventually leads to a GraphRecursionError in LangGraph. The guard blocks listener re-entry within the same execution context, breaking the unintended cycle while preserving normal listener behavior.

  Changes:
  - Add a ContextVar _IN_LISTENER with scoped set/reset around listener invocations.
  - Short-circuit listener execution if already inside a listener context.
  - Apply the same protection to both sync and async tracers.

  Rationale:
  - Prevent implicit “listener -> re-run -> listener …” loops that inflate recursion depth and trip LangGraph’s recursion_limit.
  - Provide a safe, minimal, and backwards-compatible safeguard without altering public APIs.

  Backwards compatibility:
  - Fully backwards compatible. Existing listeners that do not re-enter will behave unchanged. Listeners that attempted to re-enter will now no-op on nested calls, avoiding unbounded recursion.

  Tests/validation:
  - Manually verified with multiple stacked PII middlewares where listeners previously re-entered and caused GraphRecursionError; the behavior is now stable and non-recursive.
  - Lint passes locally.

- Issue:  #33740


mdrxy:
could you add a MRE that would fail without this change, perhaps via a unit test?

could you resolve all CI conflicts please?

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #33756, opened 2025-10-31
feat(core): automatically count and store meta for tool call count

mdrxy:
Adds automatic tool call counting to tracing by means of a new `store_tool_call_count_in_run()`, which calls on newly added `count_tool_calls_in_run()`.

Runs on successful LLM completion. Does not run on errored runs.


libs/core/langchain_core/tracers/base.py
```diff
@@ -37,13 +37,28 @@ class BaseTracer(_TracerCore, BaseCallbackHandler, ABC):
     def _persist_run(self, run: Run) -> None:
         """Persist a run."""
 
+    def _store_tool_call_metadata(self, run: Run) -> None:
+        """Store tool call count in run metadata automatically."""
+        try:
+            # Avoid circular imports
```
hinthornw: Utils doesn't import from base, so this shouldn't be needed right?

libs/core/langchain_core/tracers/base.py
```diff
@@ -37,13 +37,28 @@ class BaseTracer(_TracerCore, BaseCallbackHandler, ABC):
     def _persist_run(self, run: Run) -> None:
         """Persist a run."""
 
+    def _store_tool_call_metadata(self, run: Run) -> None:
```
hinthornw: I'd put this all in `_complete_llm_run` in tracers/core. We don't need to do this on all runs IMO

libs/core/langchain_core/tracers/utils.py
```diff
@@ -0,0 +1,91 @@
...
+def get_tool_call_count_from_run(run: Run) -> int | None:
+    """Get the tool call count from run metadata if available.
+
+    Args:
+        run: The `Run` object to check.
+
+    Returns:
+        The tool call count if stored in metadata, otherwise `None`.
+    """
+    extra = getattr(run, "extra", {}) or {}
```
hinthornw: Unnecessarily defensive right?

libs/core/langchain_core/tracers/utils.py
```diff
@@ -0,0 +1,68 @@
+"""Utility functions for working with Run objects and tracers."""
```
mdrxy:
```suggestion
"""Utility functions for working with `Run` objects and tracers."""
```

libs/core/langchain_core/tracers/core.py
```diff
@@ -283,6 +284,7 @@ def _complete_llm_run(self, response: LLMResult, run_id: UUID) -> Run:
                     )
         llm_run.end_time = datetime.now(timezone.utc)
         llm_run.events.append({"name": "end", "time": llm_run.end_time})
+        store_tool_call_count_in_run(llm_run)
```
jacoblee93: nit: I really dislike side-effects but I understand they're a thing in Python

libs/core/langchain_core/tracers/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    outputs = getattr(run, "outputs", {})
+    if isinstance(outputs, dict) and "messages" in outputs:
+        messages = outputs["messages"]
+        if messages:
+            tool_call_count += _count_tool_calls_in_messages(messages)
+
+    return tool_call_count
+
+
+def store_tool_call_count_in_run(run: Run, *, always_store: bool = False) -> int:
```
jacoblee93: Does this really need to be its own function

jacoblee93 (dismissed):

libs/core/langchain_core/tracers/core.py
```diff
@@ -284,6 +284,16 @@ def _complete_llm_run(self, response: LLMResult, run_id: UUID) -> Run:
         llm_run.end_time = datetime.now(timezone.utc)
         llm_run.events.append({"name": "end", "time": llm_run.end_time})
 
+        tool_call_count = 0
+        for generations in response.generations:
+            for generation in generations:
+                if hasattr(generation, "message"):
+                    msg = generation.message
+                    if hasattr(msg, "tool_calls"):
```
hinthornw, line 292: Could this ever be null

PR #33772, opened 2025-11-01
feat(cli): add log-level option, reload flag, and env-based defaults for serve

santosh5541:
### 🧩 Summary

This PR enhances the **LangChain CLI** to improve developer experience and flexibility.

**Key changes**
- Added global `--log-level` option (`DEBUG`, `INFO`, `WARNING`, etc.)
- Added `--reload` / `--no-reload` flag to `serve` command
- Introduced environment-based defaults:
  - `LANGCHAIN_CLI_HOST`
  - `LANGCHAIN_CLI_PORT`
- Added validation for port range
- Unified logging output for all subcommands
- Kept backward compatibility with existing `serve()` signatures (gracefully ignores `reload` if unsupported)

### DIFF
--- a/libs/cli/langchain_cli/cli.py
+++ b/libs/cli/langchain_cli/cli.py
@@ -1,11 +1,10 @@
-"""LangChain CLI."""
-
 from __future__ import annotations
 
 import logging
 import os
+from collections.abc import Callable
 from enum import Enum
-from typing import Annotated, Any, Callable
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Could you raise an issue with the larger problem you're trying to solve?

`serve` interacts with LangServe which is a legacy library— see its [readme](https://github.com/langchain-ai/langserve).

Issue #33757, opened 2025-10-31
Support cache_control for ChatOpenAI


jingyugao:
### Checked other resources

- [x] This is a feature request, not a bug report or usage question.
- [x] I added a clear and descriptive title that summarizes the feature request.
- [x] I used the GitHub search to find a similar feature request and didn't find it.
- [x] I checked the LangChain documentation and API reference to see if this feature already exists.
- [x] This is not related to the langchain-community package.

### Feature Description

Claude support Prompt-Caching feature, which can optimizes api usage and reduces processing time.
AnthropicPromptCachingMiddleware can perfectly support this feature。
However,there are some other claude model provider,such as openrouter. Which can only use ChatOpenAI client.

### Use Case

I want to add an ChatOpenAIPromptCachingMiddleware,just like AnthropicPromptCachingMiddleware.

### Proposed Solution

First, add an ChatOpenAIPromptCachingMiddleware, all code is same.
Second, modify ChatOpenAI to support "cache_control" in kwargs.
Code may be like this. I have tested in my AI agent.

```
    def _get_request_payload(
        self,
        input_: LanguageModelInput,
        *,
        stop: list[str] | None = None,
        **kwargs: Any,
    ) -> dict:
        wrap_text = lambda text :{"type":"text","text":text} if isinstance(text,str) else {**text}
        if "cache_control" in payload:
            dup_message = {** payload["messages"][-1]}
            last_message_content = dup_message["content"]
            if isinstance(last_message_content,str):
                dup_message["content"]=[wrap_text(last_message_content)]
            elif isinstance(last_message_content,list):
                dup_message["content"] = [wrap_text(content) for content in last_message_content]
            dup_message["content"][-1]["cache_control"] = payload.pop("cache_control")
            payload["messages"][-1] = dup_message
        return payload

```

mdrxy:
Thanks. We are not not planning to overload `ChatOpenAI` any further, but rather will release a dedicated integration for OpenRouter as they have diverged from the chat completions standard. See #34328 for tracking

PR #33773, opened 2025-11-01
feat(openai): support cache control for openai

jingyugao:
issue https://github.com/langchain-ai/langchain/issues/33757
support cache control for ChatOpenAI,for some other claude provider(open router) can use cache control.


ccurme (commented):
Cache control on chat completions messages is an OpenRouter-specific concept.

It's not tenable to have ChatOpenAI accommodate all variants of chat completions APIs.

Ideally we would have a separate `langchain-openrouter` package that supports all OpenRouter features. It could inherit BaseChatOpenAI and extend it with support for cache control and other features.

PR #33777, opened 2025-11-01
chore: Rename 'completion' to 'text' in base.parse_with_prompt()

KushalBhurve:
This PR implements the planned refactoring to improve clarity and consistency in the BaseOutputParser interface.

The parameter name for the LLM output in BaseOutputParser.parse_with_prompt has been renamed from completion to text.

Why this change?
Clarity and Consistency: Using text is a more generic and accurate term for the raw string output from a Language Model (LLM), regardless of whether it represents a 'completion' (like in older models) or a more general 'response' (which might include chat-style outputs or function calls that are part of the text). This aligns with the terminology used in related libraries.

Documentation: The corresponding docstring and internal call to self.parse() have been updated to reflect the new parameter name.

This change is purely a refactor and maintains the existing functionality and public interface, only changing the internal parameter name and documentation within the method signature. The TODO comment # TODO: rename 'completion' -> 'text'. has been resolved.

### DIFF
--- a/libs/core/langchain_core/output_parsers/base.py
+++ b/libs/core/langchain_core/output_parsers/base.py
@@ -294,10 +294,9 @@ class BaseOutputParser(
         """
         return await run_in_executor(None, self.parse, text)
 
-    # TODO: rename 'completion' -> 'text'.
     def parse_with_prompt(
         self,
-        completion: str,
+        text: str,
         prompt: PromptValue,  # noqa: ARG002
     ) -> Any:
         """Parse the output of an LLM call with the input prompt for context.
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this. This is a public method, so although I'm not sure we document its use anywhere, this is a breaking change (not a refactor) because any callers that passed arguments by names will raise TypeError:
```python
parser.parse_with_prompt(completion="foo", prompt=bar)
```
Resolving the commented todo should be incorporated into a breaking release.

I don't think it's high priority (and was missed in v1) because output parsers are not used much in isolation anymore now that major providers support tool calling and dedicated structured output features.

PR #33793, opened 2025-11-03
test(anthropic): add vcr to `test_search_result_tool_message`

mdrxy:
To fix nondeterministic results causing integration testing to sometimes fail

Also speeds up from 10s to 0.5


libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -1700,6 +1700,8 @@ def test_files_api_pdf(block_format: str) -> None:
     _ = llm.invoke([input_message])
 
 
+@pytest.mark.default_cassette("test_search_result_tool_message.yaml.gz")
```
ccurme:
```suggestion
```
Nothing wrong with this line but I don't think it's needed here. It's typically used when we're parametrizing a test and want to force that the same cassette gets used for all parameterizations.

ccurme (approved):

PR #33796, opened 2025-11-03
fix(langchain_v1): disable summarization when max_tokens_before_summary is None

karamvirsingh1998:
**Description:** 

Fix `SummarizationMiddleware` to properly disable all summarization when `max_tokens_before_summary=None`, as documented in the docstring. 

Previously, when `max_tokens_before_summary=None`, the middleware would still trigger summarization if the number of messages exceeded `messages_to_keep` (default: 20). This violated the documented behavior that states: "If `None`, summarization is disabled."

The fix adds an early return when `max_tokens_before_summary` is None, ensuring summarization is completely disabled regardless of message count or token count.

**Changes:**
- Added explicit check in `before_model()` to return early when `max_tokens_before_summary` is None
- Added test case with 25 messages (exceeding `messages_to_keep=20`) to verify summarization remains disabled

**Issue:** N/A 
**Dependencies:** None

---

**Testing:**
- ✅ Added new test case: `test_summarization_middleware_no_summarization_cases` with 25 messages
- ✅ Existing tests continue to pass
- ✅ Verified fix

### DIFF
--- a/libs/langchain_v1/langchain/agents/middleware/summarization.py
+++ b/libs/langchain_v1/langchain/agents/middleware/summarization.py
@@ -104,11 +104,12 @@ class SummarizationMiddleware(AgentMiddleware):
         messages = state["messages"]
         self._ensure_message_ids(messages)
 
+        # If max_tokens_before_summary is None, summarization is disabled
+        if self.max_tokens_before_summary is None:
+            return None
+
         total_tokens = self.token_counter(messages)
-        if (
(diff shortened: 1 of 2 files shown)


karamvirsingh1998:
@ccurme  Please REVIEW THIS

ccurme (commented):
Hello, thanks for this!

We are updating the interface of SummarizationMiddleware in https://github.com/langchain-ai/langchain/pull/33825 and will address it there.

PR #33827, opened 2025-11-04
chore: correct `langchain-community` references

mdrxy:
fix docstrings that referenced community versions of now-native packages


mdrxy:
codspeed is wrong

PR #33835, opened 2025-11-05
docs: add feature support guide to chat models template

BushGommern:
## Description  
Adds comprehensive guidance for filling out the feature support table in the Chat Models template. This helps new contributors understand what each feature means and how to determine if a provider supports it.  
  
## Changes  
- Added detailed explanations for all 10 features in the table  
- Provided complete examples from real providers (OpenAI)  
- Created a validation checklist for contributors  
- Updated TODO comments to reference the new guide  
  
## Type of Change  
- [x] Documentation update  
- [ ] Bug fix  
- [ ] New feature  
- [ ] Breaking change  
  
## Testing  
- [x] Verified all feature descriptions are accurate  
- [x] Checked that examples match actual provider capabilities  
- [x] Ensured markdown renders correctly in Jupyter  
  
## Checklist  
- [x] Documentation follows project standards  
- [x] No breaking changes to existing templates  
- [x] Examples are from real, production providers  
- [x] Commit message follows Conventional Commits format


libs/cli/langchain_cli/integration_template/docs/chat.ipynb
```diff
@@ -26,7 +26,94 @@
...
+        "\n",
+        "#### JSON mode\n",
+        "- **Definition**: Model has a dedicated JSON output mode\n",
+        "- **Difference from structured output**: Only guarantees valid JSON, not schema conformance\n",
+        "- **Example**: OpenAI's `response_format={\"type\": \"json_object\"}`\n",
+        "\n",
+        "#### Image input\n",
+        "- **Definition**: Model supports image inputs (multimodal)\n",
+        "- **How to check**: Look for image_url or base64 image support in API\n",
+        "- **Example**: GPT-4 Vision, Claude 3 with vision\n",
```
ccurme, line 59: These models are several years old, why use them as examples instead of modern models?

ccurme (commented):
Thanks for this. We shouldn't put this in the integration template because it is not an expected part of the integration page. It would make more sense as part of a docs page on implementing custom chat models (we don't have much on that in the new docs site, closest we have is probably the [API reference](https://reference.langchain.com/python/langchain_core/language_models/)).

PR #33640, opened 2025-10-22
chore(langchain): Support for SystemMessage in create_agent

keenborder786:
- **Description:** Updated Function Signature of `create_agent`, the system prompt can be both a list and string. I see no harm in doing this, since SystemMessage accepts both.
- **Issue:** #33630

PR #33889, opened 2025-11-07
revert: Support for SystemMessage in create_agent

ccurme:
Reverts langchain-ai/langchain#33640

Introduces lint errors into langchain-anthropic

Should incorporate into 1.1 instead of patch release.


jvjfranca-auramind:
Any idea when this will be released? its critical for us.

ccurme:
@jvjfranca-auramind would you mind sharing your use case? Interested to hear.

This is a delicate change— any middleware that is assuming a string for `system_prompt` will need to be updated.

jvjfranca-auramind:
Sure, @ccurme 

Our system prompt contains dynamic components that are generated at runtime and include metadata. Since the prompt consists of a large set of instructions, we need to split it into chunks to leverage AWS Bedrock's prompt caching feature.
Currently, we cannot cache the system prompt because the dynamic parts invalidate the cache. To solve this, we need to structure it as a list where we can mark specific sections for caching, allowing Bedrock to cache the static portions while dynamic content updates independently.

Its a huge cost savings.

Thanks!

PR #33899, opened 2025-11-08
fix(chroma): resolve OpenCLIP + Chroma image embedding test regression

AmazingcatAndrew:
**Description:**  
Fixes the OpenCLIP × Chroma regression that caused nested embedding errors when adding or searching image data.  
The test case `test_openclip_chroma_embed_no_nesting_error` has been restored and verified to work correctly with the current LangChain core dependencies.  
Functional validation confirms that `similarity_search_by_image` now returns correct, metadata‑preserving results.  

**Issue:**  
Fixes #33851

**Dependencies:**  
No new dependencies introduced.  

**Testing:**  
All tests under  
```bash
uv run --group test pytest tests/unit_tests
```  
result:
```
30 passed in 91.26s (0:01:31)
```
have passed successfully using Python 3.13.9 and uv‑managed environment.  
This confirms that the regression has been fixed.  

Running  
```bash
make test
```  
still produces cleanup‑time `AttributeError: 'ProactorEventLoop' object has no attribute '_ssock'` on Windows (Python 3.13+).  
This is a benign asyncio teardown message rather than a functional failure.  
`uv run pytest` closes event loops immediately after tests, while `make test` invokes pytest through a secondary process layer that leaves a background loop alive at interpreter shutdown.  
This difference in teardown behavior explains the extra messages seen only when using `make test`.  

**Summary:**  
- Verified the OpenCLIP + Chroma image pipeline works correctly.  
- `uv run --group test pytest` fully passes; the fix is complete.  
- The residual `_ssock` warnings occur only during Windows asyncio cleanup and are not related to this code change.  

This is my first time contributing code, please contact me with any questions

---


AmazingcatAndrew:
I tried many ways but AttributeError: '_WindowsSelectorEventLoop' object has no attribute '_ssock'  still exists. I'll be very grateful if you have any ideas on it.

Issue #33883, opened 2025-11-07
Strip message ID to improve cache performance


mdrxy:
### Checked other resources

- [x] This is a feature request, not a bug report or usage question.
- [x] I added a clear and descriptive title that summarizes the feature request.
- [x] I used the GitHub search to find a similar feature request and didn't find it.
- [x] I checked the LangChain documentation and API reference to see if this feature already exists.
- [x] This is not related to the langchain-community package.

### Feature Description

Implement message ID normalization for improved caching efficiency with chat models.

Messages have unique IDs which are used as part of the cache keys generation process, causing functionally identical messages to create separate cache entries.

This feature would add a `_strip_ids_from_messages` function to normalize messages before cache key generation, ensuring that messages with identical content produce cache hits regardless of their unique identifiers.

### Use Case

Messages with identical content but different IDs bypass cache hits, leading to reduced cache effectiveness and hit rates.

### Additional Context

Original attempt in #32642

mdrxy:
Closed by #33915

PR #33915, opened 2025-11-10
fix(core): strip message IDs from cache keys using `model_copy`

aroun-coumar:
**Description:**  

*Closes #[33883](https://github.com/langchain-ai/langchain/issues/33883)*

Chat model cache keys are generated by serializing messages via `dumps(messages)`. The optional `BaseMessage.id` field (a UUID used solely for tracing/threading) is included in this serialization, causing functionally identical messages to produce different cache keys. This results in repeated API calls, cache bloat, and degraded performance in production workloads (e.g., agents, RAG chains, long conversations).

This change normalizes messages **only for cache key generation** by stripping the nonsemantic `id` field using Pydantic V2’s `model_copy(update={"id": None})`. The normalization is applied in both synchronous and asynchronous cache paths (`_generate_with_cache` / `_agenerate_with_cache`) immediately before `dumps()`.

```python
normalized_messages = [
    msg.model_copy(update={"id": None})
    if getattr(msg, "id", None) is not None
    else msg
    for msg in messages
]
prompt = dumps(normalized_messages)


mdrxy (changes requested):
The async path fix is missing

Could you please add tests?

Otherwise looks good!

Issue #33920, opened 2025-11-11
[langchain-anthropic] Code execution tool result blocks missing from streaming responses


henryl:
## Issue Description

Code execution tool result blocks (`bash_code_execution_tool_result`, `text_editor_code_execution_tool_result`) are present in non-streaming responses but are **missing from streaming responses** when using `langchain-anthropic` with Anthropic's code execution tools.

## Environment

- **Package**: `langchain-anthropic==1.0.2`
- **Dependencies**:
  - `anthropic==0.72.0`
  - `langchain-core==1.0.4`
- **Python**: 3.13
- **Model**: `claude-sonnet-4-20250514`
- **Beta**: `code-execution-2025-08-25`

## Expected Behavior

When streaming with `.astream()`, tool result content blocks should be included in the stream, similar to how tool use blocks are streamed. The Anthropic API returns these result blocks, and they should be exposed during streaming.

Expected result block types:
- `bash_code_execution_tool_result` - for bash command execution results
- `text_editor_code_execution_tool_result` - for file operation results
- `code_execution_tool_result` - for Python-only execution (legacy)

## Actual Behavior

**Non-streaming (`.invoke()`)**: ✅ Result blocks present
```python
Content blocks: 7
├─ Block 0: text
├─ Block 1: server_tool_use
├─ Block 2: text_editor_code_execution_tool_result ✅
├─ Block 3: text
├─ Block 4: server_tool_use
├─ Block 5: bash_code_execution_tool_result ✅
└─ Block 6: text
```

**Streaming (`.astream()`)**: ❌ Result blocks missing
```python
Tool use blocks found: 2 ✅
Result blocks found: 0 ❌
```

Only `server_tool_use` blocks are streamed. Result blocks are absent during streaming but appear in the final consolidated message.

## Reproduction Script

```python
#!/usr/bin/env python3
"""
Reproduction script for missing code execution tool result blocks during streaming.

Requirements:
    pip install langchain-anthropic python-dotenv

Environment:
    ANTHROPIC_API_KEY must be set
"""

import asyncio
import os

from dotenv import load_dotenv
from langchain_anthropic import ChatAnthropic


async def test_non_streaming():
    """Test non-streaming - result blocks SHOULD appear"""
    print("=" * 80)
    print("TEST 1: Non-Streaming (.invoke())")
    print("=" * 80)

    model = ChatAnthropic(
        model="claude-sonnet-4-20250514",
        betas=["code-execution-2025-08-25"],
    )

[...]

PR #33925, opened 2025-11-11
feat(anthropic): support code_execution_20250825

keenborder786:
- **Description:** The creation of `tool_result` and `tool_call` blocks in streaming response in `_make_message_chunk_from_anthropic_event` has been made dynamic in order to avoid hard-coding the tools e.g `bash_code_execution_tool_result` and `code_execution_tool_result` (currently we were missing `bash_code_execution_tool_result` which causing the issue below). But Anthropic always have a suffix `*_tool_result` and `*_tool_resullt` so we just use that.
  - **Issue:** #33920


ccurme (approved):
thank you!

PR #33932, opened 2025-11-12
fix(core): Resolve 'FieldInfo is not JSON serializable' error in Pydantic v2

jitokim:
Fixes TypeError when generating JSON schemas: "Object of type 'FieldInfo'
is not JSON serializable". This occurs when model_json_schema() includes
internal Pydantic metadata in the output.

Solution: Add mode='serialization' parameter to model_json_schema() to
ensure only JSON-serializable data structures are generated.

Changes:
- Modified _convert_pydantic_to_openai_function() to pass mode='serialization'
- Added test_pydantic_fieldinfo_serialization() to verify correct behavior
- Resolves nested Pydantic v2 model conversion issues

Impact:
- Fixes schema generation for nested Pydantic v2 models
- No breaking changes (Pydantic v1 code path unchanged)
- Previously failing test_convert_to_openai_function_nested_v2 now passes


libs/core/tests/unit_tests/utils/test_function_calling.py
```diff
@@ -1168,3 +1170,63 @@ class MyModel(BaseModel):
...
+
+        required_field: str = Field(..., description="A required field")
+        optional_field: str | None = Field(None, description="An optional field")
+        default_field: int = Field(default=42, description="A field with default")
+
+    # Convert to OpenAI function format
+    result = convert_to_openai_function(MyModel)
+
+    # This should not raise an error if FieldInfo objects are properly serialized
+    json_str = json.dumps(result)
```
eyurtsev: can we make this test dumb and check result against a dict with a single assert?

We want to be able to see immediately what the expected value of result is instead of having to read through different assertions

jitokim: @eyurtsev  Thanks for reivew

sydney-runkle:
Is this test failing on `master`? What does the schema look like w/ `mode='validation'`? I'm surprised...

jitokim:
> Is this test failing on `master`? What does the schema look like w/ `mode='validation'`? I'm surprised...

@sydney-runkle  Thanks for the thorough question.

I pushed another commit that forces JsonOutputParser to call model_json_schema(mode="serialization") and adds an explicit repro test. Here’s what I found:

  branch: master

  - JsonOutputParser still uses the validation-mode schema on master.
  - When a Pydantic v2 model leaves behind a FieldInfo object, json.dumps raises TypeError: Object of type FieldInfo
    is not JSON serializable.
  - Screenshot shows TEST_FILE=tests/unit_tests/output_parsers/
    test_json.py::test_json_parser_uses_serialization_schema_mode failing on master with the validation-mode schema
    containing the injected FieldInfo.
    
image

---
branch : jitokim:fix/pydantic-fieldinfo-serialization

  - After forcing serialization mode in JsonOutputParser, the same test passes.
  - The screenshot includes both schema snippets: validation-mode schema contains the injected FieldInfo,
    serialization-mode schema omits it.
  - This matches the real-world TypeError we saw, just surfaced at the parser level.

image

---

For reference, langchain-google already had to patch the same issue downstream:
https://github.com/langchain-ai/langchain-google/pull/1317

mdrxy:
Thanks for working on this and for adding the repro details.

I was able to reproduce the `FieldInfo is not JSON serializable` failure on `master`, so the underlying issue is real. That said, I don’t think this PR is merge-ready in its current form:

- The branch now has conflicts and would need a rebase.
- The `JsonOutputParser` repro is useful, but the added `test_pydantic_fieldinfo_serialization` for `convert_to_openai_function` appears to pass on `master`, so it doesn’t demonstrate that path is currently broken.
- Switching `convert_to_openai_function()` to `mode="serialization"` may be semantically risky, since tool/function schemas describe inputs, where Pydantic’s validation schema is generally the more appropriate mode.
- The broad `except TypeError` fallback around `model_json_schema(mode="serialization")` could hide unrelated schema-generation errors.

I’m going to close this PR for now, but the issue is worth revisiting. A narrower follow-up PR focused on `JsonOutputParser` with a failing repro would be welcome. If there’s a concrete `convert_to_openai_function` repro from a real integration, we can evaluate that separately.

PR #33934, opened 2025-11-12
chore(cli,standard-tests,text-splitters): fix some ruff TC rules

cbornet:


mdrxy:
we can comma separate scopes now!

mdrxy (approved):

PR #33939, opened 2025-11-12
fix(langchain, openai): fix create_agent / response_format for Responses API

ccurme:
`create_agent` / `response_format` does not currently work for the Responses API.

Here we make these changes:
- `langchain`: pass JSON schema instead of OpenAI format for `response_format`
- `langchain-openai`: accept JSON schema (and TypedDict) for `response_format` in `bind_tools`.

We also beef up testing. We take `langchain` integration tests for `create_agent` / `response_format` (which never run), fix them up, add VCR cassettes and move them to unit tests so that they run during extended tests as part of PR CI.


sydney-runkle (approved):
This looks awesome. Could we also confirm this works w/ grok (might require a change in that provider package as well)?

Thank you!

ccurme:
> Could we also confirm this works w/ grok (might require a change in that provider package as well)?

It does (for both xAI's chat completions and Responses APIs), because ChatXAI inherits BaseChatOpenAI and uses its bind_tools.

Issue #33903, opened 2025-11-09
astream is not sending token in python 3.10


jgasparetti:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I read what a minimal reproducible example is (https://stackoverflow.com/help/minimal-reproducible-example).
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Example Code

```python

import os

from langchain.agents import create_agent
from langgraph.checkpoint.memory import InMemorySaver
from langchain_core.messages import HumanMessage
from langchain_ollama import ChatOllama

import logging

logger = logging.getLogger(__name__)

__config = {"configurable": {"thread_id": "1"} }

__agent = create_agent(
    tools=[],
    model=ChatOllama(
        temperature=temperature,
        base_url=os.getenv("LLM_URL"),
        stream=True,
    ),
    system_prompt="You are a good agent",
    checkpointer=InMemorySaver(),
)

async def chat(self, input: str):
    async for token, metadata in __agent.astream(
        input={"messages": [HumanMessage(content=input)]},
        config=__config,
        stream_mode="messages",
    ):
        logger.info("--------------------------------")
        logger.info(f"token: {token}")
        logger.info(f"metadata: {metadata}")
        logger.info("--------------------------------")
        if metadata["langgraph_node"] == "model":
            if token.content_blocks and token.content_blocks[0]["type"] == "text":
                yield token.content_blocks[0]["text"]
            elif token.content_blocks and token.content_blocks[0]["type"] == "tool_call_chunk" and token.content_blocks[0]["name"]:
                logger.info(f"calling tool: {token.content_blocks[0]['name']}")
        elif metadata["langgraph_node"] == "tool":
            logger.info(f"tool result: {token.content_blocks[0]['text']}")

```

### Description

In python 3.10 the _agent.astream_ function is not sending each token but send only the full message.
I am developing with a development PC with 3.10, however I tested the same code with python 3.12 on the same development pc and it is working perfectly. 
However I can not upgrade python since I have to use a Jetson Orin with python 3.10 pre-installed. 

### System Info

System Information
------------------
[...]

PR #33942, opened 2025-11-13
fix: stream iterator results as they complete

abbasyed:
**Description:** Fixes a bug where `AgentExecutorIterator` batches all tool execution results instead of streaming them as they complete. The code was using `asyncio.gather()` which waits for all concurrent tool calls to finish before yielding any results, causing delays and blocking in async iteration. This is particularly noticeable on Windows with Python 3.10's ProactorEventLoop but affects all platforms. 

  The fix replaces `asyncio.gather()` with `asyncio.as_completed()` to yield results immediately as each tool completes, and adds explicit event loop yielding for better task scheduling. The code even had a TODO comment acknowledging this limitation.

  **Issue:** N/A (bug reported in community, no tracked issue)

  **Dependencies:** None

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -4700,15 +4700,14 @@ class RunnableLambda(Runnable[Input, Output]):
     ) -> Output:
         if inspect.isgeneratorfunction(self.func):
             output: Output | None = None
-            with contextlib.closing(
-                call_func_with_variable_args(
-                    cast("Callable[[Input], Iterator[Output]]", self.func),
-                    input_,
-                    config,
-                    run_manager,
-                    **kwargs,
-                )
-            ) as generator:
+            generator = call_func_with_variable_args(
+                cast("Callable[[Input], Iterator[Output]]", self.func),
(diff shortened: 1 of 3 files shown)


eyurtsev (commented):
We're not making improvements in AgentExecutor -- this executor is nearly 2 years old and has a lot of known design issues.

The agent executor is included in the code for backwards compatibility reasons, but users should swap to the current agent implementation

See docs here: https://docs.langchain.com/oss/python/langchain/overview

eyurtsev:
https://github.com/langchain-ai/langchain/issues/33903 - is not for AgentExecutor but using the new abstraction

abbasyed:
> We're not making improvements in AgentExecutor -- this executor is nearly 2 years old and has a lot of known design issues.
> 
> The agent executor is included in the code for backwards compatibility reasons, but users should swap to the current agent implementation
> 
> See docs here: https://docs.langchain.com/oss/python/langchain/overview

Got it, thanks for clarifying! I'll close this PR and focus on the current agent implementation instead. Appreciate the pointer to the docs!

PR #33968, opened 2025-11-14
refactor(standard-tests): improve VCR config

cbornet:
Use of the fixture `_base_vcr_config` is deprecated with alternative function `base_vcr_config()`
This way:
* we don't need to import `_base_vcr_config` seen as unused (which leads to ruff violations PLC0414 and F811)
* we don't need to make a copy since a new dict is created at each function invocation

### DIFF
--- a/libs/standard-tests/langchain_tests/conftest.py
+++ b/libs/standard-tests/langchain_tests/conftest.py
@@ -8,6 +8,7 @@ from typing import TYPE_CHECKING, Any, cast
 
 import pytest
 import yaml
+from langchain_core._api.deprecation import deprecated
 from vcr import VCR
 from vcr.persisters.filesystem import CassetteNotFoundError
 from vcr.request import Request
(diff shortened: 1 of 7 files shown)


libs/standard-tests/langchain_tests/conftest.py
```diff
@@ -116,6 +116,12 @@ def _base_vcr_config() -> dict:
 
 
 @pytest.fixture(scope="session")
-def vcr_config(_base_vcr_config: dict) -> dict:
+@deprecated("1.0.1", alternative="base_vcr_config", removal="2.0")
```
mdrxy:
```suggestion
@deprecated("1.0.3", alternative="base_vcr_config", removal="2.0")
```

mdrxy (dismissed):

cbornet:
@mdrxy can be merged ?

PR #34077, opened 2025-11-23
feat(text-splitters): add MySQL Language support for RecursiveCharacterTextSplitter Class with delimiter aware splitting

Emmanuel-Arokiaraj:
**Description:**
Enhanced `RecursiveCharacterTextSplitter` for `Language.MYSQL` by adding a comprehensive list of SQL separators including DDL, DML, and control-flow constructs. This enables more logical splitting of MySQL scripts—especially those containing stored procedures, triggers, functions, and multiline statements—while preserving structural integrity and block boundaries.

Additionally, improved delimiter handling so MySQL‐specific separators are evaluated before the default newline/space behavior, minimizing incorrect splits. Updated unit tests accordingly and adjusted `chunk_size` to avoid unintended recursive breaking during splitting.

**Issue:** N/A

**Dependencies:** None

---

### **Test Notes**

For the MySQL test `test_mysql_query_splits`, the previous configuration used `chunk_size=100`, but the first expected block exceeds 135 characters. With the default recursive behavior, this resulted in excessive splitting. Updated example usage:

```python
splitter = RecursiveCharacterTextSplitter.from_language(
    language=Language.MYSQL,
    chunk_size=500,
    chunk_overlap=0,
)
```


ccurme (commented):
SQL is tricky.
- Many of the tokens here are not specific to the MySQL flavor of SQL. Do we maintain separate sets for each variant?
- Case sensitivity
- You can have comments, whitespace, etc. between tokens
- Keywords inside strings (e.g., `create temporary table`, `create or replace view`)

This will ultimately be brittle and I'm not sure we want to go down the road of maintaining it. Thanks very much for submitting it, though. Consider opening an issue so we can capture community interest.

PR #34080, opened 2025-11-24
fix(text-splitters): reverse preserved elements iterator in `HTMLSemanticPreservingSplitter`

RohanDisa:
**Description:** Reversed the order of preserved elements iteration in HTMLSemanticPreservingSplitter and added a unit test to validate the behavior.  

**Issue:** #31568  

**Dependencies:** None


libs/text-splitters/tests/unit_tests/test_text_splitters.py
```diff
@@ -3973,6 +3974,79 @@ def test_html_splitter_keep_separator_default() -> None:
     assert documents == expected
 
 
+@pytest.mark.requires("bs4")
+def test_html_splitter_preserved_elements_reverse_order() -> None:
```
ccurme, line 3978: This test passes on master and doesn't appear to exercise the new logic.

ccurme (approved):

PR #34091, opened 2025-11-24
fix(langchain): never exceed `keep` threshold in SummarizationMiddleware

ccurme:


ccurme:
Can revive if needed, bigger issue is single large ToolMessages.

PR #34101, opened 2025-11-25
fix: support Gemini structured output in ProviderStrategy

RohanDisa:
Description:
Adds Gemini-specific handling in ProviderStrategy.to_model_kwargs() to support structured output. The change includes adding response_schema and response_mime_type for Gemini models, and updating the factory to pass the model instance when calling to_model_kwargs(). This prevents JSONDecodeError when using ProviderStrategy with Gemini, while keeping OpenAI default behavior unchanged.

Issue: langchain-ai/langchain-google#1395

Dependencies: No new dependencies;


ccurme (commented):
Thanks for submitting this!

This should be implemented in [langchain-google](https://github.com/langchain-ai/langchain-google).

Those chat models should accept `response_format=<json schema>` in addition to the format returned here.

[This function](https://github.com/langchain-ai/langchain-google/blob/93bd5e6906137e0cd7bdd51146c2990421db5592/libs/genai/langchain_google_genai/chat_models.py#L2504) should pop that out of kwargs and format it as is done here.

A minimal solution would accept `response_format=<json schema>` and `response_format=<the format currently returned here>`. Ideally we would also accept `response_format=<Pydantic BaseModel>` and `response_format=<TypedDict>` but that can be added later.

Note that Gemini currently does not support simultaneous tool calling and structured output, so this can only work in the case with no tools.

PR #34112, opened 2025-11-27
fix(core): Raise informative error for reserved 'config' tool argument

dumko2001:
This PR adds a validation check to `StructuredTool.from_function` to raise a `ValueError` when a user defines a tool with the reserved argument name `config`. This prevents silent name collisions with `RunnableConfig` that previously led to confusing downstream `TypeErrors` during execution.

Fixes #34029

**Changes:**
- Added "fail fast" validation in `libs/core/langchain_core/tools/structured.py`.
- Updated `@tool` docstrings in `convert.py` to warn users.
- Added regression test `test_tool_init_config_arg_raises_error` in `test_tools.py`.

**Breaking Changes:**
- [x] Yes
*Note: This strictly affects code that uses `config` as a tool argument. Previously, this code would likely crash at runtime or behave unexpectedly. It will now raise a clear `ValueError` at definition time.*


ccurme (commented):
I'm not sure we want to do block this categorically. This currently does not raise:
```python
from langchain_core.tools import StructuredTool

def bad_tool(config: str) -> str:
    """Tool description."""
    return "I will fail"

StructuredTool.from_function(bad_tool)  # does not raise
```
and we can use such tools just fine in some cases:
```python
from langchain.tools import tool


@tool
def my_tool(config: int) -> None:
    """Tool description."""
    pass


model = init_chat_model("openai:gpt-5.2").bind_tools(
    [my_tool], tool_choice="my_tool"
)

response = model.invoke("config is 3")
response.tool_calls
# [{'name': 'my_tool',
#   'args': {'config': 3},
#   'id': 'call_osGtDmXvP1zzblOs0i77l3u1',
#   'type': 'tool_call'}]
```

PR #34120, opened 2025-11-27
feat(core,anthropic): `extras` on `BaseTool`

mdrxy:
added `extras` as an attribute on `BaseTool` to serve as ground to attach metadata to tool instances. Used specifically in the case of Claude's new optional tool options such as deferred loading for [tool search](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool) or input examples, e.g.:

```python
@tool(extras={"defer_loading": True})
def search_files(query: str) -> str:
    """Search through files in the workspace.

    Args:
        query: Search query
    """
    return f"Found 3 files matching '{query}'"
```

Notably, this supports [executable callbacks for tools](https://github.com/langchain-ai/langchain/pull/34279):

```python
@tool(extras={"provider_tool_definition": tool_spec})
def bash(*, command: str, restart: bool = False, **kw):
    """Execute a bash command."""
    if restart:
        return "Bash session restarted"
    try:
        result = subprocess.run(
...
```

Chat models should process known extras when constructing request payloads.


libs/core/langchain_core/tools/convert.py
```diff
@@ -130,6 +135,14 @@ def tool(
             parse parameter descriptions from Google Style function docstrings.
         error_on_invalid_docstring: If `parse_docstring` is provided, configure
             whether to raise `ValueError` on invalid Google Style docstrings.
+        extras: Optional provider-specific extra fields for the tool.
```
ccurme, line 138: We might document that provider packages should handle these in bind_tools.

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -147,6 +147,56 @@ class AnthropicTool(TypedDict):
...
 
+# Allowlist of valid Anthropic-specific extra fields
+_ANTHROPIC_EXTRA_FIELDS: set[str] = {
+    "cache_control",
+    "defer_loading",
+    "input_examples",
+}
+
+
+def _validate_anthropic_extras(extras: dict[str, Any]) -> dict[str, Any]:
```
ccurme: (nit) Would we ever want to put extras from more than one provider on a single tool?

Instead of validating and applying all, would we want `bind_tools` to just filter and apply known tools?

libs/partners/anthropic/tests/integration_tests/test_chat_models.py
```diff
@@ -2163,6 +2163,90 @@ def test_tool_search_with_deferred_tools() -> None:
     assert calculator_tool.get("defer_loading") is True
 
 
+def test_tool_search_with_langchain_tools_extras() -> None:
+    """Test tool search with LangChain tools using extras parameter."""
+    from langchain_core.tools import tool
```
ccurme: why not import top level?

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -142,6 +142,56 @@ class AnthropicTool(TypedDict):
...
+                if not isinstance(value, dict):
+                    msg = f"'cache_control' must be a dict, got {type(value).__name__}"
+                    raise ValueError(msg)
+            elif key == "defer_loading":
+                if not isinstance(value, bool):
+                    msg = f"'defer_loading' must be a bool, got {type(value).__name__}"
+                    raise ValueError(msg)
+            elif key == "input_examples" and not isinstance(value, list):
+                msg = f"'input_examples' must be a list, got {type(value).__name__}"
+                raise ValueError(msg)
```
ccurme: anthropic gives an informative error message if the types are wrong, IMO it's preferable to let it come from them instead of maintaining expected types here.

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -2568,12 +2602,24 @@ def get_weather(location: str) -> str:
             See LangChain [docs](https://docs.langchain.com/oss/python/integrations/chat/anthropic#strict-tool-use)
             for more detail.
         """  # noqa: E501
-        formatted_tools = [
-            tool
-            if _is_builtin_tool(tool)
-            else convert_to_anthropic_tool(tool, strict=strict)
-            for tool in tools
-        ]
+        formatted_tools: list[dict[str, Any] | type | Callable | BaseTool] = []
```
ccurme: the formatted tools are all either built-ins (dicts) or our `AnthropicTool` typed dict. will refactor this

ccurme (dismissed):

ccurme (approved):

PR #34121, opened 2025-11-27
fix(core): fix for `exclude_inputs` `POST`

mdrxy:
- Added logic to set the inputs_is_truthy flag when creating runs; flag is then used in the existing PATCH logic:

  ```python
  run.patch(exclude_inputs=run.extra.get("inputs_is_truthy", False))
  ```

- Test

---

Normal invocations:
- `POST`: Contains real input like `{"input": "real input"}`
- `inputs_is_truthy = True`
- `PATCH`: `exclude_inputs=True` (inputs not sent again)

Streaming invocations:
- `POST`: Contains placeholder `{"input": ""}`
- `inputs_is_truthy = False`
- `PATCH`: `exclude_inputs=False` (real accumulated inputs sent)


mdrxy:
Superseding with #34416

PR #34125, opened 2025-11-27
fix(fireworks): models used in tests & naming schema

mdrxy:
it appears their docs are wrong? will wait a few days and see


libs/partners/fireworks/langchain_fireworks/chat_models.py
```diff
@@ -287,9 +287,7 @@ class ChatFireworks(BaseChatModel):
         ```python
         from langchain_fireworks.chat_models import ChatFireworks
 
-        fireworks = ChatFireworks(
-            model_name="accounts/fireworks/models/llama-v3p1-8b-instruct"
-        )
+        fireworks = ChatFireworks(model_name="fireworks/llama-v3p3-70b-instruct")
```
RobertoBarrosoLuque: This model is deprecated from serverless, would recommend we use `accounts/fireworks/models/gpt-oss-120b` or another current serverless model instead

PR #34150, opened 2025-11-30
feat(langchain): add included_tools parameter to ClearToolUsesEdit

### LINKED ISSUE #34145: ClearToolUsesEdit add option to select specific tools instead of having to exclude tools
### Checked other resources

- [x] This is a feature request, not a bug report or usage question.
- [x] I added a clear and descriptive title that summarizes the feature request.
- [x] I used the GitHub search to find a similar feature request and didn't find it.
- [x] I checked the LangChain documentation and API reference to see if this feature already exists.
- [x] This is not related to the langchain-community package.

### Package (Required)

- [x] langchain
- [ ] langchain-openai
- [ ] langchain-anthropic
- [ ] langchain-classic
- [ ] langchain-core
- [ ] langchain-cli
- [ ] langchain-tests
- [ ] langchain-chroma
- [ ] langchain-deepseek
- [ ] langchain-exa
- [ ] langchain-fireworks
- [ ] langchain-groq
- [ ] langchain-huggingface
- [ ] langchain-mistralai
- [ ] langchain-nomic
- [ ] langchain-ollama
- [ ] langchain-perplexity
- [ ] langchain-prompty
- [ ] langchain-qdrant
- [ ] langchain-xai

### Feature Description

The current signature of the ClearToolUsesEdit middleware is as follows:

```python
ClearToolUsesEdit(
    trigger: int = 100000,
    clear_at_least: int = 0,
    keep: int = 3,
    clear_tool_inputs: bool = False,
    exclude_tools: Sequence[str] = (),
    placeholder: str = DEFAULT_TOOL_PLACEHOLDER,
)
```

I propose to add a parameter `include_tools: Sequence[str] = ()` which allows the user to explicitly select which tools this middleware should be applied to. From my point of view this is a much more intuitive design than excluding tools. The cleanest way would be to replace `exclude_tools` with `include_tools`, but during a deprecation period one could allow either one or the other parameter.

### Use Case

In my workflows, there are often tools that generate a lot of context and other tools, that don't. One example is a tool that reads code from the user. I want to remove this code from the model's context as soon as newer code is provided. I therefore only want to target the tool output of this specific `read_code_tool`. Since my agent uses 5-10 tools, I currently have to exclude all other tools which creates a lot of overhead, especially when adding new tools to the agent. It therefore would be great if one had to only select the targeted tool once.

### Proposed Solution

I think the implementation would be straight forward and I would be happy to take this over.

### Alternatives Considered

_No response_

### Additional Context

_No response_

valentingoelz:
Fixes #34145 

added a parameter `included_tools` to the `ClearToolUsesEdit` class. Let me know if the validation should be done differently.


sydney-runkle (changes requested):
This looks good, could you please add some tests?

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

Issue #30580, opened 2025-03-31
I can't get any output from Qwen's thinking models QwQ series


yigit353:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

```python
from langchain_openai import ChatOpenAI
import os

llm = ChatOpenAI(
    api_key=os.getenv("DASHSCOPE_API_KEY"), # If you have not configured the environment variable, replace DASHSCOPE_API_KEY with your API key
    base_url="https://dashscope-intl.aliyuncs.com/compatible-mode/v1", # Replace https://dashscope-intl.aliyuncs.com/compatible-mode/v1 with the base_url of the DashScope SDK
    model="qwq-plus", # Use qwn-plus as an example. You can use other models in the model list: https://www.alibabacloud.com/help/en/model-studio/getting-started/models
    max_completion_tokens=3000
    )

query = """Answer this question or if you can't, tell me which additional information you need:
how to get reasoning_content without stream when using chatopenai? I am using the qwq32 reason model which can output reasoning_content when using openai SDK"""

messages = [
    # {"role":"system","content":"You are a helpful assistant."}, 
    {"role":"user","content":query}
]

response = llm.stream(messages)

print("\n" + "=" * 20 + "Reasoning Process" + "=" * 20 + "\n")
for chunk in response:
        print(chunk)
        # print(chunk.model_dump_json())

llm.invoke(messages)
```

### Error Message and Stack Trace (if applicable)

====================Reasoning Process====================

content='' additional_kwargs={} response_metadata={} id='run-00c44c92-10a5-41e1-8449-94d4dd615006'
content='' additional_kwargs={} response_metadata={} id='run-00c44c92-10a5-41e1-8449-94d4dd615006'
content='' additional_kwargs={} response_metadata={} id='run-00c44c92-10a5-41e1-8449-94d4dd615006'
content='' additional_kwargs={} response_metadata={} id='run-00c44c92-10a5-41e1-8449-94d4dd615006'
content='' additional_kwargs={} response_metadata={} id='run-00c44c92-10a5-41e1-8449-94d4dd615006'
[...]

Issue #29513, opened 2025-01-30
Cannot retrieve reasoning_content while streaming


yigit353:
### Checked other resources

- [x] I added a very descriptive title to this issue.
- [x] I searched the LangChain documentation with the integrated search.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).

### Example Code

When I run the simple code:

```python
from dotenv import load_dotenv
from langchain_core.prompts import ChatPromptTemplate
from langchain_deepseek import ChatDeepSeek

load_dotenv()

llm = ChatDeepSeek(model="deepseek-reasoner", temperature=0, streaming=True)

prompt = ChatPromptTemplate(
    [
        (
            "system",
            "You are a helpful assistant that translates {input_language} to {output_language}.",
        ),
        ("human", "{input}"),
    ]
)

params = {
    "input_language": "English",
    "output_language": "German",
    "input": "I love programming.",
}

messages = prompt.invoke(params)

for chunk in llm.stream(messages):
    print(chunk)
```
I get empty reasoning_content while streaming. However, I need it to display in UI in real-time.
For that I wrote my own solution:
```python
import asyncio
from typing import Any, Optional, Iterator
from typing import AsyncIterator

from langchain.schema import HumanMessage, AIMessage, SystemMessage
from langchain_core.messages import AIMessageChunk
from langchain_core.outputs import ChatGenerationChunk, LLMResult
from langchain_openai import ChatOpenAI


class DeepseekChatOpenAI(ChatOpenAI):
    async def _astream(
            self,
            messages: Any,
            stop: Optional[Any] = None,
            run_manager: Optional[Any] = None,
            **kwargs: Any,
    ) -> AsyncIterator[AIMessageChunk]:
        openai_messages = []
        for msg in messages:
            if isinstance(msg, HumanMessage):
                openai_messages.append({"role": "user", "content": msg.content})
            elif isinstance(msg, AIMessage):
                openai_messages.append({"role": "assistant", "content": msg.content})
            elif isinstance(msg, SystemMessage):
                openai_messages.append({"role": "system", "content": msg.content})
            else:
                raise ValueError(f"Unsupported message type: {type(msg)}")

        params = {
            "model": self.model_name,
            "messages": openai_messages,
            **self.model_kwargs,
            **kwargs,
            "extra_body": {
[...]

PR #34152, opened 2025-12-01
fix(openai): support reasoning_content in streaming delta chunks

codeonbush:
## Description

This PR adds support for the `reasoning_content` field in streaming delta chunks from the Chat Completions API.

### Problem

The `_convert_delta_to_message_chunk` function does not handle the `reasoning_content` field, which is returned by:

- **LiteLLM** when proxying Gemini models with `reasoning_effort` enabled

- **DeepSeek** reasoning models (deepseek-reasoner)

- **vLLM** with reasoning parser enabled

- Other OpenAI-compatible providers that support reasoning/thinking output

This causes the reasoning/thinking content to be completely lost during streaming.

Related issues: #29513, #31326, #32981, #30580

### Solution

Add handling for `reasoning_content` in `_convert_delta_to_message_chunk`, storing it in `additional_kwargs` (consistent with how the Responses API handles reasoning content).

### Changes

1. Modified `_convert_delta_to_message_chunk` to extract `reasoning_content` from delta and store it in `additional_kwargs`

2. Added comprehensive unit tests covering various scenarios
3. Includes import formatting changes required by ruff in CI

### Testing

Added unit tests for:

- Delta with both `reasoning_content` and `content`

- Delta with only `reasoning_content`
 
- Delta without `reasoning_content` (normal case)

- Delta with `reasoning_content` and `tool_calls`

## Checklist

- [x] Added relevant unit tests

- [x] Followed conventional commit format

- [x] Code changes are minimal and focused

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -23,14 +23,7 @@ from io import BytesIO
 from json import JSONDecodeError
 from math import ceil
 from operator import itemgetter
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Literal,
-    TypeAlias,
-    TypeVar,
-    cast,
-)
+from typing import TYPE_CHECKING, Any, Literal, TypeAlias, TypeVar, cast
 from urllib.parse import urlparse
 
 import certifi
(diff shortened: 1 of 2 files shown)


mdrxy:
`reasoning_content` is not returned by the OpenAI or Azure OpenAI Chat Completions APIs. OpenAI's Chat Completions API [does not support reasoning output](https://platform.openai.com/docs/guides/reasoning?api-mode=chat#reasoning-summaries). OpenAI does offer reasoning summaries as part of its Responses API, and that is supported by ChatOpenAI (see docs [here](https://python.langchain.com/docs/integrations/chat/openai/#reasoning-output)).

`langchain-openai` would become unwieldy if we attempted to accommodate all variants of Chat Completions APIs inside it. So instead we just extend `BaseChatOpenAI` in other libraries (e.g., `langchain-deepseek`) and parse out reasoning content there.

This is an area of ongoing development. We may consider adding a base openai-like-compatiblity class for situations like these, or continue to develop dedicated classes (e.g. #34328)

codeonbush:
Thanks for the clear feedback and sharing the design plans! I get keeping langchain-openai lean, and totally align with the dedicated library/base class approach. Appreciate it!

PR #34161, opened 2025-12-01
fix(langchain): (SummarizationMiddleware) adjust token counts based on model

ccurme:


libs/langchain_v1/langchain/agents/middleware/summarization.py
```diff
@@ -119,6 +120,12 @@
 """
 
 
+def _get_approximate_token_counter(model: BaseChatModel) -> TokenCounter:
```
eyurtsev, line 123: Could you add a comment explaining what's going on here? This code isn't self-explanatory. Where is 3.3 from?

PR #34167, opened 2025-12-02
chore(ollama,groq): Filtering Parameters in bind_tools for Ollama and Groq

keenborder786:
- To prevent unsupported parameters from being passed to bind_tools, we now filter them out, which also resolves the error when using ProviderStrategy.
- Fixes #34155


keenborder786:
@ccurme is this looking good?

ccurme (approved):
Regarding Ollama (and possibly Groq): we should add handling for `response_format`, similar to what's done here for Anthropic: https://github.com/langchain-ai/langchain/blob/05ba853548b7a12046ea73041de8cbee51a5ba0a/libs/partners/anthropic/langchain_anthropic/chat_models.py#L2189-L2204

ccurme:
Thanks!

PR #34196, opened 2025-12-03
chore(langchain): qwen support in _init_chat_model_helper

keenborder786:
- Fixes #34183
- Simple Addition of langchain_qwen in init_chat_model
- Related PR in Docs: https://github.com/langchain-ai/docs/pull/1732


TBice123123:
1. There is no langchain-qwen. You probably mean langchain-qwq.
2. langchain-qwq is better suited for use with ChatQwen, as it supports more models, whereas QWQ only supports the QWQ model.
3. langchain-qwq is a third-party library with low update frequency, so it's not really suitable to be included in init_chat_model. The integrations in init_chat_model are generally officially maintained, and langchain-qwq is not one of them.

keenborder786:
Please see: https://docs.langchain.com/oss/python/integrations/chat/qwen

TBice123123:
I wrote this content myself—how could I not know it? Look closely; isn't it langchain-qwq?

TBice123123:
First of all, I greatly appreciate your support for ChatQwen. I made the majority of the contributions to ChatQwen, while the original author @yigit353  primarily contributed to ChatQwQ. However, I believe that since ChatQwen was developed by a third party and is not officially maintained, it shouldn't be included in `init_chat_model`. Moreover, it lacks a JavaScript version, which would lead to inconsistencies between the two language implementations if it were included—a situation I think the official team would discourage.

keenborder786:
@TBice123123 okay, so what do you suggest, should we ask Langchain Team to officially support this package then?

keenborder786:
@ccurme any suggestions?

keenborder786:
@mdrxy any suggestion from you?

ccurme (commented):
LangChain maintainers don't maintain or have access to `langchain-qwen`, so I'd prefer not to add it here for now.

Let me know if this is blocking any usage, you should be able to run
```python
from langchain_qwen import ChatQwQ

model = ChatQwQ(model=model, **kwargs)
```
and use this in `create_agent` or other places.

PR #34197, opened 2025-12-04
fix(langchain): make `init_chat_model` work with huggingface provider

ujjwaltwri:
### Summary

`init_chat_model(model="...", model_provider="huggingface")` currently fails because `ChatHuggingFace` is instantiated with a `model_id` instead of a Hugging Face LLM instance. This leads to a validation error on the `llm` field.

This PR updates the `huggingface` branch of `_init_chat_model_helper` to construct a `HuggingFaceEndpoint` and pass it as `llm` into `ChatHuggingFace`, matching the expectations of `langchain-huggingface`.

### Changes

- For `model_provider="huggingface"`, create a `HuggingFaceEndpoint(repo_id=model, task="text-generation", **kwargs)` and wrap it in `ChatHuggingFace(llm=llm)`.
- Add a unit test for `init_chat_model(..., model_provider="huggingface")` in `test_chat_models.py` that verifies the correct parameters are forwarded to the Hugging Face endpoint.

### Testing

- `pytest libs/langchain_v1/tests/unit_tests/chat_models/test_chat_models.py -k "huggingface"`

> Note: running the full test suite locally with `pytest libs/langchain_v1 -k "init_chat_model"` in my environment raised a `no option named 'only_extended'` error from pytest configuration. The targeted unit test above passes.


ujjwaltwri:
This PR fixes the HuggingFace integration in init_chat_model for langchain_v1. When model_provider="huggingface" is used, the previous implementation passed the model string directly into ChatHuggingFace, which is invalid because the HF integration expects an instantiated HuggingFaceEndpoint as the llm argument. This PR corrects that behavior by constructing the endpoint with the appropriate parameters and forwarding them properly. A targeted unit test was added to verify that repo_id, task, and temperature are forwarded to the endpoint, and the test was updated to skip cleanly when the optional dependency is not installed. Minor mypy adjustments were made to satisfy the valid-type rule for the dynamic subclass used in testing. All formatting, linting, and tests now pass in CI.

ccurme (commented):
Thanks for submitting this. Closing in favor of https://github.com/langchain-ai/langchain/pull/34259 (to also support other HF backends).

PR #34198, opened 2025-12-04
fix(openai): handle missing 'text' key in responses API content blocks

j3r0lin:
Fix KeyError when processing assistant message content blocks in `_construct_responses_api_input` function. The code was directly accessing `block["text"]` without checking if the key exists, which could cause a KeyError when processing malformed or incomplete content blocks.

This change adds defensive checks using `block.get("text")` to safely handle cases where the "text" key might be missing from content blocks with type "text" or "output_text".


mdrxy:
Why would this ever be the case? What usage prompted this?

j3r0lin:
I found that when using the GPT-5.1 model, occasionally the first chunk of its response is [{ "type": "text", index: -1 }], but this happens very rarely.

CleanShot 2025-12-04 at 10 22 20@2x

mdrxy:
Is this using the OpenAI API directly or wrapping via some proxy/router service?

j3r0lin:
OpenAI API directly

ccurme (approved):
Thanks for this. I just wrote up an issue here: https://github.com/langchain-ai/langchain/issues/34311

(This PR band-aids the issue.)

PR #34204, opened 2025-12-04
style(core): fix mypy no-any-return violations

cbornet:
* FIxed where possible
* Used `cast` when not possible to fix

### DIFF
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -358,7 +358,7 @@ class BaseChatModel(BaseLanguageModel[AIMessage], ABC):
 
     @cached_property
     def _serialized(self) -> dict[str, Any]:
-        return dumpd(self)
+        return cast("dict[str, Any]", dumpd(self))
 
     # --- Runnable methods ---
 
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -358,6 +358,9 @@ class BaseChatModel(BaseLanguageModel[AIMessage], ABC):
 
     @cached_property
     def _serialized(self) -> dict[str, Any]:
+        # self is always a Serializable object in this case, thus the result is
+        # guaranteed to be a dict since dumps uses the default callback, which uses
+        # obj.to_json which always returns TypedDict subclasses
         return cast("dict[str, Any]", dumpd(self))
 
     # --- Runnable methods ---
--- a/libs/core/langchain_core/language_models/chat_models.py
+++ b/libs/core/langchain_core/language_models/chat_models.py
@@ -461,7 +461,7 @@ class BaseChatModel(BaseLanguageModel[AIMessage], ABC):
 
         # Check if a runtime streaming flag has been passed in.
         if "stream" in kwargs:
-            return kwargs["stream"]
+            return bool(kwargs["stream"])
 
         if "streaming" in self.model_fields_set:
             streaming_value = getattr(self, "streaming", None)
--- a/libs/core/langchain_core/messages/base.py
+++ b/libs/core/langchain_core/messages/base.py
@@ -302,7 +302,7 @@ class BaseMessage(Serializable):
         from langchain_core.prompts.chat import ChatPromptTemplate  # noqa: PLC0415
 
         prompt = ChatPromptTemplate(messages=[self])
-        return prompt + other
+        return prompt.__add__(other)
 
     def pretty_repr(
         self,
--- a/libs/core/langchain_core/messages/utils.py
+++ b/libs/core/langchain_core/messages/utils.py
@@ -65,9 +65,9 @@ logger = logging.getLogger(__name__)
 def _get_type(v: Any) -> str:
     """Get the type associated with the object for serialization purposes."""
     if isinstance(v, dict) and "type" in v:
-        return v["type"]
+        return cast("str", v["type"])
     if hasattr(v, "type"):
-        return v.type
+        return cast("str", v.type)
     msg = (
         f"Expected either a dictionary with a 'type' key or an object "
         f"with a 'type' attribute. Instead got type {type(v)}."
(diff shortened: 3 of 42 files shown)


libs/core/langchain_core/language_models/chat_models.py
```diff
@@ -461,7 +461,7 @@ def _should_stream(
 
         # Check if a runtime streaming flag has been passed in.
         if "stream" in kwargs:
-            return kwargs["stream"]
+            return bool(kwargs["stream"])
```
cbornet, line 468: calling bool() here doesn't change the behavior as `_should_stream` is always called in a `if` condition.

libs/core/langchain_core/messages/utils.py
```diff
@@ -1019,6 +1020,30 @@ def list_token_counter(messages: Sequence[BaseMessage]) -> int:
     raise ValueError(msg)
 
 
+_SingleMessage = BaseMessage | str | dict[str, Any]
+_T = TypeVar("_T", bound=_SingleMessage)
+# A sequence of _SingleMessage that is NOT a bare str
```
cbornet, line 1125: This is needed because otherwise `str` matches `Sequence[str]`

libs/core/langchain_core/messages/base.py
```diff
@@ -302,7 +302,7 @@ def __add__(self, other: Any) -> ChatPromptTemplate:
         from langchain_core.prompts.chat import ChatPromptTemplate  # noqa: PLC0415
 
         prompt = ChatPromptTemplate(messages=[self])
-        return prompt + other
+        return prompt.__add__(other)
```
cbornet, line 305: mypy doesn't figure the `+` operator...

libs/core/langchain_core/runnables/base.py
```diff
@@ -2417,7 +2418,7 @@ async def _atransform_stream_with_config(
                     iterator = iterator_
                 try:
                     while True:
-                        chunk = await coro_with_context(py_anext(iterator), context)
+                        chunk = await coro_with_context(anext(iterator), context)
```
cbornet, line 2428: we now use python 3.10+ so `py_anext` is not needed anymore

libs/core/langchain_core/runnables/utils.py
```diff
@@ -129,23 +129,38 @@ def asyncio_accepts_context() -> bool:
...
+    return asyncio.create_task(coro)
+
+
+def coro_with_context(coro: Awaitable[_T], context: Context) -> Awaitable[_T]:
     """Await a coroutine with a context.
 
     Args:
         coro: The coroutine to await.
         context: The context to use.
-        create_task: Whether to create a task.
```
mdrxy, line 143: Removing `create_task` is technically breaking (even though it wasn't exposed in `__init__.py`)

mdrxy: Let's add a deprecation warning if `create_task=True` is passed.

libs/core/langchain_core/utils/env.py
```diff
@@ -24,7 +23,7 @@ def env_var_is_set(env_var: str) -> bool:
 
 
 def get_from_dict_or_env(
-    data: dict[str, Any],
```
mdrxy, line 27: I think `values` is `dict[str, Any]` that can contain floats, objects, etc., not `dict[str, str]`.

libs/core/langchain_core/utils/json.py
```diff
@@ -160,7 +160,7 @@ def parse_json_markdown(
 
 
 def _parse_json(
-    json_str: str, *, parser: Callable[[str], Any] = parse_partial_json
+    json_str: str, *, parser: Callable[[str], dict[str, Any]] = parse_partial_json
```
mdrxy: `parse_partial_json` is typed to return `Any`, not `dict[str, Any]`

mdrxy:
^outside of the above, looks good!

libs/core/langchain_core/utils/env.py
cbornet: From its signature `get_from_dict_or_env` must return `str` and we return the value of the `data` dict.
So data should be `dict[str, str]` or else `get_from_dict_or_env` should be changed to return `Any`.

libs/core/langchain_core/utils/json.py
cbornet: Yes but mypy misses it and doesn't complain, so there's no need to cast.

cbornet: The other possibility would be to make `_parse_json`/`parse_json_markdown` return `Any` which would be closer to the truth. And have callers cast or check the type the result. There are callers of `parse_json_markdown` in `core`, `langchain-classic` and `community`.
It seems some callers already expect `parse_json_markdown` to return something different than a `dict` and have `isinstance` checks (eg. `JSONAgentOutputParser.parse`)
LMK what you prefer.

libs/core/langchain_core/utils/env.py
mdrxy: Changing to `dict[str, str]` would be a breaking change. I understand the type inconsistency concern. If we want stricter correctness, we could add a runtime check, but I'd lean toward keeping `dict[str, Any]` for compatibility for now.

libs/core/langchain_core/utils/json.py
mdrxy: Mypy not catching something doesn't make the types correct. I think making both return `Any` is a more honest approach.

libs/core/langchain_core/prompts/string.py
```diff
@@ -330,6 +334,10 @@ async def aformat_prompt(self, **kwargs: Any) -> PromptValue:
         """
         return StringPromptValue(text=await self.aformat(**kwargs))
 
+    @override
+    @abstractmethod
```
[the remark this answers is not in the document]
cbornet: Hi Copilot, you missed that this is an override of the parent method which is already abstract so it doesn't break anything 😉.

libs/core/langchain_core/utils/json.py
cbornet: done

libs/core/langchain_core/utils/env.py
cbornet: I changed to call `str(data[key])` so the signature doesn't break. Can you take a look again ?

mdrxy (approved):
TY!

PR #34149, opened 2025-11-30
feat(langchain,openai): add strict flag to ProviderStrategy structured output

towseef41:
## Summary
- ensure ProviderStrategy propagates the strict flag into provider response_format so strict JSON schema validation can be requested
- document the new strict argument while keeping defaults unchanged
- add a unit test covering default vs strict response_format payloads

Fixes # https://github.com/langchain-ai/langchain/issues/34146

## Testing
- python3 -m uv run --group test pytest tests/unit_tests/agents/test_response_format.py -k provider_strategy_strict_flag
- python3 -m uv run --all-groups mypy langchain/agents/structured_output.py
- make format
- make lint
- make test

Issue #34146, opened 2025-11-29
`ProviderStrategy` does not allow passing `strict=True` into the OpenAI JSON Schema `response_format`


hjerbii:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Package (Required)

- [x] langchain
- [ ] langchain-openai
- [ ] langchain-anthropic
- [ ] langchain-classic
- [ ] langchain-core
- [ ] langchain-cli
- [ ] langchain-model-profiles
- [ ] langchain-tests
- [ ] langchain-text-splitters
- [ ] langchain-chroma
- [ ] langchain-deepseek
- [ ] langchain-exa
- [ ] langchain-fireworks
- [ ] langchain-groq
- [ ] langchain-huggingface
- [ ] langchain-mistralai
- [ ] langchain-nomic
- [ ] langchain-ollama
- [ ] langchain-perplexity
- [ ] langchain-prompty
- [ ] langchain-qdrant
- [ ] langchain-xai
- [ ] Other / not sure / general

### Example Code (Python)

```python
from langchain.agents import create_agent
from langchain.agents.structured_output import ProviderStrategy
from langchain_openai import ChatOpenAI
from pydantic import BaseModel

class ResponseFormat(BaseModel):
    status: str
    message: str

llm = ChatOpenAI(model="gpt-4o")

agent = create_agent(
    model=llm,
    tools=[],
    system_prompt="Test",
    response_format=ProviderStrategy(ResponseFormat),
)

agent.invoke({"messages": [{"role": "user", "content": "hi"}]})
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

* I am using langchain 1.1.0
* When using `create_agent(..., response_format=ProviderStrategy(MySchema))`, LangChain generates a response_format payload for OpenAI with (lancghain/agents/structured_output.py):

<img width="860" height="845" alt="Image" src="https://github.com/user-attachments/assets/71fba321-3f49-4556-83e6-b2746f419b0d" />

[...]

PR #34205, opened 2025-12-04
fix(openai): add strict flag to correctly use structured-output of openai provider

thdesc:
- Ensure `ChatOpenAI.bind_tools` forwards its `strict` flag into `_convert_to_openai_response_format(...)`, so the resulting `response_format` includes OpenAI’s `"strict": true` setting when requested.
 - With this change, `langchain.agents.create_agent` can now drive OpenAI’s Structured Output mode purely by providing a Pydantic `response_format`: the factory already passes `strict=True` when it selects `ProviderStrategy` ([see here](https://github.com/langchain-ai/langchain/blob/b95cb770e8a52bd53d787562fcb999d300f8de9b/libs/langchain_v1/langchain/agents/factory.py#L1050)) and that flag now propagates all the way to the OpenAI API.
 
 
 Code to reproduce:
```python
from pydantic import BaseModel, Field, ConfigDict

class ProductRating(BaseModel):
    rating: int = Field(..., description="Rating from 1-5", ge=1, le=5)
    comment: str = Field(description="Review comment")
    model_config = ConfigDict(extra="forbid")

agent = create_agent(
    model="gpt-5-mini",
    tools=[],
    response_format=ProductRating,
    system_prompt="You are a helpful assistant that parses product reviews. Do not make any field or value up.",
)

response = agent.invoke({
    "messages": [{"role": "user", "content": "Toto"}]
})
```
This could raise an error because the structured output is not forced.

With this PR, the structured output is correctly set and no error will be raised with the provided code.


thdesc:
This PR could solve this issue #34146.

ccurme (commented):
Hi, thanks for submitting this!

We can do this without a breaking change (turning on strict mode for users when it was previously disabled) by supporting
```
ProviderStrategy(schema, strict=True)
```
as is done here: https://github.com/langchain-ai/langchain/pull/34149

Closing in favor of that approach. We'll plan to release this in langchain 1.2.

PR #34208, opened 2025-12-04
fix(core): `RunnablePick` method return types

cbornet:
Following https://github.com/langchain-ai/langchain/pull/31321, the `Output` type of `RunnablePick` is `Any`.


mdrxy (approved):
TY!

PR #34211, opened 2025-12-04
chore(core): use anext and deprecate py_anext

cbornet:
LangChain uses Python 3.10+ so `py_anext` isn't needed anymore.

### DIFF
--- a/libs/core/langchain_core/utils/aiter.py
+++ b/libs/core/langchain_core/utils/aiter.py
@@ -34,7 +34,7 @@ _no_default = object()
 
 
 # https://github.com/python/cpython/blob/main/Lib/test/test_asyncgen.py#L54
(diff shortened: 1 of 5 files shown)


libs/core/langchain_core/utils/aiter.py
```diff
@@ -26,13 +26,15 @@
...
+from langchain_core._api.deprecation import deprecated
+
 T = TypeVar("T")
 
 _no_default = object()
 
 
 # https://github.com/python/cpython/blob/main/Lib/test/test_asyncgen.py#L54
-# before 3.10, the builtin anext() was not available
+@deprecated(since="1.1.0", removal="2.0.0")
```
mdrxy:
```suggestion
@deprecated(since="1.1.2", removal="2.0.0")
```

PR #34225, opened 2025-12-05
chore(core): improve typing of messages utils functions

cbornet:
With this we get the correct types for `_runnable_support` annotated functions.
* return list[BaseMessage] when messages is not None
* return Runnable when messages is None
* typing of function args

### DIFF
--- a/libs/core/langchain_core/messages/utils.py
+++ b/libs/core/langchain_core/messages/utils.py
@@ -15,12 +15,16 @@ import json
 import logging
 import math
 from collections.abc import Callable, Iterable, Sequence
-from functools import partial
+from functools import partial, wraps
 from typing import (
     TYPE_CHECKING,
     Annotated,
     Any,
+    Concatenate,
     Literal,
+    ParamSpec,
+    Protocol,
+    TypeVar,
     cast,
     overload,
 )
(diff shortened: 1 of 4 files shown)


libs/core/langchain_core/messages/utils.py
```diff
@@ -384,36 +387,64 @@ def convert_to_messages(
...
+    exclude_types: Sequence[str | type[BaseMessage]] | None = None,
+    include_ids: Sequence[str] | None = None,
+    exclude_ids: Sequence[str] | None = None,
+    exclude_tool_calls: Sequence[str] | bool | None = None,
+) -> RunnableLambda[
+    Iterable[MessageLikeRepresentation] | PromptValue, list[BaseMessage]
+]: ...
+
+
+@_runnable_support  # type: ignore[misc]
```
cbornet: The overloads are correct but don't correspond to the impl because of the decorator.

libs/core/tests/unit_tests/messages/test_utils.py
```diff
@@ -192,7 +202,7 @@ def test_filter_message_exclude_tool_calls() -> None:
     assert expected == actual
 
     # test explicitly excluding all tool calls
-    actual = filter_messages(messages, exclude_tool_calls={"1", "2"})
+    actual = filter_messages(messages, exclude_tool_calls=["1", "2"])
```
cbornet, line 205: It found some incorrect typing 😃

libs/core/langchain_core/messages/utils.py
```diff
@@ -384,36 +387,64 @@ def convert_to_messages(
     return [_convert_to_message(m) for m in messages]
 
 
-def _runnable_support(func: Callable) -> Callable:
-    @overload
-    def wrapped(
```
cbornet: Unfortunately decorator signatures don't propagate to wrapped function.

libs/core/tests/unit_tests/messages/test_utils.py
```diff
@@ -508,13 +518,14 @@ def test_trim_messages_invoke() -> None:
...
-        max_tokens=10, token_counter=FakeTokenCountingModel().bind(foo="bar")
+        max_tokens=10,
+        token_counter=FakeTokenCountingModel().bind(foo="bar"),  # type: ignore[call-overload]
     )
     trimmer.invoke([HumanMessage("foobar")])
 
 
 def test_trim_messages_bad_token_counter() -> None:
-    trimmer = trim_messages(max_tokens=10, token_counter={})
+    trimmer = trim_messages(max_tokens=10, token_counter={})  # type: ignore[call-overload]
```
cbornet, line 528: There is now typing for `_runnable_support` annotated method args.

PR #34229, opened 2025-12-06
fix(xai): Fix 400 error for AIMessages with tool_calls but no content

### LINKED ISSUE #34140: ChatXAI: 400 error when AIMessage has tool_calls but no content
## Description

When using `ChatXAI` with LangGraph agents, xAI's API returns a 400 error if an `AIMessage` contains `tool_calls` but has empty/null `content`:

```
openai.BadRequestError: Error code: 400 - {'code': 'Client specified an invalid argument', 
'error': 'Invalid request content: Each message must have at least one content element.'}
```

## Environment

- `langchain-xai`: 1.1.0
- `langchain-core`: 1.1.0
- Model: `grok-4-1-fast-reasoning`
- Using LangGraph with tool-calling agents

## Reproduction

This occurs when using ChatXAI in a LangGraph agent that makes tool calls. After several tool call rounds, the message history contains AIMessages with only `tool_calls` and no `content` field. xAI's API is stricter than OpenAI's and rejects these messages.

## Workaround

Subclass ChatXAI and ensure content is always set:

```python
from langchain_xai import ChatXAI
from langchain_core.messages import AIMessage

class ChatXAISanitized(ChatXAI):
    def _convert_messages_to_dicts(self, messages):
        for msg in messages:
            if isinstance(msg, AIMessage) and not msg.content and msg.tool_calls:
                msg.content = ""  # xAI requires content, even if empty
        return super()._convert_messages_to_dicts(messages)
```

## Suggested Fix

`ChatXAI._convert_messages_to_dicts` (or the underlying message conversion) should ensure all messages have at least an empty content field to satisfy xAI's API requirements.

arrdel:
Fixes #34140

xAI API requires all messages to have at least an empty content field. When using LangGraph agents with tool calling, AIMessage objects can have tool_calls but no content, causing 400 errors.

This fix ensures content is set to empty string instead of None for AIMessages with tool_calls.

### DIFF
--- a/libs/partners/xai/langchain_xai/chat_models.py
+++ b/libs/partners/xai/langchain_xai/chat_models.py
@@ -5,16 +5,18 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Literal, TypeAlias, cast
 
 import openai
-from pydantic import BaseModel, ConfigDict, Field, SecretStr, model_validator
-from typing_extensions import Self
-
 from langchain_core.messages import AIMessage, AIMessageChunk, BaseMessage
 from langchain_core.utils import secret_from_env
 from langchain_openai.chat_models.base import (
     BaseChatOpenAI,
     _convert_from_v1_to_chat_completions,
+)
+from langchain_openai.chat_models.base import (
     _convert_message_to_dict as _openai_convert_message_to_dict,
 )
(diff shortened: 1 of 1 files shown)


ccurme (commented):
Thanks for this, left a comment on the parent issue.

PR #34279, opened 2025-12-10
feat(anthropic): accept `TypedDict` for built-in tool types

mdrxy:
Widen `bind_tools` to accept `TypedDict` via `Mapping` so that users may import and use Anthropic's built-in tool types:

```python
import subprocess

from anthropic.types.beta import BetaToolBash20250124Param
from langchain.tools import tool

tool_spec = BetaToolBash20250124Param(
    name="bash",
    type="bash_20250124",
    strict=True,
)

@tool(extras={"provider_tool_definition": tool_spec})
def bash(*, command: str, restart: bool = False, **kw):
    """Execute a bash command."""
    if restart:
        return "Bash session restarted"
    try:
        result = subprocess.run(
            command,
            shell=True,
            capture_output=True,
            text=True,
            timeout=30,
        )
        return result.stdout + result.stderr
    except Exception as e:
        return f"Error: {e}"

# Bind bash tool to your model
```


libs/partners/anthropic/langchain_anthropic/tools/code_execution.py
```diff
@@ -0,0 +1,83 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from anthropic.types.beta import (
+        BetaCacheControlEphemeralParam,
+        BetaCodeExecutionTool20250825Param,
+    )
+
+
+def code_execution_20250825(
```
ccurme: I'm not sure I see the point of maintaining / documenting these constructors. is it possible to just allow / document passing in `BetaCodeExecutionTool20250825Param` directly?

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -1520,83 +1542,98 @@ class Joke(BaseModel):
...
-                        "enabled": True,
-                        "allowed_tools": ["ask_question"],
-                    },
-                    "authorization_token": "PLACEHOLDER",  # optional authorization
+                    "url": "https://docs.langchain.com/mcp",
+                    "name": "LangChain Docs",
+                    # "tool_configuration": {  # optional configuration
+                    #     "enabled": True,
+                    #     "allowed_tools": ["ask_question"],
+                    # },
```
ccurme: out of scope but I'm not sure the new MCP connector supports this

libs/partners/anthropic/langchain_anthropic/tools/types.py
```diff
@@ -0,0 +1,475 @@
...
+from __future__ import annotations
+
+from typing import Literal, TypedDict
+
+# =============================================================================
+# Memory Tool Command Types
+# =============================================================================
+
+
+class MemoryViewCommand(TypedDict):
```
ccurme: ~How do these get used?~ (saw the next PR)

Here too I'm hesitant to commit to maintaining these.
- If Anthropic maintains them in their SDK, can we leverage that?
- If they do not (e.g., they seem to just [document how to implement](https://platform.claude.com/docs/en/agents-and-tools/tool-use/memory-tool)), then trade-offs seem unfavorable. We might find ourselves blocking use because our types get stale.

Open to discussing!

libs/partners/anthropic/langchain_anthropic/tools/code_execution.py
mdrxy: my initial thoughts before implementing was that I agree; however I think the typing / autocomplete / devx benefits are valuable here. Also, since tools are versioned, I'm not as concerned about upstream breaking changes. I think the maintenance burden is low.

libs/partners/anthropic/langchain_anthropic/tools/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+    import subprocess
+
+    from langchain_anthropic import ChatAnthropic, tools
+    from langchain.messages import HumanMessage, ToolMessage
+
+
+    def execute_bash(*, command: str | None = None, restart: bool = False, **kw):
+        if restart:
+            return "Bash session restarted"
+        result = subprocess.run(
```
ccurme: if this is the example we document we should thoroughly warn / caveat. might make sense to use a less dangerous example.

libs/partners/anthropic/langchain_anthropic/tools/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+    # Initial request
+    response = model_with_bash.invoke([query])
+
+    # Process tool calls in a loop until no more tool calls
+    messages = [
+        query,
+        response,
+    ]
+
+    while response.tool_calls:
```
ccurme: I don't think we need to document custom tool call loops here

libs/partners/anthropic/langchain_anthropic/tools/__init__.py
```diff
@@ -0,0 +1,174 @@
...
+        # Execute each tool call
+        for tool_call in response.tool_calls:
+            # Invoke the tool with the args from the model
+            result = bash_tool.invoke(tool_call["args"])
+
+            # Add the tool result to messages
+            messages.append(ToolMessage(content=result, tool_call_id=tool_call["id"]))
```
ccurme: `bash_tool.invoke(tool_call)` gives a ToolMessage directly

libs/partners/anthropic/langchain_anthropic/tools/code_execution.py
ccurme: if you have any prompts / workflows so that when the next tool drops the update is mechanical, that would be useful

libs/partners/anthropic/langchain_anthropic/chat_models.py
```diff
@@ -3121,8 +3131,22 @@ def convert_to_anthropic_tool(
...
     """
+    # Check for provider tool definition in extras (client-side execution tools).
+    # Built-in tools (bash, computer, text_editor, memory) store their Anthropic-
+    # specific definition in extras. These have a different shape (type + name)
+    # than regular tools (name + input_schema), hence AnthropicBuiltinTool.
+    if (
+        isinstance(tool, BaseTool)
+        and hasattr(tool, "extras")
+        and isinstance(tool.extras, dict)
+        and "provider_tool_definition" in tool.extras
```
ccurme, line 3161: (nit) isn't `tool` implicit

mdrxy: it's not impossible for a `type` or `Callable` to have `extras`

this is a guard against that

PR #34283, opened 2025-12-10
chore(core): fix some types related to ToolCallChunk

cbornet:


libs/core/langchain_core/messages/block_translators/anthropic.py
```diff
@@ -245,11 +246,16 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
                 ):
                     # Isolated chunk
-                    tool_call_chunk: types.ToolCallChunk = (
-                        message.tool_call_chunks[0].copy()  # type: ignore[assignment]
+                    chunk = message.tool_call_chunks[0]
+                    tool_call_chunk = ToolCallChunk(
+                        name=chunk.get("name"),
+                        id=chunk.get("id"),
+                        args=chunk.get("args"),
+                        type="tool_call_chunk",
```
cbornet, line 254: small behavior change as it forces type to `tool_call_chunk `. Is it OK ?
Before, if `message.tool_call_chunks[0]` had a different type, that type was used in the resulting `ToolCallChunk`, breaking typing constraint.

libs/core/langchain_core/messages/block_translators/anthropic.py
```diff
@@ -245,11 +246,16 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
                     )
-                    if "type" not in tool_call_chunk:
-                        tool_call_chunk["type"] = "tool_call_chunk"
+                    index = chunk.get("index")
+                    if index is not None:
```
cbornet, line 260: small behavior change as we don't set  `index` if it's None . Is it OK ?
Before, if `index` was `None`, it was set in the resulting `ToolCallChunk`, breaking typing constraint.

libs/core/langchain_core/messages/block_translators/anthropic.py
```diff
@@ -245,17 +245,20 @@ def _iter_blocks() -> Iterable[types.ContentBlock]:
...
+                        type="tool_call_chunk",
                     )
                     if "caller" in block:
-                        if "extras" not in tool_call_chunk:
-                            tool_call_chunk["extras"] = {}
-                        tool_call_chunk["extras"]["caller"] = block["caller"]
-
-                    if "type" not in tool_call_chunk:
-                        tool_call_chunk["type"] = "tool_call_chunk"
+                        tool_call_chunk["extras"] = {"caller": block["caller"]}
```
cbornet, line 257: small behavior change. We don't check if `message.tool_call_chunks[0]` has an `extras` key since from its typing definition is hasn't. I also checked in code that nothing sets an `extras` key on the message tool_call_chunks. Is it OK ?

cbornet:
cc @mdrxy since I see that you worked on ToolCallChunk recently.

ccurme (approved):
Thanks!

PR #34313, opened 2025-12-12
feat(openai): support OpenRouter reasoning_details in AIMessage

xiangnuans:
Add support for OpenRouter's `reasoning_details` field as provider-specific metadata stored in `additional_kwargs`. This allows OpenRouter users to access structured reasoning metadata (tokens, steps, etc.) without modifying the message content structure.

## Changes

- Store `reasoning_details` in `additional_kwargs` for both non-streaming and streaming scenarios
- Handle `reasoning_details` at response level in `_create_chat_result`
- Add comprehensive tests including streaming delta merging verification
- Rely on LangChain's `merge_dicts()` for automatic delta merging when chunks are combined via `AIMessageChunk.__add__()`

This follows the same pattern as other provider-specific fields and maintains backward compatibility.

## Type

🆕 New Feature

## Testing

- [x] Ran `make format` - All checks passed
- [x] Ran `make lint` - All checks passed  
- [x] Ran `make test` - All tests pass
- [x] Tests cover non-streaming, streaming, and delta merging scenarios

## Additional Notes

**AI Contribution Disclaimer**: This contribution was created with the assistance of AI agents. The implementation follows LangChain's architecture patterns and has been thoroughly tested.

The implementation relies on LangChain's existing `merge_dicts()` mechanism for streaming delta merging, which automatically handles nested dict/list merging when chunks are combined via `AIMessageChunk.__add__()`. This ensures that multiple streaming deltas with `reasoning_details` are properly merged without requiring custom merge logic.


mdrxy:
Thanks. We are not not planning to overload this integration meant for OpenAI, but rather will release a dedicated integration for OpenRouter as they have diverged from the chat completions standard. See #34328 for tracking

xiangnuans:
Thanks for the clarification — that makes sense 👍

I understand the concern about not overloading the OpenAI integration, especially given how OpenRouter has diverged from the Chat Completions standard.

I’ll take this implementation as a basis for a dedicated OpenRouter integration, aligned with the direction outlined in #34328.
The current PR already treats reasoning_details as provider-specific metadata and relies on LangChain’s existing streaming merge semantics, so it should translate cleanly into a standalone integration.

I’ll follow up on #34328 with a proposal or initial implementation once the expected structure for the OpenRouter package is clearer.
Thanks for the guidance!

PR #34318, opened 2025-12-12
fix(core): prevent inherited docstrings for class-based tools

PrinceThummar011:
## Description
This PR addresses issue #32066 by preventing class-based tools from automatically inheriting docstrings from their parent classes.

Previously, if a child tool class did not define its own docstring, it would inherit one from its parent `BaseModel` or base tool class. This resulted in generic or incorrect descriptions being passed to LLMs (e.g., "Base class for tools..."), confusing the model.

## Changes
- Updated `_infer_arg_descriptions` in `libs/core/langchain_core/tools/base.py` to check the class `__dict__` for a docstring.
- Updated `StructuredTool.from_function` in `libs/core/langchain_core/tools/structured.py` to default to an empty string if the docstring is missing, rather than falling back to the parent.

## Notes
This PR supersedes and fixes the approach attempted in PR #33540. It implements the logic without removing existing comments or code structure, ensuring a cleaner diff and avoiding the issues flagged in the previous review.

Fixes #32066


sydney-runkle (changes requested):
Could you please add tests confirming functionality?

PR #34236, opened 2025-12-07
fix(langchain): reasoning_content/reasoning miss error #34153 issue

csuwl:
1. fix(langchain): fix #34153 issue :reasoning_content/reasoning miss error

2. PR description:
This submission is to fix this issue #34153 #34166.
When using Deepseek or OpenAI, the agent reports an error: indicating that the request is missing 'reasnoning_content'.
I added parameter 'reasoning_content'  in '_convert_message_to_dict' and '_convert_dict_to_message' function. Tested on Deepseek, agent worked well.

Issue #34328, opened 2025-12-12
OpenRouter / LiteLLM / OpenAI API compatibility across providers / proxies


mdrxy:
### Privileged issue

- [x] I am a LangChain maintainer, or was asked directly by a LangChain maintainer to create an issue here.

### Issue Content

Tracking for upcoming OpenRouter integration.

It is not practical for `ChatOpenAI` to handle each addition that third party providers make to the Chat Completions standard.

## Related issues

#30530

- User is attempting to access a LiteLLM based proxy via `ChatOpenAI`. `reasoning_content` and `thinking_blocks` fields aren't captured by `ChatOpenAI`.
- LiteLLM adds thinking/reasoning on top of the Chat Completions API
- We recommended to use [`langchain-litellm`](https://github.com/Akshay-Dongare/langchain-litellm/)

#31325

- User is attempting to access OpenRouter via `ChatAnthropic`
- OpenRouter does not support Anthropic's Messages API. OpenRouter's API is a fork of the OpenAI Chat Completions API.
- Recommended to use `ChatOpenAI` with warning.

#32967

- User is attempting to access OpenRouter via `ChatOpenAI`
- Structured output via `create_agent` fails because OpenRouter doesn't support the OpenAI SDK's `beta.chat.completions.parse` method

#32977

- User is attempting to access OpenRouter via `ChatOpenAI`
- `with_structured_output` fails with `JSONDecodeError` when OpenRouter returns malformed/incomplete JSON. Even with `include_raw=True`, the raw response is inaccessible on error.

#32981

- User is attempting to access OpenRouter via `ChatOpenAI`
- Reasoning content is not being captured by `ChatOpenAI`
  - Commenter states `ChatDeepSeek` is also not working

#33643

- User requests native LiteLLM and OpenRouter integrations in LangChain.

#33672

- User wants `ChatOpenAI` to support extracting Qwen3's thinking/reasoning process (like it does for DeepSeek). `reasoning_content` is not captured. `ChatTongyi` doesn't allow custom `base_url` for local models. The official OpenAI SDK works fine for this.

#33757

- User wants to use OpenRouter via `ChatOpenAI` to support `cache_control` (e.g. for Anthropic models)

#33834

- User is attempting to access models from Volcengine via `ChatOpenAI` foro Doubao models (`doubao-seed-1-6-251015`)
  - Volcengine may support Responses API

#34056

[...]

PR #34347, opened 2025-12-13
fix(openai): handle streaming reasoning tokens.

jimfcarroll:
This adds the ability to handle reasoning tokens returned from Deepseek like and openai reasoning models. It complements this PR: https://github.com/langchain-ai/langchain/pull/34236

I ran this against llama.cpp running a DeepSeek variant and I'm parsing the results using a CallbackHandler like this:

```python
   # Called once per token (LangChain-level)
    async def on_llm_new_token(self, token : str| list | None, **kwargs):
        if token:
            if isinstance(token, str):
                if self.mode != 2:
                    print(flush=True)
                    print("RESPONDING", flush=True)
                    self.mode = 2
                self.tokens.append(token)
                print(token, end="", flush=True)
            elif isinstance(token, list):
                reasoning : str | None = next((d['reasoning'] for d in token if 'reasoning' in d), None)
                if reasoning:
                    if self.mode != 1:
                        print(flush=True)
                        print("REASONING", flush=True)
                        self.mode=1
                    print(reasoning, end="", flush=True)
```

If someone points me to an appropriate unit test to extend that works in a similar area, I'll update it to check this explicitly.

### DIFF
--- a/libs/partners/openai/langchain_openai/chat_models/base.py
+++ b/libs/partners/openai/langchain_openai/chat_models/base.py
@@ -77,6 +77,7 @@ from langchain_core.messages.block_translators.openai import (
     _convert_from_v03_ai_message,
     convert_to_openai_data_block,
 )
+from langchain_core.messages.content import ReasoningContentBlock
 from langchain_core.messages.tool import tool_call_chunk
 from langchain_core.output_parsers import JsonOutputParser, PydanticOutputParser
 from langchain_core.output_parsers.openai_tools import (
@@ -372,6 +373,17 @@ def _convert_delta_to_message_chunk(
         if "name" in function_call and function_call["name"] is None:
             function_call["name"] = ""
         additional_kwargs["function_call"] = function_call
+
+    reasoning_content_block : ReasoningContentBlock | None = None
+    if _dict.get("reasoning_content"):
+        reasoning_content_block = ReasoningContentBlock(reasoning=cast(
+            str, _dict.get("reasoning_content") or ""
+        ))
+    elif _dict.get("reasoning"):
+        reasoning_content_block = ReasoningContentBlock(reasoning=cast(
+            str, _dict.get("reasoning") or ""
+        ))
+
     tool_call_chunks = []
     if raw_tool_calls := _dict.get("tool_calls"):
         try:
@@ -392,6 +404,7 @@ def _convert_delta_to_message_chunk(
     if role == "assistant" or default_class == AIMessageChunk:
         return AIMessageChunk(
             content=content,
+            content_blocks=[ reasoning_content_block ] if reasoning_content_block else None,
             additional_kwargs=additional_kwargs,
             id=id_,
             tool_call_chunks=tool_call_chunks,  # type: ignore[arg-type]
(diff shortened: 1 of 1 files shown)


jimfcarroll:
This linting error prevents the ability to splat parameters:

```
[ "langchain_openai" = "" ] || mkdir -p .mypy_cache && uv run --all-groups mypy langchain_openai --cache-dir .mypy_cache
langchain_openai/chat_models/base.py:404: error: Argument 5 to "AIMessageChunk" has incompatible type "**dict[str, str]"; expected "dict[Any, Any]"  [arg-type]
langchain_openai/chat_models/base.py:404: error: Argument 5 to "AIMessageChunk" has incompatible type "**dict[str, str]"; expected "Literal['AIMessageChunk']"  [arg-type]
langchain_openai/chat_models/base.py:404: error: Argument 5 to "AIMessageChunk" has incompatible type "**dict[str, str]"; expected "list[ToolCall]"  [arg-type]
langchain_openai/chat_models/base.py:404: error: Argument 5 to "AIMessageChunk" has incompatible type "**dict[str, str]"; expected "list[InvalidToolCall]"  [arg-type]
langchain_openai/chat_models/base.py:404: error: Argument 5 to "AIMessageChunk" has incompatible type "**dict[str, str]"; expected "UsageMetadata | None"  [arg-type]
langchain_openai/chat_models/base.py:404: error: Argument 5 to "AIMessageChunk" has incompatible type "**dict[str, str]"; expected "Literal['last'] | None"  [arg-type]
```

This means each combination of conditionals would need to explicitly result in a different call to the AIMessageChunk creation rather than be able to incrementally create the parameter list.

jimfcarroll:
Okay. Looking at the code more deeply I think I'm now in alignment with how reasoning tokens are meant to be added to an AIMessageChunk.

jimfcarroll:
Okay. Now I'm getting this:

```
 langchain_openai/chat_models/base.py:405: error: Unexpected keyword argument "content_blocks" for "AIMessageChunk"  [call-arg]
 ```

Even though it's obvious AIMessageChunk.__init__ is meant to take content_blocks as a parameter.

This code is a mess. I can't add the parameter to pass content_blocks. However, content_blocks can be derived from additional_kwargs in the default, but not correctly for openai which expects content blocks in the content field, and even that doesn't work correctly for the following reasons:

When the AIMessageChunk.response_metadata["model_provider"] is "openai" it uses the translator from langchain_core.messages.block_translators.opeanai.translate_content_chunk which looks like this:

```python
def translate_content_chunk(message: AIMessageChunk) -> list[types.ContentBlock]:
    """Derive standard content blocks from a message chunk with OpenAI content."""
    if isinstance(message.content, str):
        return _convert_to_v1_from_chat_completions_chunk(message)
    message = _convert_from_v03_ai_message(message)  # type: ignore[assignment]
    return _convert_to_v1_from_responses(message)
```

The gating of the call to _convert_to_v1_from_chat_completions_chunk(message) only when the message.content is a string prevents the call when you set message.content to a list of content blocks BUT the code in _convert_to_v1_from_chat_completions_chunk(message) can handle it either way.

What is the intention here? There are two confusing paths both 1/2 implemented and they conflict with each other:

1. use additional_kwargs and then change the openai content_block translate to pick them off of there the same way the default BaseMessage does.
2. Add specific content blocks like ReasoningContentBlocks to the content field of AIMessageChunk and fix the translate to handle that.

My personal preference would be (2) for two reasons, it seems obvious AIMessage[Chunk].content could be a list[dict]. And it automatically ends up on "token" correctly in the callback handler method on_llm_new_token.

mdrxy:
#34328

PR #34351, opened 2025-12-14
fix(langchain): Explicitly disable callbacks in `LLMToolSelectorMiddleware`

Manju080:
Fixes #34139

When using `LLMToolSelectorMiddleware` with stream_mode="messages", the internal tool selection tokens (the JSON output) are leaked into the final user stream. This happens because the internal LLM call inherits the parent's callback manager, which streams every token generated, even those intended only for middleware logic.

This PR introduces a fix to prevent the leakage:

1. Added `disable_streaming` parameter to the middleware init (defaults to `True`) to give users control.
2. Explicitly cleared Callbacks, instead of relying on model.bind(streaming= `False`) which is not respected by all model providers (e.g ChatOllama still leaks callback events) this implementation passes config={"callbacks": []} to the internal invoke and ainvoke calls.

I verified this locally using a reproduction script with ChatOllama (llama3.2).
Before: Output contained mixed internal JSON tokens `({"tools": ...})`.
After: Output is clean and contains only the final response tokens.

### DIFF
--- a/libs/langchain_v1/langchain/agents/middleware/tool_selection.py
+++ b/libs/langchain_v1/langchain/agents/middleware/tool_selection.py
@@ -140,6 +140,10 @@ class LLMToolSelectorMiddleware(AgentMiddleware):
             always_include: Tool names to always include regardless of selection.
 
                 These do not count against the `max_tools` limit.
+            disable_streaming: Whether to disable streaming for the tool selection.
+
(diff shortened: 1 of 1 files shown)


Manju080:
@sydney-runkle could you please look into this PR and let me know about any changes and updates.

PR #34368, opened 2025-12-15
feat(langchain): llm tool selector keyerror 

riunyfir:
Description: 
`LLMToolSelectorMiddleware` raises `KeyError: 'tools'` when the model response is missing the expected `'tools'` key. This occurs when the LLM doesn't follow the structured output schema, especially with complex or edge-case prompts. The error interrupts downstream processing and requires patching the middleware. 
Solution:
Added defensive checks in `_process_selection_response` to handle malformed responses:

1. **Missing 'tools' key**: Check if `'tools'` exists before accessing it
2. **Invalid 'tools' type**: Validate that `'tools'` is a list type
3. **Graceful fallback**: When response is malformed, fallback to using all available tools (respecting `max_tools` limit if set) instead of crashing
4. **Warning logs**: Log warnings when fallback occurs to aid debugging

The middleware now handles edge cases without crashing, improving resilience to model output variations.
Changes

- Modified `_process_selection_response` in `tool_selection.py` to add defensive checks
- Added comprehensive test cases for missing 'tools' key and invalid type scenarios
- Added warning logging for debugging malformed responses

### Areas Requiring Careful Review

1. **Fallback behavior**: The current implementation falls back to using all available tools when response is malformed. Consider:
   - Is this the desired behavior, or should we raise an exception?
   - Should we respect `max_tools` limit in fallback (currently yes)?
   - Should we still include `always_include` tools in fallback (currently yes)?

2. **Error handling strategy**: Currently we log warnings and continue. Consider:
   - Should we raise a more specific exception instead?
   - Should we provide a configuration option to control fallback vs. exception behavior?

3. **Test coverage**: Added tests for missing key and invalid type, but consider:
   - Should we test with empty 'tools' list?
   - Should we test with `always_include` tools in fallback scenario?
 Disclaimer
This contribution was created with assistance from AI agents. The changes have been reviewed and tested, but please pay special attention to the fallback logic and error handling strategy during code review.
Issue: Fixes : #34358 
Dependencies: None

### DIFF
--- a/libs/langchain_v1/langchain/agents/factory.py
+++ b/libs/langchain_v1/langchain/agents/factory.py
@@ -20,9 +20,9 @@ from langgraph._internal._runnable import RunnableCallable
 from langgraph.constants import END, START
 from langgraph.graph.state import StateGraph
 from langgraph.prebuilt.tool_node import ToolCallWithContext, ToolNode
-from langgraph.runtime import Runtime
+from langgraph.runtime import Runtime  # noqa: TC002
 from langgraph.types import Command, Send
-from langgraph.typing import ContextT
+from langgraph.typing import ContextT  # noqa: TC002
 from typing_extensions import NotRequired, Required, TypedDict
 
 from langchain.agents.middleware.types import (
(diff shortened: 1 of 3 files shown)


ccurme (commented):
Duplicated with https://github.com/langchain-ai/langchain/pull/34369.

PR #34369, opened 2025-12-15
fix: handle missing `tools` key in `LLMToolSelectorMiddleware` response

riunyfir:
Description:
This PR fixes a `KeyError: 'tools'` that occurs in `LLMToolSelectorMiddleware` when the LLM returns a response missing the expected `'tools'` key. The middleware now gracefully handles malformed or incomplete responses by falling back to using all available tools (respecting the `max_tools` limit) instead of raising an exception.
 Problem：

When using `LLMToolSelectorMiddleware` with an agent, the middleware calls an LLM to select relevant tools before the main model invocation. However, if the LLM response doesn't strictly follow the structured output schema (e.g., missing the `'tools'` key or having an invalid type), the code would raise a `KeyError` at line 229 in `tool_selection.py`, interrupting all downstream processing.

This issue occurs intermittently, especially with complex prompts or edge cases where the LLM doesn't strictly adhere to the expected schema format.

Solution:
The fix adds defensive checks in `_process_selection_response()` to handle three scenarios:

1. **Missing 'tools' key**: When the response dictionary doesn't contain a `'tools'` key, the middleware logs a warning and falls back to using all available tools (up to `max_tools` limit).

2. **Invalid 'tools' type**: When the `'tools'` key exists but is not a list, the middleware logs a warning and falls back to using all available tools.

3. **Normal case**: When the response is valid, the existing logic continues to work as before.

This approach ensures backward compatibility while making the middleware more resilient to LLM response variations. The fallback behavior (using all available tools) is reasonable because:
- It maintains functionality even when tool selection fails
- The `max_tools` limit is still respected
- `always_include` tools are still added as expected
- Users get a warning in logs to help debug LLM response issues

Changes

- Modified `_process_selection_response()` in `libs/langchain_v1/langchain/agents/middleware/tool_selection.py` to check for missing or invalid `'tools'` key before accessing it
- Added warning logs when malformed responses are detected
- Added two new test cases to verify the fix handles missing and invalid `'tools'` keys correctly
Issue: Fixes : #34358 
Dependencies: None

### DIFF
--- a/libs/langchain_v1/langchain/agents/middleware/tool_selection.py
+++ b/libs/langchain_v1/langchain/agents/middleware/tool_selection.py
@@ -235,8 +235,8 @@ class LLMToolSelectorMiddleware(AgentMiddleware):
             tools_list = valid_tool_names
         elif not isinstance(response["tools"], list):
             logger.warning(
-                f"Model response 'tools' key is not a list (got {type(response['tools'])}). "
-                "Using all available tools for selection."
+                "Model response 'tools' key is not a list (got %s). Using all available tools for selection.",
+                type(response["tools"]),
             )
             tools_list = valid_tool_names
         else:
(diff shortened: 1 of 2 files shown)


mdrxy:
is this the same as #34368?

ccurme (commented):
Commented on the issue. This is reasonable but I think we want to retry the structured model invocation instead of degrading performance with a warning.

PR #34375, opened 2025-12-15
fix(langchain): handle malformed LLM response in `LLMToolSelectorMiddleware`

thromel:
## Summary

- Add validation in `_process_selection_response` to check that the LLM response contains a valid `tools` list before processing
- Raise a clear `ValueError` with actionable guidance instead of cryptic `KeyError: 'tools'`
- Add unit tests for malformed response handling (missing key and null value cases)

## Root Cause

When using `with_structured_output()`, some models may not strictly follow the schema and could return responses without the expected `tools` key. The previous code directly accessed `response["tools"]` without validation, causing intermittent `KeyError` failures.

## Test Plan

- [x] Added `test_malformed_response_missing_tools_key` - verifies `ValueError` is raised when `tools` key is missing
- [x] Added `test_malformed_response_null_tools` - verifies `ValueError` is raised when `tools` is `None`
- [x] All 12 existing tests pass
- [x] Lint checks pass

Fixes #34358

### DIFF
--- a/libs/langchain_v1/langchain/agents/middleware/tool_selection.py
+++ b/libs/langchain_v1/langchain/agents/middleware/tool_selection.py
@@ -223,10 +223,20 @@ class LLMToolSelectorMiddleware(AgentMiddleware):
         request: ModelRequest,
     ) -> ModelRequest:
         """Process the selection response and return filtered `ModelRequest`."""
+        # Validate response structure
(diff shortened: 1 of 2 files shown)


ccurme (commented):
Thanks for this, commented on the issue. We should retry the structured model invocation and then raise an error with a message like this one.

PR #34377, opened 2025-12-15
fix(core): add `return_exceptions` type hinting

zmievsa:
Adds `@overload` type hints to `batch` and `abatch` methods so that type checkers correctly infer `list[Output]` when `return_exceptions=False` (default) and `list[Output | Exception]` when `return_exceptions=True`.

### Before

```python
# Type checker infers: list[Output]
results = runnable.batch(inputs)

# Type checker still infers: list[Output] (incorrect - could contain exceptions)
results = runnable.batch(inputs, return_exceptions=True)
for r in results:
    if isinstance(r, Exception):  # Type checker warns: condition always false
        handle_error(r)
```

### After

```python
# Type checker infers: list[Output]
results = runnable.batch(inputs)

# Type checker correctly infers: list[Output | Exception]
results = runnable.batch(inputs, return_exceptions=True)
for r in results:
    if isinstance(r, Exception):  # No warning, type narrowing works correctly
        handle_error(r)
```

### DIFF
--- a/libs/core/langchain_core/runnables/base.py
+++ b/libs/core/langchain_core/runnables/base.py
@@ -997,6 +997,7 @@ class Runnable(ABC, Generic[Input, Output]):
                 for future in futures:
                     future.cancel()
 
+    @overload
     async def abatch(
         self,
         inputs: list[Input],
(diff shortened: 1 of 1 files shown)


zmievsa:
Closing because I am now realizing that langchain's design is fundamentally not type safe here. I understand now why only the iterator return option has the correct type hints. Ah :(

mdrxy:
@zmievsa I appreciate the effort; it's true that this is a bit involved. I've saved this issue to our internal backlog of typing changes we'd like to make for future releases.

PR #34379, opened 2025-12-16
fix(openai): handle function_call content in token counting

CyberNinjaSaurav:


CyberNinjaSaurav:
Fixes a crash in get_num_tokens_from_messages when using the OpenAI Responses API.

libs/partners/openai/tests/unit_tests/test_token_counter.py
```diff
@@ -0,0 +1,22 @@
+def test_token_counter_handles_function_call():
+    from langchain_openai import ChatOpenAI
+    from langchain_core.messages import HumanMessage
```
mdrxy: Should be at top-level

mdrxy:
Could you please edit the PR body to provide a more comprehensive description? Also, please address failing CI checks.

libs/partners/openai/tests/unit_tests/test_token_counter.py
```diff
@@ -0,0 +1,23 @@
...
+def test_token_counter_handles_function_call():
+
+
+    llm = ChatOpenAI(model="gpt-4o", use_responses_api=True)
+
+    messages = [
+        HumanMessage(
+            content=[
+                {
+                    "type": "function_call",
```
ccurme: this looks like chat completions format, not Responses?

ccurme (approved):
I think this is already supported, I've updated the PR to reflect this with a test, but let me know if I've misunderstood.

The original changes were adding a legacy chat completions function call.

Responses [function calls](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_function_tool_call.py) are formatted differently and should already be accounted for, in that they are reflected in the AIMessage `tool_calls`, and those are translated into chat completions format in `_convert_message_to_dict`.

PR #34396, opened 2025-12-16
fix(openai): filter function_call blocks in token counting

sujalmh:
The Responses API returns messages with `function_call`  content blocks which cause `get_num_tokens_from_messages` to raise a `ValueError: Unrecognized content block type.` This occurs because the token counter internally uses the chat/completions format logic, which does not recognize these blocks.

This PR updates `_format_message_content` to filter out `function_call` blocks when formatting messages for the chat/completions API (which includes the token counter). These blocks are redundant for token counting as the actual tool calls are already handled via the separate `tool_calls` field, ensuring accurate counts without errors.

Fixes #34378

Breaking Changes: None.


libs/partners/openai/langchain_openai/chat_models/base.py
```diff
@@ -237,7 +237,10 @@ def _format_message_content(
             if (
                 isinstance(block, dict)
                 and "type" in block
-                and block["type"] in ("tool_use", "thinking", "reasoning_content")
+                and (
+                    block["type"] in ("tool_use", "thinking", "reasoning_content")
+                    or (block["type"] == "function_call" and api == "chat/completions")
```
sargismarkosyan: Could you please include the 'code_interpreter_call' type as well?

sujalmh: Thanks for the feedback!

I've added `code_interpreter_call` to the filtered types as requested.

PR #34399, opened 2025-12-17
fix(core): populate default args from tool's `args_schema`

nathannewyen:
## Summary
- Fixes issue where Pydantic default values from `args_schema` were not passed to tool functions when the caller omits optional arguments
- Modified `_parse_input()` in `libs/core/langchain_core/tools/base.py` to include fields with non-None defaults
- Added unit tests to verify default args behavior for both sync and async tools

## Problem
When a tool has an `args_schema` with default values:
```python
class SearchArgs(BaseModel):
    query: str = Field(..., description="Search query")
    page: int = Field(default=1, description="Page number")
    size: int = Field(default=10, description="Results per page")

@tool("search", args_schema=SearchArgs)
def search_tool(query: str, page: int, size: int) -> str:
    return f"query={query}, page={page}, size={size}"

# This threw: TypeError: search_tool() missing 2 required positional arguments
search_tool.invoke({"query": "test"})
```

The defaults from `args_schema` were being discarded because `_parse_input()` filtered validated results to only include keys from the original input.

## Solution
Changed the filtering logic to:
1. Include all fields that were in the original input (validated)
2. Also include fields with non-None defaults from the Pydantic schema

This applies user-defined defaults (like `Field(default=1)`) while excluding synthetic fields from `*args`/`**kwargs` which have `default=None`.

## Test plan
- [x] Added `test_tool_args_schema_default_values` - tests sync tool with defaults
- [x] Added `test_tool_args_schema_default_values_async` - tests async tool with defaults
- [x] All existing tests pass (150 passed, 4 skipped)
- [x] Lint passes

Fixes #34384


nathannewyen:
The CI failure in `test_startup_command_failure` is unrelated to this PR. That test is in `libs/langchain_v1/tests/unit_tests/agents/middleware/implementations/test_shell_tool.py` which tests shell tool middleware - completely separate from our changes in `libs/core/langchain_core/tools/base.py`.

The test fails because `exit 1` terminates the shell process, then when trying to write a marker to stdin, it gets `BrokenPipeError` instead of the expected `RuntimeError`. This appears to be a flaky test or pre-existing issue in the shell middleware error handling.

Could a maintainer please re-run the CI or mark this as a known flaky test?

mdrxy (dismissed):
Thank you!

Issue #34398, opened 2025-12-17
Add BuiltinToolsMiddleWare to support native llm builtin tools like web_search, code_execution, file_content, etc.


tetv:
### Checked other resources

- [x] This is a feature request, not a bug report or usage question.
- [x] I added a clear and descriptive title that summarizes the feature request.
- [x] I used the GitHub search to find a similar feature request and didn't find it.
- [x] I checked the LangChain documentation and API reference to see if this feature already exists.
- [x] This is not related to the langchain-community package.

### Package (Required)

- [x] langchain
- [ ] langchain-openai
- [ ] langchain-anthropic
- [ ] langchain-classic
- [ ] langchain-core
- [ ] langchain-cli
- [ ] langchain-model-profiles
- [ ] langchain-tests
- [ ] langchain-text-splitters
- [ ] langchain-chroma
- [ ] langchain-deepseek
- [ ] langchain-exa
- [ ] langchain-fireworks
- [ ] langchain-groq
- [ ] langchain-huggingface
- [ ] langchain-mistralai
- [ ] langchain-nomic
- [ ] langchain-ollama
- [ ] langchain-perplexity
- [ ] langchain-prompty
- [ ] langchain-qdrant
- [ ] langchain-xai
- [ ] Other / not sure / general

### Feature Description

LLMs providers now supports several builtin tools, such as web_search, code_interpreter, etc.
However each one has a way to setup then. If we could have a middleware that adds this based on the LLM automatically, it would be easier to switch LLMs provider (for example using init_chat_model.

E.g.:

antropic:
 - {"type": "web_search_20250305", "name": "web_search", "max_uses": 100}
 - {"type": "code_execution_20250825", "name": "code_execution"}

openai:
- {"type": "web_search"}
- {"type": "code_interpreter", "container": {"type": "auto"}}

google_genai:
  - {"google_search": {}}
  - {"code_execution": {}}

etc.

### Use Case

LLMs providers now supports several builtin tools, such as web_search, code_interpreter, etc.
However each one has a way to setup. If we could have a middleware that adds this based on the LLM automatically, it would be easier to switch LLMs provider.

### Proposed Solution

BuiltinToolsMiddleWare(include_tools: ["web_search", "code_execution", ...])

**Note:**
Not all LLMs support the same builtin tools.
For example Antropic supports web_fetch, but Openain doesn't.
For those cases we could have another configuration to remove/error in the __init__ for unsupported_behaviour.

[...]

PR #34401, opened 2025-12-17
feat(langchain): Added BuiltinToolsMiddleware

keenborder786:
- Added `BuiltinToolsMiddleware` which allows to add builtin tools depending on the provider and tool selected.
- #34398


tetv:
for xai / grok you can use:

{ "type": "web_search" }
{ "type": "code_interpreter" }

for xai / grok, there is also do x_search:



https://docs.x.ai/docs/guides/tools/search-tools

ccurme (commented):
I think instead of langchain middleware keeping track of provider tool definitions, where multiple providers support similar tools (web search, code execution) we should implement standard types for these in langchain-core, then translate appropriately in the respective integration.

this is requires designing schemas that cover common provider functionality + support provider-specific fields, e.g.,
```python
class UserLocation(TypedDict, total=False):
    type: Required[Literal["approximate"]]

    city: Optional[str]
    """The city of the user."""

    country: Optional[str]
    """
    The two letter
    [ISO country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) of the
    user.
    """

    region: Optional[str]
    """The region of the user."""

    timezone: Optional[str]
    """The [IANA timezone](https://nodatime.org/TimeZones) of the user."""


class WebSearchTool(TypedDict):

    type: Literal["web_search"]

    user_location: NotRequired[UserLocation]
    """Parameters for the user's location.

    Used to provide more relevant search results.

    Currently only used by OpenAI and Anthropic.
    """
```

keenborder786:
@ccurme that's a good idea.

PR #34402, opened 2025-12-17
fix(core): Add config isolation utilities to prevent middleware streaming leaks

lwtaiyty:
## Description

This PR adds `isolate_config()` and `aisolate_config()` context managers to prevent model calls in middleware from unexpectedly inheriting parent streaming callbacks.

### Problem

When making language model calls (e.g., `model.ainvoke()`) inside agent middleware hooks like `abefore_model()`, the model tokens unexpectedly stream to the parent agent's output. 

This happens because LangChain's `ensure_config()` function automatically inherits context variables from `var_child_runnable_config` (defined in `langchain_core/runnables/config.py`), which includes streaming callback handlers. When an agent is invoked with `astream()`, any model calls in middleware inherit these streaming callbacks, causing their output to leak into the agent stream.

**Reproduction:**

```python
class TestMiddleware(AgentMiddleware):
    def __init__(self):
        self.model = ChatOpenAI()
    
    async def abefore_model(self, state, runtime):
        # This leaks into the parent stream ❌
        response = await self.model.ainvoke("What's the capital of France?")
        return None

agent = create_agent(model=ChatOpenAI(), middleware=[TestMiddleware()])
async for event in agent.astream({"messages": [...]}, stream_mode=["messages"]):
    # Unexpectedly receives tokens from TestMiddleware's model call
    print(event)

### DIFF
--- a/libs/core/langchain_core/runnables/__init__.py
+++ b/libs/core/langchain_core/runnables/__init__.py
@@ -40,6 +40,8 @@ if TYPE_CHECKING:
         get_config_list,
         patch_config,
         run_in_executor,
+        isolate_config,
+        aisolate_config,
     )
     from langchain_core.runnables.fallbacks import RunnableWithFallbacks
     from langchain_core.runnables.history import RunnableWithMessageHistory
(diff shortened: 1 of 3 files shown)


mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #34403, opened 2025-12-17
fix(langchain): add reducer to PlanningState.todos for concurrent updates

### LINKED ISSUE #96: INVALID_CONCURRENT_GRAPH_UPDATE Error for Todos State
### Problem
The deepagents library was encountering a INVALID_CONCURRENT_GRAPH_UPDATE error from LangGraph when multiple nodes in the agent's graph attempted to update the [todos] state key simultaneously. This error occurs because LangGraph's default behavior doesn't know how to resolve conflicts when parallel processes (e.g., tool calls) try to modify the same state property in the same step.

Error Details:

Error Message: At key 'todos': Can receive only one value per step. Use an Annotated key to handle multiple values.
Root Cause: The [todos field in [DeepAgentState] was defined as [NotRequired[list[Todo]]], which doesn't specify a reducer function. When multiple tools or nodes (e.g., in a fan-out execution) update the [todos] list concurrently, LangGraph raises this error to prevent ambiguous state resolution.
Impact: This prevented the agent from functioning correctly in scenarios involving parallel tool invocations, such as when the AI oncall bot processes multiple tasks or tools simultaneously.
Solution
To resolve this, I modified the [DeepAgentState] class in [state.py] to use an annotated type with a reducer:

Added Import: Imported the [operator] module to enable the use of [operator.add]  as a reducer.
Updated State Definition: Changed [todos: NotRequired[list[Todo]]] to [todos: Annotated[list[Todo], operator.add]]. This tells LangGraph to merge concurrent updates to the [todos] list by appending them (using [operator.add], rather than overwriting or erroring out.
Changes Made:

[state.py]
Added: [import operator]
Modified: [todos: Annotated[list[Todo], operator.add]]
Why This Fix Works
[operator.add] for lists performs concatenation, ensuring that multiple updates to [todos] are combined into a single, coherent list.
This aligns with LangGraph's recommended approach for handling concurrent state updates, as documented in their troubleshooting guide.
The fix is minimal and targeted, preserving the existing behavior while enabling parallel execution.

riccardo-algorime:
## Summary

The `todos` field in `PlanningState` was missing a reducer, causing `INVALID_CONCURRENT_GRAPH_UPDATE` errors when multiple nodes attempt to update the state simultaneously during parallel tool invocations.

**Changes:**
- Added `_todos_reducer` function with last-write-wins semantics
- Updated `PlanningState.todos` annotation to include the reducer
- Added unit tests for the reducer function

## Root Cause

LangGraph requires annotated state fields to have reducers when they may receive concurrent updates. The `todos` field lacked this, triggering:

```
At key 'todos': Can receive only one value per step. Use an Annotated key to handle multiple values.
```

## Solution

Since `write_todos` replaces the entire todo list (not appends), the reducer implements last-write-wins semantics:

```python
def _todos_reducer(_current: list[Todo] | None, new: list[Todo]) -> list[Todo]:
    return new
```

## Test Plan

- [x] Added unit tests for `_todos_reducer`
- [x] All existing tests pass
- [x] Linting passes
- [x] Type checking passes

Fixes langchain-ai/deepagents#96

### DIFF
--- a/libs/langchain_v1/langchain/agents/middleware/todo.py
+++ b/libs/langchain_v1/langchain/agents/middleware/todo.py
@@ -33,10 +33,27 @@ class Todo(TypedDict):
     """The current status of the todo item."""
 
 
+def _todos_reducer(_current: list[Todo] | None, new: list[Todo]) -> list[Todo]:
+    """Reducer for todos state - last write wins.
+
+    This reducer is required by LangGraph to handle concurrent state updates.
+    Since write_todos replaces the entire todo list rather than appending,
+    we use last-write-wins semantics.
+
+    Args:
+        _current: The current todo list (unused - last write wins).
+        new: The new todo list to use.
+
+    Returns:
+        The new todo list.
+    """
+    return new
+
+
 class PlanningState(AgentState):
     """State schema for the todo middleware."""
 
-    todos: Annotated[NotRequired[list[Todo]], OmitFromInput]
+    todos: NotRequired[Annotated[list[Todo], _todos_reducer, OmitFromInput]]
     """List of todo items for tracking task progress."""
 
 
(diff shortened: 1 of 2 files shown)


riccardo-algorime:
hello?

eyurtsev (changes requested):
It's unclear that this is a correct solution / design

riccardo-algorime:
What Is unclear?
Deepagenta framework doesnt work at all with most models becouse of this,expecislly gemini

PR #34435, opened 2025-12-20
style(langchain): add ruff rules ARG

cbornet:
In this order:
* used `@override` when overriding a parent method.
* prefixed param with `_` when the param could be renamed.
* used `*_args, **_kwargs` when it was not possible to rename (eg: protocol)
* used `_ = some_variable` when the variable name is inspected (in tools)


libs/langchain_v1/tests/unit_tests/agents/test_system_message.py
```diff
@@ -754,25 +754,19 @@ def dynamic_system_prompt_middleware(request: ModelRequest) -> SystemMessage:
...
 
-        def metadata_middleware(request: ModelRequest) -> SystemMessage:
-            """Return SystemMessage with metadata."""
-            return SystemMessage(
-                content="You are a helpful assistant",
-                additional_kwargs={"temperature": 0.7, "model": "gpt-4"},
-                response_metadata={"region": "us-east"},
-            )
-
-        request = _make_request()
```
cbornet, line 808: There was no use creating the request.
It's only testing `SystemMessage` in reality.

cbornet:
Hi, coud this be reviewed ?

mdrxy (approved):

PR #34451, opened 2025-12-22
docs(deepseek): clarify model names (V3, R1) in docstring

kelvinvelasquez-SDE:
## Description

Improves documentation for the DeepSeek integration by clarifying which model names map to which DeepSeek releases.

## Changes

- **Modified `chat_models.py`**: Updated docstring to clarify that `'deepseek-chat'` uses DeepSeek-V3 and `'deepseek-reasoner'` uses DeepSeek-R1.

## Context

This is a documentation-only change. Initially, this PR attempted to add a `deepseek-v3` profile, but after reviewer feedback, I learned that:
1. `deepseek-v3` is not a valid API model ID (the correct ID is `deepseek-chat`)
2. Profile data is auto-generated from models.dev and shouldn't be manually edited

The PR was updated to focus solely on improving the docstring for user clarity.

### DIFF
--- a/libs/partners/deepseek/langchain_deepseek/chat_models.py
+++ b/libs/partners/deepseek/langchain_deepseek/chat_models.py
@@ -56,7 +56,8 @@ class ChatDeepSeek(BaseChatOpenAI):
 
     Key init args — completion params:
         model:
-            Name of DeepSeek model to use, e.g. `'deepseek-chat'`.
+            Name of DeepSeek model to use, e.g. "deepseek-chat", "deepseek-v3",
+            or "deepseek-reasoner".
         temperature:
             Sampling temperature.
         max_tokens:
--- a/libs/partners/deepseek/langchain_deepseek/chat_models.py
+++ b/libs/partners/deepseek/langchain_deepseek/chat_models.py
@@ -57,7 +57,7 @@ class ChatDeepSeek(BaseChatOpenAI):
     Key init args — completion params:
         model:
             Name of DeepSeek model to use, e.g. "deepseek-chat", "deepseek-v3",
-            or "deepseek-reasoner".
+            or `'deepseek-reasoner`'.
         temperature:
             Sampling temperature.
         max_tokens:
--- a/libs/partners/deepseek/langchain_deepseek/chat_models.py
+++ b/libs/partners/deepseek/langchain_deepseek/chat_models.py
@@ -56,7 +56,7 @@ class ChatDeepSeek(BaseChatOpenAI):
 
     Key init args — completion params:
         model:
-            Name of DeepSeek model to use, e.g. "deepseek-chat", "deepseek-v3",
+            Name of DeepSeek model to use, e.g. `'deepseek-chat'`, `'deepseek-v3'`,
             or `'deepseek-reasoner`'.
         temperature:
             Sampling temperature.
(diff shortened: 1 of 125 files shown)


libs/partners/deepseek/langchain_deepseek/chat_models.py
```diff
@@ -56,7 +56,8 @@ class ChatDeepSeek(BaseChatOpenAI):
 
     Key init args — completion params:
         model:
-            Name of DeepSeek model to use, e.g. `'deepseek-chat'`.
+            Name of DeepSeek model to use, e.g. "deepseek-chat", "deepseek-v3",
```
mdrxy:
```suggestion
            Name of DeepSeek model to use, e.g. `'deepseek-chat'`, `'deepseek-v3'`,
```

libs/partners/deepseek/langchain_deepseek/chat_models.py
```diff
@@ -56,7 +56,8 @@ class ChatDeepSeek(BaseChatOpenAI):
...
+            or "deepseek-reasoner".
```
mdrxy:
```suggestion
            or `'deepseek-reasoner`'.
```

mdrxy:
It would appear this model doesn't exist upstream at https://models.dev/ under the DeepSeek provider. Making the change here is not suitable since it would be overridden the next time we generate from `libs/model-profiles`. Would you mind opening a PR in the upstream models.dev project?

kelvinvelasquez-SDE:
Thanks for the feedback @mdrxy! You are right â€” I will open a PR upstream at models.dev to add the DeepSeek-V3 profile there first. Once that is merged, this PR can either be updated or closed as needed. I will link the upstream PR here once it is ready.

kelvinvelasquez-SDE:
After investigating further, I realize that `deepseek-v3` is not actually a valid model ID in DeepSeek's API. The official model names are `deepseek-chat` (which is DeepSeek-V3 internally since Dec 2024) and `deepseek-reasoner` (DeepSeek-R1).

Looking at models.dev, the existing `deepseek-chat.toml` already represents V3 (release_date: 2024-12-26 matches V3's launch).

Given this, I think the best approach would be to:
1. Remove the `_profiles.py` changes (since `deepseek-v3` isn't a real API endpoint)
2. Keep only the docstring clarification mentioning that `deepseek-chat` uses DeepSeek-V3 under the hood

Would this revised scope be acceptable? Or should I close this PR entirely?

I appreciate your guidance on the correct architecture â€” it helped me understand how langchain integrates with upstream model profiles.

kelvinvelasquez-SDE:
I've updated the PR based on the feedback:

**Changes made:**
- âœ… Removed the `deepseek-v3` profile from `_profiles.py` (since it's not a valid API model ID and would be overwritten by auto-generation)
- âœ… Removed the corresponding unit test
- âœ… Improved the docstring to clarify that `deepseek-chat` uses DeepSeek-V3 under the hood, and `deepseek-reasoner` uses DeepSeek-R1

The PR now focuses solely on documentation improvement, which should be safe and helpful for users who may not know which model name maps to which DeepSeek release.

Ready for re-review when you have a moment. Thanks again for the guidance! ðŸ™

kelvinvelasquez-SDE:
Note: The CodSpeed Benchmarks check is failing, but this appears to be an infrastructure issue rather than a code problem â€” this PR only modifies documentation (docstring), so there is no code to benchmark. All other checks (lint, tests, pydantic compatibility) are passing. âœ…

mdrxy:
I think it may be best to leave this ambiguous so as not to expand the surface area of what is dependent on upstream sources to maintain

PR #34461, opened 2025-12-23
fix(core): URL-encode bgColor in mermaid API calls

nathannewyen:
URL-encode the bgColor parameter to fix 400 errors from mermaid.ink API.

The `!` character in `!white` was not encoded, causing API failures.

Fixes #34444


mdrxy (approved):
Thanks! My only ask would be that unit tests get added, but this can come later. Ideally we unit test to ensure URL encodings are done properly for all commands.

PR #33852, opened 2025-11-06
fix(core): implement ChatPromptTemplate.save by delegating to BasePromptTemplate

antznette1:
- **Implemented** 
`ChatPromptTemplate.save` by delegating to `BasePromptTemplate.save`, matching existing patterns in other prompt templates, enabling serialization to JSON/YAML consistent with other prompt templates.

- **Aligns** 
`ChatPromptTemplate` with `FewShotPromptTemplate`’s pattern.

- **Changes**
`libs/core/langchain_core/prompts/chat.py`
    - Implemented `save(self, file_path)` to call `super().save(file_path)`.

- **Behavior**
    - No breaking changes.
    - New behavior: 
`ChatPromptTemplate.save("path/to/file.yaml")` now works as expected.

- **Test**
    - “Added a roundtrip unit test for ChatPromptTemplate.save to JSON/YAML.”

- **Issue**
Fixes #32637

PR #32654, opened 2025-08-22
fix(core): implement `ChatPromptTemplate.save()` method

keyuchen21:
## Summary
Fixes #32637 - ChatPromptTemplate.save() method now properly saves templates to files instead of raising NotImplementedError.

## Problem
The `save()` method in `ChatPromptTemplate` was not implemented and simply raised `NotImplementedError`, preventing users from saving their chat prompt templates to files.

## Solution
- Implemented a working `save()` method that uses `dumpd()` for proper serialization
- The method supports both JSON and YAML formats
- Ensures saved templates can be correctly loaded back using the `load()` function

## Changes
1. **Implemented ChatPromptTemplate.save() method**:
   - Uses `dumpd()` from `langchain_core.load` for proper serialization
   - Supports JSON and YAML file formats
   - Creates parent directories if they don't exist
   - Validates file extensions and raises appropriate errors

2. **Added comprehensive unit tests**:
   - Test saving/loading to JSON format
   - Test saving/loading to YAML format  
   - Test with MessagesPlaceholder
   - Test error handling for invalid file extensions
   - Test error handling for partial variables
   - Test directory creation

## Test plan
Added unit tests that verify:
- Templates can be saved to JSON and loaded back correctly
- Templates can be saved to YAML and loaded back correctly
- Proper error handling for invalid file extensions
- Proper error handling when partial variables are present
- Parent directories are created automatically when saving to nested paths
- Templates with MessagesPlaceholder work correctly

All tests pass successfully.

## Example Usage
```python
from langchain_core.prompts import ChatPromptTemplate
from langchain_core.load import load

# Create a chat prompt template
template = ChatPromptTemplate.from_messages([
    ("system", "You are a helpful assistant named {assistant_name}."),
    ("human", "{question}")
])

# Save to file
template.save("my_prompt.json")

# Load it back
with open("my_prompt.json") as f:
    import json
    loaded_template = load(json.load(f))

# Use the loaded template
messages = loaded_template.format_messages(
    assistant_name="Claude",
    question="What can you help me with?"
)
```

PR #34465, opened 2025-12-23
fix(core): implement `ChatPromptTemplate.save`

pallaprolus:
## Description
Implements the `save` method for `ChatPromptTemplate` which previously raised `NotImplementedError`. 
This fixes issue #32637.

## Implementation Details
- Implemented `save` in `libs/core/langchain_core/prompts/chat.py` using `langchain_core.load.dumpd` for robust serialization of complex chat messages.
- Updated `libs/core/langchain_core/prompts/loading.py` to support loading prompts saved with the `lc` (LangChain serialization) format, ensuring round-trip compatibility. 
  - This addresses the "round-trip" issues raised in previous attempts (e.g., #33631).

## Verification
- Added a regression test `test_chat_prompt_template_save_load` in `tests/unit_tests/prompts/test_chat.py`.
- Verified that `load_prompt` can correctly deserialize the saved JSON file.

## Related Issues
- Fixes #32637
- Supersedes #33631, #32654, #33852


mdrxy:
Can you describe why this should supersede #33631?

pallaprolus:
Thanks for the question, @mdrxy!

The main reason this PR should supersede #33631 is that it fixes the **round-trip compatibility** issue that stalled the previous PR.

Specifically, PR #33631's approach (falling back to the generic `BasePromptTemplate.save`) produces a serialization format that triggers a `KeyError: 'prompt'` when reloaded via `load_prompt`. Maintainer @ccurme demonstrated this with a failing test (commit `c637b70`).

This PR (#34465) resolves this by:
1. Using `langchain_core.load.dumpd` for robust serialization of chat messages into the `lc` format.
2. 2. Updating the loading logic in `prompts/loading.py` to correctly handle this format, ensuring that `save` and `load_prompt` work seamlessly together for `ChatPromptTemplate`.

ccurme (commented):
Thanks for this, and appreciate you following up with the comment on the other PR, but I've closed the underlying issue as these routines are old and we have better options for serialization now.

PR #34466, opened 2025-12-23
test(core): URL-encode bgColor parameter in mermaid.ink API calls

Mr-Neutr0n:
## Problem

The `draw_mermaid_png()` function fails with HTTP 400 when using named background colors like `white`. This is because named colors get prefixed with `!` (e.g., `!white`) but this special character is not URL-encoded before being added to the API URL.

As reported in #34444, the URL parameter `bgColor=!white` causes mermaid.ink to return a 400 Bad Request error.

## Solution

URL-encode the `background_color` parameter using `urllib.parse.quote()` before constructing the API URL. This ensures special characters like `!` are properly encoded as `%21`.

## Changes

- Added `import urllib.parse` 
- URL-encode `background_color` value with `urllib.parse.quote(str(background_color), safe="")`
- Added 2 unit tests:
  - `test_mermaid_bgcolor_url_encoding`: Verifies named colors are properly encoded
  - `test_mermaid_bgcolor_hex_not_encoded`: Verifies hex colors work correctly

## Testing

```bash
pytest tests/unit_tests/runnables/test_graph.py::test_mermaid_bgcolor_url_encoding -v
pytest tests/unit_tests/runnables/test_graph.py::test_mermaid_bgcolor_hex_not_encoded -v
```

Both tests pass.

Fixes #34444

---
*This contribution was made with AI assistance (Claude).*

### DIFF
--- a/libs/core/langchain_core/runnables/graph_mermaid.py
+++ b/libs/core/langchain_core/runnables/graph_mermaid.py
@@ -8,6 +8,7 @@ import random
 import re
 import string
 import time
+import urllib.parse
 from dataclasses import asdict
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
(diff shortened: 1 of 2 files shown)


mdrxy:
Related: #34461

PR #34468, opened 2025-12-23
test(langchain): enforce STTI-001 argument provenance invariant

joshuavetos:
This PR adds a small integration test that exposes a gap in how tool calls are validated today.

Right now, tool arguments are checked for schema correctness only. That means a model can pass a value that is type-valid but never came from any prior tool output in the session, and the tool still executes.

The test demonstrates this by allowing a delete call for an ID that was previously listed, and rejecting one that was never observed.

Nothing here depends on prompts, model behavior, or safety tuning. It’s just about whether tool arguments are required to come from earlier tool results.

The test fails under current behavior and is meant to make that gap explicit.

### DIFF
--- a/libs/langchain/tests/integration_tests/test_stti_001.py
+++ b/libs/langchain/tests/integration_tests/test_stti_001.py
@@ -1,0 +1,34 @@
+import pytest
+
+class ProvenanceError(Exception):
+    pass
+
+def execute(call):
+    return "executed"
+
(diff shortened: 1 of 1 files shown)


joshuavetos:
Closing this PR. This proposal requires design discussion first.
I am opening an issue to discuss the STTI-001 invariant at an architectural level.

PR #34470, opened 2025-12-24
fix(mistralai): ensure JSON serialization for `ToolMessage` content

VictorBravo9er:
---

### **Problem**

When using `langchain-mistralai` with MCP tools, tool outputs are propagated as `ToolMessage.content`.
For MCP servers, `message.content` is a **list of structured chunks** (e.g. `[{type: "text", text: "...", id: "..."}]`), not a plain string.

The Mistral Chat Completions API **strictly requires `tool.content` to be a string**.
Passing structured content results in a hard API failure:

```
HTTPStatusError: 422
Input should be a valid string
```

This happens even though:

* The MCP tool itself returns valid output
* LangChain correctly deserializes the tool response
* The failure occurs only at the Mistral API boundary

This is consistent with the documented limitation that **Mistral does not support JSON / structured tool message content**.

---

### **Root Cause**

In `langchain-mistralai`, `ToolMessage` objects are serialized as-is:

```python
if isinstance(message, ToolMessage):
    return {
        "role": "tool",
        "content": message.content,  # ❌ may be list[dict], not str
        "name": message.name,
        "tool_call_id": ...
    }
```

When `message.content` is a list of chunks, this violates Mistral’s schema and triggers a 422 error.

---

### **Fix**

Force serialization of tool message content into a valid string before sending it to Mistral:

```python
if isinstance(message, ToolMessage):
    return {
        "role": "tool",
        "content": json.dumps(message.content, ensure_ascii=False),
        "name": message.name,
        "tool_call_id": ...
    }
```

This guarantees:

* `tool.content` is always a string
* No schema violations at the API boundary
* Compatibility with MCP tools and structured outputs

---

### **Why this approach**

* Matches Mistral API requirements exactly
* Aligns with LangChain’s existing behavior for providers without structured tool message support
* Does **not** alter tool execution or MCP semantics
* Keeps the fix narrowly scoped to the Mistral adapter

---

### **Impact**

* Fixes 422 errors when using MCP tools with Mistral
* Unblocks agent + tool workflows without requiring user-side hacks
* No breaking changes for providers that already support structured content

---

### **Testing**

* Verified with `mcp-server-time` via `MultiServerMCPClient`
* Confirmed failure before change and success after serialization
* No regressions observed for non-tool messages

---


VictorBravo9er:
Please do mention if there is a better way of serializing the ToolMessage content, if any.

mdrxy:
Could you please add an integration test that fails without this fix?

VictorBravo9er:
> Could you please add an integration test that fails without this fix?

[test](https://github.com/VictorBravo9er/langchain-mistralai-errors.git)
Please do have a look.

VictorBravo9er:
> > Could you please add an integration test that fails without this fix?
> 
> https://github.com/VictorBravo9er/langchain-mistralai-errors.git please do have a look.

This langsmith trace should help as well.
[trace](https://smith.langchain.com/public/38983490-bad1-467c-bf65-0fb363b526b4/r)

PR #34472, opened 2025-12-24
feat(langchain): add AND/OR trigger logic to summarization middleware

osasisorae:
I added TriggerConfig support to SummarizationMiddleware so Python can mirror LangChain.js semantics: tuples still work, dicts require AND across metrics (e.g. {"tokens": 4000, "messages": 10}), and lists apply OR across clauses.

I refactored _should_summarize to evaluate normalized clauses (AND inside, OR across clauses) and documented the new syntax inline.

I wrote regression tests covering AND clauses, mixed clause lists, and the empty-mapping error to prove the new behavior.

Testing
I created a local venv, installed the editable LangChain packages + test deps, and ran test_summarization.py (all 36 tests pass; only the existing deprecation warning remains).

Fixes #34465.

### DIFF
--- a/libs/langchain_v1/langchain/agents/middleware/summarization.py
+++ b/libs/langchain_v1/langchain/agents/middleware/summarization.py
@@ -118,6 +118,12 @@ Example:
     ```
 """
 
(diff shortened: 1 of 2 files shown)


mdrxy:
Related #34576

PR #34476, opened 2025-12-24
style(core): add ruff rules TC

cbornet:
* Fixed a few TC
* Added a few Pydantic classes to `flake8-type-checking.runtime-evaluated-base-classes` (not as much as I would have imagined)
* Added a few `noqa: TC`
* Activated TC rules

### DIFF
--- a/libs/core/langchain_core/language_models/base.py
+++ b/libs/core/langchain_core/language_models/base.py
@@ -17,8 +17,8 @@ from typing import (
 from pydantic import BaseModel, ConfigDict, Field, field_validator
 from typing_extensions import TypedDict, override
 
-from langchain_core.caches import BaseCache
-from langchain_core.callbacks import Callbacks
+from langchain_core.caches import BaseCache  # noqa: TC001
+from langchain_core.callbacks import Callbacks  # noqa: TC001
 from langchain_core.globals import get_verbose
 from langchain_core.messages import (
     AIMessage,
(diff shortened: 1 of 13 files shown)


libs/core/langchain_core/language_models/base.py
```diff
@@ -113,7 +113,7 @@ def _get_verbosity() -> bool:
 
 
 class BaseLanguageModel(
-    RunnableSerializable[LanguageModelInput, LanguageModelOutputVar], ABC
+    RunnableSerializable[LanguageModelInput, LanguageModelOutputVar], BaseModel, ABC
```
cbornet: `RunnableSerializable` is in `runtime-evaluated-base-classes` but ruff doesn't seem to see it...
Also ruff could see that `BaseLanguageModel` itself is in `runtime-evaluated-base-classes`...

libs/core/langchain_core/language_models/base.py
```diff
@@ -17,8 +17,8 @@
 from pydantic import BaseModel, ConfigDict, Field, field_validator
 from typing_extensions import TypedDict, override
 
-from langchain_core.caches import BaseCache
-from langchain_core.callbacks import Callbacks
+from langchain_core.caches import BaseCache  # noqa: TC001
+from langchain_core.callbacks import Callbacks  # noqa: TC001
```
cbornet, line 21: Those are needed  by `BaseLanguageModel`
`RunnableSerializable` is in `runtime-evaluated-base-classes` but ruff doesn't seem to see it...
Also ruff could see that `BaseLanguageModel` itself is in `runtime-evaluated-base-classes`...

mdrxy (approved):
Thank you! Love small wins like these.

PR #34492, opened 2025-12-26
style(core): add ruff RUF012 rule

cbornet:


libs/core/tests/unit_tests/prompts/__snapshots__/test_chat.ambr
```diff
@@ -51,8 +51,6 @@
             'title': 'Id',
           }),
           'invalid_tool_calls': dict({
-            'default': list([
```
cbornet, line 54: NB: default is useless for type array in JSON Schema.

mdrxy (approved):

PR #34514, opened 2025-12-28
feat(langchain): Middleware Depends On Other Middlewares

keenborder786:
# Middleware Dependency Resolution - Simple Explanation

## The Problem
You have middleware that needs to run in a specific order. Some middleware depend on others.

## Simple Example

Let's say you have 3 middleware:

```
Middleware C → depends on → Middleware B
Middleware B → depends on → Middleware A
Middleware A → no dependencies
```

You give them to the system in **random order**: `[C, B, A]`

The system needs to figure out the **correct order**: `[A, B, C]`

## The Graph

```
Input Order:  C → B → A  (wrong order!)
              ↓   ↓   ↓
              
Dependencies: C needs B
              B needs A
              A needs nothing
              
              ↓   ↓   ↓
              
Correct Order: A → B → C  ✓
```

## Visual Dependency Graph

```
    A (no dependencies)
    ↓
    B (depends on A)
    ↓
    C (depends on B)
```

### Step 1: Find who has no dependencies
```
A ✓ (no dependencies - can run first!)
B ✗ (needs A)
C ✗ (needs B)
```

### Step 2: Process A, then unlock B
```
A ✓ (done)
B ✓ (A is done, so B can run now!)
C ✗ (still needs B)
```

### Step 3: Process B, then unlock C
```
A ✓ (done)
B ✓ (done)
C ✓ (B is done, so C can run now!)
```

## Final Result
**Ordered list: [A, B, C]** ✓

---

## Real Code Example

```python
# You provide (wrong order):
middleware = [
    MiddlewareC(),  # needs B
    MiddlewareB(),  # needs A  
    MiddlewareA(),  # needs nothing
]

# System returns (correct order):
[
    MiddlewareA(),  # First
    MiddlewareB(),  # Second
    MiddlewareC(),  # Third
]
```

### Basic Dependency

```python
from langchain.agents.middleware import ModelRetryMiddleware

class MyMiddleware(AgentMiddleware):
    def __init__(self):
        super().__init__(depends_on=(ModelRetryMiddleware,))

# ModelRetryMiddleware will be auto-instantiated and run first
agent = create_agent(
    model="openai:gpt-4",
    middleware=[MyMiddleware()]
)
```

### Custom Instance Dependency

```python
# Use a pre-configured instance
retry = ModelRetryMiddleware(max_retries=5, backoff_factor=2.0)

class MyMiddleware(AgentMiddleware):
    def __init__(self, retry_instance):
        super().__init__(depends_on=(retry_instance,))

# Your custom retry configuration will be used
agent = create_agent(
    model="openai:gpt-4",
    middleware=[MyMiddleware(retry)]
)
```

### Chain of Dependencies

```python
# C depends on B, B depends on A
class MiddlewareA(AgentMiddleware):
    def __init__(self):
        super().__init__()  # No dependencies

class MiddlewareB(AgentMiddleware):
    def __init__(self):
        super().__init__(depends_on=(MiddlewareA,))

class MiddlewareC(AgentMiddleware):
    def __init__(self):
        super().__init__(depends_on=(MiddlewareB,))

# Provide in any order - system will sort them
agent = create_agent(
    model="openai:gpt-4",
    middleware=[
        MiddlewareC(),  # Will run third
        MiddlewareA(),  # Will run first
        MiddlewareB(),  # Will run second
    ]
)
```

### Circular Dependency Detection

```python
class MiddlewareA(AgentMiddleware):
    def __init__(self):
        super().__init__(depends_on=(MiddlewareB,))

class MiddlewareB(AgentMiddleware):
    def __init__(self):
        super().__init__(depends_on=(MiddlewareA,))

# This will raise: ValueError: Circular dependency detected in middleware
agent = create_agent(
    model="openai:gpt-4",
    middleware=[MiddlewareA(), MiddlewareB()]
)
```

### Failed Instantiation

If a dependency class cannot be instantiated with default parameters, it will be logged and skipped:

```python
class ProblematicMiddleware(AgentMiddleware):
    def __init__(self, required_param):  # No default value!
        super().__init__()
        self.param = required_param

class MyMiddleware(AgentMiddleware):
    def __init__(self):
        # This will fail silently and be skipped
        super().__init__(depends_on=(ProblematicMiddleware,))
```

**Solution**: Provide the instance explicitly:

```python
problematic = ProblematicMiddleware(required_param="value")
my_middleware = MyMiddleware()
my_middleware.depends_on = (problematic,)
```

Issue: #33885


keenborder786:
@sydney-runkle

jgabriellima:
I’ve been working on a very similar problem with langchain-middleware-stack (focused on declarative middleware composition for LangChain Deep Agents), so this PR immediately resonated.

The topological sort approach makes a lot of sense 👍

Just sharing a couple of things I ran into while dealing with this in practice:

- Having `depends_on` in the constructor can make things a bit rigid over time. It worked well at first, but became tricky when the same middleware needed slightly different positioning across environments or teams.

- Also, I learned the hard way to be careful with silent failures during dependency resolution. If something gets skipped quietly, it can lead to subtle issues that only show up at runtime.

Not blockers at all, just edge cases that showed up for me after the system grew a bit. Really nice direction overall, this is definitely a problem worth solving.

Happy to share more details or compare approaches if that’s useful I think there’s a lot of overlap here and probably room to align ideas.

PyPI: https://pypi.org/project/langchain-middleware-stack/
Doc: https://medium.com/@joaogabriellima/declarative-middleware-composition-for-deep-agent-stacks-c4b71ba843d2

mdrxy:
Thanks for the thoughtful contribution. The idea is interesting, but after reviewing it we don't think it's the right fit for the core library at this time.

We're not currently accepting new middleware into the `langchain` monorepo — the package is already very crowded with a large dependency surface, and each addition increases the maintenance burden for the core team.

That said, we'd love to see this continue as a standalone package. You can publish it to PyPI and users can install and use it directly with `create_agent()`:

```python
pip install YOUR-PACKAGE
```

```python
from YOUR_PACKAGE import YOURMODULE

agent = create_agent(
    model="anthropic:claude-opus-4-6",
    tools=[...],
    middleware=[
        YOURMODULE(
            key=value,
            ...
        ),
    ],
)
```

The middleware interface (`AgentMiddleware`) is public and stable — everything you need to build against is already exposed.

Our team is still working on finding the ideal way to recommend community integration packages to our users — if you have any feedback there, let me know.

PR #34515, opened 2025-12-28
fix(openai): improve error handling for null choices in OpenAI-compatible APIs

xiangnuans:
Fixes #32252

Improves error handling and diagnostic information when OpenAI-compatible APIs (like vLLM) return null choices. The error messages now include more context about the error type, message, and response structure to help users diagnose API compatibility issues.

**Changes:**
- Enhanced `_create_chat_result` to check for null choices before processing error fields
- Added detailed error messages that include error type, message, and response keys
- Added comprehensive test cases for null choices scenarios (with error field, without error field, and with string error)

**Note on uv.lock changes:**
The `uv.lock` file was updated as part of resolving merge conflicts with upstream master, not due to adding new dependencies.

**AI Disclaimer:**
This contribution was created with assistance from AI tools for code implementation and testing.


mdrxy:
Please see #34328

PR #34520, opened 2025-12-29
chore(core): fix some TC rules

cbornet:
Depends on https://github.com/langchain-ai/langchain/pull/36470

This makes mypy crash !!

PR #34521, opened 2025-12-29
chore(deps): bump mypy to v1.19 and ruff to v1.14

cbornet:
* Set mypy to >=1.19.1,<1.20
* Set ruff to >=0.14.10,<0.15


libs/cli/pyproject.toml
```diff
@@ -38,14 +38,16 @@ dev = [
     "pytest-watcher>=0.3.4,<1.0.0"
 ]
 lint = [
-    "ruff>=0.13.1,<0.14",
-    "mypy>=1.18.1,<1.19"
+    "ruff>=0.14.10,<0.15.0"
```
mdrxy, line 41: This should be noted as well

mdrxy:
Will this crash like #34520?

cbornet:
No, this one doesn't crash.

libs/cli/pyproject.toml
cbornet: I updated the title and description.

PR #34526, opened 2025-12-29
chore(partners): remove redundant TODO comments from pyproject.toml f…

weiii668:


mdrxy:
Why? These are rules that need to be completed.

PR #34530, opened 2025-12-29
chore(core): improve typing of Runnable `__or__`

cbornet:
`Runnable.__or__`, `Runnable.__ror__`, and their `RunnableSequence` and `StructuredPrompt` overrides previously erased composition types: the right-hand operand was typed `Runnable[Any, Other]`, so piping two runnables together always produced `RunnableSerializable[Input, Any]`. Type information was lost at every `|`, which is why chains so often needed a `chain: Runnable = ...` annotation just to recover usable inference.

This adds `@overload`s so the `Output` of one step flows into the `Input` of the next and the composed result carries the real `Output` type through. `Runnable[int, str] | Runnable[str, float]` now infers `RunnableSerializable[int, float]` instead of `[int, Any]`. `coerce_to_runnable` gains overloads so a `Mapping` resolves to `RunnableParallel` while everything else stays a `Runnable`. As a knock-on effect, dozens of now-unnecessary `: Runnable` annotations were dropped from the test suite.

Runtime behavior is unchanged — this is a typing-only change.

## Impact on type-checked code

Most users will simply get better inference. Two changes can require a small adjustment if you run a type checker (`mypy`, `pyright`):

### Stricter operand matching in `|`

The right-hand side of `|` is now typed `Runnable[Output, Other]` rather than `Runnable[Any, Other]`, so the right operand's declared **input** must match the left operand's **output**. This is more accurate, but it surfaces a common pattern that was previously silent: piping a step that outputs a plain `dict` into a step whose declared input is a more specific type (for example a `TypedDict`). It still works at runtime; the checker now reports an `[operator]` error.

If you hit this, narrow the boundary with a `cast` (or an explicit annotation):

```python
from typing import Any, cast

from langchain_core.runnables import Runnable

# upstream outputs a dict; downstream declares a narrower input type
chain = cast("Runnable[Any, MyInput]", upstream) | downstream
```

### `list` → `Sequence` on `RunnableEach` / `map()`

`Runnable.map()` and the `invoke` / `ainvoke` methods of `RunnableEach` now accept `Sequence[Input]` instead of `list[Input]`. Callers are unaffected — a `list` is a `Sequence`, and tuples or other sequences now type-check too. The only thing to adjust: if you **subclass** `RunnableEach` (or `RunnableEachBase`) and override these methods with a `list[...]` parameter, widen the annotation to `Sequence[...]` so the override stays compatible with the base signature.


libs/core/tests/unit_tests/runnables/test_runnable.py
```diff
@@ -529,7 +530,7 @@ def test_passthrough_assign_schema() -> None:
 
     invalid_seq_w_assign = (
         RunnablePassthrough.assign(context=itemgetter("question") | retriever)
-        | fake_llm
+        | fake_llm  # type: ignore[operator]
```
cbornet, line 533: I guess we are doing that on purpose

libs/core/langchain_core/runnables/base.py
```diff
@@ -613,14 +613,35 @@ def get_prompts(
...
     def __or__(
         self,
-        other: Runnable[Any, Other]
-        | Callable[[Iterator[Any]], Iterator[Other]]
-        | Callable[[AsyncIterator[Any]], AsyncIterator[Other]]
-        | Callable[[Any], Other]
-        | Mapping[str, Runnable[Any, Other] | Callable[[Any], Other] | Any],
-    ) -> RunnableSerializable[Input, Other]:
+        other: Callable[[Output], Runnable[Output, Other]]
+        | Callable[[Output], Awaitable[Runnable[Output, Other]]],
```
cbornet, line 628: This is needed for instance when `chain = input_map | router` (see `test_higher_order_lambda_runnable`)

and

```python
def router(params: dict[str, Any]) -> Runnable[dict[str, Any], str]:
```

libs/core/langchain_core/runnables/base.py
```diff
@@ -765,7 +801,7 @@ def as_bytes(x: Any) -> bytes:
         # Import locally to prevent circular import
         from langchain_core.runnables.passthrough import RunnablePick  # noqa: PLC0415
 
-        return self | RunnablePick(keys)
+        return self | RunnablePick(keys)  # type: ignore[operator, no-any-return]
```
cbornet, line 814: I think mypy is right: RunnablePick needs a `dict[str, Any]` as input so it cannot be piped to any `Runnable`.
Shouldn't it be implemented only for specific subclasses ?

mdrxy (approved):

PR #34536, opened 2025-12-30
fix(core): ensure tool callbacks fire when tools run inside `Runnable`

Vishal-Krishna-Kumar:
### Summary
Tools executed inside `RunnableLambda` did not emit `tool_start` and `tool_end`
callbacks when callbacks were provided via `RunnableConfig`.

This caused inconsistent tracing behavior where chain callbacks fired
but tool callbacks were silently dropped.

### Fix
- Ensure `BaseTool.run()` and `BaseTool.arun()` inherit callbacks, tags, and metadata
  from `RunnableConfig` when not explicitly provided.

### Tests
- Added regression test `test_tool_used_as_runnable_emits_chain_and_tool_events`
- Test fails on `master` and passes with this change

### Breaking Changes
None


libs/core/langchain_core/tools/base.py
```diff
@@ -893,6 +893,15 @@ def run(
         Raises:
             ToolException: If an error occurs during tool execution.
         """
+        #  FIX: inherit callbacks/tags/metadata from Runnable config
```
mdrxy, line 896:
```suggestion
```

libs/core/langchain_core/tools/base.py
```diff
@@ -893,6 +893,15 @@ def run(
...
+        config = ensure_config(config)
+        if callbacks is None:
+            callbacks = config.get("callbacks")
```
mdrxy:
```suggestion
        callbacks = config.get("callbacks") if callbacks is None else callbacks
```

mdrxy:
Could you resolve CI failures?

